PR #263, opened 2018-03-22
Remove more N3DS References

N00byKing:
Let me start this off by saying: **PLEASE, BE CAREFUL**.

This is probably very dirty, but because I have updated my Switch to 4.1.0 already (I love Odyssey sue me) I am unable to do extensive tests on it. I used SpaceNX and that one Tic-Tac-Toe Game as tests, which behave the same way after this commit.

Now, what this PR does is just remove some Citra-Specific stuff, nothing fancy, however I don't know whether or not they are still _used_ (hence the carefulness). I couldn't find any references in the code, but hey. I suck. So who knows? I guess you do so have fun out there

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -23,7 +23,6 @@
 namespace Memory {
 
 static std::array<u8, Memory::VRAM_SIZE> vram;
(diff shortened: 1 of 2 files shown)


N00byKing:
woop i guess it isnt

PR #271, opened 2018-03-23
Implement Pull #3528 from citra: use nvidia graphics automatically on laptops with optimus (with AMD support)

N00byKing:


lioncash:
Vague commit messages (the first two in particular).

N00byKing:
Changed commit names.

Dont merge yet doe

N00byKing:
Instead of on Discord, I think its fine If we can just clear the thing up here.

Now, what is the "thing"? I have two versions for this PR, they are on my Repo under branches 3528 and 3528_alt. 3528 is what made it into citra, 3528_alt is based off a closed PR (not merged) that was made after citra-emu/citra#3528, which is citra-emu/citra#3540.

Personally, I prefer 3540 as that one does not introduce code duplication, but with a different file name instead of optimus.cpp, as that term is nvidia specific (something along the lines of switchablegfx.cpp dunno, i suck at name-giving). One noticable difference is that my version of that PR (3540) has only one "extern" Statement. That is the case as I do not believe that splitting them up makes a difference, I might be wrong about that however.

I'd like to ask everyone to discuss which version to use here. If 3528 is chosen, you can go ahead and merge this PR. If 3528_alt is, then I will close this one and open another one.

Thank you for reading ~~even though my grammar sucks bad~~ !

bunnei:
I'm not sure the reasons why 3528 was chosen over 3540 on Citra. However, this seems fine to me ..

N00byKing:
@bunnei 3528 was made before i think, which is why 3540 was closed

N00byKing:
'kay, I can get behind that

PR #273, opened 2018-03-24
GPU: Added code to unswizzle textures and ported the surface viewer from citra

Subv:
Things left to do:

- [ ] Detect when textures are not swizzled (There seems to be a register for this)
- [ ] Decode the textures to a format supported by the host GPU
- [ ] Analyze the shaders to know which textures are actually used

### DIFF
--- a/src/video_core/debug_utils/debug_utils.cpp
+++ b/src/video_core/debug_utils/debug_utils.cpp
@@ -23,8 +23,6 @@
 
 namespace Tegra {
 
-std::shared_ptr<DebugContext> g_debug_context;
-
 void DebugContext::DoOnEvent(Event event, void* data) {
     {
         std::unique_lock<std::mutex> lock(breakpoint_mutex);
--- a/src/video_core/debug_utils/debug_utils.cpp
+++ b/src/video_core/debug_utils/debug_utils.cpp
@@ -1,0 +1,66 @@
+// Copyright 2014 Citra Emulator Project
+// Licensed under GPLv2
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <condition_variable>
+#include <cstdint>
+#include <cstring>
+#include <fstream>
+#include <map>
+#include <mutex>
+#include <string>
+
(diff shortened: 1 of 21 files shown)


src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <condition_variable>
+#include <cstdint>
+#include <cstring>
+#include <fstream>
+#include <map>
+#include <mutex>
+#include <stdexcept>
```
lioncash: As far as I can tell, nothing uses stdexcept

src/video_core/debug_utils/debug_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "common/common_types.h"
+#include "common/file_util.h"
+#include "common/logging/log.h"
+#include "common/math_util.h"
+#include "common/vector_math.h"
+#include "video_core/debug_utils/debug_utils.h"
+
+namespace Tegra {
+
+std::shared_ptr<DebugContext> g_debug_context;
```
lioncash: Can we avoid introducing globals?

Subv: This code was ported almost verbatim from citra.
The reason this is a global and not a member of GPU is that a reference to it is stored within each BreakpointObserver subclass, and these are constructed at startup before the GPU instance is even created.

lioncash: Is that truly the best way to do it though? Given this is introducing it, should we not try and do this in a better way that doesn't specifically rely on it being global before we build anything on top of it?

Like, yeah, it's from citra, but that doesn't mean it's necessarily a good idea to perform a verbatim copy. This just introduces the same mistakes made previously.

Hexagon12:
@Subv Can you please rebase this PR, since it has conflicts?

PR #276, opened 2018-03-24
Change Telemetry Names to yuzu and remove links to citra

N00byKing:


src/yuzu_cmd/default_ini.h
```diff
@@ -173,12 +173,12 @@ gdbstub_port=24689
...
 enable_telemetry =
 # Endpoint URL for submitting telemetry data
-telemetry_endpoint_url = https://services.citra-emu.org/api/telemetry
+telemetry_endpoint_url =
 # Endpoint URL to verify the username and token
-verify_endpoint_url = https://services.citra-emu.org/api/profile
-# Username and token for Citra Web Service
+verify_endpoint_url =
+# Username and token for yuzu Web Service
 # See https://services.citra-emu.org/ for more info
```
N00byKing: Unsure what to do with that Line, any ideas?

FearlessTobi: Maybe change it to Yuzu as placeholder, even if the link doesn't exist yet

jroweboy:
@chris062689 might want to look this over

chris062689:
There's a few things that are probably considered outside of the scope of this PR, but still important for the Web API. 
Things that I think should be done in Citra and then ported over.

- https://services.citra-emu.org/ is no longer the correct endpoint, we should be using https://api.citra-emu.org/telemetry and https://api.citra-emu.org/profile

(Change these to yuzu-emu.org for yuzu, obviously) 

- We decided that we would only store a single endpoint URL, and the rest would be hard coded. This is to reduce the amount of config settings that the users' won't need to touch.
 
In the config it would list `https://api.citra-emu.org` and the actual endpoint would be hardcoded.

Aside from that, this looks fine.
The endpoints for yuzu are listening, so when anyone wants to tackle that, they're up and available.

src/yuzu_cmd/default_ini.h
bunnei: eh this is fine, we'll change it later when the yuzu telemetry endpoint exists.

PR #252, opened 2018-03-19
 Implement Pull #3064 from citra: Clean all format warnings (Yuzu-specific format warnings cleared too)

N00byKing:
Last Issue of #40!

Rejoice!

PR #277, opened 2018-03-24
Clean more Warnings, fix yuzu-emu/yuzu#252

N00byKing:


src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -170,7 +170,7 @@ class IDirectory final : public ServiceFramework<IDirectory> {
         IPC::RequestParser rp{ctx};
         const u64 unk = rp.Pop<u64>();
 
-        LOG_DEBUG(Service_FS, "called, unk=0x%llx", unk);
+        LOG_DEBUG(Service_FS, "called, unk=0x%lx", unk);
```
MerryMage: `LOG_DEBUG(Service_FS, "called, unk=0x%" PRIx64, unk);`

N00byKing: I understand what you mean, but shouldnt it be 32?

MerryMage: is `unk` not a `u64`?

N00byKing:
its unsigned long not long long (thats where the warning came from)

src/core/hle/service/filesystem/fsp_srv.cpp
N00byKing: its unsigned long not long long (thats where the warning came from)

bunnei: @MerryMage is right - %lx will only print a 32-bit value. `unk` is a 64-bit value. Applies here and elsewhere - @N00byKing can you please clean this up? I noticed in other places we are now logging 64-bit values as %lx, and therefore it's only printing the lower 32 bits.

lioncash:
Can you please prefix your commits properly like how almost everyone else does it? i.e. prefixing the thing you're changing to the message.

e.g. If it's a general change to core, then `core: <thing done>` would be the message. If it's localized in some_file.cpp/.h, then it's `some_file: <thing done>`. It makes it easier to filter out (or find) commits in the git log (and provides context for changes without needing actually read the diff).

N00byKing:
@lioncash will do

src/core/hle/service/filesystem/fsp_srv.cpp
N00byKing: Addressed

N00byKing: The Logging Problems have been fixed, and I have changed the PR title to mention that

N00byKing:
Fuck. I have no idea what just happened. Gimme a sec

Hexagon12:
What's the status of this PR?

N00byKing:
@Hexagon12 *should* be fine

daniellimws:
Minor request here, could you update the logging to use the new macros? e.g. `NGLOG_INFO` instead of `LOG_INFO`.

N00byKing:
Tbh, while I thought about it, I think it's out of scope for this PR. *After* it has been merged, maybe

daniellimws:
Sure then

N00byKing:
By now this PR is so outdated conflicts are pretty much everywhere. Closing

PR #265, opened 2018-03-23
Tegra progress 2

bunnei:
There's lots going on in this space, review at your own peril. I don't really expect anyone to review this line by line, but for the most part, it's code taken from Citra.
At a really high level, this PR:
* Adds some of Citra's texture forwarding and framebuffer handling stuff
* Uses it for framebuffer rendering (although it's currently flushed every frame, so there is no net gains yet)
* Implements some stuff that I explicitly removed with #254 
* Misc. video_core cleanup
* Adds some Maxwell format decoding stuff

PR #254, opened 2018-03-20
Port Citra OpenGL rasterizer code

bunnei:
This ports over a lot of boilerplate OpenGL rasterizer code from Citra. This is based on the latest work by @Phanto-m (https://github.com/citra-emu/citra/pull/3499) with generated GLSL vertex shaders.

I chose to port this code over (rather than starting from complete scratch) because I don't think we have any graphics experts on the team, and Citra's approach to rendering is at least a good example - even if it's a lot simpler than what we will eventually need for Switch.

For the most part, any 3DS-specific implementations have been replaced with `UNIMPLEMENTED`. A lot of this will be implemented in subsequent iterations, and I'm sure a lot will go away. I'm sure there is also some 3DS-specific junk still in here. Please don't bother reviewing too closely, unless you see anything completely wrong.

General approach to how we will implement rendering in yuzu:
[%] Implement sufficient register/vertex/texture/shader decodings for Switch GPU
[%] Port over boilerplate OpenGL rasterizer code from Citra
[ ] Hook up vertex data with OpenGL rasterizer
[ ] Implement textures with OpenGL rasterizer
[ ] Add "fake" pass-through shaders for proof-of-concept rendering (e.g. for Puyo Puyo Tetris)
[ ] Implement framebuffer copies
[ ] Implement shader decompiler (to GLSL) for Maxwell vertex shaders
[ ] Implement shader decompiler (to GLSL) for Maxwell fragment shaders

PR #279, opened 2018-03-25
Tegra progress 3

bunnei:
Part three of my work to retrofit Citra's OpenGL renderer for use with Switch rendering (previous parts were #265 and #254). Now we're starting to make progress 👍 
This change gets the bits in place for very early triangle drawing (NOTE: No games will render yet by design, so it's not worth trying them with this!). As mentioned in prior changes, there is a lot going on in this space and it's tough to review - But this change starts adding real Switch-specific rendering code (as opposed to a lot of boilerplate, as the prior two changes were), so I'd appreciate more thorough reviews with this one. General summary of changes:
* Hookup passthrough vertex/fragment shaders (these will be replaced by shader decompiler soon)
* Decode game-provided vertex attributes into arrays and hook up to the above shaders
* Add support for copying (and swizzling) hardware rendered framebuffers back to guest RAM
* Add support for proper framebuffer caching and texture forwarding, thereby removing the hack to always flush/invalidate this region in memory
* Lots of refactoring and cleanup

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -22,6 +22,7 @@
 #include "video_core/renderer_opengl/gl_shader_gen.h"
 #include "video_core/renderer_opengl/renderer_opengl.h"
 
+using Maxwell = Tegra::Engines::Maxwell3D::Regs;
 using PixelFormat = SurfaceParams::PixelFormat;
 using SurfaceType = SurfaceParams::SurfaceType;
 
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -142,37 +142,55 @@ RasterizerOpenGL::~RasterizerOpenGL() {
     }
 }
 
-static constexpr std::array<GLenum, 4> vs_attrib_types{
-    GL_BYTE,          // VertexAttributeFormat::BYTE
-    GL_UNSIGNED_BYTE, // VertexAttributeFormat::UBYTE
-    GL_SHORT,         // VertexAttributeFormat::SHORT
-    GL_FLOAT          // VertexAttributeFormat::FLOAT
-};
-
 void RasterizerOpenGL::AnalyzeVertexArray(bool is_indexed) {
     const auto& regs = Core::System().GetInstance().GPU().Maxwell3D().regs;
-    const auto& vertex_attributes = regs.vertex_attrib_format;
 
     if (is_indexed) {
         UNREACHABLE();
     }
-    const u32 vertex_num = regs.vertex_buffer.count;
 
-    vs_input_size = 0;
-    u32 max_offset{};
-    for (const auto& attrib : vertex_attributes) {
-        if (max_offset >= attrib.offset) {
-            continue;
-        }
-        max_offset = attrib.offset;
-        vs_input_size = max_offset + attrib.SizeInBytes();
-    }
-    vs_input_size *= vertex_num;
+    // TODO(bunnei): Add support for 1+ vertex arrays
+    vs_input_size = regs.vertex_buffer.count * regs.vertex_array[0].stride;
 }
 
 void RasterizerOpenGL::SetupVertexArray(u8* array_ptr, GLintptr buffer_offset) {
     MICROPROFILE_SCOPE(OpenGL_VAO);
-    UNIMPLEMENTED();
+    const auto& regs = Core::System().GetInstance().GPU().Maxwell3D().regs;
+    const auto& memory_manager = Core::System().GetInstance().GPU().memory_manager;
+
+    state.draw.vertex_array = hw_vao.handle;
+    state.draw.vertex_buffer = stream_buffer->GetHandle();
+    state.Apply();
+
+    // TODO(bunnei): Add support for 1+ vertex arrays
+    const auto& vertex_array{regs.vertex_array[0]};
+    ASSERT_MSG(vertex_array.enable, "vertex array 0 is disabled?");
+    ASSERT_MSG(!vertex_array.divisor, "vertex array 0 divisor is unimplemented!");
+    for (unsigned index = 1; index < Maxwell::NumVertexArrays; ++index) {
+        ASSERT_MSG(!regs.vertex_array[index].enable, "vertex array %d is unimplemented!", index);
+    }
+
+    // Use the vertex array as-is, assumes that the data is formatted correctly for OpenGL.
+    // Enables the first 16 vertex attributes always, as we don't know which ones are actually used
+    // until shader time. Note, Tegra technically supports 32, but we're cappinig this to 16 for now
+    // to avoid OpenGL errors.
+    for (unsigned index = 0; index < 16; ++index) {
+        auto& attrib = regs.vertex_attrib_format[index];
+        glVertexAttribPointer(index, attrib.ComponentCount(), MaxwellToGL::VertexType(attrib),
+                              GL_FALSE, vertex_array.stride,
+                              reinterpret_cast<GLvoid*>(buffer_offset + attrib.offset));
+        glEnableVertexAttribArray(index);
+        hw_vao_enabled_attributes[index] = true;
+    }
+
+    // Copy vertex array data
+    const u32 data_size{vertex_array.stride * regs.vertex_buffer.count};
+    const VAddr data_addr{memory_manager->PhysicalToVirtualAddress(vertex_array.StartAddress())};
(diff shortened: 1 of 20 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -142,37 +142,55 @@ RasterizerOpenGL::~RasterizerOpenGL() {
...
+        ASSERT_MSG(!regs.vertex_array[index].enable, "vertex array %d is unimplemented!", index);
+    }
+
+    // Use the vertex array as-is, assumes that the data is formatted correctly for OpenGL.
+    // Enables the first 16 vertex attributes always, as we don't know which ones are actually used
+    // until shader time. Note, Tegra technically supports 32, but we're cappinig this to 16 for now
+    // to avoid OpenGL errors.
+    for (unsigned index = 0; index < 16; ++index) {
+        auto& attrib = regs.vertex_attrib_format[index];
+        glVertexAttribPointer(index, attrib.ComponentCount(), MaxwellToGL::VertexType(attrib),
```
Subv: Is there any attribute type that isn't a float? We might want to use glVertexAttribIPointer for those cases.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -142,37 +142,55 @@ RasterizerOpenGL::~RasterizerOpenGL() {
...
+                              GL_FALSE, vertex_array.stride,
```
Subv: Shouldn't this normalization parameter be set to GL_TRUE for certain types of attributes? (UNORM/NORM attribs)

src/video_core/renderer_opengl/gl_rasterizer.cpp
bunnei: Discussed via Discord. `glVertexAttribPointer` will upload all data as floats, which is sufficient for now. If we want integer data, we can use `glVertexAttribIPointer`... but that decision will be made when we're actually implementing shaders.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1348,5 +1398,33 @@ void RasterizerCacheOpenGL::UnregisterSurface(const Surface& surface) {
...
+    const u64 page_start = addr >> Memory::PAGE_BITS;
+    const u64 page_end = page_start + num_pages;
+
+    // Interval maps will erase segments if count reaches 0, so if delta is negative we have to
+    // subtract after iterating
+    const auto pages_interval = PageMap::interval_type::right_open(page_start, page_end);
+    if (delta > 0)
+        cached_pages.add({pages_interval, delta});
+
+    for (auto& pair : RangeFromInterval(cached_pages, pages_interval)) {
```
lioncash: should this be a const reference?

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1093,18 +1057,104 @@ SurfaceRect_Tuple RasterizerCacheOpenGL::GetSurfaceSubRect(const SurfaceParams&
...
+    bool using_color_fb, bool using_depth_fb, const MathUtil::Rectangle<s32>& viewport) {
+    const auto& regs = Core::System().GetInstance().GPU().Maxwell3D().regs;
+    const auto& memory_manager = Core::System().GetInstance().GPU().memory_manager;
+    const auto& config = regs.rt[0];
+
+    // TODO(bunnei): This is hard corded to use just the first render buffer
+    LOG_WARNING(Render_OpenGL, "hard-coded for render target 0!");
+
+    // update resolution_scale_factor and reset cache if changed
+    static u16 resolution_scale_factor = GetResolutionScaleFactor();
```
lioncash: Note that this is thread unsafe

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -138,47 +143,276 @@ RasterizerOpenGL::~RasterizerOpenGL() {
...
+    // Viewport can have negative offsets or larger dimensions than our framebuffer sub-rect. Enable
+    // scissor test to prevent drawing outside of the framebuffer region
+    state.scissor.enabled = true;
+    state.scissor.x = draw_rect.left;
+    state.scissor.y = draw_rect.bottom;
+    state.scissor.width = draw_rect.GetWidth();
+    state.scissor.height = draw_rect.GetHeight();
+    state.Apply();
+
+    // Draw the vertex batch
```
lioncash: Should this be put into its own function, like GetPrimitiveMode() or something?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -138,47 +143,276 @@ RasterizerOpenGL::~RasterizerOpenGL() {
...
+                                                  viewport_rect_unscaled.top * res_scale,
+                                              surfaces_rect.bottom, surfaces_rect.top)), // Top
+        static_cast<u32>(MathUtil::Clamp<s32>(static_cast<s32>(surfaces_rect.left) +
+                                                  viewport_rect_unscaled.right * res_scale,
+                                              surfaces_rect.left, surfaces_rect.right)), // Right
+        static_cast<u32>(MathUtil::Clamp<s32>(static_cast<s32>(surfaces_rect.bottom) +
+                                                  viewport_rect_unscaled.bottom * res_scale,
+                                              surfaces_rect.bottom, surfaces_rect.top))}; // Bottom
+
+    // Bind the framebuffer surfaces
```
lioncash: This seems like something that can go in its own function, since its relatively non-dependent on the surrounding code aside from the surfaces, which only act as inputs.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -138,47 +143,276 @@ RasterizerOpenGL::~RasterizerOpenGL() {
...
+            // clear stencil attachment
+            glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0, 0);
+        }
+    } else {
+        // clear both depth and stencil attachment
+        glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0,
+                               0);
+    }
+
+    // Sync the viewport
```
lioncash: Should this be put in its own SyncViewport function?

src/yuzu/debugger/graphics/graphics_surface.cpp
```diff
@@ -339,9 +339,9 @@ void GraphicsSurfaceWidget::OnUpdate() {
...
 
         surface_address = rt.Address();
-        surface_width = rt.horiz;
-        surface_height = rt.vert;
-        if (rt.format != 0) {
+        surface_width = rt.width;
+        surface_height = rt.height;
+        if (rt.format != Tegra::RenderTargetFormat::NONE) {
             surface_format =
                 ConvertToTextureFormat(static_cast<Tegra::RenderTargetFormat>(rt.format));
```
Subv: The cast should be removed

Subv:
Note: DrawScreens() must be moved to the inside of the if clause in SwapBuffers to prevent drawing from potentially deleted textures. We should only draw if there's a framebuffer to draw from, otherwise we should display the previous frame (which is the default behavior when you don't draw anything)

mailwl:
why Screens? we have the only one

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
bunnei: Ported from citra, will leave a TODO in the code explaining to fix this.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
bunnei: Addressed

src/video_core/renderer_opengl/gl_rasterizer.cpp
bunnei: Addressed

src/video_core/renderer_opengl/gl_rasterizer.cpp
bunnei: Addressed

src/yuzu/debugger/graphics/graphics_surface.cpp
bunnei: Addressed

bunnei:
🎈

PR #280, opened 2018-03-25
Minor changes to VI, PL, HID, and AUDREN

bunnei:
Just some local changes I've had for a while that should get merged, nothing particularly exciting.

### DIFF
--- a/src/core/hle/service/audio/audren_u.cpp
+++ b/src/core/hle/service/audio/audren_u.cpp
@@ -151,12 +151,52 @@ private:
     Kernel::SharedPtr<Kernel::Event> system_event;
 };
 
+class IAudioDevice final : public ServiceFramework<IAudioDevice> {
+public:
+    IAudioDevice() : ServiceFramework("IAudioDevice") {
+        static const FunctionInfo functions[] = {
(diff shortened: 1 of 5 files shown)


src/core/hle/service/vi/vi.cpp
```diff
@@ -27,8 +27,8 @@ struct DisplayInfo {
     char display_name[0x40]{"Default"};
     u64 unknown_1{1};
     u64 unknown_2{1};
-    u64 width{1920};
-    u64 height{1080};
+    u64 width{1280};
```
mailwl: is it hw-tested? on switchbrew: The width/height for the "Default" Display is the resolution for 1080p even when in handheld-mode.

bunnei: Hmm thanks... maybe this should be reverted. Everywhere else, we treat resolution as 1280x720 (handheld mode) so I think we thought this change made sense, but perhaps not...

Hexagon12:
Thank you very much for taking on IAudioDevice!

I tried doing it once, but there wasn't any documentation at that time and I had to give up.

PR #284, opened 2018-03-27
Add config for "Docked" mode and various settings cleanup

bunnei:
* Adds a config setting for "Docked" mode (versus Handheld).
* Defaults to Docked mode (Note: This indicates higher performance mode, which may make games render at 1080p as opposed to 720p).
* Leaves 1280 x 720 default window size as-is. In my opinion, until we actually have games rendering, the bigger window size is just inconvenient for those of us with lower resolution displays.
* Changes CPU setting dropdown to just "Enable CPU JIT" checkbox, which is what Citra does. This is simpler for users, and we'll never support 3 CPU backends. This will also move everyone to Dynarmic (who previously manually set Unicorn), which is desirable since Unicorn is somewhat busted now due to not properly supporting the `svcSetHeapSize` and `svcMapMemory` stuff.

Obligatory UI change screenshot:
image

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -148,19 +148,15 @@ System::ResultStatus System::Init(EmuWindow* emu_window, u32 system_mode) {
 
     current_process = Kernel::Process::Create("main");
 
-    switch (Settings::values.cpu_core) {
-    case Settings::CpuCore::Unicorn:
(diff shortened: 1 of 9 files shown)


bunnei:
:shipit:

Issue #218, opened 2018-02-23
pfba crash with Xbyak error (with dynarmic)


greggameplayer:
## Hardware specs :
- Nvidia GeForce GTX 960
- AMD FX-6300
- 8 GB Of Ram
- Windows 7 Pro

## Issue :
when i run the homebrew pfba yuzu after couples of seconds crash with

"Terminate called after throwing an instance of 'Xbyak::Error'
what(): label is too far
This application has resquested the runtime to terminate it in an unusual way.
Please contact the application's support team for more information."

## Log :
the major part of the log because there is a bug causes the non-fully registration of the log with the command "yuzu.exe 2> log.txt"
[Click Here](https://gist.github.com/greggameplayer/c8931afccd74b970c868a8d122167374)

Cpasjuste:
Hi, this error was not present a few days ago. I'm not sure if this comes from a libnx change or a yuzu change.

greggameplayer:
ok, so can you correct that please ?

Cpasjuste:
No, this is probably out of my knowledge (and time).

Hexagon12:
@MerryMage should be working on a fix for this, since I sent the proper stack trace and debugging info a while back.

greggameplayer:
@Hexagon12 thanks for the information

jroweboy:
We just pushed a fix for this. Please retest

Hexagon12:
This issue is now fixed.

Cpasjuste:
Hi, tried today, it seems the problem still exist : 

terminate called after throwing an instance of 'Xbyak::Error'
  what():  label is too far
Abandon (core dumped)

jroweboy:
@cpasjuste make sure you update your submodules as it's a crash in dynarmic

Cpasjuste:
Yep, cloned the repo "from scratch", I'll take a closer look.

Le mer. 28 mars 2018 à 16:00, James Rowe <notifications@github.com> a
écrit :

PR #285, opened 2018-03-27
dynarmic: Update to 12a1020

MerryMage:
Bugfix: Use a near jmp to end instead of short jmp in emit_X64_floating_point.


jroweboy:
This should fix #218

PR #292, opened 2018-03-30
NFP, SVC, and AudRen progress

bunnei:
Nothing terribly ground-breaking here - just some minor stuff that gets BOTW booting further.


bunnei:
🎈

PR #296, opened 2018-03-31
Fix stack region, implement FSP GetSize/SetSize, and some stubs

bunnei:
Stack region fix is the most important thing here - Without this, stack region could be overlapping with the adjacent memory region (TLS). This gets Cave Story and Binding of Isaac further.
Other stuff is just some misc changes that also get these games running further.


src/core/file_sys/disk_filesystem.cpp
```diff
@@ -174,8 +174,9 @@ u64 Disk_Storage::GetSize() const {
 }
 
 bool Disk_Storage::SetSize(const u64 size) const {
```
N00byKing: Is the return whether or not it succeeded?
If yes, shouldnt it return a resultcode?
Sorry if i'm being annoying / missing something...

bunnei: That is technically what it was there for, yes - I probably won't use it though since we don't know the error codes for this yet

N00byKing: I Understand

src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -268,8 +268,11 @@ std::vector<u8> HLERequestContext::ReadBuffer() const {
 
 size_t HLERequestContext::WriteBuffer(const void* buffer, size_t size) const {
     const bool is_buffer_b{BufferDescriptorB().size() && BufferDescriptorB()[0].Size()};
-
-    ASSERT_MSG(size <= GetWriteBufferSize(), "Size %lx is too big", size);
+    const size_t buffer_size{GetWriteBufferSize()};
+    if (size > buffer_size) {
+        LOG_CRITICAL(Core, "size=%08x is greater than buffer_size=%08x", size, buffer_size);
```
lioncash: These should likely be `%zu`

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -150,6 +150,25 @@ class IFile final : public ServiceFramework<IFile> {
         IPC::ResponseBuilder rb{ctx, 2};
         rb.Push(RESULT_SUCCESS);
     }
+
+    void SetSize(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        const u64 size = rp.Pop<u64>();
+        backend->SetSize(size);
+        LOG_DEBUG(Service_FS, "called, size=0x%ld, length=0x%ld", size);
```
lioncash: Should be using PRIu64. This is also missing the length argument to complement the formatting specifier.

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -150,6 +150,25 @@ class IFile final : public ServiceFramework<IFile> {
...
+
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(RESULT_SUCCESS);
+    }
+
+    void GetSize(Kernel::HLERequestContext& ctx) {
+        const u64 size = backend->GetSize();
+        LOG_DEBUG(Service_FS, "called, size=0x%ld", size);
```
lioncash: PRIu64

PR #297, opened 2018-04-01
hid: Write empty touch screen state.

bunnei:
Games check that the structs for the touchscreen in HID shared memory are valid, even if they don't use the touch screen. This change updates them, although does not implement actual touch input. With this change and #296, Cave Story+ now runs.

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -65,13 +65,14 @@ private:
     }
 
     void UpdatePadCallback(u64 userdata, int cycles_late) {
-        SharedMemory* mem = reinterpret_cast<SharedMemory*>(shared_mem->GetPointer());
+        SharedMemory mem{};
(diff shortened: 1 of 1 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -65,22 +65,23 @@ class IAppletResource final : public ServiceFramework<IAppletResource> {
     }
 
     void UpdatePadCallback(u64 userdata, int cycles_late) {
-        SharedMemory* mem = reinterpret_cast<SharedMemory*>(shared_mem->GetPointer());
+        SharedMemory mem{};
+        std::memcpy(&mem, shared_mem->GetPointer(), sizeof(SharedMemory));
```
jroweboy: why copy this to a temp variable just to copy it back later instead of just working directly on the sharedmemory location?

Subv: GetPointer will eventually go away, we should avoid treating guest memory as if it were always continuous from the host's PoV, that's not necessarily the case.

PR #298, opened 2018-04-01
Update CMakeLists.txt

kdlz4161:
Updates CMakeLists to use Qt 5.10.1 instead of Qt 5.7


jroweboy:
There is already a pull request open for this.

PR #262, opened 2018-03-22
Logging: Add fmtlib-based macros

daniellimws:
Add a new set of logging macros based on fmtlib
Similar but not exactly the same as citra-emu/citra#3533, with the main difference being:
* Citra currently uses a different version of fmt, which does not support `FMT_VARIADIC` anymore so
`make_args` is used instead. On the other hand, yuzu uses fmt 4.1.0 which doesn't have `make_args` yet, so `FMT_VARIADIC` is used.

Also, to prove that this works, "yuzu starting..." is changed to be logged with the new macro in yuzu/main.cpp

As discussed in Discord, do we want to do the migration all at once in this PR? Or, make it a two-step process, have this merged then work on the migration.

The main reason that in citra the process is split is due to the number of PRs in the queue, but there seems to be significantly less over here, so it is possible to do all at once.

PR #306, opened 2018-04-03
logging: Use variadic template instead of FMT_VARIADIC

daniellimws:
Due to premature merging of #262 I think the build may be failing right now. Should merge this ASAP to fix it.


bunnei:
lgtm... ty!

PR #307, opened 2018-04-03
logging: Fix new fmtlib-based definitions.

bunnei:
Should fix the current build breaks that were accidentally introduced with #262.


daniellimws:
Oh oops just opened #306 which is a complete duplicate.

bunnei:
Ah thanks for pushing a fix so fast, will close this in favor of yours

PR #309, opened 2018-04-04
core, main.h: Abort on 32Bit ROMs

### LINKED ISSUE #302: yuzu should bail early (and display an error to the user) for 32-bit apps
This is a good task for someone new who wants to make a contribution to yuzu. yuzu only supports aarch64 apps, which is like 99% of Switch games/homebrew. However, it's possible for apps to be 32-bit (e.g. Mario Kart 8 is). We currently try to run them, but instead should bail out early and display an error to the user. We already detect whether an app is 32-bit (see https://github.com/yuzu-emu/yuzu/blob/master/src/core/file_sys/program_metadata.cpp#L95 ), so this should be a really simple change.

Note, we probably will support 32-bit Switch apps *some day*... but that will likely be a long time in the future and is not a priority right now.

N00byKing:
Originally thought that someone could test it before i submit the PR, changed my mind though because leaving the issue open like that will cause people to duplicate efforts.

(If it works) Closes #302

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -92,6 +92,8 @@ System::ResultStatus System::Load(EmuWindow* emu_window, const std::string& file
             return ResultStatus::ErrorLoader_ErrorEncrypted;
         case Loader::ResultStatus::ErrorInvalidFormat:
             return ResultStatus::ErrorLoader_ErrorInvalidFormat;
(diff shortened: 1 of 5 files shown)


N00byKing:
@Hexagon12 Addressed

PR #312, opened 2018-04-06
Update fmtlib to fix msvc warnings

jroweboy:
Additionally, when updating fmtlib, there was a change in fmtlib broke
how the old logging macro was overloaded, so this works around that by
just naming the fmtlib macro impl something different


bunnei:
Thanks!

PR #319, opened 2018-04-10
Various service name fixes - part 1

Hexagon12:
This multi-part PR tries to implement as much new service names as possible without adding / stubbing new services.  Most of these are taken from SwitchBrew - http://switchbrew.org/index.php?title=Services_API .


src/core/hle/service/audio/audin_u.cpp
```diff
@@ -33,6 +38,8 @@ AudInU::AudInU() : ServiceFramework("audin:u") {
     static const FunctionInfo functions[] = {
         {0x00000000, nullptr, "ListAudioIns"},
         {0x00000001, nullptr, "OpenAudioIn"},
+        {0x00000003, nullptr, "OpenAudioInAuto"},
```
mailwl: maybe change to decimal as other service functions?

PR #320, opened 2018-04-11
Service/SSL: update service according switchbrew

mailwl:


lioncash:
The small fixes and the stubbing should go in their own PR. If you open a PR to update the SSL service, then just do exactly that -- update the service. There's also no rationale given for the stubbing.

mailwl:
this set of changes lets Lego games not to crash

bunnei:
@lioncash I think it's unfair to close this when we don't hold this standard everywhere. @mailwl in future, please try to breakup unrelated changes into multiple PRs. However, for this change, I'm OK with it - there are some fixes here, and we're still at the point of emulation where we are allowing stubs (we wouldn't have anything but homebrew running if we didn't)

lioncash:
@mailwl You should indicate stuff like that in the commit message to provide context for the actual change in the future

mailwl:
i will move unrelated stuff to another PR. don't merge yet

PR #321, opened 2018-04-11
Various service name fixes - part 2

Hexagon12:
A follow-up of #319 .

>This multi-part PR tries to implement as much new service names as possible without adding / stubbing new services. Most of these are taken from SwitchBrew - http://switchbrew.org/index.php?title=Services_API .


Hexagon12:
oh fuck

PR #325, opened 2018-04-13
Service/vi: Fix normal_params_size in GetDisplayResolution

Hexagon12:
The service requires normal_params_size to be 4, not 6. Fixes an issue in Minecraft Story Mode.


mailwl:
It's wrong! w & h - u64 values both, so, change should be in other place (push<u64> instead of push<u32>)

Hexagon12:
@mailwl You can open a PR yourself if you think that my PR is wrong.
~~Also pls implement NFP IUser~~

mailwl:
i will later. but use switchbrew for reference in future: http://switchbrew.org/index.php?title=Display_services#GetDisplayResolution

greggameplayer:
yes it might be wrong because this error stay in kirby star allies

PR #327, opened 2018-04-14
Fix the stuck in fullscreen bug

adityaruplaha:
Direct port of #3611 from Citra.
Original PR link: https://github.com/citra-emu/citra/pull/3611

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -776,9 +776,11 @@ void GMainWindow::closeEvent(QCloseEvent* event) {
         return;
     }
 
-    UISettings::values.geometry = saveGeometry();
+    if (ui.action_Fullscreen->isChecked()) {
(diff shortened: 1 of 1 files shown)


src/yuzu/main.cpp
```diff
@@ -1,4 +1,5 @@
 // Copyright 2014 Citra Emulator Project
+// Copyright 2014 Citra Emulator Project
```
N00byKing: Double Copyright header?

adityaruplaha:
Gimme a microsecond to fix it.

src/yuzu/main.cpp
daniellimws: wanna change it to Yuzu?

N00byKing: As both Projects share codebases, the Copyright header is only changed when major differences occur IIRC.

PR #333, opened 2018-04-15
shaders: Expose hints about used const buffers.

bunnei:
Updates shader generation to keep track of uniforms, and declares the used const buffers. Does the necessary plumbing to return these buffers back to GL land. Blah blah blah

**NOTE: NO GAMES WILL RENDER WITH THIS CHANGE**

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -140,6 +140,11 @@ public:
         return declarations.GetResult() + shader.GetResult();
     }
 
+    /// Returns entries in the shader that are useful for external functions
+    ShaderEntries GetEntries() const {
+        return {GetConstBuffersDeclarations()};
+    }
+
 private:
     /// Gets the Subroutine object corresponding to the specified address.
     const Subroutine& GetSubroutine(u32 begin, u32 end) const {
@@ -186,10 +191,9 @@ private:
     }
 
     /// Generates code representing a uniform (C buffer) register.
-    std::string GetUniform(const Uniform& reg) const {
-        std::string index = std::to_string(reg.index);
-        return "uniform_" + index + "[" + std::to_string(reg.offset >> 2) + "][" +
-               std::to_string(reg.offset & 3) + "]";
+    std::string GetUniform(const Uniform& reg) {
+        declr_const_buffers[reg.index].MarkAsUsed(reg.index, reg.offset);
+        return 'c' + std::to_string(reg.index) + '[' + std::to_string(reg.offset) + ']';
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -439,6 +443,19 @@ class GLSLGenerator {
         GenerateDeclarations();
     }
 
+    /// Returns a list of constant buffer declarations
+    std::vector<ConstBufferEntry> GetConstBuffersDeclarations() const {
+        std::vector<ConstBufferEntry> result;
+        for (unsigned index = 0; index < declr_const_buffers.size(); ++index) {
```
lioncash: `unsigned` should be a `size_t`

This could likely also just be:

```cpp
std::copy_if(declr_const_buffers.begin(), declr_const_buffers.end(), std::back_inserter(result),
             [](const auto& entry) { return entry.IsUsed(); });
```

or:

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -186,10 +191,9 @@ class GLSLGenerator {
     }
 
     /// Generates code representing a uniform (C buffer) register.
-    std::string GetUniform(const Uniform& reg) const {
-        std::string index = std::to_string(reg.index);
-        return "uniform_" + index + "[" + std::to_string(reg.offset >> 2) + "][" +
-               std::to_string(reg.offset & 3) + "]";
+    std::string GetUniform(const Uniform& reg) {
+        declr_const_buffers[reg.index].MarkAsUsed(reg.index, reg.offset);
+        return "c" + std::to_string(reg.index) + "[" + std::to_string(reg.offset) + "]";
```
lioncash: `'c'`, `'['`, and `']'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -463,6 +480,17 @@ class GLSLGenerator {
                                  ") out vec4 " + GetOutputAttribute(index) + ";");
         }
         declarations.AddLine("");
+
+        unsigned const_buffer_layout{};
```
lioncash: `= 0;` is likely more idiomatic, if that matters

src/video_core/renderer_opengl/gl_shader_gen.h
```diff
@@ -16,6 +18,40 @@ constexpr size_t MAX_PROGRAM_CODE_LENGTH{0x1000};
 
 using ProgramCode = std::array<u64, MAX_PROGRAM_CODE_LENGTH>;
 
+class ConstBufferEntry {
+public:
+    ConstBufferEntry() = default;
```
lioncash: The constructor can likely be omitted here.

PR #341, opened 2018-04-16
file_sys: Add HFS/PFS helper component

shinyquagsire23:
Useful for loading files from XCI (HFS) or NCA (PFS). Currently not tied to any loaders.

### DIFF
--- a/src/core/file_sys/partition_filesystem.cpp
+++ b/src/core/file_sys/partition_filesystem.cpp
@@ -1,0 +1,124 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <cinttypes>
+#include "common/file_util.h"
+#include "common/logging/log.h"
+#include "core/file_sys/partition_filesystem.h"
+#include "core/loader/loader.h"
+
+namespace FileSys {
+
+Loader::ResultStatus PartitionFilesystem::Load(const std::string& file_path, size_t offset) {
+    FileUtil::IOFile file(file_path, "rb");
+    if (!file.IsOpen())
+        return Loader::ResultStatus::Error;
+
+    // At least be as large as the header
+    if (file.GetSize() < sizeof(Header))
+        return Loader::ResultStatus::Error;
+
+    // For cartridges, HFSs can get very large, so we need to calculate the size up to
+    // the actual content itself instead of just blindly reading in the entire file.
+    Header pfs_header;
+    if (!file.ReadBytes(&pfs_header, sizeof(Header)))
+        return Loader::ResultStatus::Error;
+
+    bool is_hfs = (memcmp(pfs_header.magic.data(), "HFS", 3) == 0);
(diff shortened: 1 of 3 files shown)


src/core/file_sys/partition_filesystem.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    Loader::ResultStatus result = Load(file_data);
+    if (result != Loader::ResultStatus::Success)
+        LOG_ERROR(Service_FS, "Failed to load PFS from file %s!", file_path.c_str());
+
+    return result;
+}
+
+Loader::ResultStatus PartitionFilesystem::Load(const std::vector<u8> file_data, size_t offset) {
+    size_t total_size = static_cast<size_t>(file_data.size() - offset);
```
lioncash: The result of this expression would be a size_t, so the cast isn't necessary

src/core/file_sys/partition_filesystem.cpp
```diff
@@ -0,0 +1,124 @@
...
+        return Loader::ResultStatus::Error;
+
+    Loader::ResultStatus result = Load(file_data);
+    if (result != Loader::ResultStatus::Success)
+        LOG_ERROR(Service_FS, "Failed to load PFS from file %s!", file_path.c_str());
+
+    return result;
+}
+
+Loader::ResultStatus PartitionFilesystem::Load(const std::vector<u8> file_data, size_t offset) {
```
lioncash: `file_data` was likely intended to be a const reference.

src/core/file_sys/partition_filesystem.cpp
```diff
@@ -0,0 +1,124 @@
...
+            &file_data[strtab_offset + entry.fs_entry.strtab_offset]));
+        pfs_entries.push_back(entry);
+    }
+
+    content_offset = strtab_offset + pfs_header.strtab_size;
+
+    return Loader::ResultStatus::Success;
+}
+
+u32 PartitionFilesystem::GetNumEntries(void) const {
```
lioncash: `void` in empty parameter lists isn't necessary in C++

src/core/file_sys/partition_filesystem.cpp
```diff
@@ -0,0 +1,124 @@
...
+    size_t entries_offset = offset + sizeof(Header);
+    size_t entry_size = is_hfs ? sizeof(HFSEntry) : sizeof(PFSEntry);
+    size_t strtab_offset = entries_offset + (pfs_header.num_entries * entry_size);
+    for (u16 i = 0; i < pfs_header.num_entries; i++) {
+        FileEntry entry;
+
+        memcpy(&entry.fs_entry, &file_data[entries_offset + (i * entry_size)], sizeof(FSEntry));
+        entry.name = std::string(reinterpret_cast<const char*>(
+            &file_data[strtab_offset + entry.fs_entry.strtab_offset]));
+        pfs_entries.push_back(entry);
```
lioncash: `entry` can be `std::move`d here.

PR #339, opened 2018-04-16
GPU: Further improve shaders, textures, and blending

bunnei:
This is a Frankenstein PR of GPU work to [get games rendering](https://yuzu-emu.org/entry/first-games-are-running/). There are quite a few hacks in here, and some solid work. I plan on slowly breaking this up into separate more meaningful PRs. So in that light, I suggest reviewing followon PRs, not this one (this is more or less here to just provide Canary builds of games running).

PR #344, opened 2018-04-17
Shader decompiler changes part 2

bunnei:
This is a continuation of getting the various changes in #339 merged into mainline yuzu. In this PR, we bring over all non-hacky changes related to shader decompilation and GLSL shader generation. I've also addressed any related comments on #339. Here is a summary of changes:
* Implement shader instructions IPA, TEXS
* Implement immediate mode versions of shader instructions FMUL, FADD, FFMA
* Implement more MUFU sub operations
* Allow fragment shader to use vertex position
* General cleanup (logging, warnings, etc.)

**As was mentioned in #342, if you are already using changes to run games, I do not recommend trying to merge this branch. This does not add anything functionally new that isn't in #339.**

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -13,7 +13,7 @@ namespace Tegra {
 namespace Shader {
 
 struct Register {
-    Register() = default;
+    constexpr Register() = default;
 
     constexpr Register(u64 value) : value(value) {}
 
@@ -40,13 +40,13 @@ struct Register {
     }
 
 private:
-    u64 value;
+    u64 value{};
 };
 
 union Attribute {
     Attribute() = default;
 
-    constexpr Attribute(u64 value) : value(value) {}
+    constexpr explicit Attribute(u64 value) : value(value) {}
 
     enum class Index : u64 {
         Position = 7,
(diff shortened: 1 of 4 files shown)


src/video_core/engines/shader_bytecode.h
```diff
@@ -71,6 +68,19 @@ union Attribute {
     u64 value;
 };
 
+union Sampler {
+    Sampler() = default;
```
lioncash: This should also likely be constexpr and just assign `0` to `value` in the union body. Otherwise this is kind of inconsistent API

src/video_core/engines/shader_bytecode.h
```diff
@@ -71,6 +68,19 @@ union Attribute {
...
+
+    constexpr Sampler(u64 value) : value(value) {}
```
lioncash: This should probably also be `explicit` unless there's a good reason for it not to be

src/video_core/engines/shader_bytecode.h
```diff
@@ -307,18 +309,30 @@ union Instruction {
...
         BitField<45, 1, u64> negate_b;
         BitField<46, 1, u64> abs_a;
         BitField<48, 1, u64> negate_a;
         BitField<49, 1, u64> abs_b;
         BitField<50, 1, u64> abs_d;
+        BitField<56, 1, u64> negate_imm;
+
+        float GetImm20() const {
+            float result{};
+            u32 imm{static_cast<u32>(imm20)};
```
Subv: Is this always unsigned?

bunnei: It's not unsigned, it's a 20-bit part of a float. The sign bit is stored in a separate component.

bunnei: (note, u32 is just  used for bit manipulation, it's memcpy'd into a float below, which can be negative)

PR #347, opened 2018-04-18
gl_shader_gen: Support vertical/horizontal viewport flipping.

bunnei:
This is a continuation of getting the various changes in #339 merged into mainline yuzu. In this PR, we bring over the change to support flipping the viewport vertically (or horizontally) using negative viewport scales. This is not possible with just `glViewPort`, so we flip the final output in the vertex shader. This was tested with Cave Story+. I've also addressed any related comments on #339.

**As was mentioned in other PRs, if you are already using changes to run games, I do not recommend trying to merge this branch. This does not add anything functionally new that isn't in #339.**

### DIFF
--- a/src/video_core/engines/maxwell_3d.h
+++ b/src/video_core/engines/maxwell_3d.h
@@ -319,7 +319,14 @@ public:
                     }
                 } rt[NumRenderTargets];
 
-                INSERT_PADDING_WORDS(0x80);
+                f32 viewport_scale_x;
+                f32 viewport_scale_y;
+                f32 viewport_scale_z;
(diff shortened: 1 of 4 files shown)


src/video_core/engines/maxwell_3d.h
```diff
@@ -319,7 +319,14 @@ class Maxwell3D final {
                     }
                 } rt[NumRenderTargets];
 
-                INSERT_PADDING_WORDS(0x80);
+                f32 viewport_scale_x;
```
Subv: There are 16 sets of viewport registers, please add them all

PR #348, opened 2018-04-18
Technically, yuzu can boot commercial games

jlachniet:
Clarifies the yuzu cannot play commercial games to any reasonable extent, rather than not at all.


simonmkwii:
This is extremely petty and unimportant. Can you please not abuse pull requests for useless stuff like this?

LogPod:
:shipit:

N00byKing:
IMO, all forms of help are welcome

PR #353, opened 2018-04-19
GPU: Add support for the DXT23 and DXT45 compressed texture formats.

Subv:
The respective commit in #339 can be dropped.

Original work by @jroweboy


bunnei:
lgtm, tested that this works with puyo 👍

PR #367, opened 2018-04-20
math_util: Remove the Clamp() function

lioncash:
C++17 adds clamp() to the standard library, so we can remove ours in favor of it.


bunnei:
Please fix clang-ci errors

PR #369, opened 2018-04-20
ShaderGen: Implemented fsetp/kil and predicated instruction execution.

Subv:
We'll also ignore 'sched' as it appears on a fixed interval (once every 4 instructions).

Predicated execution uses bool variables named 'pX' where X is the predicate id.
fsetp sets the predicate values based on some condition.

This fixes the text in Cave Story and probably other things.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -13,6 +13,9 @@ namespace Tegra {
 namespace Shader {
 
 struct Register {
+    // Register 255 is special cased to always be 0
+    static constexpr size_t ZeroIndex = 255;
+
     constexpr Register() = default;
 
     constexpr Register(u64 value) : value(value) {}
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -109,6 +109,8 @@ union OpCode {
 
         FSETP_R = 0x5BB,
         FSETP_C = 0x4BB,
+        FSETP_IMM = 0x36B,
+        FSETP_NEG_IMM = 0x37B,
         EXIT = 0xE30,
         KIL = 0xE33,
 
@@ -124,6 +126,7 @@ union OpCode {
         Ffma,
         Flow,
         Memory,
+        FloatPredicate,
         Unknown,
     };
 
@@ -164,6 +167,9 @@ union OpCode {
         case Id::FSETP_C:
         case Id::KIL:
             return op4;
+        case Id::FSETP_IMM:
+        case Id::FSETP_NEG_IMM:
+            return Id::FSETP_IMM;
         }
 
         switch (op5) {
@@ -241,8 +247,9 @@ union OpCode {
         info_table[Id::FMUL_C] = {Type::Arithmetic, "fmul_c"};
         info_table[Id::FMUL_IMM] = {Type::Arithmetic, "fmul_imm"};
         info_table[Id::FMUL32_IMM] = {Type::Arithmetic, "fmul32_imm"};
-        info_table[Id::FSETP_C] = {Type::Arithmetic, "fsetp_c"};
-        info_table[Id::FSETP_R] = {Type::Arithmetic, "fsetp_r"};
+        info_table[Id::FSETP_C] = {Type::FloatPredicate, "fsetp_c"};
+        info_table[Id::FSETP_R] = {Type::FloatPredicate, "fsetp_r"};
+        info_table[Id::FSETP_IMM] = {Type::FloatPredicate, "fsetp_imm"};
         info_table[Id::EXIT] = {Type::Trivial, "exit"};
         info_table[Id::IPA] = {Type::Trivial, "ipa"};
         info_table[Id::KIL] = {Type::Flow, "kil"};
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -258,17 +260,77 @@ class GLSLGenerator {
...
+     * @param value The expression value to assign to the predicate.
+     */
+    void SetPredicate(u64 pred, const std::string& value) {
+        using Tegra::Shader::Pred;
+        // Can't assign to the constant predicate.
+        ASSERT(pred != static_cast<u64>(Pred::UnusedIndex));
+
+        std::string variable = 'p' + std::to_string(pred);
+        shader.AddLine(variable + " = " + value + ';');
+        declr_predicates.insert(variable);
```
lioncash: `variable` can be `std::move`d here.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -258,17 +260,77 @@ class GLSLGenerator {
...
+     */
+    std::string GetPredicateCondition(Instruction instr) const {
+        using Tegra::Shader::Pred;
+        ASSERT(instr.pred != Pred::UnusedIndex);
+
+        std::string variable =
+            'p' + std::to_string(static_cast<u64>(instr.pred.pred_index.Value()));
+
+        if (instr.negate_pred) {
+            return "!(" + variable + ")";
```
lioncash: `')'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -258,17 +260,77 @@ class GLSLGenerator {
...
         const Instruction instr = {program_code[offset]};
 
         shader.AddLine("// " + std::to_string(offset) + ": " + OpCode::GetInfo(instr.opcode).name);
 
+        using Tegra::Shader::Pred;
+        ASSERT_MSG(instr.pred.full_pred != Pred::NeverExecute,
+                   "NeverExecute predicate not implemented");
+
+        if (instr.pred.pred_index != static_cast<u64>(Pred::UnusedIndex)) {
+            shader.AddLine("if (" + GetPredicateCondition(instr) + ")");
```
lioncash: `')'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -258,17 +260,77 @@ class GLSLGenerator {
...
+            shader.AddLine("{");
```
lioncash: `'{'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -432,14 +494,70 @@ class GLSLGenerator {
             }
             break;
         }
+        case OpCode::Type::FloatPredicate: {
+            std::string op_a = instr.fsetp.neg_a ? "-" : "";
+            op_a += GetRegister(instr.gpr8);
 
+            if (instr.fsetp.abs_a) {
+                op_a = "abs(" + op_a + ")";
```
lioncash: `')'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -432,14 +494,70 @@ class GLSLGenerator {
...
+                           instr.fsetp.pred39 == static_cast<u64>(Pred::UnusedIndex),
+                       "Compound predicates are not implemented");
+
+            // We can't use the constant predicate as destination.
+            ASSERT(instr.fsetp.pred3 != static_cast<u64>(Pred::UnusedIndex));
+
+            using Tegra::Shader::PredCondition;
+            switch (instr.fsetp.cond) {
+            case PredCondition::LessThan:
+                SetPredicate(instr.fsetp.pred3, "(" + op_a + ") < (" + op_b + ")");
```
lioncash: `'('` `')'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -432,14 +494,70 @@ class GLSLGenerator {
...
+                break;
+            case PredCondition::Equal:
+                SetPredicate(instr.fsetp.pred3, "(" + op_a + ") == (" + op_b + ")");
```
lioncash: `'('` `')'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -458,6 +576,12 @@ class GLSLGenerator {
         }
         }
 
+        // Close the predicate condition scope.
+        if (instr.pred.pred_index != static_cast<u64>(Pred::UnusedIndex)) {
+            --shader.scope;
+            shader.AddLine("}");
```
lioncash: `'}'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -587,6 +711,11 @@ class GLSLGenerator {
             declarations.AddLine("");
             ++const_buffer_layout;
         }
+
+        declarations.AddLine("");
```
lioncash: `AddNewLine()`

src/video_core/engines/shader_bytecode.h
```diff
@@ -13,6 +13,9 @@ namespace Tegra {
 namespace Shader {
 
 struct Register {
+    // Register 255 is special cased to always be 0
+    static constexpr size_t ZeroIndex = 255;
```
bunnei: Should we add an `Index` enum, like with `Attribute`? (I guess `Color` could also be in here)

bunnei: Also, why `size_t`?

Subv: I don't see the need for an enum, specially since we know that Color outputs are not hardcoded (There seems to be a configuration field that we haven't found yet)

As for `size_t`, is there a reason for it not to be so?

PR #370, opened 2018-04-20
Kernel: Reworked the new kernel synchronization primitives.

Subv:
Properly implemented svcArbitrateLock, svcArbitrateUnlock, svcWaitProcessWideKey and svcSignalProcessWideKey.

These implementations were based on reverse engineering. We should write a suite of hwtests to completely verify every edge case.

I didn't see any regressions in my games, but please test it as much as possible.


src/core/hle/kernel/thread.h
```diff
@@ -109,6 +109,19 @@ class Thread final : public WaitObject {
      */
     void BoostPriority(u32 priority);
 
+    /*
```
bunnei: nit: These aren't docstrings

bunnei:
Tested a bunch of games, didn't find any obvious regresssions

PR #381, opened 2018-04-23
GetSharedFontInOrderOfPriority

ogniK5377:
Added support for pl:u#GetSharedFontInOrderOfPriority


src/core/hle/service/ns/pl_u.cpp
```diff
@@ -116,4 +116,34 @@ void PL_U::GetSharedMemoryNativeHandle(Kernel::HLERequestContext& ctx) {
...
+    std::vector<u32> languages;
+    std::vector<u32> offsets;
+    std::vector<u32> sizes;
+
+    Memory::Write32(ctx.BufferDescriptorB()[0].Address(), language_code);
+    Memory::Write32(ctx.BufferDescriptorB()[1].Address(),
+                    SHARED_FONT_REGIONS[language_code].offset);
+    Memory::Write32(ctx.BufferDescriptorB()[2].Address(), SHARED_FONT_REGIONS[language_code].size);
+
+    for (int i = 0, offset = 4; i < SHARED_FONT_REGIONS.size(); i++) {
```
lioncash: this should be a `size_t`, otherwise it'll cause a signed/unsigned comparison warning.

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -116,4 +116,35 @@ void PL_U::GetSharedMemoryNativeHandle(Kernel::HLERequestContext& ctx) {
...
+    IPC::ResponseBuilder rb{ctx, 4};
+    rb.Push(RESULT_SUCCESS);
+    rb.Push<u8>(static_cast<u8>(LoadState::Done)); // Fonts Loaded
+    rb.Push<u32>(static_cast<u32>(SHARED_FONT_REGIONS.size()));
+
+    std::vector<u32> languages;
+    std::vector<u32> offsets;
+    std::vector<u32> sizes;
+
+    Memory::Write32(ctx.BufferDescriptorB()[0].Address(), language_code);
```
Subv: Store these buffer addresses in variables with meaningful names

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -116,4 +116,35 @@ void PL_U::GetSharedMemoryNativeHandle(Kernel::HLERequestContext& ctx) {
...
+    Memory::Write32(ctx.BufferDescriptorB()[1].Address(),
+                    SHARED_FONT_REGIONS[language_code].offset);
```
Subv: What happens if you pass a language_code that's bigger than the size of SHARED_FONT_REGIONS?

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -116,4 +116,35 @@ void PL_U::GetSharedMemoryNativeHandle(Kernel::HLERequestContext& ctx) {
...
+    Memory::Write32(ctx.BufferDescriptorB()[2].Address(), SHARED_FONT_REGIONS[language_code].size);
+
+    int offset = 4;
```
Subv: Once you have created variables for the addresses, you can increment that directly without using this offset. Also, use sizeof

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -116,4 +116,37 @@ void PL_U::GetSharedMemoryNativeHandle(Kernel::HLERequestContext& ctx) {
     rb.PushCopyObjects(shared_font_mem);
 }
 
+void PL_U::GetSharedFontInOrderOfPriority(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+    const u32 language_code{rp.Pop<u32>()};
+    NGLOG_DEBUG(Service_NS, "called, language_code=%d", language_code);
+    IPC::ResponseBuilder rb{ctx, 4};
+    if (language_code > SHARED_FONT_REGIONS.size()) {
+        rb.Push(ResultCode(-1));                          // TODO(ogniK): Find real result code
```
baka0815: Maybe add a NGLOG_WARNING() here that the`` language_code > size()``?

PR #384, opened 2018-04-23
 Nvdrv/nvhost-as-gpu: Implemented the ioctl REMAP command. 

Subv:
It takes a previously-reserved (AllocateSpace) GPU memory address and maps it to the address of the nvmap object passed to Remap.

This is used by 1-2 Switch.


src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -56,6 +61,36 @@ u32 nvhost_as_gpu::AllocateSpace(const std::vector<u8>& input, std::vector<u8>&
...
+    size_t num_entries = input.size() / sizeof(IoctlRemapEntry);
+
+    NGLOG_WARNING(Service_NVDRV, "(STUBBED) called, num_entries=0x{:X}", num_entries);
+
+    std::vector<IoctlRemapEntry> entries(num_entries);
+    std::memcpy(entries.data(), input.data(), input.size());
+
+    auto& gpu = Core::System::GetInstance().GPU();
+
+    for (auto& entry : entries) {
```
lioncash: Unless I'm mistaken, this can be const qualified

PR #388, opened 2018-04-24
Refactor rasterizer cache

bunnei:
This is a big refactor of our OpenGL rasterizer cache to be based on GPU addresses instead of CPU addresses. Summary of changes:
* Update MemoryManager naming to not use PAddr / PhysicalAddress
* Use GPUVAddr for Flush/Invalidate methods in rasterizer/etc.
* Use GPUVAddr for internally figuring out framebuffer/texture surfaces
* Flush/invalidate compressed textures properly
* Use configurable texture block height wherever possible instead of fixed 8 from Citra

For games we currently run, there should be no change in functionality. However, please keep an eye out for any perf regression.
Suggest reviewing commit-by-commit.


src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -82,6 +84,26 @@ struct SurfaceParams {
...
+    static constexpr unsigned int GetCompresssionFactor(PixelFormat format) {
+        if (format == PixelFormat::Invalid)
+            return 0;
+
+        constexpr std::array<unsigned int, MaxPixelFormat> compression_factor_table = {
+            1, // ABGR8
+            1, // B5G6R5
+            1, // A2B10G10R10
+            4, // DXT1
+            4, // DXT23
```
bunnei: Is this right?

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -82,6 +84,26 @@ struct SurfaceParams {
...
+            4, // DXT45
```
bunnei: Is this right?

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -945,6 +960,33 @@ Surface RasterizerCacheOpenGL::GetSurface(const SurfaceParams& params, ScaleMatc
     return surface;
 }
 
+boost::optional<Tegra::GPUVAddr> RasterizerCacheOpenGL::TryFindFramebufferGpuAddress(
```
bunnei: This function is not ideal, but it's what I came up with after getting pretty burnt out on working on this :)
Open to suggestions on better options... (although this will probably work fine at least until one of the below asserts is hit)

src/core/memory.cpp
```diff
@@ -394,19 +403,29 @@ void RasterizerFlushVirtualRegion(VAddr start, u64 size, FlushMode mode) {
...
-        case FlushMode::Flush:
-            rasterizer->FlushRegion(overlap_start, overlap_size);
-            break;
-        case FlushMode::Invalidate:
-            rasterizer->InvalidateRegion(overlap_start, overlap_size);
-            break;
-        case FlushMode::FlushAndInvalidate:
-            rasterizer->FlushAndInvalidateRegion(overlap_start, overlap_size);
-            break;
+        for (const auto& gpu_address : gpu_addresses) {
```
bunnei: There is often multiple GPU regions (mirrors?) mapped per CPU region. I'm flushing them all, I'm not sure if this is the best way to handle this, though?

src/video_core/memory_manager.cpp
```diff
@@ -84,6 +90,17 @@ boost::optional<VAddr> MemoryManager::GpuToCpuAddress(GPUVAddr gpu_addr) {
     return base_addr + (gpu_addr & PAGE_MASK);
 }
 
+std::vector<GPUVAddr> MemoryManager::CpuToGpuAddress(VAddr cpu_addr) {
```
Subv: This function can be const

src/core/memory.cpp
```diff
@@ -325,15 +325,24 @@ u8* GetPhysicalPointer(PAddr address) {
...
+        boost::optional<VAddr> maybe_vaddr =
+            Core::System::GetInstance().GPU().memory_manager->GpuToCpuAddress(gpu_addr);
+        // The GPU <-> CPU virtual memory mapping is not 1:1
+        if (!maybe_vaddr) {
+            LOG_ERROR(HW_Memory,
+                      "Trying to flush a cached region to an invalid physical address %08X",
+                      gpu_addr);
+            continue;
+        }
+        VAddr vaddr = *maybe_vaddr;
```
Subv: A single GPU page can span multiple CPU pages since the GPU page size is greater than the CPU page size, right now you're only marking the first page of such mapping as cached.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -342,6 +347,11 @@ bool SurfaceParams::CanTexCopy(const SurfaceParams& texcopy_params) const {
     return FromInterval(texcopy_params.GetInterval()).GetInterval() == texcopy_params.GetInterval();
 }
 
+VAddr SurfaceParams::GetCpuAddr() const {
+    const auto& gpu = Core::System::GetInstance().GPU();
+    return cpu_addr.get_value_or(*gpu.memory_manager->GpuToCpuAddress(addr));
```
Subv: At some assert here so that both cpu_addr and addr are not set at the same time

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1390,7 +1401,7 @@ void RasterizerCacheOpenGL::UnregisterSurface(const Surface& surface) {
     surface_cache.subtract({surface->GetInterval(), SurfaceSet{surface}});
 }
 
-void RasterizerCacheOpenGL::UpdatePagesCachedCount(VAddr addr, u64 size, int delta) {
+void RasterizerCacheOpenGL::UpdatePagesCachedCount(Tegra::GPUVAddr addr, u64 size, int delta) {
     const u64 num_pages =
         ((addr + size - 1) >> Memory::PAGE_BITS) - (addr >> Memory::PAGE_BITS) + 1;
```
Subv: All this code that deals with pages should use the PAGE_BITS in Tegra::MemoryManager instead. Keep in mind that a GPU memory page is bigger than a CPU memory page

src/video_core/renderer_opengl/gl_rasterizer_cache.h
Subv: yes

src/video_core/renderer_opengl/gl_rasterizer_cache.h
Subv: yes

Subv: Now that FormatTuple's compression factor isn't used, you should remove it

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -255,6 +275,14 @@ struct SurfaceParams {
     // Returns the region of the biggest valid rectange within interval
     SurfaceInterval GetCopyableInterval(const Surface& src_surface) const;
 
+    u32 GetActualWidth() const {
```
Subv: Add a comment about these functions. It can get confusing really quickly

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -235,7 +235,7 @@ SurfaceParams SurfaceParams::FromInterval(SurfaceInterval interval) const {
     } else {
         // 1 row
         ASSERT(aligned_end - aligned_start == stride_tiled_bytes);
-        const u64 tiled_alignment = BytesInPixels(is_tiled ? 8 * 8 : 1);
+        const u64 tiled_alignment = BytesInPixels(is_tiled ? block_height * block_height : 1);
```
Subv: This is wrong, tiles aren't Height * Height, there is also a configurable tile width in the TIC configuration.

src/video_core/memory_manager.cpp
```diff
@@ -8,90 +8,112 @@
...
-    return base_addr + (paddr & PAGE_MASK);
+
+    if (base_addr == static_cast<u64>(PageStatus::Allocated)) {
+        return {};
+    }
+
+    return base_addr + (gpu_addr & PAGE_MASK);
+}
+
+std::vector<GPUVAddr> MemoryManager::CpuToGpuAddress(VAddr cpu_addr) {
```
lioncash: This can be a const member function as far as I can tell.

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -82,6 +84,26 @@ struct SurfaceParams {
         Invalid = 4,
     };
 
+    static constexpr unsigned int GetCompresssionFactor(PixelFormat format) {
```
lioncash: Why not use `u32` here for consistency with the other functions added below.

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -82,6 +84,26 @@ struct SurfaceParams {
...
+        if (format == PixelFormat::Invalid)
+            return 0;
+
+        constexpr std::array<unsigned int, MaxPixelFormat> compression_factor_table = {
```
lioncash: these should be initialized with double bracing `compression_factor_table{{ stuff }};` otherwise it'll cause warnings with clang.

Hexagon12:
Please fix the conflicts.

src/video_core/renderer_opengl/gl_rasterizer_cache.h
bunnei: addressed

src/video_core/renderer_opengl/gl_rasterizer_cache.h
bunnei: addressed

src/video_core/memory_manager.cpp
bunnei: addressed

bunnei:
Still a few issues before this is ready for mergin:
* Address Subv's comment with block width, as well as further testing of this part of the change
* BoI assert's when going in game, need to debug still
* BoI title screen now has a thin black bar at the top, possibly related to the block width/height change, need to debug still
* Verify this does not break hb apps

src/core/memory.cpp
```diff
@@ -325,15 +325,24 @@ u8* GetPhysicalPointer(PAddr address) {
...
-void RasterizerMarkRegionCached(VAddr start, u64 size, bool cached) {
-    if (start == 0) {
+void RasterizerMarkRegionCached(Tegra::GPUVAddr gpu_addr, u64 size, bool cached) {
+    if (gpu_addr == 0) {
         return;
     }
 
-    u64 num_pages = ((start + size - 1) >> PAGE_BITS) - (start >> PAGE_BITS) + 1;
-    VAddr vaddr = start;
+    u64 num_pages = ((gpu_addr + size - 1) >> PAGE_BITS) - (gpu_addr >> PAGE_BITS) + 1;
```
Subv: This is still using CPU pages

src/core/memory.cpp
Subv: This has not been addressed

src/video_core/memory_manager.cpp
```diff
@@ -8,90 +8,112 @@
...
 
-VAddr MemoryManager::PhysicalToVirtualAddress(PAddr paddr) {
-    VAddr base_addr = PageSlot(paddr);
+boost::optional<VAddr> MemoryManager::GpuToCpuAddress(GPUVAddr gpu_addr) {
+    VAddr base_addr = PageSlot(gpu_addr);
     ASSERT(base_addr != static_cast<u64>(PageStatus::Unmapped));
-    return base_addr + (paddr & PAGE_MASK);
+
+    if (base_addr == static_cast<u64>(PageStatus::Allocated)) {
+        return {};
```
Subv: boost::none

PR #389, opened 2018-04-24
Service/FS: implement IFileSystem::RenameFile

mailwl:
This PR allow to run PyNX homebrew

### DIFF
--- a/src/core/hle/service/filesystem/fsp_srv.cpp
+++ b/src/core/hle/service/filesystem/fsp_srv.cpp
@@ -239,7 +239,7 @@ public:
             {2, &IFileSystem::CreateDirectory, "CreateDirectory"},
             {3, nullptr, "DeleteDirectory"},
             {4, nullptr, "DeleteDirectoryRecursively"},
-            {5, nullptr, "RenameFile"},
+            {5, &IFileSystem::RenameFile, "RenameFile"},
             {6, nullptr, "RenameDirectory"},
             {7, &IFileSystem::GetEntryType, "GetEntryType"},
             {8, &IFileSystem::OpenFile, "OpenFile"},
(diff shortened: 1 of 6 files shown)


src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -300,6 +300,26 @@ class IFileSystem final : public ServiceFramework<IFileSystem> {
         rb.Push(backend->CreateDirectory(name));
     }
 
+    void RenameFile(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+
+        std::vector<u8> buffer;
```
lioncash:
```cpp
std::vector<u8> buffer(ctx.BufferDescriptorX()[0].Size());
```

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -300,6 +300,26 @@ class IFileSystem final : public ServiceFramework<IFileSystem> {
...
+        buffer.resize(ctx.BufferDescriptorX()[0].Size());
+        Memory::ReadBlock(ctx.BufferDescriptorX()[0].Address(), buffer.data(), buffer.size());
+        auto end = std::find(buffer.begin(), buffer.end(), '\0');
```
lioncash: If everything going through here guaranteed to be null-terminated? As in, even if someone purposefully put in non-null-terminated data, would it still come through non-null terminated? If so, this case should probably be guarded against.

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -300,6 +300,26 @@ class IFileSystem final : public ServiceFramework<IFileSystem> {
...
+        std::string src_name(buffer.begin(), end);
+
+        buffer.resize(ctx.BufferDescriptorX()[1].Size());
+        Memory::ReadBlock(ctx.BufferDescriptorX()[1].Address(), buffer.data(), buffer.size());
```
lioncash: This seems to duplicate the above code. Perhaps consider a lambda function like:

```cpp
auto read_string = [&ctx](size_t index) {
  const auto descriptor = ctx.BufferDescriptorX()[index];
  std::vector<u8> data(descriptor.Size());
  Memory::ReadBlock(descriptor.Address(), buffer.data(), buffer.size());
  const auto end = std::find(buffer.begin(), buffer.end(), '\0');
  return std::string(buffer.begin(), end);
};
```

lioncash:
Given it was merged while I was reviewing it, nothing above really matters I guess, but the first one is still something that should be considered.

PR #392, opened 2018-04-24
service: Move logging macros over to the new fmt-compatible ones

lioncash:
Fairly self-explanatory, moves the macros over to the new fmt-compatible ones.


lioncash:
@bunnei done

bunnei:
👍 ty

PR #404, opened 2018-04-26
Add MOV32I instruction. 1-2 Switch uses it.

MoochMcGee:


bunnei:
I assume MOVI is move integer? If so, this impl is wrong

PR #408, opened 2018-04-27
gl_shader_decompiler: Add GLSLRegisterManager class to track register state.

bunnei:
Next part of my changes for supporting integer shader instructions. With this change, we add the `GLSLRegisterManager` class, which is used to track each `GLSLRegister` instance. This class now encapsulates all register `Set*` and `Get*` methods, handling integer/float conversions as needed.

At this point, there are still no functional changes - no new instructions implemented, but should be no regressions with currently supported shaders that use floats.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -41,7 +41,7 @@ enum class ExitMethod {
 struct Subroutine {
     /// Generates a name suitable for GLSL source code.
     std::string GetName() const {
-        return "sub_" + std::to_string(begin) + "_" + std::to_string(end);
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -301,6 +444,49 @@ class GLSLGenerator {
         }
     }
 
+    /// Generates code to use for a swizzle operation.
+    static std::string GetSwizzle(u64 elem) {
+        std::string swizzle = ".";
+        swizzle += "xyzw"[elem];
```
lioncash: This function should probably have a debug assert or something to make sure `elem` stays in range of the string (I realize this is just moved code though)

PR #409, opened 2018-04-27
general: Convert assertion macros over to be fmt-compatible

lioncash:
After this, only LOG_GENERIC calls need to be addressed, then we can kick out the old logging system entirely


bunnei:
awesome, thanks for doing this tedious stuff 👍

PR #410, opened 2018-04-27
core/renderer_opengl: Replace usages of LOG_GENERIC with fmt-capable equivalents

lioncash:
After this, a follow-up PR will remove the old printf-based logging system.


jroweboy:
One thing to note is we *do* get errors from gl drivers through that. I'm worried that people will stop noticing these errors and warnings if they are on debug (as the default log level will hide them) 

To be frank, i don't know that anyone has done much about the GL warnings and errors though. So maybe its fine to put it under debug and call it a day.

lioncash:
Yeah, I felt that after opening the PR, so I've added it back with the logging done in each case through the appropriate fmt-equivalent logger.

PR #417, opened 2018-04-29
set/am: Fix code for getting language codes

bunnei:
This fixes the text in 1-2-Switch to be the returned language (as opposed to always Japanese).

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -12,6 +12,7 @@
 #include "core/hle/service/apm/apm.h"
 #include "core/hle/service/filesystem/filesystem.h"
 #include "core/hle/service/nvflinger/nvflinger.h"
+#include "core/hle/service/set/set.h"
 #include "core/settings.h"
 
 namespace Service::AM {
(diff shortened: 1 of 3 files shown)


src/core/hle/service/am/am.cpp
```diff
@@ -537,10 +537,16 @@ void IApplicationFunctions::SetTerminateResult(Kernel::HLERequestContext& ctx) {
 }
 
 void IApplicationFunctions::GetDesiredLanguage(Kernel::HLERequestContext& ctx) {
+    // TODO(bunnei): This should be configurable, and return one of the language codes that are
+    // available with SET::GetAvailableLanguageCodes
+    u64 language_code{};
+    std::array<char, 8> language_code_str = {{"en-US"}};
```
lioncash: Wouldn't it be better to use an enum like:

```cpp
enum class LanguageCodes : u64 {
    EN_US = 0x00000053552D6E65,
    ... etc
};
```

So you don't need to copy it? (not that it matters performance-wise, but it can be reused between this func and the func in set).

bunnei: mmk

PR #419, opened 2018-04-29
set/am: Fix code for getting language codes

KP122222:
Pull request #418 - set/Pull request #418 - set/am: Fix code for getting language codes


bunnei:
What are you trying to do with this change? Looks like you're just dupping my PR.

PR #422, opened 2018-04-29
Shader instructions MOV_C, MOV_R, and several minor GPU things

bunnei:
With these changes, we now get the loading graphic on Minecraft

image


src/video_core/engines/maxwell_3d.cpp
```diff
@@ -206,7 +205,7 @@ void Maxwell3D::DrawArrays() {
         debug_context->OnEvent(Tegra::DebugContext::Event::FinishedPrimitiveBatch, nullptr);
     }
 
-    const bool is_indexed{regs.index_array.count && !regs.vertex_buffer.count};
+    const bool is_indexed{regs.index_array.count != 0};
```
Subv: This can fail if you first use DrawElements and then use DrawArrays, IMO the solution here is to zero those registers after a draw call

PR #427, opened 2018-05-01
ipc: Add support for PopIpcInterface() method.

bunnei:
This can be used for domain objects as inputs to service functions.

Edit: Admittedly, this is a little sloppy, and not 100% tested. Waiting for @ogniK5377 to confirm that it works correctly.  This is meant to be used with something like `ILibraryAppletAccessor::PushInData`, where we'd want to do:
`std::shared_ptr<AM::IStorage> storage = rp.PopIpcInterface<AM::IStorage>();`

### DIFF
--- a/src/core/hle/kernel/hle_ipc.h
+++ b/src/core/hle/kernel/hle_ipc.h
@@ -202,6 +202,16 @@ public:
         domain_objects.emplace_back(std::move(object));
     }
 
+    template <typename T>
+    std::shared_ptr<T> GetDomainRequestHandler(size_t index) const {
+        return std::static_pointer_cast<T>(domain_request_handlers[index]);
+    }
+
+    void SetDomainRequestHandlers(
+        const std::vector<std::shared_ptr<SessionRequestHandler>>& handlers) {
+        domain_request_handlers = handlers;
+    }
+
     /// Clears the list of objects so that no lingering objects are written accidentally to the
     /// response buffer.
     void ClearIncomingObjects() {
@@ -245,6 +255,8 @@ private:
     unsigned data_payload_offset{};
     unsigned buffer_c_offset{};
     u32_le command{};
+
+    std::vector<std::shared_ptr<SessionRequestHandler>> domain_request_handlers;
 };
 
 } // namespace Kernel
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/hle_ipc.h
```diff
@@ -202,6 +202,15 @@ class HLERequestContext {
         domain_objects.emplace_back(std::move(object));
     }
 
+    template <typename T>
+    std::shared_ptr<T> GetDomainRequestHandler(size_t index) const {
+        return std::static_pointer_cast<T>(domain_request_handlers[index]);
+    }
+
+    void SetDomainRequestHandlers(std::vector<std::shared_ptr<SessionRequestHandler>> handlers) {
+        domain_request_handlers = handlers;
```
lioncash: `handlers` can be `std::move`d here

bunnei: We can't move this, ServerSession is the real owner of these request handles. Context will go out of scope and destroy them

wwylele: @bunnei the `handlers` is passed in by value (= a copy of the owning list), so you can safely move it without touching the original ownership from `ServerSession`. Otherwise you might want to pass it by const reference.

src/core/hle/kernel/hle_ipc.h
```diff
@@ -202,6 +202,16 @@ class HLERequestContext {
         domain_objects.emplace_back(std::move(object));
     }
 
+    template <typename T>
+    std::shared_ptr<T> GetDomainRequestHandler(size_t index) const {
+        return std::static_pointer_cast<T>(domain_request_handlers[index]);
+    }
+
+    void SetDomainRequestHandlers(
+        const std::vector<std::shared_ptr<SessionRequestHandler>> handlers) {
```
lioncash: note that this is missing an ampersand for a reference, so this is just taking it by value still

lioncash: though I still don't see why this can't be moved if making a copy is desirable (the member variable isn't a reference either), it's still going to function as expected in that scenario (though obviously, moving const qualified values doesn't make sense).

PR #436, opened 2018-05-06
Initial support for multi-core

bunnei:
With this change, we add initial support for multi-core. This is very preliminary, and as such is disabled by default - but I'd like to get the basic changes merged, and we can build upon it. In summary, this change:
* Adds a new config setting to enable multicore, which is disabled by default
* Creates an ARM core and scheduler for each of Switch's 4 CPU cores
* Creates a thread to run each of the CPU cores
* Improves/implements several SVCs such that guest thread threads are scheduled on the appropriate core
* Adds a barrier mechanism to run each of the cores in lockstep
* Various refactoring/fixes to work with 4 cores/threads

We do **not** currently implement the ARM exclusive access load/store instructions with this change, so  we know any games that rely on this behavior will be broken (e.g. BOTW). @MerryMage we are looking for guidance and/or an implementation to support this scenario across several Dynarmic instances.
 
That being said, all games that previously rendered in yuzu should work with multicore enabled. Please test with the feature both enabled and disabed (ideally, disabled should work exactly  as before). At this time, I wouldn't expect this to give much performance improvement (if any at all), but please keep an eye out for any major performance regressions.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -5,10 +5,6 @@
 #include <memory>
 #include <utility>
 #include "common/logging/log.h"
-#ifdef ARCHITECTURE_x86_64
-#include "core/arm/dynarmic/arm_dynarmic.h"
-#endif
-#include "core/arm/unicorn/arm_unicorn.h"
 #include "core/core.h"
 #include "core/core_timing.h"
 #include "core/gdbstub/gdbstub.h"
@@ -33,9 +29,6 @@ System::~System() = default;
 
 System::ResultStatus System::RunLoop(bool tight_loop) {
     status = ResultStatus::Success;
-    if (!cpu_core) {
-        return ResultStatus::ErrorNotInitialized;
-    }
 
     if (GDBStub::IsServerEnabled()) {
         GDBStub::HandlePacket();
@@ -52,24 +45,7 @@ System::ResultStatus System::RunLoop(bool tight_loop) {
         }
     }
 
-    // If we don't have a currently active thread then don't execute instructions,
-    // instead advance to the next event and try to yield to the next thread
-    if (Kernel::GetCurrentThread() == nullptr) {
-        NGLOG_TRACE(Core_ARM, "Idling");
-        CoreTiming::Idle();
-        CoreTiming::Advance();
-        PrepareReschedule();
-    } else {
-        CoreTiming::Advance();
-        if (tight_loop) {
-            cpu_core->Run();
-        } else {
-            cpu_core->Step();
-        }
-    }
-
-    HW::Update();
-    Reschedule();
+    cpu_cores[0]->RunLoop(tight_loop);
 
     return status;
 }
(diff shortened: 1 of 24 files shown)


src/core/core.cpp
```diff
@@ -133,39 +113,22 @@ System::ResultStatus System::Load(EmuWindow* emu_window, const std::string& file
...
 System::ResultStatus System::Init(EmuWindow* emu_window, u32 system_mode) {
     NGLOG_DEBUG(HW_Memory, "initialized OK");
 
     CoreTiming::Init();
 
     current_process = Kernel::Process::Create("main");
 
-    if (Settings::values.use_cpu_jit) {
-#ifdef ARCHITECTURE_x86_64
-        cpu_core = std::make_shared<ARM_Dynarmic>();
```
Subv: The dynarmic and unicorn includes can be removed from this file

src/core/core_cpu.h
```diff
@@ -0,0 +1,74 @@
...
+
+class ARM_Interface;
+
+namespace Kernel {
+class Scheduler;
+}
+
+namespace Core {
+
+static constexpr unsigned NUM_CPU_CORES{4};
```
lioncash: const/constexpr values are static by default

src/core/core_cpu.h
```diff
@@ -0,0 +1,74 @@
...
+public:
+    Cpu(std::shared_ptr<CpuBarrier> cpu_barrier, size_t core_index);
+
+    void RunLoop(bool tight_loop = true);
+
+    void SingleStep();
+
+    void PrepareReschedule();
+
+    ARM_Interface& ArmInterface() {
```
lioncash: Should this also have a const overload?

bunnei: addressed

src/core/core.h
```diff
@@ -105,22 +108,24 @@ class System {
 
     PerfStats::Results GetAndResetPerfStats();
 
-    /**
-     * Gets a reference to the emulated CPU.
-     * @returns A reference to the emulated CPU.
-     */
-    ARM_Interface& CPU() {
-        return *cpu_core;
+    ARM_Interface& CurrentArmInterface() {
```
Subv: Add documentation for these functions

src/core/core_cpu.cpp
```diff
@@ -0,0 +1,119 @@
...
+        NGLOG_TRACE(Core, "Core-{} idling", core_index);
+
+        if (IsMainCore()) {
+            CoreTiming::Idle();
+            CoreTiming::Advance();
+        }
+
+        PrepareReschedule();
+    } else {
+        if (IsMainCore()) {
```
Subv: I worry this may cause issues if the main core is idling but the other cores are not.
It might lead to events firing too early, I'm not sure if this is going to be a problem

src/core/hle/kernel/thread.h
```diff
@@ -240,6 +245,11 @@ class Thread final : public WaitObject {
     // available. In case of a timeout, the object will be nullptr.
     std::function<WakeupCallback> wakeup_callback;
 
+    std::shared_ptr<Scheduler> scheduler;
+
+    u32 ideal_core{0xFFFFFFFF};
+    u64 mask{0x1};
```
Subv: Name this affinity_mask

src/core/core.cpp
bunnei: addressed

MerryMage:
> @MerryMage we are looking for guidance and/or an implementation to support this scenario across several Dynarmic instances.
I'll implement this soon™.

PR #443, opened 2018-05-17
Added IPC RequestWithContext & ControlWithContext

ogniK5377:
As far as I know, "Token" can be completely ignored and can still use our default IPC parser for Request & Control. This is needed for 5.x.x sysmodules and games too will need this soon enough when they start becoming 5.x.x

### DIFF
--- a/src/core/hle/ipc.h
+++ b/src/core/hle/ipc.h
@@ -32,6 +32,8 @@ enum class CommandType : u32 {
     Close = 2,
     Request = 4,
     Control = 5,
+    RequestWithContext = 6,
+    ControlWithContext = 7,
     Unspecified,
 };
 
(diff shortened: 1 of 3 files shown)


greggameplayer:
@ogniK5377 
Good PR. if you have the time can you add this to ipc code : 

0 | Invalid
1 | LegacyRequest
3 | LegacyControl

ogniK5377:
Is there anything that actually uses legacy?

greggameplayer:
@ogniK5377 according to switchbrew : 

> It is speculated that these are part of an older message processing system where headers were further partitioned

and according to switchbrew these are handled by calling these functions :

> - "nn::sf::hipc::detail::HipcMessageBufferAccessor::ParseHeader"
> - "nn::sf::hipc::server::HipcServerSessionManager::ProcessMessage"
> - "nn::sf::hipc::Reply"
> - "nn::sf::hipc::server::HipcServerSessionManager::RegisterServerSessionToWaitBase"

PR #455, opened 2018-05-20
Correct audio command numbers & add or rename some functions

greggameplayer:
I've made these changes according to switchbrew website : 
http://switchbrew.org/index.php?title=Audio_services

### DIFF
--- a/src/core/hle/service/audio/audin_u.cpp
+++ b/src/core/hle/service/audio/audin_u.cpp
@@ -35,11 +35,8 @@ public:
 
 AudInU::AudInU() : ServiceFramework("audin:u") {
     static const FunctionInfo functions[] = {
-        {0, nullptr, "ListAudioIns"},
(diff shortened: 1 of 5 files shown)


src/core/hle/service/audio/audin_u.cpp
```diff
@@ -38,6 +38,7 @@ AudInU::AudInU() : ServiceFramework("audin:u") {
         {0, nullptr, "ListAudioIns"},
         {1, nullptr, "OpenAudioIn"},
         {3, nullptr, "OpenAudioInAuto"},
+        {2, nullptr, "Unknown"},
```
Hexagon12: This has to be in the correct place, between 1 and 3.

greggameplayer: ok, i fix that and i correct the clang format error

greggameplayer:
@Hexagon12 done

PR #478, opened 2018-05-28
Use Ninja for Travis builds

janisozaur:
`ninja` is _the fast_ build system. It trades run-time complexity of `make` for simple rules with very limited logic that are generated by a more sentient system, like CMake. In practice that means more of CPUs' duty cycle is spent in the compiler and not in the build system resulting in faster builds.

[This build](https://travis-ci.org/yuzu-emu/yuzu/builds/384673023) completed in 13m47s, [one of the recent builds](https://travis-ci.org/yuzu-emu/yuzu/builds/384502149) took 14m19s. Note the improvement may not be as pronounced due to noisy environment and the quoted time including downloading and setup of build environment.

### DIFF
--- a/.travis/linux/docker.sh
+++ b/.travis/linux/docker.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -ex
 
 apt-get update
-apt-get install -y build-essential git libqt5opengl5-dev libsdl2-dev libssl-dev python qtbase5-dev wget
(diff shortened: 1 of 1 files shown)


NekokoTime:
Pull request #1 - Pull request #479 - Use Ninja for Travis builds

bunnei:
Can you add a PR description?

PR #483, opened 2018-05-30
Several GPU fixes to boot Sonic Mania

bunnei:
image


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -851,6 +852,10 @@ class GLSLGenerator {
                 regs.SetRegisterToInteger(instr.gpr0, instr.conversion.is_signed, 0, op_a, 1, 1);
                 break;
             }
+            case OpCode::Id::F2F_R: {
+                regs.SetRegisterToFloat(instr.gpr0, 0, regs.GetRegisterAsFloat(instr.gpr20), 1, 1);
```
Subv: Doesn't this have an abs flag in bit 49? We don't seem to assert or check it.

bunnei: oops mistake

greggameplayer:
this PR permit also Tiny Barbarian DX to boot but with weird textures

PR #484, opened 2018-05-30
Services/nvdrv: add '/dev/nvhost-nvdec' device

mailwl:
This device used in game Monster Energy Supercross and some other

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_nvdec.h
+++ b/src/core/hle/service/nvdrv/devices/nvhost_nvdec.h
@@ -1,0 +1,38 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <cstdlib>
+#include <cstring>
+#include <vector>
+#include "common/common_types.h"
+#include "core/hle/service/nvdrv/devices/nvdevice.h"
+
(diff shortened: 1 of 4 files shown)


Subv:
I would prefer it if we had a generic NvChannel class that implements the boilerplate ioctls and other channels just inherited from it (nvhost-gpu, nvhost-nvdec, nvhost-vic, etc)

greggameplayer:
@mailwl if you want to put more Ioctls Channel here is the list of these : 
http://switchbrew.org/index.php?title=NV_services#Channel_Ioctls

src/core/hle/service/nvdrv/devices/nvhost_nvdec.h
```diff
@@ -0,0 +1,38 @@
...
+
+class nvhost_nvdec final : public nvdevice {
+public:
+    nvhost_nvdec() = default;
+    ~nvhost_nvdec() override = default;
+
+    u32 ioctl(Ioctl command, const std::vector<u8>& input, std::vector<u8>& output) override;
+
+private:
+    enum class IoctlCommand : u32_le {
```
lioncash: This will break on a big-endian system, as `u32_le` won't be a primitive arithmetic type.

Hexagon12:
With this, 1-2-Switch now needs mm:u.

bunnei:
Seems fine to me... will leave it up to someone to make a separate change refactoring with a generic NvChannel class.

PR #487, opened 2018-05-30
add IPC CommandType & Some HID FunctionInfo

greggameplayer:
all these changes has been made according to switchbrew website.

## add IPC CommandType : 
- Invalid as cmd number 0
- LegacyRequest as cmd number 1
- LegacyControl as cmd number 3

## add some HID FunctionInfo : 
add functions useful for 4.0.0+ and 5.0.0+

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -329,6 +329,7 @@ public:
             {130, nullptr, "SwapNpadAssignment"},
             {131, nullptr, "IsUnintendedHomeButtonInputProtectionEnabled"},
             {132, nullptr, "EnableUnintendedHomeButtonInputProtection"},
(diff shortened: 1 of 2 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -329,19 +329,56 @@ class Hid final : public ServiceFramework<Hid> {
             {130, nullptr, "SwapNpadAssignment"},
             {131, nullptr, "IsUnintendedHomeButtonInputProtectionEnabled"},
             {132, nullptr, "EnableUnintendedHomeButtonInputProtection"},
+            {133, nullptr, "SetNpadJoyAssignmentModeSingleWithDestination"}, // 5.0.0+
```
Hexagon12: No need for the `// 5.0.0+` or similar marks here and below.

greggameplayer: ok i remove it

PR #488, opened 2018-05-30
Kernel/SVC: Corrected the behavior of svcSetThreadCoreMask for core values -2 and -3.

Subv:
-2 means 'Use the default core from the parent process'.
-3 means 'Don't change the ideal core, only update the mask'.

### DIFF
--- a/src/core/hle/kernel/errors.h
+++ b/src/core/hle/kernel/errors.h
@@ -21,7 +21,9 @@ enum {
 
     // Confirmed Switch OS error codes
     MisalignedAddress = 102,
+    InvalidProcessorId = 113,
     InvalidHandle = 114,
+    InvalidCombination = 116,
     Timeout = 117,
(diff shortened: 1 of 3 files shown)


mailwl:
maybe add 2 more changes?
assigned to itself: https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/kernel/thread.cpp#L463
comparing u32 and s32: https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/kernel/thread.cpp#L474

Subv:
@mailwl I changed the mask = mask assignment, nice catch! How did anything even work at all?
The signed/unsigned comparison is IMO out of scope for this particular PR

mailwl:
just check on online compuler, (u32)0xFFFFFFFF == -1, so, it's good. it's only (u32 < 0) check always false

PR #490, opened 2018-05-30
add check for GL extension support

### LINKED ISSUE #336: Yuzu should report / assert if a needed GL extension is required
We should report this (and most probably assert) if the user's GPU doesn't support a required extension. A current example of this is introduced in the constbuffer PR (#328) with the inclusion of the ARB_shader_storage_buffer_object extension.

BreadFish64:
Checks if the hardware supports supports the required OpenGL extensions before the game attempts to boot.
image
resolves #336

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -335,6 +335,24 @@ void GMainWindow::OnDisplayTitleBars(bool show) {
     }
 }
 
+bool GMainWindow::SupportsRequiredGLExtensions() {
+    QStringList unsupported_ext;
+
+    if (!GLAD_GL_ARB_program_interface_query)
+        unsupported_ext.append("ARB_program_interface_query");
+    if (!GLAD_GL_ARB_separate_shader_objects)
+        unsupported_ext.append("ARB_separate_shader_objects");
+    if (!GLAD_GL_ARB_shader_storage_buffer_object)
+        unsupported_ext.append("ARB_shader_storage_buffer_object");
+    if (!GLAD_GL_ARB_vertex_attrib_binding)
+        unsupported_ext.append("ARB_vertex_attrib_binding");
+
+    for (const QString& ext : unsupported_ext)
+        NGLOG_CRITICAL(Frontend, "Unsupported GL extension: {}", ext.toStdString());
+
+    return unsupported_ext.empty();
+}
+
 bool GMainWindow::LoadROM(const QString& filename) {
     // Shutdown previous session if the emu thread is still active...
     if (emu_thread != nullptr)
@@ -347,6 +365,14 @@ bool GMainWindow::LoadROM(const QString& filename) {
         QMessageBox::critical(this, tr("Error while initializing OpenGL 3.3 Core!"),
                               tr("Your GPU may not support OpenGL 3.3, or you do not "
                                  "have the latest graphics driver."));
+        return false;
+    }
+
+    if (!SupportsRequiredGLExtensions()) {
+        QMessageBox::critical(
+            this, tr("Error while initializing OpenGL Core!"),
+            tr("Your GPU may not support one or more required OpenGL extensions. Please "
+               "ensure you have the latest graphics driver. See the log for more details."));
         return false;
     }
 
--- a/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
+++ b/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
@@ -78,6 +78,24 @@ void EmuWindow_SDL2::Fullscreen() {
     SDL_MaximizeWindow(render_window);
 }
 
+bool EmuWindow_SDL2::SupportsRequiredGLExtensions() {
+    std::vector<std::string> unsupported_ext;
+
+    if (!GLAD_GL_ARB_program_interface_query)
+        unsupported_ext.push_back("ARB_program_interface_query");
+    if (!GLAD_GL_ARB_separate_shader_objects)
+        unsupported_ext.push_back("ARB_separate_shader_objects");
+    if (!GLAD_GL_ARB_shader_storage_buffer_object)
+        unsupported_ext.push_back("ARB_shader_storage_buffer_object");
+    if (!GLAD_GL_ARB_vertex_attrib_binding)
+        unsupported_ext.push_back("ARB_vertex_attrib_binding");
+
+    for (const std::string& ext : unsupported_ext)
+        NGLOG_CRITICAL(Frontend, "Unsupported GL extension: {}", ext);
+
+    return unsupported_ext.empty();
+}
+
 EmuWindow_SDL2::EmuWindow_SDL2(bool fullscreen) {
     InputCommon::Init();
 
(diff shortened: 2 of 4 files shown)


src/yuzu/main.cpp
```diff
@@ -350,6 +350,27 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
+        unsupported_ext.append("ARB_vertex_attrib_binding");
+
+    if (!unsupported_ext.empty()) {
+        for (QString ext : unsupported_ext)
+            NGLOG_CRITICAL(Frontend, "Unsupported GL extension: {}", ext.toStdString());
+        QMessageBox::critical(this, tr("Error while initializing OpenGL Core!"),
+                              tr("Your GPU does may not support one or more required OpenGL "
+                                 "extensions. Please ensure you "
+                                 "have the latest graphics driver. "
+                                 "See the "
```
CodingKoopa: These lines are broken kinda oddly?

BreadFish64: blame clang-format I guess?

CodingKoopa: You can edit it into one long string, and let `clang-format` reformat it.

BreadFish64: I see

src/yuzu/main.cpp
```diff
@@ -350,6 +350,25 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
 
+    QStringList unsupported_ext;
+
+    if (!GLAD_GL_ARB_program_interface_query)
+        unsupported_ext.append("ARB_program_interface_query");
+    if (!GLAD_GL_ARB_shader_storage_buffer_object)
+        unsupported_ext.append("ARB_shader_storage_buffer_object");
+    if (!GLAD_GL_ARB_vertex_attrib_binding)
+        unsupported_ext.append("ARB_vertex_attrib_binding");
+
```
bunnei: Should we also check for https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_separate_shader_objects.txt ?

src/yuzu/main.cpp
```diff
@@ -350,6 +350,25 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
+    if (!unsupported_ext.empty()) {
+        for (QString ext : unsupported_ext)
+            NGLOG_CRITICAL(Frontend, "Unsupported GL extension: {}", ext.toStdString());
+        QMessageBox::critical(
+            this, tr("Error while initializing OpenGL Core!"),
+            tr("Your GPU does may not support one or more required OpenGL extensions. Please "
+               "ensure you have the latest graphics driver. See the log for more details."));
```
bunnei: Is it worth mentioning yuzu generally requires a GPU with support for OpenGL 4.1?
(which I've never considered to be an unreasonable requirement, since Switch supports like OpenGL 4.6...)

src/yuzu/main.cpp
```diff
@@ -350,6 +350,27 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
+        unsupported_ext.append("ARB_shader_storage_buffer_object");
+    if (!GLAD_GL_ARB_vertex_attrib_binding)
+        unsupported_ext.append("ARB_vertex_attrib_binding");
+
+    if (!unsupported_ext.empty()) {
+        for (QString ext : unsupported_ext)
+            NGLOG_CRITICAL(Frontend, "Unsupported GL extension: {}", ext.toStdString());
+        QMessageBox::critical(
+            this, tr("Error while initializing OpenGL Core!"),
+            tr("Your GPU does may not support one or more required OpenGL extensions. Please "
```
Subv: `Your GPU does may not`

Subv:
What about yuzu-cmd?

NekokoTime:
This fix this other OpenGL problem
anotacao 7

greggameplayer:
@NekokoTime this has nothing in related with that problem and it's not a problem it's just a warning.

NekokoTime:
@greggameplayer oky I deleted it, I thought it was a problem.

BreadFish64:
@Subv checks added 👍

src/yuzu/main.cpp
Hexagon12: If we go higher than 4.1, macOS is gone.

src/yuzu/main.cpp
```diff
@@ -350,6 +350,27 @@ bool GMainWindow::LoadROM(const QString& filename) {
...
+        unsupported_ext.append("ARB_program_interface_query");
+    if (!GLAD_GL_ARB_separate_shader_objects)
+        unsupported_ext.append("ARB_separate_shader_objects");
+    if (!GLAD_GL_ARB_shader_storage_buffer_object)
+        unsupported_ext.append("ARB_shader_storage_buffer_object");
+    if (!GLAD_GL_ARB_vertex_attrib_binding)
+        unsupported_ext.append("ARB_vertex_attrib_binding");
+
+    if (!unsupported_ext.empty()) {
+        for (QString ext : unsupported_ext)
```
lioncash: This should be `const QString&`, since copies aren't necessary here.

src/yuzu/main.cpp
```diff
@@ -350,6 +350,27 @@ bool GMainWindow::LoadROM(const QString& filename) {
         return false;
     }
 
+    QStringList unsupported_ext;
```
lioncash: This should all probably go within it's own local function:

```cpp
bool SupportsRequiredGLExtensions()
```
which is then called here like:

```cpp
if (!SupportsRequiredGLExtensions()) {
    // Display error
}
```

src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
```diff
@@ -128,6 +128,23 @@ EmuWindow_SDL2::EmuWindow_SDL2(bool fullscreen) {
         exit(1);
     }
 
+    std::vector<std::string> unsupported_ext;
```
lioncash: Ditto about being its own function

src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
```diff
@@ -128,6 +128,23 @@ EmuWindow_SDL2::EmuWindow_SDL2(bool fullscreen) {
...
+        unsupported_ext.push_back("ARB_program_interface_query");
+    if (!GLAD_GL_ARB_separate_shader_objects)
+        unsupported_ext.push_back("ARB_separate_shader_objects");
+    if (!GLAD_GL_ARB_shader_storage_buffer_object)
+        unsupported_ext.push_back("ARB_shader_storage_buffer_object");
+    if (!GLAD_GL_ARB_vertex_attrib_binding)
+        unsupported_ext.push_back("ARB_vertex_attrib_binding");
+
+    if (!unsupported_ext.empty()) {
+        for (std::string ext : unsupported_ext)
```
lioncash: `const std::string&`

bunnei:
@BreadFish64 can you please address the above comments?

PR #492, opened 2018-05-31
Service/time: implement posix time to calendar conversion

mailwl:
addittionally, some changes in service/time structures

### DIFF
--- a/src/core/hle/service/time/time.cpp
+++ b/src/core/hle/service/time/time.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <chrono>
+#include <ctime>
 #include "common/logging/log.h"
 #include "core/core_timing.h"
 #include "core/hle/ipc_helpers.h"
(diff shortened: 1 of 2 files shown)


src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -350,7 +350,7 @@ std::string HLERequestContext::Description() const {
         }
         s << ']';
     }
-    s << ", data_size:" << command_header->data_size.Value();
+    s << ", data_size: 0x" << command_header->data_size.Value();
```
Hexagon12: Unintentional change?

mailwl: yes, data_size is printed as hex, and may confuse. will remove

src/core/hle/service/time/time.cpp
```diff
@@ -103,23 +106,69 @@ class ITimeZoneService final : public ServiceFramework<ITimeZoneService> {
...
         rb.Push(RESULT_SUCCESS);
         rb.PushRaw(calendar_time);
         rb.PushRaw(additional_info);
     }
+
+    void PosixToCalendar(u64 posix_time, CalendarTime& calendar_time,
+                         CalendarAdditionalInfo& additional_info, const TimeZoneRule& /*rule*/) {
+        std::time_t t(posix_time);
+        std::tm* tm = std::localtime(&t);
+        if (tm) {
```
lioncash:
```cpp
if (!tm) {
    return;
}
```

mailwl:
fixed

PR #497, opened 2018-06-02
GPU: Implemented the DXN1 (BC4) texture format.

Subv:


greggameplayer:
@Subv can you implement DXN2 which is the TextureFormat value 0x28 ?

greggameplayer:
@Subv bit width of DXN2 also known as BC5 are about 128, according to : 
https://msdn.microsoft.com/en-us/library/windows/desktop/bb694531(v=vs.85).aspx

greggameplayer:
and there is some informations about some other textures formats to implement like R32, R32_G32

Subv:
`@Subv can you implement also with this PR DXN2 which is the TextureFormat value 0x28 ?`
No.
It'll be implemented when we find a game that uses it, and in a separate PR.

greggameplayer:
@Subv ok no problem, but we should add all the textures formats even if we don't know a game that use them.

PR #499, opened 2018-06-03
am: Implement CreateStorage, PushInData, etc.

bunnei:
Basic RE work done by @ogniK5377 is the basis for this implementation. These were tested with SMO, which now gets to drawing code with these changes.

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <cinttypes>
+#include <stack>
 #include "core/file_sys/filesystem.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/event.h"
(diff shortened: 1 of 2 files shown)


greggameplayer:
@bunnei also seem to permit Rayman Legends Definitive Edition to go further

greggameplayer:
@bunnei for me Super Mario Odyssey work the same as before

src/core/hle/service/am/am.cpp
```diff
@@ -412,72 +505,17 @@ void ILibraryAppletCreator::CreateLibraryApplet(Kernel::HLERequestContext& ctx)
...
+void ILibraryAppletCreator::CreateStorage(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+    const u64 size{rp.Pop<u64>()};
+    std::vector<u8> buffer(size);
 
-        rb.Push(RESULT_SUCCESS);
-        rb.PushIpcInterface<AM::IStorageAccessor>(buffer);
+    IPC::ResponseBuilder rb{rp.MakeBuilder(2, 0, 1)};
+    rb.Push(RESULT_SUCCESS);
+    rb.PushIpcInterface<AM::IStorage>(buffer);
```
Subv: std::move

src/core/hle/service/am/am.cpp
```diff
@@ -348,6 +349,87 @@ void ICommonStateGetter::GetPerformanceMode(Kernel::HLERequestContext& ctx) {
...
+    }
+
+private:
+    std::vector<u8> buffer;
+
+    void Open(Kernel::HLERequestContext& ctx) {
+        IPC::ResponseBuilder rb{ctx, 2, 0, 1};
+
+        rb.Push(RESULT_SUCCESS);
+        rb.PushIpcInterface<AM::IStorageAccessor>(buffer);
```
Subv: Each call to this function copies the buffer?

bunnei: For now, yes - until we understand better how this works. Note, the behavior of this was not changed with this PR.

PR #500, opened 2018-06-04
GPU: Partial implementation of long GPU queries.

Subv:
Long queries write a 128-bit result value to memory, which consists of a 64 bit query value and a 64 bit timestamp.

In this implementation, only select=Zero of the Crop unit is implemented, this writes the query sequence as a 64 bit value, and a 0u64 value for the timestamp, since we emulate an infinitely fast GPU.

This specific type was hwtested, but more rigorous tests should be performed in the future for the other types.

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -156,16 +156,15 @@ void Maxwell3D::ProcessQueryGet() {
     // TODO(Subv): Support the other query units.
     ASSERT_MSG(regs.query.query_get.unit == Regs::QueryUnit::Crop,
                "Units other than CROP are unimplemented");
-    ASSERT_MSG(regs.query.query_get.short_query,
-               "Writing the entire query result structure is unimplemented");
 
     u32 value = Memory::Read32(*address);
-    u32 result = 0;
+    u64 result = 0;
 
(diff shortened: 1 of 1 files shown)


greggameplayer:
make 1.2.Switch in single core go on the intro but it crash after the warning screen with tried to unwrap empty ResultVal

greggameplayer:
also make Arms go further only in single core now we need to implement texture format 23 which is X8Z24_X20V4S8__COV8R8V. Now all the game in the previous builds with the error "writing the entire query result structure is unimplemented" which are now getting further on single core does all in multi cores "shader not linked"error.

Hexagon12:
> tried to unwrap empty ResultVal

It wants to create save data. Relaunch the game and it should work like normal.

>implement texture format 23 which is X8Z24_X20V4S8__COV8R8V

That is incorrect. Yuzu displays texture types as decimal, but the textures in SwitchBrew are in hex. Anyway, it actually needs BC7U (0x17).

greggameplayer:
@Hexagon12 i have correct that i realize this later after the comment.

greggameplayer:
@Hexagon12 no it don't work it does this error all the time for example on mario+rabbids kingdom battle even if i relaunch the game

Issue #493, opened 2018-05-31
shader not linked [Error]


greggameplayer:
## Hardware Specs : 
- Nvidia GeForce GTX 960
- Intel I7-7700K @ 4.5Ghz
- 8 GB Of Ram 3200Mhz
- Windows 10 Pro

## The issue : 
when you launch 1.2.Switch on multi-cores, the game crash with shader not linked.
but on single-core this error doesn't appear.

## List of games with this error : 
- 1.2.Switch
- Arms
- Lego World
- Lego Ninjago the movie videogame
- Super Mario Odyssey

## The last part of the log : 
![Log Image](https://image.noelshack.com/fichiers/2018/22/5/1527809392-capture2.png)

Hexagon12:
You can't get to the opening screen, right? What yuzu version are you using?

greggameplayer:
yes and I'm currently using HEAD-bdd68fc.

Elenedeath:
same for me with a similar configuration
Nvidia GTX 970
Intel I7-6700K
24 go de ram@3000mhz
W10 Pro 64bits

greggameplayer:
@Elenedeath canary build and nightly build are the same currently so it's that why The binding of isaac run at the same fps on both builds types.

greggameplayer:
this error appear on newer versions too.

Elenedeath:
I found why i got the error on all version, it's due to check the "enable multicore" option.
when you uncheck it, the error disappear but i got another error

"[  12.394379] Debug <Critical> video_core/engines/maxwell_3d.cpp:operator():160: Assertion Failed!
Writing the entire query result structure is unimplemented"

greggameplayer:
@Elenedeath this error is due to the fact that not all the maxwell 3d engine have been implement

Subv:
The 'Shader not linked' error when multicore is enabled is caused by yuzu trying to perform OpenGL state manipulation from a thread that isn't the current OpenGL thread. I don't know if the games are submitting commands from multiple threads or if they're just submitting commands from a single thread that isn't running in the host EmuThread.

greggameplayer:
@Subv currently yuzu submitting commands from a single thread or multiple ?

Hexagon12:
This issue is fixed now.

Requesting to close.

PR #503, opened 2018-06-04
Service/nfp:user : stub some functions.

mailwl:
Used by Zelda: BoTW

### DIFF
--- a/src/core/hle/service/nfp/nfp.cpp
+++ b/src/core/hle/service/nfp/nfp.cpp
@@ -5,6 +5,7 @@
 #include "common/logging/log.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/event.h"
+#include "core/hle/service/hid/hid.h"
 #include "core/hle/service/nfp/nfp.h"
 #include "core/hle/service/nfp/nfp_user.h"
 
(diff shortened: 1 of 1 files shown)


src/core/hle/service/nfp/nfp.cpp
```diff
@@ -33,24 +34,87 @@ class IUser final : public ServiceFramework<IUser> {
...
+
         IPC::ResponseBuilder rb{ctx, 2};
         rb.Push(RESULT_SUCCESS);
     }
+
+    void ListDevices(Kernel::HLERequestContext& ctx) {
+        NGLOG_WARNING(Service_NFP, "(STUBBED) called");
+        IPC::ResponseBuilder rb{ctx, 3};
+        rb.Push(RESULT_SUCCESS);
+        rb.Push<u32>(0);
```
Hexagon12: Any idea why we're pushing 0 here? We are basically saying that there's no NFC controllers on the system.

src/core/hle/service/nfp/nfp.cpp
```diff
@@ -33,24 +34,87 @@ class IUser final : public ServiceFramework<IUser> {
...
+        IPC::ResponseBuilder rb{ctx, 3};
+        rb.Push(RESULT_SUCCESS);
+        rb.Push<u32>(static_cast<u32>(device_state));
+    }
+
+    void GetNpadId(Kernel::HLERequestContext& ctx) {
+        NGLOG_WARNING(Service_NFP, "(STUBBED) called");
+        IPC::ResponseBuilder rb{ctx, 3};
+        rb.Push(RESULT_SUCCESS);
+        rb.Push<u32>(0);
```
Hexagon12: 0 probably isn't a valid ID. 1 would be better.

Hexagon12:
With this, BotW now tries rendering only on multicore, but since #493 is active, you would get the `Shader not linked` error.

mailwl:
how about Kirby? it has tons unimplemented functions, i don't even get to `AttachAvailabilityChangeEvent`

mailwl:
it's not enough, latter - unimplemented Ioctl2

src/core/hle/service/nfp/nfp.cpp
```diff
@@ -33,24 +35,116 @@ class IUser final : public ServiceFramework<IUser> {
...
+        IPC::RequestParser rp{ctx};
+        const u32 array_size = rp.Pop<u32>();
+
+        ctx.WriteBuffer(&device_handle, sizeof(device_handle));
+
+        NGLOG_WARNING(Service_NFP, "(STUBBED) called, array_size={}", array_size);
+
+        IPC::ResponseBuilder rb{ctx, 3};
+        rb.Push(RESULT_SUCCESS);
+        rb.Push<u32>(1);
```
Hexagon12: Please revert this back to 0.

PR #504, opened 2018-06-04
GDB Stub should work now.

Hedges:


bunnei:
@Hedges just curious (without looking much at the code) - What is the reason why this was closed?

Hedges:
I had to clang-format and make it platform agnostic, re-submitted now.

PR #516, opened 2018-06-04
GPU: Implemented the F2I_R shader instruction.

Subv:
The float rounding modes in F2F are unimplemented and will assert.


greggameplayer:
@Subv good implementation this is needed for splatoon 2

PR #522, opened 2018-06-05
Service/MM: add service and stub some functions

mailwl:
Used in some games, for example, Monster Energy Supercross

### DIFF
--- a/src/core/hle/service/mm/mm_u.cpp
+++ b/src/core/hle/service/mm/mm_u.cpp
@@ -1,0 +1,50 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <sstream>
+#include <string>
+#include "common/logging/log.h"
+#include "core/hle/ipc_helpers.h"
+#include "core/hle/kernel/client_session.h"
+#include "core/hle/service/mm/mm_u.h"
+
+namespace Service::MM {
+
+void InstallInterfaces(SM::ServiceManager& service_manager) {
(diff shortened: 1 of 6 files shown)


src/core/hle/service/mm/mm_u.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <sstream>
```
lioncash: `<sstream>` and `<string>` aren't used in this file.

src/core/hle/service/mm/mm_u.h
```diff
@@ -0,0 +1,29 @@
...
+public:
+    MM_U();
+    ~MM_U() = default;
+
+private:
+    void Initialize(Kernel::HLERequestContext& ctx);
+    void SetAndWait(Kernel::HLERequestContext& ctx);
+    void Get(Kernel::HLERequestContext& ctx);
+
+    u32 value;
```
lioncash: This should have a default initializer and also have a comment explaining what it's used for, given this is a pretty generic name.

src/core/hle/service/mm/mm_u.h
```diff
@@ -0,0 +1,29 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <vector>
```
lioncash: `<vector>` isn't used in this header.

PR #526, opened 2018-06-05
Use Ninja for MSVC AppVeyor builds

janisozaur:


janisozaur:
Cool, many thanks for merging it. I kinda expected some comments or questions.

I hope it will reduce the turnaround time for you.

bunnei:
@janisozaur this change has been reverted, because with it, necessary runtime libraries (SDL and Qt DLLs) were no longer being copied to the binary directories. Please correct this issue and resubmit the change - thank you!

bunnei:
@janisozaur given that the PR was open for over two weeks without any comments, I would suspect that no one really has a strong opinion, nor the build system expertise to advise on this (which was my case, anyways).

PR #532, opened 2018-06-06
 gl_shader_decompiler: Implement LD_C instruction.

bunnei:


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -654,7 +654,18 @@ u32 RasterizerOpenGL::SetupConstBuffers(Maxwell::ShaderStage stage, GLuint progr
         buffer_draw_state.bindpoint = current_bindpoint + bindpoint;
 
         boost::optional<VAddr> addr = gpu.memory_manager->GpuToCpuAddress(buffer.address);
-        std::vector<u8> data(used_buffer.GetSize() * sizeof(float));
+
+        std::vector<u8> data;
+        if (used_buffer.IsUsed()) {
```
Subv: What's this check for? if it's in the entries vector then it's used.

src/video_core/engines/shader_bytecode.h
```diff
@@ -354,12 +349,21 @@ union Instruction {
         }
     } bra;
 
+    union {
+        BitField<20, 14, u64> offset;
+        BitField<34, 5, u64> index;
+    } cbuf34;
+
+    union {
+        BitField<20, 16, u64> offset;
```
Subv: This one should use s64, the 16 bit offset is a signed integer.

PR #533, opened 2018-06-06
Common/string_util: add StringFromBuffer() function

mailwl:
convert input buffer (`std::vector<u8>`) to string, stripping zero chars


mailwl:
Fixed

PR #554, opened 2018-06-09
Rasterizer: Use UBOs instead of SSBOs for uploading const buffers.

Subv:
This should help a bit with GPU performance once we're GPU-bound.

### DIFF
--- a/src/video_core/renderer_opengl/gl_state.cpp
+++ b/src/video_core/renderer_opengl/gl_state.cpp
@@ -223,7 +223,7 @@ void OpenGLState::Apply() const {
             if (current.enabled != new_state.enabled || current.bindpoint != new_state.bindpoint ||
                 current.ssbo != new_state.ssbo) {
                 if (new_state.enabled) {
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -223,7 +223,7 @@ void OpenGLState::Apply() const {
             if (current.enabled != new_state.enabled || current.bindpoint != new_state.bindpoint ||
                 current.ssbo != new_state.ssbo) {
                 if (new_state.enabled) {
-                    glBindBufferBase(GL_SHADER_STORAGE_BUFFER, new_state.bindpoint, new_state.ssbo);
+                    glBindBufferBase(GL_UNIFORM_BUFFER, new_state.bindpoint, new_state.ssbo);
```
NekokoTime: (GL_SHADER_STORAGE_BUFFER, 3, ssbo);

greggameplayer: This comment has no sense please delete it.

bunnei:
Adding canary-merge to make this easier to test. Will merge tomorrow if I don't find any regressions.

PR #556, opened 2018-06-10
GPU: Partially implemented the Maxwell DMA engine.

Subv:
Only tiled->linear and linear->tiled copies that aren't offsetted are supported for now. 
Queries are not supported.
Swizzled copies are not supported.


MoochMcGee:
you should probably implement linear->linear copies too, while you're at it, as it seems relatively foolproof.

src/video_core/engines/maxwell_dma.h
```diff
@@ -0,0 +1,162 @@
...
+#include "common/bit_field.h"
+#include "common/common_funcs.h"
+#include "common/common_types.h"
+#include "video_core/gpu.h"
+#include "video_core/memory_manager.h"
+
+namespace Tegra {
+namespace Engines {
+
+#define MAXWELLDMA_REG_INDEX(field_name)                                                           \
```
lioncash: This can probably go in the cpp file to keep it localized

PR #557, opened 2018-06-12
hid: Update all layouts and only show handheld as connected, fixes libnx input for P1_AUTO

shinyquagsire23:
Probably breaks SMO but it's not like the input is working there anyhow, find a better "fix" pls.


Hexagon12:
This fixes both LibNX input AND SMO.

mailwl:
@Hexagon12 i don't understand the nature of fix, but i think yuzu is nintendo switch emulator, not homebrew emulator.
(i don't with this PR, just thinking out loud)

PR #558, opened 2018-06-12
GPU: Implemented the iadd32i shader instruction.

Subv:


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1005,6 +1005,28 @@ class GLSLGenerator {
...
+            std::string op_a = regs.GetRegisterAsInteger(instr.gpr8);
+
+            if (instr.iadd32i.negate_a)
+                op_a = '-' + op_a;
+
+            std::string op_b = '(' + std::to_string(instr.alu.imm20_32.Value()) + ')';
+
+            switch (opcode->GetId()) {
+            case OpCode::Id::IADD32I:
+                regs.SetRegisterToInteger(instr.gpr0, true, 0, op_a + " + " + op_b, 1, 1);
```
bunnei: Pass bit 54 (saturate) into `SetRegisterToInteger` we get the assert there

bunnei: (and it's a smaller diff when we go to implement this)

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1005,6 +1005,28 @@ class GLSLGenerator {
             break;
         }
 
+        case OpCode::Type::ArithmeticIntegerImmediate: {
+            ASSERT_MSG(!instr.iadd32i.saturate, "Unimplemented");
+
+            std::string op_a = regs.GetRegisterAsInteger(instr.gpr8);
+
+            if (instr.iadd32i.negate_a)
+                op_a = '-' + op_a;
```
bunnei: This isn't used by all decodings of iadd32i, right? e.g. https://github.com/envytools/envytools/blob/master/envydis/gm107.c#L2118

Subv: It is, the first decoding just has this hardcoded to 0

PR #559, opened 2018-06-12
Services/FS: Return the correct error code when trying to mount a nonexistent savedata.

Subv:
We still have to test what exactly AM::EnsureSaveData does.


Hexagon12:
Please fix Clang-format.

PR #560, opened 2018-06-13
Qt: Removed the Registers widget.

Subv:
It was crashing and nobody actually uses this.


bunnei:
No sense keeping cruft around "just because it might be used some day". Someone can always add it back if they have a use for it in the future (which I'd guess is unlikely).

PR #561, opened 2018-06-13
Avoid initializing single-joycon layouts with handheld controller

DarkLordZach:
This fixes the odyssey crash introduced by #557.

(this explanation might be 100% wrong sorry)

So, when initializing the layouts for controllers in the hid module, right now it initializes all layouts on all controllers. However, it is not possible with the hardware to have a single joycon layout (id # 3, 4, or 5) with the handheld controller (Switch OS does not allow it). I added a check in the loop to see if its the handheld and if it is, not try to make the extra layouts. Apparently odyssey uses this and crashes on it for some reason.

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -16,6 +16,14 @@
 
 namespace Service::HID {
 
+int GetNumberLayoutsForController(size_t controller_id) {
+    if (controller_id == Controller_Handheld) {
+        return HID_NUM_LAYOUTS_HANDHELD;
+    } else {
+        return HID_NUM_LAYOUTS;
+    }
+}
+
 // Updating period for each HID device.
 // TODO(shinyquagsire23): These need better values.
 constexpr u64 pad_update_ticks = CoreTiming::BASE_CLOCK_RATE / 10000;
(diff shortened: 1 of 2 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -83,7 +83,10 @@ class IAppletResource final : public ServiceFramework<IAppletResource> {
         controller_header.left_color_buttons = JOYCON_BUTTONS_NEON_BLUE;
 
         for (size_t controller = 0; controller < mem.controllers.size(); controller++) {
-            for (int index = 0; index < HID_NUM_LAYOUTS; index++) {
+            for (int index = 0;
+                 index <
+                 (controller != Controller_Handheld ? HID_NUM_LAYOUTS : HID_NUM_LAYOUTS_HANDHELD);
```
lioncash: You may want to break this out into an appropriate function or something that returns the value based on a given value. This looks pretty gross as a loop continuation condition.

DarkLordZach: ok

src/core/hle/service/hid/hid.cpp
```diff
@@ -16,6 +16,14 @@
 
 namespace Service::HID {
 
+int GetNumberLayoutsForController(size_t controller_id) {
```
lioncash: `static int ...`

shinyquagsire23:
I'm gonna say this once and for any other PRs which try and do this, there are 7 layouts, there are always 7 layouts, and any less is incorrect. This will break libnx on the default layout, which had index 6. Find a better fix.

lioncash:
Given @shinyquagsire23 has done RE work on the HID stuff, I'm more inclined to agree with them (and, well, the fact this would break libnx, which I didn't think of at the time initially).

DarkLordZach:
Let me know when you finish that hw test @shinyquagsire23

bunnei:
This seems to fix SMO without breaking libnx, so seems fine to me

shinyquagsire23:
LGTM, though I'd like to figure out a better fix but I can figure that out on my own time I suppose. Hopefully nothing really needs that layout I guess.

PR #564, opened 2018-06-15
gl_shader_decompiler: Implement LOP32I LogicOperation PassB.

bunnei:


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -946,6 +946,10 @@ class GLSLGenerator {
                                               '(' + op_a + " ^ " + std::to_string(imm) + ')', 1, 1);
                     break;
                 }
+                case Tegra::Shader::LogicOperation::PassB: {
+                    regs.SetRegisterToInteger(instr.gpr0, true, 0, op_a, 1, 1);
```
Subv: I'm rather sure this is wrong, the PassB operation should assign the value of the second operand, not the first

PR #566, opened 2018-06-15
gl_shader_gen: Set position.w to 1.

bunnei:


Subv:
Are we really sure this is needed? I was under the impression that position.w would end up as 1.0 anyway in the fragment shader after interpolation

PR #568, opened 2018-06-17
gl_shader_decompiler: Implement LOP instructions.

bunnei:


src/video_core/engines/shader_bytecode.h
```diff
@@ -229,6 +229,12 @@ union Instruction {
             BitField<42, 1, u64> negate_pred;
         } fmnmx;
 
+        union {
+            BitField<39, 1, u64> invert_a;
+            BitField<40, 1, u64> invert_b;
+            BitField<41, 2, LogicOperation> operation;
```
Subv: You're missing the PRED48 field, i don't know what it does, but you should at least add an assert about it.
Same with the 5c40_1 tab.

PR #575, opened 2018-06-19
Implement GetAvailableLanguageCodes2

greggameplayer:
tested on Kirby Star Allies

### DIFF
--- a/src/core/hle/service/set/set.cpp
+++ b/src/core/hle/service/set/set.cpp
@@ -12,9 +12,6 @@
 namespace Service::Set {
 
 void SET::GetAvailableLanguageCodes(Kernel::HLERequestContext& ctx) {
-    IPC::RequestParser rp{ctx};
-    u32 id = rp.Pop<u32>();
-
     static constexpr std::array<LanguageCode, 17> available_language_codes = {{
         LanguageCode::JA,
         LanguageCode::EN_US,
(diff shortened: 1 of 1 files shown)


src/core/hle/service/set/set.cpp
```diff
@@ -43,14 +43,46 @@ void SET::GetAvailableLanguageCodes(Kernel::HLERequestContext& ctx) {
     NGLOG_DEBUG(Service_SET, "called");
 }
 
+void SET::GetAvailableLanguageCodes2(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+    u32 id = rp.Pop<u32>();
```
mailwl: unused, maybe log this?

FearlessTobi: This is basically a copy of the GetAvailableLanguageCodes implementation and the value doesn't get logged there. So if it should be changed, then in both functions.

greggameplayer: ok so i delete this in both functions

src/core/hle/service/set/set.cpp
```diff
@@ -12,9 +12,35 @@
...
+    ctx.WriteBuffer(available_language_codes.data(), available_language_codes.size());
+
+    IPC::ResponseBuilder rb{ctx, 4};
+    rb.Push(RESULT_SUCCESS);
+    rb.Push(static_cast<u64>(available_language_codes.size()));
+
+    NGLOG_DEBUG(Service_SET, "called");
+}
 
+void SET::GetAvailableLanguageCodes2(Kernel::HLERequestContext& ctx) {
```
bunnei: This is just a copy paste of `GetAvailableLanguageCodes` .. just call that directly if it's the same function

greggameplayer: ok just the buffer type is different

greggameplayer:
bug in squash

greggameplayer:
@bunnei done

PR #578, opened 2018-06-21
[RFC] Virtual Filesystem

DarkLordZach:
So this obviously isn't done, just wanted to see if i was going in a good direction before integrating it into various other classes. The idea is that if the files are abstracted away from classes that use them, it is easier to implement things such as layeredfs, dlc, updates, and encryption.

### DIFF
--- a/src/core/file_sys/vfs.cpp
+++ b/src/core/file_sys/vfs.cpp
@@ -35,23 +35,25 @@ VfsDirectory::operator bool() {
 std::shared_ptr<VfsFile> VfsDirectory::GetFile(const std::string& name) {
     auto files = GetFiles();
     auto iter = std::find_if(files.begin(), files.end(),
-                             [&name](auto file1) { return name == file1.GetName(); });
+                             [&name](auto file1) { return name == file1->GetName(); });
     return iter == files.end() ? nullptr : std::move(*iter);
 }
 
 std::shared_ptr<VfsDirectory> VfsDirectory::GetSubdirectory(const std::string& name) {
     auto subs = GetSubdirectories();
     auto iter = std::find_if(subs.begin(), subs.end(),
-                             [&name](auto file1) { return name == file1.GetName(); });
+                             [&name](auto file1) { return name == file1->GetName(); });
     return iter == subs.end() ? nullptr : std::move(*iter);
 }
 
 u64 VfsDirectory::GetSize() {
     auto files = GetFiles();
-    auto file_total = std::accumulate(files.begin(), files.end(), 0);
+    auto file_total = std::accumulate(files.begin(), files.end(), 0ull,
+                                      [](auto f1, auto f2) { return f1 + f2->GetSize(); });
 
     auto sub_dir = GetSubdirectories();
-    auto subdir_total = std::accumulate(sub_dir.begin(), sub_dir.end(), 0);
+    auto subdir_total = std::accumulate(sub_dir.begin(), sub_dir.end(), 0ull,
+                                        [](auto f1, auto f2) { return f1 + f2->GetSize(); });
 
     return file_total + subdir_total;
 }
--- a/src/core/file_sys/vfs.cpp
+++ b/src/core/file_sys/vfs.cpp
@@ -1,0 +1,7 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/file_sys/vfs.h"
+
+namespace FileSys {}
--- a/src/core/file_sys/vfs.cpp
+++ b/src/core/file_sys/vfs.cpp
@@ -2,72 +2,110 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
 #include <numeric>
 #include "core/file_sys/vfs.h"
 
 namespace FileSys {
 
-VfsFile::operator bool() {
-    return IsGood();
+VfsFile::~VfsFile() = default;
+
+VfsDirectory::~VfsDirectory() = default;
+
+boost::optional<u8> VfsFile::ReadByte(size_t offset) const {
+    u8 out{};
+    size_t size = Read(&out, 1, offset);
+    if (size == 1)
+        return out;
+
+    return boost::none;
 }
 
-boost::optional<u8> VfsFile::ReadByte(u64 offset) {
-    auto vec = ReadBytes(offset, 1);
-    if (vec.empty())
-        return boost::none;
-    return vec[0];
+std::vector<u8> VfsFile::ReadBytes(size_t size, size_t offset) const {
+    std::vector<u8> out(size);
+    size_t read_size = Read(out.data(), size, offset);
+    out.resize(read_size);
+    return out;
 }
 
-std::vector<u8> VfsFile::ReadAllBytes() {
-    return ReadBytes(0, GetSize());
+std::vector<u8> VfsFile::ReadAllBytes() const {
+    return ReadBytes(GetSize());
 }
 
-u64 VfsFile::ReplaceBytes(const std::vector<u8>& data) {
-    if (!Resize(data.size()))
-        return 0;
-    return WriteBytes(data, 0);
+bool VfsFile::WriteByte(u8 data, size_t offset) {
+    return 1 == Write(&data, 1, offset);
 }
 
-VfsDirectory::operator bool() {
-    return IsGood();
+size_t VfsFile::WriteBytes(std::vector<u8> data, size_t offset) {
+    return Write(data.data(), data.size(), offset);
 }
 
-std::shared_ptr<VfsFile> VfsDirectory::GetFile(const std::string& name) {
-    auto files = GetFiles();
-    auto iter = std::find_if(files.begin(), files.end(),
-                             [&name](auto file1) { return name == file1->GetName(); });
-    return iter == files.end() ? nullptr : std::move(*iter);
+std::shared_ptr<VfsFile> VfsDirectory::GetFileRelative(const std::filesystem::path& path) const {
+    if (path.parent_path() == path.root_path())
+        return GetFile(path.filename().string());
+
+    auto parent = path.parent_path().string();
+    parent.replace(path.root_path().string().begin(), path.root_path().string().end(), "");
+    const auto index = parent.find_first_of('\\');
+    const auto first_dir = parent.substr(0, index), rest = parent.substr(index + 1);
+    const auto sub = GetSubdirectory(first_dir);
+    if (sub == nullptr)
+        return nullptr;
+    return sub->GetFileRelative(path.root_path().string() + rest);
 }
 
-std::shared_ptr<VfsDirectory> VfsDirectory::GetSubdirectory(const std::string& name) {
-    auto subs = GetSubdirectories();
-    auto iter = std::find_if(subs.begin(), subs.end(),
-                             [&name](auto file1) { return name == file1->GetName(); });
+std::shared_ptr<VfsFile> VfsDirectory::GetFileAbsolute(const std::filesystem::path& path) const {
+    if (IsRoot())
+        return GetFileRelative(path);
+
+    return GetParentDirectory()->GetFileAbsolute(path);
+}
+
+std::shared_ptr<VfsFile> VfsDirectory::GetFile(const std::string& name) const {
+    const auto& files = GetFiles();
+    const auto& iter = std::find_if(files.begin(), files.end(), [&name](const auto& file1) {
+        return name == file1->GetName();
+    });
+    return iter == files.end() ? nullptr : *iter;
+}
+
+std::shared_ptr<VfsDirectory> VfsDirectory::GetSubdirectory(const std::string& name) const {
+    const auto& subs = GetSubdirectories();
+    const auto& iter = std::find_if(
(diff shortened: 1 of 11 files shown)


src/core/file_sys/vfs.h
```diff
@@ -0,0 +1,105 @@
...
+    ~VfsFile();
+};
+
+struct VfsDirectory : NonCopyable {
+    virtual bool IsReady() = 0;
+    virtual bool IsGood() = 0;
+    virtual operator bool();
+    virtual void ResetState() = 0;
+
+    virtual std::vector<std::shared_ptr<VfsFile>> GetFiles() = 0;
```
lioncash: Some of these getters are likely ideally const qualified as well.

src/core/file_sys/vfs.h
```diff
@@ -0,0 +1,105 @@
...
+    virtual std::shared_ptr<VfsFile> GetFile(const std::string& name);
+
+    virtual std::vector<std::shared_ptr<VfsDirectory>> GetSubdirectories() = 0;
```
lioncash: Ditto

src/core/file_sys/vfs_offset.cpp
```diff
@@ -0,0 +1,69 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/file_sys/vfs_offset.h"
+
+namespace FileSys {
+
+OffsetVfsFile::OffsetVfsFile(std::unique_ptr<VfsFile>&& file, u64 offset, u64 size)
+    : file(std::move(file)), offset(offset), size(size) {}
```
lioncash: this will cause variable shadowing warnings, the parameters should be `file_`, `offset_`, and `size_`

src/core/file_sys/vfs_offset.h
```diff
@@ -0,0 +1,32 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+#include "core/file_sys/vfs.h"
```
lioncash: We usually add a newline after pragma once.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/common_paths.h"
+#include "common/logging/log.h"
+#include "core/file_sys/vfs_real.h"
+
+namespace FileSys {
+
+RealVfsFile::RealVfsFile(const std::string& path, const char openmode[])
```
lioncash: Ditto about variable shadowing warnings.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+void RealVfsFile::ResetState() {
+    backing.Clear();
+}
+
+std::string RealVfsFile::GetName() {
+    std::string part_path, name, extention;
+    Common::SplitPath(path, &part_path, &name, &extention);
+    return name + "." + extention;
```
lioncash: `'.'`

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+RealVfsDirectory::RealVfsDirectory(const std::string& path, const char openmode[])
+    : path(path[path.size() - 1] == DIR_SEP_CHR ? path.substr(0, path.size() - 1) : path) {
+    FileUtil::FSTEntry entry;
+    if (0 != FileUtil::ScanDirectoryTree(path, entry))
+        NGLOG_CRITICAL(Service_FS, "Failure to initialize file.");
+    for (FileUtil::FSTEntry child : entry.children) {
+        if (child.isDirectory)
+            subdirectories.emplace_back(
```
lioncash: this and the else should be braced, given they travel more than one line.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+    std::string part_path, o_name, extention;
+    Common::SplitPath(path, &part_path, &o_name, &extention);
+    backing = FileUtil::IOFile(part_path + name, mode.c_str());
+    return out;
+}
+
+RealVfsDirectory::RealVfsDirectory(const std::string& path, const char openmode[])
+    : path(path[path.size() - 1] == DIR_SEP_CHR ? path.substr(0, path.size() - 1) : path) {
+    FileUtil::FSTEntry entry;
+    if (0 != FileUtil::ScanDirectoryTree(path, entry))
```
lioncash:
```cpp
FileUtil::ScanDirectoryTree(path, entry) != 0
```

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+bool RealVfsFile::Rename(const std::string& name) {
+    auto out = FileUtil::Rename(GetName(), name);
+    std::string part_path, o_name, extention;
+    Common::SplitPath(path, &part_path, &o_name, &extention);
+    backing = FileUtil::IOFile(part_path + name, mode.c_str());
+    return out;
+}
+
+RealVfsDirectory::RealVfsDirectory(const std::string& path, const char openmode[])
```
lioncash: Ditto about variable shadowing warnings.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+    : path(path[path.size() - 1] == DIR_SEP_CHR ? path.substr(0, path.size() - 1) : path) {
```
lioncash:
```cpp
path(path.back() == DIR_SEP_CHR ? path.substr(0, path.size() - 1) : path)
```

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+    FileUtil::FSTEntry entry;
+    if (0 != FileUtil::ScanDirectoryTree(path, entry))
+        NGLOG_CRITICAL(Service_FS, "Failure to initialize file.");
+    for (FileUtil::FSTEntry child : entry.children) {
```
lioncash: You likely want to do `const FileUtil::FSTEntry& child` here, otherwise this does wasteful copying.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+    FileUtil::CreateDir(path + DIR_SEP + name);
+    subdirectories.emplace_back(
+        std::make_shared<RealVfsDirectory>(path + DIR_SEP + name, mode.c_str()));
+    return subdirectories[subdirectories.size() - 1];
+}
+
+std::shared_ptr<VfsFile> RealVfsDirectory::CreateFile(const std::string& name) {
+    FileUtil::CreateEmptyFile(path + DIR_SEP + name);
+    files.emplace_back(std::make_shared<RealVfsFile>(path + DIR_SEP + name, mode.c_str()));
+    return files[files.size() - 1];
```
lioncash:
```cpp
return files.back();
```

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+        return nullptr;
+
+    return std::make_shared<RealVfsDirectory>(path.substr(0, last), mode.c_str());
+}
+
+std::shared_ptr<VfsDirectory> RealVfsDirectory::CreateSubdirectory(const std::string& name) {
+    FileUtil::CreateDir(path + DIR_SEP + name);
+    subdirectories.emplace_back(
+        std::make_shared<RealVfsDirectory>(path + DIR_SEP + name, mode.c_str()));
+    return subdirectories[subdirectories.size() - 1];
```
lioncash:
```cpp
return subdirectories.back();
```

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+bool RealVfsFile::IsWritable() {
+    return mode.find('w') != std::string::npos;
+}
+
+bool RealVfsFile::IsReadable() {
+    return mode.find('r') != std::string::npos;
+}
+
+std::vector<u8> RealVfsFile::ReadBytes(u64 offset, u64 length) {
+    backing.Seek(offset, SEEK_SET);
```
lioncash: This doesn't account for the seek failing.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+    std::vector<u8> out(length);
+    backing.ReadBytes(out.data(), length);
```
lioncash: This doesn't account for the read not being fully successful.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+    return out;
+}
+
+u64 RealVfsFile::WriteBytes(const std::vector<u8>& data, u64 offset) {
+    backing.Seek(offset, SEEK_SET);
```
lioncash: This doesn't account for the seek failing.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+std::shared_ptr<VfsDirectory> RealVfsDirectory::CreateSubdirectory(const std::string& name) {
+    FileUtil::CreateDir(path + DIR_SEP + name);
+    subdirectories.emplace_back(
+        std::make_shared<RealVfsDirectory>(path + DIR_SEP + name, mode.c_str()));
+    return subdirectories[subdirectories.size() - 1];
+}
+
+std::shared_ptr<VfsFile> RealVfsDirectory::CreateFile(const std::string& name) {
+    FileUtil::CreateEmptyFile(path + DIR_SEP + name);
```
lioncash: This doesn't account for the case where this function doesn't succeed.

src/core/file_sys/vfs.h
```diff
@@ -0,0 +1,105 @@
...
+        memcpy(out_vec.data(), vec.data(), vec.size());
+        return out_vec;
+    }
+
+    template <typename T>
+    u64 ReadArray(T* data, u64 offset, u64 number_elements) {
+        static_assert(std::is_trivially_copyable<T>(),
+                      "Given array does not consist of trivially copyable objects");
+
+        std::vector<T> vec = ReadArray<T>(offset, number_elements);
```
lioncash: It seems a bit wasteful here to allocate a vector instead of having an overload that just copies data out into a `data` pointer.

src/core/file_sys/vfs.cpp
```diff
@@ -0,0 +1,73 @@
...
+    auto files = GetFiles();
+    auto iter = std::find_if(files.begin(), files.end(),
+                             [&name](auto file1) { return name == file1->GetName(); });
+    return iter == files.end() ? nullptr : std::move(*iter);
+}
+
+std::shared_ptr<VfsDirectory> VfsDirectory::GetSubdirectory(const std::string& name) {
+    auto subs = GetSubdirectories();
+    auto iter = std::find_if(subs.begin(), subs.end(),
+                             [&name](auto file1) { return name == file1->GetName(); });
```
lioncash: This should likely be `const auto&` so a copy isn't made.

src/core/file_sys/vfs.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+VfsDirectory::operator bool() {
+    return IsGood();
+}
+
+std::shared_ptr<VfsFile> VfsDirectory::GetFile(const std::string& name) {
+    auto files = GetFiles();
+    auto iter = std::find_if(files.begin(), files.end(),
+                             [&name](auto file1) { return name == file1->GetName(); });
```
lioncash: This should likely be `const auto&` so a copy isn't made.

src/core/file_sys/vfs_offset.cpp
```diff
@@ -0,0 +1,69 @@
...
+std::vector<u8> OffsetVfsFile::ReadBytes(u64 r_offset, u64 r_length) {
+    return file->ReadBytes(offset + r_offset, std::min(r_offset + r_length, size));
+}
+
+u64 OffsetVfsFile::WriteBytes(const std::vector<u8>& data, u64 r_offset) {
+    auto end = data.end();
+    if (data.size() + r_offset > size)
+        end = data.begin() + size - r_offset;
+
+    return file->WriteBytes(std::vector<u8>(data.begin(), end), r_offset + offset);
```
lioncash: Does it really make sense to allocate a std::vector for this, when you already have the beginning and end? This seems it would be better suited to an overload that accepts iterators instead, avoiding the need to allocate.

src/core/file_sys/vfs_real.cpp
```diff
@@ -0,0 +1,161 @@
...
+    return out;
+}
+
+u64 RealVfsFile::WriteBytes(const std::vector<u8>& data, u64 offset) {
+    backing.Seek(offset, SEEK_SET);
+    return backing.WriteBytes(data.data(), data.size());
+}
+
+bool RealVfsFile::Rename(const std::string& name) {
+    auto out = FileUtil::Rename(GetName(), name);
```
lioncash: If the rename fails, should it not bail out immediately instead of continuing onward?

PR #580, opened 2018-06-22
Fixed RequestUpdateAudioRenderer deadlocks and calculated section sizes properly

ogniK5377:
This fixes RequestUpdateAudioRenderer deadlocks in games like Puyo Puyo Tetris and games which require a proper section size in games such as Retro City Rampage. This fix causes various games to start rendering or trying to render. This is an iteration on #577

### DIFF
--- a/src/core/hle/service/audio/audren_u.cpp
+++ b/src/core/hle/service/audio/audren_u.cpp
@@ -120,7 +120,7 @@ private:
         NGLOG_WARNING(Service_Audio, "(STUBBED) called");
     }
 
-    enum class MemoryPoolStates : u32_le {
+    enum class MemoryPoolStates : u32 { // Should be LE
         Invalid = 0x0,
         Unknown = 0x1,
         RequestDetatch = 0x2,
--- a/src/core/hle/service/audio/audren_u.cpp
+++ b/src/core/hle/service/audio/audren_u.cpp
@@ -315,7 +315,7 @@ void AudRenU::GetAudioRendererWorkBufferSize(Kernel::HLERequestContext& ctx) {
                             (params.unknown8 + 6),
                         0x40);
 
-    if (IsFeatureSupported(AudioFeatures::Splitter, params.magic)) {
+    if (IsFeatureSupported(AudioFeatures::Splitter, params.revision)) {
         u32 count = params.unknownC + 1;
         u64 node_count = Common::AlignUp(count, 0x40);
         u64 node_state_buffer_sz =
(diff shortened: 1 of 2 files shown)


src/core/hle/service/audio/audren_u.cpp
```diff
@@ -116,6 +128,24 @@ class IAudioRenderer final : public ServiceFramework<IAudioRenderer> {
...
 
+    struct MemoryPoolInfo {
+        u64_le pool_address;
+        u64_le pool_size;
+        u32_le pool_state;
+        INSERT_PADDING_WORDS(3); // Unknown
+    };
+    static_assert(sizeof(MemoryPoolInfo) == 0x20, "MemoryPoolInfo has wrong size");
+
+    enum MemoryPoolStates : u32 {
```
mailwl: suggest convert to `enum class`  and remove `MPS_` part

src/core/hle/service/audio/audren_u.cpp
```diff
@@ -108,14 +120,32 @@ class IAudioRenderer final : public ServiceFramework<IAudioRenderer> {
         NGLOG_WARNING(Service_Audio, "(STUBBED) called");
     }
 
+    enum class MemoryPoolStates : u32_le {
```
bunnei: Base type of `u32_le` is meaningless, since this would not work anyways on a BE system (underlying enum types must be the base types, e.g. int/char/etc. this just works because u32_le on your system is `unsigned int`)

src/core/hle/service/audio/audren_u.h
```diff
@@ -12,6 +12,24 @@ class HLERequestContext;
 
 namespace Service::Audio {
 
+struct AudioRendererParameters {
+    u32_le sample_rate;
+    u32_le sample_count;
+    u32_le unknown8;
+    u32_le unknownC;
```
bunnei: nit: Don't mix case

src/core/hle/service/audio/audren_u.cpp
```diff
@@ -108,14 +120,32 @@ class IAudioRenderer final : public ServiceFramework<IAudioRenderer> {
         NGLOG_WARNING(Service_Audio, "(STUBBED) called");
     }
 
+    enum class MemoryPoolStates : u32 { // Should be LE
+        Invalid = 0x0,
+        Unknown = 0x1,
+        RequestDetatch = 0x2,
```
mailwl: Detach without 't'? both for 0x2 and 0x3

PR #585, opened 2018-06-24
Improve directory creation in WindowsCopyFiles.cmake

janisozaur:


bunnei:
Change lgtm... @janisozaur were you planning on making the above suggestion with this PR, or should this be merged as-is?

PR #587, opened 2018-06-25
Send the correct RequestUpdateAudioRenderer revision in the output header

ogniK5377:
Hardware test on a 5.1.0 console shows the latest revision is sent instead of the one provided by the AudioRendererParameter. As of 5.1.0, REV4 is the latest revision.

### DIFF
--- a/src/core/hle/service/audio/audren_u.cpp
+++ b/src/core/hle/service/audio/audren_u.cpp
@@ -150,7 +150,7 @@ private:
         UpdateDataHeader() {}
 
         UpdateDataHeader(const AudioRendererParameter& config) {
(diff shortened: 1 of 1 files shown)


src/core/hle/service/audio/audren_u.cpp
```diff
@@ -150,7 +150,7 @@ class IAudioRenderer final : public ServiceFramework<IAudioRenderer> {
         UpdateDataHeader() {}
 
         UpdateDataHeader(const AudioRendererParameter& config) {
-            revision = config.revision;
+            revision = Common::MakeMagic('R', 'E', 'V', '4');
```
Subv: Please add a comment saying that this is the current revision as of 5.1.0

PR #588, opened 2018-06-25
Service/Audio: add hwopus service, stub GetWorkBufferSize function

mailwl:
Used in game BlazBlue and some other

### DIFF
--- a/src/core/hle/service/audio/hwopus.cpp
+++ b/src/core/hle/service/audio/hwopus.cpp
@@ -1,0 +1,29 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/logging/log.h"
+#include "core/hle/ipc_helpers.h"
+#include "core/hle/kernel/hle_ipc.h"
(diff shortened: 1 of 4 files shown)


src/core/hle/service/audio/hwopus.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "core/hle/kernel/hle_ipc.h"
+#include "core/hle/service/audio/hwopus.h"
+
+namespace Service::Audio {
+
+void HwOpus::GetWorkBufferSize(Kernel::HLERequestContext& ctx) {
+    NGLOG_WARNING(Service_Audio, "(STUBBED) called");
+    IPC::ResponseBuilder rb{ctx, 3};
+    rb.Push(RESULT_SUCCESS);
+    rb.Push<u32>(0x4000);
```
Hexagon12: According to SwitchBrew:

> Takes two u32s SampleRate and ChannelCount packed as an u64.

Where do you take the 0x4000 from, exactly?

mailwl: just random number, to pass crash in game.
i don't know algorithm for calculating size, so, input parameters unused

PR #594, opened 2018-06-27
gl_rasterizer: Workaround for when exceeding max UBO size.

bunnei:


bunnei:
@Subv this is a fix for #573 breaking SMO, which has a UBO size of 0x40000. Purely FYI, if you think this needs a better fix, let me know.

PR #595, opened 2018-06-27
Rewrite the OpenGL rasterizer cache

bunnei:
This is more or less a complete rewrite of the OpenGL rasterizer cache. This used to cache "surfaces", which represent an emulated texture or framebuffer. These can be more or less used interchangeably - for example, a framebuffer can be used as a texture (for a "render-to-texture" effect). Much like our old cache, 
a "surface" is a somewhat generic construct, that is both texture and framebuffer agnostic.

**Why rewrite it?** Our previous rasterizer cache was based on [Citra's implementation](https://github.com/citra-emu/citra/blob/master/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp). This served us for a while as an easy drop in, but it quickly became unmanageable. 3DS requires a more complicated cache (won't go into details here), but needless to say, the complexity is an unused burden with yuzu. Furthermore, it was designed with 3DS constraints in mind that are not applicable with Switch (specifically, Citra has no HW compressed texture support, and a fixed tile size). Loosening these constraints within the complexities of the Citra cache quickly became unmanageable.

**What's different?** This implementation simplifies things. This potentially could be at the cost of accuracy, but it's probably safe to assume that we do not need to account for the edge cases that were handled by the previous implementation. In summary:
* We retain constructs of `Surface`, `SurfaceParams`, and the primary interface for the rasterizer cache
* We have a single flat map of surfaces for the surface cache, with is keyed based on `SurfaceParams`
* Complex scenarios such as subsets of surfaces, reinterpreting surface formats, etc. are no longer supported (although there are some asserts when this type of thing happens)
* Textures are cached (as they were before), and refreshed on memory writes
* Render buffers (once flagged as such) live only in cache space, they are never flushed to emulated memory nor are they invalidated by it, however they can be directly used as textures
* Various fixes for compressed textures and ASTC textures
* To compensate for a potential loss of accuracy, we've introduced a new setting, `use_accurate_framebuffers`. When enabled, this will always load textures from memory and flush render buffers. This is useful for debugging, but that's about it - swizzling a 1080p framebuffer on every draw is painfully slow, and so is software decoding ASTC

**What's improved?** I tested all of my games, and there are no regressions (note: this was after a lot of work, lol). There are graphical fixes in almost every game, due to our buggy old implementation. Performance should be on par with the old implementation, and/or a little better in some places (due to render buffers being permanently cached). 

An obligatory screenshot - since SMO is looking much better with these changes (and running twice as fast!)

image

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -29,6 +29,28 @@ struct FormatTuple {
     bool compressed;
 };
 
+SurfaceParams::SurfaceParams(const Tegra::Texture::FullTextureInfo& config)
+    : addr(config.tic.Address()), is_tiled(config.tic.IsTiled()),
+      block_height(is_tiled ? config.tic.BlockHeight() : 0),
+      pixel_format(PixelFormatFromTextureFormat(config.tic.format)),
+      component_type(ComponentTypeFromTexture(config.tic.r_type.Value())),
+      type(GetFormatType(pixel_format)),
+      width(Common::AlignUp(config.tic.Width(), GetCompressionFactor(pixel_format))),
+      height(Common::AlignUp(config.tic.Height(), GetCompressionFactor(pixel_format))) {
+
+    // TODO(Subv): Different types per component are not supported.
+    ASSERT(config.tic.r_type.Value() == config.tic.g_type.Value() &&
+           config.tic.r_type.Value() == config.tic.b_type.Value() &&
+           config.tic.r_type.Value() == config.tic.a_type.Value());
+}
+
+SurfaceParams::SurfaceParams(const Tegra::Engines::Maxwell3D::Regs::RenderTargetConfig& config)
+    : addr(config.Address()), is_tiled(true),
+      block_height(Tegra::Texture::TICEntry::DefaultBlockHeight),
+      pixel_format(PixelFormatFromRenderTargetFormat(config.format)),
+      component_type(ComponentTypeFromRenderTarget(config.format)),
+      type(GetFormatType(pixel_format)), width(config.width), height(config.height) {}
+
 static constexpr std::array<FormatTuple, SurfaceParams::MaxPixelFormat> tex_format_tuples = {{
     {GL_RGBA8, GL_RGBA, GL_UNSIGNED_INT_8_8_8_8_REV, false},                    // ABGR8
     {GL_RGB, GL_RGB, GL_UNSIGNED_SHORT_5_6_5_REV, false},                       // B5G6R5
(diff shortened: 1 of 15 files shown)


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -29,29 +29,37 @@ struct FormatTuple {
...
-      type(GetFormatType(pixel_format)),
-      width(Common::AlignUp(config.tic.Width(), GetCompressionFactor(pixel_format))),
-      height(Common::AlignUp(config.tic.Height(), GetCompressionFactor(pixel_format))),
-      size_in_bytes(SizeInBytes()) {
-
-    // TODO(Subv): Different types per component are not supported.
-    ASSERT(config.tic.r_type.Value() == config.tic.g_type.Value() &&
-           config.tic.r_type.Value() == config.tic.b_type.Value() &&
-           config.tic.r_type.Value() == config.tic.a_type.Value());
+/*static*/ SurfaceParams SurfaceParams::CreateForTexture(
```
Subv: what's with that commented out static?

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -29,29 +29,37 @@ struct FormatTuple {
...
 }
 
-SurfaceParams::SurfaceParams(const Tegra::Engines::Maxwell3D::Regs::RenderTargetConfig& config)
-    : addr(config.Address()), is_tiled(true),
-      block_height(Tegra::Texture::TICEntry::DefaultBlockHeight),
-      pixel_format(PixelFormatFromRenderTargetFormat(config.format)),
-      component_type(ComponentTypeFromRenderTarget(config.format)),
-      type(GetFormatType(pixel_format)), width(config.width), height(config.height),
-      size_in_bytes(SizeInBytes()) {}
+/*static*/ SurfaceParams SurfaceParams::CreateForFramebuffer(
```
Subv: same

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -317,220 +233,79 @@ struct SurfaceParams {
...
-
-    u64 PixelsInBytes(u64 size) const {
-        return size * CHAR_BIT / GetFormatBpp(pixel_format);
-    }
-
-    u64 BytesInPixels(u64 pixels) const {
-        return pixels * GetFormatBpp(pixel_format) / CHAR_BIT;
+    size_t SizeInBytes() const {
+        const u32 compression_factor{GetCompressionFactor()};
+        ASSERT(width % compression_factor == 0);
```
Subv: I think these asserts are going to fail for Splatoon 2, iirc the squid texture size is not a multiple of its compression factor

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
bunnei: Just a style thing, indicates this is a static class function from the CPP, as opposed to a class method

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
bunnei: Just a style thing, indicates this is a static class function from the CPP, as opposed to a class method

src/video_core/renderer_opengl/gl_rasterizer_cache.h
bunnei: Splatoon 2 works fine, we always alignup texture width/height to compression factor

bunnei:
@Subv your comments were addressed in replies, if you feel like there should be changes, feel free to ping me and I will make a followup PR.

PR #599, opened 2018-06-30
gl_rasterizer: Use GPU virtual addr when possible for framebuffers.

bunnei:


src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -85,8 +85,8 @@ u32 nvhost_as_gpu::Remap(const std::vector<u8>& input, std::vector<u8>& output)
         u64 size = static_cast<u64>(entry.pages) << 0x10;
         ASSERT(size <= object->size);
 
-        Tegra::GPUVAddr returned = gpu.memory_manager->MapBufferEx(object->addr, offset, size);
-        ASSERT(returned == offset);
+        object->gpu_addr = gpu.memory_manager->MapBufferEx(object->cpu_addr, offset, size);
```
Subv: You set the GPU address here and in MapBufferEx, but you don't unset it when Unmap is called

Subv: Also, the logic is wrong here, MapBufferEx and Remap can be called multiple times with the same nvmap object at different offsets and it'll return different addresses, right now you're just overwriting the gpu address to whatever was used last

Hexagon12:
Please fix conflicts kthx

Subv:
I think this PR should remove the first and last commits, leaving only the ` gl_rasterizer_cache: Track page counts in CPU virtual addresses. ` commit.

Hexagon12:
Is this PR still relevant?

PR #607, opened 2018-07-02
Logging - Customizable backends

jroweboy:
See https://github.com/citra-emu/citra/pull/3568 for more info.

* Hides the yuzu log window by default on release mode.
* Adds a setting to show the log window
* Adds a ui field to change log level
* Writing logs to console/disk is now on a background thread. Calling a log macro simply acquires a mutex and adds the entry to a queue for the backend to write.
* Logs are written to disk, but only the first 50Mb (to keep from eating up hard drive space if it spams too much)
* New button in Debug to show log location 

Obligatory screenshot: 
yuzu_2018-07-02_11-14-18


src/common/logging/filter.cpp
```diff
@@ -65,14 +65,14 @@ bool Filter::ParseFilterRule(const std::string::const_iterator begin,
                              const std::string::const_iterator end) {
     auto level_separator = std::find(begin, end, ':');
     if (level_separator == end) {
-        NGLOG_ERROR(Log, "Invalid log filter. Must specify a log level after `:`: %s",
-                    std::string(begin, end).c_str());
+        LOG_ERROR(Log, "Invalid log filter. Must specify a log level after `:`: %s",
```
Subv: Shouldn't this be changed to use {} instead of %s and do away with the c_str()?

src/common/logging/filter.cpp
```diff
@@ -65,14 +65,14 @@ bool Filter::ParseFilterRule(const std::string::const_iterator begin,
...
+                  std::string(begin, end).c_str());
         return false;
     }
 
     const Level level = GetLevelByName(level_separator + 1, end);
     if (level == Level::Count) {
-        NGLOG_ERROR(Log, "Unknown log level in filter: %s", std::string(begin, end).c_str());
+        LOG_ERROR(Log, "Unknown log level in filter: %s", std::string(begin, end).c_str());
```
Subv: same

src/common/logging/filter.cpp
```diff
@@ -83,7 +83,7 @@ bool Filter::ParseFilterRule(const std::string::const_iterator begin,
 
     const Class log_class = GetClassByName(begin, level_separator);
     if (log_class == Class::Count) {
-        NGLOG_ERROR(Log, "Unknown log class in filter: %s", std::string(begin, end).c_str());
+        LOG_ERROR(Log, "Unknown log class in filter: %s", std::string(begin, end).c_str());
```
Subv: same

PR #608, opened 2018-07-02
GPU: Implemented the depth buffer and depth test + culling

Subv:
Games that use the depth test will most likely draw a black screen until depth clearing is implemented.


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -149,11 +145,17 @@ void MortonCopy(u32 stride, u32 block_height, u32 height, u8* gl_buffer, Tegra::
...
-            SurfaceParams::TextureFormatFromPixelFormat(format), stride, height, block_height);
-
-        std::memcpy(gl_buffer, data.data(), data.size());
+        if (SurfaceParams::GetFormatType(format) == SurfaceType::ColorTexture) {
+            auto data = Tegra::Texture::UnswizzleTexture(
+                *gpu.memory_manager->GpuToCpuAddress(addr),
+                SurfaceParams::TextureFormatFromPixelFormat(format), stride, height, block_height);
+            std::memcpy(gl_buffer, data.data(), data.size());
+        } else {
+            auto data = Tegra::Texture::UnswizzleDepthTexture(
```
bunnei: Since this is an enum with more than 1 value, can you change this to a switch statement and assert on default? If we don't need the other enum values, we can just make this a boolean, then an if-statement is OK

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -307,7 +307,9 @@ void RasterizerOpenGL::DrawArrays() {
     // TODO(bunnei): Implement these
     const bool has_stencil = false;
     const bool using_color_fb = true;
-    const bool using_depth_fb = false;
+
```
bunnei: nit: What's with this line? Move the TODO to just the has_stencil line if that's the only place it now applies?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -349,6 +349,8 @@ void RasterizerOpenGL::DrawArrays() {
     // Sync the blend state registers
     SyncBlendState();
 
+    SyncCullMode();
```
bunnei: nit: Either add a comment to this line, or remove the comments from the preceding two calls and group these together (that may be the better option, since this code is quite clear)... by OCD does not like the commenting inconsistency

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
Subv: There may be other values (if we decide to separate Depth from DepthStencil formats) but for now there aren't. IMO the enum is much more descriptive than a simple boolean

PR #609, opened 2018-07-02
GPU: Implemented the CLEAR_BUFFERS register.

Subv:
Support clearing both the color and the depth buffer.
Depends on #608

### DIFF
--- a/src/video_core/engines/maxwell_3d.h
+++ b/src/video_core/engines/maxwell_3d.h
@@ -436,7 +436,12 @@ public:
                     u32 count;
                 } vertex_buffer;
 
-                INSERT_PADDING_WORDS(0x99);
+                INSERT_PADDING_WORDS(1);
+
+                float clear_color[4];
+                float clear_depth;
+
+                INSERT_PADDING_WORDS(0x93);
 
                 struct {
                     u32 address_high;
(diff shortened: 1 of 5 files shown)


src/video_core/engines/maxwell_3d.h
```diff
@@ -781,13 +803,15 @@ class Maxwell3D final {
 
 #define ASSERT_REG_POSITION(field_name, position)                                                  \
     static_assert(offsetof(Maxwell3D::Regs, field_name) == position * 4,                           \
-                  "Field " #field_name " has invalid position")
+                  "Field " #field_name " has invalid position");
```
bunnei: What's with this change?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -297,6 +297,77 @@ bool RasterizerOpenGL::AccelerateDrawBatch(bool is_indexed) {
     return true;
 }
 
+void RasterizerOpenGL::Clear() {
+    const auto& regs = Core::System().GetInstance().GPU().Maxwell3D().regs;
+
+    // TODO(bunnei): Implement these
+    const bool has_stencil = false;
```
bunnei: Whats with the big copy paste with this function? (note: the same comment applies with this TODO as with your last PR). Can we refactor part of this into some `SetupRenderBuffers` helper function that is shared code?

PR #613, opened 2018-07-03
Add qt windowsvistastyle dll to the build

jroweboy:
See https://github.com/citra-emu/citra/pull/3716 for what happened to the dll


bunnei:
tested after my force push, lgtm

PR #615, opened 2018-07-03
Ignore asserts - DO NOT CHECK IN

bunnei:
**DO NOT MERGE**

Using this PR to disable assertions in canary builds to make it easier for non-developers to test games.
This is more-or-less just a trial, if it goes well, we should probably come up with a more robust solution.

This replaces #596, after the merge conflicts of #607.


jroweboy:
Closing this in favor of tagging https://github.com/yuzu-emu/yuzu/pull/1536 which also ignores asserts but prints skipped asserts in the log.

PR #625, opened 2018-07-04
GPU: Implemented the IMNMX shader instruction.

Subv:
It's similar to the FMNMX instruction but it works on integers.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -194,6 +194,13 @@ enum class UniformType : u64 {
     Double = 5,
 };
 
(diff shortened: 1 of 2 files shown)


src/video_core/engines/shader_bytecode.h
```diff
@@ -194,6 +194,13 @@ enum class UniformType : u64 {
     Double = 5,
 };
 
+enum class IMinMaxExchange : u64 {
```
bunnei: nit: `IntegerMinMaxChange`? e.g. if you're gonna lose the abbreviation on Min and Max, lose it on Integer too. Otherwise, this now sounds like an interface

PR #637, opened 2018-07-07
Port #3335 and #3373 from Citra: "Small SDL fixes" and "Print the actual error preventing SDL from working"

FearlessTobi:
I know we usually shall not port different PRs in one PR, but those two both touch the SDL frontend and are pretty small, so I hope it's okay in this case.

See https://github.com/citra-emu/citra/pull/3335 and https://github.com/citra-emu/citra/pull/3373 for more details.

### DIFF
--- a/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
+++ b/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
@@ -126,7 +126,7 @@ EmuWindow_SDL2::EmuWindow_SDL2(bool fullscreen) {
                          SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI);
 
     if (render_window == nullptr) {
-        LOG_CRITICAL(Frontend, "Failed to create SDL2 window! Exiting...");
+        LOG_CRITICAL(Frontend, "Failed to create SDL2 window! %s", SDL_GetError());
         exit(1);
     }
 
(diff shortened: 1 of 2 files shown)


baka0815:
Just wanted to note that citra-emu/citra#3335 is only taken partly (fixing include of ``getopt.h`` in ``yuzu.cpp``), but not fixing the ``break;?``

FearlessTobi:
That's because the feature with the missing `break;` isn't in yuzu, so I don'thave to make that change here. (Or am I wrong?)

src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
```diff
@@ -126,7 +126,7 @@ EmuWindow_SDL2::EmuWindow_SDL2(bool fullscreen) {
                          SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI);
 
     if (render_window == nullptr) {
-        LOG_CRITICAL(Frontend, "Failed to create SDL2 window! Exiting...");
+        LOG_CRITICAL(Frontend, "Failed to create SDL2 window! %s", SDL_GetError());
```
wwylele: Aren't we using new fmtlib style formatting string (`"{}"`)?

FearlessTobi: Oh, good catch. I'll fix this tomorrow.

baka0815:
@FearlessTobi yes, the ``break`` isn't necessary, because all arguments are currently handled correctly.
I just wanted to point that out. Sorry for the confusion!

bunnei:
@FearlessTobi ping to update this please!

FearlessTobi:
@bunnei Fixed.

PR #638, opened 2018-07-07
Implement exclusive monitor

MerryMage:
Global exclusive monitor across cores.

NOTE: Unicorn support is nonexistent.

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic.h
+++ b/src/core/arm/dynarmic/arm_dynarmic.h
@@ -6,15 +6,18 @@
 
 #include <memory>
 #include <dynarmic/A64/a64.h>
+#include <dynarmic/A64/exclusive_monitor.h>
 #include "common/common_types.h"
 #include "core/arm/arm_interface.h"
+#include "core/arm/exclusive_monitor.h"
 #include "core/arm/unicorn/arm_unicorn.h"
 
 class ARM_Dynarmic_Callbacks;
+class DynarmicExclusiveMonitor;
 
 class ARM_Dynarmic final : public ARM_Interface {
 public:
-    ARM_Dynarmic();
+    ARM_Dynarmic(std::shared_ptr<ExclusiveMonitor> exclusive_monitor, size_t core_index);
     ~ARM_Dynarmic();
 
     void MapBackingMemory(VAddr address, size_t size, u8* memory,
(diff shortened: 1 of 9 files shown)


src/core/core.h
```diff
@@ -130,6 +136,11 @@ class System {
         return *CurrentCpuCore().Scheduler();
     }
 
+    /// Gets the scheduler for the CPU core that is currently running
```
Subv: the comment needs to be updated

Hexagon12:
This PR breaks all Unity based games (Yuzu throws a unhandled exception at 0xFFFFFFFFFFFFFFF and the call stack in VS is empty)

bunnei:
@Hexagon12 can you provide more details? For example:
* Can you provide some examples of games that are broken?
* Are these games broken even with multicore off?

src/core/arm/dynarmic/arm_dynarmic.h
```diff
@@ -39,15 +42,41 @@ class ARM_Dynarmic final : public ARM_Interface {
...
 
+    size_t core_index;
+    std::shared_ptr<DynarmicExclusiveMonitor> exclusive_monitor;
+
     Memory::PageTable* current_page_table = nullptr;
 };
+
+class DynarmicExclusiveMonitor final : public ExclusiveMonitor {
+public:
+    DynarmicExclusiveMonitor(size_t core_count);
```
lioncash:
```cpp
explicit DynarmicExclusiveMonitor(size_t core_count);
```

src/core/arm/exclusive_monitor.h
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+
+class ExclusiveMonitor {
+public:
+    virtual ~ExclusiveMonitor() = default;
```
lioncash: Note that if this isn't put in a cpp file, it will cause a -Wweak-vtables warning.

src/core/arm/exclusive_monitor.h
```diff
@@ -0,0 +1,22 @@
...
+
+    virtual void SetExclusive(size_t core_index, u64 addr) = 0;
+    virtual void ClearExclusive() = 0;
+
+    virtual bool ExclusiveWrite8(size_t core_index, u64 vaddr, u8 value) = 0;
+    virtual bool ExclusiveWrite16(size_t core_index, u64 vaddr, u16 value) = 0;
+    virtual bool ExclusiveWrite32(size_t core_index, u64 vaddr, u32 value) = 0;
+    virtual bool ExclusiveWrite64(size_t core_index, u64 vaddr, u64 value) = 0;
+    virtual bool ExclusiveWrite128(size_t core_index, u64 vaddr,
+                                   std::array<std::uint64_t, 2> value) = 0;
```
lioncash: `<array>` should be included.

bunnei:
@MerryMage can you test any of the games that @Hexagon12 mentioned above? Please let me know if you need help debugging - I at least have Snipper Clips and can give it a try.

MerryMage:
Will investigate

Subv:
Note that Unity games depend on dynarmic writing to TPIDR_EL0 properly, which it does not since the IR opcode that writes it is being deleted by the dead code elimination optimization.

A fix would be to add the A64SetTPIDR opcode to Inst::WritesToCPSR.

Subv:
From Discord
```
oh, i can reproduce the crash @Hexagon12:IModerate talked about in Snipperclips with that PR
though it's a weird one
VS says Exception thrown at 0x000001D40C5CC1CD in yuzu.exe: 0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF. occurred
the instruction at that address is 000001D40C5CC1CD  movaps      xmmword ptr [r8],xmm0
but r8 is 0x000000c6e00ffad8
so i don't know what's going on?
https://gist.github.com/Subv/27213506dfd5c2ef92825adb103f052f here's the generated block that's causing problems
```

Subv:
The problem is that this movaps https://github.com/MerryMage/dynarmic/blob/master/src/backend_x64/a64_emit_x64.cpp#L901 might access unaligned memory. We can either ensure code.ABI_PARAM3 is always aligned to 16 bytes, or use movups.

Subv:
Note: https://github.com/Subv/yuzu/commit/023a826b87b03644a1a1a48f5bf714eb97d37f89 is required after this is merged

PR #597, opened 2018-06-27
Virtual Filesystem

DarkLordZach:
NOTE: Still testing and fixing bugs.. wanted to get review input on code itself. I'm aware of a crash related to save data right now... should be done tonight.

Main Features:
- Fully abstract away the impl. of the files and directories.
- This only adds the classes and ports over loaders/services to use it.
- Eventual goal: Encryption/updates/xci/etc.

PR #639, opened 2018-07-08
Revert "Virtual Filesystem (#597)"

bunnei:
This reverts commit 77c684c1140f6bf3fb7d4560d06d2efb1a2ee5e2.

@DarkLordZach this seems to break save game creation. E.g. delete your nand folder, and try to run Puyo Puyo Tetris. I'm not familiar enough with the code to easily debug it, but please resubmit this change once you have identified the problem - as this is definitely something we'd like to have. I'm reverting since you are AFK for several days, otherwise I would be happy to debug with you 👍


Subv:
The issue with save files happens because for some reason the CreateDir function doesn't work recursively now, ie, trying to create `nand/save/ID/00000000` will fail if `nand/save/ID` does not exist.

I do have a problem with the way the VFS handles archives, most archives (and more specifically, SaveData) also accept a structure which specifies *which* save to open. The way VFS was set up made it so only the current program's save file could ever be opened.

PR #645, opened 2018-07-10
More improvements to GDBStub

Hedges:
- Debugging of threads should work correctly with source and assembly level stepping and modifying registers and memory, meaning threads and callstacks are fully clickable in VS.
- List of modules is available to the client, with assumption that .nro and .nso are backed up by an .elf with symbols, while deconstructed ROMs keep N names.
- Initial support for floating point registers.


Hexagon12:
Please squash your commits into a few parts

Hedges:
It is actually not so many actual changes as there are just merges from master and I would rather keep it as a single commit which it conceptually is.

src/core/loader/nso.cpp
```diff
@@ -147,6 +149,11 @@ VAddr AppLoader_NSO::LoadModule(const std::string& name, const std::vector<u8>&
     codeset->memory = std::make_shared<std::vector<u8>>(std::move(program_image));
     Core::CurrentProcess()->LoadModule(codeset, load_base);
 
+    // Register module with GDBStub
+    std::string filename;
+    Common::SplitPath(codeset->name, nullptr, &filename, nullptr);
+    GDBStub::RegisterModule((filename + ".elf").c_str(), load_base, load_base);
```
lioncash: This seems to duplicate the code also used in nro.cpp. Perhaps it would be better to add another registration function to the GDBStub namespace.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -144,6 +181,7 @@ static u32 latest_signal = 0;
 static bool memory_break = false;
 
 static Kernel::Thread* current_thread = nullptr;
+static unsigned current_core = 0;
```
lioncash: `u32`

src/core/gdbstub/gdbstub.cpp
```diff
@@ -335,6 +393,21 @@ static void LongToGdbHex(u8* dest, u64 v) {
     }
 }
 
+/**
```
lioncash: This should be removed if it's not being used.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -171,13 +209,29 @@ static std::map<u64, Breakpoint> breakpoints_execute;
...
+    Module module;
+    strncpy(module.name, name, sizeof(module.name));
+    module.beg = beg;
+    module.end = end;
+    modules.push_back(module);
+}
+
 static Kernel::Thread* FindThreadById(int id) {
-    for (int core = 0; core < Core::NUM_CPU_CORES; core++) {
+    for (unsigned core = 0; core < Core::NUM_CPU_CORES; core++) {
```
lioncash: `u32`

src/core/gdbstub/gdbstub.cpp
```diff
@@ -367,6 +440,24 @@ static u64 GdbHexToLong(const u8* src) {
     return output;
 }
 
+/**
```
lioncash: Ditto

src/core/gdbstub/gdbstub.cpp
```diff
@@ -534,12 +625,17 @@ static void HandleQuery() {
...
+        if (modules.size()) {
+            buffer += ";qXfer:libraries:read+";
+        }
+        SendReply(buffer.c_str());
     } else if (strncmp(query, "Xfer:features:read:target.xml:",
                        strlen("Xfer:features:read:target.xml:")) == 0) {
         SendReply(target_xml);
     } else if (strncmp(query, "Offsets", strlen("Offsets")) == 0) {
-        std::string buffer = fmt::format("TextSeg={:0x}", Memory::PROCESS_IMAGE_VADDR);
+        std::string buffer;
```
lioncash: Why was this broken across two lines?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -739,6 +866,16 @@ static void ReadRegisters() {
 
     bufptr += 8;
 
+    for (int reg = UC_ARM64_REG_Q0; reg <= UC_ARM64_REG_Q0 + 31; reg++) {
+        LongToGdbHex(bufptr + reg * 16, RegRead(reg, current_thread));
+    }
+
+    bufptr += (32 * 32);
```
lioncash: the parentheses here are superfluous

src/core/gdbstub/gdbstub.cpp
```diff
@@ -806,6 +956,10 @@ static void ReadMemory() {
         SendReply("E01");
     }
 
+    if ((addr < Memory::PROCESS_IMAGE_VADDR) || (addr >= Memory::MAP_REGION_VADDR_END)) {
```
lioncash: Ditto

src/core/gdbstub/gdbstub.cpp
```diff
@@ -1192,8 +1349,12 @@ void SetCpuStepFlag(bool is_step) {
 
 void SendTrap(Kernel::Thread* thread, int trap) {
     if (send_trap) {
+        if (!halt_loop || (current_thread == thread)) {
```
lioncash: Ditto

src/core/gdbstub/gdbstub.cpp
```diff
@@ -171,13 +209,29 @@ static std::map<u64, Breakpoint> breakpoints_execute;
 static std::map<u64, Breakpoint> breakpoints_read;
 static std::map<u64, Breakpoint> breakpoints_write;
 
+struct Module {
+    char name[128];
```
lioncash: Any reason this doesn't just take a std::string? The register function can simply std::move the `name` parameter, if it's changed to a `std::string`, resulting in only one copy like the current code (or none at all, if the argument is std::moved into the function when calling it.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -171,13 +209,29 @@ static std::map<u64, Breakpoint> breakpoints_execute;
...
+};
+
+static std::vector<Module> modules;
+
+void RegisterModule(const char* name, PAddr beg, PAddr end) {
+    Module module;
+    strncpy(module.name, name, sizeof(module.name));
+    module.beg = beg;
+    module.end = end;
+    modules.push_back(module);
```
lioncash: With the above, module can also be std::moved

lioncash:
Just provided review on the changes itself. The rebasing stuff was already commented on.

src/core/gdbstub/gdbstub.cpp
Hedges: IMHO, I do not think the parentheses are superfluous inside ifs like the above and the next one. Sure, there is operator precedence etc, but it certainly is more readable when divided into logical blocks.

src/core/gdbstub/gdbstub.cpp
Hedges: Yeah, replacing it with std::string is a good idea, although std::moving things seems like unnecessary micro-optimization as this function is used so rarely.

src/core/gdbstub/gdbstub.cpp
lioncash: We essentially don't perform this form of sectioning anywhere else in the codebase in a widespread manner, as far as I remember (and there's also code added that does not parenthesize as well [example](https://github.com/yuzu-emu/yuzu/pull/645/files#diff-95c7701747859f90564fae1b6df32688R226)), so I do not agree.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -554,40 +588,59 @@ static void HandleQuery() {
         SendReply(val.c_str());
     } else if (strncmp(query, "sThreadInfo", strlen("sThreadInfo")) == 0) {
         SendReply("l");
+    } else if (strncmp(query, "Xfer:threads:read", strlen("Xfer:threads:read")) == 0) {
+        std::string buffer;
+        buffer += "l<?xml version=\"1.0\"?>";
+        buffer += "<threads>";
+        for (int core = 0; core < Core::NUM_CPU_CORES; core++) {
```
lioncash: I missed this before, but shouldn't this be a `u32` as well?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -554,40 +588,59 @@ static void HandleQuery() {
...
+            auto threads = Core::System::GetInstance().Scheduler(core)->GetThreadList();
```
lioncash: This should likely be `const auto&` to make it explicit the thread list is returned by reference (and only intended to be read).

src/core/gdbstub/gdbstub.cpp
```diff
@@ -554,40 +588,59 @@ static void HandleQuery() {
...
+            for (auto thread : threads) {
```
lioncash: Likewise this could likely be a `const auto&`

src/core/gdbstub/gdbstub.cpp
```diff
@@ -711,8 +771,13 @@ static void ReadRegister() {
         LongToGdbHex(reply, RegRead(id, current_thread));
     } else if (id == CPSR_REGISTER) {
         IntToGdbHex(reply, (u32)RegRead(id, current_thread));
+    } else if (id >= UC_ARM64_REG_Q0 && id < FPSCR_REGISTER) {
+        LongToGdbHex(reply, RegRead(id, current_thread));
+    } else if (id == FPSCR_REGISTER) {
+        LongToGdbHex(reply, RegRead(998, current_thread));
     } else {
-        return SendReply("E01");
+        // return SendReply("E01");
```
lioncash: Was this intended to be removed instead?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -554,40 +588,59 @@ static void HandleQuery() {
...
+                                thread->GetThreadId(), core, thread->GetThreadId());
+            }
+        }
+        buffer += "</threads>";
+        SendReply(buffer.c_str());
+    } else if (strncmp(query, "Xfer:libraries:read", strlen("Xfer:libraries:read")) == 0) {
+        std::string buffer;
+        buffer += "l<?xml version=\"1.0\"?>";
+        buffer += "<library-list>";
+        for (auto module : modules) {
```
lioncash: This should likely be `const auto&` otherwise this copies each module instance.

src/core/gdbstub/gdbstub.cpp
Hedges: Right, missed it.

src/core/gdbstub/gdbstub.cpp
Hedges: It is meant as a comment for the time being.

src/core/gdbstub/gdbstub.cpp
Hedges: You are right about all the const& stuff above although it should actually make no difference because of the way compiler works and also because those are not time-critical pieces of code. I will however tidy it up, cheers mate :)

src/core/gdbstub/gdbstub.cpp
```diff
@@ -171,13 +176,34 @@ static std::map<u64, Breakpoint> breakpoints_execute;
...
+    module.end = end;
+    modules.push_back(std::move(module));
+}
+
 static Kernel::Thread* FindThreadById(int id) {
-    for (int core = 0; core < Core::NUM_CPU_CORES; core++) {
-        auto threads = Core::System::GetInstance().Scheduler(core)->GetThreadList();
+    for (u32 core = 0; core < Core::NUM_CPU_CORES; core++) {
+        const auto& threads = Core::System::GetInstance().Scheduler(core)->GetThreadList();
         for (auto thread : threads) {
```
lioncash: Is it not applicable to use `auto&` here as well?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -543,8 +577,8 @@ static void HandleQuery() {
         SendReply(buffer.c_str());
     } else if (strncmp(query, "fThreadInfo", strlen("fThreadInfo")) == 0) {
         std::string val = "m";
-        for (int core = 0; core < Core::NUM_CPU_CORES; core++) {
-            auto threads = Core::System::GetInstance().Scheduler(core)->GetThreadList();
+        for (u32 core = 0; core < Core::NUM_CPU_CORES; core++) {
+            const auto& threads = Core::System::GetInstance().Scheduler(core)->GetThreadList();
             for (auto thread : threads) {
```
lioncash: Ditto but with `const auto&`

src/core/gdbstub/gdbstub.cpp
```diff
@@ -554,40 +588,59 @@ static void HandleQuery() {
         SendReply(val.c_str());
     } else if (strncmp(query, "sThreadInfo", strlen("sThreadInfo")) == 0) {
         SendReply("l");
+    } else if (strncmp(query, "Xfer:threads:read", strlen("Xfer:threads:read")) == 0) {
+        std::string buffer;
+        buffer += "l<?xml version=\"1.0\"?>";
+        buffer += "<threads>";
+        for (int core = 0; core < Core::NUM_CPU_CORES; core++) {
+            const auto& threads = Core::System::GetInstance().Scheduler(core)->GetThreadList();
+            for (auto thread : threads) {
```
lioncash: Ditto but with `const auto&`

src/core/gdbstub/gdbstub.cpp
lioncash: Unfortunately this is not the case, as each instance contains a std::string. Which does make a difference (each copy is a potential allocation point, which requires code generation. It also makes it a point in code that can potentially throw an exception, which is avoidable with a reference): [example](https://godbolt.org/g/UbhsNo).

You've said it multiple times, however just because it's not time-critical code doesn't mean trivial things that aren't out-of-the-way shouldn't be avoided or addressed.

Hedges: Argh, std::string... I keep forgetting I am not using mine...

bunnei:
@Hedges whats with the dynarmic update that is part of these changes?

Hedges:
@bunnei Something must have gone wrong as I did not intend to update it. Diff shows no change, so hopefully it can be ignored.

lioncash:
It can't. It would downgrade the version of dynarmic being used

bunnei:
@Hedges yeah - you need to remove that from your commits. I suggest creating a NEW branch from latest master, and then cherry-picking just the commits you need for your change, then force-pushing to replace this branch

Hedges:
@bunnei Right, makes sense.

jroweboy:
@bunnei if they have allow edits from maintainers you can do the rebase yourself and force push the branch to their remote

PR #649, opened 2018-07-12
Audout "Auto" functions

ogniK5377:
Audout autos are identical to their counterpart except for the buffer type which yuzu already handles for us.


src/core/hle/service/audio/audout_u.cpp
```diff
@@ -96,7 +96,7 @@ class IAudioOut final : public ServiceFramework<IAudioOut> {
         rb.PushCopyObjects(buffer_event);
     }
 
-    void AppendAudioOutBuffer(Kernel::HLERequestContext& ctx) {
+    void AppendAudioOutBufferImpl(Kernel::HLERequestContext& ctx) {
```
baka0815: Why the rename to ``*Impl``?

PR #653, opened 2018-07-12
More improvements to GDBStub

Hedges:
- Debugging of threads should work correctly with source and assembly level stepping and modifying registers and memory, meaning threads and callstacks are fully clickable in VS.
- List of modules is available to the client, with assumption that .nro and .nso are backed up by an .elf with symbols, while deconstructed ROMs keep N names.
- Initial support for floating point registers.

### DIFF
--- a/src/common/string_util.cpp
+++ b/src/common/string_util.cpp
@@ -134,7 +134,7 @@ bool SplitPath(const std::string& full_path, std::string* _pPath, std::string* _
     size_t dir_end = full_path.find_last_of("/"
 // windows needs the : included for something like just "C:" to be considered a directory
 #ifdef _WIN32
-                                            ":"
+                                            "\\:"
 #endif
     );
     if (std::string::npos == dir_end)
--- a/src/core/arm/unicorn/arm_unicorn.cpp
+++ b/src/core/arm/unicorn/arm_unicorn.cpp
@@ -193,11 +193,11 @@ void ARM_Unicorn::ExecuteInstructions(int num_instructions) {
         }
         Kernel::Thread* thread = Kernel::GetCurrentThread();
         SaveContext(thread->context);
-        if (last_bkpt_hit) {
+        if (last_bkpt_hit || (num_instructions == 1)) {
             last_bkpt_hit = false;
             GDBStub::Break();
+            GDBStub::SendTrap(thread, 5);
         }
-        GDBStub::SendTrap(thread, 5);
     }
 }
 
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -61,10 +61,12 @@ const u32 SIGTERM = 15;
 const u32 MSG_WAITALL = 8;
 #endif
 
-const u32 X30_REGISTER = 30;
+const u32 LR_REGISTER = 30;
 const u32 SP_REGISTER = 31;
 const u32 PC_REGISTER = 32;
 const u32 CPSR_REGISTER = 33;
+const u32 UC_ARM64_REG_Q0 = 34;
+const u32 FPSCR_REGISTER = 66;
 
 // For sample XML files see the GDB source /gdb/features
 // GDB also wants the l character at the start
@@ -130,6 +132,8 @@ static const char* target_xml =
     </flags>
     <reg name="cpsr" bitsize="32" type="cpsr_flags"/>
   </feature>
+  <feature name="org.gnu.gdb.aarch64.fpu">
+  </feature>
 </target>
 )";
 
@@ -144,6 +148,7 @@ static u32 latest_signal = 0;
 static bool memory_break = false;
 
 static Kernel::Thread* current_thread = nullptr;
+static u32 current_core = 0;
 
 // Binding to a port within the reserved ports range (0-1023) requires root permissions,
 // so default to a port outside of that range.
@@ -171,13 +176,34 @@ static std::map<u64, Breakpoint> breakpoints_execute;
 static std::map<u64, Breakpoint> breakpoints_read;
 static std::map<u64, Breakpoint> breakpoints_write;
 
+struct Module {
+    std::string name;
+    PAddr beg;
+    PAddr end;
+};
+
+static std::vector<Module> modules;
+
+void RegisterModule(std::string name, PAddr beg, PAddr end, bool add_elf_ext) {
+    Module module;
+    if (add_elf_ext) {
+        Common::SplitPath(name, nullptr, &module.name, nullptr);
+        module.name += ".elf";
+    } else {
+        module.name = std::move(name);
+    }
+    module.beg = beg;
+    module.end = end;
+    modules.push_back(std::move(module));
+}
+
 static Kernel::Thread* FindThreadById(int id) {
-    for (int core = 0; core < Core::NUM_CPU_CORES; core++) {
-        auto threads = Core::System::GetInstance().Scheduler(core)->GetThreadList();
-        for (auto thread : threads) {
+    for (u32 core = 0; core < Core::NUM_CPU_CORES; core++) {
+        const auto& threads = Core::System::GetInstance().Scheduler(core)->GetThreadList();
+        for (auto& thread : threads) {
             if (thread->GetThreadId() == id) {
-                current_thread = thread.get();
-                return current_thread;
+                current_core = core;
+                return thread.get();
             }
         }
     }
@@ -197,6 +223,8 @@ static u64 RegRead(int id, Kernel::Thread* thread = nullptr) {
         return thread->context.pc;
     } else if (id == CPSR_REGISTER) {
         return thread->context.cpsr;
+    } else if (id > CPSR_REGISTER && id < FPSCR_REGISTER) {
+        return thread->context.fpu_registers[id - UC_ARM64_REG_Q0][0];
     } else {
         return 0;
     }
@@ -215,6 +243,8 @@ static void RegWrite(int id, u64 val, Kernel::Thread* thread = nullptr) {
         thread->context.pc = val;
     } else if (id == CPSR_REGISTER) {
         thread->context.cpsr = val;
+    } else if (id > CPSR_REGISTER && id < FPSCR_REGISTER) {
+        thread->context.fpu_registers[id - (CPSR_REGISTER + 1)][0] = val;
     }
 }
 
@@ -534,7 +564,11 @@ static void HandleQuery() {
         SendReply("T0");
     } else if (strncmp(query, "Supported", strlen("Supported")) == 0) {
         // PacketSize needs to be large enough for target xml
-        SendReply("PacketSize=2000;qXfer:features:read+");
+        std::string buffer = "PacketSize=2000;qXfer:features:read+;qXfer:threads:read+";
+        if (modules.size()) {
+            buffer += ";qXfer:libraries:read+";
+        }
+        SendReply(buffer.c_str());
     } else if (strncmp(query, "Xfer:features:read:target.xml:",
                        strlen("Xfer:features:read:target.xml:")) == 0) {
         SendReply(target_xml);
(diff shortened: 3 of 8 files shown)


src/common/string_util.cpp
```diff
@@ -134,7 +134,7 @@ bool SplitPath(const std::string& full_path, std::string* _pPath, std::string* _
     size_t dir_end = full_path.find_last_of("/"
 // windows needs the : included for something like just "C:" to be considered a directory
 #ifdef _WIN32
-                                            ":"
+                                            "\\:"
```
Subv: why was this changed?

src/core/gdbstub/gdbstub.cpp
```diff
@@ -534,7 +564,11 @@ static void HandleQuery() {
         SendReply("T0");
     } else if (strncmp(query, "Supported", strlen("Supported")) == 0) {
         // PacketSize needs to be large enough for target xml
-        SendReply("PacketSize=2000;qXfer:features:read+");
+        std::string buffer = "PacketSize=2000;qXfer:features:read+;qXfer:threads:read+";
+        if (modules.size()) {
```
Subv: !modules.empty()

bunnei:
Much better. Thanks for updating! Will give this another day for folks to comment on, and hopefully merge tomorrow.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -711,8 +771,13 @@ static void ReadRegister() {
         LongToGdbHex(reply, RegRead(id, current_thread));
     } else if (id == CPSR_REGISTER) {
         IntToGdbHex(reply, (u32)RegRead(id, current_thread));
+    } else if (id >= UC_ARM64_REG_Q0 && id < FPSCR_REGISTER) {
+        LongToGdbHex(reply, RegRead(id, current_thread));
+    } else if (id == FPSCR_REGISTER) {
+        LongToGdbHex(reply, RegRead(998, current_thread));
     } else {
-        return SendReply("E01");
+        // return SendReply("E01");
```
Subv: Why is this commented out? If it's dead code then remove it, if it is not then leave a TODO.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -759,10 +834,17 @@ static void WriteRegister() {
         RegWrite(id, GdbHexToLong(buffer_ptr), current_thread);
     } else if (id == CPSR_REGISTER) {
         RegWrite(id, GdbHexToInt(buffer_ptr), current_thread);
+    } else if (id >= UC_ARM64_REG_Q0 && id < FPSCR_REGISTER) {
+        RegWrite(id, GdbHexToLong(buffer_ptr), current_thread);
+    } else if (id == FPSCR_REGISTER) {
+        RegWrite(998, GdbHexToLong(buffer_ptr), current_thread);
     } else {
-        return SendReply("E01");
+        // return SendReply("E01");
```
Subv: same

src/core/gdbstub/gdbstub.cpp
```diff
@@ -759,10 +834,17 @@ static void WriteRegister() {
...
+        RegWrite(997, GdbHexToLong(buffer_ptr), current_thread);
     }
 
+    Core::System::GetInstance().ArmInterface(current_core).LoadContext(current_thread->context);
```
Subv: Is this supposed to forcefully context-switch to the current thread? This will break a bunch of stuff that our scheduler is supposed to manage.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -773,18 +855,24 @@ static void WriteRegisters() {
...
+        } else if (reg >= UC_ARM64_REG_Q0 && reg < FPSCR_REGISTER) {
+            RegWrite(reg, GdbHexToLong(buffer_ptr + i * 16), current_thread);
+        } else if (reg == FPSCR_REGISTER) {
+            RegWrite(998, GdbHexToLong(buffer_ptr + i * 16), current_thread);
         } else {
             UNIMPLEMENTED();
         }
     }
 
+    Core::System::GetInstance().ArmInterface(current_core).LoadContext(current_thread->context);
```
Subv: see my previous comment.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -840,16 +932,17 @@ static void WriteMemory() {
...
+    send_trap = true;
 
     memory_break = is_memory_break;
 }
 
 /// Tell the CPU that it should perform a single step.
 static void Step() {
+    if (command_length > 1) {
+        RegWrite(PC_REGISTER, GdbHexToLong(command_buffer + 1), current_thread);
+        Core::System::GetInstance().ArmInterface(current_core).LoadContext(current_thread->context);
```
Subv: See previous comments.

src/common/string_util.cpp
Hedges: It did not actually work as it was before.

src/core/gdbstub/gdbstub.cpp
Hedges: I do not think it is important enough to be corrected right now.

src/core/gdbstub/gdbstub.cpp
Hedges: Just a comment, again - is that really so important at this very moment?

src/core/gdbstub/gdbstub.cpp
Hedges: Ditto

src/core/gdbstub/gdbstub.cpp
Hedges: Yes, this and the next one are supposed to forcefully switch thread contexts. Remember it is a debugger and its user should know what they are doing!

src/core/gdbstub/gdbstub.cpp
Subv: If it's not corrected during review before merging, it'll realistically never be corrected. This has been shown with many changes we've had in both citra and yuzu.

src/core/gdbstub/gdbstub.cpp
Hedges: To be precise - there are places where we actually switch contexts, what we are doing here is updating it, still - assumes advanced user who knows what they are doing.

src/core/gdbstub/gdbstub.cpp
Subv: See my previous reply, this comment says absolutely nothing about the code and even confuses the reader.

src/core/gdbstub/gdbstub.cpp
Subv: Forcefully loading the context without even changing the current thread is a complete hack in my opinion. Go through the scheduler if you want to change the current thread.

Hedges: Why so? We are updating the context between Unicorn steps and are a debugger so I do not consider it hackier than anything else system level :)

Subv: Using it only to avoid calling the ARM_Interface's SetReg function repeatedly is fine, provided that the current thread doesn't actually change.

Using it for anything that actually changes the current running thread from thread A to thread B is a massive hack that *will* break everything as soon as execution enters the HLE kernel again.

Subv: At the very least there should be a comment here indicating what this is actually intending to do. As you can see from this conversation, it is currently confusing.

src/core/gdbstub/gdbstub.cpp
Hedges: I can agree in general, but I am planning to continue working on them, so the change will happen. BTW, That is a matter of deciding what the purpose of version control systems is - should it only contain 'final' versions or is work in progress allowed?

Subv: In my opinion a PR should only be accepted if it is as reasonably complete as possible, or very self contained. Merging PRs in a piecewise fashion while leaving stuff as easy to fix as a leftover comment is not okay.

src/core/gdbstub/gdbstub.cpp
Hedges: Right, there is no running thread at this point, so there is no hack neither danger of breaking anything. Best would be to actually try it and then discuss - seems to work for me, wonder if it would work for you :)

Regarding the comment - if you update registers then you need to send them back to the emulation core which is kinda self-documenting behavior in my books.

src/core/gdbstub/gdbstub.cpp
Hedges: Well, the comment is not a leftover one - it is there on purpose to mark a POI.

src/core/gdbstub/gdbstub.cpp
Subv: It may be self documenting to you, the person who wrote the code, but adding a comment does not hurt.

Just because something works doesn't mean it is the correct (read: not a hack) way of doing it. 

What do you mean there is no running thread? There is always a running thread unless the core is idling or no binary is loaded, could you please elaborate on that?

src/core/gdbstub/gdbstub.cpp
Subv: Then leave a TODO explaining what is going on.

src/core/gdbstub/gdbstub.cpp
Hedges: I think you misread me - I meant that I was afraid you were not getting what it was doing, this is why I suggested that you tried it.

This code is executed between emulation steps, so 'the guest' is not running, so there are no running threads until the next step.

Hedges: I am afraid I missed your point. What I said was we were between emulation steps at this point so no guest threads were running/emulated. And - if you look above - I explained that I meant updating the context rather than switching it.

Hedges:
Yeah, it is confusing when Github is doing this… Hope that I found the right one and my answer plus comments I added are okay.

src/core/gdbstub/gdbstub.cpp
Hedges: A few more things. Rewriting registers is an extreme case and does not happen on a regular basis, a client has to deliberately request it. A debugger can use help from OS in such cases if this is needed, but there are also h/w debuggers (see e.g. devkits) that can do it without OS support, so they are effectively way more powerful. What I am trying to achieve is the best support for debugging, so I am going with emulating the h/w support as above, does that make sense?

src/core/arm/unicorn/arm_unicorn.cpp
```diff
@@ -193,11 +193,11 @@ void ARM_Unicorn::ExecuteInstructions(int num_instructions) {
         }
         Kernel::Thread* thread = Kernel::GetCurrentThread();
         SaveContext(thread->context);
-        if (last_bkpt_hit) {
+        if (last_bkpt_hit || (num_instructions == 1)) {
```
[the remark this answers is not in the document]
Hedges: I do not need to - they are conceptually the same. In both cases we will poke the debugger that will react accordingly.

src/core/gdbstub/gdbstub.cpp
```diff
@@ -711,8 +771,13 @@ static void ReadRegister() {
         LongToGdbHex(reply, RegRead(id, current_thread));
     } else if (id == CPSR_REGISTER) {
         IntToGdbHex(reply, (u32)RegRead(id, current_thread));
+    } else if (id >= UC_ARM64_REG_Q0 && id < FPSCR_REGISTER) {
+        LongToGdbHex(reply, RegRead(id, current_thread));
+    } else if (id == FPSCR_REGISTER) {
+        LongToGdbHex(reply, RegRead(998, current_thread));
```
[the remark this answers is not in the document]
Hedges: Well, it is again some deliberate decision to leave it as it is. This is wip version and as such contains things that catch your attention and are meant to be replaced/improved. I am still not sure we see the VCS same way.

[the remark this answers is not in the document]
MerryMage: I generally see a PR as a request for a merge, and hence usually one would expect code to be in a mergeable state. Perhaps mark your PR as [WIP] if this is not the case.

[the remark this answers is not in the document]
Hedges: It actually is in a very mergeable state and self contained. The strange 997 and 998 are only used to tag my next targets, though I can change them if it is really that important, but then again - I am afraid we are missing point of having a VCS.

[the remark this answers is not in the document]
Hedges: Amended.

Hedges:
Cool, cheers!

src/core/arm/unicorn/arm_unicorn.cpp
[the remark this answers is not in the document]
Hedges: Ah, I have only got now what you meant. So, it made no difference when I was assuming Unicorn being the only one interface used for debugging. I have started to interface to Dynarmic now and I can see how it has to be refined, thanks :)

PR #659, opened 2018-07-14
gl_rasterizer_cache: Implement depth format Z16_UNORM.

bunnei:


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -112,6 +112,8 @@ static constexpr std::array<FormatTuple, SurfaceParams::MaxPixelFormat> tex_form
     {GL_DEPTH24_STENCIL8, GL_DEPTH_STENCIL, GL_UNSIGNED_INT_24_8, ComponentType::UNorm,
      false},                                                                            // S8Z24
     {GL_DEPTH_COMPONENT32F, GL_DEPTH_COMPONENT, GL_FLOAT, ComponentType::Float, false}, // Z32F
+    {GL_DEPTH_COMPONENT16, GL_DEPTH_COMPONENT, GL_UNSIGNED_SHORT, ComponentType::UNorm,
```
Subv: I am pretty sure the type here should be GL_HALF_FLOAT

MoochMcGee: But then that wouldn't be a UNORM, would it? ;)

bunnei: I *think* what I have is right. `Z16_UNORM` is from envytools - why would this be a float format? Furthermore, data seems fine in renderdoc - although tbf, this is not critical for the game I am testing with, so I'm not 100% confident. Please lmk if you feel strongly that this is wrong.

PR #664, opened 2018-07-14
Minor logging improvements

jroweboy:
* Don't lock the logging queue while writing. Removes the speed penalty for having the console window open
* If the logger has any classes set to debug, it will now dump all logs in the queue on exit instead of the last 100.


bunnei:
Changes lgtm - please resolve CI build break though

PR #671, opened 2018-07-16
scheduler: Clear exclusive state when switching contexts

MerryMage:
Exclusive state should not be shared between disparate threads.

On Horizon this would be implicit in the `ERET` instruction when the kernel returns to usermode.

Split off from #638.


bunnei:
LGTM- ping @Hexagon12 to confirm the regressions from #638 are not here.

PR #673, opened 2018-07-17
nvflinger: Fix for BufferQueue event handling.

bunnei:
Gets SteamWorld Dig 2 booting, probably fixes some other things.

image

### DIFF
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -29,7 +29,8 @@ void SessionRequestHandler::ClientDisconnected(SharedPtr<ServerSession> server_s
 
 SharedPtr<Event> HLERequestContext::SleepClientThread(SharedPtr<Thread> thread,
                                                       const std::string& reason, u64 timeout,
-                                                      WakeupCallback&& callback) {
+                                                      WakeupCallback&& callback,
+                                                      Kernel::SharedPtr<Kernel::Event> event) {
 
     // Put the client thread to sleep until the wait event is signaled or the timeout expires.
     thread->wakeup_callback =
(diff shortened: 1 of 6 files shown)


src/core/hle/kernel/hle_ipc.h
```diff
@@ -118,10 +118,12 @@ class HLERequestContext {
      * @param callback Callback to be invoked when the thread is resumed. This callback must write
      * the entire command response once again, regardless of the state of it before this function
      * was called.
+     * @param event Event to use to wake up the thread. If unspecified, an event will be created.
      * @returns Event that when signaled will resume the thread and call the callback function.
      */
     SharedPtr<Event> SleepClientThread(SharedPtr<Thread> thread, const std::string& reason,
-                                       u64 timeout, WakeupCallback&& callback);
+                                       u64 timeout, WakeupCallback&& callback,
+                                       Kernel::SharedPtr<Kernel::Event> event = {});
```
Subv: use nullptr instead of {}

src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -41,9 +42,14 @@ SharedPtr<Event> HLERequestContext::SleepClientThread(SharedPtr<Thread> thread,
...
 
-    auto event = Kernel::Event::Create(Kernel::ResetType::OneShot, "HLE Pause Event: " + reason);
+    if (!event) {
+        // Create event if not provided
+        event = Kernel::Event::Create(Kernel::ResetType::OneShot, "HLE Pause Event: " + reason);
+    }
+
     thread->status = THREADSTATUS_WAIT_HLE_EVENT;
     thread->wait_objects = {event};
+    event->Clear();
```
Subv: move this to after the if, before       thread->status = THREADSTATUS_WAIT_HLE_EVENT;

PR #679, opened 2018-07-18
game_list: Remove unnecessary QString initialization in KeyReleaseEater

lioncash:
QString initializes to an empty string by default, so this does nothing meaningful. While we're at it, use a constructor initializer list for initializing the gamelist member variable.


FearlessTobi:
If necessary, can you backport this PR and the other one to Citra?

lioncash:
Sure thing, I'll do it later today

PR #680, opened 2018-07-18
decoders: Fix calc of swizzle image_width_in_gobs.

bunnei:
Fixes textures in SteamWorld:

image

### DIFF
--- a/src/video_core/textures/decoders.cpp
+++ b/src/video_core/textures/decoders.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <cmath>
 #include <cstring>
 #include "common/assert.h"
 #include "core/memory.h"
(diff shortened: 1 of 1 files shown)


src/video_core/textures/decoders.cpp
```diff
@@ -17,7 +18,7 @@ namespace Texture {
  * Taken from the Tegra X1 TRM.
  */
 static u32 GetSwizzleOffset(u32 x, u32 y, u32 image_width, u32 bytes_per_pixel, u32 block_height) {
-    u32 image_width_in_gobs = image_width * bytes_per_pixel / 64;
+    u32 image_width_in_gobs = static_cast<u32>(std::ceil((image_width * bytes_per_pixel) / 64.f));
```
gdkchan: You could use a rounded integer division, like:
```cpp
u32 image_width_in_gobs = (image_width * bytes_per_pixel + 63) / 64;
```
Should be a bit faster.

bunnei: Indeed -thx

PR #683, opened 2018-07-18
Touchscreen Support

DarkLordZach:
Currently only single touch

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -5,6 +5,7 @@
 #include <atomic>
 #include "common/logging/log.h"
 #include "core/core_timing.h"
+#include "core/frontend/emu_window.h"
 #include "core/frontend/input.h"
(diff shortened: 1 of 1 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -164,7 +164,21 @@ class IAppletResource final : public ServiceFramework<IAppletResource> {
...
+        TouchScreenEntryTouch touch_entry{};
+        float x;
+        float y;
+        bool pressed;
+        std::tie(x, y, pressed) = touch_device->GetStatus();
+        touch_entry.x = static_cast<u16>(x * Layout::ScreenUndocked::Width);
+        touch_entry.y = static_cast<u16>(y * Layout::ScreenUndocked::Height);
+
+        if (pressed) {
+            touchscreen.entries[curr_entry].header.num_touches = 1;
```
Subv: leave a TODO saying that multitouch is not implemented

PR #689, opened 2018-07-18
program_metadata: Take std::vector by const reference in Load()

lioncash:
This is only ever read from in the function itself, it's not moved or copied anywhere else, so this is kind of wasteful. Given it has a const qualifier, it was likely intended to be a reference anyways, however the ampersand was forgotten.


bunnei:
@lioncash Please resolve conflicts

lioncash:
This one's no longer relevant

PR #695, opened 2018-07-19
NRO Assets and NACP File Format

DarkLordZach:
Part of a larger change, but its probably better to split it up.

### DIFF
--- a/src/core/file_sys/control_metadata.cpp
+++ b/src/core/file_sys/control_metadata.cpp
@@ -1,0 +1,42 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/string_util.h"
+#include "common/swap.h"
+#include "core/file_sys/control_metadata.h"
+
+namespace FileSys {
+
+std::string LanguageEntry::GetApplicationName() const {
+    return Common::StringFromFixedZeroTerminatedBuffer(application_name.data(), 0x200);
+}
+
+std::string LanguageEntry::GetDeveloperName() const {
+    return Common::StringFromFixedZeroTerminatedBuffer(developer_name.data(), 0x100);
+}
+
+NACP::NACP(VirtualFile file_) : file(std::move(file_)), raw(std::make_unique<RawNACP>()) {
+    file->ReadObject(raw.get());
+}
+
+const LanguageEntry& NACP::GetLanguageEntry(Language language) const {
+    return raw->language_entries.at(static_cast<u8>(language));
+}
+
+std::string NACP::GetApplicationName(Language language) const {
+    return GetLanguageEntry(language).GetApplicationName();
+}
+
+std::string NACP::GetDeveloperName(Language language) const {
+    return GetLanguageEntry(language).GetDeveloperName();
+}
+
+u64 NACP::GetTitleId() const {
+    return raw->title_id;
+}
+
+std::string NACP::GetVersionString() const {
+    return Common::StringFromFixedZeroTerminatedBuffer(raw->version_string.data(), 0x10);
+}
+} // namespace FileSys
--- a/src/core/file_sys/control_metadata.h
+++ b/src/core/file_sys/control_metadata.h
@@ -1,0 +1,81 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <memory>
+#include <string>
+#include "common/common_funcs.h"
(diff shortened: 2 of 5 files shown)


src/core/file_sys/control_metadata.cpp
```diff
@@ -0,0 +1,44 @@
...
+std::string LanguageEntry::GetApplicationName() {
+    return Common::StringFromFixedZeroTerminatedBuffer(application_name.data(), 0x200);
+}
+
+std::string LanguageEntry::GetDeveloperName() {
+    return Common::StringFromFixedZeroTerminatedBuffer(developer_name.data(), 0x100);
+}
+
+NACP::NACP(VirtualFile file_) : file(std::move(file_)) {
+    RawNACP nacp{};
```
Subv: You can use 
```
raw = std::make_unique<RawNCP>();
file->ReadObject(raw.get());
```

src/core/file_sys/control_metadata.h
```diff
@@ -0,0 +1,79 @@
...
+
+namespace FileSys {
+
+// A localized entry containing strings within the NACP.
+// One for each language of type Language.
+struct LanguageEntry {
+    std::array<char, 0x200> application_name;
+    std::array<char, 0x100> developer_name;
+
+    std::string GetApplicationName();
```
lioncash: These two functions can likely be const member functions

src/core/file_sys/control_metadata.h
```diff
@@ -0,0 +1,79 @@
...
+    Taiwanese = 13,
+    Chinese = 14,
+};
+
+// A class representing the format used my NX metadata files, typically named Control.nacp.
+// These store application name, dev name, title id, and other miscellaneous data.
+class NACP {
+public:
+    explicit NACP(VirtualFile file);
+    LanguageEntry GetLanguageEntry(Language language = Language::AmericanEnglish);
```
lioncash: All of these getters can likely be const member functions.

src/core/file_sys/control_metadata.h
```diff
@@ -0,0 +1,79 @@
...
+    Dutch = 8,
+    CanadianFrench = 9,
+    Portugese = 10,
+    Russian = 11,
+    Korean = 12,
+    Taiwanese = 13,
+    Chinese = 14,
+};
+
+// A class representing the format used my NX metadata files, typically named Control.nacp.
```
lioncash: my -> by

src/core/file_sys/control_metadata.cpp
```diff
@@ -0,0 +1,43 @@
...
+std::string LanguageEntry::GetDeveloperName() {
+    return Common::StringFromFixedZeroTerminatedBuffer(developer_name.data(), 0x100);
+}
+
+NACP::NACP(VirtualFile file_) : file(std::move(file_)) {
+    raw = std::make_unique<RawNACP>();
+    file->ReadObject(raw.get());
+}
+
+LanguageEntry NACP::GetLanguageEntry(Language language) {
```
lioncash: This should likely have an overload returning by reference and by const reference (or only const-reference if it should be immutable), otherwise this does a pointlessly large copy.

src/core/file_sys/control_metadata.cpp
```diff
@@ -0,0 +1,43 @@
...
+std::string LanguageEntry::GetApplicationName() {
+    return Common::StringFromFixedZeroTerminatedBuffer(application_name.data(), 0x200);
+}
+
+std::string LanguageEntry::GetDeveloperName() {
+    return Common::StringFromFixedZeroTerminatedBuffer(developer_name.data(), 0x100);
+}
+
+NACP::NACP(VirtualFile file_) : file(std::move(file_)) {
+    raw = std::make_unique<RawNACP>();
```
lioncash: `raw` can go in the constructor initializer list as well.

src/core/file_sys/control_metadata.h
```diff
@@ -0,0 +1,79 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
```
lioncash: `<memory>` and `<string>` should also be included.

src/core/file_sys/control_metadata.h
```diff
@@ -0,0 +1,79 @@
...
+#include "common/common_funcs.h"
```
lioncash: This should likely be `common_types.h` instead.

DarkLordZach: nah, theres the INSERT_PADDING_BYTES macro in RawNACP

PR #697, opened 2018-07-19
gl_state: Temporarily disable culling and depth test.

bunnei:
Stole the idea from Ryujinx... Until we thoroughly understand how these work (and have games that really need them), we should just disable them. Depth test and culling issues hide games that might be rendering,  and make it difficult to keep track of what might be rendering (and what isn't).

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -795,7 +795,9 @@ void RasterizerOpenGL::SyncClipCoef() {
 void RasterizerOpenGL::SyncCullMode() {
     const auto& regs = Core::System::GetInstance().GPU().Maxwell3D().regs;
 
-    state.cull.enabled = regs.cull.enabled != 0;
+    // TODO(bunnei): Enable the below once more things work - until then, this may hide regressions
+    // state.cull.enabled = regs.cull.enabled != 0;
+    state.cull.enabled = false;
 
     if (state.cull.enabled) {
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -74,14 +74,15 @@ OpenGLState::OpenGLState() {
...
+    // TODO(bunnei): Enable the below once more things work - until then, this may hide regressions
     // Culling
-    if (cull.enabled != cur_state.cull.enabled) {
-        if (cull.enabled) {
-            glEnable(GL_CULL_FACE);
-        } else {
-            glDisable(GL_CULL_FACE);
-        }
-    }
+    // if (cull.enabled != cur_state.cull.enabled) {
```
Subv: I suggest you disable this in SyncCullMode instead

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -795,7 +795,9 @@ void RasterizerOpenGL::SyncClipCoef() {
 void RasterizerOpenGL::SyncCullMode() {
     const auto& regs = Core::System::GetInstance().GPU().Maxwell3D().regs;
 
-    state.cull.enabled = regs.cull.enabled != 0;
+    // TODO(bunnei): Enable the below once more things work - until then, this may hide regressions
+    // state.cull.enabled = regs.cull.enabled != 0;
+    state.cull.enabled = 0;
```
Subv: false?

PR #712, opened 2018-07-19
 fsp_srv: Misc individual changes

lioncash:
This includes a few relatively contained changes that don't really warrant being broken into their own individual PRs.

### DIFF
--- a/src/core/hle/service/filesystem/fsp_srv.cpp
+++ b/src/core/hle/service/filesystem/fsp_srv.cpp
@@ -138,15 +138,15 @@ private:
         const std::vector<u8> data = ctx.ReadBuffer();
 
         ASSERT_MSG(
-            data.size() <= length,
+            static_cast<s64>(data.size()) <= length,
             "Attempting to write more data than requested (requested={:016X}, actual={:016X}).",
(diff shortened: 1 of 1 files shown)


src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -133,17 +135,16 @@ class IFile final : public ServiceFramework<IFile> {
...
 
         ASSERT_MSG(
             data.size() <= length,
             "Attempting to write more data than requested (requested={:016X}, actual={:016X}).",
             length, data.size());
 
-        std::copy(data.begin(), data.end(), actual_data.begin());
         // Write the data to the Storage backend
-        auto written = backend->WriteBytes(data, offset);
+        std::vector<u8> actual_data(data.begin(), data.begin() + length);
```
Subv: What happens in the case length > data.size()? won't it try to copy from out of bounds?

lioncash: I can add an assertion in for that case if it's preferable (or clamp length, though that seems kind of iffy).

PR #725, opened 2018-07-19
pl_u: Specify correct size for buffers in GetSharedFontInOrderOfPriority()

lioncash:
This `WriteBuffer()` overload expects its size argument to be in bytes, not elements.


lioncash:
Error is related to downloading the qt binary

bunnei:
Yep- reran build.

PR #726, opened 2018-07-19
hle_ipc: Introduce generic WriteBuffer overload for multiple container types

lioncash:
This introduces a slightly more generic variant of WriteBuffer(). Notably, this variant doesn't constrain the arguments to only accepting std::vector instances. It accepts whatever adheres to the ContiguousContainer concept in the C++ standard library.

This essentially means, `std::array`, `std::string`, and `std::vector` can be used directly with this interface. The interface no longer forces you to solely use containers that dynamically allocate.

To ensure our overloads play nice with one another, we only enable the container-based WriteBuffer if the argument is not a pointer, otherwise we fall back to the pointer-based one.

This overload makes it less error-prone when writing elements out, as one must ensure that the container size is multiplied by the element size in order to make the size be in terms of bytes. This overload does that work for us.

### DIFF
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -301,10 +301,6 @@ size_t HLERequestContext::WriteBuffer(const void* buffer, size_t size, int buffe
     return size;
 }
 
-size_t HLERequestContext::WriteBuffer(const std::vector<u8>& buffer, int buffer_index) const {
-    return WriteBuffer(buffer.data(), buffer.size(), buffer_index);
-}
-
 size_t HLERequestContext::GetReadBufferSize(int buffer_index) const {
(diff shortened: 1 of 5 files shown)


src/core/hle/service/acc/acc.cpp
```diff
@@ -96,15 +96,15 @@ void Module::Interface::GetUserExistence(Kernel::HLERequestContext& ctx) {
 void Module::Interface::ListAllUsers(Kernel::HLERequestContext& ctx) {
     LOG_WARNING(Service_ACC, "(STUBBED) called");
     constexpr std::array<u128, 10> user_ids{DEFAULT_USER_ID};
-    ctx.WriteBuffer(user_ids.data(), user_ids.size());
+    ctx.WriteBuffer(user_ids);
```
Subv: This is going to cause errors, but not because of this change, our implementation of this function is _very_ wrong

PR #741, opened 2018-07-20
ipc_helpers: Add PushEnum() member function to ResponseBuilder

lioncash:
Allows pushing strongly-typed enum members without the need to always cast them at the call sites.

Note that we *only* allow strongly-typed enums in this case. The reason for this is that strongly typed enums have a guaranteed defined size, so the size of the data being pushed is always deterministic. With regular enums this can be a little more error-prone, so we disallow them.

This function simply uses the underlying type of the enum to determine the size of the data. For example, if an enum is defined as:

```cpp
enum class SomeEnum : u16 {
  SomeEntry
};
```

if `PushEnum(SomeEnum::SomeEntry);` is called, then it will push a u16-size amount of data.

### DIFF
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -175,6 +175,25 @@ public:
     void Push(const First& first_value, const Other&... other_values);
 
     /**
+     * Helper function for pushing strongly-typed enumeration values.
+     *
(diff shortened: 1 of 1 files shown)


lioncash:
@Subv should we disallow pushing u64-sized enums like we do with citra?

PR #757, opened 2018-07-21
string_util: Remove unnecessary ifdef in UTF16ToUTF8() and UTF8ToUTF16()

lioncash:
VS 2017 provides the char16_t and char32_t types, so the ifdef isn't necessary anymore.


lioncash:
will close this, evidently MSVC doesn't completely provide them.

PR #771, opened 2018-07-22
shader_bytecode: Add constexpr to default constructor of Attribute and Sampler

lioncash:
We already have a constexpr constructor that takes a paremeter, so the other constructor should have it too.


lioncash:
GCC seems to lose its marbles over this, so I'll close this for now

PR #676, opened 2018-07-17
Virtual Filesystem 2: Electric Boogaloo

DarkLordZach:
Now including:
- Fully-virtualized filesystem allowing for fancy fs-related stuff to be easier (Updates, DLC, new formats, encryption, et cetera. Keep your eyes peeled)
- `VirtualFile` and `VirtualDir` typedefs to reprent files and dirs respectively
- Basic implementations: `RealVfsFile`/`RealVfsDirectory` for files on the user's computer, `OffsetVfsFile` for a file at specified offset and size in another, `ReadOnlyVfsDirectory` a partial implementation that defaults all methods that only make sense for writable dirs.
- Move NCA implementation to its own file `content_archive.[h|cpp]`
- A fix for NCA file loading (they now boot -- it was an issue with the romfs offset)
- Cleanup in the `core/file_sys` directory/namespace
- Documentation is in `vfs.h`
- `VfsDirectoryServiceWrapper` covers relevant VfsDirectory methods in ones that return `ResultCode`/`ResultVal` for easier interoperability with the Switch's FS services.
- Fixes random save-data corruption issues (confirmed found in Binding of Isaac)
- Remove unused path implementation from citra

PR #776, opened 2018-07-23
VFS Regression and Accuracy Fixes

DarkLordZach:
Fixes all of the resgressions caused by #676.

- Sanitize paths being passed into `VfsDirectoryServiceWrapper`. (Allows stardew valley to progress to character create screen).
- Use the correct interpretation of `FileSys::Mode` as a bitmask and not mutex. 
- Delete files in the dir that is requested, not the root dir (Fixes farming simulator).

Tested on: Isaac, ARMS, SMO, Cave Story, Stardew, Farming Simulator.

Sorry to anyone disrupted by the regressions from #676, I'm working hard to fix them but it takes time. If you notice anything else wrong related to VFS, feel free to leave an issue in github or mention me in the yuzu discord.

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -826,7 +826,7 @@ std::string_view GetPathWithoutTop(std::string_view path) {
     }
 
     while (path[0] == '\\' || path[0] == '/') {
-        path.remove_suffix(1);
+        path.remove_prefix(1);
         if (path.empty()) {
             return path;
         }
@@ -868,6 +868,15 @@ std::string_view RemoveTrailingSlash(std::string_view path) {
     }
 
     return path;
+}
+
+std::string_view SanitizePath(std::string_view path_) {
+    std::string path(path_);
+    std::replace(path.begin(), path.end(), '\\', '/');
+    path.erase(std::unique(path.begin(), path.end(),
+                           [](char c1, char c2) { return c1 == '/' && c2 == '/'; }),
+               path.end());
+    return RemoveTrailingSlash(path);
 }
 
 IOFile::IOFile() {}
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -870,7 +870,7 @@ std::string_view RemoveTrailingSlash(std::string_view path) {
     return path;
 }
 
-std::string_view SanitizePath(std::string_view path_) {
+std::string SanitizePath(std::string_view path_) {
     std::string path(path_);
     std::replace(path.begin(), path.end(), '\\', '/');
     path.erase(std::unique(path.begin(), path.end(),
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -876,7 +876,7 @@ std::string_view SanitizePath(std::string_view path_) {
     path.erase(std::unique(path.begin(), path.end(),
                            [](char c1, char c2) { return c1 == '/' && c2 == '/'; }),
                path.end());
-    return RemoveTrailingSlash(path);
+    return std::string(RemoveTrailingSlash(path));
 }
 
 IOFile::IOFile() {}
--- a/src/core/file_sys/vfs_real.cpp
+++ b/src/core/file_sys/vfs_real.cpp
@@ -13,24 +13,33 @@
 
 namespace FileSys {
 
-static std::string PermissionsToCharArray(Mode perms) {
-    std::string out;
-    switch (perms) {
-    case Mode::Read:
-        out += "r";
-        break;
-    case Mode::Write:
-        out += "r+";
-        break;
-    case Mode::Append:
-        out += "a";
-        break;
+static std::string ModeFlagsToString(Mode mode) {
+    std::string mode_str;
+    u32 mode_flags = static_cast<u32>(mode);
+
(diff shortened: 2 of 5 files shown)


src/core/file_sys/vfs_real.cpp
```diff
@@ -93,7 +102,8 @@ RealVfsDirectory::RealVfsDirectory(const std::string& path_, Mode perms_)
       path_components(FileUtil::SplitPathComponents(path)),
       parent_components(FileUtil::SliceVector(path_components, 0, path_components.size() - 1)),
       perms(perms_) {
-    if (!FileUtil::Exists(path) && (perms == Mode::Write || perms == Mode::Append))
+    // Write|Append
+    if (!FileUtil::Exists(path) && (static_cast<u32>(perms) & 0x6) > 0)
```
Subv: Use the enum when checking the bitflags, add a new WriteAppend member to it if necessary.

Don't use > 0 when checking for bitflags, just use `(static_cast<u32>(perms) & static_cast<u32>(Mode::WriteAppend))`

Subv: There are a bunch of other places in this file where permission checks need to be changed to a bitcheck instead of an equality check, please do them all.

src/common/file_util.cpp
```diff
@@ -870,6 +870,15 @@ std::string_view RemoveTrailingSlash(std::string_view path) {
     return path;
 }
 
+std::string_view SanitizePath(std::string_view path_) {
+    std::string path(path_);
+    std::replace(path.begin(), path.end(), '\\', '/');
+    path.erase(std::unique(path.begin(), path.end(),
+                           [](char c1, char c2) { return c1 == '/' && c2 == '/'; }),
+               path.end());
+    return RemoveTrailingSlash(path);
```
lioncash: This will not do what you expect. `std::string_view` is a non-owning view into a string and you're passing in a local. This is effectively returning the buffer of the string on the stack. This will be junk memory once the function returns.

src/core/file_sys/vfs_real.cpp
```diff
@@ -13,24 +13,33 @@
...
-        break;
-    case Mode::Write:
-        out += "r+";
-        break;
-    case Mode::Append:
-        out += "a";
-        break;
+static std::string ModeFlagsToString(Mode mode) {
+    std::string mode_str;
+    u32 mode_flags = static_cast<u32>(mode);
```
lioncash: You may just want to consider defining `operator&` and whatever other operators are necessary for Mode, so that you don't need to cast all the time.

src/common/file_util.cpp
```diff
@@ -876,7 +876,7 @@ std::string_view SanitizePath(std::string_view path_) {
     path.erase(std::unique(path.begin(), path.end(),
                            [](char c1, char c2) { return c1 == '/' && c2 == '/'; }),
                path.end());
-    return RemoveTrailingSlash(path);
+    return std::string(RemoveTrailingSlash(path));
```
Subv: Make this function return std::string, otherwise you're still returning a view into a soon-to-be-destroyed object.

src/core/file_sys/vfs_real.cpp
```diff
@@ -53,11 +60,11 @@ std::shared_ptr<VfsDirectory> RealVfsFile::GetContainingDirectory() const {
 }
 
 bool RealVfsFile::IsWritable() const {
-    return perms == Mode::Append || perms == Mode::Write;
+    return perms & Mode::WriteAppend;
```
lioncash: These can case warnings on some compilers, since this isn't technically a boolean return value, but a u32.

this should be:

```cpp
return (perms & Mode::WriteAppend) != 0;
```

src/core/file_sys/vfs_real.cpp
```diff
@@ -53,11 +60,11 @@ std::shared_ptr<VfsDirectory> RealVfsFile::GetContainingDirectory() const {
...
 }
 
 bool RealVfsFile::IsReadable() const {
-    return perms == Mode::Read || perms == Mode::Write;
+    return perms & Mode::ReadWrite;
```
lioncash: Ditto

src/core/file_sys/vfs_real.cpp
```diff
@@ -120,11 +127,11 @@ std::vector<std::shared_ptr<VfsDirectory>> RealVfsDirectory::GetSubdirectories()
 }
 
 bool RealVfsDirectory::IsWritable() const {
-    return perms == Mode::Write || perms == Mode::Append;
+    return perms & Mode::WriteAppend;
```
lioncash: Ditto

src/core/file_sys/vfs_real.cpp
```diff
@@ -120,11 +127,11 @@ std::vector<std::shared_ptr<VfsDirectory>> RealVfsDirectory::GetSubdirectories()
...
 }
 
 bool RealVfsDirectory::IsReadable() const {
-    return perms == Mode::Read || perms == Mode::Write;
+    return perms & Mode::ReadWrite;
```
lioncash: Ditto

PR #789, opened 2018-07-24
maxwell_to_gl: Implement Texture::WrapMode::Border.

bunnei:


Subv:
This won't actually work until we implement the border color here https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/renderer_opengl/gl_rasterizer.cpp#L632

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -601,7 +601,6 @@ void RasterizerOpenGL::SamplerInfo::Create() {
     sampler.Create();
     mag_filter = min_filter = Tegra::Texture::TextureFilter::Linear;
     wrap_u = wrap_v = Tegra::Texture::WrapMode::Wrap;
-    border_color_r = border_color_g = border_color_b = border_color_a = 0;
```
Subv: keep the initial default value of 0

PR #804, opened 2018-07-24
svc: Log parameters in SetMemoryAttribute()

lioncash:
Provides slightly more context than only logging out the address value.


lioncash:
failure seems unrelated

PR #826, opened 2018-07-25
service: Add erpt and eupld services

lioncash:
Adds the skeletons for both services based off information on Switch Brew.


Subv:
Please fix clang-format

PR #829, opened 2018-07-26
GPU: Allow using R16F as a render target format.

Subv:


greggameplayer:
@Subv Can you also do R16UI, R16I, R16S & R16 RenderTargetFormats ?

Subv:
@greggameplayer we haven't found any games that use them, so i don't see a reason to add them yet.

greggameplayer:
@Subv So you rush the work instead of doing it right even if we haven't currently any games who use these they will be needed anyway.

PR #842, opened 2018-07-27
Initial implementation of Audio Core

bunnei:
Summary of changes:
* Adds audio_core project for audio related things
* Adds basic state management for audio streams and buffers
* Hooks things up for `IAudioOut`
* Other `IAudioOut` emulation improvements

Should fix most games that had `audout` issues. Timing seems to be very sensitive. Due to our core timing being not very accurate, expect things that depend on `audout` timing to be janky (e.g. framerate in SnipperClips, cutscene video playback in One Piece, etc.). Thanks to @ogniK5377 for doing a lot of the initial work for this impl.

Some things that now work:
image
image
image


src/audio_core/audio_out.h
```diff
@@ -0,0 +1,46 @@
...
+namespace AudioCore {
+
+using StreamPtr = std::shared_ptr<Stream>;
+
+/**
+ * Represents an audio playback interface, used to open and play audio streams
+ */
+class AudioOut {
+public:
+    AudioOut() = default;
```
lioncash: This can be omitted, since there's no other constructor present.

src/audio_core/buffer.h
```diff
@@ -0,0 +1,37 @@
...
+        return data;
+    }
+
+    /// Returns the buffer tag, this is provided by the game to the audout service
+    Tag GetTag() const {
+        return tag;
+    }
+
+private:
+    const Tag tag;
```
lioncash: Note that this being `const` internally prevents the move constructor from working and will always result in copies being made.

src/audio_core/stream.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "common/assert.h"
+#include "common/logging/log.h"
+#include "core/core_timing.h"
+#include "core/core_timing_util.h"
+
+#include "audio_core/stream.h"
+
+namespace AudioCore {
+
+static constexpr size_t MaxAudioBufferCount{32};
```
lioncash: `static` can be omitted here.

src/audio_core/stream.h
```diff
@@ -0,0 +1,89 @@
...
+    /// Plays the next queued buffer in the audio stream, starting playback if necessary
+    void PlayNextBuffer();
+
+    /// Releases the actively playing buffer, signalling that it has been completed
+    void ReleaseActiveBuffer();
+
+    /// Gets the number of core cycles when the specified buffer will be released
+    s64 GetBufferReleaseCycles(const Buffer& buffer) const;
+
+    const int sample_rate;                  ///< Sample rate of the stream
```
lioncash: Ditto about move constructors, as well as with `format` below.

src/core/hle/service/audio/audout_u.cpp
```diff
@@ -5,195 +5,201 @@
...
+    IPC::RequestParser rp{ctx};
+    auto params{rp.PopRaw<AudoutParams>()};
+    if (params.channel_count <= 2) {
+        // Mono does not exist for audout
+        params.channel_count = 2;
+    } else {
+        params.channel_count = 6;
     }
+    if (!params.sample_rate) {
+        params.sample_rate = DefaultSampleRate;
```
gdkchan: Unused?

src/core/hle/service/audio/audout_u.cpp
```diff
@@ -5,195 +5,201 @@
...
-        const u64 key{rp.Pop<u64>()};
-        queue_keys.insert(queue_keys.begin(), key);
+        const auto& input_buffer{ctx.ReadBuffer()};
+        ASSERT_MSG(input_buffer.size() == sizeof(AudioBuffer),
+                   "AudioBuffer input is an invalid size!");
+        AudioBuffer audio_buffer{};
+        std::memcpy(&audio_buffer, input_buffer.data(), sizeof(AudioBuffer));
+        const u64 tag{rp.Pop<u64>()};
+
+        std::vector<u8> data;
```
```cpp
std::vector<u8> data(audio_buffer.buffer_size)
```

src/core/hle/service/audio/audout_u.cpp
```diff
@@ -5,195 +5,201 @@
...
-
-        if (queue_keys.size()) {
-            key = queue_keys.back();
-            queue_keys.pop_back();
+        LOG_DEBUG(Service_Audio, "called {}", ctx.Description());
+        IPC::RequestParser rp{ctx};
+        const u64 max_count{ctx.GetWriteBufferSize() / sizeof(u64)};
+        const auto released_buffers{audio_core.GetReleasedBuffers(stream, max_count)};
+
+        std::vector<u64> tags;
```
lioncash: This could also be:

```cpp
std::vector<u64> tags = released_buffers;
tags.resize(max_count);
```

which gets rid of the need for the loop.

src/audio_core/stream.cpp
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return false;
+}
+
+bool Stream::ContainsBuffer(Buffer::Tag tag) const {
+    ASSERT_MSG(false, "Unimplemented");
+    return {};
+}
+
+std::vector<Buffer::Tag> Stream::GetReleasedBuffers(size_t max_count) {
```
lioncash: It doesn't seem like a good naming choice to use `Get` here without clarification, especially when it's popping the internal `released_buffer` queue (more than getting is being done). One naming scheme that might work here is `GetAndReleaseBuffers();`

src/core/hle/service/audio/audout_u.h
```diff
@@ -25,6 +37,9 @@ class AudOutU final : public ServiceFramework<AudOutU> {
     void ListAudioOutsImpl(Kernel::HLERequestContext& ctx);
     void OpenAudioOutImpl(Kernel::HLERequestContext& ctx);
 
+    const std::array<char, 10> DefaultDevice{{"DeviceOut"}};
```
lioncash: This and the member variable below can be `static constexpr` (or be moved to the cpp file and be made `constexpr`)

src/core/hle/service/audio/audout_u.cpp
```diff
@@ -5,195 +5,201 @@
...
 
     void StartAudioOut(Kernel::HLERequestContext& ctx) {
-        LOG_WARNING(Service_Audio, "(STUBBED) called");
+        LOG_DEBUG(Service_Audio, "called");
 
-        // Start audio
-        audio_out_state = AudioState::Started;
+        if (stream->IsPlaying()) {
+            IPC::ResponseBuilder rb{ctx, 2};
+            rb.Push(ResultCode(ErrorModule::Audio, 2));
```
Subv: Add an enum for that error description

PR #843, opened 2018-07-27
Add some HID commands

Hexagon12:
Adds `IsSixAxisSensorAtRest` and `MergeSingleJoyAsDualJoy`, which are used by Kirby Star Allies and Has Been Heroes.

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -326,7 +326,7 @@ public:
             {79, &Hid::SetGyroscopeZeroDriftMode, "SetGyroscopeZeroDriftMode"},
             {80, nullptr, "GetGyroscopeZeroDriftMode"},
             {81, nullptr, "ResetGyroscopeZeroDriftMode"},
-            {82, nullptr, "IsSixAxisSensorAtRest"},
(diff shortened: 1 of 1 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -455,6 +455,12 @@ class Hid final : public ServiceFramework<Hid> {
         LOG_WARNING(Service_HID, "(STUBBED) called");
     }
 
+    void IsSixAxisSensorAtRest(Kernel::HLERequestContext& ctx) {
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(RESULT_SUCCESS);
```
lioncash: This function probably pushes more than just a result value, and shouldn't be stubbed like this.

lioncash: to be more specific, this should likely be:

```cpp
rb.Push(RESULT_SUCCESS);
// could also be false, depending on what you want to stub it as.
rb.Push(true);
```

bunnei: +1

Hexagon12: Done, not sure why this comment isn't closing off.

PR #848, opened 2018-07-28
Implement R16S & R16UI & R16I RenderTargetFormats & PixelFormats and more (R16_UNORM needed by Fate Extella)

greggameplayer:
- Implement R16_UNORM & R16_SNORM & R16_UINT & R16_SINT RenderTargetFormats

- Implement R16S & R16UI & R16I PixelFormat

- Add all the current implemented RenderTargetFormat in RenderTargetBytesPerPixel

- Create DepthFormatBytesPerPixel function and add all the current DepthFormat inside.

R16_UNORM is needed by Fate Extella


greggameplayer:
Yes but we're gonna need these anyway, so it's better to do it now.

lioncash:
You're missing the point of what @Subv saying. If you have no test case, you have no idea if the code is doing what you expect of it.

greggameplayer:
@lioncash Yes but it's based on opengl info so normally it's good

src/video_core/gpu.cpp
```diff
@@ -33,17 +33,51 @@ u32 RenderTargetBytesPerPixel(RenderTargetFormat format) {
...
+        return 2;
+    case RenderTargetFormat::R8_UNORM:
+        return 1;
     default:
         UNIMPLEMENTED_MSG("Unimplemented render target format {}", static_cast<u32>(format));
     }
 }
 
+u32 DepthFormatBytesPerPixel(DepthFormat format) {
+
```
lioncash: Unnecessary newline

src/video_core/gpu.cpp
```diff
@@ -33,17 +33,51 @@ u32 RenderTargetBytesPerPixel(RenderTargetFormat format) {
...
+    case RenderTargetFormat::R16_FLOAT:
+        return 2;
+    case RenderTargetFormat::R8_UNORM:
+        return 1;
     default:
         UNIMPLEMENTED_MSG("Unimplemented render target format {}", static_cast<u32>(format));
     }
 }
 
+u32 DepthFormatBytesPerPixel(DepthFormat format) {
```
lioncash: This is missing a prototype in the header along with a doc comment.

lioncash:
It's kind of incredible that you're still managing to miss the point that the code is set up to find those exact issues. When code isn't asserting on the data being processed it's much more difficult to find.

It would also be great if you'd stop arguing for fixing potential issues later rather than sooner. That's a terrible way to debug issues. "It may not do what you expect", "It's probably OK, because OpenGL", "The point of the code is to help establish test cases", "We can do that later". Do you *really* not see anything at all wrong with that line of thinking?

greggameplayer:
@lioncash if the texture bad render we can see what this is in RenderDoc

lioncash:
This also adds an extra step in the process for little gain at that point. It's also assuming we'd only want to know what textures a game uses if it renders incorrectly. It's good to establish which games use a certain texture format even if they display properly because it allows knowing which games to specifically test if the relevant code is ever changed in the future without having to comb through games, or wait until a user reports that it's broken. This makes doing that much more difficult than it otherwise needs to be.

greggameplayer:
Ok ,so i wait the bunnei thought about this PR

khang06:
"we usually only implement things as we come across games that use them"

R16_UNORM used in star allies

DarkLordZach:
Please rebase

PR #849, opened 2018-07-28
XCI and Encrypted NCA Support

DarkLordZach:
First of all, a massive thanks to @SciresM and his amazing [hactool](https://github.com/sciresm/hactool). Without it, this probably would not have been possible.

**Technical Overview**
- Add [mbedtls](https://tls.mbed.org/) dependency for crypto operations. (Crypto++ from Citra couldn't be used due to a lack of AES-XTS support).
- `KeyManager` to contain/load keys. When needed, keys are loaded from the same place hactool loads them from. 
- Keys are stored in `%YUZU_DIR%/keys/prod.keys` if using production, `%YUZU_DIR%/keys/dev.keys` if using dev, and `%YUZU_DIR%/keys/title.keys` for titlekeys. (There is a fallback to the hactool config dir if the yuzu versions cannot be found)
- Dev mode (using dev keys) can be enabled by finding the `use_dev_mode` key in qt/sdl2 config (depending on frontend).
- Encryption implemented using VFS. `EncryptionLayer` is an incomplete class that overrides everything except `Read` in `VfsFile` to cut down on duplicate code in subclasses.
- `CTREncryptionLayer` for CTR-mode encryption scheme.
- `Key128`, `Key256`, `SHA256Hash` for convenience (typedefs of `std::array<u8, *>`)
- `AESCipher<KeyType>` added as a C++ wrapper around mbedtls.

**Feature Overview**
- Supports the running of encrypted NCA files with the correct NCA header key and key area key. **NCAs using titlekey crypto are not supported with this PR.** (decided to make that separate)
- By extension, supports the execution of XCI files. 
- Adds support in the game list for the recognition and display of XCI/eNCA files.

Tested on: 1-2-Switch, Cave Story+, DOOM, Hyrule Warriors, I Am Setsuna, Minecraft Story Mode, Puyo Puyo Tetrix, Rocket League, Snipperclips Plus, Sonic Forces, Binding of Isaac, and Tiny Barbarian DX.

PR #850, opened 2018-07-28
Add Icons and Metadata Support

DarkLordZach:
Adds:
- Icons on the gamelist. Size is configurable in settings.
Icons are loaded from (in priority order):
    1. The game file (in the case of NRO or XCI (stay tuned #849) formats)
    2. The game directory (for deconstructed rom dir games. Looks for any png/jpg file, prefers ones named icon_AmericanEnglish.dat/.jpg)
    3. Scans all game dirs for Control-type NCAs. These store icon/metadata for game ncas on the switch. If one is found, all games matching the title id without another icon will use the one from the control nca.

- Metadata (such as title id and name)
Loaded same as icons, except looks for a file called `control.nacp` in the directory.

- The two rows that show in the game list are configurable (from Filename, filetype, title name, or title id). To hide the second row, make its display setting the same as the first.

- If you use the Control-type NCA option described above, you'll notice that they show up in the game list as type 'unknown' because they have a valid extention. An option to hide 'unknown' files has been added to settings. It is default-disabled.

**Quick Note:** Just extracting a game to directory is not enough, you must also find the matching NCA of type Control (different from game) and extract the contents to the game dir.

**EDIT**: Add XCI support. As XCI files come with the control NCA packed in, there is no need for extra files with them (all you need is the keys).

### DIFF
--- a/src/core/file_sys/content_archive.cpp
+++ b/src/core/file_sys/content_archive.cpp
@@ -140,6 +140,10 @@ VirtualFile NCA::Decrypt(NCASectionHeader header, VirtualFile in, u64 starting_o
 }
 
 NCA::NCA(VirtualFile file_) : file(std::move(file_)) {
+    if (file == nullptr) {
+        status = Loader::ResultStatus::ErrorInvalidFormat;
+        return;
+    }
     if (sizeof(NCAHeader) != file->ReadObject(&header))
         LOG_ERROR(Loader, "File reader errored out during header read.");
 
--- a/src/core/file_sys/control_metadata.h
+++ b/src/core/file_sys/control_metadata.h
@@ -62,7 +62,7 @@ enum class Language : u8 {
     Chinese = 14,
 };
 
-static std::array<std::string, 15> LANGUAGE_NAMES = {
+static constexpr std::array<const char*, 15> LANGUAGE_NAMES = {
     "AmericanEnglish", "BritishEnglish", "Japanese",
     "French",          "German",         "LatinAmericanSpanish",
     "Spanish",         "Italian",        "Dutch",
--- a/src/core/file_sys/control_metadata.h
+++ b/src/core/file_sys/control_metadata.h
@@ -62,6 +62,13 @@ enum class Language : u8 {
     Chinese = 14,
 };
 
+static std::array<std::string, 15> LANGUAGE_NAMES = {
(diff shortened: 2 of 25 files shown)


src/core/file_sys/content_archive.cpp
```diff
@@ -153,6 +162,10 @@ VirtualDir NCA::GetExeFS() const {
     return exefs;
 }
 
+VirtualDir NCA::GetSection(u8 section_id) {
```
lioncash: const member function

src/core/file_sys/control_metadata.h
```diff
@@ -62,6 +62,13 @@ enum class Language : u8 {
     Chinese = 14,
 };
 
+static std::array<std::string, 15> LANGUAGE_NAMES = {
```
lioncash: `static constexpr`. This can likely also just be `const char*`

src/yuzu/configuration/configure_gamelist.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "core/settings.h"
+#include "ui_configure_gamelist.h"
+#include "ui_settings.h"
+#include "yuzu/configuration/configure_gamelist.h"
+
+ConfigureGameList::ConfigureGameList(QWidget* parent)
+    : QWidget(parent), ui(new Ui::ConfigureGameList) {
+    ui->setupUi(this);
+
+    static std::vector<std::pair<u32, std::string>> default_icon_sizes{
```
lioncash: `static const`

src/yuzu/configuration/configure_gamelist.cpp
```diff
@@ -0,0 +1,62 @@
...
+    };
+
+    for (const auto& size : default_icon_sizes) {
+        ui->icon_size_combobox->addItem(QString::fromStdString(size.second + " (" +
+                                                               std::to_string(size.first) + "x" +
+                                                               std::to_string(size.first) + ")"),
+                                        size.first);
+    }
+
+    static std::vector<std::string> row_text_names{
```
lioncash: `static const`

EmulationChannel:
@DarkLordZach Ah Possibilities of showing the XCI GAMES icons equal to ah from the GAME SWITCH box?

PR #851, opened 2018-07-28
Added ability to change username & language code in the settings ui. Added IProfile::Get and SET::GetLanguageCode for libnx tests

ogniK5377:
Currently, the additional languages added to the language list need to be translated. By default the username will be "yuzu" and the language will be english
2018-07-28_16-24-8937c5af-9140-4272-8cad-739c989e56aa-d48d6310

### DIFF
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -10,6 +10,7 @@
 #include "core/hle/service/acc/acc_su.h"
 #include "core/hle/service/acc/acc_u0.h"
 #include "core/hle/service/acc/acc_u1.h"
+#include "core/settings.h"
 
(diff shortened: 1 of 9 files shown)


src/core/hle/service/acc/acc.cpp
```diff
@@ -46,14 +48,36 @@ class IProfile final : public ServiceFramework<IProfile> {
     }
 
 private:
+    void Get(Kernel::HLERequestContext& ctx) {
```
Subv: I'm rather sure this function returns not only the ProfileBase but also something else.

ogniK5377: Used LibNX as a reference for this function as it's implemented there

PR #852, opened 2018-07-28
Port #3758 from Citra: "Add missing std::string import in text_formatter"

FearlessTobi:
See https://github.com/citra-emu/citra/pull/3758 for more details.


bunnei:
I'll just fix the commit message on merge, FYI (but definitely please correct for future PRs)

PR #855, opened 2018-07-28
Audio output backend based on cubeb

bunnei:
Adds an audio output backend based on [cubeb](https://github.com/kinetiknz/cubeb). Summary of the changes:
* Adds cubeb as a dependency, much like Citra this can be disabled via a CMake flag
* Adds `Sink` interface, this is the main interface for an audio output, much like Citra
* Adds `SinkStream` interface, this is the a sub-interface of `Sink` for outputting a single stream of audio via the Sink interface. The idea is that we will support multiple output streams, and the audio library will do the mixing for us.
* Implements `Sink` and `SinkStream` with cubeb, and integrates this into `AudioOut`.
* Integrates this all with `audout:u` service for audio output. The idea though is that we should be able to use this design with `audren`, etc. as well in the near future.
* As a side-note, I moved the `AudioOut` creation to `audout:u`. This is because right now we need to initialize cubeb on the same thread that we create streams on. This will need to be improved if we want `audren`, etc. to use this.

Games that use the `audout:u` service should now play audio. It works really well if the games are running at or near 60fps. Games tested:
* One Piece
* Snipper Clips
* Sonic Mania
* Minecraft Story Mode

I also tested both SDL and Qt frontends.

Edit: No audio tab in settings yet. I'll add that either as a followon change or separate PR.

### DIFF
--- a/src/audio_core/CMakeLists.txt
+++ b/src/audio_core/CMakeLists.txt
@@ -2,6 +2,8 @@ add_library(audio_core STATIC
     audio_out.cpp
     audio_out.h
     buffer.h
+    cubeb_sink.cpp
+    cubeb_sink.h
     null_sink.h
     stream.cpp
     stream.h
@@ -14,3 +16,8 @@ add_library(audio_core STATIC
 create_target_directory_groups(audio_core)
 
 target_link_libraries(audio_core PUBLIC common core)
+
+if(ENABLE_CUBEB)
+    target_link_libraries(audio_core PRIVATE cubeb)
+    target_compile_definitions(audio_core PRIVATE -DHAVE_CUBEB=1)
+endif()
--- a/src/audio_core/CMakeLists.txt
+++ b/src/audio_core/CMakeLists.txt
@@ -2,8 +2,13 @@ add_library(audio_core STATIC
     audio_out.cpp
     audio_out.h
     buffer.h
+    null_sink.h
     stream.cpp
     stream.h
+    sink.h
+    sink_details.cpp
+    sink_details.h
+    sink_stream.h
 )
 
 create_target_directory_groups(audio_core)
--- a/src/audio_core/buffer.h
+++ b/src/audio_core/buffer.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <memory>
 #include <vector>
 
 #include "common/common_types.h"
@@ -33,5 +34,7 @@ private:
     Tag tag;
     std::vector<u8> data;
 };
+
+using BufferPtr = std::shared_ptr<Buffer>;
 
 } // namespace AudioCore
--- a/src/audio_core/cubeb_sink.cpp
+++ b/src/audio_core/cubeb_sink.cpp
@@ -1,0 +1,190 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <cstring>
+
+#include "audio_core/cubeb_sink.h"
+#include "audio_core/stream.h"
+#include "common/logging/log.h"
+
+namespace AudioCore {
+
+class SinkStreamImpl final : public SinkStream {
+public:
+    SinkStreamImpl(cubeb* ctx, cubeb_devid output_device) : ctx{ctx} {
+        cubeb_stream_params params;
+        params.rate = 48000;
+        params.channels = GetNumChannels();
+        params.format = CUBEB_SAMPLE_S16NE;
+        params.layout = CUBEB_LAYOUT_STEREO;
+
+        u32 minimum_latency = 0;
+        if (cubeb_get_min_latency(ctx, &params, &minimum_latency) != CUBEB_OK) {
+            LOG_CRITICAL(Audio_Sink, "Error getting minimum latency");
+        }
+
+        if (cubeb_stream_init(ctx, &stream_backend, "yuzu Audio Output", nullptr, nullptr,
+                              output_device, &params, std::max(512u, minimum_latency),
+                              &SinkStreamImpl::DataCallback, &SinkStreamImpl::StateCallback,
+                              this) != CUBEB_OK) {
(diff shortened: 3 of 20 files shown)


src/audio_core/CMakeLists.txt
```diff
@@ -2,10 +2,22 @@ add_library(audio_core STATIC
...
+    sink_stream.h
 )
 
 create_target_directory_groups(audio_core)
 
 target_link_libraries(audio_core PUBLIC common core)
+
+if(ENABLE_CUBEB)
+    target_link_libraries(audio_core PRIVATE cubeb)
+    add_definitions(-DHAVE_CUBEB=1)
```
lioncash: It would likely be best to use `target_add_compile_options` here and make the definition public, so it only applies to targets that link in audio_core

src/audio_core/buffer.h
```diff
@@ -34,4 +34,6 @@ class Buffer {
     std::vector<u8> data;
 };
 
+using BufferPtr = std::shared_ptr<Buffer>;
```
lioncash: `<memory>` should be included.

src/audio_core/cubeb_sink.cpp
```diff
@@ -0,0 +1,185 @@
...
+        if (cubeb_get_min_latency(ctx, &params, &minimum_latency) != CUBEB_OK) {
+            LOG_CRITICAL(Audio_Sink, "Error getting minimum latency");
+        }
+
+        if (cubeb_stream_init(ctx, &stream_backend, "yuzu Audio Output", nullptr, nullptr,
+                              output_device, &params, std::max(512u, minimum_latency),
+                              &SinkStreamImpl::DataCallback, &SinkStreamImpl::StateCallback,
+                              this) != CUBEB_OK) {
+            LOG_CRITICAL(Audio_Sink, "Error initializing cubeb stream");
+            return;
```
lioncash: Shouldn't there be a way to check if the initialization failed?

src/audio_core/null_sink.h
```diff
@@ -0,0 +1,27 @@
...
+
+#pragma once
+
+#include "audio_core/sink.h"
+
+namespace AudioCore {
+
+class NullSink final : public Sink {
+public:
+    NullSink(std::string) {}
```
lioncash: This can be defaulted as well

src/audio_core/cubeb_sink.cpp
```diff
@@ -0,0 +1,185 @@
...
+    const size_t frames_to_write{
+        std::min(impl->queue.size() / impl->GetNumChannels(), static_cast<size_t>(num_frames))};
+
+    memcpy(buffer, impl->queue.data(), frames_to_write * sizeof(s16) * impl->GetNumChannels());
+    impl->queue.erase(impl->queue.begin(),
+                      impl->queue.begin() + frames_to_write * impl->GetNumChannels());
+
+    if (frames_to_write < num_frames) {
+        // Fill the rest of the frames with silence
+        memset(buffer + frames_to_write * sizeof(s16) * impl->GetNumChannels(), 0,
```
lioncash: `<cstring>` should be included

src/audio_core/cubeb_sink.cpp
```diff
@@ -0,0 +1,185 @@
...
+                                  void* output_buffer, long num_frames) {
+    SinkStreamImpl* impl = static_cast<SinkStreamImpl*>(user_data);
+    u8* buffer = reinterpret_cast<u8*>(output_buffer);
+
+    if (!impl) {
+        return {};
+    }
+
+    const size_t frames_to_write{
+        std::min(impl->queue.size() / impl->GetNumChannels(), static_cast<size_t>(num_frames))};
```
lioncash: `<algorithm>` should be included.

src/audio_core/cubeb_sink.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+        queue.reserve(queue.size() + sample_count * GetNumChannels());
+
+        if (num_channels == 2) {
+            // Copy as-is
+            std::copy(samples, samples + sample_count * GetNumChannels(),
+                      std::back_inserter(queue));
+        } else if (num_channels == 6) {
+            // Downsample 6 channels to 2
+            for (size_t i = 0; i < sample_count * num_channels; i += num_channels) {
```
lioncash: Not sure if it matters, but you can do:

```cpp
const size_t sample_count_copy_size = sample_count * num_channels * 2;

queue.reserve(sample_count_copy_size);
```

to reserve the block of memory ahead of time, so reallocation isn't necessary within the loop for long copies.

src/audio_core/cubeb_sink.cpp
bunnei: I don't think it really should fail, unless something is really wrong, so I'm not sure the callers should have to handle that. The logging is mostly here just so we can catch/correct if any problems come up.

src/audio_core/null_sink.h
bunnei: What do you mean here? Setting this to default gives me a compile error...

src/audio_core/CMakeLists.txt
bunnei: Can you provide an example of what this would look like? I'm not very proficient with CMake, beyond the basics. Thanks!

src/audio_core/cubeb_sink.cpp
bunnei: addressed

src/audio_core/cubeb_sink.cpp
bunnei: addressed

src/audio_core/buffer.h
bunnei: addressed

src/audio_core/CMakeLists.txt
lioncash:
```cmake
target_compile_definitions(audio_core PRIVATE -DHAVE_CUBEB=1)
```

src/audio_core/null_sink.h
lioncash: What's the compile error?

PR #869, opened 2018-07-31
Corrected a few error cases detected by asan/ubsan

Subv:
Thanks @DarkLordZach for running it on yuzu!


src/video_core/macro_interpreter.h
```diff
@@ -91,6 +91,10 @@ class MacroInterpreter final {
         u32 GetBitfieldMask() const {
             return (1 << bf_size) - 1;
         }
+
+        s32 GetBranchTarget() const {
+            return immediate * sizeof(u32);
```
lioncash: The result should be explicitly casted, since this is technically truncating a 64-bit value implicitly.

PR #870, opened 2018-07-31
arm_dynarmic: Correct initializer list order

lioncash:
Amends the initializer list to be in the same order that each variable would be initialized in.


MerryMage:
Looking at this, `MakeJit()` depends on `exclusive_monitor`, `cb` and `core_index`, so perhaps `jit` should be reordered to be last (in the class header as well).

lioncash:
@MerryMage done

PR #871, opened 2018-07-31
audio_core: Add configuration settings.

bunnei:
Adds audio configuration settings, and adds an "Audio" tab for setting them from the Qt frontend. This is similar to Citra, with a couple of differences:
* There is no time stretching, so we do not have this setting
* Sink selection and output device are disabled when the emulator is running, as these are not as easy to implement with Switch (we have more audio streams/state to keep track of)

Obligatory UI screenshot:
image

### DIFF
--- a/src/audio_core/stream.cpp
+++ b/src/audio_core/stream.cpp
@@ -2,14 +2,17 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include "common/assert.h"
-#include "common/logging/log.h"
-#include "core/core_timing.h"
-#include "core/core_timing_util.h"
+#include <algorithm>
+#include <cmath>
 
 #include "audio_core/sink.h"
 #include "audio_core/sink_details.h"
 #include "audio_core/stream.h"
+#include "common/assert.h"
+#include "common/logging/log.h"
+#include "core/core_timing.h"
+#include "core/core_timing_util.h"
+#include "core/settings.h"
 
 namespace AudioCore {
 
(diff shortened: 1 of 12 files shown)


src/yuzu/configuration/configure_audio.cpp
```diff
@@ -0,0 +1,90 @@
...
+    this->setConfiguration();
+    connect(ui->output_sink_combo_box,
+            static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this,
+            &ConfigureAudio::updateAudioDevices);
+
+    ui->output_sink_combo_box->setEnabled(!Core::System::GetInstance().IsPoweredOn());
+    ui->audio_device_combo_box->setEnabled(!Core::System::GetInstance().IsPoweredOn());
+}
+
+ConfigureAudio::~ConfigureAudio() {}
```
lioncash: This can be defaulted.

src/core/settings.h
```diff
@@ -137,6 +137,11 @@ struct Values {
 
     std::string log_filter;
 
+    // Audio
+    std::string sink_id;
+    std::string audio_device_id;
+    float volume;
```
lioncash: Should there be some form of sanitizing the volume value? (e.g. if someone sets it to a silly value).

src/audio_core/stream.cpp
```diff
@@ -56,6 +58,23 @@ s64 Stream::GetBufferReleaseCycles(const Buffer& buffer) const {
...
+static std::vector<s16> GetVolumeAdjustedSamples(const std::vector<u8>& data) {
+    std::vector<s16> samples(data.size() / sizeof(s16));
+    std::memcpy(samples.data(), data.data(), data.size());
+
+    if (Settings::values.volume == 1.0f) {
+        return samples;
+    }
+
+    // Implementation of a volume slider with a dynamic range of 60 dB
+    const float volume_scale_factor{std::exp(6.90775f * Settings::values.volume) * 0.001f};
```
MerryMage: `std::expf`

src/core/settings.h
bunnei: Added a clamp(0.0, 1.0) to audio core

PR #883, opened 2018-08-01
audren_u: Remove incorrect wave_buffer updates.

bunnei:
The current code to update here is inaccurate, and it's breaking games. We'll be improving this soon, but for now, I think it's better to just not update. Also, some minor cleanups. Fixes Disgaea 5 (goes in game now), and likely some other games that are stuck on audren.

image

### DIFF
--- a/src/core/hle/service/audio/audren_u.cpp
+++ b/src/core/hle/service/audio/audren_u.cpp
@@ -72,8 +72,8 @@ private:
                     buf.data() + sizeof(UpdateDataHeader) + config.behavior_size,
                     memory_pool_count * sizeof(MemoryPoolInfo));
 
-        std::vector<VoiceInfo> voice_info(worker_params.voice_count);
-        std::memcpy(voice_info.data(),
+        std::vector<VoiceInfo> voice_infos(worker_params.voice_count);
+        std::memcpy(voice_infos.data(),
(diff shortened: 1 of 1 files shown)


src/core/hle/service/audio/audren_u.cpp
```diff
@@ -94,18 +94,19 @@ class IAudioRenderer final : public ServiceFramework<IAudioRenderer> {
...
-                for (u32 buff_idx = 0; buff_idx < voice_info[i].wave_buffer_count; buff_idx++) {
-                    voice_status_list[i].played_sample_count +=
-                        (voice_info[i].wave_buffer[buff_idx].end_sample_offset -
-                         voice_info[i].wave_buffer[buff_idx].start_sample_offset) /
-                        2;
-                    voice_status_list[i].wave_buffer_consumed++;
-                }
+            } else if (voice_info.play_state == PlayState::Started) {
+                // TODO(bunnei): Properly update wave_buffer state here once playback has started.
+                // Doing this inaccurately breaks games.
```
lioncash: "breaks games (such as Disgaea 5)" would serve to document an example game that gets broken in the meantime

bunnei:
FYI this change breaks Puyo Puyo Tetris from going in game.

bunnei:
Superceded by #925

PR #886, opened 2018-08-01
Implement RGB32F PixelFormat

greggameplayer:
Used by games like Go Vacation


lioncash:
The game should be put in the commit message as well, so it's documented in the log as an example game that uses it.

bunnei:
I'll just add that the commit message on merge

PR #892, opened 2018-08-02
video_core: Make global EmuWindow instance part of the base renderer …

lioncash:
Makes the global a member of the RendererBase class. We also change this to be a reference. Passing any form of null pointer to these functions is incorrect entirely, especially given the code itself assumes that the pointer would always be in a valid state.

This also makes it easier to follow the lifecycle of instances being used, as we explicitly interact the renderer with the rasterizer, rather than it just operating on a global pointer.


bunnei:
Another 👌

PR #893, opened 2018-08-02
service: Add the psc services

lioncash:
Adds the basic skeleton for the psc services based off the information provided by Switch Brew.


lioncash:
New commits simply add a `"common/logging/log.h"` include that I almost forgot about in the cpp file for the psc services, everything else is the same.

PR #895, opened 2018-08-02
 sink_details: std::move std::function instances 

lioncash:
Given std::function is allowed to potentially allocate, these should be std::move'd to prevent potential reallocation (should that ever happen).


lioncash:
Failure is unrelated

PR #898, opened 2018-08-02
service: Add migration services

lioncash:
Adds the basic skeleton for the mig:usr service based off information provided by Switch Brew.


lioncash:
Failure seems unrelated

PR #900, opened 2018-08-02
math_util: Always initialize members of Rectangle

lioncash:
Prevents potentially using the members uninitialized.


lioncash:
Failure seems unrelated

PR #901, opened 2018-08-02
gl_shader_manager: Take ShaderSetup instances by const reference in UseProgrammableVertexShader() and UseProgrammableFragmentShader()

lioncash:
Avoids performing unnecessary copies of 65560 byte sized ShaderSetup instances, considering it's only used as part of lookup and not modified.

Given the parameters were already const, it's likely taking these parameters by reference was intended but the ampersand was forgotten.


bunnei:
Nice find 👌

PR #922, opened 2018-08-04
 CMakeLists: Change MSVC14 variable to MSVC_VERSION

lioncash:
Use of the MSVC14 variable is discouraged in the CMake documentation


MerryMage:
`MSVC_VERSION GREATER_EQUAL 1910 AND MSVC_VERSION LESS 1920`

lioncash:
Done

PR #925, opened 2018-08-04
Implement audren audio output

bunnei:
This implements audio playback for audren_u, e.g. `AudioRenderer`. This should add audio output for most games. Similar to our audout_u implementation, there is no time stretching, etc., so this relies on games running at or around 60fps to sound correct.

This leverages the same AudioCore backend, the `AudioOut` class, that we use for `audout_u`. Similar to Ryujinx's impl, we use a single 48000Hz 2-channel stream. Voices are mixed into this stream (regardless of sample rate) with 1-channel buffers being upsampled to 2-channel. I chose to go this route because creating lots of smaller streams for voices (configured with their correct sample rate and channel count) did not work out too well - some games ended up created way to many streams, which led to cubeb being unstable. Initial support is for PCM16 and ADPCM.

That being said, this seems to work well enough. In the future, we might want to create a new stream for each unique sample rate, and mix into the right one - but this can be done later.

Tested with: Cave Story+, the Binding of Isaac, Puyo Puyo Tetris, Disgaea 5, Super Mario Odyssey, and Stardew Valley

Thanks @ogniK5377 for most of the RE work for this, as well as Ryujinx, whose impl. inspired ours.

### DIFF
--- a/src/audio_core/CMakeLists.txt
+++ b/src/audio_core/CMakeLists.txt
@@ -1,6 +1,8 @@
 add_library(audio_core STATIC
     audio_out.cpp
     audio_out.h
+    audio_renderer.cpp
+    audio_renderer.h
     buffer.h
     cubeb_sink.cpp
     cubeb_sink.h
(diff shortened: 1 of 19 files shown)


EmulationChannel:
@bunnei Probabilities Splatoon 2 Audios?

PR #934, opened 2018-08-06
 core_timing: Make GetGlobalTimeUs() return std::chrono::microseconds

lioncash:
Enforces the time unit being returned and also allows using the standard time utilities to manipulate it.

### DIFF
--- a/src/core/perf_stats.cpp
+++ b/src/core/perf_stats.cpp
@@ -76,7 +76,7 @@ double PerfStats::GetLastFrameTimeScale() {
 void FrameLimiter::DoFrameLimiting(microseconds current_system_time_us) {
     // Max lag caused by slow frames. Can be adjusted to compensate for too many slow frames. Higher
     // values increase the time needed to recover and limit framerate again after spikes.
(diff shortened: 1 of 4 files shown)


src/core/perf_stats.cpp
```diff
@@ -74,18 +73,18 @@ double PerfStats::GetLastFrameTimeScale() {
     return duration_cast<DoubleSecs>(previous_frame_length).count() / FRAME_LENGTH;
 }
 
-void FrameLimiter::DoFrameLimiting(u64 current_system_time_us) {
+void FrameLimiter::DoFrameLimiting(microseconds current_system_time_us) {
     // Max lag caused by slow frames. Can be adjusted to compensate for too many slow frames. Higher
     // values increase the time needed to recover and limit framerate again after spikes.
-    constexpr microseconds MAX_LAG_TIME_US = 25ms;
+    constexpr microseconds MAX_LAG_TIME_US = 25us;
```
MerryMage: `25ms` looks more correct here?

PR #935, opened 2018-08-06
Port fixup of "gdbstub: Get rid of a few signed/unsigned comparisons" from citra

zhaowenlan1779:
According to @wwylele comment, there's no need to cast to u32

See citra-emu/citra#4031, #723 for details.

### DIFF
--- a/src/core/gdbstub/gdbstub.cpp
+++ b/src/core/gdbstub/gdbstub.cpp
@@ -743,7 +743,7 @@ static bool IsDataAvailable() {
     fd_set fd_socket;
 
     FD_ZERO(&fd_socket);
-    FD_SET(static_cast<u32>(gdbserver_socket), &fd_socket);
+    FD_SET(gdbserver_socket, &fd_socket);
 
     struct timeval t;
     t.tv_sec = 0;


lioncash:
This does apply on Windows, however, which uses a u32 value internally, not an int value.

zhaowenlan1779:
If that's the case, I may have to port that back to citra again 🤔

wwylele:
The real problem is the inconsistent definition of a socket type between [linux](http://man7.org/linux/man-pages/man2/socket.2.html)(`int`) and [windows](https://docs.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-socket)(`SOCKET`). If you are going to fix warnings, the correct way is to define `gdbserver_socket` (and any other socket) with a unified type, which can be wrapped in `#ifdef` + `using`

zhaowenlan1779:
@wwylele Yes, I realized that after reading mingw source code for that - it's uint, and for linux it's often int.
That's quite annoying, and I feel it's a bit foolish to introduce some platform-dependent code only to fix warnings, but probably that's the best way for now.

PR #937, opened 2018-08-06
Service/Audio: audout_a.cpp: remove pragma once

mailwl:


lioncash:
The clang-format directives are supposed to remain there, the pragma however is fine to be removed.

mailwl:
done

PR #938, opened 2018-08-06
GDBStub works with both Unicorn and Dynarmic now

Hedges:


MerryMage:
It would likely be better (performance-wise, but also design-wise) to add a function like the following to `System`:

```c++
void InvalidateCpuCaches() {
    for (auto& cpu : cpu_cores) {
        cpu->ArmInterface().ClearInstructionCache();
    }
}
```

And call that when the instruction cache needs to be invalidated, rather than having a check in `arm_dynarmic.cpp`.

Hedges:
Agreed, looks much cleaner this way, thanks.

Hedges:
Argh, my git went haywire again, will have to redo the PR!

PR #941, opened 2018-08-06
GDBStub works with both Unicorn and Dynarmic now

Hedges:


src/core/gdbstub/gdbstub.cpp
```diff
@@ -173,6 +173,7 @@ struct Breakpoint {
     bool active;
     VAddr addr;
     u64 len;
+    u8 inst[4];
```
lioncash: `std::array`

src/core/gdbstub/gdbstub.cpp
```diff
@@ -988,6 +994,10 @@ static bool CommitBreakpoint(BreakpointType type, VAddr addr, u64 len) {
     breakpoint.active = true;
     breakpoint.addr = addr;
     breakpoint.len = len;
+    Memory::ReadBlock(addr, breakpoint.inst, 4);
+    static const u8 btrap[] = {0xd4, 0x20, 0x7d, 0x00};
```
lioncash:
```cpp
Memory::ReadBlock(addr, breakpoint.inst.data(), breakpoint.inst.size());
static constexpr std::array<u8, 4> btrap{{0xd4, 0x20, 0x7d, 0x0}};
Memory::WriteBlock(addr, btrap.data(), btrap.size());
```

PR #948, opened 2018-08-06
CMakeLists: Make mbedtls and cubeb not install headers and libraries

hcorion:
This PR makes it so that `make install` does not install the mbedtls or cubeb headers and static libraries.

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -36,7 +36,7 @@ add_subdirectory(lz4/contrib/cmake_unofficial)
 target_include_directories(lz4_static INTERFACE ./lz4/lib)
 
 # mbedtls
-add_subdirectory(mbedtls)
+add_subdirectory(mbedtls EXCLUDE_FROM_ALL)
 target_include_directories(mbedtls PUBLIC ./mbedtls/include)
 
 # MicroProfile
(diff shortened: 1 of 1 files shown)


externals/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ add_subdirectory(lz4/contrib/cmake_unofficial)
 target_include_directories(lz4_static INTERFACE ./lz4/lib)
 
 # mbedtls
+option(INSTALL_MBEDTLS_HEADERS OFF)
```
lioncash: Would it not be sufficient to just do:

```cmake
add_subdirectory(mbedtls EXCLUDE_FROM_ALL)
```

to achieve the same behavior without needing to modify the submodule?

hcorion: Of course, that makes much more sense, I completely forgot about that, I think I'll do that for cubeb too. I'll leave the changes to the submodule in the PR, unless @DarkLordZach wants to undo the change.

lioncash: Sounds good. Ideally, the change for the submodule should be undone. We shouldn't even be using any source repository that isn't the main mbedtls one, because any change that needs to be made should be upstreamed, and all the local changes make it more annoying to update between releases of mbedtls, but for whatever reason that's not what was done.

DarkLordZach: @lioncash The main mbedtls repo is licensed under Apache, and they provide packaged versions on their website under GPL. My mbedtls repo is an upload of that for ease of use with submodules.

lioncash: that's... kind of bizarre they do that, but I guess separate repo makes sense now (though we still shouldn't be applying local changes to it and should update to [v2.12.0](https://github.com/ARMmbed/mbedtls/releases), given the amount of bugfixes it has).

PR #958, opened 2018-08-07
nvdrv: Get rid of global std::weak_ptr

lioncash:
Rather than use global state, we can simply pass the instance into the NVFlinger instance directly.


bunnei:
Needs a rebase, FYI

PR #973, opened 2018-08-08
acc: Stub GetUserCount.

bunnei:


lioncash:
Commit message should probably say what it was stubbed for, for historical lookup

PR #974, opened 2018-08-08
acc: Add missing function table entries for GetUserCount

lioncash:
Given this is stubbed within the common module in 5ac7b84, it should be added to the other relevant tables as well.


lioncash:
Failure is related to downloading qt. The build itself is actually fine

PR #990, opened 2018-08-08
 fsp_srv: Emplace entries first when building index instead of emplacing last

lioncash:
The current way were doing it would require copying a 768 character buffer (part of the Entry struct) to the new element in the vector. Given it's a plain array, std::move won't eliminate that.

Instead, we can emplace an instance directly into the destination buffer and then fill it out, avoiding the need to perform any unnecessary copies.

Given this is done in a loop, we can request the destination to allocate all of the necessary memory ahead of time, avoiding the need to potentially keep reallocating over and over on every few insertions into the vector.

This makes it slightly better when it comes to building the filesystem index, and we keep the sanitizing logic for the filename within the same struct its in, as opposed to being off in some other separate module.

### DIFF
--- a/src/core/file_sys/directory.h
+++ b/src/core/file_sys/directory.h
@@ -4,8 +4,9 @@
 
 #pragma once
 
-#include <array>
 #include <cstddef>
+#include <iterator>
+#include <string_view>
 #include "common/common_funcs.h"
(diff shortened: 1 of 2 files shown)


lioncash:
Build succeeded, it's just a server response error

Issue #939, opened 2018-08-06
yuzu nightly crashes on linux no matter what the game


ReaperOfSouls1909:
https://gist.github.com/ReaperOfSouls1909/f36ae4e63473ea0d962c61ee830898c7 backtrace

7oxicshadow:
I have built against the latest git but I am getting a different crash when loading a game:

BT:

```
Thread 22 "EmuThread" received signal SIGTRAP, Trace/breakpoint trap.
[Switching to Thread 0x7fffa6ffd700 (LWP 19822)]
assert_noinline_call<Kernel::Break(u64, u64, u64)::<lambda()> >(const Kernel::<lambda()> &) (fn=...)
    at /mnt/games/native_emulators/source_files/yuzu/src/./common/assert.h:27
27	    exit(1); // Keeps GCC's mouth shut about this actually returning
[...]

PR #999, opened 2018-08-09
gl_rasterizer_cache: Avoid iterator invalidation issues within InvalidateRegion()

lioncash:
A range-based for loop can't be used when the container being iterated is also being erased from.


Hexagon12:
Fixes #939.

PR #1002, opened 2018-08-10
textures: Refactor out for Texture/Depth FormatFromPixelFormat.

bunnei:
- As a plus, this allows us to add render target formats that do not have a corresponding texture format.


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -183,6 +183,21 @@ MathUtil::Rectangle<u32> SurfaceParams::GetRect() const {
     return {0, actual_height, width, 0};
 }
 
+/// Returns true if the specified PixelFormat is a BCn format, e.g. DTX or DXN
```
MoochMcGee: *DXT

bunnei: ah good catch-ty

MoochMcGee:
You should probably put the tile size calculations in the debugger too, but other than that, lgtm!!

bunnei:
@MoochMcGee I did not want to duplicate the code because I'm not sure anyone uses this feature. If anyone is using it, they can update it - otherwise, I might remove it.

PR #1005, opened 2018-08-10
file_sys: Add support for registration format

DarkLordZach:
This adds support for parsing directories that use nintendo's 'registered' format, namely the internal user/system NAND and SD card and searching them for NCA files. This specific PR only adds support for sys/user NAND as sd card titles use an extra layer of encryption (NAX0, but stay tuned). 

**NON-DEVS/REVIEWERS:** Unfortunately, all of my games except MK8DX + Update are on my sd card, so testing the 'drag and drop' functionality is based all on that. Everyone who can, please, use HacDiskMount to dump you nand and see what shows up in the game list. Send me your results in discord.

Adds:
- `RegisteredCache` to scan directories for metadata and use that metadata to find all NCAs within.
**NOTE**: Instead of scanning for all NCAs, I decided to rather scan for meta NCAs and use those to determine which are which. This is because updates to games are registered under the same title id as the base, not the update title id, so the only way to differentiate them is to parse the meta and see. 
- Drag & drop support in user and system NAND. You should be able to use HacDiskMount and copy the `Contents` folder of your user or system nand into `%YUZU_DIR%/nand/<>` where `<>` is `user` or `system` and, provided you have all your keys in order, it should *just* work. You **cannot** however just drop any random NCA into this folder, for the reasons in the note above, but you can:
- Add `Install to NAND` menu option in frontend. This will:
    - For XCI: Copy all the NCAs into the cache, doing some sanity checking on the meta to make sure the NCAs will be accessible in the future.
    - For NCA: Creates a new metadata file (or appends to old one with same title id) a content record and then copies the NCA in. When picking the title type for installed NCAs, `Game` will do you fine for everything right now. System archives will use `System Archive`, but that's coming soon.
- Add a partial implementation of `BISFactory`, mainly as a class to store the registered caches for system/user NAND. This is because there really should only be one instance at a time per directory. I haven't tested multiple, but it likely won't work too well. Also, the constructor of `RegisteredCache` is quite expensive, especially with larger NANDs.

There are a few more things I would like to add before this gets merged:
- [x] Search installed NAND for icon/name when it cannot be found in the games dir
- [x] Provide some form of response to the user when installing titles besides freezing the program.
install
- [x] Add ability to detect and force overwrite of already installed titles.
- [ ] More thorough testing of drag/drop NAND (see note at top)

Finally, reviewers: No commit overwrites another, it's fine to go commit by commit. I tried to break it up as much as possible, it's just a lot. Sorry

### DIFF
--- a/src/common/hex_util.h
+++ b/src/common/hex_util.h
@@ -5,6 +5,8 @@
 #pragma once
 
 #include <array>
+#include <cstddef>
+#include <string>
 #include <fmt/format.h>
 #include "common/common_types.h"
 
--- a/src/common/hex_util.h
+++ b/src/common/hex_util.h
@@ -1,0 +1,35 @@
+// Copyright 2013 Dolphin Emulator Project / 2014 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <fmt/format.h>
+#include "common/common_types.h"
+
+u8 ToHexNibble(char c1);
+
+template <size_t Size, bool le = false>
+std::array<u8, Size> HexStringToArray(std::string_view str) {
+    std::array<u8, Size> out{};
+    if constexpr (le) {
+        for (size_t i = 2 * Size - 2; i <= 2 * Size; i -= 2)
+            out[i / 2] = (ToHexNibble(str[i]) << 4) | ToHexNibble(str[i + 1]);
+    } else {
+        for (size_t i = 0; i < 2 * Size; i += 2)
+            out[i / 2] = (ToHexNibble(str[i]) << 4) | ToHexNibble(str[i + 1]);
+    }
+    return out;
+}
+
+template <size_t Size>
+std::string HexArrayToString(std::array<u8, Size> array, bool upper = true) {
+    std::string out;
+    for (u8 c : array)
+        out += fmt::format(upper ? "{:02X}" : "{:02x}", c);
+    return out;
+}
+
+std::array<u8, 0x10> operator"" _array16(const char* str, size_t len);
+std::array<u8, 0x20> operator"" _array32(const char* str, size_t len);
--- a/src/core/file_sys/nca_metadata.cpp
+++ b/src/core/file_sys/nca_metadata.cpp
@@ -10,7 +10,7 @@
 
 namespace FileSys {
 
-CNMT::CNMT(VirtualFile file_) : file(std::move(file_)), header(std::make_unique<CNMTHeader>()) {
+CNMT::CNMT(VirtualFile file) : header(std::make_unique<CNMTHeader>()) {
     if (file->ReadObject(header.get()) != sizeof(CNMTHeader))
         return;
 
@@ -41,7 +41,7 @@ CNMT::CNMT(VirtualFile file_) : file(std::move(file_)), header(std::make_unique<
 
 CNMT::CNMT(CNMTHeader header, OptionalHeader opt_header, std::vector<ContentRecord> content_records,
            std::vector<MetaRecord> meta_records)
-    : file(nullptr), header(std::make_unique<CNMTHeader>(std::move(header))),
+    : header(std::make_unique<CNMTHeader>(std::move(header))),
       opt_header(std::make_unique<OptionalHeader>(std::move(opt_header))),
       content_records(std::move(content_records)), meta_records(std::move(meta_records)) {}
 
--- a/src/core/file_sys/nca_metadata.cpp
+++ b/src/core/file_sys/nca_metadata.cpp
@@ -103,7 +103,10 @@ bool CNMT::UnionRecords(const CNMT& other) {
 std::vector<u8> CNMT::Serialize() const {
     const bool has_opt_header =
         header.type >= TitleType::Application && header.type <= TitleType::AOC;
-    std::vector<u8> out(sizeof(CNMTHeader) + (has_opt_header ? sizeof(OptionalHeader) : 0));
+    const auto dead_zone = header.table_offset + sizeof(CNMTHeader);
+    std::vector<u8> out(
+        std::max(sizeof(CNMTHeader) + (has_opt_header ? sizeof(OptionalHeader) : 0), dead_zone) +
+        content_records.size() * sizeof(ContentRecord) + meta_records.size() * sizeof(MetaRecord));
     memcpy(out.data(), &header, sizeof(CNMTHeader));
 
     // Optional Header
@@ -113,17 +116,11 @@ std::vector<u8> CNMT::Serialize() const {
 
     auto offset = header.table_offset;
 
-    const auto dead_zone = offset + sizeof(CNMTHeader) - out.size();
-    if (dead_zone > 0)
-        out.resize(offset + sizeof(CNMTHeader));
-
-    out.resize(out.size() + content_records.size() * sizeof(ContentRecord));
     for (const auto& rec : content_records) {
         memcpy(out.data() + offset + sizeof(CNMTHeader), &rec, sizeof(ContentRecord));
         offset += sizeof(ContentRecord);
     }
 
-    out.resize(out.size() + content_records.size() * sizeof(MetaRecord));
     for (const auto& rec : meta_records) {
         memcpy(out.data() + offset + sizeof(CNMTHeader), &rec, sizeof(MetaRecord));
         offset += sizeof(MetaRecord);
--- a/src/core/file_sys/nca_metadata.cpp
+++ b/src/core/file_sys/nca_metadata.cpp
@@ -4,36 +4,44 @@
 
 #include <cstring>
 #include "common/common_funcs.h"
+#include "common/logging/log.h"
 #include "common/swap.h"
 #include "content_archive.h"
 #include "core/file_sys/nca_metadata.h"
 
 namespace FileSys {
 
-CNMT::CNMT(VirtualFile file) : header(std::make_unique<CNMTHeader>()) {
-    if (file->ReadObject(header.get()) != sizeof(CNMTHeader))
+bool operator>=(TitleType lhs, TitleType rhs) {
+    return static_cast<size_t>(lhs) >= static_cast<size_t>(rhs);
+}
+
+bool operator<=(TitleType lhs, TitleType rhs) {
+    return static_cast<size_t>(lhs) <= static_cast<size_t>(rhs);
+}
+
+CNMT::CNMT(VirtualFile file) {
+    if (file->ReadObject(&header) != sizeof(CNMTHeader))
         return;
 
     // If type is {Application, Update, AOC} has opt-header.
-    if (static_cast<u8>(header->type) >= 0x80 && static_cast<u8>(header->type) <= 0x82) {
-        opt_header = std::make_unique<OptionalHeader>();
-        if (file->ReadObject(opt_header.get(), sizeof(CNMTHeader)) != sizeof(OptionalHeader)) {
-            opt_header = nullptr;
+    if (header.type >= TitleType::Application && header.type <= TitleType::AOC) {
+        if (file->ReadObject(&opt_header, sizeof(CNMTHeader)) != sizeof(OptionalHeader)) {
+            LOG_WARNING(Loader, "Failed to read optional header.");
         }
     }
 
-    for (u16 i = 0; i < header->number_content_entries; ++i) {
+    for (u16 i = 0; i < header.number_content_entries; ++i) {
         auto& next = content_records.emplace_back(ContentRecord{});
         if (file->ReadObject(&next, sizeof(CNMTHeader) + i * sizeof(ContentRecord) +
-                                        header->table_offset) != sizeof(ContentRecord)) {
+                                        header.table_offset) != sizeof(ContentRecord)) {
             content_records.erase(content_records.end() - 1);
         }
     }
 
-    for (u16 i = 0; i < header->number_meta_entries; ++i) {
+    for (u16 i = 0; i < header.number_meta_entries; ++i) {
         auto& next = meta_records.emplace_back(MetaRecord{});
         if (file->ReadObject(&next, sizeof(CNMTHeader) + i * sizeof(MetaRecord) +
-                                        header->table_offset) != sizeof(MetaRecord)) {
+                                        header.table_offset) != sizeof(MetaRecord)) {
             meta_records.erase(meta_records.end() - 1);
         }
     }
@@ -41,20 +49,19 @@ CNMT::CNMT(VirtualFile file) : header(std::make_unique<CNMTHeader>()) {
 
 CNMT::CNMT(CNMTHeader header, OptionalHeader opt_header, std::vector<ContentRecord> content_records,
            std::vector<MetaRecord> meta_records)
-    : header(std::make_unique<CNMTHeader>(std::move(header))),
-      opt_header(std::make_unique<OptionalHeader>(std::move(opt_header))),
+    : header(std::move(header)), opt_header(std::move(opt_header)),
       content_records(std::move(content_records)), meta_records(std::move(meta_records)) {}
 
 u64 CNMT::GetTitleID() const {
-    return header->title_id;
+    return header.title_id;
 }
 
 u32 CNMT::GetTitleVersion() const {
-    return header->title_version;
+    return header.title_version;
 }
 
 TitleType CNMT::GetType() const {
-    return header->type;
+    return header.type;
 }
 
 const std::vector<ContentRecord>& CNMT::GetContentRecords() const {
@@ -74,7 +81,7 @@ bool CNMT::UnionRecords(const CNMT& other) {
                                        });
         if (iter == content_records.end()) {
             content_records.emplace_back(rec);
-            ++header->number_content_entries;
+            ++header.number_content_entries;
             change = true;
         }
     }
@@ -86,7 +93,7 @@ bool CNMT::UnionRecords(const CNMT& other) {
             });
         if (iter == meta_records.end()) {
             meta_records.emplace_back(rec);
-            ++header->number_meta_entries;
+            ++header.number_meta_entries;
             change = true;
         }
     }
@@ -94,30 +101,30 @@ bool CNMT::UnionRecords(const CNMT& other) {
 }
 
 std::vector<u8> CNMT::Serialize() const {
-    if (header == nullptr)
-        return {};
-    std::vector<u8> out(sizeof(CNMTHeader));
-    out.reserve(0x100); // Avoid resizing -- average size.
-    memcpy(out.data(), header.get(), sizeof(CNMTHeader));
-    if (opt_header != nullptr) {
-        out.resize(out.size() + sizeof(OptionalHeader));
-        memcpy(out.data() + sizeof(CNMTHeader), opt_header.get(), sizeof(OptionalHeader));
+    const bool has_opt_header =
+        header.type >= TitleType::Application && header.type <= TitleType::AOC;
+    std::vector<u8> out(sizeof(CNMTHeader) + (has_opt_header ? sizeof(OptionalHeader) : 0));
+    memcpy(out.data(), &header, sizeof(CNMTHeader));
+
+    // Optional Header
+    if (has_opt_header) {
+        memcpy(out.data() + sizeof(CNMTHeader), &opt_header, sizeof(OptionalHeader));
     }
 
-    auto offset = header->table_offset;
+    auto offset = header.table_offset;
 
     const auto dead_zone = offset + sizeof(CNMTHeader) - out.size();
     if (dead_zone > 0)
         out.resize(offset + sizeof(CNMTHeader));
 
+    out.resize(out.size() + content_records.size() * sizeof(ContentRecord));
     for (const auto& rec : content_records) {
-        out.resize(out.size() + sizeof(ContentRecord));
         memcpy(out.data() + offset + sizeof(CNMTHeader), &rec, sizeof(ContentRecord));
         offset += sizeof(ContentRecord);
     }
 
+    out.resize(out.size() + content_records.size() * sizeof(MetaRecord));
     for (const auto& rec : meta_records) {
-        out.resize(out.size() + sizeof(MetaRecord));
         memcpy(out.data() + offset + sizeof(CNMTHeader), &rec, sizeof(MetaRecord));
         offset += sizeof(MetaRecord);
     }
--- a/src/core/file_sys/nca_metadata.cpp
+++ b/src/core/file_sys/nca_metadata.cpp
@@ -1,0 +1,125 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/common_funcs.h"
+#include "common/swap.h"
+#include "content_archive.h"
+#include "core/file_sys/nca_metadata.h"
+
+namespace FileSys {
+
+CNMT::CNMT(VirtualFile file_) : file(std::move(file_)), header(std::make_unique<CNMTHeader>()) {
+    if (file->ReadObject(header.get()) != sizeof(CNMTHeader))
+        return;
+
+    // If type is {Application, Update, AOC} has opt-header.
+    if (static_cast<u8>(header->type) >= 0x80 && static_cast<u8>(header->type) <= 0x82) {
+        opt_header = std::make_unique<OptionalHeader>();
+        if (file->ReadObject(opt_header.get(), sizeof(CNMTHeader)) != sizeof(OptionalHeader)) {
+            opt_header = nullptr;
+        }
+    }
+
+    for (u16 i = 0; i < header->number_content_entries; ++i) {
+        auto& next = content_records.emplace_back(ContentRecord{});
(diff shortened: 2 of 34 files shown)


src/common/hex_util.h
```diff
@@ -0,0 +1,35 @@
+// Copyright 2013 Dolphin Emulator Project / 2014 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
```
lioncash: `<cstddef>` and `<string>` needs to be included

src/core/file_sys/nca_metadata.cpp
```diff
@@ -0,0 +1,125 @@
...
+const std::vector<MetaRecord>& CNMT::GetMetaRecords() const {
+    return meta_records;
+}
+
+bool CNMT::UnionRecords(const CNMT& other) {
+    bool change = false;
+    for (const auto& rec : other.content_records) {
+        const auto iter = std::find_if(
+            content_records.begin(), content_records.end(),
+            [rec](const ContentRecord& r) { return r.nca_id == rec.nca_id && r.type == rec.type; });
```
lioncash: `This should likely be `&rec` to avoid taking it by value and doing an unnecessary copy.

src/core/file_sys/nca_metadata.cpp
```diff
@@ -0,0 +1,125 @@
...
+        if (iter == content_records.end()) {
+            content_records.emplace_back(rec);
+            ++header->number_content_entries;
+            change = true;
+        }
+    }
+    for (const auto& rec : other.meta_records) {
+        const auto iter =
+            std::find_if(meta_records.begin(), meta_records.end(), [rec](const MetaRecord& r) {
```
lioncash: ditto

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+std::string RegisteredCacheEntry::DebugInfo() const {
+    return fmt::format("title_id={:016X}, content_type={:02X}", title_id, static_cast<u8>(type));
+}
+
+bool operator<(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs) {
+    return (lhs.title_id < rhs.title_id) || (lhs.title_id == rhs.title_id && lhs.type < rhs.type);
+}
+
+static bool FollowsTwoDigitDirFormat(std::string_view name) {
+    const static std::regex two_digit_regex(
```
lioncash: `static const`, the lifetime specifier should always be first.

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+        "000000[0123456789abcdefABCDEF][0123456789abcdefABCDEF]");
+    return std::regex_match(name.begin(), name.end(), two_digit_regex);
+}
+
+static bool FollowsNcaIdFormat(std::string_view name) {
+    const static std::regex nca_id_regex("[0123456789abcdefABCDEF]+.nca");
```
lioncash: Ditto

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+                                                       std::string_view path) const {
+    if (dir->GetFileRelative(path) != nullptr)
+        return dir->GetFileRelative(path);
+    if (dir->GetDirectoryRelative(path) != nullptr) {
+        const auto nca_dir = dir->GetDirectoryRelative(path);
+        VirtualFile file = nullptr;
+
+        const auto files = nca_dir->GetFiles();
+        if (files.size() == 1 && files[0]->GetName() == "00")
+            file = files[0];
```
lioncash: If the else case is braced, the if case should be braced as well for consistency.

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+        else {
+            std::vector<VirtualFile> concat;
+            for (u8 i = 0; i < 0x10; ++i) {
+                auto next = nca_dir->GetFile(fmt::format("{:02X}", i));
+                if (next != nullptr)
```
lioncash: Ditto

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+        if (file != nullptr)
+            return file;
+    }
+    return file;
+}
+
+boost::optional<NcaID> RegisteredCache::GetNcaIDFromMetadata(u64 title_id,
+                                                             ContentRecordType type) const {
+    if (type == ContentRecordType::Meta && meta_id.find(title_id) != meta_id.end())
+        return meta_id.at(title_id);
```
lioncash: Given this already bounds checks the range with the find, the `at()` can just be a regular indexing operation.

src/core/file_sys/registered_cache.h
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * (This impl also supports substituting the nca dir for an nca file, as that's more convenient when
+ * 4GB splitting can be ignored.)
+ */
+class RegisteredCache {
+public:
+    // Parsing function defines the conversion from raw file to NCA. If there are other steps
+    // besides creating the NCA from the file (e.g. NAX0 on SD Card), that should go in a custom
+    // parsing function.
+    RegisteredCache(VirtualDir dir,
```
lioncash: `explicit RegisteredCache`

src/core/file_sys/vfs_concat.cpp
```diff
@@ -0,0 +1,93 @@
...
+VirtualFile ConcatenateFiles(std::vector<VirtualFile> files, std::string_view name) {
+    if (files.empty())
+        return nullptr;
+    if (files.size() == 1)
+        return files[0];
+
+    return std::shared_ptr<VfsFile>(new ConcatenatedVfsFile(std::move(files), name));
+}
+
+ConcatenatedVfsFile::ConcatenatedVfsFile(std::vector<VirtualFile> files_, std::string_view name)
```
lioncash: It's better interface-wise here to take a std::string and std::move it into the member, since it can avoid copying altogether depending on the calling code.

src/core/file_sys/vfs_concat.h
```diff
@@ -0,0 +1,41 @@
...
+
+// Wrapper function to allow for more efficient handling of files.size() == 0, 1 cases.
+VirtualFile ConcatenateFiles(std::vector<VirtualFile> files, std::string_view name = "");
+
+// Class that wraps multiple vfs files and concatenates them, making reads seamless. Currently
+// read-only.
+class ConcatenatedVfsFile : public VfsFile {
+    friend VirtualFile ConcatenateFiles(std::vector<VirtualFile> files, std::string_view name);
+
+    ConcatenatedVfsFile(std::vector<VirtualFile> files, std::string_view name);
```
lioncash: ditto wrt moving

src/yuzu/main.cpp
```diff
@@ -654,6 +660,89 @@ void GMainWindow::OnMenuLoadFolder() {
     }
 }
 
+void GMainWindow::OnMenuInstallToNAND() {
+    const static QString file_filter =
```
lioncash: In this case it doesn't really have any benefit to use static const, given it's UI code (and runtime translation with `tr` won't function correctly here like this if it's `static`.

src/yuzu/main.cpp
```diff
@@ -654,6 +660,89 @@ void GMainWindow::OnMenuLoadFolder() {
...
+                vfs->OpenFile(filename.toStdString(), FileSys::Mode::Read));
+            if (nca->GetStatus() != Loader::ResultStatus::Success) {
+                QMessageBox::critical(
+                    this, tr("Failed to Install NCA"),
+                    tr("The NCA file you provided is invalid. Please double-check your encryption "
+                       "keys and the file and try again."));
+                return;
+            }
+
+            const static QStringList tt_options{"System Application",
```
lioncash: `static const`

src/yuzu/main.cpp
```diff
@@ -654,6 +660,89 @@ void GMainWindow::OnMenuLoadFolder() {
...
+            if (!ok || index == -1) {
+                QMessageBox::critical(this, tr("Failed to Install NCA"),
+                                      tr("The title type you selected for the NCA is invalid."));
+                return;
+            }
+
+            if (index >= 5)
+                index += 0x80;
+
+            if (!Service::FileSystem::GetUserNANDContents()->InstallEntry(
```
lioncash: This can be inverted to operate in terms of the true case.

src/yuzu/main.cpp
```diff
@@ -654,6 +660,89 @@ void GMainWindow::OnMenuLoadFolder() {
...
+                    nca, static_cast<FileSys::TitleType>(index))) {
+                QMessageBox::critical(this, tr("Failed to Install NCA"),
```
lioncash: Given this isn't a fatal error, this should be using `warning` instead of critical.

src/yuzu/main.cpp
```diff
@@ -654,6 +660,89 @@ void GMainWindow::OnMenuLoadFolder() {
...
+            const auto xci = std::make_shared<FileSys::XCI>(
+                vfs->OpenFile(filename.toStdString(), FileSys::Mode::Read));
+            if (xci->GetStatus() != Loader::ResultStatus::Success) {
+                QMessageBox::critical(
+                    this, tr("Failed to Install XCI"),
+                    tr("The XCI file you provided is invalid. Please double-check your encryption "
+                       "keys and the file and try again."));
+                return;
+            }
+            if (!Service::FileSystem::GetUserNANDContents()->InstallEntry(xci)) {
```
lioncash: This can be inverted in terms of the true case.

src/yuzu/main.cpp
```diff
@@ -654,6 +660,89 @@ void GMainWindow::OnMenuLoadFolder() {
...
+                QMessageBox::critical(
```
lioncash: Given this isn't a fatal error this should be a warning.

src/yuzu/main.cpp
```diff
@@ -654,6 +660,89 @@ void GMainWindow::OnMenuLoadFolder() {
...
+        tr("Installable Switch File (*.nca *.xci);;Nintendo Content Archive (*.nca);;NX Cartridge "
+           "Image (*.xci)");
+    QString filename = QFileDialog::getOpenFileName(this, tr("Install File"),
+                                                    UISettings::values.roms_path, file_filter);
+    if (!filename.isEmpty()) {
+        if (filename.endsWith("xci", Qt::CaseInsensitive)) {
+            const auto xci = std::make_shared<FileSys::XCI>(
+                vfs->OpenFile(filename.toStdString(), FileSys::Mode::Read));
+            if (xci->GetStatus() != Loader::ResultStatus::Success) {
+                QMessageBox::critical(
```
lioncash: Given this isn't a fatal error this should be a warning.

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+        "SystemUpdate",
+        "BootImagePackage",
+        "BootImagePackageSafe",
+        "Application",
+        "Patch",
+        "AddOnContent",
+        "" ///< Currently unknown 'DeltaTitle'
+    };
+
+    size_t index = static_cast<size_t>(type);
```
lioncash: `auto index` given the cast already gives the destination type.

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+void RegisteredCache::ProcessFiles(const std::vector<NcaID>& ids) {
+    for (const auto& id : ids) {
+        const auto file = GetFileAtID(id);
+
+        if (file == nullptr)
+            continue;
+        const auto nca = std::make_shared<NCA>(parser(file, id));
+        if (nca->GetStatus() != Loader::ResultStatus::Success ||
+            nca->GetType() != NCAContentType::Meta)
```
lioncash: Should be braced since the conditional travels more than one line.

src/core/file_sys/registered_cache.h
```diff
@@ -0,0 +1,108 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <map>
+#include <memory>
+#include <string>
+#include <boost/container/flat_map.hpp>
+#include "common/common_funcs.h"
```
lioncash: `common/common_types.h` also needs to be included.

src/core/file_sys/registered_cache.h
```diff
@@ -0,0 +1,108 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <map>
```
lioncash: `<functional>` and `<vector>` need to be included.

src/core/file_sys/nca_metadata.h
```diff
@@ -0,0 +1,105 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
+#include "core/file_sys/vfs.h"
```
lioncash: `<vector>`, `"common/swap.h`, and `"common/common_types.h"` need to be included.

src/core/file_sys/nca_metadata.cpp
```diff
@@ -0,0 +1,125 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/common_funcs.h"
```
lioncash: `<cstring>` needs to be included.

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+    const auto iter = std::find_if(cnmt.GetContentRecords().begin(), cnmt.GetContentRecords().end(),
+                                   [type](const ContentRecord& rec) { return rec.type == type; });
+    if (iter == cnmt.GetContentRecords().end())
+        return boost::none;
+
+    return boost::make_optional(iter->nca_id);
+}
+
+void RegisteredCache::AccumulateFiles(std::vector<NcaID>& ids) const {
```
lioncash: This should likely return the vector by value.

src/core/file_sys/registered_cache.cpp
DarkLordZach: This won't work as the method is const and the `operator[]` on flat map isn't because it creates if it doesn't exist (and the overload that dosen't do that is `at`)

src/core/file_sys/vfs_concat.cpp
```diff
@@ -0,0 +1,93 @@
...
+size_t ConcatenatedVfsFile::Read(u8* data, size_t length, size_t offset) const {
+    auto entry = files.end();
+    for (auto iter = files.begin(); iter != files.end(); ++iter) {
+        if (iter->first > offset) {
+            entry = --iter;
+            break;
+        }
+    }
+
+    if (entry == files.end() && offset < files.rbegin()->first + files.rbegin()->second->GetSize())
```
Subv: What is this checking for, and why? please add a comment

src/core/file_sys/vfs_real.cpp
```diff
@@ -83,7 +83,10 @@ VirtualFile RealVfsFilesystem::OpenFile(std::string_view path_, Mode perms) {
 
 VirtualFile RealVfsFilesystem::CreateFile(std::string_view path_, Mode perms) {
     const auto path = FileUtil::SanitizePath(path_, FileUtil::DirectorySeparator::PlatformDefault);
-    if (!FileUtil::Exists(path) && !FileUtil::CreateEmptyFile(path))
+    if (!FileUtil::Exists(path) &&
```
Subv: Please split this if into multiple ones, as it is now it is a pain to read.

src/core/file_sys/nca_metadata.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "core/file_sys/nca_metadata.h"
+
+namespace FileSys {
+
+CNMT::CNMT(VirtualFile file_) : file(std::move(file_)), header(std::make_unique<CNMTHeader>()) {
+    if (file->ReadObject(header.get()) != sizeof(CNMTHeader))
+        return;
+
+    // If type is {Application, Update, AOC} has opt-header.
+    if (static_cast<u8>(header->type) >= 0x80 && static_cast<u8>(header->type) <= 0x82) {
```
Subv: Don't use magic numbers, add an enum for this

src/core/file_sys/nca_metadata.cpp
```diff
@@ -0,0 +1,125 @@
...
+                                        header->table_offset) != sizeof(MetaRecord)) {
+            meta_records.erase(meta_records.end() - 1);
+        }
+    }
+}
+
+CNMT::CNMT(CNMTHeader header, OptionalHeader opt_header, std::vector<ContentRecord> content_records,
+           std::vector<MetaRecord> meta_records)
+    : file(nullptr), header(std::make_unique<CNMTHeader>(std::move(header))),
+      opt_header(std::make_unique<OptionalHeader>(std::move(opt_header))),
```
Subv: What is this supposed to do? you're copying opt_header, then making a pointer out of it?, same with header

src/core/file_sys/nca_metadata.cpp
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+    return change;
+}
+
+std::vector<u8> CNMT::Serialize() const {
+    if (header == nullptr)
+        return {};
+    std::vector<u8> out(sizeof(CNMTHeader));
+    out.reserve(0x100); // Avoid resizing -- average size.
```
Subv: It seems pretty clear to me that the size is always `sizeof(CNMTHeader) + (opt_header ? sizeof(OptionalHeader) : 0) + content_records.size() * sizeof(ContentRecord)` so why are you guessing 0x100? Get rid of the superfluous resize calls.

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+    return fmt::format("title_id={:016X}, content_type={:02X}", title_id, static_cast<u8>(type));
+}
+
+bool operator<(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs) {
+    return (lhs.title_id < rhs.title_id) || (lhs.title_id == rhs.title_id && lhs.type < rhs.type);
+}
+
+static bool FollowsTwoDigitDirFormat(std::string_view name) {
+    const static std::regex two_digit_regex(
+        "000000[0123456789abcdefABCDEF][0123456789abcdefABCDEF]");
```
Subv: does `std::regex two_digit_regex(000000[0-9A-F]{2}, std::regex_constants::ECMAScript | std::regex_constants::icase)` not work?

src/core/file_sys/registered_cache.cpp
Subv: use `icase` and `[0-9A-F]+\\.nca`

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+        "BootImagePackage",
+        "BootImagePackageSafe",
+        "Application",
+        "Patch",
+        "AddOnContent",
+        "" ///< Currently unknown 'DeltaTitle'
+    };
+
+    size_t index = static_cast<size_t>(type);
+    if (index >= 0x80)
```
Subv: What is this doing? don't use magic numbers and add a comment

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+        return file;
+    }
+    return nullptr;
+}
+
+VirtualFile RegisteredCache::GetFileAtID(NcaID id) const {
+    VirtualFile file;
+    for (u8 i = 0; i < 4; ++i) {
+        file = OpenFileOrDirectoryConcat(
```
Subv: Please split this up, it's very difficult to read, also what are those bit checks for? add a comment.

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+            meta.insert_or_assign(nca->GetTitleId(), CNMT(file));
+            meta_id.insert_or_assign(nca->GetTitleId(), id);
+            break;
+        }
+    }
+}
+
+void RegisteredCache::AccumulateYuzuMeta() {
+    const auto dir = this->dir->GetSubdirectory("yuzu_meta");
```
Subv: I don't really like the idea of having a yuzu_meta directory, what does it contain and why is it needed? Please do a write up about this so we can discuss improvements

src/yuzu/main.cpp
```diff
@@ -654,6 +660,89 @@ void GMainWindow::OnMenuLoadFolder() {
     }
 }
 
+void GMainWindow::OnMenuInstallToNAND() {
+    const static QString file_filter =
+        tr("Installable Switch File (*.nca *.xci);;Nintendo Content Archive (*.nca);;NX Cartridge "
```
Subv: Why do we allow the installation of XCI files if they are not intended to be installed?

src/core/file_sys/nca_metadata.cpp
```diff
@@ -0,0 +1,134 @@
...
+        memcpy(out.data() + sizeof(CNMTHeader), &opt_header, sizeof(OptionalHeader));
+    }
+
+    auto offset = header.table_offset;
+
+    const auto dead_zone = offset + sizeof(CNMTHeader) - out.size();
+    if (dead_zone > 0)
+        out.resize(offset + sizeof(CNMTHeader));
+
+    out.resize(out.size() + content_records.size() * sizeof(ContentRecord));
```
Subv: Why are you resizing so many times? This is inefficient, the size will not change, calculate the total space needed at the beginning and use that.

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,473 @@
...
+}
+
+static bool FollowsTwoDigitDirFormat(std::string_view name) {
+    static const std::regex two_digit_regex("000000[0-9A-F]{2}", std::regex_constants::ECMAScript |
+                                                                     std::regex_constants::icase);
+    return std::regex_match(name.begin(), name.end(), two_digit_regex);
+}
+
+static bool FollowsNcaIdFormat(std::string_view name) {
+    static const std::regex nca_id_regex("[0-9A-F]{32}.nca", std::regex_constants::ECMAScript |
```
Subv: This dot (.) is also a regex atom, you have to escape it `\\.`

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,473 @@
...
+        "Application",
+        "Patch",
+        "AddOnContent",
+        "" ///< Currently unknown 'DeltaTitle'
+    };
+
+    auto index = static_cast<size_t>(type);
+    // If the index is after the jump in TitleType, subtract it out.
+    if (index >= static_cast<size_t>(TitleType::Application))
+        index -= 0x7B;
```
Subv: magic number?

src/core/file_sys/registered_cache.cpp
```diff
@@ -0,0 +1,473 @@
...
+    if (dir->GetDirectoryRelative(path) != nullptr) {
+        const auto nca_dir = dir->GetDirectoryRelative(path);
+        VirtualFile file = nullptr;
+
+        const auto files = nca_dir->GetFiles();
+        if (files.size() == 1 && files[0]->GetName() == "00") {
+            file = files[0];
+        } else {
+            std::vector<VirtualFile> concat;
+            for (u8 i = 0; i < 0x10; ++i) {
```
Subv: where is this 0x10 coming from?

src/core/file_sys/vfs_real.cpp
```diff
@@ -83,7 +83,9 @@ VirtualFile RealVfsFilesystem::OpenFile(std::string_view path_, Mode perms) {
 
 VirtualFile RealVfsFilesystem::CreateFile(std::string_view path_, Mode perms) {
     const auto path = FileUtil::SanitizePath(path_, FileUtil::DirectorySeparator::PlatformDefault);
-    if (!FileUtil::Exists(path) && !FileUtil::CreateEmptyFile(path))
+    const auto path_fwd = FileUtil::SanitizePath(path, FileUtil::DirectorySeparator::ForwardSlash);
+    if (!FileUtil::Exists(path) && !FileUtil::CreateFullPath(path_fwd) &&
```
Subv: Please re-check the logic of this if statement

src/core/file_sys/vfs_real.cpp
```diff
@@ -140,7 +142,9 @@ VirtualDir RealVfsFilesystem::OpenDirectory(std::string_view path_, Mode perms)
 
 VirtualDir RealVfsFilesystem::CreateDirectory(std::string_view path_, Mode perms) {
     const auto path = FileUtil::SanitizePath(path_, FileUtil::DirectorySeparator::PlatformDefault);
-    if (!FileUtil::Exists(path) && !FileUtil::CreateDir(path))
+    const auto path_fwd = FileUtil::SanitizePath(path, FileUtil::DirectorySeparator::ForwardSlash);
+    if (!FileUtil::Exists(path) && !FileUtil::CreateFullPath(path_fwd) &&
```
Subv: same

PR #1012, opened 2018-08-10
filesystem: Create directory if it dosen't exist on open

DarkLordZach:
It would seem that Ultra Street Fighter II requires this behavior on save creation.


DarkLordZach:
Hardware testing confirms that this was indeed a hack and the correct behavior is to return error code `0x2EEA02`

ogniK5377:
We have a few issues with FS and this definitely does not seem like the correct action. I think we'll need to review a few FS cmds to address folders not being created.

bunnei:
Closing, as it seems we don't need this at all anymore.

PR #1013, opened 2018-08-10
Prevent PixelFormatFromRenderTargetFormat from returning an uninitialized value

khang06:
Fixes Star Allies crashing on canary. Usually I would implement what it's trying to use, but #848 has it already. (R16_UNORM)


khang06:
closed (see discord for details)

PR #1016, opened 2018-08-10
video_core: Get rid of global variable g_toggle_framelimit_enabled

lioncash:
Instead, we make a struct for renderer settings and allow the renderer to update all of these settings, getting rid of the need for global-scoped variables.

This also uncovered a few indirect inclusions for certain headers, which this also fixes.

With this, we get rid of all main accessible global state within the video core 🎉

### DIFF
--- a/src/core/hle/service/nvflinger/buffer_queue.h
+++ b/src/core/hle/service/nvflinger/buffer_queue.h
@@ -6,6 +6,7 @@
 
 #include <vector>
 #include <boost/optional.hpp>
+#include "common/common_funcs.h"
 #include "common/math_util.h"
 #include "common/swap.h"
 #include "core/hle/kernel/event.h"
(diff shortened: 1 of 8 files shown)


jroweboy:
iirc the framelimit isn't used in video core at all (and looking at this change I don't see it being used anywhere anyway) but I do like the idea to have that RenderSetting struct ready for when we stick the upscaling factor in there.

lioncash:
Yeah, I was opting between removing it entirely and leaving it in, and I decided to leave it in because eventually, once games are going to be somewhat playable, we're going to need a place to store the "do we do frame limiting or not?" variable, and this was also a way to establish a settings interface ahead of time, so I went with that.

jroweboy:
thats understandable. in citra at least (and by extension in yuzu) the framelimiter happens in the core instead of video core. I'm fine keeping it and merging whats here because like you said it establishes a settings interface ahead of time, but i don't think this particular variable will have any use :)

PR #1017, opened 2018-08-11
New account backend to allow for future extended support

ogniK5377:
This is the start of an account management backend for acc:*. In the future, importing users from console, custom icons, and a UI will be added. Currently, it should work the same as it does now except on the new backend.


src/core/hle/service/acc/profile_manager.h
```diff
@@ -0,0 +1,111 @@
...
+    bool GetProfileBaseAndData(UUID uuid, ProfileBase& profile, std::array<u8, MAX_DATA>& data);
+    bool GetProfileBaseAndData(ProfileInfo user, ProfileBase& profile,
+                               std::array<u8, MAX_DATA>& data);
+    size_t GetUserCount();
+    bool UserExists(UUID uuid);
+    void OpenUser(UUID uuid);
+    void CloseUser(UUID uuid);
+    std::array<UUID, MAX_USERS> GetOpenUsers();
+    std::array<UUID, MAX_USERS> GetAllUsers();
+    const UUID& GetLastOpennedUser();
```
mailwl: I am not native English, "Opened" really should be with 2 "n"?

src/core/hle/service/acc/acc.h
```diff
@@ -5,14 +5,16 @@
 #pragma once
 
 #include "core/hle/service/service.h"
+#include "profile_manager.h"
```
lioncash: should be a relative include

src/core/hle/service/acc/profile_manager.h
```diff
@@ -0,0 +1,113 @@
...
+#include "common/swap.h"
+#include "core/hle/result.h"
+
+namespace Service::Account {
+constexpr size_t MAX_USERS = 8;
+constexpr size_t MAX_DATA = 128;
+
+struct UUID {
+    // UUIDs which are 0 are considered invalid!
+    u128 uuid{0, 0};
```
mailwl: maybe instead comment declare static const INVALID_UUID or something? and use here and in Invalidate()

src/core/hle/service/acc/acc.cpp
```diff
@@ -141,44 +131,57 @@ class IManagerForApplication final : public ServiceFramework<IManagerForApplicat
...
 }
 
 void Module::Interface::ListOpenUsers(Kernel::HLERequestContext& ctx) {
-    LOG_WARNING(Service_ACC, "(STUBBED) called");
-    // TODO(Subv): There is only one user for now.
-    const std::vector<u128> user_ids = {DEFAULT_USER_ID};
-    ctx.WriteBuffer(user_ids);
+    LOG_INFO(Service_ACC, "called");
+    ctx.WriteBuffer(profile_manager->GetOpenUsers());
     IPC::ResponseBuilder rb{ctx, 2};
```
mailwl: probably out of scope of this PR, but the function should return actual count of opened users as well, else we get UB. Something like:
```
rb.Push(RESULT_SUCCESS);
rb.Push((profile_manager->GerOpenedUsersCount());
```
for future, need check if output buffer size enough for all opened users

src/core/hle/service/acc/acc.cpp
```diff
@@ -49,38 +45,31 @@ class IProfile final : public ServiceFramework<IProfile> {
...
+        std::array<u8, MAX_DATA> data{};
+        if (profile_manager.GetProfileBaseAndData(user_id, profile_base, data)) {
+            ctx.WriteBuffer(data);
+            IPC::ResponseBuilder rb{ctx, 16};
+            rb.Push(RESULT_SUCCESS);
+            rb.PushRaw(profile_base);
         } else {
-            std::copy(Settings::values.username.begin(), Settings::values.username.end(),
-                      profile_base.username.begin());
+            IPC::ResponseBuilder rb{ctx, 2};
```
Subv: Add an error log when this happens

src/core/hle/service/acc/acc.cpp
```diff
@@ -49,38 +45,31 @@ class IProfile final : public ServiceFramework<IProfile> {
...
-            std::copy_n(Settings::values.username.begin(), profile_base.username.size(),
-                        profile_base.username.begin());
+        if (profile_manager.GetProfileBase(user_id, profile_base)) {
+            IPC::ResponseBuilder rb{ctx, 16};
+            rb.Push(RESULT_SUCCESS);
+            rb.PushRaw(profile_base);
         } else {
-            std::copy(Settings::values.username.begin(), Settings::values.username.end(),
-                      profile_base.username.begin());
+            IPC::ResponseBuilder rb{ctx, 2};
```
Subv: Add an error log when this happens

src/core/hle/service/acc/acc.cpp
```diff
@@ -104,7 +93,8 @@ class IProfile final : public ServiceFramework<IProfile> {
         rb.Push<u32>(jpeg_size);
     }
 
-    u128 user_id; ///< The user id this profile refers to.
+    ProfileManager& profile_manager;
```
Subv: this can be a const reference, right?

src/core/hle/service/acc/acc.cpp
```diff
@@ -141,44 +131,59 @@ class IManagerForApplication final : public ServiceFramework<IManagerForApplicat
...
-    LOG_WARNING(Service_ACC, "(STUBBED) called");
-    // TODO(Subv): There is only one user for now.
-    const std::vector<u128> user_ids = {DEFAULT_USER_ID};
-    ctx.WriteBuffer(user_ids);
-    IPC::ResponseBuilder rb{ctx, 2};
+    LOG_INFO(Service_ACC, "called");
+    ctx.WriteBuffer(profile_manager->GetAllUsers());
+    IPC::ResponseBuilder rb{ctx, 3};
     rb.Push(RESULT_SUCCESS);
+    rb.Push<u32>(static_cast<u32>(profile_manager->GetUserCount()));
```
Subv: Are you sure this returns the count? I couldn't find this on SwIPC

src/core/hle/service/acc/acc.cpp
```diff
@@ -141,44 +131,59 @@ class IManagerForApplication final : public ServiceFramework<IManagerForApplicat
...
-    LOG_WARNING(Service_ACC, "(STUBBED) called");
-    // TODO(Subv): There is only one user for now.
-    const std::vector<u128> user_ids = {DEFAULT_USER_ID};
-    ctx.WriteBuffer(user_ids);
-    IPC::ResponseBuilder rb{ctx, 2};
+    LOG_INFO(Service_ACC, "called");
+    ctx.WriteBuffer(profile_manager->GetOpenUsers());
+    IPC::ResponseBuilder rb{ctx, 3};
     rb.Push(RESULT_SUCCESS);
+    rb.Push<u32>(static_cast<u32>(profile_manager->GetOpenUserCount()));
```
Subv: same

src/core/hle/service/acc/acc.cpp
```diff
@@ -194,22 +199,18 @@ void Module::Interface::GetBaasAccountManagerForApplication(Kernel::HLERequestCo
...
-    rb.Push(RESULT_SUCCESS);
-    rb.PushRaw(DEFAULT_USER_ID);
-}
-
-Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
-    : ServiceFramework(name), module(std::move(module)) {}
+Module::Interface::Interface(std::shared_ptr<Module> module,
+                             std::shared_ptr<ProfileManager> profile_manager, const char* name)
+    : ServiceFramework(name), module(std::move(module)),
+      profile_manager(std::make_shared<ProfileManager>(*profile_manager)) {}
```
Subv: `profile_manager(std::move(profile_manager))`

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include "core/settings.h"
+
+namespace Service::Account {
+// TODO(ogniK): Get actual error codes
+constexpr ResultCode ERROR_TOO_MANY_USERS(ErrorModule::Account, -1);
+constexpr ResultCode ERROR_USER_ALREADY_EXISTS(ErrorModule::Account, -2);
+constexpr ResultCode ERROR_ARGUMENT_IS_NULL(ErrorModule::Account, 20);
+
+ProfileManager::ProfileManager() {
+    auto user_uuid = UUID{1, 0};
```
Subv: Why are you doing this in the constructor? at least add a comment about this code

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+    CreateNewUser(user_uuid, Settings::values.username);
+    OpenUser(user_uuid);
+}
+
+size_t ProfileManager::AddToProfiles(const ProfileInfo& user) {
+    if (user_count >= MAX_USERS) {
```
Subv: Don't do this, use boost::optional and return boost::none to signify the error

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+        return std::numeric_limits<size_t>::max();
+    }
+    profiles[user_count] = std::move(user);
```
Subv: I didn't know you could std::move const references, that sounds dangerous, @lioncash ?

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+    return user_count++;
+}
+
+bool ProfileManager::RemoveProfileAtIdx(size_t index) {
```
Subv: Don't use needless abbreviations, either use `RemoveProfileAtIndex` or simply `RemoveProfile`

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+    if (index >= MAX_USERS || index >= user_count) {
+        return false;
+    }
+    if (index < user_count - 1) {
```
Subv: What is this code supposed to do? I assume `profiles` is a circular buffer? Won't what you're doing invalidate all indices after you remove a profile? Doing this by hand seems very fishy, is there a reason why you didn't use remove-erase?

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+        std::rotate(profiles.begin() + index, profiles.begin() + index + 1, profiles.end());
+    }
+    profiles.back() = {};
+    user_count--;
+    return true;
+}
+
+ResultCode ProfileManager::AddUser(ProfileInfo user) {
+    if (AddToProfiles(user) == std::numeric_limits<size_t>::max()) {
```
Subv: Continuing on from a previous comment, check for boost::none here

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+        return ERROR_ARGUMENT_IS_NULL;
+    }
+    if (std::any_of(profiles.begin(), profiles.end(),
+                    [&uuid](const ProfileInfo& profile) { return uuid == profile.user_uuid; })) {
+        return ERROR_USER_ALREADY_EXISTS;
+    }
+    ProfileInfo prof_inf;
+    prof_inf.user_uuid = std::move(uuid);
+    prof_inf.username = std::move(username);
+    prof_inf.data = std::array<u8, MAX_DATA>();
```
Subv: data = {}

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+        return ERROR_ARGUMENT_IS_NULL;
+    }
+    if (username[0] == 0x0) {
+        return ERROR_ARGUMENT_IS_NULL;
+    }
+    if (std::any_of(profiles.begin(), profiles.end(),
+                    [&uuid](const ProfileInfo& profile) { return uuid == profile.user_uuid; })) {
+        return ERROR_USER_ALREADY_EXISTS;
+    }
+    ProfileInfo prof_inf;
```
Subv: `information` or simply `profile`

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+ResultCode ProfileManager::AddUser(ProfileInfo user) {
+    if (AddToProfiles(user) == std::numeric_limits<size_t>::max()) {
+        return ERROR_TOO_MANY_USERS;
+    }
+    return RESULT_SUCCESS;
+}
+
+ResultCode ProfileManager::CreateNewUser(UUID uuid, std::array<u8, 0x20>& username) {
```
Subv: Having this array be a reference is very misleading, any callers wouldn't expect this function to leave their input array in an unspecified state (due to the std::move)

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+        std::copy_n(username.begin(), username_output.size(), username_output.begin());
+    } else {
+        std::copy(username.begin(), username.end(), username_output.begin());
+    }
+    return CreateNewUser(uuid, username_output);
+}
+
+size_t ProfileManager::GetUserIndex(const UUID& uuid) const {
+    if (!uuid) {
+        return std::numeric_limits<size_t>::max();
```
Subv: Please don't use numeric_limits for this, use an optional type like boost::optional

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+    return static_cast<size_t>(std::distance(profiles.begin(), iter));
+}
+
+size_t ProfileManager::GetUserIndex(ProfileInfo user) const {
+    return GetUserIndex(user.user_uuid);
+}
+
+bool ProfileManager::GetProfileBase(size_t index, ProfileBase& profile) const {
+    if (index >= MAX_USERS) {
+        profile.Invalidate();
```
Subv: why?

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -0,0 +1,194 @@
...
+std::array<UUID, MAX_USERS> ProfileManager::GetAllUsers() const {
+    std::array<UUID, MAX_USERS> output;
+    std::transform(profiles.begin(), profiles.end(), output.begin(),
+                   [](const ProfileInfo& p) { return p.user_uuid; });
+    return output;
+}
+
+std::array<UUID, MAX_USERS> ProfileManager::GetOpenUsers() const {
+    std::array<UUID, MAX_USERS> output;
+    std::transform(profiles.begin(), profiles.end(), output.begin(), [](const ProfileInfo& p) {
```
Subv: pin @lioncash , is there a better way to do this?

src/core/hle/service/acc/profile_manager.h
```diff
@@ -0,0 +1,111 @@
...
+        return uuid[0] == rhs.uuid[0] && uuid[1] == rhs.uuid[1];
+    }
+
+    bool operator!=(const UUID& rhs) {
+        return uuid[0] != rhs.uuid[0] || uuid[1] != rhs.uuid[1];
+    }
+
+    // TODO(ogniK): Properly generate uuids based on RFC-4122
+    const UUID& Generate() {
+        uuid[0] = (static_cast<u64>(std::rand()) << 32) | std::rand();
```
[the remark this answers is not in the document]
Subv: Please don't use std::rand, use one of the random generators from <random>

src/core/hle/service/acc/profile_manager.h
```diff
@@ -0,0 +1,114 @@
...
+    std::array<u8, MAX_DATA> data; // TODO(ognik): Work out what this is
+    bool is_open;
+};
+
+struct ProfileBase {
+    UUID user_uuid;
+    u64_le timestamp;
+    std::array<u8, 0x20> username;
+
+    void Invalidate() {
```
Subv: What does this do and why is it needed? add documentation

src/core/hle/service/acc/profile_manager.h
```diff
@@ -0,0 +1,114 @@
...
+    bool CanSystemRegisterUser() const;
+
+private:
+    std::array<ProfileInfo, MAX_USERS> profiles{};
+    size_t user_count = 0;
+    size_t AddToProfiles(const ProfileInfo& profile);
+    bool RemoveProfileAtIdx(size_t index);
+    UUID last_opened_user{0, 0};
+};
+using ProfileManagerPtr = std::unique_ptr<ProfileManager>;
```
Subv: You never actually use this alias

src/core/hle/service/acc/acc.cpp
Subv: libnx doesn't seem to have this field, and in the code that i saw during RE, games loop over the data buffer checking which UUIDs are not 0 up to a maximum iteration index to count the total amount of users.

PR #1033, opened 2018-08-12
audio_core: Interpolate

MerryMage:
Let's do some simple interpolation.

Higher quality algorithm is possible in the future (but we've not bothered with that in Citra yet either).

Wrote it using `double`s instead of using fixed-point math like I usually would just to make it simple to understand.


src/audio_core/algorithm/interpolate.cpp
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    const size_t num_frames = input.size() / 2;
+    double position = state.position;
+    std::vector<s16> output;
+    output.reserve(static_cast<size_t>(input.size() / ratio + 4));
+
+    const auto sample = [&](double f, s16 x0l, s16 x0r, s16 x1l, s16 x1r) {
+        // Linear interpolation.
+        double l = x0l * f + x1l * (1 - f);
```
lioncash: `l` and `r` can be const here.

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -0,0 +1,58 @@
...
+        double r = x0r * f + x1r * (1 - f);
+        output.emplace_back(static_cast<s16>(l));
+        output.emplace_back(static_cast<s16>(r));
+    };
+
+    if (position < 0) {
+        const double f = std::fmod(position + 1, 1);
+        sample(f, state.last_frame[0], state.last_frame[1], input[0], input[1]);
+        position = std::max<double>(position + ratio, 0);
```
lioncash: `std::max(position + ratio, 0.0);`

MerryMage:
Implementation has changed quite a bit since last review. Please re-review.

Apologies for also including `f6819e0`. If a separate PR is warranted, I'll do that.

bunnei:
Still LGTM

src/audio_core/algorithm/filter.cpp
```diff
@@ -0,0 +1,79 @@
...
+CascadingFilter CascadingFilter::LowPass(double cutoff, size_t cascade_size) {
+    std::vector<Filter> cascade(cascade_size);
+    for (size_t i = 0; i < cascade_size; i++) {
+        cascade[i] = Filter::LowPass(cutoff, CascadingBiquadQ(cascade_size, i));
+    }
+    return CascadingFilter{cascade};
+}
+
+CascadingFilter::CascadingFilter() = default;
+CascadingFilter::CascadingFilter(const std::vector<Filter>& filters) : filters(filters) {}
```
lioncash: It would likely be better to take by value and std::move in the initializer list, to allow avoiding copies entirely depending on the external calling code.

src/audio_core/algorithm/filter.h
```diff
@@ -0,0 +1,61 @@
...
+#include "common/common_types.h"
+
+namespace AudioCore {
+
+/// Digial biquad filter:
+///
+///          b0 + b1 z^-1 + b2 z^-2
+///  H(z) = ------------------------
+///          a0 + a1 z^-1 + b2 z^-2
+struct Filter {
```
lioncash: Should probably be a class, since it has invariants.

src/audio_core/algorithm/filter.h
```diff
@@ -0,0 +1,61 @@
...
+    /// Coefficients are in normalized form (a0 = 1.0).
+    double a1, a2, b0, b1, b2;
+    /// Input History
+    std::array<std::array<double, channel_count>, 3> in;
+    /// Output History
+    std::array<std::array<double, channel_count>, 3> out;
+};
+
+/// Cascade filters to build up higher-order filters from lower-order ones.
+struct CascadingFilter {
```
lioncash: Ditto

bunnei:
Tested with Cave Story and Binding of Isaac, the interpolation sounds 👌

PR #1035, opened 2018-08-12
GetAudioDeviceServiceWithRevisionInfo

ogniK5377:
As we're not handling any anything about the revision data for GetAudioDeviceServiceWithRevisionInfo, it's currently marked as stubbed. However, for games, this shouldn't affect the result. Proper revision info would be more for homebrew.


mailwl:
as i saw in games, this function calls right after GetAudioDevice(). Probably better create one unique audio device as class var and return it in both functions?

JohnGodgames:
**Bloodstained: Curse of the Moon**
`Service <Error> core\hle\service\service.cpp:ReportUnimplementedFunction:149: unknown / unimplemented function 'GetAudioDeviceServiceWithRevisionInfo': port='audren:u' cmd_buf={[0]=0x4, [1]=0xC, [2]=0x0, [3]=0x0, [4]=0x49434653, [5]=0x0, [6]=0x4, [7]=0x0, [8]=0x34564552}`

ogniK5377:
@JohnGodgames you need to merge the PR :P

JohnGodgames:
@ogniK5377  I know.I just wanted to point out which game needed it :P

PR #1040, opened 2018-08-12
gl_shader_decompiler: Implement XMAD instruction.

bunnei:
Please review very closely - I'm not entirely sure about some stuff here. Tested with SMO, seems to mostly improve things, although not drastically.

References for this impl:
* https://github.com/envytools/envytools/blob/master/docs/hw/graph/maxwell/cuda/int.rst#multiply-add-xmad
* https://github.com/envytools/envytools/blob/master/envydis/gm107.c
* https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics/Gal/Shader/ShaderDecodeAlu.cs#L1070

@Hexagon12 (and others) - Please test with other games that use XMAD and let me know if there are any improvements/regressions. Thanks!

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -200,6 +200,14 @@ enum class IMinMaxExchange : u64 {
     XHi = 3,
 };
 
+enum class XmadMode : u64 {
+    None = 0,
+    CLo = 1,
+    CHi = 2,
+    CSfu = 3,
+    CBcc = 4,
+};
+
 enum class FlowCondition : u64 {
     Always = 0xF,
     Fcsm_Tr = 0x1C, // TODO(bunnei): What is this used for?
@@ -457,6 +465,18 @@ union Instruction {
     } bra;
 
     union {
+        BitField<20, 16, u64> imm20_16;
+        BitField<36, 1, u64> product_shift_left;
+        BitField<37, 1, u64> merge_37;
+        BitField<48, 1, u64> sign_a;
+        BitField<49, 1, u64> sign_b;
+        BitField<50, 3, XmadMode> mode;
+        BitField<52, 1, u64> high_b;
+        BitField<53, 1, u64> high_a;
+        BitField<56, 1, u64> merge_56;
+    } xmad;
+
+    union {
         BitField<20, 14, u64> offset;
         BitField<34, 5, u64> index;
     } cbuf34;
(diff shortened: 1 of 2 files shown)


src/video_core/engines/shader_bytecode.h
```diff
@@ -456,6 +464,18 @@ union Instruction {
...
+    union {
+        BitField<20, 16, u64> imm20_16;
+        BitField<36, 1, u64> product_shift_left;
+        BitField<37, 1, u64> merge_37;
+        BitField<48, 1, u64> sign_a;
+        BitField<49, 1, u64> sign_b;
+        BitField<50, 3, XmadMode> mode;
+        BitField<52, 1, u64> high_b;
+        BitField<53, 1, u64> high_a;
+        BitField<56, 1, u64> merge_56;
```
bunnei: envydis has this in a different bit for XMAD_CR ... no idea if this is right

src/video_core/engines/shader_bytecode.h
```diff
@@ -456,6 +464,18 @@ union Instruction {
         }
     } bra;
 
+    union {
+        BitField<20, 16, u64> imm20_16;
```
bunnei: envydis has this as 19-bit imm starting at bit 20, but that seems wrong - it overlaps with product_shift_left and merge_37 bits. So not 100% on this, but for reference:
```
pendingchaosToday at 4:09 PM
it seems none of the bits are a negate flag
the immediate does seem to start at bit 20
and trying to give an immediate of 65536 wraps around to 0, so I guess it's a 16-bit field
```

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1630,6 +1632,99 @@ class GLSLGenerator {
...
+            default: {
+                LOG_CRITICAL(HW_GPU, "Unhandled XMAD mode: {}",
+                             static_cast<u32>(instr.xmad.mode.Value()));
+                UNREACHABLE();
+            }
+            }
+
+            std::string sum{product + " + " + op_c};
+            if (is_merge) {
+                sum = "((" + sum + " & 0xFFFF) | (" + src2 + "<< 16))";
```
Subv: I think you'll have to wrap `sum` in parentheses to get the correct precedence, same with src2

src/video_core/engines/shader_bytecode.h
Subv: I am okay with this

src/video_core/engines/shader_bytecode.h
Subv: I am also okay with this

PR #1043, opened 2018-08-13
Use an approximated amortized amount of ticks when advancing timing.

Subv:
We divide the number of ticks to add by the number of cores (4) to obtain a more or less rough estimate of the actual number of ticks added. This assumes that all 4 cores are doing similar work. Previously we were adding ~4 times the number of ticks, thus making the games think that time was going way too fast.

This lets us bypass certain hangs in some games like Breath of the Wild.

We should modify our CoreTiming to support multiple cores (both running in a single thread, and in multiple host threads).

BoTW in singlecore now behaves similarly to having multicore on, but hopefully without the race conditions.

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic.cpp
@@ -86,7 +86,16 @@ public:
     }
 
     void AddTicks(u64 ticks) override {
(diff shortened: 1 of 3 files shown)


bunnei:
looks...great...to...me...
image

PR #1046, opened 2018-08-13
Added missing channel devices

ogniK5377:


greggameplayer:
@ogniK5377 can you add the others ioctls inside these ?

PR #1054, opened 2018-08-13
common/misc: use windows.h

zhaowenlan1779:
linux-mingw does not really like this.


lioncash:
This seems like an issue that should also be filed with the toolchain (the actual Windows libraries use `Windows.h`, so mingw should probably be making a symlink for that on Linux platforms).

PR #1064, opened 2018-08-14
common/telemetry: Migrate core-independent info gathering to common

lioncash:
Previously core itself was the library containing the code to gather common information (build info, CPU info, and OS info), however all of this isn't core-dependent and can be moved to the common code and use the common interfaces. We can then just call those functions from the core instead.

This will allow replacing our CPU detection with Xbyak's which has better detection facilities than ours (without needing to make core have Xbyak as a dependency) in a followup. It also keeps more architecture-dependent code in common instead of core.

### DIFF
--- a/src/common/telemetry.cpp
+++ b/src/common/telemetry.cpp
@@ -3,7 +3,14 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <cstring>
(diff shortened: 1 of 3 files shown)


lioncash:
Failure isn't related to the build itself.

PR #1065, opened 2018-08-14
qt: Add filename and title id to window title while running

DarkLordZach:
Pretty self explanatory.

Example:
newtbar


Soikow:
Please merge this

DarkLordZach:
@FearlessTobi Comments addressed. Now displays game name (from metadata) if possible, if not the filename is shown instead. Title ID was removed.

PR #1075, opened 2018-08-15
loader/{nca, xci}: Remove unnecessary includes and unused member variables

lioncash:
Many of these aren't necessary and will cause this file to be required to be recompiled whenever any changes to those files are made, which lengthens compile times for no reason.


lioncash:
Failure is unrelated

PR #1079, opened 2018-08-15
loader: Make ResultStatus directly compatible with fmt

lioncash:
We can make the enum class type compatible with fmt by providing an overload of `operator<<`.

While we're at it, perform proper bounds checking. If something exceeds the array, it should be a hard fail, because it's, without a doubt, a programmer error in this case.


lioncash:
Failure is unrelated

PR #1082, opened 2018-08-15
CoreTiming: Keep Cores synced

B3n30:
Keep Cores synced by having each core it's own downcount and let the main core run an intermediate run if it's slice_length was cropped

This is untested again, and has probably a lot of bugs. So please test and give me feadback

### DIFF
--- a/src/core/core_timing.cpp
+++ b/src/core/core_timing.cpp
@@ -93,7 +93,7 @@ void UnregisterAllEvents() {
 
 void Init() {
     std::unique_lock<std::mutex> lock(mutex);
-    std::fill(downcount.begin(), downcount.end(), MAX_SLICE_LENGTH);
+    downcount.fill(MAX_SLICE_LENGTH);
     slice_length = MAX_SLICE_LENGTH;
     next_slice_length = MAX_SLICE_LENGTH;
     global_timer = 0;
@@ -183,8 +183,7 @@ void ForceExceptionCheck(s64 cycles) {
     if (downcount[0] > cycles) {
         // downcount is always (much) smaller than MAX_INT so we can safely cast cycles to an int
         // here. Account for cycles already executed by adjusting the g.slice_length
-        s64 crop_time = downcount[0] - static_cast<int>(cycles);
-        slice_length -= crop_time;
+        slice_length -= downcount[0] - static_cast<int>(cycles);
         downcount[0] = static_cast<int>(cycles);
         if (next_slice_length <= 0)
             next_slice_length = MAX_SLICE_LENGTH;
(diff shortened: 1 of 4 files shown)


src/core/core_timing.h
```diff
@@ -82,6 +81,8 @@ void MoveEvents();
 /// Pretend that the main CPU has executed enough cycles to reach the next event.
 void Idle();
 
+bool MainSliceWasCropped();
```
lioncash: This should have a documentation comment just like the functions above and below it.

src/core/core_timing.cpp
```diff
@@ -193,11 +200,19 @@ void MoveEvents() {
 }
 
 void Advance() {
+    std::unique_lock<std::mutex> lock(mutex);
+    size_t current_core = Core::System::GetInstance().CurrentCoreIndex();
```
lioncash: `const size_t`

src/core/core_timing.cpp
```diff
@@ -176,11 +179,15 @@ void RemoveNormalAndThreadsafeEvent(const EventType* event_type) {
...
 void ForceExceptionCheck(s64 cycles) {
     cycles = std::max<s64>(0, cycles);
-    if (downcount > cycles) {
+    std::unique_lock<std::mutex> lock(mutex);
+    if (downcount[0] > cycles) {
         // downcount is always (much) smaller than MAX_INT so we can safely cast cycles to an int
         // here. Account for cycles already executed by adjusting the g.slice_length
-        slice_length -= downcount - static_cast<int>(cycles);
-        downcount = static_cast<int>(cycles);
+        s64 crop_time = downcount[0] - static_cast<int>(cycles);
```
lioncash: `const s64`

src/core/core_timing.cpp
```diff
@@ -89,10 +92,11 @@ void UnregisterAllEvents() {
 }
 
 void Init() {
-    downcount = MAX_SLICE_LENGTH;
+    std::unique_lock<std::mutex> lock(mutex);
+    std::fill(downcount.begin(), downcount.end(), MAX_SLICE_LENGTH);
```
lioncash:
```cpp
downcount.fill(MAX_SLICE_LENGTH);
```

B3n30:
All events scheduled from another core then the main core will always have a delay (with max being MAX_SLICE_LENGTH) and I can't see any way to fix this. 

In my opinion this is the best approach to sync multiple cores without destroying performance completely

FernandoS27:
Seems to be crashing as of  4d7e166 master.

B3n30:
Sorry forgot about that PR.
CoreTiming needs a major rewrite to work properly with the multicore implementation of yuzu.

PR #1086, opened 2018-08-16
core/file_sys: Replace includes with forward declarations where applicable

lioncash:
Avoids proliferating includes through other headers, lessening the total amount of files that need to be rebuilt if those headers are changed. This also resolves a few indirect inclusions.


bunnei:
@lioncash can you rebase please

lioncash:
Most of these were done in other PRs I made, so I can just close this one.

PR #1088, opened 2018-08-16
PL:U Added BFTTF loading(Loading from System NAND dumps)

ogniK5377:
We can now load system bfttf fonts from system archives AND shared memory dumps. This allows people who have installed their system nand dumps to yuzu to automatically get shared font support. We also now don't hard code the offsets or the sizes of the shared fonts and it's all calculated for us now.

### DIFF
--- a/src/core/hle/service/ns/pl_u.cpp
+++ b/src/core/hle/service/ns/pl_u.cpp
@@ -47,7 +47,7 @@ std::vector<FontRegion>
     SHARED_FONT_REGIONS{}; // Automatically populated based on shared_fonts dump or system archives
 
 const FontRegion& GetSharedFontRegion(size_t index) {
-    if (index < SHARED_FONT_REGIONS.size() || SHARED_FONT_REGIONS.empty()) {
+    if (index >= SHARED_FONT_REGIONS.size() || SHARED_FONT_REGIONS.empty()) {
         // No font fallback
         return EMPTY_REGION;
     }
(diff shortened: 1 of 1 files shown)


src/core/hle/service/ns/pl_u.cpp
```diff
@@ -41,25 +104,68 @@ PL_U::PL_U() : ServiceFramework("pl:u") {
...
+    if (nand->HasEntry(
+            static_cast<u64>(FontArchives::Standard), // Rebuild shared fonts from data ncas
+            FileSys::ContentRecordType::Data)) {
+        size_t offset = 0;
+        shared_font = std::make_shared<std::vector<u8>>(SHARED_FONT_MEM_SIZE);
+        for (auto font : SHARED_FONTS) {
+            const auto nca =
+                nand->GetEntry(static_cast<u64>(font.first), FileSys::ContentRecordType::Data);
+            if (!nca) {
+                LOG_CRITICAL(Service_NS, "Failed to find {:016X}! Skipping",
```
lioncash: These should probably just be regular error logs. Critical should be for behavior where the application is very likely going to crash (or be unstable).

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -40,26 +103,72 @@ PL_U::PL_U() : ServiceFramework("pl:u") {
...
-    FileUtil::IOFile file(filepath, "rb");
-
-    shared_font = std::make_shared<std::vector<u8>>(SHARED_FONT_MEM_SIZE);
-    if (file.IsOpen()) {
-        // Read shared font data
-        ASSERT(file.GetSize() == SHARED_FONT_MEM_SIZE);
-        file.ReadBytes(shared_font->data(), shared_font->size());
+    const auto nand = FileSystem::GetSystemNANDContents();
+    if (nand->HasEntry(
+            static_cast<u64>(FontArchives::Standard), // Rebuild shared fonts from data ncas
```
DarkLordZach: What you have is fine, however you may want to consider something like `const auto nca = nand->GetEntry()` and instead of HasEntry check if nca is nullptr. This is because HasEntry just does `GetEntry() != nullptr`. Sorry for not mentioning this earlier.

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -40,26 +103,72 @@ PL_U::PL_U() : ServiceFramework("pl:u") {
...
+        shared_font = std::make_shared<std::vector<u8>>(SHARED_FONT_MEM_SIZE);
+        for (auto font : SHARED_FONTS) {
+            const auto nca =
+                nand->GetEntry(static_cast<u64>(font.first), FileSys::ContentRecordType::Data);
+            if (!nca) {
+                LOG_ERROR(Service_NS, "Failed to find {:016X}! Skipping",
+                          static_cast<u64>(font.first));
+                continue;
+            }
+            const auto romfs = FileSys::ExtractRomFS(nca->GetRomFS());
```
DarkLordZach: If decryption fails or the nca is malformed, `nca->GetRomFS()` may return nullptr. ExtractRomFS doesn't check for this and will likely crash.

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -40,26 +103,72 @@ PL_U::PL_U() : ServiceFramework("pl:u") {
...
+                continue;
+            }
+            const auto font_fp = romfs->GetFile(font.second);
+            if (!romfs) {
+                LOG_ERROR(Service_NS, "{:016X} has no file \"{}\"! Skipping",
+                          static_cast<u64>(font.first), font.second);
+                continue;
+            }
+            const auto font_data = font_fp->ReadAllBytes();
+            std::vector<u32> font_data_u32(font_data.size() / sizeof(u32));
```
DarkLordZach: You should be able to read directly into this using the overload of Read/ReadBytes that takes a pointer (on the font file)

PR #1089, opened 2018-08-16
Shaders: Corrected the 'abs' and 'neg' bit usage in the float arithmetic instructions.

Subv:
We should definitely audit our shader generator for more errors like this.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -741,6 +741,30 @@ private:
         return op->second;
     }
 
+    /**
+     * Transforms the input string GLSL operand into one that applies the abs() function and negates
+     * the output if necessary. When both abs and neg are true, the negation will be applied after
+     * taking the absolute value.
+     * @param operand The input operand to take the abs() of, negate, or both.
+     * @param abs Whether to apply the abs() function to the input operand.
+     * @param neg Whether to negate the input operand.
+     * @returns String corresponding to the operand after being transformed by the abs() and
+     * negation operations.
+     */
+    static std::string GetOperandAbsNeg(const std::string& operand, bool abs, bool neg) {
+        std::string result = operand;
+
+        if (abs) {
+            result = "abs(" + result + ')';
+        }
+
+        if (neg) {
+            result = "-(" + result + ')';
+        }
+
+        return result;
+    }
+
     /*
      * Returns whether the instruction at the specified offset is a 'sched' instruction.
      * Sched instructions always appear before a sequence of 3 instructions.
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -741,6 +741,30 @@ class GLSLGenerator {
...
+     * Transforms the input string GLSL operand into one that applies the abs() function and negates
+     * the output if necessary. When both abs and neg are true, the negation will be applied after
+     * taking the absolute value.
+     * @param operand The input operand to take the abs() of, negate, or both.
+     * @param abs Whether to apply the abs() function to the input operand.
+     * @param neg Whether to negate the input operand.
+     * @returns String corresponding to the operand after being transformed by the abs() and
+     * negation operations.
+     */
+    std::string GetOperandAbsNeg(const std::string& operand, bool abs, bool neg) {
```
lioncash: This can be a static function.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -878,36 +895,35 @@ class GLSLGenerator {
...
+                // MOV does not have neither 'abs' nor 'neg' bits.
                 regs.SetRegisterToFloat(instr.gpr0, 0, op_b, 1, 1);
                 break;
             }
 
             case OpCode::Id::FMUL_C:
             case OpCode::Id::FMUL_R:
             case OpCode::Id::FMUL_IMM: {
+                // FMUL does not have 'abs' bits and only the second operand has a 'neg' bit.
+                op_b = GetOperandAbsNeg(op_b, false, instr.alu.negate_a);
```
bunnei: Does negate_a here actually mean negate_b? Or is this wrong? (if it's not wrong, we need a different alias for this bit)

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -878,36 +895,35 @@ class GLSLGenerator {
...
             case OpCode::Id::FADD_IMM: {
+                op_a = GetOperandAbsNeg(op_a, instr.alu.abs_a, instr.alu.negate_a);
+                op_b = GetOperandAbsNeg(op_b, instr.alu.abs_b, instr.alu.negate_b);
                 regs.SetRegisterToFloat(instr.gpr0, 0, op_a + " + " + op_b, 1, 1,
                                         instr.alu.saturate_d);
                 break;
             }
             case OpCode::Id::MUFU: {
+                // MUFU only has 'abs' and 'neg' bits for the second operand.
+                op_b = GetOperandAbsNeg(op_b, instr.alu.abs_a, instr.alu.negate_a);
```
bunnei: Again, should we alias these differently so the code does not look wrong?

Subv:
Updated

PR #1095, opened 2018-08-16
filesystem: Add support for loading of system archives

DarkLordZach:
Mainly:
- Implement `OpenDataStorageByDataId`, which is the primary accessor for system archives. 
- Rewrite `RomFSFactory` to make opening anything besides the current romfs easier (using the NAND/SDMC registered caches)

To install system archives, you can use the `Install to NAND` menu option if you have the NCA, just make sure you change the type to `System Archive`. Alternatively, you can copy the `Contents` folder from the `SYSTEM` partition of a nand dump and paste it in `%YUZU_DIR%/nand/system`. You'll need keys (obv), but this is easier and covers all system archives (shared fonts included for #1088). 

This also gets Ultra Street Fighter II booting:
stretfirghter2
boww


src/core/file_sys/registered_cache.cpp
```diff
@@ -77,12 +77,13 @@ static ContentRecordType GetCRTypeFromNCAType(NCAContentType type) {
     case NCAContentType::Control:
         return ContentRecordType::Control;
     case NCAContentType::Data:
+    case static_cast<NCAContentType>(0x05): ///< Seems to be used on some system archives
```
bunnei: This is weird... can we just add 0x05 to the enum? E.g. `Data_Unknown5` or something? (and move the comment there)

src/core/file_sys/romfs_factory.cpp
```diff
@@ -17,9 +19,41 @@ RomFSFactory::RomFSFactory(Loader::AppLoader& app_loader) {
...
+        }
+        const auto romfs = res->GetRomFS();
+        if (romfs == nullptr) {
+            // TODO(DarkLordZach): Find the right error code to use here
+            return ResultCode(-1);
+        }
+        return MakeResult<VirtualFile>(romfs);
+    }
+    default:
+        UNIMPLEMENTED_MSG("Unimplmented storage_id={:02X}", static_cast<u8>(storage));
```
bunnei: nit: Spelled wrong

PR #1106, opened 2018-08-19
Shaders: Write all the enabled color outputs when a fragment shader exits.

Subv:
We were only writing to the first render target before.
Note that this is only the GLSL side of the implementation, supporting multiple render targets requires more changes in the OpenGL renderer.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -26,6 +26,7 @@ using Tegra::Shader::Sampler;
 using Tegra::Shader::SubOp;
 
 constexpr u32 PROGRAM_END = MAX_PROGRAM_CODE_LENGTH;
+constexpr u32 PROGRAM_HEADER_SIZE = 0x50;
 
 class DecompileFail : public std::runtime_error {
 public:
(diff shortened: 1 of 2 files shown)


Hexagon12:
Please rebase this PR

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -815,6 +828,32 @@ class GLSLGenerator {
...
+        ASSERT_MSG(header.writes_samplemask == 0, "Samplemask write is unimplemented");
+
+        // Write the color outputs using the data in the shader registers, disabled
+        // rendertargets/components are skipped in the register assignment.
+        u32 current_reg = 0;
+        for (u32 render_target = 0; render_target < Maxwell3D::Regs::NumRenderTargets;
+             ++render_target) {
+            for (u32 component = 0; component < 4; ++component) {
+                u32 bit = render_target * 4 + component;
+                if (header.enabled_color_outputs & (1 << bit)) {
```
bunnei: Nit: Move this check to `FragmentHeader`, and add a helper function:
<pre>
bool IsColorChannelEnabled(size_t render_target, size_t component) const {
    // ..
}
</pre>
IMO more readable and we do this type of pattern elsewhere [move helper functions like this to the class rather than inlining the code].

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -815,6 +828,32 @@ class GLSLGenerator {
...
+                    shader.AddLine("color[" + std::to_string(render_target) + "][" +
```
bunnei: nit: use `fmt::format`, its more readable IMO

PR #1111, opened 2018-08-19
Patch the new Dynarmic's setup for GDBStub

Hedges:


MerryMage:
What is this intended to do?

Hedges:
This is to keep Dynarmic falling back to interpreter for unpredictable instructions as it used to - only when GDBStub is active - which is what we use for s/w breakpoints. I will have a look at a better solution like using exceptions, but this seems like a good patch for now.

MerryMage:
The reason I am doubtful this helps is because the flag doesn't affect all unpredictable instructions; it currently (should) only affect STXR, STLXR, STXP and STLXP. The relevant commit is [here](https://github.com/MerryMage/dynarmic/commit/550d662f0ece868209f0f760dddc8b2ad4698b17).

If this flag does anything else, this is likely either an error in my logic, in dynarmic, or in my memory, and in any case I would prefer for that to be fixed properly, if possible.

It is for the above reasons that this smells like a hack to me, since it shouldn't be necessary.

Please help me understand why this is necessary.

Hedges:
Closing as unnecessary.

PR #1114, opened 2018-08-19
vulkan: Initial implementation

ReinUsesLisp:
It does the following:
* Split EmuWindow for SDL2 into an OpenGL and Vulkan backend
* Add an SDL2 config entry for graphics backend (0: OpenGL, 1: Vulkan)
* Create a Vulkan instance, device and swapchain
* Present to the screen homebrew's framebuffer pixels

It doesn't:
* Implement any of the rasterizer functionality
* Implement a memory allocator
* Choose a device and test for its features (it just chooses the first one)
* Implement the Qt frontend

It's disabled by default in cmake (I think it would trick appveyor otherwise).
I'm not the C++ kind of guy (I'm more used to C). Anything that you think is wrong (code-wise or in CMake files), please let me know and I'll try to fix it.


ReaperOfSouls1909:
our lord and savior Rein

EmulationChannel:
Yuzu Suport Vulkan API
Oh my God 🥳

src/yuzu_cmd/emu_window/emu_window_sdl2_vk.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include <vulkan/vulkan.h>
+#include "common/assert.h"
+#include "common/logging/log.h"
+#include "common/scm_rev.h"
+#include "yuzu_cmd/emu_window/emu_window_sdl2_vk.h"
+
+static VkBool32 DebugCallback(VkDebugReportFlagsEXT flags, VkDebugReportObjectTypeEXT object_type,
+                              uint64_t object, size_t location, int32_t message_code,
+                              const char* layer_prefix, const char* message, void* user_data) {
+    switch (flags) {
```
lioncash: Given these are flags, these should probably be tested for with bitwise AND, considering multiple can be used together, rather than with a switch statement.

This should likely also check if  the `VK_DEBUG_REPORT_PERFORMANCE_WARNING_BIT_EXT` bit is set and print out to the debug log. Given it reports potential suboptimal use of Vulkan.

`VK_DEBUG_REPORT_INFORMATION_BIT_EXT` should likely be checked for and pushed to the debug log or trace log.

`VK_DEBUG_REPORT_DEBUG_BIT_EXT` should likely also be checked for and have its messages be pushed to the debug log or trace log.

src/yuzu_cmd/emu_window/emu_window_sdl2_vk.cpp
```diff
@@ -0,0 +1,146 @@
...
+    std::string window_title = fmt::format("yuzu {} | {}-{} (Vulkan)", Common::g_build_name,
+                                           Common::g_scm_branch, Common::g_scm_desc);
+    render_window =
+        SDL_CreateWindow(window_title.c_str(),
+                         SDL_WINDOWPOS_UNDEFINED, // x position
+                         SDL_WINDOWPOS_UNDEFINED, // y position
+                         Layout::ScreenUndocked::Width, Layout::ScreenUndocked::Height,
+                         SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI | SDL_WINDOW_VULKAN);
+
+    std::vector<const char*> enabled_extensions{};
```
lioncash: These can be moved farther down to where they're actually used. They also don't need the `{}` initialized. This is done automatically, since their constructor is defined to initialize the container.

src/video_core/renderer_vulkan/vk_swapchain.cpp
```diff
@@ -0,0 +1,326 @@
...
+}
+
+VkImage VulkanSwapchain::GetImage(u32 image_index) const {
+    return images[image_index];
+}
+
+VkFence VulkanSwapchain::GetFence(u32 image_index) const {
+    return fences[image_index];
+}
+VkRenderPass VulkanSwapchain::GetRenderPass() const {
```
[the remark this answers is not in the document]
ReinUsesLisp: They are declared in `vulkan.h` as a pointer to an opaque structure, that's why I prefer returning them by value.

Wunkolo:
Have you guys considered [vulkan.hpp](https://github.com/KhronosGroup/Vulkan-Hpp) to try and tuck away some of the little vulkan-headaches.

ReinUsesLisp:
Addressed most of the feedback. There's still the `void*` passed to the rasterizer interface (it's there because ScreenInfo was an OpenGL-specific structure). What should I do about that one? I could move all the framebuffer code (in OpenGL and Vulkan) to the rasterizer, but doesn't really belong there.

@Wunkolo I never thought objectively about why not to use `vulkan.hpp` (since I've always used the C API). That depends on what the maintainers think about it.

ReinUsesLisp:
Addressed feedback and ported it to `vulkan.hpp`. There's no need for depending on `vulkan.hpp`'s repo since it's already included in LunarG's SDK.
I used the exception based API catching error results manually on initialization.
New things might be wrong code-wise because I had to rewrite most of the Vulkan code.

iOS4all:
Just a random question. Can we get this vulkan and make it running on Nintendo Switch itself as a part of hombrew? To take advantage of it in terms of emulators.
Thanks.

ReinUsesLisp:
Rebased and updated. I didn't add any significant change but there are some improvements, like silencing validation layer warnings and the usage of unique pointers from `vulkan.hpp`.

ReinUsesLisp:
This a very old base and would require non-trivial work to update without adding extra code.

PR #1121, opened 2018-08-20
Rasterizer: Use PBOs to reinterpret texture formats when games re-use the same memory.

Subv:
We were previously using a framebuffer blit, which internally converts the formats, instead of using the raw data and interpreting it as the new format.

This shouldn't have much impact on anything since most reinterpretations happen because games stopped using a texture and re-used the memory for another unrelated one that they're about to clear.


bunnei:
Tested a bunch, seems fine.

Issue #1116, opened 2018-08-20
Error in last canary build


zemzema:
Canary Build - 16943fb 

./yuzu
qt5ct: using qt5ct plugin
qt5ct: D-Bus global menu: no
Qt Concurrent has caught an exception thrown from a worker thread.
This is not supported, exceptions thrown in worker threads must be
caught before control returns to Qt Concurrent.
terminate called after throwing an instance of 'std::logic_error'
  what():  Invalid hex digit
Aborted (core dumped)

DarkLordZach:
This is an issue where the key file parser will forcibly fail when the key is not 32 hex digits long. It's a relic from testing. It should be fixed soon, but in the meantime make sure all of your keys and rights IDs are the correct length

zemzema:
Checked everything, problem still exist. This is the first time that it's not working.

Drive251:
@zemzema , it seem it was fixed

DarkLordZach:
@zemzema I am confident that you have either: Incorrect length keys or a file in yuzu's nand not named properly. That is the only way to generate that error. The PR only prevents yuzu from hard crashing due to this.

PR #1128, opened 2018-08-21
hex_util: Replace logic_errors with LOG_CRITICAL

DarkLordZach:
Makes it so malformed hex strings do not crash the entire program.

While this does make it so yuzu won't hard crash on malformed hex (which is bad), if you see the log spammed with `ToHexNibble` and `Not of correct size`, you should investigate the keys as that's probably why games aren't booting.

Address issue quasi-outlined in #1116. While I am still confident that the keys are of incorrect length, this should at least allow yuzu to start and maybe provide insight into specifically what's wrong.

### DIFF
--- a/src/common/hex_util.cpp
+++ b/src/common/hex_util.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include "common/hex_util.h"
+#include "common/logging/log.h"
 
 namespace Common {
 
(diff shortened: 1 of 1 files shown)


lioncash:
These should be `LOG_ERROR`. `LOG_CRITICAL` is for situations where the application is unstable or likely to crash.

src/common/hex_util.cpp
```diff
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include "common/hex_util.h"
+#include "logging/log.h"
```
lioncash: `common/logging/log.h`. Includes should always use the full relative path from the `src` directory.

src/common/hex_util.cpp
```diff
@@ -13,18 +14,23 @@ u8 ToHexNibble(char c1) {
...
-    throw std::logic_error("Invalid hex digit");
+    LOG_CRITICAL(Common, "Invalid hex digit: 0x{:02X}", c1);
+    return 0;
 }
 
 std::array<u8, 16> operator""_array16(const char* str, size_t len) {
-    if (len != 32)
-        throw std::logic_error("Not of correct size.");
+    if (len != 32) {
+        LOG_CRITICAL(Common, "Not of correct size (expected={d}, actual={d}).", 32, len);
```
lioncash: This should indicate what specifically is not of the correct size. Also why not just `{}` for the formatting specifiers? (the 32 can be made part of the string directly, since it's already known what you want to print)

src/common/hex_util.cpp
```diff
@@ -13,18 +14,23 @@ u8 ToHexNibble(char c1) {
...
+        return {};
+    }
     return HexStringToArray<16>(str);
 }
 
 std::array<u8, 32> operator""_array32(const char* str, size_t len) {
-    if (len != 64)
-        throw std::logic_error("Not of correct size.");
+    if (len != 64) {
+        LOG_CRITICAL(Common, "Not of correct size (expected={d}, actual={d}).", 64, len);
```
lioncash: Ditto. (64 can also be made part of the string directly).

DarkLordZach:
@bunnei just saw, will do

DarkLordZach:
Done

PR #1135, opened 2018-08-21
gl_rasterizer_cache: Use CPU address for caching surfaces, and misc. changes.

bunnei:
* Caches rasterizer surfaces based on their true address, not GPU virtual address.
* Adds support for compressed surface copies. This probably shouldn't be used, by I accidentally added it and it seems to work.
* Removes assert for color <-> depth copies, since this should work now.

Unfortunately, I believe these changes introduce a bit of a performance hit. There are probably more superfluous surface copies happening now, such as when a region of memory is reused as a render target and then cleared.

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_gpu.h
+++ b/src/core/hle/service/nvdrv/devices/nvhost_gpu.h
@@ -10,6 +10,7 @@
 #include "common/common_types.h"
 #include "common/swap.h"
 #include "core/hle/service/nvdrv/devices/nvdevice.h"
+#include "video_core/memory_manager.h"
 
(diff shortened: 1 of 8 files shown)


degasus:
I wonder about the motivation of this patch. The GPU ressource caches should have a high cache hit rate, so translating on every access is just another overhead. It is even worse: Most writes to GPU ressources are from the GPU itself in Maxwell3D::ProcessCBData.

Which kind of mappings do exist in general:
- Are there ressources, which are split in main memory but continious in GPU addresses?
- Are there ressources, which are mapped twice to the GPU and used with both addresses?

PR #1138, opened 2018-08-21
config: Removed setting icon config

tech4me:
This config update causes not ideal behavior. i.e. If your first added a directory to the gamelist(which will have an icon size of 0) thereafter the icon size will remain to be zero until the config file is manually changed again. (Even if a proper game dump is imported into NAND)


DarkLordZach:
This has the side effect of the game list configuration box being unable to change the icon size. Thanks for bringing this issue to my attention though, I'll look into it later.

tech4me:
@DarkLordZach ok, If there is a better way then this can be closed.

DarkLordZach:
@tech4me Sorry if I came off as rude, but this is definitely a larger issue than that. If you want to contact me in the discord, I could give you some tips on how to solve this. (Or if you want I could do it.)

PR #1142, opened 2018-08-21
Port #4013 from Citra: "Init logging sooner so we dont miss some logs on startup"

FearlessTobi:
See https://github.com/citra-emu/citra/pull/4013 for more details.


jroweboy:
This is wrong, InitLogging is a static function defined here: https://github.com/yuzu-emu/yuzu/pull/1142/files#diff-41128cb4d8543fc3dc355acefa28cd61R1101 hence why its not compiling

FearlessTobi:
Compilation is fixed now. PR should be good to merge.

PR #1150, opened 2018-08-21
Swap "Plus" with "Minus" on the controller GUI

literalmente-game:
Major fix /s requested by Hydro


FearlessTobi:
The layouts need to be renamed as well.

PR #1157, opened 2018-08-22
 gl_shader_gen: Use a std::vector to represent program code instead of std::array

lioncash:
While convenient as a std::array, it's also quite a large set of data as well (32KB). It being an array also means data cannot be std::moved. Any situation where the code is being set or relocated means that a full copy of that 32KB data must be done.

If we use a std::vector we do need to allocate on the heap, however, it does allow us to std::move the data we have within the std::vector into another std::vector instance, eliminating the need to always copy the program data (as std::move in this case would just transfer the pointers and bare necessities over to the new vector instance).

That, and we aren't just tossing 32KB (64KB in the case of declaring a ShaderSetup instance) onto the stack at one time.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_gen.h
+++ b/src/video_core/renderer_opengl/gl_shader_gen.h
@@ -115,7 +115,7 @@ struct ShaderEntries {
 using ProgramResult = std::pair<std::string, ShaderEntries>;
 
 struct ShaderSetup {
-    ShaderSetup(ProgramCode program_code) {
+    explicit ShaderSetup(ProgramCode program_code) {
         program.code = std::move(program_code);
     }
 
--- a/src/video_core/renderer_opengl/gl_shader_gen.h
+++ b/src/video_core/renderer_opengl/gl_shader_gen.h
@@ -9,14 +9,14 @@
 #include <type_traits>
 #include <utility>
 #include <vector>
+#include <boost/functional/hash.hpp>
 #include "common/common_types.h"
 #include "common/hash.h"
 
 namespace GLShader {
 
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_gen.h
```diff
@@ -146,13 +146,18 @@ struct ShaderSetup {
 
 private:
     u64 GetNewHash() const {
+        size_t hash = 0;
+
+        const u64 hash_a = Common::ComputeHash64(program.code.data(), program.code.size());
+        boost::hash_combine(hash, hash_a);
```
jroweboy: does this do anything useful? couldn't we just `size_t hash = static_cast<size_t>(Common::ComputeHash64(program.code.data(), program.code.size()));` and then hash_combine only if has_program_b?

lioncash: Yes, it keeps the combined value uniform algorithmically by applying the combining algorithm to all data involved. We want that algorithm applied uniformly across all data (which is also why all of the examples on [boost's](https://www.boost.org/doc/libs/1_68_0/doc/html/hash/combine.html) page explaining hash_combine are set up that way).

e.g. Currently in 1.68.0, hash_combine does the following:

```cpp
inline void hash_combine_impl(boost::uint64_t& h, boost::uint64_t k)
{
    const boost::uint64_t m = UINT64_C(0xc6a4a7935bd1e995);
    const int r = 47;

    k *= m;
    k ^= k >> r;
    k *= m;

    h ^= k;
    h *= m;

    // Completely arbitrary number, to prevent 0's
    // from hashing to 0.
    h += 0xe6546b64;
}
```

PR #1160, opened 2018-08-22
gl_rasterizer_cache: Several improvements

bunnei:
Several improvements to the rasterizer cache, mostly to improve performance.

* The biggest change here is to "remember" previously used unique surfaces (a unique surface is defined by unique set of address+format+type+width+height+etc.). This avoids scenarios where surfaces are constantly created/destroyed because they are being reused with different parameters. Note, there is still only one "active" surface at a time for a given cached address.
* Uses BlitTexture instead of PBOs to copy surfaces when the format is the same, I assume this is a little bit faster (please correct if I am wrong).
* Implements compressed texture copies in RecreateSurface. This should be mostly unused, but I happened to implement it by mistake.

Suggest reviewing commit by commit. I noticed a slight perf increase in SMO (couple fps in game), not sure how this translates to other titles - Please test. CC @degasus if I'm doing anything wrong/unnecessary here.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -800,8 +800,6 @@ Surface RasterizerCacheOpenGL::RecreateSurface(const Surface& surface,
     // Verify surface is compatible for blitting
     const auto& params{surface->GetSurfaceParams()};
     ASSERT(params.type == new_params.type);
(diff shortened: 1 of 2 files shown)


degasus:
+1 for the surface cache. Freeing ressources is a terrible slow path in drivers, so caching them is very needed.

And indeed, blitting is *by far* faster than abusing a PBO.

But honestly, I'm not familiar with this code, so I doubt I'm able to review it well...

PR #1161, opened 2018-08-23
Add precompiled header support with cotire

jroweboy:
In order to speed up local compile times, cotire will automagically create precompiled headers for each of the build targets. Precompiled headers will work in all major supported compilers (MSVS, Xcode, and GCC that i know of) and should produce some minor recompilation benefits for development. 

I also took the liberty to fix a few annoying useless messages that were in the logs.

For more info about cotire see https://github.com/sakra/cotire

### DIFF
--- a/src/core/file_sys/errors.h
+++ b/src/core/file_sys/errors.h
@@ -17,6 +17,15 @@ enum {
 };
 }
 
+// When using mingw precompiled headers, it pulls in some dark corners of the stdlib that defines
+// these as macros
+#ifdef ERROR_PATH_NOT_FOUND
+#undef ERROR_PATH_NOT_FOUND
+#endif
+#ifdef ERROR_FILE_NOT_FOUND
+#undef ERROR_FILE_NOT_FOUND
+#endif
+
 constexpr ResultCode ERROR_PATH_NOT_FOUND(ErrorModule::FS, ErrCodes::NotFound);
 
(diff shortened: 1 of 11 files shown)


src/core/file_sys/errors.h
```diff
@@ -17,6 +17,15 @@ enum {
 };
 }
 
+// When using mingw precompiled headers, it pulls in some dark corners of the stdlib that defines
+// these as macros
+#ifdef ERROR_PATH_NOT_FOUND
```
lioncash: It may be worth changing the identifiers to `ERROR_PATH_CANNOT_BE_FOUND` and `ERROR_FILE_CANNOT_BE_FOUND`, in place of these undefs.

src/core/loader/linker.cpp
```diff
@@ -12,6 +12,11 @@
 
 namespace Loader {
 
+// When using mingw precompiled headers, it pulls in some dark corners of the stdlib that defines
+// RELATIVE as a macro
+#ifdef RELATIVE
```
lioncash: Enums should be using `SentenceCase` where applicable, so it may be more worthwhile changing to that than undefing stuff we don't control.

Hexagon12:
Please rebase the PR

jroweboy:
This didn't really bring much to the table, so its probably not worth the maintenance headache either.

PR #1165, opened 2018-08-23
renderer_opengl: Implement a new shader cache.

bunnei:
This rewrites our shader cache to be a bit more efficient. Summary of changes:
* Refactor the existing rasterizer cache/invalidation code out of the surface cache, so we can reuse it to cache and invalidate things other than textures and framebuffers.
* Use this to cache shaders based on address. This avoids having to memcpy and hash a 32Kb shader for each shader stage every draw call.
* Caches the results of `glGetProgramResourceIndex` and `glGetUniformLocation` so we do not need to call them every stage of every draw. 

With my testing, there seems to be about a 2X speedup in SMO ingame (so still single digit/low double digit framerates). Please test thoroughly - I'm interested to know of any perf or accuracy regressions, or if this even helps a lot in other games.

Thanks @degasus for the tips for this approach.

### DIFF
--- a/src/video_core/rasterizer_cache.h
+++ b/src/video_core/rasterizer_cache.h
@@ -1,0 +1,116 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <unordered_map>
+#include <boost/icl/interval_map.hpp>
+#include <boost/range/iterator_range.hpp>
+
+#include "common/common_types.h"
+#include "core/memory.h"
+#include "video_core/memory_manager.h"
+
+template <class T>
+class RasterizerCache : NonCopyable {
+public:
+    /// Mark the specified region as being invalidated
+    void InvalidateRegion(Tegra::GPUVAddr region_addr, size_t region_size) {
+        for (auto iter = cached_objects.cbegin(); iter != cached_objects.cend();) {
+            const auto& object{iter->second};
+
+            ++iter;
+
+            if (object->GetAddr() <= (region_addr + region_size) &&
+                region_addr <= (object->GetAddr() + object->GetSizeInBytes())) {
+                // Regions overlap, so invalidate
+                Unregister(object);
+            }
+        }
+    }
+
+protected:
+    /// Tries to get an object from the cache with the specified address
+    T TryGet(Tegra::GPUVAddr addr) const {
+        const auto& search{cached_objects.find(addr)};
+        if (search != cached_objects.end()) {
+            return search->second;
+        }
+
+        return nullptr;
+    }
+
+    /// Gets a reference to the cache
+    const std::unordered_map<Tegra::GPUVAddr, T>& GetCache() const {
+        return cached_objects;
+    }
+
+    /// Register an object into the cache
+    void Register(const T& object) {
+        const auto& search{cached_objects.find(object->GetAddr())};
+        if (search != cached_objects.end()) {
+            // Registered already
+            return;
+        }
+
+        cached_objects[object->GetAddr()] = object;
+        UpdatePagesCachedCount(object->GetAddr(), object->GetSizeInBytes(), 1);
+    }
+
+    /// Unregisters an object from the cache
+    void Unregister(const T& object) {
+        const auto& search{cached_objects.find(object->GetAddr())};
+        if (search == cached_objects.end()) {
+            // Unregistered already
+            return;
+        }
+
+        UpdatePagesCachedCount(object->GetAddr(), object->GetSizeInBytes(), -1);
+        cached_objects.erase(search);
+    }
+
+private:
+    using PageMap = boost::icl::interval_map<u64, int>;
+
+    template <typename Map, typename Interval>
+    constexpr auto RangeFromInterval(Map& map, const Interval& interval) {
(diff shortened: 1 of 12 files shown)


src/video_core/rasterizer_cache.h
```diff
@@ -0,0 +1,118 @@
...
+#include "common/common_types.h"
+#include "core/memory.h"
+#include "video_core/memory_manager.h"
+
+template <class T>
+class RasterizerCache : NonCopyable {
+public:
+    /// Mark the specified region as being invalidated
+    void InvalidateRegion(Tegra::GPUVAddr region_addr, size_t region_size) {
+        for (auto iter = cached_objects.cbegin(); iter != cached_objects.cend();) {
```
degasus: Good for now, but we should keep in mind to get something in n * log n instead of n².

src/video_core/rasterizer_cache.h
```diff
@@ -0,0 +1,118 @@
...
+
+protected:
+    /// Tries to get an object from the cache with the specified address
+    T TryGet(Tegra::GPUVAddr addr) const {
+        const auto& search{cached_objects.find(addr)};
+        if (search != cached_objects.end()) {
+            return search->second;
+        }
+
+        return {};
```
degasus: Do you think it is better to return a nullptr instead? Else it might be hard to differ an entry at 0x0 with no entry.

src/video_core/rasterizer_cache.h
```diff
@@ -0,0 +1,118 @@
...
+    constexpr auto RangeFromInterval(Map& map, const Interval& interval) {
+        return boost::make_iterator_range(map.equal_range(interval));
+    }
+
+    /// Increase/decrease the number of object in pages touching the specified region
+    void UpdatePagesCachedCount(Tegra::GPUVAddr addr, u64 size, int delta) {
+        const u64 num_pages = ((addr + size - 1) >> Tegra::MemoryManager::PAGE_BITS) -
+                              (addr >> Tegra::MemoryManager::PAGE_BITS) + 1;
+        const u64 page_start = addr >> Tegra::MemoryManager::PAGE_BITS;
+        const u64 page_end = page_start + num_pages;
```
degasus: This is the only user of num_pages. So I'd suggest to calculate page_start and page_end and to skip num_pages.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -719,23 +688,20 @@ std::tuple<u8*, GLintptr, u32> RasterizerOpenGL::SetupConstBuffers(
...
 
         // Now configure the bindpoint of the buffer inside the shader
-        const std::string buffer_name = used_buffer.GetName();
-        const GLuint index =
-            glGetProgramResourceIndex(program, GL_UNIFORM_BLOCK, buffer_name.c_str());
-        if (index != GL_INVALID_INDEX) {
-            glUniformBlockBinding(program, index, current_bindpoint + bindpoint);
-        }
+        glUniformBlockBinding(shader->GetProgramHandle(),
+                              shader->GetProgramResourceIndex(used_buffer.GetName()),
```
degasus: Is there any reason why you indix the ressource index map with the buffer name? IMO some integers will fit better here than a std::string

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -745,12 +711,9 @@ u32 RasterizerOpenGL::SetupTextures(Maxwell::ShaderStage stage, GLuint program,
         u32 current_bindpoint = current_unit + bindpoint;
 
         // Bind the uniform to the sampler.
-        GLint uniform = glGetUniformLocation(program, entry.GetName().c_str());
-        if (uniform == -1) {
-            continue;
-        }
 
-        glProgramUniform1i(program, uniform, current_bindpoint);
+        glProgramUniform1i(shader->GetProgramHandle(), shader->GetUniformLocation(entry.GetName()),
```
degasus: Same here, the uniform location shall not be indexed by a string.

src/video_core/rasterizer_cache.h
```diff
@@ -0,0 +1,118 @@
...
+                region_addr <= (object->GetAddr() + object->GetSizeInBytes())) {
+                // Regions overlap, so invalidate
+                Unregister(object);
+            }
+        }
+    }
+
+protected:
+    /// Tries to get an object from the cache with the specified address
+    T TryGet(Tegra::GPUVAddr addr) const {
```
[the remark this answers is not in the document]
bunnei: We will be, but I'm still tracking down some issues with that so for now it will be a separate change. It matters less for shaders, fwiw

src/video_core/renderer_opengl/gl_rasterizer.cpp
bunnei: Right now, this is how our shader management works - everything is referenced via string. This would have been a bigger change, so I thought it made sense for it to be a different one (although I'm not entirely sure it's necessary)

src/video_core/renderer_opengl/gl_rasterizer.cpp
bunnei: Right now, this is how our shader management works - everything is referenced via string. This would have been a bigger change, so I thought it made sense for it to be a different one (although I'm not entirely sure it's necessary)

PR #1178, opened 2018-08-25
file_sys: Add Nintendo Submissions Package (NSP) file format

DarkLordZach:
**NOTE**: This is *not* to satisfy pirates or anything like that. This is because of the new SD to NSP dumper tool that has appeared and the fact that NSP will likely become the format of choice for advanced dumper homebrew (easy to use and contains lots of info).

- Supports both types of NSPs, which I am dubbing:
    - `Extracted Type`: Contains a romfs, main executable and other library files needed by the game. Bootable, but no icons or text names.
    - `Not Extracted Type`: Contains a collection of NCAs, at least one that is Meta-type describing what other NCAs are in the package. Also optionally contains a ticket (if the game uses titlekey crypto).

- Replace `secure` XCI partition processing with NSP. This is because it will eliminate duplicate code and it adds proper support for Rev1+ carts (gamecards with updates in the secure partition).
- Add support for installing NSP files (only of `Not Extracted Type`)

### DIFF
--- a/src/core/file_sys/submission_package.h
+++ b/src/core/file_sys/submission_package.h
@@ -13,9 +13,10 @@
 #include "core/file_sys/romfs_factory.h"
 #include "core/file_sys/vfs.h"
 #include "core/loader/loader.h"
-#include "romfs_factory.h"
 
 namespace FileSys {
+
+class PartitionFilesystem;
 
 class NSP : public ReadOnlyVfsDirectory {
 public:
--- a/src/core/file_sys/submission_package.h
+++ b/src/core/file_sys/submission_package.h
@@ -10,6 +10,7 @@
 #include "common/common_types.h"
 #include "common/swap.h"
 #include "core/file_sys/content_archive.h"
+#include "core/file_sys/romfs_factory.h"
 #include "core/file_sys/vfs.h"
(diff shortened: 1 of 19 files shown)


src/core/loader/nsp.cpp
```diff
@@ -21,22 +21,27 @@ namespace Loader {
 AppLoader_NSP::AppLoader_NSP(FileSys::VirtualFile file)
     : AppLoader(file), nsp(std::make_unique<FileSys::NSP>(file)),
       title_id(nsp->GetProgramTitleID()) {
+
```
lioncash: This newline isn't necessary since it's the start of the function.

src/core/file_sys/submission_package.h
```diff
@@ -0,0 +1,72 @@
...
+#include <array>
+#include <map>
+#include <vector>
+#include "common/common_types.h"
+#include "common/swap.h"
+#include "core/file_sys/content_archive.h"
+#include "core/file_sys/romfs_factory.h"
+#include "core/file_sys/vfs.h"
+#include "core/loader/loader.h"
+#include "romfs_factory.h"
```
lioncash: `core/file_sys/romfs_factory.h`

PR #1180, opened 2018-08-26
set: Fixed GetAvailableLanguageCodes() to follow the max_entries

tech4me:
Rightnow, in games use GetAvailableLanguageCodes(), there is a WriteBuffer() with size larger than the buffer_size. (Core Critical core\hle\kernel\hle_ipc.cpp:WriteBuffer:296: size (0000000000000088) is greater than buffer_size (0000000000000078))

0x88 = 17(languages) * 8
0x78 = 15(languages) * 8

GetAvailableLanguageCodes() can only support 15 languages.
After firmware 4.0.0 there are 17 supported language instead of 15, to enable this GetAvailableLanguageCodes2() need to be used.
So GetAvailableLanguageCodes() will be caped at 15 languages.
Reference:
http://switchbrew.org/index.php/Settings_services

### DIFF
--- a/src/core/hle/service/set/set.cpp
+++ b/src/core/hle/service/set/set.cpp
@@ -32,24 +32,59 @@ constexpr std::array<LanguageCode, 17> available_language_codes = {{
     LanguageCode::ZH_HANT,
 }};
 
+constexpr size_t pre4_0_0_max_entries = 0xF;
+constexpr size_t post4_0_0_max_entries = 0x40;
+
 LanguageCode GetLanguageCodeFromIndex(size_t index) {
     return available_language_codes.at(index);
 }
 
-void SET::GetAvailableLanguageCodes(Kernel::HLERequestContext& ctx) {
-    ctx.WriteBuffer(available_language_codes);
+template <size_t size>
+static std::array<LanguageCode, size> MakeLanguageCodeSubset() {
+    std::array<LanguageCode, size> arr;
+    std::copy_n(available_language_codes.begin(), size, arr.begin());
+    return arr;
+}
 
+static void PushResponseLanguageCode(Kernel::HLERequestContext& ctx, size_t max_size) {
     IPC::ResponseBuilder rb{ctx, 3};
     rb.Push(RESULT_SUCCESS);
-    rb.Push(static_cast<u32>(available_language_codes.size()));
+    if (available_language_codes.size() > max_size)
+        rb.Push(static_cast<u32>(max_size));
+    else
+        rb.Push(static_cast<u32>(available_language_codes.size()));
+}
+
+void SET::GetAvailableLanguageCodes(Kernel::HLERequestContext& ctx) {
+    if (available_language_codes.size() > pre4_0_0_max_entries)
+        ctx.WriteBuffer(MakeLanguageCodeSubset<pre4_0_0_max_entries>());
+    else
+        ctx.WriteBuffer(available_language_codes);
+
+    PushResponseLanguageCode(ctx, pre4_0_0_max_entries);
+
+    LOG_DEBUG(Service_SET, "called");
+}
+
+void SET::GetAvailableLanguageCodes2(Kernel::HLERequestContext& ctx) {
+    if (available_language_codes.size() > post4_0_0_max_entries)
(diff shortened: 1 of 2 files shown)


src/core/hle/service/set/set.cpp
```diff
@@ -36,8 +36,19 @@ LanguageCode GetLanguageCodeFromIndex(size_t index) {
     return available_language_codes.at(index);
 }
 
+template <size_t size>
+static constexpr std::array<LanguageCode, size> sub_language_codes() {
```
lioncash: Our code formatting uses `TitleCase` for function names (as can be seen by essentially every other function in the file). This is likely better named as `MakeLanguageCodeSubset`

src/core/hle/service/set/set.cpp
```diff
@@ -36,8 +36,19 @@ LanguageCode GetLanguageCodeFromIndex(size_t index) {
...
+    std::array<LanguageCode, size> arr;
+    for (size_t i = 0; i < size; ++i)
```
lioncash: This is more concise as:

```cpp
std::copy_n(available_language_codes.begin(), size, arr.begin());
```

the function will need to have constexpr removed, but this doesn't matter, given it's not a highly used function.

src/core/hle/service/set/set.cpp
```diff
@@ -36,8 +36,19 @@ LanguageCode GetLanguageCodeFromIndex(size_t index) {
...
+        arr.at(i) = available_language_codes.at(i);
+    return arr;
+}
+
 void SET::GetAvailableLanguageCodes(Kernel::HLERequestContext& ctx) {
-    ctx.WriteBuffer(available_language_codes);
+    if (available_language_codes.size() > pre4_0_0_max_entries)
```
lioncash: This can be extracted into its own function so it's not duplicated in `GetAvailableLanguageCodes2`

src/core/hle/service/set/set.cpp
```diff
@@ -46,10 +57,40 @@ void SET::GetAvailableLanguageCodes(Kernel::HLERequestContext& ctx) {
...
 
+void SET::GetAvailableLanguageCodes2(Kernel::HLERequestContext& ctx) {
+    if (available_language_codes.size() > post4_0_0_max_entries)
+        ctx.WriteBuffer(sub_language_codes<post4_0_0_max_entries>());
+    else
+        ctx.WriteBuffer(available_language_codes);
+
+    IPC::ResponseBuilder rb{ctx, 3};
+    rb.Push(RESULT_SUCCESS);
+    if (available_language_codes.size() > post4_0_0_max_entries)
```
lioncash: This can be extracted to its own function so its not duplicated in `GetAvailableLanguageCodes2`, `GetAvailableLanguageCodesCount`, and `GetAvailableLanguageCodeCount2`

src/core/hle/service/set/set.h
```diff
@@ -30,6 +30,9 @@ enum class LanguageCode : u64 {
 };
 LanguageCode GetLanguageCodeFromIndex(size_t idx);
 
+constexpr size_t pre4_0_0_max_entries = 0xF;
```
lioncash: These can be placed in the cpp file, as they aren't needed anywhere else.

src/core/hle/service/set/set.cpp
tech4me: Yep. std::copy_n is only constexpr in C++20. I guess I will get rid of the constexpr

src/core/hle/service/set/set.cpp
tech4me: This I just plan to leave it for now. Because MakeLanguageCodeSubset template function's template argument need to be constexpr which is hard to pass in through another function without some hack.
ie. 
```C++
void foo(size_t max_size) {
    MakeLanguageCodeSubset<max_size>();
}

constexpr size_t bar = 1;
foo(bar);
```
Would result in compile time error.

lioncash: Ah fair enough

tech4me:
@lioncash Fixed!

bunnei:
Thanks for the fix!

PR #1196, opened 2018-08-28
.travis: Use Citras ccache for builds instead of yuzus

FearlessTobi:
Makes the code cleaner and makes it easier to port other changes from Citra that build ontop of this (e.g. travis mingw builds).


bunnei:
@FearlessTobi please rebase

PR #1199, opened 2018-08-29
Various fixes and improvements to rasterizer cache

bunnei:
Summary of changes:

- Fixes bug with multiple rasterizer caches where multiple page maps might clash with eachother, we should only have one rasterizer cache page map for all of surface cache, shader cache, etc.
- Fixes calculation of ending page in `UpdatePagesCachedCount`.
- Cache in the rasterizer based on CPU address, not GPU address. From the perspective of the GPU memory manager, the CPU address is the "physical address". Games can map multiple GPU regions to the same CPU address region.
- Optimizes cache access using a two-level cache. First level of the cache is keyed on starting CPU page, second level is keyed on starting CPU address. We now invalidate if the starting page of a cached region is written to. 

Suggest reviewing commit by commit. Fixes some rendering issues with Disgaea 5, I'm sure others. Please test thoroughly, looking for both accuracy improvements/regressions and or performance improvements/regressions.

Supersedes #1197.

image

cc @degasus

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_gpu.h
+++ b/src/core/hle/service/nvdrv/devices/nvhost_gpu.h
@@ -10,6 +10,7 @@
 #include "common/common_types.h"
 #include "common/swap.h"
 #include "core/hle/service/nvdrv/devices/nvdevice.h"
+#include "video_core/memory_manager.h"
 
 namespace Service::Nvidia::Devices {
 
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -279,10 +279,9 @@ static constexpr auto RangeFromInterval(Map& map, const Interval& interval) {
     return boost::make_iterator_range(map.equal_range(interval));
 }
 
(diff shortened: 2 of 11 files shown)


src/core/hle/service/nvdrv/devices/nvhost_gpu.h
```diff
@@ -10,6 +10,7 @@
 #include "common/common_types.h"
 #include "common/swap.h"
 #include "core/hle/service/nvdrv/devices/nvdevice.h"
+#include "video_core/memory_manager.h"
```
lioncash: Is this necessary in the header?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -274,6 +274,41 @@ bool RasterizerOpenGL::AccelerateDrawBatch(bool is_indexed) {
     return true;
 }
 
+template <typename Map, typename Interval>
+constexpr auto RangeFromInterval(Map& map, const Interval& interval) {
```
lioncash: This should also be static.

src/core/hle/service/nvdrv/devices/nvhost_gpu.h
bunnei: Yes, it was included via memory.h before

PR #1203, opened 2018-08-30
gl_shader_decompiler: Implement POPC

Lakumakkara:
Used by Xenoblade Chronicles 2 and Yooka-Laylee.


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1302,6 +1302,12 @@ class GLSLGenerator {
                                           "((" + op_a + " << " + shift + ") + " + op_b + ')', 1, 1);
                 break;
             }
+            case OpCode::Id::POPC_C:
+            case OpCode::Id::POPC_R:
+            case OpCode::Id::POPC_IMM: {
+                regs.SetRegisterToInteger(instr.gpr0, true, 0, "bitCount(" + op_b + ')', 1, 1);
```
bunnei: According to envydis, this also has `ON(40, inv)` (https://github.com/envytools/envytools/blob/master/envydis/gm107.c#L2113).
Can you add: `ASSERT_MSG(!instr.popc.inv, "Unimplemented");`
So we can find when this is used and implement? Thanks!

Lakumakkara: I tested this just now. What it does is that it takes the bitwise complement of the operand before calculating the popcount.
For example: `POPC R9, 0x1.INV ;` returns 31.

PR #1205, opened 2018-08-31
Various fixes and improvements to rasterizer cache 2: Electric Boogaloo

bunnei:
Summary of changes:

- Fixes bug with multiple rasterizer caches where multiple page maps might clash with eachother, we should only have one rasterizer cache page map for all of surface cache, shader cache, etc.
- Fixes calculation of ending page in `UpdatePagesCachedCount`.
- Cache in the rasterizer based on CPU address, not GPU address. From the perspective of the GPU memory manager, the CPU address is the "physical address". Games can map multiple GPU regions to the same CPU address region.
- Optimizes cache access and invalidation using `boost::interval_map`.
- Fixes an issue where `surface_reserve` was not being used for `RecreateSurface`, meaning lots of surfaces could still be needlessly created and destroyed.
- Changes the "use accurate framebuffers" setting to perform accurate framebuffer copies, and remove the always invalidate and flush behavior. This was broken anyways, and it is not practical to have an always invalidate and flush mode with Switch emulation (even simple 2D games become a slide show).
- I've yet to see accurate framebuffer copies fix anything, so for now, the above setting is still default off. I encourage folks to test with it on and off, because it's still plausible that it fixes things.

Generally, this should improve performance quite a bit. Fixes some rendering issues with Disgaea 5, I'm sure others. Please test thoroughly, looking for both accuracy improvements/regressions and or performance improvements/regressions.

Supersedes #1197 and #1199.

cc @degasus


src/video_core/rasterizer_cache.h
```diff
@@ -16,63 +18,66 @@
...
-        for (auto iter = cached_objects.cbegin(); iter != cached_objects.cend();) {
-            const auto& object{iter->second};
+    /// Mark the specified page as being invalidated
+    void InvalidateRegion(VAddr addr, u64 size) {
+        if (size == 0)
+            return;
 
-            ++iter;
+        const ObjectInterval interval{addr, addr + size};
+        for (auto& pair : boost::make_iterator_range(object_cache.equal_range(interval))) {
```
degasus: I wonder which way is better:
- Iterating over the full range and assembling the std::set like done here.
- Just call Unregister on the first object, and do the map lookup object_cache.equal_range(interval) within the loop.

I guess both are very similiar, especially for the common case to just invalidate one object. But what will happen if a big blog gets updated, which overlaps lots of small objects?

src/video_core/rasterizer_cache.h
```diff
@@ -4,113 +4,80 @@
...
 protected:
     /// Tries to get an object from the cache with the specified address
-    T TryGet(Tegra::GPUVAddr addr) const {
-        const auto& search{cached_objects.find(addr)};
-        if (search != cached_objects.end()) {
-            return search->second;
+    T TryGet(VAddr addr) const {
+        const ObjectInterval interval{addr};
+        for (auto& pair : boost::make_iterator_range(object_cache.equal_range(interval))) {
+            for (auto& cached_object : pair.second) {
```
degasus: Do you think there is a point to use a interval_map<VAddr, std::map<VAddr, T>> instead? This would allow a direct access here. Through this set shall be small anyways.

bunnei: Going to leave as-is for now, and experiment with this (and some other changes) for a followup improvement.

src/video_core/rasterizer_cache.h
bunnei: Going to leave as-is for now, and experiment with this (and some other changes) for a followup improvement.

PR #1213, opened 2018-09-01
filesystem/maxwell_3d: Various changes to boot Project Octopath Traveller

DarkLordZach:
The timestamp CoreTiming change is full credit to @gdkchan 

This allows octopath to boot and produces the following dazzling screenshots:

### DIFF
--- a/src/core/hle/service/filesystem/fsp_srv.cpp
+++ b/src/core/hle/service/filesystem/fsp_srv.cpp
@@ -455,7 +455,7 @@ FSP_SRV::FSP_SRV() : ServiceFramework("fsp-srv") {
         {34, nullptr, "GetCacheStorageSize"},
         {51, &FSP_SRV::MountSaveData, "MountSaveData"},
         {52, nullptr, "OpenSaveDataFileSystemBySystemSaveDataId"},
-        {53, nullptr, "OpenReadOnlySaveDataFileSystem"},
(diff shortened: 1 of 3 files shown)


src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -516,6 +527,16 @@ void FSP_SRV::Initialize(Kernel::HLERequestContext& ctx) {
...
 }
 
+void FSP_SRV::OpenFileSystemWithPatch(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+
+    const auto type = rp.PopRaw<FileSystemType>();
+    const auto title_id = rp.PopRaw<u64>();
+
+    IPC::ResponseBuilder rb{ctx, 2, 0, 0};
+    rb.Push(ResultCode(-1));
```
lioncash: Shouldn't this log that it's stubbed?

bunnei:
Thanks to @gdkchan for the QueryGet change! This did not come from me.

src/core/hle/service/filesystem/fsp_srv.cpp
lioncash: The variables should likely also have `[[maybe_unused]]` beside them, or be logged out as well, to prevent unused variable warnings.

PR #1217, opened 2018-09-01
renderer_opengl: Implement a buffer cache.

degasus:
The idea of this cache is to avoid redundant uploads. So we are going
to cache the uploaded buffers within the stream_buffer and just reuse
the old pointers.
The next step is to implement a VBO cache on GPU memory, but for now,
I want to check the overhead of the cache management. Fetching the
buffer over PCI-E should be quite fast.

### DIFF
--- a/src/video_core/CMakeLists.txt
+++ b/src/video_core/CMakeLists.txt
@@ -22,6 +22,7 @@ add_library(video_core STATIC
     rasterizer_interface.h
     renderer_base.cpp
     renderer_base.h
+    renderer_opengl/gl_buffer_cache.cpp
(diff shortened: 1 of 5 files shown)


bunnei:
Removing canary merge - Sounds like there's been some good testing of this. It seems like there are no accuracy regressions, but that performance can be worse.

@degasus Any ideas on how to improve performance with this? Should we experiment with only caching with large buffers, for example?

degasus:
I've limited the cache to objects bigger than 4 kB. I hope those objects won't overlap that much with regard to the CPU page size. Honestly, I haven't done a clean verification, but it should be big enough to not yield a performance regression. Likely not a big performance increase either...

PR #1219, opened 2018-09-01
Build - Upload fewer artifacts

jroweboy:
Appveyor has a limit on artifact retention, and we hit the limit all the
time, so just lower the number of build artifacts to just the final zip

For reference, this cuts the total artifacts size down to 44MB from 104MB per build.


Dantttt:
Seems a good solution!

PR #1236, opened 2018-09-04
Update microprofile scopes.

degasus:
Blame the subsystems which deserve the blame :)

The updated list is not complete, just the ones I've spotted on random sampling the stack trace.

### DIFF
--- a/src/audio_core/stream.cpp
+++ b/src/audio_core/stream.cpp
@@ -10,6 +10,7 @@
 #include "audio_core/stream.h"
 #include "common/assert.h"
 #include "common/logging/log.h"
+#include "common/microprofile.h"
 #include "core/core_timing.h"
 #include "core/core_timing_util.h"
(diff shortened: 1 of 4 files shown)


src/audio_core/stream.cpp
```diff
@@ -94,7 +95,10 @@ void Stream::PlayNextBuffer() {
     CoreTiming::ScheduleEventThreadsafe(GetBufferReleaseCycles(*active_buffer), release_event, {});
 }
 
+MICROPROFILE_DEFINE(AudioOutput, "Audio", "ReleaseActiveBuffer", MP_RGB(100, 100, 255));
+
 void Stream::ReleaseActiveBuffer() {
+    MICROPROFILE_SCOPE(AudioOutput);
```
degasus: This was actually spotted by accident. Audio stretching with pitch correction should not be done on the main thread.

PR #1237, opened 2018-09-04
Optimizations

degasus:
Two small and trivial 1% optimizations. Please read the commit messages of the patches.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -441,8 +441,8 @@ void System::SetGPUDebugContext(std::shared_ptr<Tegra::DebugContext> context) {
     impl->debug_context = std::move(context);
 }
 
-std::shared_ptr<Tegra::DebugContext> System::GetGPUDebugContext() const {
-    return impl->debug_context;
+Tegra::DebugContext* System::GetGPUDebugContext() const {
+    return impl->debug_context.get();
 }
 
 void System::SetFilesystem(std::shared_ptr<FileSys::VfsFilesystem> vfs) {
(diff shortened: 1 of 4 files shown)


src/video_core/command_processor.cpp
```diff
@@ -47,8 +49,6 @@ void GPU::WriteReg(u32 method, u32 subchannel, u32 value, u32 remaining_params)
         return;
     }
 
-    ASSERT(bound_engines.find(subchannel) != bound_engines.end());
-
     const EngineID engine = bound_engines[subchannel];
```
degasus: It is also possible to store the function pointers in the std:array. But I'm not sure if the switch is slower or faster than this kind of indirect branch.

src/core/core.cpp
```diff
@@ -441,8 +441,8 @@ void System::SetGPUDebugContext(std::shared_ptr<Tegra::DebugContext> context) {
     impl->debug_context = std::move(context);
 }
 
-std::shared_ptr<Tegra::DebugContext> System::GetGPUDebugContext() const {
-    return impl->debug_context;
+Tegra::DebugContext* System::GetGPUDebugContext() const {
```
lioncash: This can be a reference, given it's required to be valid for the entire system execution lifetime.

degasus: Reference to the shared_ptr? AFAIK the object itself might be a nullptr.

lioncash: Ah true, disregard my comment then. I forgot that the command-line version of yuzu doesn't set a debug context.

PR #1246, opened 2018-09-06
gl_rasterizer: Use baseInstance instead of moving the buffer points.

degasus:
This hopefully helps our cache not to redundant upload the vertex buffer.

But bad luck, it has no impact at all on SMO. It still might help to fix the worst case.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -151,11 +151,6 @@ void RasterizerOpenGL::SetupVertexArrays() {
         Tegra::GPUVAddr start = vertex_array.StartAddress();
         const Tegra::GPUVAddr end = regs.vertex_array_limit[index].LimitAddress();
 
-        if (regs.instanced_arrays.IsInstancingEnabled(index) && vertex_array.divisor != 0) {
-            start += static_cast<Tegra::GPUVAddr>(vertex_array.stride) *
-                     (gpu.state.current_instance / vertex_array.divisor);
-        }
-
         ASSERT(end > start);
         const u64 size = end - start + 1;
         const GLintptr vertex_buffer_offset = buffer_cache.UploadMemory(start, size);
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -493,11 +490,26 @@ void RasterizerOpenGL::DrawArrays() {
...
+            glDrawElementsInstancedBaseVertexBaseInstance(
+                primitive_mode, regs.index_array.count,
+                MaxwellToGL::IndexFormat(regs.index_array.format),
+                reinterpret_cast<const void*>(index_buffer_offset), 1, base_vertex,
+                gpu.state.current_instance);
+        } else {
+            glDrawElementsBaseVertex(primitive_mode, regs.index_array.count,
+                                     MaxwellToGL::IndexFormat(regs.index_array.format),
+                                     reinterpret_cast<const void*>(index_buffer_offset),
+                                     base_vertex);
```
degasus: If you want, I can always use the BaseInstance call here. It would yield cleaner code, but a tiny bit more overhead in the driver.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -168,7 +164,7 @@ void RasterizerOpenGL::SetupVertexArrays() {
             // Tell OpenGL that this is an instanced vertex buffer to prevent accessing different
             // indexes on each vertex. We do the instance indexing manually by incrementing the
             // start address of the vertex buffer.
-            glVertexBindingDivisor(index, 1);
```
Subv: Update the comment

degasus:
From ARB_base_instance:
    1) Does <baseinstance> offset gl_InstanceID?
    RESOLVED: No. gl_InstanceID always starts from zero and counts up by one
    for each instance rendered.

bunnei:
LGTM. Tested a bunch of games and rebased for you 👍

PR #1256, opened 2018-09-06
Initial support for non-2D textures

bunnei:
This is a bunch of changes to add initial (and partial) support for non-2D textures. This includes a combination of changes to the rasterizer cache (to properly track and upload non-2D textures) and to the generated shaders (to properly use non-2D textures). I tried to make these changes in such a way that there are not any regressions for things that used to render, but no promises.

Through this, I found that we need to do more accurate texture copies - which means there is a fairly big performance hit (but I'm sure we can optimize this better later).

This does not implement 3D or cubemap textures yet, as I haven't worked out all the details for these.

Suggest reviewing commit by commit. Please test thoroughly, keeping an eye out for both performance regressions and rendering regressions.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -7,6 +7,7 @@
 
 #include "common/alignment.h"
 #include "common/assert.h"
+#include "common/logging/log.h"
 #include "common/microprofile.h"
 #include "common/scope_exit.h"
 #include "core/core.h"
@@ -54,8 +55,8 @@ static VAddr TryGetCpuAddr(Tegra::GPUVAddr gpu_addr) {
     params.depth = config.tic.Depth();
     params.unaligned_height = config.tic.Height();
     params.size_in_bytes = params.SizeInBytes();
-    params.cache_width = Common::AlignUp(params.width, 16);
-    params.cache_height = Common::AlignUp(params.height, 16);
+    params.cache_width = Common::AlignUp(params.width, 8);
+    params.cache_height = Common::AlignUp(params.height, 8);
     params.target = SurfaceTargetFromTextureType(config.tic.texture_type);
     return params;
 }
@@ -74,8 +75,8 @@ static VAddr TryGetCpuAddr(Tegra::GPUVAddr gpu_addr) {
     params.depth = 1;
     params.unaligned_height = config.height;
     params.size_in_bytes = params.SizeInBytes();
-    params.cache_width = Common::AlignUp(params.width, 16);
-    params.cache_height = Common::AlignUp(params.height, 16);
+    params.cache_width = Common::AlignUp(params.width, 8);
+    params.cache_height = Common::AlignUp(params.height, 8);
     params.target = SurfaceTarget::Texture2D;
     return params;
 }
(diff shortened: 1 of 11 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1753,10 +1754,36 @@ class GLSLGenerator {
                 break;
             }
             case OpCode::Id::TEX: {
-                const std::string op_a = regs.GetRegisterAsFloat(instr.gpr8);
-                const std::string op_b = regs.GetRegisterAsFloat(instr.gpr8.Value() + 1);
-                const std::string sampler = GetSampler(instr.sampler);
-                const std::string coord = "vec2 coords = vec2(" + op_a + ", " + op_b + ");";
+                ASSERT_MSG(instr.tex.array == 0, "TEX arrays unimplemented");
+                Tegra::Shader::TextureType texture_type{instr.tex.texture_type};
+                std::string coord{};
```
lioncash: braces aren't necessary here.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1778,20 +1805,65 @@ class GLSLGenerator {
                 break;
             }
             case OpCode::Id::TEXS: {
-                const std::string op_a = regs.GetRegisterAsFloat(instr.gpr8);
-                const std::string op_b = regs.GetRegisterAsFloat(instr.gpr20);
-                const std::string sampler = GetSampler(instr.sampler);
-                const std::string coord = "vec2 coords = vec2(" + op_a + ", " + op_b + ");";
+                std::string coord{};
```
lioncash: ditto

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1778,20 +1805,65 @@ class GLSLGenerator {
...
                 break;
             }
             case OpCode::Id::TLDS: {
-                const std::string op_a = regs.GetRegisterAsInteger(instr.gpr8);
-                const std::string op_b = regs.GetRegisterAsInteger(instr.gpr20);
-                const std::string sampler = GetSampler(instr.sampler);
-                const std::string coord = "ivec2 coords = ivec2(" + op_a + ", " + op_b + ");";
+                ASSERT(instr.tlds.GetTextureType() == Tegra::Shader::TextureType::Texture2D);
+                ASSERT(instr.tlds.IsArrayTexture() == false);
+                std::string coord{};
```
lioncash: ditto

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -166,6 +171,26 @@ static constexpr std::array<FormatTuple, SurfaceParams::MaxPixelFormat> tex_form
...
+    case SurfaceParams::SurfaceTarget::Texture3D:
+        return GL_TEXTURE_3D;
+    case SurfaceParams::SurfaceTarget::Texture1DArray:
+        return GL_TEXTURE_1D_ARRAY;
+    case SurfaceParams::SurfaceTarget::Texture2DArray:
+        return GL_TEXTURE_2D_ARRAY;
+    case SurfaceParams::SurfaceTarget::TextureCubemap:
+        return GL_TEXTURE_CUBE_MAP;
+    }
+    LOG_CRITICAL(Render_OpenGL, "Unimplemented texture target={}", static_cast<u32>(target));
```
lioncash: `common/logging/log.h` needs to be included.

bunnei: addressed

bunnei:
@degasus it turns out we need to use the PBO code path for more accurate surface copies (see commit 9444660c6800a1b43dd48a9ce8184eacf769939b). This fixes lots of graphical issues, particularly in SMO. Any ideas on how we can optimize this?

bunnei:
Big improvements with the latest iteration - I think I fixed most accuracy regressions (e.g. with Octopath being stuck), and I also removed the need for the PBO codepath. As such, this should fix most of the performance regressions previously introduced.

PR #1262, opened 2018-09-08
video_core: Arithmetic overflow warning fix for gl_rasterizer

zero334:
-  Fixed warnings, which were indicating a possible incorrect behavior from integral
 promotion rules and types larger than those in which arithmetic is
 typically performed (used C-style casts for numeric types).
 
-  Added const for variables where possible and meaningful.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -152,7 +152,7 @@ void RasterizerOpenGL::SetupVertexArrays() {
         const Tegra::GPUVAddr end = regs.vertex_array_limit[index].LimitAddress();
 
         if (regs.instanced_arrays.IsInstancingEnabled(index) && vertex_array.divisor != 0) {
-            start += (Tegra::GPUVAddr)vertex_array.stride *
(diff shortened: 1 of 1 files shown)


lioncash:
Our convention is to use C++ casts, not C-style casts.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -493,7 +493,8 @@ void RasterizerOpenGL::DrawArrays() {
         const GLint base_vertex{static_cast<GLint>(regs.vb_element_base)};
 
         // Adjust the index buffer offset so it points to the first desired index.
-        index_buffer_offset += regs.index_array.first * regs.index_array.FormatSizeInBytes();
+        index_buffer_offset +=
+            (GLintptr)regs.index_array.first * regs.index_array.FormatSizeInBytes();
```
lioncash: This should also be casting FormatSizeInBytes, otherwise this is doing `signed * unsigned`. Which can cause sign conversion warnings.

zero334:
Thanks for the hints @lioncash 
I will immediately apply some changes.

PR #1263, opened 2018-09-08
shader_decompiler:  Implemented (Partially) Texture Processing Modes

FernandoS27:
After some heavy reverse engineering. We can up with the special texture modes. Currently I have only implemented them in TEX instruction but they are also used in other instructions. I'm sending this PR early for test purposes.

Big thanks to Rodrigo for writing and dumping the needed tests.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -254,6 +254,15 @@ enum class TextureQueryType : u64 {
     BorderColor = 22,
 };
 
+enum class TextureProcessMode : u64 {
+    None = 0,
+    LZ = 1,  // Unknown, appears to be the same as none.
+    LB = 2,  // Load Bias.
+    LL = 3,  // Load LOD (LevelOfDetail)
+    LBA = 6, // Load Bias. The A is unknown, does not appear to differ with LB
+    LLA = 7  // Load LOD. The A is unknown, does not appear to differ with LL
+};
+
 enum class IpaInterpMode : u64 { Linear = 0, Perspective = 1, Flat = 2, Sc = 3 };
 enum class IpaSampleMode : u64 { Default = 0, Centroid = 1, Offset = 2 };
 
@@ -522,6 +531,7 @@ union Instruction {
         BitField<28, 1, u64> array;
         BitField<29, 2, TextureType> texture_type;
         BitField<31, 4, u64> component_mask;
+        BitField<55, 3, TextureProcessMode> process_mode;
 
         bool IsComponentEnabled(size_t component) const {
             return ((1ull << component) & component_mask) != 0;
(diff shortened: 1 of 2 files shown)


src/video_core/engines/shader_bytecode.h
```diff
@@ -244,6 +244,8 @@ enum class TextureType : u64 {
     TextureCube = 3,
 };
 
+enum class TextureProcessMode : u64 { None = 0, LZ = 1, LB = 2, LL = 3, LBA = 6, LLA = 7 };
```
Subv: Add comments about what each of these mean

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1781,15 +1781,39 @@ class GLSLGenerator {
...
-                const std::string texture = "texture(" + sampler + ", coords)";
+                std::string texture;
 
+                switch (instr.tex.process_mode) {
+                case Tegra::Shader::TextureProcessMode::None: {
+                    texture = "texture(" + sampler + ", coords)";
+                    break;
+                }
+                case Tegra::Shader::TextureProcessMode::LB:
+                case Tegra::Shader::TextureProcessMode::LBA: {
```
Subv: What's the difference between LB and LBA? Is that not known or just irrelevant? (Add a comment or a todo about it regardless)

FernandoS27:
I've added documentation to it. Feel free to review it.

src/video_core/engines/shader_bytecode.h
```diff
@@ -254,6 +254,15 @@ enum class TextureQueryType : u64 {
     BorderColor = 22,
 };
 
+enum class TextureProcessMode : u64 {
+    None = 0,
+    LZ = 1,  // Unknown, appears to be the same as none.
```
Subv: I think this may be LevelZero, as in, it always fetches from the 0th mipmap.

Are you sure the L stands for Load and not Level?

FernandoS27: would LevelBias make sense then?

I think LZ could be tested as mipmap 0 but there are other instructions where LZ is the default none. Like in TEXS.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1786,15 +1786,45 @@ class GLSLGenerator {
                     coord = "vec2 coords = vec2(" + x + ", " + y + ");";
                     texture_type = Tegra::Shader::TextureType::Texture2D;
                 }
+                const std::string op_c = regs.GetRegisterAsFloat(instr.gpr20);
```
Subv: fwiw, it's unknown if different texture types use different registers for this field, add a TODO or put it inside each of the switch cases.

PR #1264, opened 2018-09-08
video_core: Optimize the command processor.

degasus:
This PR tries to force inline the hot loop generation of the command processor. With the main loop within the command processor, it is easier to profile the CPU overhead of video_core vs SVC.

### DIFF
--- a/src/video_core/command_processor.h
+++ b/src/video_core/command_processor.h
@@ -7,6 +7,7 @@
 #include <type_traits>
 #include "common/bit_field.h"
 #include "common/common_types.h"
+#include "video_core/memory_manager.h"
 
 namespace Tegra {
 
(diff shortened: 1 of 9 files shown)


src/video_core/command_processor.h
```diff
@@ -19,6 +20,22 @@ enum class SubmissionMode : u32 {
...
+        BitField<8, 2, u32> unk1;
+        BitField<10, 21, u32> sz;
+        BitField<31, 1, u32> unk2;
+    };
+
+    GPUVAddr Address() const {
+        return (static_cast<GPUVAddr>(gpu_va_hi) << 32) | entry0;
+    }
+};
+static_assert(sizeof(CommandHeaderList) == 8, "IoctlGpfifoEntry is incorrect size");
```
lioncash: This isn't the name of the struct anymore.

src/video_core/command_processor.h
```diff
@@ -19,6 +20,22 @@ enum class SubmissionMode : u32 {
     IncreaseOnce = 5
 };
 
+struct CommandHeaderList {
```
lioncash: This isn't really a "list", and should probably have a better name.

PR #1271, opened 2018-09-08
GPU: Basic implementation of the Kepler Inline Memory engine (p2mf).

Subv:
This engine writes data from a FIFO register into the configured address.

It is used by nouveau.

### DIFF
--- a/src/video_core/engines/kepler_memory.cpp
+++ b/src/video_core/engines/kepler_memory.cpp
@@ -1,0 +1,45 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/logging/log.h"
+#include "core/memory.h"
+#include "video_core/engines/kepler_memory.h"
+
+namespace Tegra::Engines {
+
+KeplerMemory::KeplerMemory(MemoryManager& memory_manager) : memory_manager(memory_manager) {}
+KeplerMemory::~KeplerMemory() = default;
+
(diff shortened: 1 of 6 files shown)


src/video_core/engines/kepler_memory.cpp
```diff
@@ -0,0 +1,46 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/logging/log.h"
+#include "core/memory.h"
+#include "video_core/engines/kepler_memory.h"
+
+namespace Tegra {
```
lioncash: This can be:

```cpp
namespace Tegra::Engines {

...

} // namespace Tegra::Engines
```

src/video_core/engines/kepler_memory.h
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Tegra::Engines {
+
+#define KEPLERMEMORY_REG_INDEX(field_name)                                                         \
+    (offsetof(Tegra::Engines::KeplerMemory::Regs, field_name) / sizeof(u32))
+
+class KeplerMemory final {
+public:
+    KeplerMemory(MemoryManager& memory_manager);
+    ~KeplerMemory() = default;
```
lioncash: This should likely be defaulted in the cpp file, so it makes forward declarations less likely to cause a compilation error if a non-trivial type is ever added to this class.

bunnei:
@Subv needs a rebase.

PR #1273, opened 2018-09-08
Shaders: Implemented multiple-word loads and stores to and from attribute memory.

Subv:
This seems to be an optimization performed by nouveau.


bunnei:
@Subv ping to address review comments.

PR #1274, opened 2018-09-08
Add (Currently broken.) after "Enable multi-core"

MysticExile:
Should be good now

### DIFF
--- a/src/yuzu/configuration/configure_general.ui
+++ b/src/yuzu/configuration/configure_general.ui
@@ -61,7 +61,7 @@
           <item>
            <widget class="QCheckBox" name="use_multi_core">
             <property name="text">
-             <string>Enable multi-core</string>
+             <string>Enable multi-core (Currently broken.)</string>
             </property>
            </widget>
           </item>


src/yuzu/configuration/configure_general.ui
```diff
@@ -61,7 +61,7 @@
           <item>
            <widget class="QCheckBox" name="use_multi_core">
             <property name="text">
-             <string>Enable multi-core</string>
+             <string>Enable multi-core (Currently broken.)</string>
```
bunnei: Should be:
`Enable multi-core (currently broken)`

bunnei:
On second thought, let's just remove this from the UI for now. (We should still leave the config setting, FWIW... it just shouldn't be exposed to users at all if its broken). Thanks!

MysticExile:
@bunnei I’ll close it for now then. And I’ll open a new PR later this day to entirely remove the option from the UI. Thanks for reviewing!

jroweboy:
@bunnei removing the option from the UI means that for the small number of people that have launched citra in the past with the option enabled, they won't be able to turn it off without wiping their config

PR #1279, opened 2018-09-09
shader_decompiler: Implemented (Partialy) Control Codes and CSETP

FernandoS27:
After some intense RE, we were able to get the behavior of CSETP and Control Codes. This PR only implements the base Control Codes, CSETP and an example of their use, as many instructions actually modify them.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -240,6 +240,41 @@ enum class FlowCondition : u64 {
     Fcsm_Tr = 0x1C, // TODO(bunnei): What is this used for?
 };
 
+enum class ControlCode : u64 {
+    F = 0,
+    LT = 1,
+    EQ = 2,
+    LE = 3,
+    GT = 4,
+    NE = 5,
+    GE = 6,
+    Num = 7,
+    Nan = 8,
+    LTU = 9,
+    EQU = 10,
+    LEU = 11,
+    GTU = 12,
+    NEU = 13,
+    GEU = 14,
+    //
+    OFF = 16,
+    LO = 17,
+    SFF = 18,
+    LS = 19,
+    HI = 20,
+    SFT = 21,
+    HS = 22,
+    OFT = 23,
+    CSM_TA = 24,
+    CSM_TR = 25,
+    CSM_MX = 26,
+    FCSM_TA = 27,
+    FCSM_TR = 28,
+    FCSM_MX = 29,
+    RLE = 30,
+    RGT = 31,
+};
+
 enum class PredicateResultMode : u64 {
     None = 0x0,
     NotZero = 0x3,
@@ -735,6 +770,7 @@ union Instruction {
         BitField<36, 5, u64> index;
     } cbuf36;
 
+    BitField<47, 1, u64> generates_cc;
     BitField<61, 1, u64> is_b_imm;
     BitField<60, 1, u64> is_b_gpr;
     BitField<59, 1, u64> is_c_gpr;
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -581,6 +581,15 @@ union Instruction {
     } pset;
 
     union {
+        BitField<0, 3, u64> pred0;
+        BitField<3, 3, u64> pred3;
+        BitField<8, 5, ControlCode> cc; // flag in cc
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2014,31 +2033,55 @@ class GLSLGenerator {
...
+                if (instr.psetp.pred0 != static_cast<u64>(Pred::UnusedIndex)) {
+                    // Set the secondary predicate to the result of !Predicate OP SecondPredicate,
+                    // if enabled
+                    SetPredicate(instr.psetp.pred0,
+                                 "!(" + predicate + ") " + combiner + " (" + second_pred + ')');
+                }
+                break;
+            }
+            case OpCode::Id::CSETP: {
+                std::string pred =
```
lioncash: These can be const.

src/video_core/engines/shader_bytecode.h
```diff
@@ -984,6 +1030,7 @@ class OpCode {
             INST("010010110101----", Id::ISET_C, Type::IntegerSet, "ISET_C"),
             INST("0011011-0101----", Id::ISET_IMM, Type::IntegerSet, "ISET_IMM"),
             INST("0101000010010---", Id::PSETP, Type::PredicateSetPredicate, "PSETP"),
+            INST("010100001010----", Id::PSETP, Type::PredicateSetPredicate, "CSETP"),
```
Subv: Wrong opcode

Hexagon12: *Id::CSETP*

FernandoS27: Control Codes is indeed a 32 bits flag register. I implemented it using 32 1 bit variables to abuse GLSL compiler's deadcode elimination and thus don't affect performance so heavily. As of right now this implementation does not implement everything, there's still a bunch of flags to implement. i just wanted to write a single example on how it is done.

I'll amend the title to signify it's partial.

FernandoS27:
Changes made.

FernandoS27:
Rebased

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1576,6 +1591,11 @@ class GLSLGenerator {
 
                 regs.SetRegisterToInteger(instr.gpr0, instr.conversion.is_output_signed, 0, op_a, 1,
                                           1, instr.alu.saturate_d, 0, instr.conversion.dest_size);
+                if (instr.generates_cc.Value() != 0) {
+                    const std::string neucondition =
+                        "( " + regs.GetRegisterAsInteger(instr.gpr0, 0, false) + " != 0 )";
```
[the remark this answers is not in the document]
FernandoS27: You are right

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1641,6 +1673,10 @@ class GLSLGenerator {
 
                 regs.SetRegisterToInteger(instr.gpr0, instr.conversion.is_output_signed, 0, op_a, 1,
                                           1, instr.alu.saturate_d, 0, instr.conversion.dest_size);
+                if (instr.generates_cc.Value() != 0) {
```
Subv: I think this code might be wrong, it doesn't take into account the saturation stuff nor the destination size.

I think it's better if you pass generates_cc to SetRegister and do the flag setting there.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -351,6 +359,24 @@ class GLSLRegisterManager {
         shader.AddLine(dest + " = " + src + ';');
     }
 
+    std::string GetControlCode(const Tegra::Shader::ControlCode cc) const {
+        switch (cc) {
+        case Tegra::Shader::ControlCode::NEU:
+            return "!(" + GetInternalFlag(InternalFlag::ZeroFlag) + ')';
+        default:
```
Subv: log + UNREACHABLE in the default case.

FernandoS27:
that shouldn't be a problem if the IR is SSA. It's a piece of cake to eliminate deadcode in that format.

bunnei:
Thanks!

PR #1287, opened 2018-09-10
Use ARB_multi_bind for uniform buffers

ReinUsesLisp:
It should also be used with glBindTexture, but I leave that for someone more experienced in the codebase (and it may be a bad idea while non-2D texture families are being implemented).

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -642,15 +642,17 @@ void RasterizerOpenGL::SamplerInfo::SyncWithConfig(const Tegra::Texture::TSCEntr
 u32 RasterizerOpenGL::SetupConstBuffers(Maxwell::ShaderStage stage, Shader& shader,
                                         u32 current_bindpoint) {
     MICROPROFILE_SCOPE(OpenGL_UBO);
-    const u64 max_binds = Tegra::Engines::Maxwell3D::Regs::MaxConstBuffers;
     const auto& gpu = Core::System::GetInstance().GPU();
     const auto& maxwell3d = gpu.Maxwell3D();
     const auto& shader_stage = maxwell3d.state.shader_stages[static_cast<size_t>(stage)];
     const auto& entries = shader->GetShaderEntries().const_buffer_entries;
 
-    std::array<GLuint, max_binds> bind_buffers = {};
-    std::array<GLintptr, max_binds> bind_offsets = {};
-    std::array<GLsizeiptr, max_binds> bind_sizes = {};
+    constexpr u64 max_binds = Tegra::Engines::Maxwell3D::Regs::MaxConstBuffers;
+    std::array<GLuint, max_binds> bind_buffers;
+    std::array<GLintptr, max_binds> bind_offsets;
+    std::array<GLsizeiptr, max_binds> bind_sizes;
+
+    ASSERT_MSG(entries.size() <= max_binds, "Exceeded expected number of binding points.");
 
     // Upload only the enabled buffers from the 16 constbuffers of each shader stage
     for (u32 bindpoint = 0; bindpoint < entries.size(); ++bindpoint) {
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -151,6 +151,7 @@ class RasterizerOpenGL : public VideoCore::RasterizerInterface {
     void SyncLogicOpState();
 
     bool has_ARB_direct_state_access = false;
+    bool has_ARB_multi_bind = false;
```
Subv: You're making this a required extension but this variable isn't used at all. Add the requirement to the list of mandatory extensions in both frontends (Qt and SDL)

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -634,17 +637,23 @@ void RasterizerOpenGL::SamplerInfo::SyncWithConfig(const Tegra::Texture::TSCEntr
 u32 RasterizerOpenGL::SetupConstBuffers(Maxwell::ShaderStage stage, Shader& shader,
                                         u32 current_bindpoint) {
     MICROPROFILE_SCOPE(OpenGL_UBO);
+    const u64 max_binds = Tegra::Engines::Maxwell3D::Regs::MaxConstBuffers;
     const auto& gpu = Core::System::GetInstance().GPU();
     const auto& maxwell3d = gpu.Maxwell3D();
     const auto& shader_stage = maxwell3d.state.shader_stages[static_cast<size_t>(stage)];
     const auto& entries = shader->GetShaderEntries().const_buffer_entries;
 
+    std::array<GLuint, max_binds> buffers = {};
```
Subv: These could use better names, tbh

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -672,15 +681,20 @@ u32 RasterizerOpenGL::SetupConstBuffers(Maxwell::ShaderStage stage, Shader& shad
...
-        glBindBufferRange(GL_UNIFORM_BUFFER, current_bindpoint + bindpoint,
-                          buffer_cache.GetHandle(), const_buffer_offset, size);
-
         // Now configure the bindpoint of the buffer inside the shader
         glUniformBlockBinding(shader->GetProgramHandle(),
                               shader->GetProgramResourceIndex(used_buffer),
                               current_bindpoint + bindpoint);
+
+        // Prepare values for multibind
+        buffers[bindpoint] = buffer_cache.GetHandle();
```
Subv: Maybe assert that bindpoint is less than the array size

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -634,17 +637,23 @@ void RasterizerOpenGL::SamplerInfo::SyncWithConfig(const Tegra::Texture::TSCEntr
...
                                         u32 current_bindpoint) {
     MICROPROFILE_SCOPE(OpenGL_UBO);
+    const u64 max_binds = Tegra::Engines::Maxwell3D::Regs::MaxConstBuffers;
     const auto& gpu = Core::System::GetInstance().GPU();
     const auto& maxwell3d = gpu.Maxwell3D();
     const auto& shader_stage = maxwell3d.state.shader_stages[static_cast<size_t>(stage)];
     const auto& entries = shader->GetShaderEntries().const_buffer_entries;
 
+    std::array<GLuint, max_binds> buffers = {};
+    std::array<GLintptr, max_binds> offsets = {};
```
bunnei: You can remove the `= {}` here and below - this is the default behavior for `std::array`

ReinUsesLisp:
Addressed feedback. About std::array, [this](http://www.cplusplus.com/reference/array/array/array/) says that it won't be initialized.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -634,17 +637,23 @@ void RasterizerOpenGL::SamplerInfo::SyncWithConfig(const Tegra::Texture::TSCEntr
 u32 RasterizerOpenGL::SetupConstBuffers(Maxwell::ShaderStage stage, Shader& shader,
                                         u32 current_bindpoint) {
     MICROPROFILE_SCOPE(OpenGL_UBO);
+    const u64 max_binds = Tegra::Engines::Maxwell3D::Regs::MaxConstBuffers;
```
lioncash: This should be `constexpr` otherwise the compiler is allowed to reject the std::array declarations below (const != always usable at compile-time, though most compilers permit it if the compiler can deduce it ahead of time).

PR #1288, opened 2018-09-10
Remove the multi-core option from the UI

MysticExile:
There we are.


FearlessTobi:
Just gonna repeat here what @jroweboy said on the initial PR: If we merge this PR, this can lead to problems for people that have multicore still enabled.

MysticExile:
I didn't see that comment. Isn't it disabled by default? So anyone who has it enabled probably isn't using yuzu anyways. And anyone who still somehow has it enabled and uses Yuzu we could just help in the discord server (editing the qt-config file).

bunnei:
Eh - I'm OK with what @jroweboy said. It's really not worth the effort to change this setting, and I prefer we keep it around for now (just without it being exposed to users).
Standard guidance (if things are really broken for a user without any indication why) should include wiping out user settings.

PR #1289, opened 2018-09-10
shader_decompiler: Implemented LEA and PSET

FernandoS27:
Here's the last 2 instructions used by SMO. They are not 100% complete as it's almost impossible to implement a seedless random in glsl for the case of PSET.BF. Thankfully, I haven't found a shader that uses it. LEA was reversed engineered and NVIDIA's LEA calculates an effective adress the next manner LEA a b c = a + b*2^c.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -1505,6 +1505,73 @@ private:
                                           1, 1);
                 break;
             }
+            case OpCode::Id::LEA_R2:
+            case OpCode::Id::LEA_R1:
+            case OpCode::Id::LEA_IMM:
+            case OpCode::Id::LEA_RZ:
+            case OpCode::Id::LEA_HI: {
+                std::string op_a;
+                std::string op_b;
+                std::string op_c;
+
+                switch (opcode->GetId()) {
+                case OpCode::Id::LEA_R2: {
+                    op_a = regs.GetRegisterAsInteger(instr.gpr20);
+                    op_b = regs.GetRegisterAsInteger(instr.gpr39);
+                    op_c = std::to_string(instr.lea.r2.entry_a);
+                    break;
+                }
+
+                case OpCode::Id::LEA_R1: {
+                    const bool neg = instr.lea.r1.neg != 0;
+                    op_a = regs.GetRegisterAsInteger(instr.gpr8);
+                    if (neg)
+                        op_a = "-(" + op_a + ')';
+                    op_b = regs.GetRegisterAsInteger(instr.gpr20);
+                    op_c = std::to_string(instr.lea.r1.entry_a);
+                    break;
+                }
+
+                case OpCode::Id::LEA_IMM: {
+                    const bool neg = instr.lea.imm.neg != 0;
+                    op_b = regs.GetRegisterAsInteger(instr.gpr8);
+                    if (neg)
+                        op_b = "-(" + op_b + ')';
+                    op_a = std::to_string(instr.lea.imm.entry_a);
+                    op_c = std::to_string(instr.lea.imm.entry_b);
+                    break;
+                }
+
+                case OpCode::Id::LEA_RZ: {
+                    const bool neg = instr.lea.rz.neg != 0;
+                    op_b = regs.GetRegisterAsInteger(instr.gpr8);
+                    if (neg)
+                        op_b = "-(" + op_b + ')';
+                    op_a = regs.GetUniform(instr.lea.rz.cb_index, instr.lea.rz.cb_offset,
+                                           GLSLRegister::Type::Integer);
+                    op_c = std::to_string(instr.lea.rz.entry_a);
+
+                    break;
+                }
+
+                case OpCode::Id::LEA_HI:
+                default: {
+                    op_b = regs.GetRegisterAsInteger(instr.gpr8);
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1505,6 +1505,75 @@ class GLSLGenerator {
...
+                switch (opcode->GetId()) {
+                case OpCode::Id::LEA_R2: {
+                    op_a = regs.GetRegisterAsInteger(instr.gpr20);
+                    op_b = regs.GetRegisterAsInteger(instr.gpr39);
+                    op_c = std::to_string(instr.lea.r2.entry_a);
+                    break;
+                }
+
+                case OpCode::Id::LEA_R1: {
+                    neg = (instr.lea.r1.neg != 0);
```
lioncash: This doesn't need to be parenthesized, since it's an assignment.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1505,6 +1505,75 @@ class GLSLGenerator {
...
+                    op_a = regs.GetRegisterAsInteger(instr.gpr8);
+                    if (neg)
+                        op_a = "-(" + op_a + ')';
+                    op_b = regs.GetRegisterAsInteger(instr.gpr20);
+                    op_c = std::to_string(instr.lea.r1.entry_a);
+                    break;
+                }
+
+                case OpCode::Id::LEA_IMM: {
+                    neg = (instr.lea.imm.neg != 0);
```
lioncash: ditto

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1505,6 +1505,75 @@ class GLSLGenerator {
...
+                    op_b = regs.GetRegisterAsInteger(instr.gpr8);
+                    if (neg)
+                        op_b = "-(" + op_b + ')';
+                    op_a = std::to_string(instr.lea.imm.entry_a);
+                    op_c = std::to_string(instr.lea.imm.entry_b);
+                    break;
+                }
+
+                case OpCode::Id::LEA_RZ: {
+                    neg = (instr.lea.rz.neg != 0);
```
lioncash: ditto

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1505,6 +1505,75 @@ class GLSLGenerator {
...
+                case OpCode::Id::LEA_HI:
+                default: {
+                    op_b = regs.GetRegisterAsInteger(instr.gpr8);
+                    op_a = std::to_string(instr.lea.imm.entry_a);
+                    op_c = std::to_string(instr.lea.imm.entry_b);
+                    LOG_CRITICAL(HW_GPU, "Unhandled LEA subinstruction: {}", opcode->GetName());
+                    UNREACHABLE();
+                }
+                }
+                if ((instr.lea.pred48 != static_cast<u64>(Pred::UnusedIndex)) &&
```
lioncash: These two conditions also don't need to be individually parenthesized,

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2087,6 +2156,28 @@ class GLSLGenerator {
             }
             break;
         }
+        case OpCode::Type::PredicateSetRegister: {
+            // TODO: when the bf flag is set, the instruction generates random values
```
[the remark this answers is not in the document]
FernandoS27: I'll prepare a set of results, but they always differ on execution.

FernandoS27:
Changes made.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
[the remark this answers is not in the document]
Subv: The test was flawed, the assembler was using an incorrect bit for the BF flag.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1505,6 +1505,75 @@ class GLSLGenerator {
...
+                case OpCode::Id::LEA_HI:
+                default: {
+                    op_b = regs.GetRegisterAsInteger(instr.gpr8);
+                    op_a = std::to_string(instr.lea.imm.entry_a);
+                    op_c = std::to_string(instr.lea.imm.entry_b);
+                    LOG_CRITICAL(HW_GPU, "Unhandled LEA subinstruction: {}", opcode->GetName());
+                    UNREACHABLE();
+                }
+                }
+                if (instr.lea.pred48 != static_cast<u64>(Pred::UnusedIndex) &&
```
Subv: It seems github removed my comment about how pred48 can not be NeverExecute.

FernandoS27: well NeverExecute is actualy PredicateAlwaysFalse. Either way, both aren't writeable.

Subv: Yes, and it's defined as a 4-bit field with value 0xF. `instr.lea.pred48` is a 3 bit field, if you truncate 0xF to 3 bits you get 7 which is Pred::UnusedIndex.

Remove the check for NeverExecute, it's useless.

PR #1293, opened 2018-09-11
externals: Place font data within cpp files

lioncash:
This places the font data within cpp files, which mitigates the possibility of the font data being duplicated within the binary if it's referred to in more than one translation unit in the future. It also stores the data within a std::array, which is more flexible when it comes to operating with the standard library. It also puts proper header guards in the headers.

Furthermore, it makes the data arrays const. This is what we want, as it allows the compiler to store the data within the read-only segment. As it is, having several large sections of mutable data like this just leaves spots in memory that we can accidentally write to (via accidental overruns, what have you) and actually have it work. This ensures the font data remains the same no matter what.

This was generated with the following Python script that I made [here](https://gist.github.com/lioncash/d0e1fd8f8390f2462ba4361ed0236f16)


lioncash:
Failure is unrelated to these changes

PR #1294, opened 2018-09-11
gl_rasterizer: Use ARB_texture_storage.

degasus:
It allows us to use texture views and it reduces the overhead within the GPU driver.

But it disallows us to reallocate the texture, but we don't do so anyways.

In the end, it is the new way to allocate textures, so there is no need to use the old way.


degasus:
This conflicts with #1287

PR #1311, opened 2018-09-13
Optimized Texture Swizzling

FernandoS27:
So while I was studying the swizzling algorithm for the gpu decoders, I noticed it could be optimized a lot. So I made a fast version of the normal swizzling algorithm. It reduces Surface Load by around +-60%, it has a catch though, it does not work on surfaces with a BPP(Bytes per Pixel) of 3. For this cases, we fallback to the old implementation.

### DIFF
--- a/src/video_core/textures/decoders.cpp
+++ b/src/video_core/textures/decoders.cpp
@@ -46,6 +46,48 @@ void CopySwizzledData(u32 width, u32 height, u32 bytes_per_pixel, u32 out_bytes_
     }
 }
 
+template <std::size_t N, std::size_t M>
+struct alignas(64) SwizzleTable {
+    constexpr SwizzleTable() {
+        for (u32 y = 0; y < N; ++y) {
+            for (u32 x = 0; x < M; ++x) {
+                const u32 x2 = x * 16;
+                values[y][x] = static_cast<u16>(((x2 % 64) / 32) * 256 + ((y % 8) / 2) * 64 +
+                                                ((x2 % 32) / 16) * 32 + (y % 2) * 16);
+            }
+        }
+    }
+    const std::array<u16, M>& operator[](std::size_t index) const {
+        return values[index];
+    }
+    std::array<std::array<u16, M>, N> values{};
+};
+
+constexpr auto swizzle_table = SwizzleTable<8, 4>();
+
+void FastSwizzleData(u32 width, u32 height, u32 bytes_per_pixel, u8* swizzled_data,
+                     u8* unswizzled_data, bool unswizzle, u32 block_height) {
+    std::array<u8*, 2> data_ptrs;
+    const std::size_t stride{width * bytes_per_pixel};
+    const std::size_t image_width_in_gobs{(stride + 63) / 64};
+    const std::size_t copy_size{16};
+    for (std::size_t y = 0; y < height; ++y) {
+        const std::size_t initial_gob =
+            (y / (8 * block_height)) * 512 * block_height * image_width_in_gobs +
+            (y % (8 * block_height) / 8) * 512;
+        const std::size_t pixel_base{y * width * bytes_per_pixel};
+        const auto& table = swizzle_table[y % 8];
+        for (std::size_t xb = 0; xb < stride; xb += copy_size) {
+            const std::size_t gob_address{initial_gob + (xb / 64) * 512 * block_height};
+            const std::size_t swizzle_offset{gob_address + table[(xb / 16) % 4]};
+            const std::size_t pixel_index{xb + pixel_base};
+            data_ptrs[unswizzle] = swizzled_data + swizzle_offset;
+            data_ptrs[!unswizzle] = unswizzled_data + pixel_index;
+            std::memcpy(data_ptrs[0], data_ptrs[1], copy_size);
+        }
+    }
+}
+
 u32 BytesPerPixel(TextureFormat format) {
     switch (format) {
     case TextureFormat::DXT1:
@@ -91,8 +133,13 @@ u32 BytesPerPixel(TextureFormat format) {
 std::vector<u8> UnswizzleTexture(VAddr address, u32 tile_size, u32 bytes_per_pixel, u32 width,
                                  u32 height, u32 block_height) {
     std::vector<u8> unswizzled_data(width * height * bytes_per_pixel);
-    CopySwizzledData(width / tile_size, height / tile_size, bytes_per_pixel, bytes_per_pixel,
-                     Memory::GetPointer(address), unswizzled_data.data(), true, block_height);
+    if (bytes_per_pixel % 3 != 0 && (width * bytes_per_pixel) % 16 == 0) {
+        FastSwizzleData(width / tile_size, height / tile_size, bytes_per_pixel,
+                        Memory::GetPointer(address), unswizzled_data.data(), true, block_height);
+    } else {
+        CopySwizzledData(width / tile_size, height / tile_size, bytes_per_pixel, bytes_per_pixel,
+                         Memory::GetPointer(address), unswizzled_data.data(), true, block_height);
+    }
     return unswizzled_data;
 }
 


FernandoS27:
Changes made. Now my code looks gorgeous! Thanks @lioncash

FernandoS27:
This PR can cause heap corruption. I'm closing it until I find the cause of it

src/video_core/textures/decoders.cpp
```diff
@@ -46,6 +46,62 @@ void CopySwizzledData(u32 width, u32 height, u32 bytes_per_pixel, u32 out_bytes_
...
+
+constexpr auto morton_table = MortonTable<8, 4>();
+
+void FastSwizzleData(u32 width, u32 height, u32 bytes_per_pixel, u8* swizzled_data,
+                     u8* unswizzled_data, bool unswizzle, u32 block_height) {
+    std::array<u8*, 2> data_ptrs;
+    const u32 image_width_in_gobs{(width * bytes_per_pixel + 63) / 64};
+    u32 stride{width};
+    u32 i = 0;
+    while (stride % 2 == 0) {
```
degasus: Have you seen any texture with a stride not being a multiple of 16 bytes? If not, please fall back to the slow path for them. GPUs have some stride requirement, and it is often bigger than 4 bytes. So for this swizzle pattern, forcing the stride as a multiple of 16 bytes makes sense.

degasus: And while you're at it, use a better name for "i"

src/video_core/textures/decoders.cpp
```diff
@@ -46,6 +46,62 @@ void CopySwizzledData(u32 width, u32 height, u32 bytes_per_pixel, u32 out_bytes_
...
+    }
+    const std::array<u16, M>& operator[](std::size_t index) const {
+        return values[index];
+    }
+    std::array<std::array<u16, M>, N> values{};
+};
+
+constexpr auto morton_table = MortonTable<8, 4>();
+
+void FastSwizzleData(u32 width, u32 height, u32 bytes_per_pixel, u8* swizzled_data,
```
degasus: Do you think __restrict for both pointers helps here?

degasus:
Honestly, I was afraid seeing the headline of this PR. It sounds like some general code optimizations without looking at the algorithm. But in fact, after reviewing the code, I think it is even close to the hardware than the old implementation. And more important, it isn't more complex than the slow swizzler. 

But you've made some assumption to optimize this function: power-of-two bbp, even width and even height
I think this was a great move, but this is the best time to verify the capabilities of the hardware. If they match (IMO quite likely), we could drop the old swizzler instead.

src/video_core/textures/decoders.cpp
lioncash: Please don't use it if possible (or at all). It's really not worth adding a means to invoke undefined behavior to the exposed interface (for likely minimal gain).

src/video_core/textures/decoders.cpp
```diff
@@ -46,6 +46,62 @@ void CopySwizzledData(u32 width, u32 height, u32 bytes_per_pixel, u32 out_bytes_
...
+        copy_size = bytes_per_pixel;
+    }
+    for (std::size_t y = 0; y < height; ++y) {
+        const std::size_t initial_gob =
+            (y / (8 * block_height)) * 512 * block_height * image_width_in_gobs +
+            (y % (8 * block_height) / 8) * 512;
+        const std::size_t pixel_base{y * width * bytes_per_pixel};
+        const auto& table = morton_table[y % 8];
+        for (std::size_t x = 0; x < width; x += x_step) {
+            const std::size_t x2{bytes_per_pixel * x};
```
degasus: You don't use "x" at all. Might you just loop over:
for (std::size_t x2 = 0; x2 < width*bytes_per_pixel; x += x_step*bytes_per_pixel)

src/video_core/textures/decoders.cpp
```diff
@@ -46,6 +46,62 @@ void CopySwizzledData(u32 width, u32 height, u32 bytes_per_pixel, u32 out_bytes_
...
+            const std::size_t gob_address{initial_gob + (x2 / 64) * 512 * block_height};
+            const std::size_t swizzle_offset{gob_address + table[(x2 / 16) % 4]};
+            const std::size_t pixel_index{x2 + pixel_base};
+            data_ptrs[unswizzle] = swizzled_data + swizzle_offset;
+            data_ptrs[!unswizzle] = &unswizzled_data[pixel_index];
```
degasus: Please use the same syntax for both pointers.

src/video_core/textures/decoders.cpp
```diff
@@ -46,6 +46,62 @@ void CopySwizzledData(u32 width, u32 height, u32 bytes_per_pixel, u32 out_bytes_
...
+    std::array<std::array<u16, M>, N> values{};
+};
+
+constexpr auto morton_table = MortonTable<8, 4>();
+
+void FastSwizzleData(u32 width, u32 height, u32 bytes_per_pixel, u8* swizzled_data,
+                     u8* unswizzled_data, bool unswizzle, u32 block_height) {
+    std::array<u8*, 2> data_ptrs;
+    const u32 image_width_in_gobs{(width * bytes_per_pixel + 63) / 64};
+    u32 stride{width};
```
degasus: Have you missed a times bytes_per_pixel here?

degasus:
While you're at this stuff, do you think you're able to draw an image of the swizzle method? This always makes it much easier for me to understand the code.

src/video_core/textures/decoders.cpp
```diff
@@ -46,6 +46,62 @@ void CopySwizzledData(u32 width, u32 height, u32 bytes_per_pixel, u32 out_bytes_
...
 }
 
+template <std::size_t N, std::size_t M>
+struct alignas(64) MortonTable {
+    constexpr MortonTable() {
+        for (u32 y = 0; y < N; ++y) {
+            for (u32 x = 0; x < M; ++x) {
+                const u32 x2 = x * 16;
+                values[y][x] = static_cast<u16>(((x2 % 64) / 32) * 256 + ((y % 8) / 2) * 64 +
+                                                ((x2 % 32) / 16) * 32 + (y % 2) * 16 + (x2 % 16));
```
degasus: as x2 = x*16, x2%16 is always 0.

src/video_core/textures/decoders.cpp
```diff
@@ -91,8 +147,13 @@ u32 BytesPerPixel(TextureFormat format) {
 std::vector<u8> UnswizzleTexture(VAddr address, u32 tile_size, u32 bytes_per_pixel, u32 width,
                                  u32 height, u32 block_height) {
     std::vector<u8> unswizzled_data(width * height * bytes_per_pixel);
-    CopySwizzledData(width / tile_size, height / tile_size, bytes_per_pixel, bytes_per_pixel,
-                     Memory::GetPointer(address), unswizzled_data.data(), true, block_height);
+    if (bytes_per_pixel % 3 != 0 && width % 2 == 0 && height % 2 == 0) {
```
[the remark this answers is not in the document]
degasus: You should maybe try to also assert on:
`(width * bytes_per_pixel) % 16 == 0`
and
`height % (block_height * 8) == 0`

src/video_core/textures/decoders.cpp
```diff
@@ -46,6 +46,62 @@ void CopySwizzledData(u32 width, u32 height, u32 bytes_per_pixel, u32 out_bytes_
     }
 }
 
+template <std::size_t N, std::size_t M>
+struct alignas(64) MortonTable {
+    constexpr MortonTable() {
+        for (u32 y = 0; y < N; ++y) {
+            for (u32 x = 0; x < M; ++x) {
+                const u32 x2 = x * 16;
+                values[y][x] = static_cast<u16>(((x2 % 64) / 32) * 256 + ((y % 8) / 2) * 64 +
```
degasus: Just a terrible bad sketch of this swizzle here: https://photos.app.goo.gl/MPfXRFgB5aLRhSFaA

src/video_core/textures/decoders.cpp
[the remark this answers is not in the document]
FernandoS27: Zelda breath of the wild has some odd textures that aren't multiples of 16. Some are barely multiples of 2.

FernandoS27:
@degasus changes made.
Well the swizzle method is not morton code, atleast not the traditional one. If you take the table and print the results you'll see that it's similar to morton code but values are ordered linearly rather than in Z-order.

Also. We cannot get rid of the old method because textures with incoherent strides and heights exist, you can easily see them while loading in Zelda BotW. Same case for textures with byte_per_pixels multiples of 3, (R8G8B8 for example).

bunnei:
Looks good to me, thanks. I am fine with using the fast path where applicable, and falling back to the slower existing method for non-2D, etc. un/swizzles.

PR #1317, opened 2018-09-14
Fix some major memory leak problems

balika011:


src/core/hle/kernel/thread.h
```diff
@@ -260,8 +260,6 @@ class Thread final : public WaitObject {
     // available. In case of a timeout, the object will be nullptr.
     std::function<WakeupCallback> wakeup_callback;
 
-    std::shared_ptr<Scheduler> scheduler;
```
lioncash: This should likely be a `weak_ptr` or a regular pointer/reference. We shouldn't be calling back into the system just to get the scheduler instance.

src/core/hle/kernel/session.h
```diff
@@ -22,6 +22,5 @@ class Session final {
 public:
     ClientSession* client = nullptr; ///< The client endpoint of the session.
     ServerSession* server = nullptr; ///< The server endpoint of the session.
-    SharedPtr<ClientPort> port;      ///< The port that this session is associated with (optional).
```
lioncash: Uh no, this is actually kind of important for closing connections from the emulated instance or will be in the future if it's not being used for that case right now.

src/core/hle/kernel/server_session.cpp
```diff
@@ -26,10 +26,6 @@ ServerSession::~ServerSession() {
     // This destructor will be called automatically when the last ServerSession handle is closed by
     // the emulated application.
 
-    // Decrease the port's connection count.
-    if (parent->port)
-        parent->port->ConnectionClosed();
```
lioncash: This will break proper kernel handling of excess connections to the given port (active connections will never decrement).

src/yuzu/main.cpp
```diff
@@ -658,6 +658,7 @@ void GMainWindow::ShutdownGame() {
     emu_frametime_label->setVisible(false);
 
     emulation_running = false;
+    config->Reload();
```
lioncash: The commit should probably explain what this is fixing in particular (i.e. what actually causes the crash without this). We shouldn't be needing to reload the *whole* config after shutting down.

balika011: We need to call `ReloadInputDevices();˙ or UpdatePadCallback will crash because LoadInputDevices won't be called.
Pushed a better fix for that problem.

balika011:
What are the certain circumstances?

src/core/hle/kernel/thread.cpp
```diff
@@ -430,14 +436,16 @@ void Thread::ChangeCore(u32 core, u64 mask) {
...
+        if (auto schlr = scheduler.lock())
+            schlr->RemoveThread(this);
         next_scheduler->AddThread(this, current_priority);
     }
 
     processor_id = *new_processor_id;
 
     // If the thread was ready, unschedule from the previous core and schedule on the new core
-    scheduler->UnscheduleThread(this, current_priority);
+    if (auto schlr = scheduler.lock())
```
Subv: You can make a helper member function that retrieves the scheduler. lock() should always succeed so you can ASSERT that there, too.

src/core/hle/service/sm/sm.h
```diff
@@ -26,14 +26,12 @@ class Controller;
...
 public:
-    explicit SM(std::shared_ptr<ServiceManager> service_manager);
+    explicit SM();
     ~SM() override;
 
 private:
     void Initialize(Kernel::HLERequestContext& ctx);
     void GetService(Kernel::HLERequestContext& ctx);
-
-    std::shared_ptr<ServiceManager> service_manager;
```
Subv: You can turn this into a weak_ptr instead of making the class access global state.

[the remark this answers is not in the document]
balika011: At first that was why ServiceManager was kept in memory because SM was kept in the memory. It was fixed as i fixed the problem in ServiceFrameworkBase, but i left it this way. It can be reverted now, and i will.

bunnei:
Needs a rebase, as well as a re-review from @Subv and @lioncash

src/core/hle/kernel/thread.cpp
Subv: ping on this comment

DarkLordZach:
@balika011 please rebase

CarlKenner:
I've been working on the thread / scheduler memory leak issue independently, because I didn't realise it was a known issue with an existing patch. You fixed it the same way I was thinking of fixing it.

I'm impressed with your work. But if you'll forgive some nitpicking...

After giving it some more thought, I think I'd go with making the Cpu core object explicitly own the Scheduler (like it does the ARM), and use naked Scheduler* pointers or Scheduler& references everywhere else. The Cpu core class would either have a unique_ptr, or just a naked pointer that is explicitly deleted in the destructor and marked as owned with a comment. I don't think it ever makes sense for the Scheduler to outlive the Cpu core, and we're freeing the Cpu in only one place where we want the Scheduler to be unambiguously freed. Preventing crashes if our design logic is wrong by giving objects an extra lease of life sounds like a good idea, but actually, it just makes things harder to debug. If we are using objects after we're supposed to have shut down, it's actually better for us for that to crash spectacularly so we can find and fix it, IMHO. And of course we want things to be clean and efficient. All the checks we had to add for the weak pointers made it slightly messy and confusing, IMHO.

Another change I'd make, is changing the order in which things are Shutdown in KernelCore::Impl::Shutdown (kernel.cpp), so the process list and current process are freed AFTER the handle_table and thread_wakeup_callback_handle_table. Because that's the order in which they currently are freed in real life, due to the thread tables owning a reference to their Process. I think we should make the code order match what's actually happening for better readability.

I can confirm the UpdatePadCallback crash, so I'm glad you fixed that.

I haven't looked into the other memory leak or your fix for that, so I can't comment on that yet.

balika011:
@DarkLordZach feel free to do so

Hexagon12:
Superseded by #1890 + #1901.

PR #1319, opened 2018-09-14
audio_core: Replace includes with forward declarations where applicable.

lioncash:
Avoids propagating includes in headers where avoidable.


lioncash:
done

PR #1322, opened 2018-09-14
gl_rasterizer_cache: Implement cubemap textures.

bunnei:
There are a lot of changes here - please review commit by commit. I'm fairly confident that this is not quite right, see various TODOs (and open to suggestions or ideas). It's mostly working though. Summary of changes:
* We now make sure compatible surfaces are the same texture target and depth. This was left out of the cache lookup before because it broke things.
* Adds a workaround code path for using Texture2DArray textures (with depth of 1) as Texture2D when the shader sampler is not using array. This fixes a rendering issue in SMO.
* Implements TextureType::TextureCube mode in TEXS shader instruction.
* Implements LoadGLBuffer for Texture2DArray (somewhat unrelated to the rest of the changes here, but we needed the same code for TextureCubemap).
* Updates BlitTextures function to support non-Texture2D surfaces.
* Implements render target -> TextureCubemap reuse by copying the render target to each of the 6 cubemap surfaces. This is used heavily by SMO, and fixes rendering issues. This is likely wrong - we need to figure out which face this is supposed to be copied to (this is the main thing that I was not quite certain about). This likely explains some of the artifacts we still see with cubemaps.

Example rendering improvements with these changes:
image
image
image


greggameplayer:
@bunnei Rebase please

bunnei:
I think that we've worked out enough of the issues with this that it's pretty close to being mergable. Going to tag canary - please speak up if this causes any regressions [including both accuracy and perf].

bunnei:
@degasus we could use a bit of OpenGL guidance here. The specific usecase that this change has been tested with is SMO rendering 6 color attachments in a single pass, and then reusing it as a cubemap texture. To do the "reuse" we need to either blit (or copy) each of the 6 attachments to each face of the cube. However, for this to work, we need to remember each attachment that each face was rendered with for the blit/copy (see https://github.com/yuzu-emu/yuzu/pull/1322/files#diff-91973f62f37e3ace9aaadc4bb5ff94cfR460, for example, as well as https://github.com/yuzu-emu/yuzu/pull/1322/files#diff-91973f62f37e3ace9aaadc4bb5ff94cfR582 for the copy version of this code). Is this right? It seems like it shouldn't matter what the original attachment was that these were rendered with - since we are just using scratch surfaces here for copying/blitting. Any suggestions on how to implement this without having to "remember" the original attachment?

bunnei:
Reminder that I could use some reviews here :P

Hexagon12:
TLOZ: Breath of the Wild seems to crash with this PR on the BlitSurface function.

Steps to reproduce:
1. Use master + this PR
2. Load the game up until the save select screen.
3. The game should crash as soon as you confirm on the save load.

Call stack:
fiksavimas

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -413,47 +448,203 @@ static bool BlitTextures(GLuint src_tex, const MathUtil::Rectangle<u32>& src_rec
...
 
-    if (type == SurfaceType::ColorTexture) {
-        glFramebufferTexture2D(GL_READ_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, src_tex,
-                               0);
-        glFramebufferTexture2D(GL_READ_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0,
-                               0);
+    if (src_params.type == SurfaceType::ColorTexture) {
+        switch (src_params.target) {
+        case SurfaceParams::SurfaceTarget::Texture2D:
+            glFramebufferTexture2D(GL_READ_FRAMEBUFFER, GL_COLOR_ATTACHMENT0 + src_attachment,
```
ReinUsesLisp: I think it'd be better to just use GL_COLOR_ATTACHMENT0 to blit textures. If that gives trouble, you can store 6 texture views (ARB_texture_view) to the cubemap and blit one of those instead. That way you can just send a texture handle and remove `src_attachment`.

bunnei:
Adding canary-merge tag back, as the last commit should prevent BotW from crashing yuzu.

slashiee:
Splatoon 2 crashes on the PR just like how BotW did before: on the BlitSurface function. I'm using v4.0.0 with a save file that has access to the plaza.

Steps to reproduce:
1. On the PR, load up the game
2. The game will crash when loading the plaza

bunnei:
@slashiee very helpful - thanks for testing. The latest iteration should fix the issues with Splatoon 2.

PR #1331, opened 2018-09-16
Implement ASTC_2D_8X8 

raven02:
Heavily used by Bayonetta 2 

Compared with master .

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.h
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.h
@@ -70,19 +70,20 @@ struct SurfaceParams {
         RG8S = 42,
         RG32UI = 43,
         R32UI = 44,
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -253,6 +255,7 @@ struct SurfaceParams {
             16,  // RG8S
             64,  // RG32UI
             32,  // R32UI
+            16,  // ASTC_2D_8X8
```
greggameplayer: @raven02 It should be 32 because you're using RGBA8 which are 8 Red Pixels + 8 Green Pixels + 8 Blue Pixels + 8 Alpha Pixels

Dantttt: astc is a special case,the code is not wrong

PR #1334, opened 2018-09-17
Invalid default value of username in yuzu_cmd

PhiBabin:
The default username value for yuzu_cmd was "" instead of  "yuzu".
It caused the creation of an userprofile with uninitialized stack data as username.
The username is never used, so it's not really an issue currently.

I also added a nullptr check in `RegisteredCacheUnion`. The cache can contains nullptr entries.

### DIFF
--- a/src/core/file_sys/registered_cache.cpp
+++ b/src/core/file_sys/registered_cache.cpp
@@ -555,9 +555,11 @@ VirtualFile RegisteredCacheUnion::GetEntryUnparsed(RegisteredCacheEntry entry) c
 
 VirtualFile RegisteredCacheUnion::GetEntryRaw(u64 title_id, ContentRecordType type) const {
     for (const auto& c : caches) {
-        const auto res = c->GetEntryRaw(title_id, type);
-        if (res != nullptr)
-            return res;
+        if (c != nullptr) {
+            const auto res = c->GetEntryRaw(title_id, type);
+            if (res != nullptr)
+                return res;
+        }
     }
 
     return nullptr;
--- a/src/core/file_sys/registered_cache.cpp
+++ b/src/core/file_sys/registered_cache.cpp
@@ -555,11 +555,12 @@ VirtualFile RegisteredCacheUnion::GetEntryUnparsed(RegisteredCacheEntry entry) c
 
 VirtualFile RegisteredCacheUnion::GetEntryRaw(u64 title_id, ContentRecordType type) const {
     for (const auto& c : caches) {
-        if (c != nullptr) {
-            const auto res = c->GetEntryRaw(title_id, type);
-            if (res != nullptr)
-                return res;
+        if (c == nullptr) {
+            continue;
         }
+        const auto res = c->GetEntryRaw(title_id, type);
+        if (res != nullptr)
+            return res;
     }
 
     return nullptr;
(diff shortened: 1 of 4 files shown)


src/yuzu_cmd/config.cpp
```diff
@@ -125,6 +125,8 @@ void Config::ReadValues() {
 
     // System
     Settings::values.use_docked_mode = sdl2_config->GetBoolean("System", "use_docked_mode", false);
+    Settings::values.username = sdl2_config->Get("System", "username", "yuzu");
+    Settings::values.username = Settings::values.username == "" ? "yuzu" : Settings::values.username;
```
lioncash: There's no real need to use a ternary here if the else case is just a no-op

```cpp
if (Settings::values.username.empty()) {
    Settings::values.username = "yuzu";
}
```

src/core/file_sys/registered_cache.cpp
```diff
@@ -555,9 +555,11 @@ VirtualFile RegisteredCacheUnion::GetEntryUnparsed(RegisteredCacheEntry entry) c
 
 VirtualFile RegisteredCacheUnion::GetEntryRaw(u64 title_id, ContentRecordType type) const {
     for (const auto& c : caches) {
-        const auto res = c->GetEntryRaw(title_id, type);
-        if (res != nullptr)
-            return res;
+        if (c != nullptr) {
```
lioncash: this can be inverted like:

```cpp
if (c == nullptr) {
    continue;
}
```

Which avoids the need to nest (and makes it nicer for breakpoints during debugging)

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -92,6 +92,8 @@ ResultCode ProfileManager::CreateNewUser(UUID uuid, const ProfileUsername& usern
 /// we're loading a string straight from the config
 ResultCode ProfileManager::CreateNewUser(UUID uuid, const std::string& username) {
     ProfileUsername username_output;
```
lioncash: This should likely be:

```cpp
ProfileUsername username_output{};
```

which zeroes out the data, making the assertion unnecessary (which shouldn't be used as a bounds checking mechanism for user input anyway).

PhiBabin: It won't help, since `CreateNewUser(UUID uuid, const ProfileUsername& username)` does not accept a empty string and will return an error (src/core/hle/service/acc/profile_manager.cpp:75). 
The error is not check by the constructor of `ProfileManager()`, thus no user will be created and no error will be throw.
One possibility is to add an assert in that constructor.

src/yuzu_cmd/config.cpp
lioncash: This still applies

src/core/hle/service/acc/profile_manager.cpp
lioncash: It's likely more suited for the constructor. than within this function (it's part of the system behavior to return an error in this case, so it makes no sense to directly assert valid behavior), so the check should be placed around the return value of the function, not in it. The data should also be zeroed out to make that case deterministic, rather than letting it occupy whatever remnant junk data is on the stack.

src/core/file_sys/registered_cache.cpp
```diff
@@ -555,6 +555,9 @@ VirtualFile RegisteredCacheUnion::GetEntryUnparsed(RegisteredCacheEntry entry) c
 
 VirtualFile RegisteredCacheUnion::GetEntryRaw(u64 title_id, ContentRecordType type) const {
     for (const auto& c : caches) {
+        if (c == nullptr) {
```
DarkLordZach: Under normal circumstances RCU cannot contain nullptr entries. It does in yuzu-cmd because of a missing call to CreateFactories. Simply adding a nullptr check without addressing the actual issue will lead to stange errors in the future (missing sysarchives/dlc for one). I have added the necessary init call in #1337, please remove nullptr check.

PR #1340, opened 2018-09-17
Implement a Better Ignore Assert

FernandoS27:
**DO NOT MERGE**

This one prints the assert message to the log unlike #615. At least it helps us figure if an assert was broken from user log.

### DIFF
--- a/src/common/assert.cpp
+++ b/src/common/assert.cpp
@@ -4,8 +4,4 @@
 
 #include "common/assert.h"
 
-#include "common/common_funcs.h"
-
-void assert_handle_failure() {
-    Crash();
-}
+void assert_handle_failure() {}


src/common/assert.h
```diff
@@ -28,18 +28,14 @@ __declspec(noinline, noreturn)
...
+    }
 
 #define ASSERT_MSG(_a_, ...)                                                                       \
-    do                                                                                             \
-        if (!(_a_)) {                                                                              \
-            assert_noinline_call([&] { LOG_CRITICAL(Debug, "Assertion Failed!\n" __VA_ARGS__); }); \
-        }                                                                                          \
-    while (0)
+    if (!(_a_)) {                                                                                  \
+        LOG_CRITICAL(Debug, "Assertion Failed!\n" __VA_ARGS__);                                    \
```
FearlessTobi: Is the `\n` needed?

FernandoS27: ¯\_(ツ)_/¯ it was there, i just kept it that way.

FearlessTobi: You should change it then. As far as I understand, it's not needed.

FernandoS27: done.

greggameplayer: @FearlessTobi \n is use to put the crash explanation in a new line to make it more readable

ogniK5377:
I disagree with this PR mainly because when you have a chain of unimplemented things passing data to each other you can end up with emulator crashes due to uninitialized/bad memory from functions which don't return info which they're suppose to

FernandoS27:
@ogniK5377 this PR is meant to replace #615 which is notrmaly merged into Canary. The only difference between this one amnd the other is that this one prints the error messages.

ogniK5377:
I know but trying to print messages which are full of uninitialized data will lead to emulator crashes whereas if the message isn't printed the game will continue to run just fine

bunnei:
Superseded

PR #1347, opened 2018-09-18
Implemented fatal:u properly

ogniK5377:
fatal:u now is properly implemented with all the ipc cmds. Error reports/Crash reports are also now implemented for fatal:u. Crash reports save to yuzu/logs/crash_reports/
The register dump is currently known as sys modules send all zeros. If there are any non zero values for the "registers" or the unknown values, let me know!

### DIFF
--- a/src/core/hle/service/fatal/fatal.cpp
+++ b/src/core/hle/service/fatal/fatal.cpp
@@ -2,11 +2,15 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <array>
+#include <cstring>
+#include <ctime>
 #include <fmt/time.h>
 #include "common/common_paths.h"
 #include "common/file_util.h"
 #include "common/logging/log.h"
 #include "common/scm_rev.h"
+#include "common/swap.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/process.h"
 #include "core/hle/service/fatal/fatal.h"
@@ -46,20 +50,8 @@ enum class FatalType : u32_le {
     ErrorScreen = 2,
 };
 
-void GenerateErrorReport(ResultCode error_code, const FatalInfo& info) {
-    auto title_id = Core::CurrentProcess()->program_id;
-
-    const std::string crashreport_dir =
-        FileUtil::GetUserPath(FileUtil::UserPath::LogDir) + "crash_logs";
-
-    FileUtil::CreateFullPath(crashreport_dir);
-
-    const std::time_t t = std::time(nullptr);
-
-    const std::string crashreport_filename =
-        fmt::format("{}/{:016x}-{:%F-%H%M%S}.log", crashreport_dir, title_id, *std::localtime(&t));
-
-    auto file = FileUtil::IOFile(crashreport_filename, "wb");
+static void GenerateErrorReport(ResultCode error_code, const FatalInfo& info) {
+    const auto title_id = Core::CurrentProcess()->program_id;
     std::string crash_report =
         fmt::format("Yuzu {}-{} crash report\n"
                     "Title ID:                        {:016x}\n"
@@ -95,12 +87,26 @@ void GenerateErrorReport(ResultCode error_code, const FatalInfo& info) {
 
     LOG_ERROR(Service_Fatal, "{}", crash_report);
 
-    if (!file.IsOpen()) {
-        LOG_ERROR(Service_Fatal, "Failed to save error report to {}", crashreport_filename);
-    } else {
+    const std::string crashreport_dir =
+        FileUtil::GetUserPath(FileUtil::UserPath::LogDir) + "crash_logs";
+
+    if (!FileUtil::CreateFullPath(crashreport_dir)) {
+        LOG_ERROR(
+            Service_Fatal,
+            "Unable to create crash report directory. Possible log directory permissions issue.");
+        return;
+    }
+
+    const std::time_t t = std::time(nullptr);
+    const std::string crashreport_filename =
+        fmt::format("{}/{:016x}-{:%F-%H%M%S}.log", crashreport_dir, title_id, *std::localtime(&t));
+
+    auto file = FileUtil::IOFile(crashreport_filename, "wb");
+    if (file.IsOpen()) {
         file.WriteString(crash_report);
-        file.Close();
         LOG_ERROR(Service_Fatal, "Saving error report to {}", crashreport_filename);
+    } else {
+        LOG_ERROR(Service_Fatal, "Failed to save error report to {}", crashreport_filename);
     }
 }
 
@@ -113,8 +119,8 @@ void ThrowFatalError(ResultCode error_code, FatalType fatal_type, const FatalInf
     case FatalType::ErrorScreen:
         UNREACHABLE();
         break;
-    case FatalType::ErrorReport: // Should not throw a fatal screen but should generate an error
-                                 // report
+        // Should not throw a fatal screen but should generate an error report
+    case FatalType::ErrorReport:
         GenerateErrorReport(error_code, info);
         break;
     };
--- a/src/core/hle/service/fatal/fatal.cpp
+++ b/src/core/hle/service/fatal/fatal.cpp
@@ -2,8 +2,13 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <fmt/time.h>
+#include "common/common_paths.h"
+#include "common/file_util.h"
 #include "common/logging/log.h"
+#include "common/scm_rev.h"
 #include "core/hle/ipc_helpers.h"
+#include "core/hle/kernel/process.h"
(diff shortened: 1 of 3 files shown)


src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +20,139 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+static_assert(sizeof(FatalInfo) == 0x250, "FatalInfo is an invalid size");
+
+enum class FatalType : u32_le {
+    ErrorReportAndScreen = 0,
+    ErrorReport = 1,
+    ErrorScreen = 2,
+};
+
+void GenerateErrorReport(ResultCode error_code, const FatalInfo& info) {
+    auto title_id = Core::CurrentProcess()->program_id;
```
lioncash: This can be const.

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +20,139 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+        FileUtil::GetUserPath(FileUtil::UserPath::LogDir) + "crash_logs";
+
+    FileUtil::CreateFullPath(crashreport_dir);
+
+    const std::time_t t = std::time(nullptr);
+
+    const std::string crashreport_filename =
+        fmt::format("{}/{:016x}-{:%F-%H%M%S}.log", crashreport_dir, title_id, *std::localtime(&t));
+
+    auto file = FileUtil::IOFile(crashreport_filename, "wb");
```
lioncash: This should probably be moved down to the conditional where its success status is checked. This doesn't need to be opened before the text to push out is made.

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +20,139 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+    ErrorScreen = 2,
+};
+
+void GenerateErrorReport(ResultCode error_code, const FatalInfo& info) {
+    auto title_id = Core::CurrentProcess()->program_id;
+
+    const std::string crashreport_dir =
+        FileUtil::GetUserPath(FileUtil::UserPath::LogDir) + "crash_logs";
+
+    FileUtil::CreateFullPath(crashreport_dir);
```
lioncash: Perhaps change this to:

```cpp
if (!FileUtil::CreateFullPath(crashreport_dir)) {
    LOG_ERROR(Service_Fatal, "Unable to create crash report directory. Possible log directory permissions issue.");
    return;
}
```

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +20,139 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+};
+static_assert(sizeof(FatalInfo) == 0x250, "FatalInfo is an invalid size");
+
+enum class FatalType : u32_le {
+    ErrorReportAndScreen = 0,
+    ErrorReport = 1,
+    ErrorScreen = 2,
+};
+
+void GenerateErrorReport(ResultCode error_code, const FatalInfo& info) {
```
lioncash: This can be static.

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +20,139 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+        crash_report += fmt::format("Unknown 10:                      0x{:016x}\n", info.unk10);
+    }
+
+    LOG_ERROR(Service_Fatal, "{}", crash_report);
+
+    if (!file.IsOpen()) {
+        LOG_ERROR(Service_Fatal, "Failed to save error report to {}", crashreport_filename);
+    } else {
+        file.WriteString(crash_report);
+        file.Close();
```
lioncash: This doesn't need to be explicitly closed, this'll happen automatically.

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +20,139 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+        }
+        crash_report += fmt::format("\nUnknown 7:                       0x{:016x}\n", info.unk7);
+        crash_report += fmt::format("Unknown 8:                       0x{:016x}\n", info.unk8);
+        crash_report += fmt::format("Unknown 9:                       0x{:016x}\n", info.unk9);
+        crash_report += fmt::format("Unknown 10:                      0x{:016x}\n", info.unk10);
+    }
+
+    LOG_ERROR(Service_Fatal, "{}", crash_report);
+
+    if (!file.IsOpen()) {
```
lioncash: This can be inverted in terms of the true case

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +20,139 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+}
+
+void ThrowFatalError(ResultCode error_code, FatalType fatal_type, const FatalInfo& info) {
+    LOG_ERROR(Service_Fatal, "Threw fatal error type {}", static_cast<u32>(fatal_type));
+    switch (fatal_type) {
+    case FatalType::ErrorReportAndScreen:
+        GenerateErrorReport(error_code, info);
+        [[fallthrough]];
+    case FatalType::ErrorScreen:
+        UNREACHABLE();
```
lioncash: This is currently reachable with the above error category. Should this be UNIMPLEMENTED instead?

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +20,139 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+        break;
+    case FatalType::ErrorReport: // Should not throw a fatal screen but should generate an error
```
lioncash: The comment can go on top of the case to avoid wonky wrapping.

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -2,8 +2,13 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <fmt/time.h>
```
lioncash: `<array>`, `<cstring>`, `<ctime>`, and `"common/swap.h"` need to be included

src/core/hle/service/fatal/fatal.cpp
ogniK5377: UNREACHABLE here is so execution is stopped. Since we don't have an error screen at all we ignore it.

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +24,142 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+    std::array<u64_le, 32> backtrace{};
+    u64_le unk7{};
+    u64_le unk8{};
+    u32_le backtrace_size{};
+    u32_le unk9{};
+    u32_le unk10{}; // TODO(ogniK): Is this even used or is it just padding?
+};
+static_assert(sizeof(FatalInfo) == 0x250, "FatalInfo is an invalid size");
+
+enum class FatalType : u32_le {
```
Subv: This kind of inheritance will break in non-LE platforms

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +24,142 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+    auto file = FileUtil::IOFile(crashreport_filename, "wb");
+    if (file.IsOpen()) {
+        file.WriteString(crash_report);
+        LOG_ERROR(Service_Fatal, "Saving error report to {}", crashreport_filename);
+    } else {
+        LOG_ERROR(Service_Fatal, "Failed to save error report to {}", crashreport_filename);
+    }
+}
+
+void ThrowFatalError(ResultCode error_code, FatalType fatal_type, const FatalInfo& info) {
```
Subv: this can be static

src/core/hle/service/fatal/fatal.cpp
```diff
@@ -15,16 +24,142 @@ Module::Interface::Interface(std::shared_ptr<Module> module, const char* name)
...
+    rb.Push(RESULT_SUCCESS);
+}
+
 void Module::Interface::ThrowFatalWithPolicy(Kernel::HLERequestContext& ctx) {
+    LOG_ERROR(Service_Fatal, "called");
     IPC::RequestParser rp(ctx);
-    u32 error_code = rp.Pop<u32>();
-    LOG_WARNING(Service_Fatal, "(STUBBED) called, error_code=0x{:X}", error_code);
+    auto error_code = rp.Pop<ResultCode>();
+    auto fatal_type = rp.PopRaw<FatalType>();
```
Subv: Don't we have a PopEnum function?

PR #1360, opened 2018-09-19
Port citra-emu/citra#3979 game_list: move SearchField to game_list_p.h and fix untranslated text

FearlessTobi:
See citra-emu/citra#3979 for more details. 

Original description: `Previously the whole filter is left untranslated. After a bit of investigating, I found that my IDE says that reference to "tr" was ambigious. I looked at the ts file and found that these texts are in the context of GameList. Probably the default tr function called was not in GameList class. This explicitly set it and fixed the untranslated text.`

### DIFF
--- a/src/yuzu/game_list.cpp
+++ b/src/yuzu/game_list.cpp
@@ -26,10 +26,10 @@
 #include "yuzu/main.h"
 #include "yuzu/ui_settings.h"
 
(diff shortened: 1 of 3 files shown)


bunnei:
Needs a rebase.

FearlessTobi:
Rebased.

zhaowenlan1779:
Actually, the original description is not very accurate. the thing got changed after addressing reviews.

PR #1369, opened 2018-09-20
Set default windows size to native res 1280x720

raven02:
I do believe that the current SetDefaultUIGeometry() in yuzu is cater for 3DS which has upper and lower screen .This change is to set the windows size as native res of switch 1280x720 when startup .


ogniK5377:
What about users with smaller screens?

lioncash:
also using raw pixel values doesn't take into account the pixel scaling for users with Hi-DPI monitors.

PR #1374, opened 2018-09-21
Added support for uncompressed NSOs

ogniK5377:
Unable to use a bitfield for flags since it won't be a trivial copy

### DIFF
--- a/src/core/loader/nso.cpp
+++ b/src/core/loader/nso.cpp
@@ -39,6 +39,11 @@ struct NsoHeader {
     u32_le bss_size;
     INSERT_PADDING_BYTES(0x1c);
     std::array<u32_le, 3> segments_compressed_size;
+
+    bool IsSegmentCompressed(size_t segment_num) const {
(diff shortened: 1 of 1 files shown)


src/core/loader/nso.cpp
```diff
@@ -105,9 +107,11 @@ VAddr AppLoader_NSO::LoadModule(FileSys::VirtualFile file, VAddr load_base) {
     Kernel::SharedPtr<Kernel::CodeSet> codeset = Kernel::CodeSet::Create(kernel, "");
     std::vector<u8> program_image;
     for (std::size_t i = 0; i < nso_header.segments.size(); ++i) {
-        const std::vector<u8> compressed_data =
+        std::vector<u8> data =
             file->ReadBytes(nso_header.segments_compressed_size[i], nso_header.segments[i].offset);
-        std::vector<u8> data = DecompressSegment(compressed_data, nso_header.segments[i]);
+        if ((nso_header.flags >> i) & 1) {
```
Subv: Add a function inside NsoHeader that checks for this

PR #1377, opened 2018-09-21
Improved Fast Swizzle and Legacy Swizzle

FernandoS27:
FastSwizzle now works with less restrictions and is better documented.  Legacy swizzle is still required for certain types of textures but should be a bit faster now.


FernandoS27:
well I've done more changes after testing and benchmarking. This should be enough to unify both methods and help DMA run faster.

PR #1378, opened 2018-09-21
process: Make a few member variables private where applicable

lioncash:
Mainly to make it easier to reason about certain aspects of a process, and to make it easier to do a few planned changes in the future. This just gets the cleanup out of the way so its not necessary to include it alongside said planned changes.


lioncash:
Failure is just appveyor being appveyor

PR #1380, opened 2018-09-21
shader_bytecode: Make operator== and operator!= of IpaMode const qualified

lioncash:
These don't affect the state of the struct and can be const member functions.


FernandoS27:
Damn I should look more into your PRs, they really teach you appropriate C++.

PR #1383, opened 2018-09-22
game_list: Add Qt SmoothTransformation to picture scaling

DarkLordZach:
As per @jroweboy 's request - makes the icons less crispy at low resolutions.

Example:
less crisp


jroweboy:
@DarkLordZach before and after pic pls ❤️

PR #1384, opened 2018-09-22
Added glObjectLabels for renderdoc for textures and shader programs

ogniK5377:
Special thanks to @ReinUsesLisp for the help

### DIFF
--- a/src/video_core/utils.h
+++ b/src/video_core/utils.h
@@ -161,4 +161,25 @@ static inline void MortonCopyPixels128(u32 width, u32 height, u32 bytes_per_pixe
     }
 }
 
+static void LabelGLObject(GLenum identifier, GLuint handle, VAddr addr) {
+    if (!GLAD_GL_KHR_debug) {
+        return; // We don't need to throw an error as this is just for debugging
+    }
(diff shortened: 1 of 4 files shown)


FernandoS27:
Hmn would be nice if we could add some info to those labels, like texture type (2d, 3d, cube-layer)

ogniK5377:
Added, thanks for the idea @FernandoS27

src/video_core/utils.h
```diff
@@ -161,4 +161,26 @@ static inline void MortonCopyPixels128(u32 width, u32 height, u32 bytes_per_pixe
     }
 }
 
+static void LabelGLObject(GLenum identifier, GLuint handle, VAddr addr,
+                          std::string extra_info = "") {
+    if (!GLAD_GL_KHR_debug) {
+        return; // We don't need to throw an error as this is just for debugging
+    }
+    const std::string nice_addr = fmt::format("0x{:016x}", addr);
+    std::string object_label{};
```
lioncash: `{}` isn't necessary for std::string. It initializes to an empty string by default.

PR #1390, opened 2018-09-23
Added Amiibo support

ogniK5377:
Yuzu now has the ability to read amiibo file dump files! Currently, this is a basic support for reading nfc tags. It would need to be expanded further on to allow writing to amiibos and reading the encrypted amiibo area. To "scan" an amiibo, simply goto `File->Load Amiibo` whilst a game is loaded and pick your amiibo dump! It's as simple as that! Please note this isn't complete amiibo support just yet, but it's a start. A few games work and others don't. More support will be added in the future for more games

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -175,8 +175,6 @@ struct System::Impl {
         GetAndResetPerfStats();
         perf_stats.BeginSystemFrame();
 
-        nfc_activate = Kernel::Event::Create(kernel, Kernel::ResetType::OneShot, "Nfc:ActivateTag");
-
         return ResultStatus::Success;
     }
 
@@ -290,10 +288,6 @@ struct System::Impl {
     std::array<std::unique_ptr<Cpu>, NUM_CPU_CORES> cpu_cores;
     std::array<std::unique_ptr<std::thread>, NUM_CPU_CORES - 1> cpu_core_threads;
     std::size_t active_core{}; ///< Active core, only used in single thread mode
-
-    /// NFC Loading
-    Kernel::SharedPtr<Kernel::Event> nfc_activate;
-    std::string nfc_filename;
 
     /// Service manager
     std::shared_ptr<Service::SM::ServiceManager> service_manager;
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -175,6 +175,8 @@ struct System::Impl {
         GetAndResetPerfStats();
         perf_stats.BeginSystemFrame();
 
+        nfc_activate = Kernel::Event::Create(kernel, Kernel::ResetType::OneShot, "Nfc:ActivateTag");
+
         return ResultStatus::Success;
     }
 
@@ -288,6 +290,10 @@ struct System::Impl {
     std::array<std::unique_ptr<Cpu>, NUM_CPU_CORES> cpu_cores;
     std::array<std::unique_ptr<std::thread>, NUM_CPU_CORES - 1> cpu_core_threads;
     std::size_t active_core{}; ///< Active core, only used in single thread mode
+
+    /// NFC Loading
+    Kernel::SharedPtr<Kernel::Event> nfc_activate;
+    std::string nfc_filename;
 
     /// Service manager
     std::shared_ptr<Service::SM::ServiceManager> service_manager;
@@ -463,6 +469,19 @@ void System::SetFilesystem(std::shared_ptr<FileSys::VfsFilesystem> vfs) {
 
 std::shared_ptr<FileSys::VfsFilesystem> System::GetFilesystem() const {
     return impl->virtual_filesystem;
+}
+
+void System::LoadAmiibo(const std::string& filename) {
+    impl->nfc_filename = filename;
+    impl->nfc_activate->Signal();
+}
+
+Kernel::SharedPtr<Kernel::Event>& System::GetNFCEvent() const {
+    return impl->nfc_activate;
+}
+
+std::string& System::GetNFCFilename() const {
+    return impl->nfc_filename;
 }
 
 System::ResultStatus System::Init(Frontend::EmuWindow& emu_window) {
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -476,11 +476,11 @@ void System::LoadAmiibo(const std::string& filename) {
     impl->nfc_activate->Signal();
 }
 
-Kernel::SharedPtr<Kernel::Event>& System::GetNFCEvent() const {
+const Kernel::SharedPtr<Kernel::Event>& System::GetNFCEvent() const {
     return impl->nfc_activate;
 }
 
-std::string& System::GetNFCFilename() const {
+const std::string& System::GetNFCFilename() const {
     return impl->nfc_filename;
 }
 
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -471,19 +471,6 @@ std::shared_ptr<FileSys::VfsFilesystem> System::GetFilesystem() const {
     return impl->virtual_filesystem;
 }
 
-void System::LoadAmiibo(const std::string& filename) {
-    impl->nfc_filename = filename;
-    impl->nfc_activate->Signal();
-}
-
-const Kernel::SharedPtr<Kernel::Event>& System::GetNFCEvent() const {
-    return impl->nfc_activate;
-}
-
-const std::string& System::GetNFCFilename() const {
-    return impl->nfc_filename;
-}
-
 System::ResultStatus System::Init(Frontend::EmuWindow& emu_window) {
     return impl->Init(emu_window);
 }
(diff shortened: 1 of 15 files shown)


src/core/hle/service/nfp/nfp.cpp
```diff
@@ -82,71 +112,145 @@ class IUser final : public ServiceFramework<IUser> {
...
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(RESULT_SUCCESS);
+    }
+
+    void GetTagInfo(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Service_NFP, "called");
+        TagInfo tag_info{};
+        Core::System& system{Core::System::GetInstance()};
+        auto nfc_file = FileUtil::IOFile(system.GetNFCFilename(), "rb");
+        size_t read_length = nfc_file.ReadBytes(tag_info.uuid.data(), sizeof(tag_info.uuid.size()));
```
Subv: What if the file fails to open?

src/core/core.cpp
```diff
@@ -174,6 +174,8 @@ struct System::Impl {
         GetAndResetPerfStats();
         perf_stats.BeginSystemFrame();
 
+        nfc_activate = Kernel::Event::Create(kernel, Kernel::ResetType::OneShot, "Nfc:ActivateTag");
```
Subv: Does this really have to live in System? that sounds like a pretty heavy leak of responsibilities and a sign of bad design (same with the direct file-opening in the nfc class).

src/core/hle/service/nfp/nfp.cpp
```diff
@@ -94,6 +94,17 @@ class IUser final : public ServiceFramework<IUser> {
...
     static_assert(sizeof(ModelInfo) == 0x40, "ModelInfo is an invalid size");
 
+    struct CommonInfo {
+        u16_be last_write_year;
+        u8 last_write_month;
+        u8 last_write_day;
+        u16_be write_counter;
+        u16_be version;
+        u32_be application_area_size;
+        INSERT_PADDING_BYTES(52);
```
FearlessTobi: We usually use decimal notation for things like this.

src/yuzu/ui_settings.h
```diff
@@ -44,6 +44,7 @@ struct Values {
     QString roms_path;
     QString symbols_path;
     QString gamedir;
+    QString amiibo_path;
```
FearlessTobi: is this actually used anywhere? Yes, it's passed into the open function but wouldn't an empty String be okay as well?

src/core/hle/service/nfp/nfp.cpp
```diff
@@ -83,71 +129,213 @@ class IUser final : public ServiceFramework<IUser> {
...
+        LOG_DEBUG(Service_NFP, "called");
+
+        Core::System& system{Core::System::GetInstance()};
+        auto nfc_file = FileUtil::IOFile(system.GetNFCFilename(), "rb");
+        IPC::ResponseBuilder rb{ctx, 2};
+        if (!nfc_file.IsOpen()) {
+            rb.Push(ErrCodes::ERR_TAG_FAILED);
+            return;
+        }
+        nfc_file.Seek(0x54, SEEK_SET);
```
Subv: What is this offset?

src/core/hle/service/nfp/nfp.cpp
```diff
@@ -83,71 +129,213 @@ class IUser final : public ServiceFramework<IUser> {
...
+        auto nfc_file = FileUtil::IOFile(system.GetNFCFilename(), "rb");
+        if (!nfc_file.IsOpen()) {
+            rb.Push(ErrCodes::ERR_TAG_FAILED);
+            return;
+        }
+        TagInfo tag_info{};
+        size_t read_length = nfc_file.ReadBytes(tag_info.uuid.data(), sizeof(tag_info.uuid.size()));
+        tag_info.uuid_length = static_cast<u8>(read_length);
+
+        tag_info.protocol = 1; // TODO(ogniK): Figure out actual values
```
Subv: Please add a better comment about where these values came from and what they might mean

src/core/hle/service/nfp/nfp.cpp
FearlessTobi: This is the offset of the Amiibo Identification Block.

FearlessTobi:
As already said on Discord, some feedback from https://github.com/citra-emu/citra/pull/4333 might apply here as well.

src/yuzu/main.cpp
```diff
@@ -1274,6 +1279,29 @@ void GMainWindow::OnConfigure() {
     }
 }
 
+void GMainWindow::OnLoadAmiibo() {
+    const QString extensions{"*.bin"};
+    const QString file_filter =
+        tr("Amiibo File") + " (" + extensions + ");;" + tr("All Files (*.*)");
+    const QString filename = QFileDialog::getOpenFileName(
+        this, tr("Load Amiibo"), UISettings::values.amiibo_path, file_filter);
```
FearlessTobi: You can simplify this like I also did for Citra.
```
const QString extensions{"*.bin"};
const QString file_filter = tr("Amiibo File (%1);; All Files (*.*)").arg(extensions);
const QString filename = QFileDialog::getOpenFileName(this, tr("Load Amiibo"), "", file_filter);`

File: src/yuzu/main.cpp [hunk 8]
```diff
@@ -1274,6 +1279,29 @@ void GMainWindow::OnConfigure() {
...
+    if (!filename.isEmpty()) {
+        Core::System& system{Core::System::GetInstance()};
+        Service::SM::ServiceManager& sm = system.ServiceManager();
+        auto nfc = sm.GetService<Service::NFP::Module::Interface>("nfp:user");
+        if (nfc != nullptr) {
+            auto nfc_file = FileUtil::IOFile(filename.toStdString(), "rb");
+            if (!nfc_file.IsOpen()) {
+                return;
+            }
+            std::vector<u8> amiibo_buffer(nfc_file.GetSize());
```
FearlessTobi: Check if the amiibo_buffer matches the size of the struct.

src/core/hle/service/nfp/nfp.cpp
```diff
@@ -156,7 +340,20 @@ void Module::Interface::CreateUserInterface(Kernel::HLERequestContext& ctx) {
     LOG_DEBUG(Service_NFP, "called");
     IPC::ResponseBuilder rb{ctx, 2, 0, 1};
     rb.Push(RESULT_SUCCESS);
-    rb.PushIpcInterface<IUser>();
+    rb.PushIpcInterface<IUser>(*this);
+}
+
+void Module::Interface::LoadAmiibo(const std::vector<u8> amiibo) {
```
FearlessTobi: Lock this function with g_hle_lock.

src/core/hle/service/nfp/nfp.h
```diff
@@ -16,6 +18,13 @@ class Module final {
         ~Interface() override;
 
         void CreateUserInterface(Kernel::HLERequestContext& ctx);
+        void LoadAmiibo(const std::vector<u8> amiibo);
+        const Kernel::SharedPtr<Kernel::Event>& GetNFCEvent() const;
+        const std::vector<u8>& GetAmiiboBuffer() const;
+
+    private:
+        Kernel::SharedPtr<Kernel::Event> nfc_tag_load{};
+        std::vector<u8> amiibo_buffer{};
```
FearlessTobi: You should add a struct for the full Amiibo data and load the buffer into it. It can probably be copy pasted from Citra. See this: https://github.com/citra-emu/citra/pull/4337/files#diff-48850a28647a4211447cc7a507cbbeddR31

src/yuzu/ui_settings.h
FearlessTobi: Question still applies.

src/core/hle/service/nfp/nfp.cpp
```diff
@@ -83,71 +134,204 @@ class IUser final : public ServiceFramework<IUser> {
...
+        rb.Push(RESULT_SUCCESS);
+        rb.PushRaw<u32>(0); // This is from the GetCommonInfo stub
+    }
+
+    bool has_attached_handle{};
+    const u64 device_handle{Common::MakeMagic('Y', 'U', 'Z', 'U')};
+    const u32 npad_id{0}; // Player 1 controller
     State state{State::NonInitialized};
     DeviceState device_state{DeviceState::Initialized};
-    Kernel::SharedPtr<Kernel::Event> activate_event;
```
FearlessTobi: What's wrong with leaving it here?

FearlessTobi:
Also are there any Amiibo functions that require the State to have a special value? This should be either hardware-tested or found out by reverse engineering the nfp module.

src/core/hle/service/nfp/nfp.cpp
ogniK5377: This isn't used anymore, we need to signal this event so we pass the event we're signalling instead

ogniK5377:
> Also are there any Amiibo functions that require the State to have a special value? This should be either hardware-tested or found out by reverse engineering the nfp module.

Afaik I haven't encountered anything like that yet, we don't have too many games with NFC support so we'll need to see in the future

PR #1396, opened 2018-09-25
loader: Add support for packed updates

DarkLordZach:
Add support for updates that are included inside of the game XCI file (commonly known as Rev1/2/3+ carts). The XCI version will only be applied if there is no installed title.

Todo Before Merge:
- [x] Test on more games (will need to manually repack updates in)

Ex:
packed updates


Hexagon12:
Please rebase

PR #1403, opened 2018-09-26
qt: Install System TitleTypes to System NAND

DarkLordZach:
Fixes an issue where installed system archive NCAs would be installed to user NAND and not recognized by games.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1055,11 +1055,21 @@ void GMainWindow::OnMenuInstallToNAND() {
             return;
         }
 
-        if (index >= 5)
-            index += 0x7B;
+        // If index is equal to or past Game, add the jump in TitleType.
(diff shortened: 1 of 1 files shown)


src/yuzu/main.cpp
```diff
@@ -1058,8 +1058,15 @@ void GMainWindow::OnMenuInstallToNAND() {
         if (index >= 5)
             index += 0x7B;
 
-        const auto res = Service::FileSystem::GetUserNANDContents()->InstallEntry(
-            nca, static_cast<FileSys::TitleType>(index), false, qt_raw_copy);
+        FileSys::InstallResult res;
+        if (index >= 0x80) {
```
Subv: I would prefer if we didn't use magic numbers here, I have no idea what '5', '0x7B' or '0x80' mean

PR #1405, opened 2018-09-26
Ignore NODEP texture mode, since it's just a hw-level optimization hint.

MoochMcGee:


MoochMcGee:
Yeah, it called this bit liveOnly, with the comment "only execute on live pixels of a quad (optimization)". This line of code has remained the exact same for 7 years, so it's probably accurate.

MoochMcGee:
This also lets DOOM get past some of the intro videos. After this, it's blocked by an unimplemented shader instruction. I'm investigating that now.

FernandoS27:
you can just use ignore-assert if you are trying to get past DOOM asserts.

MoochMcGee:
I know that, @FernandoS27, but i'm working on trying to actually FIX the problems lol.

pendingchaos:
A nouveau dev made use of this flag and found incorrect rendering when done wrongly, performance improvements and somehow a hang with Payday (I guess it used texlod): https://trello.com/c/bW7SYHtW/56-add-pass-to-set-liveonly-on-tex-instructions-when-possible
So basically it more or less worked as expected.

I think it should be fine to ignore the flag.

ReinUsesLisp:
There's the drawback of a game relying on that `incorrect rendering`. But that shouldn't be an issue since most (if not all) games run through a driver (and it would be a nightmare to emulate).

pendingchaos:
It might be possible to emulate it with gl_HelperInvocation. I don't know what the texture operation should result in for helper invocations though (garbage? the previous value of the register?).

It seems very unlikely that anything would rely on it though. I doubt any compiler would expose the flag and games probably aren't ever using hand-written assembly.

MoochMcGee:
Anybody else wanna put in their thoughts?

Hexagon12:
This PR is superseded by #1927.

PR #1406, opened 2018-09-27
gl_state: Pack sampler bindings into a single ARB_multi_bind

ReinUsesLisp:
My original idea was to just update the intervals that changed, but it proved to be more complex and it would result in more OpenGL calls. This just packs deltas into a single `glBindSamplers`.

### DIFF
--- a/src/video_core/renderer_opengl/gl_state.cpp
+++ b/src/video_core/renderer_opengl/gl_state.cpp
@@ -205,9 +205,6 @@ void OpenGLState::Apply() const {
             glActiveTexture(TextureUnits::MaxwellTexture(static_cast<int>(i)).Enum());
             glBindTexture(texture_unit.target, texture_unit.texture);
         }
-        if (texture_unit.sampler != cur_state_texture_unit.sampler) {
-            glBindSampler(static_cast<GLuint>(i), texture_unit.sampler);
-        }
         // Update the texture swizzle
         if (texture_unit.swizzle.r != cur_state_texture_unit.swizzle.r ||
             texture_unit.swizzle.g != cur_state_texture_unit.swizzle.g ||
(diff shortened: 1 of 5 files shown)


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -219,6 +216,27 @@ void OpenGLState::Apply() const {
         }
     }
 
+    // Samplers
+    {
+        bool any{};
```
bunnei: nit: `has_samplers` is probably a better name?

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -219,6 +216,27 @@ void OpenGLState::Apply() const {
...
+        std::size_t first{}, last{};
+        std::array<GLuint, 32> samplers;
+        for (std::size_t i = 0; i < std::size(samplers); ++i) {
```
bunnei: nit: `samplers.size()`?

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -219,6 +216,27 @@ void OpenGLState::Apply() const {
         }
     }
 
+    // Samplers
+    {
+        bool any{};
+        std::size_t first{}, last{};
+        std::array<GLuint, 32> samplers;
```
bunnei: nit: Use `NumTextureSamplers`? (avoid unnamed constant where possible). Note, after searching for this, it seems like it's in the wrong place - should probably be moved to the other constants defined in maxwell_3d.h.

ReinUsesLisp: It's imported from `gl_state.h`'s `texture_units` which also uses a magic number (32). I could use its `constexpr ::size()` value to avoid having another magic number. Although `gl_state.h` doesn't include `maxwell_3d.h` to remove the first unnamed constant.

src/video_core/renderer_opengl/gl_state.cpp
ReinUsesLisp: Other `for`s in the file use `std::size`. I'll leave it like that for consistency.

PR #1407, opened 2018-09-27
aoc_u: Implement functions and add support for DLC loading

DarkLordZach:
Implements the necessary aoc:u service commands that games use to list and use add on content (2, 3, 5, 7). Also makes it so `StorageId::None` will read from all locations so DLC can be loaded.

DLC will show in the game list as `DLC(...)` where ... is a list of the DLC IDs that will be loaded. The DLC ID is the title id of that particular DLC bitwise AND with 0x7FF.

If your installed DLC doesn't show, it is most likely titlekeys. Each DLC has a different titlekey than the base game.

dlc1
dlc2
dlc3

### DIFF
--- a/src/core/hle/service/aoc/aoc_u.cpp
+++ b/src/core/hle/service/aoc/aoc_u.cpp
@@ -2,7 +2,9 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
 #include <numeric>
+#include <vector>
 #include "common/logging/log.h"
 #include "core/file_sys/content_archive.h"
 #include "core/file_sys/nca_metadata.h"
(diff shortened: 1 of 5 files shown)


src/core/hle/service/aoc/aoc_u.cpp
```diff
@@ -2,41 +2,114 @@
...
-        {5, nullptr, "GetAddOnContentBaseId"},
+        {5, &AOC_U::GetAddOnContentBaseId, "GetAddOnContentBaseId"},
         {6, nullptr, "PrepareAddOnContentByApplicationId"},
-        {7, nullptr, "PrepareAddOnContent"},
+        {7, &AOC_U::PrepareAddOnContent, "PrepareAddOnContent"},
         {8, nullptr, "GetAddOnContentListChangedEvent"},
     };
     RegisterHandlers(functions);
+
+    // Accumulate AOC title ids
```
lioncash: If you're describing what you're doing, and not why you're doing it with a comment, it means you should be putting the code in a named function.

src/core/hle/service/aoc/aoc_u.cpp
```diff
@@ -2,41 +2,114 @@
...
 void AOC_U::ListAddOnContent(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+
+    const auto offset = rp.PopRaw<u32>();
+    auto count = rp.PopRaw<u32>();
+
+    const auto current = Core::System::GetInstance().CurrentProcess()->program_id;
+
+    std::vector<u32> out;
+    for (size_t i = 0; i < add_on_content.size(); ++i) {
```
lioncash: This can be `std::copy_if` with the use of `std::back_inserter`

src/core/hle/service/aoc/aoc_u.cpp
```diff
@@ -2,41 +2,114 @@
...
 #include "core/hle/service/aoc/aoc_u.h"
+#include "core/hle/service/filesystem/filesystem.h"
+#include "core/loader/loader.h"
 
 namespace Service::AOC {
 
+constexpr u64 DLC_BASE_TITLE_ID_MASK = 0xFFFFFFFFFFFFE000;
+constexpr u64 DLC_BASE_TO_AOC_ID_MASK = 0x1000;
+
+bool CheckAOCTitleIDMatchesBase(u64 base, u64 aoc) {
```
lioncash: This function can be `static`

PR #1409, opened 2018-09-27
crypto: Add support for full key derivation

DarkLordZach:
Supports:
- Deriving SBK from `fuses.bin`
- Deriving keyblobs from `BOOT0`
- Deriving various sources each from `BOOT0`, `sm` (trustzone decrypted), `pkg_decr` (package1 decrypted), and `BCPKG2-1-Normal-Main` (package2).
- Deriving full keys given appropriate source and master key
- Deriving titlekeys given `PRODINFO` and system saves `E1` and `E2`.

In essence, the ability to generate all of the keys required for yuzu given only TSEC, BIS keys 0-2 and `fuses`, `BOOT0`, `pkg_decr`, `sm`, and `BCPKG2-1-Normal-Main` in sysdata.

Adds:
- UI option to rederive keys in help menu
- Derives keys on any run *if*:
    - Keys are missing and
    - New keys can be derived
- PartitionDataManager class to manage decryption/extraction/use of data from system files in sysdata.

Ex: 
deriving


src/core/crypto/key_manager.cpp
```diff
@@ -290,59 +575,437 @@ void KeyManager::DeriveSDSeedLazy() {
...
+    Key128 latest_master{};
+    for (s8 i = 0x1F; i > 0; --i) {
+        if (GetKey(S128KeyType::Master, i) != Key128{}) {
+            latest_master = GetKey(S128KeyType::Master, i);
+            break;
+        }
+    }
+
+    const auto masters = data.GetTZMasterKeys(latest_master);
+    for (size_t i = 0; i < 0x20; ++i) {
```
lioncash: `i < masters.size();`

src/core/crypto/key_manager.cpp
```diff
@@ -290,59 +575,437 @@ void KeyManager::DeriveSDSeedLazy() {
...
+            SetKey(S128KeyType::Master, masters[i], i);
+    }
+
+    DeriveBase();
+
+    if (!data.HasPackage2())
+        return;
+
+    std::array<Key128, 0x20> package2_keys{};
+    for (size_t i = 0; i < 0x20; ++i) {
```
lioncash: `i < package2_keys.size();`

src/core/crypto/key_manager.cpp
```diff
@@ -290,59 +575,437 @@ void KeyManager::DeriveSDSeedLazy() {
...
+                      static_cast<u64>(SourceKeyType::Keyblob), i);
+    }
+
+    if (data.HasFuses())
+        SetKeyWrapped(S128KeyType::SecureBoot, data.GetSecureBootKey());
+
+    DeriveBase();
+
+    Key128 latest_master{};
+    for (s8 i = 0x1F; i > 0; --i) {
```
lioncash: This can still be a `std::size_t` given it'll terminate once it hits zero. If `>=` was intended, then this is a logic bug in the termination check (in which case `i` should be casted to the proper types in the function calls to avoid sign conversion warnings.

src/core/crypto/key_manager.h
```diff
@@ -91,6 +143,8 @@ class KeyManager {
     Key128 GetKey(S128KeyType id, u64 field1 = 0, u64 field2 = 0) const;
     Key256 GetKey(S256KeyType id, u64 field1 = 0, u64 field2 = 0) const;
 
+    Key256 GetBISKey(u8 partition_id) const;
```
lioncash: This doesn't seem to be used anywhere.

src/core/crypto/key_manager.cpp
```diff
@@ -290,59 +575,437 @@ void KeyManager::DeriveSDSeedLazy() {
...
+}
+
+void KeyManager::PopulateFromPartitionData(PartitionDataManager data) {
+    if (!BaseDeriveNecessary())
+        return;
+
+    if (!data.HasBoot0())
+        return;
+
+    for (size_t i = 0; i < 0x20; ++i) {
```
lioncash: Magic number

src/core/crypto/key_manager.h
```diff
@@ -5,11 +5,18 @@
...
 
 #include <array>
+#include <map>
 #include <string>
 #include <boost/container/flat_map.hpp>
 #include <boost/optional.hpp>
 #include <fmt/format.h>
 #include "common/common_types.h"
+#include "core/file_sys/vfs_types.h"
+#include "partition_data_manager.h"
```
lioncash: Our includes use relative paths:

`#include "core/crypto/partition_data_manager.h"`

And I'm going to preface by saying that I usually don't do callouts like this unless it's a repeated thing, but can you *please* be more consistent with this? I've had to point this same thing out over and over across quite a few recent PRs of yours. I usually let it slide for newcomers initially and just point them out because it can take some time to adjust to a codebase, but when it becomes a constant thing I need to point out to the same person, it becomes annoying.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+#include "common/assert.h"
+#include "common/common_types.h"
+#include "common/hex_util.h"
+#include "common/logging/log.h"
+#include "common/string_util.h"
+#include "core/crypto/key_manager.h"
+#include "core/crypto/partition_data_manager.h"
+#include "core/file_sys/vfs.h"
+#include "core/file_sys/vfs_offset.h"
+#include "ctr_encryption_layer.h"
```
lioncash: This should be relative like the others

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+#include "xts_encryption_layer.h"
```
lioncash: This should be relative like the others

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+    u8 priority;
+    u8 core;
+    INSERT_PADDING_BYTES(1);
+    u8 flags;
+    std::array<SectionHeader, 6> sections;
+    std::array<u32, 0x20> capabilities;
+};
+static_assert(sizeof(KIPHeader) == 0x100, "KIPHeader has incorrect size.");
+
+const static std::array<SHA256Hash, 0x10> source_hashes{
```
lioncash: This can just be constexpr, static isn't necessary.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+    "6B2ED877C2C52334AC51E59ABFA7EC457F4A7D01E46291E9F2EAA45F011D24B7"_array32, // sd_card_kek_source
+    "D482743563D3EA5DCDC3B74E97C9AC8A342164FA041A1DC80F17F6D31E4BC01C"_array32, // sd_card_save_key_source
+    "2E751CECF7D93A2B957BD5FFCB082FD038CC2853219DD3092C6DAB9838F5A7CC"_array32, // sd_card_nca_key_source
+    "1888CAED5551B3EDE01499E87CE0D86827F80820EFB275921055AA4E2ABDFFC2"_array32, // header_kek_source
+    "8F783E46852DF6BE0BA4E19273C4ADBAEE16380043E1B8C418C4089A8BD64AA6"_array32, // header_key_source
+    "D1757E52F1AE55FA882EC690BC6F954AC46A83DC22F277F8806BD55577C6EED7"_array32, // rsa_kek_seed3
+    "FC02B9D37B42D7A1452E71444F1F700311D1132E301A83B16062E72A78175085"_array32, // rsa_kek_mask0
+};
+
+const static std::array<SHA256Hash, 0x20> keyblob_source_hashes{
```
lioncash: Ditto

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // keyblob_key_source_19
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // keyblob_key_source_1A
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // keyblob_key_source_1B
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // keyblob_key_source_1C
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // keyblob_key_source_1D
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // keyblob_key_source_1E
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // keyblob_key_source_1F
+};
+
+const static std::array<SHA256Hash, 0x20> master_key_hashes{
```
lioncash: Ditto

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // master_key_19
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // master_key_1A
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // master_key_1B
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // master_key_1C
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // master_key_1D
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // master_key_1E
+    "0000000000000000000000000000000000000000000000000000000000000000"_array32, // master_key_1F
+};
+
+std::vector<u8> DecompressBLZ(const std::vector<u8>& in) {
```
lioncash: This function should be static

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+            control <<= 1;
+            if (out_index == 0)
+                return out;
+        }
+    }
+
+    return out;
+}
+
+u8 CalculateMaxKeyblobSourceHash() {
```
lioncash: Ditto

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+    for (s8 i = 0x1F; i >= 0; --i) {
+        if (keyblob_source_hashes[i] != SHA256Hash{})
+            return i + 1;
```
lioncash: This should cast to u8, otherwise this will cause both a sign conversion warning and a truncation warning, because the result of the expression here would be an `int`, not a `s8`.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+        std::array<u8, key_size> out{};
+        std::memcpy(out.data(), binary.data() + i, key_size);
+        return out;
+    }
+
+    return {};
+}
+
+std::array<Key128, 0x20> FindEncryptedMasterKeyFromHex(const std::vector<u8>& binary, Key128 key) {
```
lioncash: This can be a static function.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+    if (binary.size() < 0x10)
+        return {};
+
+    std::array<Key128, 0x20> out{};
+    AESCipher<Key128> cipher(key, Mode::ECB);
+    for (size_t i = 0; i < binary.size() - 0x10; ++i) {
+        cipher.Transcode(binary.data() + i, dec_temp.size(), dec_temp.data(), Op::Decrypt);
+        mbedtls_sha256(dec_temp.data(), dec_temp.size(), temp.data(), 0);
+
+        for (size_t k = 0; k < 0x20; ++k) {
```
lioncash: `k < out.size();`

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+        }
+    }
+
+    return out;
+}
+
+FileSys::VirtualFile FindFileInDirWithNames(const FileSys::VirtualDir& dir,
+                                            const std::string& name) {
+    auto upper = name;
+    std::transform(upper.begin(), upper.end(), upper.begin(), ::toupper);
```
lioncash:
```cpp
std::transform(upper.begin(), upper.end(), upper.begin(), [](u8 c){ return std::toupper(c); });
```
To prevent undefined behavior from a potential negative character value.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+// NOTE TO FUTURE MAINTAINERS:
+// When a new version of switch cryptography is released,
+// hash the new keyblob source and master key and add the hashes to
+// the arrays below.
+
+#include <array>
```
lioncash: `<algorithm>`, `<cstring>`, `"common/common_funcs.h"` needs to be included.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+    INIHeader ini{};
+    std::memcpy(&ini, c.data(), sizeof(INIHeader));
+    if (ini.magic != Common::MakeMagic('I', 'N', 'I', '1'))
+        return;
+
+    std::map<u64, KIPHeader> kips{};
+    u64 offset = sizeof(INIHeader);
+    for (size_t i = 0; i < ini.process_count; ++i) {
+        KIPHeader kip{};
```
lioncash: This doesn't need to be zeroed out if you're going to memcpy the full length of the data type.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+    auto c = a->ReadAllBytes();
+
+    AESCipher<Key128> cipher(package2_keys[revision], Mode::CTR);
+    cipher.SetIV(s1_iv);
+    cipher.Transcode(c.data(), c.size(), c.data(), Op::Decrypt);
+
+    // package2_decrypted[static_cast<size_t>(type)] = s1;
+
+    INIHeader ini{};
```
lioncash: This doesn't need to be zeroed out if you're going to memcpy the full length of the data type.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,586 @@
...
+        package2[static_cast<size_t>(type)],
+        package2[static_cast<size_t>(type)]->GetSize() - 0x4000, 0x4000);
+
+    Package2Header header{};
+    if (file->ReadObject(&header) != sizeof(Package2Header))
+        return;
+
+    u8 revision = 0xFF;
+    if (header.magic != Common::MakeMagic('P', 'K', '2', '1')) {
+        for (size_t i = 0; i < 0x20; ++i) {
```
lioncash: This should go in its own function.

src/core/crypto/partition_data_manager.h
```diff
@@ -0,0 +1,104 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
```
lioncash: newline after the pragma

src/core/crypto/partition_data_manager.h
```diff
@@ -0,0 +1,104 @@
...
+    NormalSub,
+    SafeModeMain,
+    SafeModeSub,
+    RepairMain,
+    RepairSub,
+};
+
+class PartitionDataManager {
+public:
+    const static u8 MAX_KEYBLOB_SOURCE_HASH;
```
lioncash: `static const`

src/core/crypto/partition_data_manager.h
```diff
@@ -0,0 +1,104 @@
...
+
+    explicit PartitionDataManager(FileSys::VirtualDir sysdata_dir);
```
lioncash: Destructor should be defaulted into the cpp file.

src/yuzu/main.h
```diff
@@ -159,6 +159,7 @@ private slots:
     void HideFullscreen();
     void ToggleWindowMode();
     void OnCoreError(Core::System::ResultStatus, std::string);
+    void OnReinitializeKeys(bool callouts);
```
lioncash: This should probably be using an enum class rather than a boolean parameter.

src/yuzu/main.cpp
```diff
@@ -1325,6 +1331,85 @@ void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string det
...
+        FileUtil::Delete(FileUtil::GetUserPath(FileUtil::UserPath::KeysDir) +
+                         "title.keys_autogenerated");
+    }
+
+    Core::Crypto::KeyManager keys{};
+    if (keys.BaseDeriveNecessary()) {
+        Core::Crypto::PartitionDataManager pdm{vfs->OpenDirectory(
+            FileUtil::GetUserPath(FileUtil::UserPath::SysDataDir), FileSys::Mode::Read)};
+
+        const auto function = [this, &keys, &pdm]() {
```
lioncash: `()` can be omitted for lambdas without arguments.

src/yuzu/main.cpp
```diff
@@ -1325,6 +1331,85 @@ void GMainWindow::OnCoreError(Core::System::ResultStatus result, std::string det
...
+            keys.PopulateFromPartitionData(pdm);
+            Service::FileSystem::CreateFactories(vfs);
+            keys.DeriveETicket(pdm);
+        };
+
+        std::vector<std::string> errors;
```
lioncash: This should likely be using a QString and make the errors translatable. Rather than concatenating all of the error strings at the end, this can be made more translator friendly by adding the formatting characters into the strings themselves.

e.g.

```cpp
tr("- Missing fuses - Cannot derive SBK\n")
```

Always provide the formatting wholesale. Adding it programmatically can make for a really weird interface for languages that may not order sentences the same way as English, or may not even use the same characters for forming visual lists, since the translator will have no idea that sort of thing will be appended to the strings.

src/core/crypto/key_manager.h
DarkLordZach: Understood

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,604 @@
...
+    u8 priority;
+    u8 core;
+    INSERT_PADDING_BYTES(1);
+    u8 flags;
+    std::array<SectionHeader, 6> sections;
+    std::array<u32, 0x20> capabilities;
+};
+static_assert(sizeof(KIPHeader) == 0x100, "KIPHeader has incorrect size.");
+
+static const std::array<SHA256Hash, 0x10> source_hashes{
```
lioncash: These arrays can just be `constexpr`. const and constexpr internally link by default on file scope values, so there's no need to specify static.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,604 @@
...
+        }
+    }
+
+    return out;
+}
+
+static u8 CalculateMaxKeyblobSourceHash() {
+    for (s8 i = 0x1F; i >= 0; --i) {
+        if (keyblob_source_hashes[i] != SHA256Hash{})
+            return static_cast<u8>(i) + 1;
```
lioncash: The cast needs to encompass the whole operation

```cpp
value_with_type_less_than_or_equal_to_int + value_less_than_or_equal_to_int
```
will always promote to a result with a type of `int`.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    return 0;
+}
+
+const u8 PartitionDataManager::MAX_KEYBLOB_SOURCE_HASH = CalculateMaxKeyblobSourceHash();
+
+template <size_t key_size = 0x10>
+std::array<u8, key_size> FindKeyFromHex(const std::vector<u8>& binary, std::array<u8, 0x20> hash) {
+    std::array<u8, 0x20> temp{};
```
lioncash: This should probably be moved down after the early return right below, given it's not used in that event.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,604 @@
...
+            return static_cast<u8>(i) + 1;
+    }
+
+    return 0;
+}
+
+const u8 PartitionDataManager::MAX_KEYBLOB_SOURCE_HASH = CalculateMaxKeyblobSourceHash();
+
+template <size_t key_size = 0x10>
+std::array<u8, key_size> FindKeyFromHex(const std::vector<u8>& binary, std::array<u8, 0x20> hash) {
```
lioncash: hash can be a const reference here to avoid copying it.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+        std::array<u8, key_size> out{};
+        std::memcpy(out.data(), binary.data() + i, key_size);
+        return out;
+    }
+
+    return {};
+}
+
+std::array<u8, 16> FindKeyFromHex16(const std::vector<u8>& binary, std::array<u8, 32> hash) {
```
lioncash: Ditto

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,604 @@
...
+    return FindKeyFromHex<0x10>(binary, hash);
+}
+
+static std::array<Key128, 0x20> FindEncryptedMasterKeyFromHex(const std::vector<u8>& binary,
+                                                              Key128 key) {
+    SHA256Hash temp{};
```
lioncash: These temporary vars should be moved after the early exit conditional below.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+    return {};
+}
+
+std::array<u8, 16> FindKeyFromHex16(const std::vector<u8>& binary, std::array<u8, 32> hash) {
+    return FindKeyFromHex<0x10>(binary, hash);
+}
+
+static std::array<Key128, 0x20> FindEncryptedMasterKeyFromHex(const std::vector<u8>& binary,
+                                                              Key128 key) {
```
lioncash: key can be a const reference here to avoid copying it unnecessarily.

src/core/crypto/partition_data_manager.cpp
```diff
@@ -0,0 +1,604 @@
...
+      kfuses(FindFileInDirWithNames(sysdata_dir, "kfuse")),
+      package2({
+          FindFileInDirWithNames(sysdata_dir, "BCPKG2-1-Normal-Main"),
+          FindFileInDirWithNames(sysdata_dir, "BCPKG2-2-Normal-Sub"),
+          FindFileInDirWithNames(sysdata_dir, "BCPKG2-3-SafeMode-Main"),
+          FindFileInDirWithNames(sysdata_dir, "BCPKG2-4-SafeMode-Sub"),
+          FindFileInDirWithNames(sysdata_dir, "BCPKG2-5-Repair-Main"),
+          FindFileInDirWithNames(sysdata_dir, "BCPKG2-6-Repair-Sub"),
+      }),
+      secure_monitor(FindFileInDirWithNames(sysdata_dir, "secmom")),
```
lioncash: Was this intended to be `secmon` rather than `secmom`?

src/core/crypto/key_manager.cpp
```diff
@@ -4,23 +4,56 @@
...
 #include "core/loader/loader.h"
 #include "core/settings.h"
 
 namespace Core::Crypto {
 
+constexpr u64 CURRENT_CRYPTO_REVISION = 0x5;
+
+using namespace Common;
+
+static const std::array<SHA256Hash, 2> eticket_source_hashes{
```
lioncash: This can just be `constexpr`, static isn't necessary since const and constexpr internally link by default on const file-scope variables.

src/core/crypto/key_manager.cpp
```diff
@@ -4,23 +4,56 @@
...
+    "B71DB271DC338DF380AA2C4335EF8873B1AFD408E80B3582D8719FC81C5E511C"_array32, // eticket_rsa_kek_source
+    "E8965A187D30E57869F562D04383C996DE487BBA5761363D2D4D32391866A85C"_array32, // eticket_rsa_kekek_source
+};
+
+static const std::map<std::pair<S128KeyType, u64>, std::string> KEYS_VARIABLE_LENGTH{
```
lioncash: Ditto, static can be omitted here.

src/core/crypto/key_manager.cpp
```diff
@@ -290,59 +649,390 @@ void KeyManager::DeriveSDSeedLazy() {
...
+    if (exefs == nullptr)
+        return;
+
+    const auto main = exefs->GetFile("main");
+    if (main == nullptr)
+        return;
+
+    const auto bytes = main->ReadAllBytes();
+
+    using namespace Common;
```
lioncash: The top of the file already has this using statement, so this likely isn't necessary.

src/core/crypto/key_manager.cpp
```diff
@@ -37,6 +70,77 @@ Key128 GenerateKeyEncryptionKey(Key128 source, Key128 master, Key128 kek_seed, K
     return out;
 }
 
+Key128 DeriveKeyblobKey(Key128 sbk, Key128 tsec, Key128 source) {
```
lioncash: `sbk` and `tsec` can be const taken by const reference.

src/core/crypto/key_manager.cpp
```diff
@@ -290,59 +649,390 @@ void KeyManager::DeriveSDSeedLazy() {
...
+
+    auto res = GetTicketblob(save1);
+    const auto res2 = GetTicketblob(save2);
+    std::copy(res2.begin(), res2.end(), std::back_inserter(res));
+
+    for (const auto& raw : res) {
+        const auto pair = ParseTicket(raw, rsa_key);
+        if (pair == boost::none)
+            continue;
+        auto [rid, key] = pair.value();
```
lioncash: You probably want to make this `const auto&` to avoid unnecessary copies

src/core/crypto/key_manager.cpp
```diff
@@ -290,59 +649,390 @@ void KeyManager::DeriveSDSeedLazy() {
...
+        SetKey(S128KeyType::HeaderKek, header_kek);
+
+        AESCipher<Key128> header_cipher(header_kek, Mode::ECB);
+        Key256 out = GetKey(S256KeyType::HeaderSource);
+        header_cipher.Transcode(out.data(), out.size(), out.data(), Op::Decrypt);
+        SetKey(S256KeyType::Header, out);
+    }
+}
+
+void KeyManager::DeriveETicket(PartitionDataManager data) {
```
lioncash: You likely want to take `data` by reference or const reference here to avoid copying the whole data manager.

src/core/crypto/key_manager.cpp
```diff
@@ -290,59 +649,390 @@ void KeyManager::DeriveSDSeedLazy() {
...
+    SetKey(id, key, field1, field2);
+}
+
+void KeyManager::SetKeyWrapped(S256KeyType id, Key256 key, u64 field1, u64 field2) {
+    if (key == Key256{})
+        return;
+    SetKey(id, key, field1, field2);
+}
+
+void KeyManager::PopulateFromPartitionData(PartitionDataManager data) {
```
lioncash: Ditto, you likely want to take `data` by reference or const reference here.

src/core/crypto/key_manager.h
lioncash: still applies

DarkLordZach: Now used in KeyManager::PopulateFromPartitonData, as it was code that was supposed to be extracted to this function but wasn't. Fixed in latest

Hexagon12:
Please rebase

src/core/crypto/key_manager.cpp
```diff
@@ -4,23 +4,56 @@
...
+#include "core/file_sys/nca_metadata.h"
+#include "core/file_sys/partition_filesystem.h"
+#include "core/file_sys/registered_cache.h"
+#include "core/hle/service/filesystem/filesystem.h"
 #include "core/loader/loader.h"
 #include "core/settings.h"
 
 namespace Core::Crypto {
 
+constexpr u64 CURRENT_CRYPTO_REVISION = 0x5;
```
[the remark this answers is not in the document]
DarkLordZach: No, they start at 0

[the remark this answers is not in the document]
CarlKenner: Then line 491 is wrong, which is the only place this constant is used.

PR #1412, opened 2018-09-28
kernel/object: Remove unnecessary std::move from DynamicObjectCast()

lioncash:
`boost::static_pointer_cast` for `boost::intrusive_ptr` (what `SharedPtr` is), takes its parameter by const reference. Given that, it means that this `std::move` doesn't actually do anything other than obscure what the function's actual behavior is, so we can remove this. To clarify, this would only do something if the parameter was either taking its argument by value, by non-const ref, or by rvalue-reference.

### DIFF
--- a/src/core/hle/kernel/object.h
+++ b/src/core/hle/kernel/object.h
@@ -6,7 +6,6 @@
 
 #include <atomic>
 #include <string>
-#include <utility>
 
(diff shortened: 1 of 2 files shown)


lioncash:
Failure isn't related to this PR

PR #1415, opened 2018-09-30
file_sys: Add support for loading IPS patches

DarkLordZach:
How to Use:
```
yuzu/load
  - <title id>
    - <mod name>
      - exefs
        - <nso build id>.ips
```
You can have multiple ips files in the exefs dir (for different versions for ex). You can also pair the exefs dir with a sibling romfs dir to make one mod with LFS and IPS.

This also changes the Add-Ons column slightly. Now instead of saying just `LayeredFS` (or `IPS` as you might have expected) It will say the name of the mod (<mod name> in ex above) and the in parenthesis what it uses (LFS and/or IPS)

Example: 
inlisti s

### DIFF
--- a/src/core/loader/nso.h
+++ b/src/core/loader/nso.h
@@ -28,7 +28,7 @@ public:
     }
 
     static VAddr LoadModule(FileSys::VirtualFile file, VAddr load_base,
-                            std::shared_ptr<FileSys::PatchManager> pm = nullptr);
+                            boost::optional<FileSys::PatchManager> pm = boost::none);
 
     ResultStatus Load(Kernel::Process& process) override;
 };
--- a/src/core/loader/nso.h
+++ b/src/core/loader/nso.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "common/common_types.h"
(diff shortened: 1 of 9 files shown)


src/core/loader/nso.h
```diff
@@ -26,7 +27,8 @@ class AppLoader_NSO final : public AppLoader, Linker {
         return IdentifyType(file);
     }
 
-    static VAddr LoadModule(FileSys::VirtualFile file, VAddr load_base);
+    static VAddr LoadModule(FileSys::VirtualFile file, VAddr load_base,
+                            std::shared_ptr<FileSys::PatchManager> pm = nullptr);
```
lioncash: This is likely more suitable as a `boost::optional` rather than a reference counted pointer. It's a bit overkill resource-wise as is.

jroweboy:
https://github.com/yuzu-emu/yuzu/pull/1415/files#diff-d80adbc892daeeefd2c0bbdc50a5bf5aR79 This line crashes when `patch_dirs.size() == 0` which should be an easy fix. Removing from canary for the time being

PR #1419, opened 2018-09-30
nso/nro: Add support for passing command-line arguments to executable

DarkLordZach:
- Adds necessary framework and structs to pass an arguments string into an NSO/NRO.
- Will pass into normal games, but only to the `rtld` NSO.
- Can be set in UI
argsstr
- Can be passed in via command line (for yuzu-cmd) using the `-p` or `--program` options.
(e.g. `yuzu-cmd -p "-h -l" program.nro` will execute the nro and pass `-h -l` to it)

- Created the following tester homebrew: https://github.com/DarkLordZach/echo-homebrew

- WIP as no libnx homebrew currently runs in yuzu because of HID problems

### DIFF
--- a/src/core/loader/nro.cpp
+++ b/src/core/loader/nro.cpp
@@ -154,12 +154,14 @@ bool AppLoader_NRO::LoadNro(FileSys::VirtualFile file, VAddr load_base) {
 
     if (!Settings::values.program_args.empty()) {
         const auto arg_data = Settings::values.program_args;
-        codeset->DataSegment().size += 0x9000;
-        NSOArgumentHeader args_header{0x9000, static_cast<u32_le>(arg_data.size()), {}};
-        program_image.resize(static_cast<u32>(program_image.size()) + 0x9000);
-        std::memcpy(program_image.data() + program_image.size() - 0x9000, &args_header,
-                    sizeof(NSOArgumentHeader));
-        std::memcpy(program_image.data() + program_image.size() - 0x8FE0, arg_data.data(),
+        codeset->DataSegment().size += NSO_ARGUMENT_DATA_ALLOCATION_SIZE;
+        NSOArgumentHeader args_header{
+            NSO_ARGUMENT_DATA_ALLOCATION_SIZE, static_cast<u32_le>(arg_data.size()), {}};
+        const auto end_offset = program_image.size();
+        program_image.resize(static_cast<u32>(program_image.size()) +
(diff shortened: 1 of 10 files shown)


src/core/loader/nro.cpp
```diff
@@ -150,6 +152,17 @@ bool AppLoader_NRO::LoadNro(FileSys::VirtualFile file, VAddr load_base) {
         codeset->segments[i].size = PageAlignSize(nro_header.segments[i].size);
     }
 
+    if (!Settings::values.program_args.empty()) {
+        const auto arg_data = Settings::values.program_args;
+        codeset->DataSegment().size += 0x9000;
```
Subv: Do not use magic numbers.

src/core/loader/nro.cpp
```diff
@@ -150,6 +152,17 @@ bool AppLoader_NRO::LoadNro(FileSys::VirtualFile file, VAddr load_base) {
...
+        NSOArgumentHeader args_header{0x9000, arg_data.size(), {}};
+        program_image.resize(static_cast<u32>(program_image.size()) + 0x9000);
+        std::memcpy(program_image.data() + program_image.size() - 0x9000, &args_header,
```
Subv: Store the end offset before resizing the vector

src/core/loader/nro.cpp
```diff
@@ -150,6 +152,17 @@ bool AppLoader_NRO::LoadNro(FileSys::VirtualFile file, VAddr load_base) {
...
+                    sizeof(NSOArgumentHeader));
+        std::memcpy(program_image.data() + program_image.size() - 0x8FE0, arg_data.data(),
```
Subv: Use offsetof or sizeof instead of hardcoding yet another magic number

src/core/loader/nso.cpp
```diff
@@ -124,6 +126,17 @@ VAddr AppLoader_NSO::LoadModule(FileSys::VirtualFile file, VAddr load_base) {
         codeset->segments[i].size = PageAlignSize(static_cast<u32>(data.size()));
     }
 
+    if (should_pass_arguments && !Settings::values.program_args.empty()) {
+        const auto arg_data = Settings::values.program_args;
+        codeset->DataSegment().size += 0x9000;
```
Subv: Same

bunnei:
@DarkLordZach is there a reason this is still [WIP] ?

PR #1423, opened 2018-10-01
 fsmitm_romfsbuild: Add support for stubbing and IPS patches in LFS

DarkLordZach:
This PR extends the current LayeredFS functionality in yuzu with two new features:

*NOTE: Both of these are placed in a directory called `romfs_ext` which is a sibling to `romfs` in a mod. This is to prevent name conflicts with developer files. This dir follows the exact directory structure as the actual romfs.*
1. Add support for stubbing files. When patching, if there exists a file with the entire name of the current file but the extension `.stub`, it will be deleted from the final image.
    - For example, if you wish to delete `texture.dds` from a romfs, make a file in the same directory under LayeredFS called `texture.dds.stub`.
2. Add support for IPS patching files. This is useful when distributing the entire replacement file would be copyright infringement. Similar to stubbing above, but with the extension `.ips`.
    - For example, to patch `menu.dat` in romfs, create a file in the same directory under LayeredFS called `menu.dat.ips`.

Fair note of caution, `stub` is propagating. If one mod has a .stub for a file, it cannot be overridden by another mod, regardless of load order.

**DEPENDENT ON PR #1415 -- WANTED TO GET REVIEW AND SUCH OUT OF THE WAY**
*Includes a rebase so it compiles/can be tested -- actual PR starts at commit b5b03fb*

### DIFF
--- a/src/core/file_sys/fsmitm_romfsbuild.cpp
+++ b/src/core/file_sys/fsmitm_romfsbuild.cpp
@@ -124,7 +124,7 @@ static u64 romfs_get_hash_table_count(u64 num_entries) {
     return count;
 }
 
-void RomFSBuildContext::VisitDirectory(VirtualDir root_romfs,
+void RomFSBuildContext::VisitDirectory(VirtualDir root_romfs, VirtualDir ext,
                                        std::shared_ptr<RomFSBuildDirectoryContext> parent) {
     std::vector<std::shared_ptr<RomFSBuildDirectoryContext>> child_dirs;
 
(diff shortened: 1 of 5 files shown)


SciresM:
Not a fan of the stubbing syntax -- what if a developer includes a file with a ".stub" extension?

I had vague plans to implement in Atmosphere a solution where a `romfs_stubs` folder was traversed, and if a file was present in there it would be patched.

I had similar plans for ips/ips32 support (`romfs_patches`).

DarkLordZach:
This current impl will leave the stubs in (ex. if a dev had a file called `x.stub`, as long as they don't have a file in the same dir called `x` it will work.) I do agree that it's ambiguous in its current form. 

Perhaps a dir called `romfs_ext` or something where `.ips` and `.stub` files can be placed (matching dir structure and everything)?

SciresM:
@DarkLordZach Yeah, `romfs_ext` is reasonable.

DarkLordZach:
Will do, thanks for the input @SciresM

PR #1424, opened 2018-10-01
ips_layer: Add support for IPSwitch executable patches

DarkLordZach:
IPSwitch (pronounced IPS-witch) is a patch description format created by @3096 to be easier to use than IPS. The format revolves around 'patch text' files which look similar to the following:
```
// Crash game if it tries to delete save
@disabled
00176ACC DEADBEEF
``` 
(credit to Dualie Artist on Splatoon Modding Hub discord)

This PR adds support for parsing these patch texts instead of requiring the user to use homebrew to compile them to IPS and then using that.

The files (extension `.pchtxt`) go into the exefs folder of a mod and can be named anything desired (as the file contains the NSO build ID within it).

A full specification as well as examples and the homebrew can be found [here](https://github.com/3096/ipswitch).

**DEPENDENT ON PR #1415 -- WANTED TO GET REVIEW AND SUCH OUT OF THE WAY**
*Includes a rebase so it compiles/can be tested -- actual PR starts at commit 3025afe*

### DIFF
--- a/src/core/file_sys/ips_layer.cpp
+++ b/src/core/file_sys/ips_layer.cpp
@@ -17,9 +17,10 @@ enum class IPSFileType {
     Error,
 };
 
-const std::map<const char*, const char*> ESCAPE_CHARACTER_MAP{
-    {"\\a", "\a"}, {"\\b", "\b"},  {"\\f", "\f"},  {"\\n", "\n"},  {"\\r", "\r"},  {"\\t", "\t"},
-    {"\\v", "\v"}, {"\\\\", "\\"}, {"\\\'", "\'"}, {"\\\"", "\""}, {"\\\?", "\?"},
+constexpr std::array<std::pair<const char*, const char*>, 11> ESCAPE_CHARACTER_MAP{
+    std::pair{"\\a", "\a"}, {"\\b", "\b"},  {"\\f", "\f"},  {"\\n", "\n"},
+    {"\\r", "\r"},          {"\\t", "\t"},  {"\\v", "\v"},  {"\\\\", "\\"},
+    {"\\\'", "\'"},         {"\\\"", "\""}, {"\\\?", "\?"},
 };
 
 static IPSFileType IdentifyMagic(const std::vector<u8>& magic) {
@@ -92,11 +93,11 @@ VirtualFile PatchIPS(const VirtualFile& in, const VirtualFile& ips) {
     return std::make_shared<VectorVfsFile>(in_data, in->GetName(), in->GetContainingDirectory());
 }
 
-IPSwitchCompiler::IPSwitchCompiler(VirtualFile patch_text_)
-    : valid(false), patch_text(std::move(patch_text_)), nso_build_id{}, is_little_endian(false),
-      offset_shift(0), print_values(false), last_comment("") {
+IPSwitchCompiler::IPSwitchCompiler(VirtualFile patch_text_) : patch_text(std::move(patch_text_)) {
     Parse();
 }
+
+IPSwitchCompiler::~IPSwitchCompiler() = default;
 
 std::array<u8, 32> IPSwitchCompiler::GetBuildID() const {
     return nso_build_id;
@@ -106,7 +107,7 @@ bool IPSwitchCompiler::IsValid() const {
     return valid;
 }
 
-static bool StartsWith(const std::string& base, const std::string& check) {
+static bool StartsWith(std::string_view base, std::string_view check) {
     return base.size() >= check.size() && base.substr(0, check.size()) == check;
 }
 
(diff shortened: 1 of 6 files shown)


src/core/file_sys/ips_layer.cpp
```diff
@@ -85,4 +87,144 @@ VirtualFile PatchIPS(const VirtualFile& in, const VirtualFile& ips) {
...
+
+std::array<u8, 32> IPSwitchCompiler::GetBuildID() const {
+    return nso_build_id;
+}
+
+bool IPSwitchCompiler::IsValid() const {
+    return valid;
+}
+
+static bool StartsWith(const std::string& base, const std::string& check) {
```
lioncash: These should be string_view instances.

src/core/file_sys/ips_layer.h
```diff
@@ -12,4 +12,30 @@ namespace FileSys {
 
 VirtualFile PatchIPS(const VirtualFile& in, const VirtualFile& ips);
 
+class IPSwitchCompiler {
+public:
+    explicit IPSwitchCompiler(VirtualFile patch_text);
```
lioncash: destructor should be defaulted in the cpp file.

src/core/file_sys/ips_layer.cpp
```diff
@@ -85,4 +87,144 @@ VirtualFile PatchIPS(const VirtualFile& in, const VirtualFile& ips) {
...
+                }
+
+                if (print_values) {
+                    LOG_INFO(Loader,
+                             "[IPSwitchCompiler ('{}')]     - Patching value at offset 0x{:08X} "
+                             "with byte string '{}'",
+                             patch_text->GetName(), offset, Common::HexVectorToString(replace));
+                }
+
+                patch.records.emplace(offset, replace);
```
lioncash: replace should be std::moved

src/core/file_sys/ips_layer.cpp
```diff
@@ -85,4 +87,144 @@ VirtualFile PatchIPS(const VirtualFile& in, const VirtualFile& ips) {
...
+            LOG_INFO(Loader, "[IPSwitchCompiler ('{}')] Parsing patch '{}' ({})",
+                     patch_text->GetName(), name, line.substr(1));
+
+            IPSwitchPatch patch{name, enabled, {}};
+
+            // Read rest of patch
+            while (true) {
+                if (i + 1 >= lines.size())
+                    break;
+                line = lines[++i];
```
lioncash: This should likely not be reusing the same variable as the outer scope.

src/core/file_sys/patch_manager.cpp
```diff
@@ -250,8 +270,22 @@ std::map<std::string, std::string, std::less<>> PatchManager::GetPatchVersionNam
     if (mod_dir != nullptr && mod_dir->GetSize() > 0) {
         for (const auto& mod : mod_dir->GetSubdirectories()) {
             std::string types;
-            if (IsDirValidAndNonEmpty(mod->GetSubdirectory("exefs")))
-                AppendCommaIfNotEmpty(types, "IPS");
+            if (IsDirValidAndNonEmpty(mod->GetSubdirectory("exefs"))) {
+                bool ips = false;
+                bool ipswitch = false;
+
+                for (const auto& file : mod->GetSubdirectory("exefs")->GetFiles()) {
```
lioncash:
```cpp
mod->GetSubdirectory("exefs")
```
is already called in the if statement condition. You can just extract that call into its own variable and use it here to deduplicate unnecessary work.

src/core/file_sys/ips_layer.cpp
```diff
@@ -17,6 +17,11 @@ enum class IPSFileType {
     Error,
 };
 
+const std::map<const char*, const char*> ESCAPE_CHARACTER_MAP{
```
lioncash: This probably shouldn't be a std::map, given causes global heap allocation before the program can even run.

src/core/file_sys/ips_layer.cpp
```diff
@@ -121,6 +138,13 @@ void IPSwitchCompiler::Parse() {
 
     for (std::size_t i = 0; i < lines.size(); ++i) {
         auto line = lines[i];
+
+        // Remove midline comments
+        if (!StartsWith(line, "//") && line.find("//") != std::string::npos) {
+            last_comment = line.substr(line.find("//") + 2);
```
lioncash: You can just extract the line.find("//") call from the if statement condition instead of duplicating the same work twice more than you need to.

src/core/file_sys/ips_layer.cpp
DarkLordZach: would `constexpr std::array<std::pair<const char*, const char*>, X` work instead?

PR #1425, opened 2018-10-02
gl_shader_decompiler: Implement geometry shaders

ReinUsesLisp:
Implements geometry shaders. The following changes had to be done:
* Explicitly set `position` location ~to 15~ to 0. Generic varyings start with an offset of 1.
* Add an extra config argument for shaders. It contains an integer telling the vertex-based shader if it has to viewport-flip or it's handled in a future stage.
* Querying a shader program from the cache requires a primitive mode argument. This is needed because in Maxwell's header there's no data about the input topology but GLSL requires one. It lazily builds shaders as needed. Initially I thought of building them all at once, but it generated some out-of-bound errors from NVidia's blob GLSL compiler.
* When a geometry shader is disabled, it gets unbound.
* Changed shader opcode commentaries to be shown in hex.

About the implementation itself, this does the following:
* Stub `ISBERD` as a copy. Hardware might be translating from a pointer, but we don't care about that.
* An `AMEM` buffer is created. I'm not really sure about the need of this one.
* `OUT.EMIT` emits a vertex. It returns the next index in `AMEM` looping in `[0; 6)`.
* `MOV_SYS` (`S2R`) with Invocation Info is stubbed as a zero set. Unimplemented ones assert.

Special geometry shader built-ins are not implemented (`gl_Layer`, `gl_ViewportLayer`).

It's important to note that I didn't see any progression in commercial games, if anyone has a commercial that currently boots and uses geometry shaders (Xenoblade, Yooka Laylee), I'd appreciate some feedback about regressions/progressions. Although as far as I know those games require lots of unimplemented stuff.

To implement them I wrote [this](https://github.com/yuzu-emu/yuzu/files/2436403/nouveauTests.7z.zip) tests.

imagen

Special thanks to nouveau and libnx contributors.

### DIFF
--- a/src/video_core/utils.h
+++ b/src/video_core/utils.h
@@ -169,16 +169,20 @@ static void LabelGLObject(GLenum identifier, GLuint handle, VAddr addr,
     const std::string nice_addr = fmt::format("0x{:016x}", addr);
     std::string object_label;
 
-    switch (identifier) {
-    case GL_TEXTURE:
-        object_label = extra_info + "@" + nice_addr;
-        break;
-    case GL_PROGRAM:
-        object_label = "ShaderProgram@" + nice_addr;
-        break;
-    default:
-        object_label = fmt::format("Object(0x{:x})@{}", identifier, nice_addr);
-        break;
+    if (extra_info.empty()) {
+        switch (identifier) {
+        case GL_TEXTURE:
+            object_label = "Texture@" + nice_addr;
+            break;
+        case GL_PROGRAM:
+            object_label = "Shader@" + nice_addr;
+            break;
+        default:
+            object_label = fmt::format("Object(0x{:x})@{}", identifier, nice_addr);
+            break;
+        }
+    } else {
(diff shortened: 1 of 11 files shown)


bunnei:
@Subv ping for a second review

src/video_core/utils.h
```diff
@@ -169,16 +169,20 @@ static void LabelGLObject(GLenum identifier, GLuint handle, VAddr addr,
...
-    case GL_PROGRAM:
-        object_label = "ShaderProgram@" + nice_addr;
-        break;
-    default:
-        object_label = fmt::format("Object(0x{:x})@{}", identifier, nice_addr);
-        break;
+    if (extra_info.empty()) {
+        switch (identifier) {
+        case GL_TEXTURE:
+            object_label = "Texture@" + nice_addr;
```
FernandoS27: You are removing the name indicative of the texture here.

ReinUsesLisp: Mind line 172.

PR #1432, opened 2018-10-03
service/lbl: Update service function table

lioncash:
Amends the lbl service table to include new names of functions that were added to Switchbrew.


lioncash:
error doesn't seem to be related to the code in this PR.

PR #1441, opened 2018-10-05
logging: Add DebuggerBackend for logging to Visual Studio

CarlKenner:
This shows the log in the Output window in Visual Studio, or whatever other debugger you use on Windows, like it does in Dolphin.

### DIFF
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -12,7 +12,8 @@
 #include <thread>
 #include <vector>
 #ifdef _WIN32
-#include <share.h> // For _SH_DENYWR
+#include <share.h>   // For _SH_DENYWR
+#include <windows.h> // For OutputDebugStringA
 #else
 #define _SH_DENYWR 0
 #endif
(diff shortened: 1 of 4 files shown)


src/common/logging/backend.cpp
```diff
@@ -145,6 +146,12 @@ void FileBackend::Write(const Entry& entry) {
     }
 }
 
+void DebuggerBackend::Write(const Entry& entry) {
+#ifdef _WIN32
+    ::OutputDebugStringA((FormatLogMessage(entry) + '\n').c_str());
```
lioncash: This should likely be:

```cpp
::OutputDebugStringA(FormatLogMessage(entry).append(1, '\n').c_str());
```

so it's not making a copy of the whole string just to append a newline character, reducing the potential amount of allocations needed.

bunnei:
Sorry - forgot about this change. Tagging as canary to make sure there is nothing unexpected, then will merge if things look good. Thanks for contributing!

PR #1442, opened 2018-10-05
text_formatter: Avoid unnecessary string temporary creation in PrintMessage()

lioncash:
operator+ for std::string creates an entirely new string, which is kind of unnecessary here if we just want to append a null terminator to the existing one.

Reduces the total amount of potential allocations that need to be done in the logging path.


CarlKenner:
I already fixed this in my pull request.

lioncash:
When I originally made the PR, it wasn't part of yours.

PR #1339, opened 2018-09-17
GPU/DMA: Flush the source region and invalidate the destination region when doing a DMA transfer.

Subv:
Also implemented the source offsets for Tiled->Linear transfers, this is needed by nouveau for proper glReadPixels behavior.

This should probably fix some random cache issues in some games. The same flushing for the 2D engine and the Copy engine will be implemented in a future PR.

This causes crashes in Splatoon, BoTW and ARMS due to our GetSwizzleOffset function being slightly broken:

```
GetSwizzleOffset(848, 1024, 1920, 4, 16) = 0x7EA000 for a 1920x1080 texture with bpp=4
That's out of bounds
```

PR #1446, opened 2018-10-06
gl_rasterizer: Implement accelerated Fermi2D copies.

bunnei:
This is an "extrapolation" of #1339. With this change, we implement a more functional Fermi2D copy. Instead of @Subv's change which flushes and loads from guest memory, we just use existing cached surfaces. This also adds a faster and simpler surface copy using `glCopyImageSubData`. Fixes several things, such as the ice effect in Super Mario Odyssey, and the blur effect in One Piece - I am sure more.

image
image

### DIFF
--- a/src/video_core/engines/fermi_2d.cpp
+++ b/src/video_core/engines/fermi_2d.cpp
@@ -4,11 +4,13 @@
 
 #include "core/memory.h"
 #include "video_core/engines/fermi_2d.h"
+#include "video_core/rasterizer_interface.h"
 #include "video_core/textures/decoders.h"
 
 namespace Tegra::Engines {
 
-Fermi2D::Fermi2D(MemoryManager& memory_manager) : memory_manager(memory_manager) {}
+Fermi2D::Fermi2D(VideoCore::RasterizerInterface& rasterizer, MemoryManager& memory_manager)
+    : memory_manager(memory_manager), rasterizer{rasterizer} {}
 
 void Fermi2D::WriteReg(u32 method, u32 value) {
     ASSERT_MSG(method < Regs::NUM_REGS,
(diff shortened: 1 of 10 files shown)


src/video_core/engines/fermi_2d.cpp
```diff
@@ -44,28 +46,9 @@ void Fermi2D::HandleSurfaceCopy() {
...
-                                  regs.src.BlockHeight());
-    } else {
-        // If the input is linear and the output is tiled, swizzle the input and copy it over.
-        Texture::CopySwizzledData(regs.src.width, regs.src.height, src_bytes_per_pixel,
-                                  dst_bytes_per_pixel, dst_buffer, src_buffer, false,
-                                  regs.dst.BlockHeight());
-    }
+    // TODO(bunnei): Accelerated copy is likely incomplete, but without flushing there is not much
+    // point in doing anything else here.
+    rasterizer.AccelerateSurfaceCopy(regs.src, regs.dst);
```
Subv: We should return a boolean in AccelerateSurfaceCopy to handle fallback cases that OpenGL can not handle, and to keep a purely software implementation

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1033,6 +1067,25 @@ Surface RasterizerCacheOpenGL::GetUncachedSurface(const SurfaceParams& params) {
...
+    ASSERT(src_params.width == dst_params.width);
+    ASSERT(src_params.height == dst_params.height);
+    ASSERT(src_params.pixel_format == dst_params.pixel_format);
+    ASSERT(src_params.block_height == dst_params.block_height);
+    ASSERT(src_params.is_tiled == dst_params.is_tiled);
+    ASSERT(src_params.depth == dst_params.depth);
+    ASSERT(src_params.target == dst_params.target);
+    ASSERT(src_params.rt.index == dst_params.rt.index);
+
+    FastCopySurface(GetSurface(src_params, true), GetSurface(dst_params, false));
```
ReinUsesLisp: You should also assert a depth of 1 since that's how it's being handled in `FastCopySurface`.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1041,6 +1094,14 @@ Surface RasterizerCacheOpenGL::RecreateSurface(const Surface& old_surface,
     // Get a new surface with the new parameters, and blit the previous surface to it
     Surface new_surface{GetUncachedSurface(new_params)};
 
+    // For compatible surfaces, we can just do fast glCopyImageSubData based copy
+    if (old_params.target == new_params.target && old_params.type == new_params.type &&
+        SurfaceParams::GetFormatBpp(old_params.pixel_format) ==
+            SurfaceParams::GetFormatBpp(new_params.pixel_format)) {
+        FastCopySurface(old_surface, new_surface);
```
ReinUsesLisp: Same here, but instead of asserting you can skip the fast path.

bunnei: Addressed

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
bunnei: Addressed

PR #1448, opened 2018-10-06
Ported #4296 from citra

ogniK5377:
This will allow us to easily remove the use of "NFC" in "System"

https://github.com/citra-emu/citra/pull/4296 for more information

### DIFF
--- a/src/core/hle/kernel/client_port.h
+++ b/src/core/hle/kernel/client_port.h
@@ -7,13 +7,13 @@
 #include <string>
 #include "common/common_types.h"
 #include "core/hle/kernel/object.h"
+#include "core/hle/kernel/server_port.h"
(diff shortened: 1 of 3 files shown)


src/core/hle/kernel/client_port.h
```diff
@@ -7,13 +7,13 @@
...
 #include "common/common_types.h"
 #include "core/hle/kernel/object.h"
+#include "core/hle/kernel/server_port.h"
 #include "core/hle/result.h"
 
 namespace Kernel {
 
 class ClientSession;
 class KernelCore;
-class ServerPort;
```
lioncash: This can still be a forward declaration if `GetServerPort()` is defined in the cpp file.

PR #1455, opened 2018-10-07
Fixed smo softlock due to incorrect effect state updating

ogniK5377:
The explanation, after effects are initialized in smo, nothing actually changes the state. It expects the state to always be initialized. With the previous testing, updating the states much like how we handle the memory pools continue to have the softlock(which is why I said it probably wasn't effects) after further examination it seems like effects need to be initialized but the state remains unchanged until further notice. For now, assertions are added for the aux buffers to see if they update, unable to check as I haven't gotten smo to actually update them yet.

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -30,7 +30,7 @@ public:
         return info;
     }
 
-    VoiceInfo& Info() {
+    VoiceInfo& GetInfo() {
         return info;
     }
 
@@ -61,7 +61,7 @@ public:
         return info;
     }
 
-    EffectInStatus& Info() {
+    EffectInStatus& GetInfo() {
         return info;
     }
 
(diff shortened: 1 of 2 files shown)


lioncash:
Explanation should go in the commit message.

src/audio_core/audio_renderer.cpp
```diff
@@ -51,9 +51,30 @@ class AudioRenderer::VoiceState {
...
 };
 
+class AudioRenderer::EffectState {
+public:
+    const EffectOutStatus& GetOutStatus() const {
+        return out_status;
+    }
+
+    const EffectInStatus& GetInfo() const {
+        info;
```
lioncash: This should have a return in it.

src/audio_core/audio_renderer.cpp
```diff
@@ -51,9 +51,30 @@ class AudioRenderer::VoiceState {
...
+    }
+
+    EffectInStatus& Info() {
```
lioncash: `GetInfo()`

src/audio_core/audio_renderer.h
```diff
@@ -174,10 +221,12 @@ class AudioRenderer {
 
 private:
     class VoiceState;
+    class EffectState;
```
lioncash: This should go above VoiceState, to keep it alphabetized.

PR #1458, opened 2018-10-08
Fixed block height settings for RenderTargets and Depth Buffers

FernandoS27:
also added block width and block depth.


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -123,10 +127,22 @@ static VAddr TryGetCpuAddr(Tegra::GPUVAddr gpu_addr) {
 /*static*/ SurfaceParams SurfaceParams::CreateForDepthBuffer(u32 zeta_width, u32 zeta_height,
                                                              Tegra::GPUVAddr zeta_address,
                                                              Tegra::DepthFormat format) {
+    auto matchBlockHeight = [](u32 height) {
```
ReinUsesLisp: PascalCase

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -123,10 +127,22 @@ static VAddr TryGetCpuAddr(Tegra::GPUVAddr gpu_addr) {
 /*static*/ SurfaceParams SurfaceParams::CreateForDepthBuffer(u32 zeta_width, u32 zeta_height,
                                                              Tegra::GPUVAddr zeta_address,
                                                              Tegra::DepthFormat format) {
+    auto MatchBlockHeight = [](u32 height) {
+        u32 bh = 0;
+        u32 h = height / 8;
+        while (h) {
+            h = h << 1;
```
lioncash: `h <<= 1;`

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -123,10 +127,22 @@ static VAddr TryGetCpuAddr(Tegra::GPUVAddr gpu_addr) {
...
+            bh++;
+        }
+        bh = std::min(bh, 4u);
```
lioncash: We generally use `U` instead of `u` to stay consistent with the use of `L` instead of `l`.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -123,10 +127,22 @@ static VAddr TryGetCpuAddr(Tegra::GPUVAddr gpu_addr) {
...
+        return 1 << bh;
```
lioncash: This should likely be `1U` instead of `1`

src/video_core/textures/texture.h
```diff
@@ -202,13 +204,27 @@ struct TICEntry {
         return depth_minus_1 + 1;
     }
 
+    u32 BlockWidth() const {
+        ASSERT(header_version == TICHeaderVersion::BlockLinear ||
```
lioncash:
```cpp
ASSERT(IsTiled());
```

This also applies below

FernandoS27:
changes made. I removed the default block height calculation on depth buffers since I found the register that sets its block dimensions; I've also updated Fermi2D Surfaces to support correct blocking.

PR #1459, opened 2018-10-09
svcBreak, Signalling to the debugger should not kill execution

ogniK5377:
When loading NROs, svcBreak is called to signal to the debugger that a new "module" is loaded. As no debugger is technically attached we shouldn't be killing the execution of the program.

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -308,12 +308,12 @@ static void Break(u64 reason, u64 info1, u64 info2) {
             Debug_Emulated,
             "Emulated program broke execution! reason=0x{:016X}, info1=0x{:016X}, info2=0x{:016X}",
             reason, info1, info2);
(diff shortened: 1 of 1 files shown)


src/core/hle/kernel/svc.cpp
```diff
@@ -303,11 +303,18 @@ static ResultCode ArbitrateUnlock(VAddr mutex_addr) {
 
 /// Break program execution
 static void Break(u64 reason, u64 info1, u64 info2) {
-    LOG_CRITICAL(
-        Debug_Emulated,
-        "Emulated program broke execution! reason=0x{:016X}, info1=0x{:016X}, info2=0x{:016X}",
-        reason, info1, info2);
-    ASSERT(false);
+    if ((reason & (1 << 31)) == 0) {
```
Subv: Add a struct and use a bitfield for this.

PR #1467, opened 2018-10-10
Fixed incorrect types for svcBreak

ogniK5377:
svcBreak reason should be a u32, not a u64.

### DIFF
--- a/src/core/hle/kernel/svc_wrap.h
+++ b/src/core/hle/kernel/svc_wrap.h
@@ -35,18 +35,18 @@ void SvcWrap() {
 
 template <ResultCode func(u32)>
 void SvcWrap() {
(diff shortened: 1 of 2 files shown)


src/core/hle/kernel/svc_wrap.h
```diff
@@ -239,4 +239,9 @@ void SvcWrap() {
     func(Param(0), Param(1), Param(2));
 }
 
+template <void func(u32, u64, u64)>
+void SvcWrap() {
+    func((u32)(Param(0) & 0xFFFFFFFF), Param(1), Param(2));
```
lioncash:
```cpp
func(static_cast<u32>(Param(0)), Param(1), Param(2));
```

PR #1468, opened 2018-10-10
qt: Add UI to manage emulated user profiles

DarkLordZach:
tl;dr multiple users and user icons

This adds a UI to manage emulated users along with config support for multiple users and loading of custom icons for users.

For icons, yuzu looks in `%YUZU_DIR%/config/users/<username>.jpg`. It must be a JPEG less than 20kb in size due to switch restrictions. Should no suitable icon be found, a gray background will be used.

The UI allows adding, renaming, and deleting users. When a new user is made, the UUID will be randomly generated. 

Additionally, the right-click open save data menu will let you pick the user.

settings
(NSO NES - icon and name in top left corner)
ppt
(Puyo Puyo - custom name in blue box in bottom left corner)

### DIFF
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -3,8 +3,11 @@
 // Refer to the license.txt file included.
 
 #include <array>
+#include "common/common_paths.h"
 #include "common/common_types.h"
+#include "common/file_util.h"
 #include "common/logging/log.h"
+#include "common/string_util.h"
 #include "common/swap.h"
 #include "core/core_timing.h"
 #include "core/hle/ipc_helpers.h"
@@ -16,6 +19,9 @@
 #include "core/hle/service/acc/profile_manager.h"
 
 namespace Service::Account {
+
+constexpr u32 MAX_JPEG_IMAGE_SIZE = 0x20000;
+
 // TODO: RE this structure
 struct UserData {
     INSERT_PADDING_WORDS(1);
@@ -26,6 +32,11 @@ struct UserData {
     INSERT_PADDING_BYTES(0x60);
 };
 static_assert(sizeof(UserData) == 0x80, "UserData structure has incorrect size");
+
+static std::string GetImagePath(const std::string& username) {
+    return FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir) + "users" + DIR_SEP + username +
+           ".jpg";
+}
 
 class IProfile final : public ServiceFramework<IProfile> {
 public:
@@ -38,6 +49,15 @@ public:
             {11, &IProfile::LoadImage, "LoadImage"},
         };
         RegisterHandlers(functions);
+
+        ProfileBase profile_base{};
+        if (profile_manager.GetProfileBase(user_id, profile_base)) {
+            image = std::make_unique<FileUtil::IOFile>(
+                GetImagePath(Common::StringFromFixedZeroTerminatedBuffer(
+                    reinterpret_cast<const char*>(profile_base.username.data()),
+                    profile_base.username.size())),
+                "rb");
+        }
     }
 
 private:
@@ -73,11 +93,11 @@ private:
     }
 
     void LoadImage(Kernel::HLERequestContext& ctx) {
-        LOG_WARNING(Service_ACC, "(STUBBED) called");
+        LOG_DEBUG(Service_ACC, "called");
         // smallest jpeg https://github.com/mathiasbynens/small/blob/master/jpeg.jpg
-        // TODO(mailwl): load actual profile image from disk, width 256px, max size 0x20000
-        constexpr u32 jpeg_size = 107;
-        static constexpr std::array<u8, jpeg_size> jpeg{
+        // used as a backup should the one on disk not exist
+        constexpr u32 backup_jpeg_size = 107;
+        static constexpr std::array<u8, backup_jpeg_size> backup_jpeg{
             0xff, 0xd8, 0xff, 0xdb, 0x00, 0x43, 0x00, 0x03, 0x02, 0x02, 0x02, 0x02, 0x02, 0x03,
             0x02, 0x02, 0x02, 0x03, 0x03, 0x03, 0x03, 0x04, 0x06, 0x04, 0x04, 0x04, 0x04, 0x04,
             0x08, 0x06, 0x06, 0x05, 0x06, 0x09, 0x08, 0x0a, 0x0a, 0x09, 0x08, 0x09, 0x09, 0x0a,
@@ -87,22 +107,38 @@ private:
             0xff, 0xcc, 0x00, 0x06, 0x00, 0x10, 0x10, 0x05, 0xff, 0xda, 0x00, 0x08, 0x01, 0x01,
             0x00, 0x00, 0x3f, 0x00, 0xd2, 0xcf, 0x20, 0xff, 0xd9,
         };
-        ctx.WriteBuffer(jpeg);
+
         IPC::ResponseBuilder rb{ctx, 3};
         rb.Push(RESULT_SUCCESS);
-        rb.Push<u32>(jpeg_size);
+
+        if (image == nullptr) {
+            ctx.WriteBuffer(backup_jpeg);
+            rb.Push<u32>(backup_jpeg_size);
+        } else {
+            const auto size = std::min<u32>(image->GetSize(), MAX_JPEG_IMAGE_SIZE);
+            std::vector<u8> buffer(size);
+            image->ReadBytes(buffer.data(), buffer.size());
+
+            ctx.WriteBuffer(buffer.data(), buffer.size());
+            rb.Push<u32>(buffer.size());
+        }
     }
 
     void GetImageSize(Kernel::HLERequestContext& ctx) {
-        LOG_WARNING(Service_ACC, "(STUBBED) called");
-        constexpr u32 jpeg_size = 107;
+        LOG_DEBUG(Service_ACC, "called");
+        constexpr u32 backup_jpeg_size = 107;
         IPC::ResponseBuilder rb{ctx, 3};
         rb.Push(RESULT_SUCCESS);
-        rb.Push<u32>(jpeg_size);
+
+        if (image == nullptr)
+            rb.Push<u32>(backup_jpeg_size);
+        else
+            rb.Push<u32>(std::min<u32>(image->GetSize(), MAX_JPEG_IMAGE_SIZE));
     }
 
     const ProfileManager& profile_manager;
     UUID user_id; ///< The user id this profile refers to.
+    std::unique_ptr<FileUtil::IOFile> image = nullptr;
 };
 
 class IManagerForApplication final : public ServiceFramework<IManagerForApplication> {
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
 #include <array>
 #include "common/common_paths.h"
 #include "common/common_types.h"
@@ -33,9 +34,9 @@ struct UserData {
 };
 static_assert(sizeof(UserData) == 0x80, "UserData structure has incorrect size");
 
-static std::string GetImagePath(const std::string& username) {
-    return FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir) + "users" + DIR_SEP + username +
-           ".jpg";
+static std::string GetImagePath(UUID uuid) {
+    return FileUtil::GetUserPath(FileUtil::UserPath::NANDDir) +
+           "/system/save/8000000000000010/su/avators/" + uuid.FormatSwitch() + ".jpg";
 }
 
 class IProfile final : public ServiceFramework<IProfile> {
@@ -49,15 +50,6 @@ public:
             {11, &IProfile::LoadImage, "LoadImage"},
         };
         RegisterHandlers(functions);
-
-        ProfileBase profile_base{};
-        if (profile_manager.GetProfileBase(user_id, profile_base)) {
-            image = std::make_unique<FileUtil::IOFile>(
-                GetImagePath(Common::StringFromFixedZeroTerminatedBuffer(
-                    reinterpret_cast<const char*>(profile_base.username.data()),
-                    profile_base.username.size())),
-                "rb");
-        }
     }
 
 private:
@@ -111,13 +103,15 @@ private:
         IPC::ResponseBuilder rb{ctx, 3};
         rb.Push(RESULT_SUCCESS);
 
-        if (image == nullptr) {
+        const FileUtil::IOFile image(GetImagePath(user_id), "rb");
+
+        if (!image.IsOpen()) {
             ctx.WriteBuffer(backup_jpeg);
             rb.Push<u32>(backup_jpeg_size);
         } else {
-            const auto size = std::min<u32>(image->GetSize(), MAX_JPEG_IMAGE_SIZE);
+            const auto size = std::min<u32>(image.GetSize(), MAX_JPEG_IMAGE_SIZE);
             std::vector<u8> buffer(size);
-            image->ReadBytes(buffer.data(), buffer.size());
+            image.ReadBytes(buffer.data(), buffer.size());
 
             ctx.WriteBuffer(buffer.data(), buffer.size());
             rb.Push<u32>(buffer.size());
@@ -130,15 +124,16 @@ private:
         IPC::ResponseBuilder rb{ctx, 3};
         rb.Push(RESULT_SUCCESS);
 
-        if (image == nullptr)
+        const FileUtil::IOFile image(GetImagePath(user_id), "rb");
+
+        if (!image.IsOpen())
             rb.Push<u32>(backup_jpeg_size);
         else
-            rb.Push<u32>(std::min<u32>(image->GetSize(), MAX_JPEG_IMAGE_SIZE));
+            rb.Push<u32>(std::min<u32>(image.GetSize(), MAX_JPEG_IMAGE_SIZE));
     }
 
     const ProfileManager& profile_manager;
     UUID user_id; ///< The user id this profile refers to.
-    std::unique_ptr<FileUtil::IOFile> image = nullptr;
 };
 
 class IManagerForApplication final : public ServiceFramework<IManagerForApplication> {
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -106,6 +106,8 @@ private:
         const FileUtil::IOFile image(GetImagePath(user_id), "rb");
 
         if (!image.IsOpen()) {
+            LOG_WARNING(Service_ACC,
+                        "Failed to load user provided image! Falling back to built-in backup...");
             ctx.WriteBuffer(backup_jpeg);
             rb.Push<u32>(backup_jpeg_size);
         } else {
@@ -126,10 +128,13 @@ private:
 
         const FileUtil::IOFile image(GetImagePath(user_id), "rb");
 
-        if (!image.IsOpen())
+        if (!image.IsOpen()) {
+            LOG_WARNING(Service_ACC,
+                        "Failed to load user provided image! Falling back to built-in backup...");
             rb.Push<u32>(backup_jpeg_size);
-        else
+        } else {
             rb.Push<u32>(std::min<u32>(image.GetSize(), MAX_JPEG_IMAGE_SIZE));
+        }
     }
 
     const ProfileManager& profile_manager;
--- a/src/core/hle/service/acc/profile_manager.cpp
+++ b/src/core/hle/service/acc/profile_manager.cpp
@@ -30,22 +30,20 @@ constexpr ResultCode ERROR_TOO_MANY_USERS(ErrorModule::Account, -1);
 constexpr ResultCode ERROR_USER_ALREADY_EXISTS(ErrorModule::Account, -2);
 constexpr ResultCode ERROR_ARGUMENT_IS_NULL(ErrorModule::Account, 20);
 
-constexpr const char* ACC_SAVE_AVATORS_BASE_PATH = "/system/save/8000000000000010/su/avators/";
+constexpr char ACC_SAVE_AVATORS_BASE_PATH[] = "/system/save/8000000000000010/su/avators/";
 
-const UUID& UUID::Generate() {
+UUID UUID::Generate() {
     std::random_device device;
     std::mt19937 gen(device());
     std::uniform_int_distribution<u64> distribution(1, std::numeric_limits<u64>::max());
-    uuid[0] = distribution(gen);
-    uuid[1] = distribution(gen);
-    return *this;
+    return UUID{distribution(gen), distribution(gen)};
 }
 
 ProfileManager::ProfileManager() {
     ParseUserSaveFile();
 
     if (user_count == 0)
-        CreateNewUser(UUID{}.Generate(), "yuzu");
+        CreateNewUser(UUID::Generate(), "yuzu");
 
     auto current = std::clamp<int>(Settings::values.current_user, 0, MAX_USERS - 1);
     if (UserExistsIndex(current))
@@ -309,10 +307,18 @@ void ProfileManager::ParseUserSaveFile() {
                               ACC_SAVE_AVATORS_BASE_PATH + "profiles.dat",
                           "rb");
 
-    ProfileDataRaw data;
-    save.Seek(0, SEEK_SET);
-    if (save.ReadBytes(&data, sizeof(ProfileDataRaw)) != sizeof(ProfileDataRaw))
+    if (!save.IsOpen()) {
+        LOG_WARNING(Service_ACC, "Failed to load profile data from save data... Generating new "
+                                 "user 'yuzu' with random UUID.");
         return;
+    }
+
+    ProfileDataRaw data;
+    if (save.ReadBytes(&data, sizeof(ProfileDataRaw)) != sizeof(ProfileDataRaw)) {
+        LOG_WARNING(Service_ACC, "profiles.dat is smaller than expected... Generating new user "
+                                 "'yuzu' with random UUID.");
+        return;
+    }
 
     for (std::size_t i = 0; i < MAX_USERS; ++i) {
         const auto& user = data.users[i];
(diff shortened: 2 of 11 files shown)


src/yuzu/configuration/configure_system.cpp
```diff
@@ -92,3 +189,67 @@ void ConfigureSystem::refreshConsoleID() {
...
 }
+
+void ConfigureSystem::SelectUser(const QModelIndex& index) {
+    Settings::values.current_user =
+        std::clamp<std::size_t>(index.row(), 0, Settings::values.users.size() - 1);
+
+    UpdateCurrentUser();
+
+    if (Settings::values.users.size() >= 2)
+        ui->pm_remove->setEnabled(true);
```
lioncash: This can collapse to:
```cpp
ui->pm_remove->setEnabled(Settings::values.users.size() >= 2);
```

src/yuzu/configuration/configure_system.cpp
```diff
@@ -92,3 +189,67 @@ void ConfigureSystem::refreshConsoleID() {
...
+        ui->pm_remove->setEnabled(false);
+
+    ui->pm_rename->setEnabled(true);
+}
+
+void ConfigureSystem::AddUser() {
+    Service::Account::UUID uuid;
+    uuid.Generate();
+
+    bool ok = false;
```
lioncash: Shouldn't the value of `ok` be checked after the `getText()` call below.

src/yuzu/configuration/config.cpp
```diff
@@ -122,7 +122,25 @@ void Config::ReadValues() {
 
     qt_config->beginGroup("System");
     Settings::values.use_docked_mode = qt_config->value("use_docked_mode", false).toBool();
-    Settings::values.username = qt_config->value("username", "yuzu").toString().toStdString();
+
+    Settings::values.users.clear();
```
lioncash: The code to load all the users should be placed within its own function.

src/yuzu/configuration/config.cpp
```diff
@@ -247,7 +265,19 @@ void Config::SaveValues() {
 
     qt_config->beginGroup("System");
     qt_config->setValue("use_docked_mode", Settings::values.use_docked_mode);
-    qt_config->setValue("username", QString::fromStdString(Settings::values.username));
+    qt_config->setValue("current_user", Settings::values.current_user);
```
lioncash: Ditto here, but with the saving code.

src/yuzu/configuration/config.cpp
```diff
@@ -247,7 +265,19 @@ void Config::SaveValues() {
...
+
+    qt_config->beginWriteArray("users", Settings::values.users.size());
+    for (std::size_t i = 0; i < Settings::values.users.size(); ++i) {
+        qt_config->setArrayIndex(i);
```
lioncash: `i` should be casted to an `int`, otherwise this will cause a truncation warning.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -2,7 +2,12 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <QGraphicsItem>
+#include <QList>
 #include <QMessageBox>
+#include <qinputdialog.h>
```
lioncash: `<QInputDialog>`

src/yuzu/configuration/configure_system.cpp
```diff
@@ -32,23 +48,104 @@ ConfigureSystem::ConfigureSystem(QWidget* parent) : QWidget(parent), ui(new Ui::
...
+                       QPixmap icon{icon_url};
+
+                       if (!icon) {
+                           icon.fill(QColor::fromRgb(0, 0, 0));
+                           icon.loadFromData(backup_jpeg.data(), backup_jpeg.size());
+                       }
+
+                       return QList{new QStandardItem{
+                           icon.scaled(64, 64, Qt::IgnoreAspectRatio, Qt::SmoothTransformation),
+                           QString::fromStdString(user.first + "\n" + user.second.Format())}};
```
lioncash: '\n'

src/yuzu/configuration/configure_system.cpp
```diff
@@ -32,23 +48,104 @@ ConfigureSystem::ConfigureSystem(QWidget* parent) : QWidget(parent), ui(new Ui::
...
+
     ui->combo_language->setCurrentIndex(Settings::values.language_index);
+
+    item_model->removeRows(0, item_model->rowCount());
+    list_items.clear();
+
+    std::transform(Settings::values.users.begin(), Settings::values.users.end(),
+                   std::back_inserter(list_items),
+                   [](const std::pair<std::string, Service::Account::UUID>& user) {
+                       const auto icon_url = QString::fromStdString(
```
lioncash: This lambda is essentially partially duplicating the `UpdateCurrentUser()` function.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -92,3 +189,67 @@ void ConfigureSystem::refreshConsoleID() {
...
+    uuid.Generate();
+
+    bool ok = false;
+    const auto username =
+        QInputDialog::getText(this, tr("Enter Username"), tr("Enter a username for the new user:"),
+                              QLineEdit::Normal, QString(), &ok);
+
+    Settings::values.users.emplace_back(username.toStdString(), uuid);
+
+    setConfiguration();
```
lioncash: It seems really janky to need to completely reload the entire user list if only one new user is added.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -92,3 +189,67 @@ void ConfigureSystem::refreshConsoleID() {
...
+    const auto new_username = QInputDialog::getText(
+        this, tr("Enter Username"), tr("Enter a new username:"), QLineEdit::Normal,
+        QString::fromStdString(Settings::values.users[user].first), &ok);
+
+    if (!ok)
+        return;
+
+    Settings::values.users[user].first = new_username.toStdString();
+
+    setConfiguration();
```
lioncash: Especially so for renaming.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -92,3 +189,67 @@ void ConfigureSystem::refreshConsoleID() {
...
+
+    if (confirm == QMessageBox::No)
+        return;
+
+    if (Settings::values.current_user == tree_view->currentIndex().row())
+        Settings::values.current_user = 0;
+
+    Settings::values.users.erase(user);
+
+    setConfiguration();
```
lioncash: Ditto for removing a user.

src/yuzu/configuration/configure_system.h
```diff
@@ -5,6 +5,11 @@
 #pragma once
 
 #include <memory>
+#include <QGraphicsScene>
```
lioncash: Many of these can be forward declarations.

src/yuzu/main.cpp
```diff
@@ -725,12 +725,33 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
         open_target = "Save Data";
         const std::string nand_dir = FileUtil::GetUserPath(FileUtil::UserPath::NANDDir);
         ASSERT(program_id != 0);
-        // TODO(tech4me): Update this to work with arbitrary user profile
-        // Refer to core/hle/service/acc/profile_manager.cpp ProfileManager constructor
-        constexpr u128 user_id = {1, 0};
+
+        QStringList list{};
```
lioncash: `{}` isn't necessary here.

src/yuzu/main.cpp
```diff
@@ -725,12 +725,33 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
         open_target = "Save Data";
         const std::string nand_dir = FileUtil::GetUserPath(FileUtil::UserPath::NANDDir);
         ASSERT(program_id != 0);
-        // TODO(tech4me): Update this to work with arbitrary user profile
-        // Refer to core/hle/service/acc/profile_manager.cpp ProfileManager constructor
-        constexpr u128 user_id = {1, 0};
+
```
lioncash: This code should probably go within its own function.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -32,23 +48,104 @@ ConfigureSystem::ConfigureSystem(QWidget* parent) : QWidget(parent), ui(new Ui::
...
+
+void ConfigureSystem::UpdateCurrentUser() {
+    const auto& current_user = Settings::values.users[Settings::values.current_user];
+    const auto icon_url =
+        QString::fromStdString(FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir) + "users" +
+                               DIR_SEP + current_user.first + ".jpg");
+    QPixmap icon{icon_url};
+
+    if (!icon) {
+        icon.fill(QColor::fromRgb(0, 0, 0));
```
lioncash: just calling `icon.fill(Qt::black);` will perform the same behavior.

src/core/hle/service/am/am.cpp
```diff
@@ -670,16 +672,16 @@ IApplicationFunctions::IApplicationFunctions() : ServiceFramework("IApplicationF
...
-        0,    0,    0,    0,    // User Id (word 1)
-        0,    0,    0,    0,    // User Id (word 2)
-        0,    0,    0,    0     // User Id (word 3)
-    }};
+    };
+
+    std::vector<u8> buffer(POP_LAUNCH_PARAMETER_BUFFER_SIZE);
 
-    std::vector<u8> buffer(data.begin(), data.end());
+    std::memcpy(buffer.data(), header_data.data(), header_data.size());
```
lioncash: `<cstring>` needs to be included.

src/core/hle/service/acc/acc.cpp
```diff
@@ -87,22 +106,38 @@ class IProfile final : public ServiceFramework<IProfile> {
...
+
+        if (image == nullptr)
+            rb.Push<u32>(backup_jpeg_size);
+        else
+            rb.Push<u32>(std::min<u32>(image->GetSize(), MAX_JPEG_IMAGE_SIZE));
     }
 
     const ProfileManager& profile_manager;
     UUID user_id; ///< The user id this profile refers to.
+    std::unique_ptr<FileUtil::IOFile> image = nullptr;
```
lioncash: Why does this need to contain a file handle the whole time it's active rather than just the data within the file?. At least that way avoids holding a handle open to a file.

src/core/hle/service/acc/acc.cpp
```diff
@@ -87,22 +106,38 @@ class IProfile final : public ServiceFramework<IProfile> {
...
+
         IPC::ResponseBuilder rb{ctx, 3};
         rb.Push(RESULT_SUCCESS);
-        rb.Push<u32>(jpeg_size);
+
+        if (image == nullptr) {
+            ctx.WriteBuffer(backup_jpeg);
+            rb.Push<u32>(backup_jpeg_size);
+        } else {
+            const auto size = std::min<u32>(image->GetSize(), MAX_JPEG_IMAGE_SIZE);
```
lioncash: `<algorithm>` needs to be included.

src/core/settings.h
```diff
@@ -113,7 +114,8 @@ static const std::array<const char*, NumAnalogs> mapping = {{
 struct Values {
     // System
     bool use_docked_mode;
-    std::string username;
+    int current_user;
+    std::vector<std::pair<std::string, Service::Account::UUID>> users;
```
lioncash: Fairly sure it's not a good idea to dump service internals into the main settings header.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -32,23 +48,104 @@ ConfigureSystem::ConfigureSystem(QWidget* parent) : QWidget(parent), ui(new Ui::
...
 void ConfigureSystem::setConfiguration() {
     enabled = !Core::System::GetInstance().IsPoweredOn();
-    ui->edit_username->setText(QString::fromStdString(Settings::values.username));
+
     ui->combo_language->setCurrentIndex(Settings::values.language_index);
+
+    item_model->removeRows(0, item_model->rowCount());
+    list_items.clear();
+
+    std::transform(Settings::values.users.begin(), Settings::values.users.end(),
```
lioncash: `<algorithm>` needs to be included.

Subv:
Where is the list of users actually stored in the switch? I'm of the opinion that we should try to keep as close compatibility as possible with the real system so following that convention would be preferable.

ogniK5377:
> Where is the list of users actually stored in the switch? I'm of the opinion that we should try to keep as close compatibility as possible with the real system so following that convention would be preferable.

The users are stored in a system save, as of currently we have no way of openning the save with our current backend and requires a lot of work. However in the future we should switch from the config -> the official syssave at some point

DarkLordZach:
@Subv fixed

src/core/hle/service/am/am.cpp
```diff
@@ -680,8 +682,10 @@ void IApplicationFunctions::PopLaunchParameter(Kernel::HLERequestContext& ctx) {
     std::vector<u8> buffer(POP_LAUNCH_PARAMETER_BUFFER_SIZE);
 
     std::memcpy(buffer.data(), header_data.data(), header_data.size());
-    const auto current_uuid = Settings::values.users[Settings::values.current_user].second.uuid;
-    std::memcpy(buffer.data() + header_data.size(), current_uuid.data(), sizeof(u128));
+
+    Account::ProfileManager profile_manager{};
+    const auto uuid = profile_manager.GetAllUsers()[Settings::values.current_user].uuid;
```
Subv: Again, don't we have a GetUser? I'll stop pointing this out.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -99,49 +112,69 @@ void ConfigureSystem::setConfiguration() {
...
-                       }
-
-                       return QList{new QStandardItem{
-                           icon.scaled(64, 64, Qt::IgnoreAspectRatio, Qt::SmoothTransformation),
-                           QString::fromStdString(user.first + "\n" + user.second.Format())}};
-                   });
-
-    for (const auto& item : list_items)
-        item_model->appendRow(item);
+    ui->pm_add->setEnabled(profile_manager->GetUserCount() < 8);
```
Subv: what is 8?

src/yuzu/configuration/config.cpp
```diff
@@ -123,23 +124,7 @@ void Config::ReadValues() {
...
-    }
-
-    qt_config->endArray();
-
-    if (Settings::values.users.empty())
-        Settings::values.users.emplace_back("yuzu", Service::Account::UUID{}.Generate());
-
-    Settings::values.current_user =
-        std::clamp(qt_config->value("current_user", 0).toInt(), 0, size);
+    Settings::values.current_user = std::clamp(qt_config->value("current_user", 0).toInt(), 0, 7);
```
Subv: don't use magic numbers, `ProfileManager::MaxUsers - 1`

src/yuzu/configuration/configure_system.cpp
```diff
@@ -99,49 +112,69 @@ void ConfigureSystem::setConfiguration() {
...
+            const auto username = Common::StringFromFixedZeroTerminatedBuffer(
+                reinterpret_cast<const char*>(profile.username.data()), profile.username.size());
+
+            return QList<QStandardItem*>{new QStandardItem{
+                GetIcon(user).scaled(64, 64, Qt::IgnoreAspectRatio, Qt::SmoothTransformation),
+                QString::fromStdString(username + '\n' + user.FormatSwitch())}};
+        });
+
+    list_items.erase(
+        std::remove_if(list_items.begin(), list_items.end(),
```
Subv: what is this doing and why? Add a comment

src/yuzu/configuration/configure_system.cpp
```diff
@@ -99,49 +112,69 @@ void ConfigureSystem::setConfiguration() {
...
+        icon.fill(Qt::black);
         icon.loadFromData(backup_jpeg.data(), backup_jpeg.size());
     }
 
+    return icon;
+}
+
+void ConfigureSystem::PopulateUserList() {
+    const auto& profiles = profile_manager->GetAllUsers();
+    std::transform(
```
Subv: You're not doing a transform operation, you're doing a filter+transform, IMO it's better to have this as a range-based for

src/yuzu/configuration/configure_system.cpp
```diff
@@ -212,44 +245,96 @@ void ConfigureSystem::AddUser() {
...
 
-    setConfiguration();
+    item_model->appendRow(new QStandardItem{
+        GetIcon(uuid).scaled(64, 64, Qt::IgnoreAspectRatio, Qt::SmoothTransformation),
+        QString::fromStdString(username.toStdString() + '\n' + uuid.FormatSwitch())});
 }
 
 void ConfigureSystem::RenameUser() {
     const auto user = tree_view->currentIndex().row();
+    ASSERT(user < 8);
```
Subv: 8?

src/yuzu/configuration/configure_system.cpp
```diff
@@ -212,44 +245,96 @@ void ConfigureSystem::AddUser() {
...
+        QInputDialog::getText(this, tr("Enter Username"), tr("Enter a new username:"),
+                              QLineEdit::Normal, QString::fromStdString(username), &ok);
 
     if (!ok)
         return;
 
-    Settings::values.users[user].first = new_username.toStdString();
+    const auto username_std = new_username.toStdString();
+    if (username_std.size() > profile.username.size())
+        std::copy_n(username_std.begin(), profile.username.size(), profile.username.begin());
```
Subv: `std::copy_n(username_std.begin(), std::min(profile.username.size(), username_std.size()), profile.username.begin());`

src/yuzu/main.cpp
```diff
@@ -726,10 +727,22 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
         ASSERT(program_id != 0);
 
-        QStringList list{};
-        std::transform(Settings::values.users.begin(), Settings::values.users.end(),
-                       std::back_inserter(list),
-                       [](const auto& user) { return QString::fromStdString(user.first); });
+        Service::Account::ProfileManager manager{};
+        const auto user_ids = manager.GetAllUsers();
+        QStringList list;
+        std::transform(
```
Subv: Again, you're not performing a map but a filter+map, this would be better off as a range-based loop

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -241,4 +264,70 @@ bool ProfileManager::CanSystemRegisterUser() const {
...
 }
 
+bool ProfileManager::RemoveUser(UUID uuid) {
+    auto index = GetUserIndex(uuid);
+    if (index == boost::none) {
+        return false;
+    }
+
+    profiles[*index] = ProfileInfo{};
+    std::stable_partition(profiles.begin(), profiles.end(),
```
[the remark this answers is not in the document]
DarkLordZach: If the user was not the last user, it has to be put at the back. From my understanding, both the class and switch operate on the property that the users array has all of the users with invalid UUIDs (0) at the end.

src/yuzu/configuration/configure_system.h
```diff
@@ -5,7 +5,16 @@
 #pragma once
 
 #include <memory>
+
+#include <QList>
 #include <QWidget>
+#include "core/hle/service/acc/profile_manager.h"
+
+class QVBoxLayout;
```
lioncash: These should be alphabetically ordered.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -24,31 +40,141 @@ static const std::array<int, 12> days_in_month = {{
...
+
+    scene->clear();
+    scene->addPixmap(
+        GetIcon(*current_user).scaled(48, 48, Qt::IgnoreAspectRatio, Qt::SmoothTransformation));
+    ui->current_user_username->setText(QString::fromStdString(username));
 }
 
 void ConfigureSystem::ReadSystemSettings() {}
 
+std::string ConfigureSystem::GetAccountUsername(Service::Account::UUID uuid) {
```
lioncash: This can be a const member function

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -239,4 +278,76 @@ bool ProfileManager::CanSystemRegisterUser() const {
...
+    return true;
+}
+
+void ProfileManager::ParseUserSaveFile() {
+    FileUtil::IOFile save(FileUtil::GetUserPath(FileUtil::UserPath::NANDDir) +
+                              ACC_SAVE_AVATORS_BASE_PATH + "profiles.dat",
+                          "rb");
+
+    ProfileDataRaw data;
+    save.Seek(0, SEEK_SET);
```
lioncash: You don't need to seek to the beginning on an immediately opened file. It's already at offset 0.

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -239,4 +278,76 @@ bool ProfileManager::CanSystemRegisterUser() const {
...
+    auto& profile = profiles[*index];
+    profile.user_uuid = profile_new.user_uuid;
+    profile.username = profile_new.username;
+    profile.creation_time = profile_new.timestamp;
+
+    return true;
+}
+
+void ProfileManager::ParseUserSaveFile() {
+    FileUtil::IOFile save(FileUtil::GetUserPath(FileUtil::UserPath::NANDDir) +
```
lioncash: There should be handling code for this not opening properly.

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -239,4 +278,76 @@ bool ProfileManager::CanSystemRegisterUser() const {
...
+    }
+
+    const auto raw_path =
+        FileUtil::GetUserPath(FileUtil::UserPath::NANDDir) + "/system/save/8000000000000010";
+    if (FileUtil::Exists(raw_path) && !FileUtil::IsDirectory(raw_path))
+        FileUtil::Delete(raw_path);
+
+    const auto path = FileUtil::GetUserPath(FileUtil::UserPath::NANDDir) +
+                      ACC_SAVE_AVATORS_BASE_PATH + "profiles.dat";
+    FileUtil::CreateFullPath(path);
```
lioncash: This is something that can fail.

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -239,4 +278,76 @@ bool ProfileManager::CanSystemRegisterUser() const {
...
+    FileUtil::IOFile save(path, "wb");
```
lioncash: There should be code to handle this failing to open.

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -239,4 +278,76 @@ bool ProfileManager::CanSystemRegisterUser() const {
...
+
+    save.Resize(sizeof(ProfileDataRaw));
+    save.Seek(0, SEEK_SET);
```
lioncash: This seek is unnecessary.

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -23,13 +42,21 @@ const UUID& UUID::Generate() {
...
 
 ProfileManager::ProfileManager() {
-    // TODO(ogniK): Create the default user we have for now until loading/saving users is added
-    auto user_uuid = UUID{1, 0};
-    ASSERT(CreateNewUser(user_uuid, Settings::values.username).IsSuccess());
-    OpenUser(user_uuid);
+    ParseUserSaveFile();
+
+    if (user_count == 0)
+        CreateNewUser(UUID{}.Generate(), "yuzu");
```
lioncash: You may want to consider just making `Generate()` a static function and changing exiting usages to operate in terms of that, given this is pretty wonky.

src/core/hle/service/acc/profile_manager.cpp
```diff
@@ -4,15 +4,34 @@
...
+    std::array<UserRaw, MAX_USERS> users;
+};
+static_assert(sizeof(ProfileDataRaw) == 0x650, "ProfileDataRaw has incorrect size.");
+
 // TODO(ogniK): Get actual error codes
 constexpr ResultCode ERROR_TOO_MANY_USERS(ErrorModule::Account, -1);
 constexpr ResultCode ERROR_USER_ALREADY_EXISTS(ErrorModule::Account, -2);
 constexpr ResultCode ERROR_ARGUMENT_IS_NULL(ErrorModule::Account, 20);
 
+constexpr const char* ACC_SAVE_AVATORS_BASE_PATH = "/system/save/8000000000000010/su/avators/";
```
lioncash:
```cpp
constexpr char ACC_SAVE_AVATORS_BASE_PATH[] = "/system/save/8000000000000010/su/avators/";
```

Is more space-efficient, given it only stores the string data. The previous code stores both the string data and a pointer address into the binary.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -92,3 +218,125 @@ void ConfigureSystem::refreshConsoleID() {
...
+    if (!ok)
+        return;
+
+    const auto username_std = new_username.toStdString();
+    if (username_std.size() > profile.username.size()) {
+        std::copy_n(username_std.begin(), std::min(profile.username.size(), username_std.size()),
+                    profile.username.begin());
+    } else {
+        std::copy(username_std.begin(), username_std.end(), profile.username.begin());
+    }
```
kyn3tic: I don't believe this properly handles when the new name is shorter than the existing name. After renaming with a short name, the resulting name retains its original length with only the first n characters replaced. The list UI will however show the new name correctly until it's reopened.

DarkLordZach: Fixed in 1c5885d by clearing the username array before copying the new name in.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -92,3 +218,126 @@ void ConfigureSystem::refreshConsoleID() {
...
+
+    profile_manager->SetProfileBase(*uuid, profile);
+
+    item_model->setItem(
+        user, 0,
+        new QStandardItem{
+            GetIcon(*uuid).scaled(64, 64, Qt::IgnoreAspectRatio, Qt::SmoothTransformation),
+            QString::fromStdString(username_std + '\n' + uuid->FormatSwitch())});
```
lioncash: This should probably be a translatable string like:

```cpp
tr("%1\n%2", 
   "%1 is the profile username, %2 is the formatted UUID (e.g. 00112233-4455-6677-8899-AABBCCDDEEFF))")
        .arg(QString::fromStdString(username_std), QString::fromStdString(uuid->FormatSwitch()));
```

so translators have control over the text flow.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -92,3 +218,126 @@ void ConfigureSystem::refreshConsoleID() {
...
+
+void ConfigureSystem::DeleteUser() {
+    const auto index = tree_view->currentIndex().row();
+    const auto uuid = profile_manager->GetUser(index);
+    ASSERT(uuid != boost::none);
+    const auto username = GetAccountUsername(*uuid);
+
+    const auto confirm = QMessageBox::question(
+        this, tr("Confirm Delete"),
+        tr("You are about to delete user with name %1. Are you sure?").arg(username.c_str()));
```
lioncash: `.arg(QString::fromStdString(username))`

src/yuzu/configuration/configure_system.h
```diff
@@ -5,7 +5,16 @@
 #pragma once
 
 #include <memory>
+
+#include <QList>
 #include <QWidget>
+#include "core/hle/service/acc/profile_manager.h"
```
lioncash: This can be replaced with forward declarations of

```cpp
namespace Service::Account {
class ProfileManager;
struct UUID;
};
```

and have the include be moved into the cpp file.

PR #1470, opened 2018-10-10
Implemented Alpha Test using Shader Emulation

FernandoS27:
Here's an implementation of alpha testing using shaders. This should be used by some games. Please test this PR and let me know if performance degrades to some heavy extent.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -643,13 +643,6 @@ private:
                                  ';');
         }
         declarations.AddNewLine();
-
-        if (stage == Maxwell3D::Regs::ShaderStage::Fragment) {
-            declarations.AddLine("uniform bool alpha_testing_active;");
-            declarations.AddLine("uniform float alpha_testing_ref;");
-            declarations.AddLine("uniform uint alpha_testing_func;");
-        }
-        declarations.AddNewLine();
     }
 
     /// Generates declarations used for geometry shaders.
@@ -1271,17 +1264,20 @@ private:
 
         ASSERT_MSG(header.ps.omap.sample_mask == 0, "Samplemask write is unimplemented");
 
-        shader.AddLine("if (alpha_testing_active) {");
+        shader.AddLine("if (alpha_test[0] != 0) {");
         ++shader.scope;
+        // We start on the register containing the alpha value in the first RT.
         u32 current_reg = 3;
         for (u32 render_target = 0; render_target < Maxwell3D::Regs::NumRenderTargets;
              ++render_target) {
+            // TODO(Blinkhawk): verify the behavior of alpha testing on hardware when
+            // multiple render targets are used.
             if (header.ps.IsColorComponentOutputEnabled(render_target, 0) ||
                 header.ps.IsColorComponentOutputEnabled(render_target, 1) ||
                 header.ps.IsColorComponentOutputEnabled(render_target, 2) ||
                 header.ps.IsColorComponentOutputEnabled(render_target, 3)) {
                 shader.AddLine(fmt::format(
-                    "if (AlphaFunc({}, alpha_testing_ref, alpha_testing_func)) discard;",
+                    "if (AlphaFunc({})) discard;",
                     regs.GetRegisterAsFloat(current_reg)));
                 current_reg += 4;
             }
@@ -3506,38 +3502,6 @@ private:
             declarations.AddLine("bool " + pred + " = false;");
         }
         declarations.AddNewLine();
-        GenerateFunctionDeclarations();
-    }
-
-    void GenerateFunctionDeclarations() {
-        if (stage == Maxwell3D::Regs::ShaderStage::Fragment) {
-            declarations.AddLine("bool AlphaFunc(in float value, in float ref, in uint func) {");
-            declarations.scope++;
-            declarations.AddLine("switch (func) {");
-            declarations.scope++;
-            declarations.AddLine("case 1:");
-            declarations.AddLine("return false;");
-            declarations.AddLine("case 2:");
-            declarations.AddLine("return value < ref;");
-            declarations.AddLine("case 3:");
-            declarations.AddLine("return value == ref;");
-            declarations.AddLine("case 4:");
-            declarations.AddLine("return value <= ref;");
-            declarations.AddLine("case 5:");
-            declarations.AddLine("return value > ref;");
-            declarations.AddLine("case 6:");
-            declarations.AddLine("return value != ref;");
-            declarations.AddLine("case 7:");
-            declarations.AddLine("return value >= ref;");
-            declarations.AddLine("case 8:");
-            declarations.AddLine("return true;");
-            declarations.AddLine("default:");
-            declarations.AddLine("return false;");
-            declarations.scope--;
-            declarations.AddLine('}');
-            declarations.scope--;
-            declarations.AddLine('}');
-        }
     }
 
 private:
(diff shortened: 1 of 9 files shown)


pendingchaos:
Just a thought, but perhaps something like
```
bool ShouldDiscard(in float value) {
   uint func = alpha_testing_func;
   func -= 1;
   if (alpha_testing_active && func <= 7) { 
      bool lt = bool(func & 1) && value < alpha_testing_ref;
      bool eq = bool(func & 2) && value == alpha_testing_ref;
      bool gt = bool(func & 4) && value > alpha_testing_ref;
      return lt || eq || gt;
   }
   return alpha_testing_active;
}
```
could be more efficient than a switch statement. Although uniform, the switch may create a if-else chain or binary search instead of a jump-table depending on how well a driver handles them.

I don't know what the generated code of each idea is though, on any driver/device, other than a vague memory of a if-else chain being created on nouveau and mesa seems to not have any (used) representation of switches in TGSI or NIR.

FernandoS27:
@pendingchaos tested your version and it's slower on NVIDIA Propietary driver.

In general, Alpha Testing is very costy, I think I'll add a setting to activate it or not.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1185,6 +1193,26 @@ class GLSLGenerator {
...
+            shader.AddLine("if (alpha_testing_active) {");
+            ++shader.scope;
+            u32 current_reg = 3;
+            for (u32 render_target = 0; render_target < Maxwell3D::Regs::NumRenderTargets;
+                 ++render_target) {
+                if (header.ps.IsColorComponentOutputEnabled(render_target, 0) ||
+                    header.ps.IsColorComponentOutputEnabled(render_target, 1) ||
+                    header.ps.IsColorComponentOutputEnabled(render_target, 2) ||
+                    header.ps.IsColorComponentOutputEnabled(render_target, 3)) {
+                    shader.AddLine(fmt::format(
```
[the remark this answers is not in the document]
FernandoS27: That just checks if any of the outputs it's enabled. If none is, then we just don't perform a test for them.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -3071,6 +3099,38 @@ class GLSLGenerator {
             declarations.AddLine("bool " + pred + " = false;");
         }
         declarations.AddNewLine();
+        GenerateFunctionDeclarations();
+    }
+
+    void GenerateFunctionDeclarations() {
+        if (stage == Maxwell3D::Regs::ShaderStage::Fragment && Settings::values.use_alpha_testing) {
+            declarations.AddLine("bool AlphaFunc(in float value, in float ref, in uint func) {");
```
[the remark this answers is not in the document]
ReinUsesLisp: Also this should be declared in `gl_shader_gen.cpp` not in the decompiler.

FernandoS27:
changes made and perf fixed.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -27,6 +27,10 @@ layout(std140) uniform vs_config {
     vec4 viewport_flip;
     uvec4 instance_id;
     uvec4 flip_stage;
+    uint alpha_test_enabled;
+    uint alpha_test_func;
+    float alpha_test_ref;
+    uint alpha_test_padding;
```
ReinUsesLisp: This is going to be packed in 64 bytes. You'll have to do something like this:
```glsl
uvec4 alpha_test;
```
Then on your code:
```glsl
uint at_enabled = alpha_test[0];
uint at_func = alpha_test[1];
float at_ref = uintBitsToFloat(alpha_test[2]);
```

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -100,6 +104,10 @@ layout (std140) uniform gs_config {
...
```
ReinUsesLisp: Just pointing the same as above, so it doesn't get lost.

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -135,6 +143,10 @@ layout (std140) uniform fs_config {
...
```
ReinUsesLisp: Ditto.

src/video_core/renderer_opengl/gl_shader_manager.cpp
```diff
@@ -16,6 +16,17 @@ void MaxwellUniformData::SetFromRegs(const Maxwell3D::State::ShaderStageInfo& sh
     viewport_flip[0] = regs.viewport_transform[0].scale_x < 0.0 ? -1.0f : 1.0f;
     viewport_flip[1] = regs.viewport_transform[0].scale_y < 0.0 ? -1.0f : 1.0f;
 
+    u32 func = static_cast<u32>(regs.alpha_test_func);
+    // Normalize the gl variants of opCompare to be the same as the normal variants
+    u32 op = static_cast<u32>(Tegra::Engines::Maxwell3D::Regs::ComparisonOp::Never);
```
Subv: Pick a more meaningful name for these variables

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1185,9 +1185,28 @@ class GLSLGenerator {
 
         ASSERT_MSG(header.ps.omap.sample_mask == 0, "Samplemask write is unimplemented");
 
+        shader.AddLine("if (alpha_test_enabled != 0) {");
+        ++shader.scope;
+        // We start on the register containing the alpha value in the first RT.
+        u32 current_reg = 3;
+        for (u32 render_target = 0; render_target < Maxwell3D::Regs::NumRenderTargets;
+             ++render_target) {
+            if (header.ps.IsColorComponentOutputEnabled(render_target, 0) ||
```
Subv: > That just checks if any of the outputs it's enabled. If none is, then we just don't perform a test for them.

That's not what this code does. You'll fail the alpha test if /any/ of the output colors fails it, is that intended? Have you verified that this is the case in hardware?

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1185,9 +1185,30 @@ class GLSLGenerator {
...
+        for (u32 render_target = 0; render_target < Maxwell3D::Regs::NumRenderTargets;
+             ++render_target) {
+            // TODO(Blinkhawk): verify the behavior of alpha testing on hardware when
+            // multiple render targets are used.
+            if (header.ps.IsColorComponentOutputEnabled(render_target, 0) ||
+                header.ps.IsColorComponentOutputEnabled(render_target, 1) ||
+                header.ps.IsColorComponentOutputEnabled(render_target, 2) ||
+                header.ps.IsColorComponentOutputEnabled(render_target, 3)) {
+                shader.AddLine(fmt::format(
+                    "if (AlphaFunc({}, uintBitsToFloat(alpha_test[2]), alpha_test[1])) discard;",
```
ReinUsesLisp: You never check for `alpha_test[0]` being enabled.
Also these values can be read from `AlphaFunc`, it makes more sense because here `alpha_test` seems like a magic name (while in `gl_shader_gen` it's a declared right above).

FernandoS27: line 1188: shader.AddLine("if (alpha_test[0] != 0) {");

ReinUsesLisp: Yes, but I set it to a variable in the example, that's obscured when passed as a parameter. Just read them from `AlphaFunc` and make it take a single parameter (alpha's value).
About line 1188 I didn't notice.

MysticExile:
Pokken Tournament DX has issues with this PR.

Hexagon12:
The issues with Pokken, L.A.Noire and Has Been Heroes have been solved as of latest commits.

PR #1478, opened 2018-10-11
Passing an invalid nmap handle to Remap should throw an error

ogniK5377:
Added error for invalid nmap handles

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
@@ -15,6 +15,11 @@
 #include "video_core/renderer_base.h"
 
 namespace Service::Nvidia::Devices {
+namespace NvErrCodes {
+enum {
+    InvalidNmapHandle = -22,
+};
+}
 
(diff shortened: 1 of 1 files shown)


src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -79,14 +84,17 @@ u32 nvhost_as_gpu::Remap(const std::vector<u8>& input, std::vector<u8>& output)
...
     for (const auto& entry : entries) {
         LOG_WARNING(Service_NVDRV, "remap entry, offset=0x{:X} handle=0x{:X} pages=0x{:X}",
                     entry.offset, entry.nvmap_handle, entry.pages);
         Tegra::GPUVAddr offset = static_cast<Tegra::GPUVAddr>(entry.offset) << 0x10;
-
         auto object = nvmap_dev->GetObject(entry.nvmap_handle);
-        ASSERT(object);
+        if (!object) {
+            LOG_CRITICAL(Service_NVDRV, "nvmap {} is an invalid handle!", entry.nvmap_handle);
+            failed_remap = true;
```
Subv: Should this really be a continue instead of a break?

PR #1479, opened 2018-10-11
Added error codes for nvmap

ogniK5377:
Just to make it a bit more console accurate


src/core/hle/service/nvdrv/devices/nvmap.cpp
```diff
@@ -66,18 +75,35 @@ u32 nvmap::IocCreate(const std::vector<u8>& input, std::vector<u8>& output) {
...
+
+    if (!params.handle) {
+        return static_cast<u32>(NvErrCodes::InvalidValue);
+    }
+
+    if ((params.align - 1) & params.align) {
+        return static_cast<u32>(NvErrCodes::InvalidValue);
+    }
+
+    if (params.align < 0x1000) {
```
Subv: No magic numbers please

PR #1480, opened 2018-10-11
Introduce 3D Swizzle seamlessly 

FernandoS27:
The way we handled textures, in general, is very complex and hard to debug/check. This PR introduces a new algorithm for handling textures with the added bonus of being a lot faster and being able to tackle 3D Textures.

Some Pros of the new version:
- Faster texture unswizzling, about 6 times faster in most cases.
- Ability to decode 3D Textures.
- Easier to infer from logic, the structure of a blocklinear texture.
- No complex adress calculations.

Cons:
- More complex algorithm than old version (6 nested loops).
- Still does not handle block_width but it's easier to implement if we ever have to use it.
- Differs from most documentation is it tackles the problem from a geometric perspective rather than an adress perspective.

What it means for the User:
- this PR opens the door for many new features and bug fixes.
- Upcoming implementations like DMA, won't hit performance that hard.
- Faster loading times and more FPS on certain games.


ReinUsesLisp:
a general nit: `adress` -> `address`

PR #1488, opened 2018-10-13
video_core: Added ASTC 5x4; 8x5 types

Hexagon12:
Adds ASTC_2D_8X5 and ASTC_2D_5X4, which are both used in Donkey Kong Country: Tropical Freeze.


greggameplayer:
@Hexagon12 can you add ASTC_2D_5X5 too because it's needed by mario+rabbids kingdom battle and nintendo labo 1 and 2 ?

Hexagon12:
5x5 is too hard for me - seems like it behaves differently than the current ASTC types.

greggameplayer:
@Hexagon12 that's what i've seen when i was trying to implement it.

PR #1494, opened 2018-10-14
aoc: Fix various bugs in current AOC implementation

DarkLordZach:
1. Returns buffer size in ListAddOnContent, which is what the signature should be and should fix the bugs in games like Rocket League where it tries to access DLC not owned and svcBreaks.
2. Reads DLC base ID from Control metadata, which is how HOS seems to do it. If this cannot be found, it falls back to base TID + 0x1000, which is also the HOS fallback.

### DIFF
--- a/src/core/hle/service/aoc/aoc_u.cpp
+++ b/src/core/hle/service/aoc/aoc_u.cpp
@@ -7,8 +7,10 @@
 #include <vector>
 #include "common/logging/log.h"
 #include "core/file_sys/content_archive.h"
+#include "core/file_sys/control_metadata.h"
 #include "core/file_sys/nca_metadata.h"
 #include "core/file_sys/partition_filesystem.h"
+#include "core/file_sys/patch_manager.h"
 #include "core/file_sys/registered_cache.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/process.h"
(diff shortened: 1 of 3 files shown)


src/core/hle/service/aoc/aoc_u.cpp
```diff
@@ -97,14 +99,25 @@ void AOC_U::ListAddOnContent(Kernel::HLERequestContext& ctx) {
...
 
 void AOC_U::GetAddOnContentBaseId(Kernel::HLERequestContext& ctx) {
     IPC::ResponseBuilder rb{ctx, 4};
     rb.Push(RESULT_SUCCESS);
-    rb.Push(Core::System::GetInstance().CurrentProcess()->GetTitleID() | DLC_BASE_TO_AOC_ID_MASK);
+    const auto title_id = Core::System::GetInstance().CurrentProcess()->GetTitleID();
+    FileSys::PatchManager pm{title_id};
+
+    const auto res = pm.GetControlMetadata();
+    if (res == std::pair<std::unique_ptr<FileSys::NACP>, FileSys::VirtualFile>{} ||
```
lioncash: This can likely just be:

```cpp
if (res.first == nullptr || (res.first == nullptr && res.second == nullptr))
```
is much more readable than constructing an empty pair.

PR #1497, opened 2018-10-14
Implement flushing in the rasterizer cache

bunnei:
This implements flushing of surfaces in the rasterizer cache, and various other improvements. Surfaces are marked as "dirty" when they are rendered to, and then flushed to emulated memory when they are accessed directly.

Flushing will allow us to more accurately implement surface copies, DMA, etc. With this change, we do not really use this much yet - So I do not really expect any accuracy improvements yet. This change is behind the "use_accurate_framebuffers" setting, as it does incur a pretty big performance hit.

This partially supersedes #1339, fixing several of the issues we had with getting that change merged - mainly to do with how surface memory is managed.

~~This is WIP still for a few reasons, but I'd like to get feedback and I know others would like to use this. I plan to wrap this up pretty quickly though (hopefully this week). Still TODO:~~
- [X] Fix performance regressions even when flushing is turned off
- [X] Validate that flushed data is correct, especially for non-2D formats
- [X] Test more games, make sure there are no regressions

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -905,8 +905,6 @@ void CachedSurface::LoadGLBuffer() {
     }
 
     ConvertFormatAsNeeded_LoadGLBuffer(gl_buffer, params.pixel_format, params.width, params.height);
-
-    dirty = false;
 }
 
 MICROPROFILE_DEFINE(OpenGL_SurfaceFlush, "OpenGL", "Surface Flush", MP_RGB(128, 192, 64));
@@ -1111,6 +1109,7 @@ Surface RasterizerCacheOpenGL::GetColorBufferSurface(std::size_t index, bool pre
 void RasterizerCacheOpenGL::LoadSurface(const Surface& surface) {
     surface->LoadGLBuffer();
     surface->UploadGLTexture(read_framebuffer.handle, draw_framebuffer.handle);
+    surface->MarkAsModified(false, *this);
 }
 
 Surface RasterizerCacheOpenGL::GetSurface(const SurfaceParams& params, bool preserve_contents) {
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -1121,10 +1121,6 @@ void RasterizerCacheOpenGL::LoadSurface(const Surface& surface) {
     surface->UploadGLTexture(read_framebuffer.handle, draw_framebuffer.handle);
 }
 
-void RasterizerCacheOpenGL::FlushSurface(const Surface& surface) {
-    surface->FlushGLBuffer();
-}
-
 Surface RasterizerCacheOpenGL::GetSurface(const SurfaceParams& params, bool preserve_contents) {
     if (params.addr == 0 || params.height * params.width == 0) {
         return {};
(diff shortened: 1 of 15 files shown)


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -333,23 +351,24 @@ static bool IsFormatBCn(PixelFormat format) {
...
-        // check the configuration for this and perform more generic un/swizzle
-        LOG_WARNING(Render_OpenGL, "need to use correct swizzle/GOB parameters!");
-        VideoCore::MortonCopyPixels128(stride, height, bytes_per_pixel, gl_bytes_per_pixel,
-                                       Memory::GetPointer(addr), gl_buffer, morton_to_gl);
+        std::vector<u8> data(gl_buffer_size);
+        Tegra::Texture::CopySwizzledData(stride / tile_size, height / tile_size, depth,
+                                         bytes_per_pixel, bytes_per_pixel, data.data(), gl_buffer,
+                                         false, block_height, block_depth);
+        const std::size_t size_to_copy{std::min(gl_buffer_size, data.size())};
+        memcpy(Memory::GetPointer(addr), data.data(), size_to_copy);
```
FernandoS27: You can shy away of using a temporal buffer once we are certain there's no memory corruption.

FernandoS27:
Ok, after testing and looking up, Good Job. However, first set flushing to it's own setting called: Accurate Direct GPU Memory. Second, my last comment. That's it.

bunnei:
Regarding adding another setting, I'd prefer to just have one "slow but accurate framebuffers" setting. Too many knobs and switches just confuses users, and we will end up with certain combinations of settings for each game (as opposed to an "it just works" experience). The existing "use accurate framebuffers" setting should be used for any accurate rasterizer cache codepaths without much concern for performance.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -853,22 +895,67 @@ void CachedSurface::LoadGLBuffer() {
...
+
+    dirty = false;
 }
 
 MICROPROFILE_DEFINE(OpenGL_SurfaceFlush, "OpenGL", "Surface Flush", MP_RGB(128, 192, 64));
 void CachedSurface::FlushGLBuffer() {
-    ASSERT_MSG(false, "Unimplemented");
+    MICROPROFILE_SCOPE(OpenGL_SurfaceFlush);
+
+    if (!Settings::values.use_accurate_framebuffers) {
```
FernandoS27: Move this to the Rasterizer, so we can use Flushing on Blitting without needing it on.

FernandoS27:
Add a timestamp to the cacheObjects, whenever they are marked as dirty, a unsigned 64 bit integer from the RasterizerCache is increased by one and returned as new timestamp for the object. The timestamp should be ignored on hash key generation.

Then FlushRegion becomes:
```
    // Accurate Flushing
    void FlushRegion(Tegra::GPUVAddr addr, size_t size) {
        if (size == 0)
            return;
        std::vector<T> objects{};
         const ObjectInterval interval{addr, addr + size};
        for (auto& pair : boost::make_iterator_range(object_cache.equal_range(interval))) {
            for (auto& cached_object : pair.second) {
                if (!cached_object)
                    continue;
                objects.push_back(cached_object);
            }
        }
        std::sort(objects.begin(), objects.end(), [](const T& a, const T& b) -> bool {
            return a->GetTimestamp() > b->GetTimestamp();
        });
        for (auto& cached_object : objects) {
            cached_object->Flush();
        }
    }
```
This should correctly emulate overlaps on flushing.

bunnei:
All issues should be addressed, final call for reviews

PR #1506, opened 2018-10-16
qt: Add a shader viewer and editor at runtime

ReinUsesLisp:
Allows viewing and editing shaders at runtime. To use it you have to set a Maxwell Breakpoint and it will shaders that are being reported by the cache. You can edit those and when the execution is resumed they will get injected back to the renderer (OpenGL). Since GLSL is the only shading language used, that's what you'll view/edit.

If a shader is invalid (e.g.: it has a syntax error), yuzu will proceed as if that shader would have been emitted by itself (aka crashing in an assert).

imagen

There's a weird issue where the cache just reports a subset of current shaders. I couldn't find it's root but it can be solved hitting the resume button a bunch of times. I know this is horrible, but it works.

There are a bunch of I'd like to add on top of it in the future:
* Execute `glslangValidator` before injecting.
* Minimal code highlight. It's hard to find stuff otherwise.
* A feature to peek in a status bar the result of `nvdisasm -b SM60` or `envydis -m gm107 -V sm60 -i` of the current instruction (it would work looking at the comment above the current line). This would ease finding encoding issues when implementing new instructions.
* A dump button to save shaders in binary form.
* A button to import and decompile a dumped shader.

As a final note, please don't abuse this to fix graphical issues in a hacky manner.

### DIFF
--- a/src/video_core/rasterizer_cache.h
+++ b/src/video_core/rasterizer_cache.h
@@ -129,6 +129,17 @@ protected:
         return ++modified_ticks;
     }
 
+    /// Returns a list of current objects
+    std::set<T> GetObjects() const {
+        std::set<T> objects;
+        for (const auto& pair : interval_cache) {
+            for (const auto& cached_object : pair.second) {
+                objects.insert(cached_object);
+            }
+        }
+        return objects;
+    }
+
 private:
     /// Returns a list of cached objects from the specified memory region, ordered by access time
     std::vector<T> GetSortedObjectsFromRegion(VAddr addr, u64 size) {
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -373,6 +373,14 @@ bool RasterizerOpenGL::AccelerateDrawBatch(bool is_indexed) {
     return true;
 }
 
+std::vector<VideoCore::ShaderInfo> RasterizerOpenGL::GetShaderList() const {
+    return shader_cache.GetShaderInfo();
+}
+
+void RasterizerOpenGL::InjectShader(Tegra::GPUVAddr addr, std::size_t code_size, const u8* code) {
+    shader_cache.InjectShader(addr, code_size, code);
+}
+
 template <typename Map, typename Interval>
 static constexpr auto RangeFromInterval(Map& map, const Interval& interval) {
     return boost::make_iterator_range(map.equal_range(interval));
--- a/src/video_core/renderer_opengl/gl_shader_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_cache.cpp
@@ -10,6 +10,8 @@
 #include "video_core/renderer_opengl/gl_shader_cache.h"
 #include "video_core/renderer_opengl/gl_shader_manager.h"
 #include "video_core/renderer_opengl/utils.h"
+#include "video_core/shader_info.h"
+#include "video_core/utils.h"
 
 namespace OpenGL {
 
@@ -57,7 +59,6 @@ CachedShader::CachedShader(VAddr addr, Maxwell::ShaderProgram program_type)
     : addr{addr}, program_type{program_type}, setup{GetShaderCode(addr)} {
 
     GLShader::ProgramResult program_result;
-    GLenum gl_type{};
 
     switch (program_type) {
     case Maxwell::ShaderProgram::VertexA:
@@ -83,18 +84,10 @@ CachedShader::CachedShader(VAddr addr, Maxwell::ShaderProgram program_type)
         return;
     }
 
+    code = program_result.first;
     entries = program_result.second;
 
-    if (program_type != Maxwell::ShaderProgram::Geometry) {
-        OGLShader shader;
-        shader.Create(program_result.first.c_str(), gl_type);
-        program.Create(true, shader.handle);
-        SetShaderUniformBlockBindings(program.handle);
-        LabelGLObject(GL_PROGRAM, program.handle, addr);
-    } else {
-        // Store shader's code to lazily build it on draw
-        geometry_programs.code = program_result.first;
-    }
+    BuildProgram();
 }
 
 GLuint CachedShader::GetProgramResourceIndex(const GLShader::ConstBufferEntry& buffer) {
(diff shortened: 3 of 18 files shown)


src/video_core/rasterizer_cache.h
```diff
@@ -74,6 +74,17 @@ class RasterizerCache : NonCopyable {
...
 
+    /// Returns a list of current objects
+    auto GetObjects() const {
+        std::set<T> objects;
+        for (const auto& pair : object_cache) {
+            for (const auto& cached_object : pair.second) {
+                objects.insert(cached_object);
+            }
+        }
+        return std::move(objects);
```
lioncash: This should just return without the move, otherwise the move can actually be inhibited here.

src/video_core/rasterizer_cache.h
```diff
@@ -74,6 +74,17 @@ class RasterizerCache : NonCopyable {
         object_cache.subtract({GetInterval(object), ObjectSet{object}});
     }
 
+    /// Returns a list of current objects
+    auto GetObjects() const {
```
lioncash: `std::set<T> GetObjects() const`

src/yuzu/debugger/shader_tools.cpp
```diff
@@ -0,0 +1,236 @@
...
+    emit BreakPointHit(event, data);
+}
+
+void ShaderToolsDialog::OnMaxwellResume() {
+    emit Resumed();
+}
+
+void ShaderToolsDialog::OnBreakPointHit(Tegra::DebugContext::Event event, void* data) {
+    auto& gpu = Core::System::GetInstance().GPU();
+    shaders = std::move(gpu.Maxwell3D().GetShaderList());
```
lioncash: There's no need to move here, this will happen automatically, since the right-hand side is an rvalue.

src/yuzu/debugger/shader_tools.cpp
```diff
@@ -0,0 +1,236 @@
...
+    modified_shaders.resize(shaders.size(), false);
```
lioncash: `false` can be omitted here.

src/yuzu/debugger/shader_tools.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+void ShaderToolsDialog::OnResumed() {
+    SaveActiveCode();
+    SaveScrollPosition();
+    SaveSelectedShader();
+    active_shader = {};
+
+    auto& gpu = Core::System::GetInstance().GPU();
+    for (std::size_t i = 0; i < shaders.size(); ++i) {
+        auto& shader = shaders[i];
```
lioncash: `const auto&`. This can also be moved after the conditional.

```cpp
if (!modified_shaders[i]) {
    continue;
}

auto& shader = shaders[i];
const auto* code = shaders[i].code.c_str();
gpu.Maxwell3D().InjectShader(shader.addr, reinterpret_cast<const u8*>(code));
```

src/yuzu/debugger/shader_tools.h
```diff
@@ -0,0 +1,87 @@
...
+
+#pragma once
+
+#include <map>
+#include <memory>
+#include <vector>
+
+#include <boost/optional.hpp>
+
+#include <QAction>
```
lioncash: Several of these can be forward declarations

src/yuzu/debugger/shader_tools.h
```diff
@@ -0,0 +1,87 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <map>
+#include <memory>
+#include <vector>
+
+#include <boost/optional.hpp>
```
lioncash: We can use regular `<optional>` now.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -363,6 +363,14 @@ bool RasterizerOpenGL::AccelerateDrawBatch(bool is_indexed) {
     return true;
 }
 
+std::vector<VideoCore::ShaderInfo> RasterizerOpenGL::GetShaderList() const {
+    return std::move(shader_cache.GetShaderInfo());
```
lioncash: Ditto wrt the move should be removed.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -134,6 +146,19 @@ GLuint CachedShader::LazyGeometryProgram(OGLProgram& target_program,
...
+VideoCore::ShaderInfo CachedShader::GetShaderInfo() const {
+    VideoCore::ShaderInfo shader_info;
+    shader_info.code = code;
+    shader_info.addr = static_cast<u64>(addr);
+    shader_info.stage = VideoCore::GetShaderStageName(program_type);
+    return shader_info;
+}
+
+void CachedShader::InjectGLSL(const GLchar* inject_code) {
+    code = std::string(inject_code);
```
lioncash: If it's possible to trivially get the length of the data as well, you may want to pass it as a parameter here as well, since the

```cpp
std::string(data, size);
```

constructor doesn't need to call `std::strlen` on the data to determine the size of it. If it's not, don't worry about it.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -149,4 +174,19 @@ Shader ShaderCacheOpenGL::GetStageProgram(Maxwell::ShaderProgram program) {
     return shader;
 }
 
+std::vector<VideoCore::ShaderInfo> ShaderCacheOpenGL::GetShaderInfo() const {
+    std::vector<VideoCore::ShaderInfo> info;
+    for (const auto& shader : GetObjects()) {
+        info.push_back(shader->GetShaderInfo());
+    }
+    return std::move(info);
```
lioncash: The move should be removed here.

bunnei:
Oops - canary conflicts with #1468

src/video_core/utils.h
```diff
@@ -187,4 +187,25 @@ static void LabelGLObject(GLenum identifier, GLuint handle, VAddr addr,
     glObjectLabel(identifier, handle, -1, static_cast<const GLchar*>(object_label.c_str()));
 }
 
+static constexpr const char* GetShaderStageName(
```
DarkLordZach: Since this is used for UI, consider using the enum in GetShaderInfo/GetShaderList and mapping to strings in the frontend. You can then use `QT_TR_NOOP` and `tr` to make the stage names translateable.

ReinUsesLisp: Initially I wanted to do that, but I got a "circular" header dependency. I will retry.

bunnei:
Needs a rebase. Apologies for letting this sit around for more than a week - once it's updated, we should be able to get it merged within the next day or so. Thanks!

bunnei:
Reminder to rebase when you get a chance, thanks!

ReinUsesLisp:
Rebased. DarkLordZach's comment has not been addressed.
I don't know why appveyor doesn't want to check (maybe because I pushed twice in a short period of time?)

bunnei:
@ReinUsesLisp needs a rebase

ReinUsesLisp:
Rebased. Addressed DarkLordZach comment. I'm using D3D names because they are more UI friendly (`Tesselation Eval` is a long name).

ReinUsesLisp:
This is not really needed with Nsight's shader promoting or RenderDoc.

PR #1512, opened 2018-10-16
gl_shader_decompiler: Implement PBK and BRK

ReinUsesLisp:
From my testing this pair behaves just like `SSY` and `SYNC`. From nouveau docs they share the same stack, but calling `SSY` -> `BRK` kills the shader (they are incompatible). This is not emulated because it's very unlikely that a driver/offline compiler would emit an invalid shader.

Renamed the `ssy` stack to `flow` for consistency.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -1095,11 +1095,13 @@ public:
         KIL,
         SSY,
         SYNC,
+        BRK,
         DEPBAR,
         BFE_C,
         BFE_R,
         BFE_IMM,
         BRA,
+        PBK,
         LD_A,
         LD_C,
         ST_A,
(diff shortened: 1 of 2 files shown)


src/video_core/engines/shader_bytecode.h
```diff
@@ -1095,11 +1095,13 @@ class OpCode {
         KIL,
         SSY,
         SYNC,
+        BRK,
```
j-selby:
```suggestion
        BNK,
```

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1156,11 +1157,11 @@ class GLSLGenerator {
      * Emits code to push the input target address to the SSY address stack, incrementing the stack
```
Subv: Please update the comment

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1169,11 +1170,11 @@ class GLSLGenerator {
      * Emits code to pop an address from the SSY address stack, setting the jump address to the
```
Subv: same

PR #1515, opened 2018-10-17
patch_manager: Add support for LayeredFS on DLC RomFS

DarkLordZach:
- Adds support to load LayeredFS patches on DLCs. Since DLCs have different title IDs from the base game, they follow the exact same patch structure except with their title ID. 
- Adds support for dumping DLC RomFS from the game list. If DLCs are found, they will appear in a list with base game for user selection.
dlclfs
*[Pictured: Splatoon 2 + Octo Expansion]*
- Remove duplicates from DLC list if a DLC is installed twice.
- When dumping the base game RomFS, avoid dumping the patched version (with updates/mods).
- Log a DLC layered FS patch to log for debugging.


src/yuzu/main.cpp
```diff
@@ -823,16 +825,24 @@ static bool RomFSRawCopy(QProgressDialog& dialog, const FileSys::VirtualDir& src
...
-
-    const auto failed = [this, &path] {
+    const auto failed = [this] {
         QMessageBox::warning(this, tr("RomFS Extraction Failed!"),
                              tr("There was an error copying the RomFS files or the user "
                                 "cancelled the operation."));
-        vfs->DeleteDirectory(path);
     };
 
+    const auto& installed = Service::FileSystem::GetUnionContents();
```
lioncash: This function return by value, so it shouldn't be taking this by reference. This code can also go in its own function which returns the std::vector.

Is this also intentionally called before reading the RomFS? If not, the call to said function can probably be moved down closer to the other code where it's used to avoid doing unnecessary work if bailing out is necessary in the two conditionals below.

src/yuzu/main.cpp
```diff
@@ -845,10 +855,39 @@ void GMainWindow::OnGameListDumpRomFS(u64 program_id, const std::string& game_pa
...
+    auto romfs_title_id = program_id;
+    std::vector<u64> romfs_tids;
+    romfs_tids.push_back(program_id);
+    for (const auto& entry : dlc_match)
+        romfs_tids.push_back(entry.title_id);
+
+    if (romfs_tids.size() > 1) {
+        QStringList list{"Base"};
+        for (std::size_t i = 1; i < romfs_tids.size(); ++i)
+            list.push_back(QString::fromStdString(fmt::format("DLC {}", romfs_tids[i] & 0x7FF)));
```
lioncash: This can just be:

```cpp
list.push_back(QStringLiteral("DLC %1").arg(romfs_tids[i] & 0x7FF));
```

src/yuzu/main.cpp
```diff
@@ -845,10 +855,39 @@ void GMainWindow::OnGameListDumpRomFS(u64 program_id, const std::string& game_pa
         return;
     }
 
-    const auto romfs =
-        loader->IsRomFSUpdatable()
-            ? FileSys::PatchManager(program_id).PatchRomFS(file, loader->ReadRomFSIVFCOffset())
-            : file;
+    auto romfs_title_id = program_id;
```
lioncash: This segment of code up to line 879 should go in its own function, returning `std::optional` that would contain the title ID as a result and would be used like so:

```cpp
const auto romfs_title_id = GetRomFSTitleID(...);
if (!romfs_title_id) {
    failed();
    return;
}
```

src/core/file_sys/registered_cache.h
```diff
@@ -50,6 +50,9 @@ constexpr u64 GetUpdateTitleID(u64 base_title_id) {
 // boost flat_map requires operator< for O(log(n)) lookups.
 bool operator<(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs);
 
+// std unique requires operator== to identify duplicates.
+bool operator==(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs);
```
lioncash: `operator!=` should also be provided for symmetry.

PR #1517, opened 2018-10-18
GPU/DMA: Flush the source region and invalidate the destination region when doing a DMA transfer.

bunnei:
Implements the rest of #1339, using the recently merged flushing code from #1497. Some of this is behind the `use_accurate_gpu_emulation` setting (as otherwise, we avoid flushing in many cases). This is mostly ported over as-is, and needs some testing - I'm not sure what games actually use this
In particular, let me know if there are things that do not work with the `use_accurate_gpu_emulation` setting turned off, and I will try to debug/fix. Also, please let me know if there are any major perf regressions (again, with the setting turned off).

cc @Hexagon12 to assist with testing.
cc @FernandoS27 to improve the swizzle code here, this is based on our old swizzling.
cc @Subv, purely FYI.

### DIFF
--- a/src/video_core/engines/maxwell_dma.cpp
+++ b/src/video_core/engines/maxwell_dma.cpp
@@ -4,12 +4,14 @@
 
 #include "core/memory.h"
 #include "video_core/engines/maxwell_dma.h"
+#include "video_core/rasterizer_interface.h"
 #include "video_core/textures/decoders.h"
 
 namespace Tegra {
 namespace Engines {
 
-MaxwellDMA::MaxwellDMA(MemoryManager& memory_manager) : memory_manager(memory_manager) {}
+MaxwellDMA::MaxwellDMA(VideoCore::RasterizerInterface& rasterizer, MemoryManager& memory_manager)
+    : memory_manager(memory_manager), rasterizer{rasterizer} {}
 
 void MaxwellDMA::WriteReg(u32 method, u32 value) {
     ASSERT_MSG(method < Regs::NUM_REGS,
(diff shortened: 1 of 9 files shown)


src/video_core/textures/decoders.cpp
```diff
@@ -237,6 +237,51 @@ std::vector<u8> UnswizzleTexture(VAddr address, u32 tile_size, u32 bytes_per_pix
...
+                      (x * bytes_per_pixel / 64) * 512 * block_height +
+                      (y % (8 * block_height) / 8) * 512;
+    x *= bytes_per_pixel;
+    u32 address = GOB_address + ((x % 64) / 32) * 256 + ((y % 8) / 2) * 64 + ((x % 32) / 16) * 32 +
+                  (y % 2) * 16 + (x % 16);
+
+    return address;
+}
+
+void SwizzleSubrect(u32 subrect_width, u32 subrect_height, u32 source_pitch, u32 swizzled_width,
```
bunnei: These need to be updated for 3d swizzle (used by Octopath, for example)

src/video_core/engines/maxwell_dma.cpp
```diff
@@ -44,38 +46,79 @@ void MaxwellDMA::HandleCopy() {
...
+            const VAddr source_line = source_cpu + line * regs.src_pitch;
+            const VAddr dest_line = dest_cpu + line * regs.dst_pitch;
+            Memory::CopyBlock(dest_line, source_line, regs.x_count);
+        }
         return;
     }
 
     ASSERT(regs.exec.enable_2d == 1);
+
+    size_t copy_size = regs.x_count * regs.y_count;
```
ogniK5377: std::size_t

PR #1524, opened 2018-10-18
rasterizer: Fix Layered Textures Loading and Cubemaps

FernandoS27:
This pull request fixes several issues concerning texture arrays and cubemaps. Before we were reading this types of textures naively without calculating the spacing between them (the mipmaps). With this PR, we correct this issue and add the ability to interpret cubemaps correctly.

As many issues this fixes, it may also bring to light new issues as this textures weren't working correctly and things not shown before will now be shown.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -78,6 +78,29 @@ void SurfaceParams::InitCacheParameters(Tegra::GPUVAddr gpu_addr_) {
     }
 }
 
+std::size_t SurfaceParams::InnerMemorySize(bool layer_only) const {
+    const u32 compression_factor{GetCompressionFactor(pixel_format)};
+    const u32 bytes_per_pixel{GetBytesPerPixel(pixel_format)};
+    u32 m_depth = (layer_only ? 1U : depth);
+    u32 m_width = std::max(1U, width / compression_factor);
+    u32 m_height = std::max(1U, height / compression_factor);
+    std::size_t size = Tegra::Texture::CalculateSize(is_tiled, bytes_per_pixel, m_width, m_height,
+                                                     m_depth, block_height, block_depth);
+    u32 m_block_height = block_height;
+    u32 m_block_depth = block_depth;
+    std::size_t block_size_bytes = 512 * block_height * block_depth; // 512 is GOB size
+    for (u32 i = 1; i < max_mip_level; i++) {
+        m_width = std::max(1U, m_width / 2);
+        m_height = std::max(1U, m_height / 2);
+        m_depth = std::max(1U, m_depth / 2);
+        m_block_height = std::max(1U, m_block_height / 2);
+        m_block_depth = std::max(1U, m_block_depth / 2);
+        size += Tegra::Texture::CalculateSize(is_tiled, bytes_per_pixel, m_width, m_height, m_depth,
+                                              m_block_height, m_block_depth);
+    }
+    return is_tiled ? Common::AlignUp(size, block_size_bytes) : size;
+}
+
 /*static*/ SurfaceParams SurfaceParams::CreateForTexture(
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -890,12 +893,25 @@ void CachedSurface::LoadGLBuffer() {
         if (params.target == SurfaceParams::SurfaceTarget::Texture2D) {
             // TODO(Blinkhawk): Eliminate this condition once all texture types are implemented.
             depth = 1U;
-            block_depth = 1U;
         }
 
-        morton_to_gl_fns[static_cast<std::size_t>(params.pixel_format)](
-            params.width, params.block_height, params.height, block_depth, depth, gl_buffer.data(),
-            gl_buffer.size(), params.addr);
+        if (params.is_layered) {
```
lioncash: This and the below equivalent in FlushGLBuffer can be generified like so:

```cpp
// This would be placed on line 363 and the two conversion arrays would be changed to use this alias.
using GLConversionArray = std::array<void (*)(u32, u32, u32, u32, u32, u8*, std::size_t, VAddr), SurfaceParams::MaxPixelFormat>;

// Give this an appropriate name
void DoThing(const GLConversionArray& functions, const SurfaceParams& params) {
    if (params.is_layered) {
        u64 offset = 0;
        u64 offset_gl = 0;
        u64 layer_size = params.LayerMemorySize();
        u64 gl_size = params.LayerSizeGL();
        for (u32 i = 0; i < depth; i++) {
            functions[static_cast<std::size_t>(params.pixel_format)](
                params.width, params.block_height, params.height, block_depth, 1,
                gl_buffer.data() + offset_gl, gl_size, params.addr + offset);
            offset += layer_size;
            offset_gl += gl_size;
        }
    } else {
        functions[static_cast<std::size_t>(params.pixel_format)](
            params.width, params.block_height, params.height, block_depth, depth,
            gl_buffer.data(), gl_buffer.size(), params.addr);
    }
}
```

Then this can just be called like:

```cpp
DoThing(morton_to_gl_fns, params);
```
here

and called like:

```cpp
DoThing(gl_to_morton_fns, params);
```

in FlushGLBuffer().

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1221,52 +1251,18 @@ Surface RasterizerCacheOpenGL::RecreateSurface(const Surface& old_surface,
...
-    }
     default:
         LOG_CRITICAL(Render_OpenGL, "Unimplemented surface target={}",
                      static_cast<u32>(new_params.target));
         UNREACHABLE();
     }
 
     return new_surface;
-}
+} // namespace OpenGL
```
lioncash: This likely wasn't intended.

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -797,6 +834,30 @@ struct SurfaceParams {
         u32 layer_stride;
         u32 base_layer;
     } rt;
+
+private:
+    std::size_t InnerMemorySize(bool layer_only = false) const {
```
lioncash: This can be implemented within the cpp file, so you don't need to include `<algorithm>`

FernandoS27: could but I feel the cpp file is overcrowded with other unrelated things to it.

lioncash: `InitCacheParameters` is implemented in the cpp file, this can just be put next to it.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
FernandoS27: that's clang format being clang format.

FernandoS27:
changes made and clang formatted.

PR #1525, opened 2018-10-19
Home button blocking stub

ogniK5377:
Used by arms


lioncash:
Your commit message should explain what it's for/what needs it.

PR #1527, opened 2018-10-19
Assert Control Flow Instructions using Control Codes

FernandoS27:
This PR just adds asserts to Flow Instructions, making sure they don't use Control Codes. (if they do, hell)


FernandoS27:
let's keep it simple as I want to add more asserts in other PRs.

bunnei:
Needs a rebase

PR #1531, opened 2018-10-20
Added auto controller switching to supported controllers and single joycon button rotation

ogniK5377:
This is a subset of the better-hid-2 changes, this fixes input in various games which don't support dual joycons. This pr will search for the next best controller which is supported by the current game

### DIFF
--- a/src/core/hle/service/hid/controllers/npad.cpp
+++ b/src/core/hle/service/hid/controllers/npad.cpp
@@ -26,7 +26,11 @@ constexpr s32 HID_JOYSTICK_MAX = 0x7fff;
 constexpr s32 HID_JOYSTICK_MIN = -0x7fff;
 constexpr std::size_t NPAD_OFFSET = 0x9A00;
 constexpr u32 BATTERY_FULL = 2;
-
+constexpr u32 NPAD_HANDHELD = 32;
+constexpr u32 NPAD_UNKNOWN = 16; // TODO(ogniK): What is this?
+constexpr u32 MAX_NPAD_ID = 7;
+constexpr Controller_NPad::NPadControllerType PREFERRED_CONTROLLER =
+    Controller_NPad::NPadControllerType::JoyDual;
 constexpr std::array<u32, 10> npad_id_list{
     0, 1, 2, 3, 4, 5, 6, 7, 32, 16,
 };
@@ -121,7 +125,7 @@ void Controller_NPad::OnInit() {
         supported_npad_id_types.resize(npad_id_list.size());
         std::memcpy(supported_npad_id_types.data(), npad_id_list.data(),
                     npad_id_list.size() * sizeof(u32));
-        AddNewController(NPadControllerType::JoyDual);
+        AddNewController(PREFERRED_CONTROLLER);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/core/hle/service/hid/controllers/npad.h
```diff
@@ -284,5 +284,7 @@ class Controller_NPad final : public ControllerBase {
     bool can_controllers_vibrate{true};
 
     void InitNewlyAddedControler(std::size_t controller_idx);
+    bool IsControllerSupported(NPadControllerType controller);
```
lioncash: This can be a const member function.

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -426,4 +483,127 @@ Controller_NPad::LedPattern Controller_NPad::GetLedPattern(u32 npad_id) {
...
+        if (!is_docked) {
+            priority_list.push_back(NPadControllerType::Handheld);
+        }
+        priority_list.push_back(NPadControllerType::ProController);
+        priority_list.push_back(NPadControllerType::JoyLeft);
+        priority_list.push_back(NPadControllerType::JoyRight);
+        priority_list.push_back(NPadControllerType::JoyDual);
+    }
+
+    for (const auto controller_type : priority_list) {
```
lioncash:
```cpp
const auto iter = std::find_if(priority_list.begin(), priority_list.end(),
                               [this](auto type) { return IsControllerSupported(type); });
if (iter == priority_list.end()) {
    UNIMPLEMENTED_MSG("Could not find supported controller!");
    return priority;
}

return *iter;
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -426,4 +483,127 @@ Controller_NPad::LedPattern Controller_NPad::GetLedPattern(u32 npad_id) {
...
+        return priority;
+    }
+    const auto is_docked = Settings::values.use_docked_mode;
+    if (is_docked && priority == NPadControllerType::Handheld) {
+        priority = NPadControllerType::JoyDual;
+        if (IsControllerSupported(priority)) {
+            return priority;
+        }
+    }
+    std::vector<NPadControllerType> priority_list{};
```
lioncash: braces aren't necessary here.

src/core/hle/service/hid/controllers/npad.h
```diff
@@ -284,5 +284,7 @@ class Controller_NPad final : public ControllerBase {
...
+    NPadControllerType DecideBestController(NPadControllerType priority);
```
lioncash: This can be a const member function.

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -426,4 +483,127 @@ Controller_NPad::LedPattern Controller_NPad::GetLedPattern(u32 npad_id) {
 void Controller_NPad::SetVibrationEnabled(bool can_vibrate) {
     can_controllers_vibrate = can_vibrate;
 }
+
+bool Controller_NPad::IsControllerSupported(NPadControllerType controller) {
+    const bool support_handheld =
+        std::find(supported_npad_id_types.begin(), supported_npad_id_types.end(), 32) !=
```
lioncash: Magic constant

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -426,4 +483,127 @@ Controller_NPad::LedPattern Controller_NPad::GetLedPattern(u32 npad_id) {
...
+        }
+        // Handheld should not be supported in docked mode
+        if (Settings::values.use_docked_mode) {
+            return false;
+        }
+
+        return true;
+    }
+    if (std::any_of(supported_npad_id_types.begin(), supported_npad_id_types.end(),
+                    [](u32 npad_id) { return npad_id < 8; })) {
```
lioncash: Ditto

PR #1533, opened 2018-10-20
Implemented Shader Local Memory

FernandoS27:
This should fix many visual issues in many games that spilled registers to local memory.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -208,6 +208,16 @@ enum class UniformType : u64 {
     Double = 5,
 };
 
+enum class StoreType : u64 {
+    Unsigned8 = 0,
+    Signed8 = 1,
+    Unsigned16 = 2,
+    Signed16 = 3,
+    Bytes32 = 4,
+    Bytes64 = 5,
+    Bytes128 = 6,
+};
+
 enum class IMinMaxExchange : u64 {
     None = 0,
     XLo = 1,
@@ -748,6 +758,18 @@ union Instruction {
     } ld_c;
 
     union {
+        BitField<48, 3, StoreType> type;
+    } ldst_sl;
+
+    union {
+        BitField<44, 2, u64> unknown;
+    } ld_l;
+
+    union {
+        BitField<44, 2, u64> unknown;
+    } st_l;
+
+    union {
         BitField<0, 3, u64> pred0;
         BitField<3, 3, u64> pred3;
         BitField<7, 1, u64> abs_a;
@@ -1209,6 +1231,7 @@ union Instruction {
     BitField<61, 1, u64> is_b_imm;
     BitField<60, 1, u64> is_b_gpr;
     BitField<59, 1, u64> is_c_gpr;
+    BitField<20, 24, s64> smem_imm;
 
     Attribute attribute;
     Sampler sampler;
@@ -1232,8 +1255,12 @@ public:
         BRA,
         PBK,
         LD_A,
+        LD_L,
+        LD_S,
         LD_C,
         ST_A,
+        ST_L,
+        ST_S,
         LDG, // Load from global memory
         STG, // Store in global memory
         TEX,
@@ -1490,8 +1517,12 @@ private:
             INST("111000110100---", Id::BRK, Type::Flow, "BRK"),
             INST("1111000011110---", Id::DEPBAR, Type::Synch, "DEPBAR"),
             INST("1110111111011---", Id::LD_A, Type::Memory, "LD_A"),
+            INST("1110111101001---", Id::LD_S, Type::Memory, "LD_S"),
+            INST("1110111101000---", Id::LD_L, Type::Memory, "LD_L"),
             INST("1110111110010---", Id::LD_C, Type::Memory, "LD_C"),
             INST("1110111111110---", Id::ST_A, Type::Memory, "ST_A"),
+            INST("1110111101011---", Id::ST_S, Type::Memory, "ST_S"),
+            INST("1110111101010---", Id::ST_L, Type::Memory, "ST_L"),
             INST("1110111011010---", Id::LDG, Type::Memory, "LDG"),
             INST("1110111011011---", Id::STG, Type::Memory, "STG"),
             INST("110000----111---", Id::TEX, Type::Memory, "TEX"),
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -278,8 +278,7 @@ public:
                         const Maxwell3D::Regs::ShaderStage& stage, const std::string& suffix,
                         const Tegra::Shader::Header& header)
         : shader{shader}, declarations{declarations}, stage{stage}, suffix{suffix}, header{header},
-          fixed_pipeline_output_attributes_used{} {
-        local_memory_size = 0;
+          fixed_pipeline_output_attributes_used{}, local_memory_size{0} {
         BuildRegisterList();
         BuildInputList();
     }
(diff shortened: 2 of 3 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -391,6 +392,24 @@ class GLSLRegisterManager {
         shader.AddLine(dest + " = " + src + ';');
     }
 
+    std::string GetLocalMemoryAsFloat(const std::string index) {
```
lioncash: The parameter can be a std::string_view. This function also doesn't affect class state and can be static.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -391,6 +392,24 @@ class GLSLRegisterManager {
...
+        return "lmem[" + index + "]";
```
lioncash: ']';

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -391,6 +392,24 @@ class GLSLRegisterManager {
...
+    }
+
+    std::string GetLocalMemoryAsInteger(const std::string index, bool is_signed = false) {
```
lioncash: ditto about std::string_view. This also doesn't affect class state either and can be made static.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -391,6 +392,24 @@ class GLSLRegisterManager {
...
+        const std::string func{is_signed ? "floatToIntBits" : "floatBitsToUint"};
+        return func + "(lmem[" + index + "])";
+    }
+
+    void SetLocalMemoryAsFloat(const std::string index, const std::string value) {
```
lioncash: Ditto about std::string_view.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -391,6 +392,24 @@ class GLSLRegisterManager {
...
+        shader.AddLine("lmem[" + index + "] = " + value + ';');
+    }
+
+    void SetLocalMemoryAsInteger(const std::string index, const std::string value, bool is_signed = false) {
```
lioncash: Ditto about std::string_view

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -391,6 +392,24 @@ class GLSLRegisterManager {
...
+        const std::string func{is_signed ? "intBitsToFloat" : "uintBitsToFloat"};
```
lioncash: This can be a `const char*`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -537,6 +561,14 @@ class GLSLRegisterManager {
         declarations.AddNewLine();
     }
 
+    /// Generates declarations for local memory.
+    void GenerateLocalMemory() {
+        if (local_memory_size > 0) {
```
```cpp
if (local_memory_size <= 0) {
    return;
}
```

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2008,6 +2044,39 @@ class GLSLGenerator {
                 shader.AddLine("}");
                 break;
             }
+            case OpCode::Id::LD_L: {
+                // Add an extra scope and declare the index register inside to prevent
+                // overwriting it in case it is used as an output of the LD instruction.
+                shader.AddLine("{");
```
lioncash: '{'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2008,6 +2044,39 @@ class GLSLGenerator {
...
+                    regs.SetRegisterToFloat(instr.gpr0, 0, op_a, 1, 1);
+                    break;
+                default:
+                    LOG_CRITICAL(HW_GPU, "LD_L Unhandled type: {}",
+                                 static_cast<unsigned>(instr.ldst_sl.type.Value()));
+                    UNREACHABLE();
+                }
+
+                --shader.scope;
+                shader.AddLine("}");
```
lioncash: `'}'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2036,6 +2105,37 @@ class GLSLGenerator {
 
                 break;
             }
+            case OpCode::Id::ST_L: {
+
+                // Add an extra scope and declare the index register inside to prevent
+                // overwriting it in case it is used as an output of the LD instruction.
+                shader.AddLine("{");
```
lioncash: '{'

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2036,6 +2105,37 @@ class GLSLGenerator {
...
+                    regs.SetLocalMemoryAsFloat("index", regs.GetRegisterAsFloat(instr.gpr0));
+                    break;
+                default:
+                    LOG_CRITICAL(HW_GPU, "ST_L Unhandled type: {}",
+                                 static_cast<unsigned>(instr.ldst_sl.type.Value()));
+                    UNREACHABLE();
+                }
+
+                --shader.scope;
+                shader.AddLine("}");
```
lioncash: `'}'`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -278,6 +278,7 @@ class GLSLRegisterManager {
                         const Maxwell3D::Regs::ShaderStage& stage, const std::string& suffix,
                         const Tegra::Shader::Header& header)
         : shader{shader}, declarations{declarations}, stage{stage}, suffix{suffix}, header{header} {
+        local_memory_size = 0;
```
lioncash: This can be assigned to the variable directly in the class body.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
FernandoS27: tried it but string_view does not support + operator.

lioncash: Yes, you append it to an existing string:

```cpp
std::string GetLocalMemoryAsFloat(std::string_view index) {
        return std::string("lmem[").append(index).append("]");
}
```

or:

```cpp
std::string GetLocalMemoryAsFloat(std::string_view index) {
        return fmt::format("lmem[{}]", index);
}
```

The reason for that is std::string_view doesn't allocate memory, so they made the case where allocation would occur non-implicit, so you have to be explicit when you want to allocate memory related to it.

FernandoS27: well I went with a reference instead to avoid copying.  If that option is better, it's better to apply it in a style pr to ALL the decompiler (we use std::string everywhere).

lioncash: std::string_view avoids copying in cases like:

```cpp
GetLocalMemoryAsFloat("SomeStringLiteral");
```

Here, this is a const char[17], however if the parameter type is a std::string, this will potentially heal-allocate and construct a std::string. `const std::string&` only avoids a copy if the type being passed is already a std::string type. `std::string_view` is what is intended to be used as a parameter if the data being passed is a string of any form and the data is only being read from (and nothing in the function needs to own the data.

This is why the `AddLine` function for our ShaderWriter class uses a std::string_view. You can write out either form of string variant without potentially making an unnecessary copy implicitly.

Swapping out `std::string` with `std::string_view` is not recommended when you actually need to make changes to the data (or actually need a copy of the `std::string` to do so), because it would end up modifying the pointed-to data. `std::string_view` is nothing more than a pointer and size to data it doesn't own.

src/video_core/engines/shader_bytecode.h
```diff
@@ -1080,6 +1102,7 @@ union Instruction {
     BitField<61, 1, u64> is_b_imm;
     BitField<60, 1, u64> is_b_gpr;
     BitField<59, 1, u64> is_c_gpr;
+    BitField<20, 24, s64> smem_imn;
```
Subv: imm

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -527,6 +548,10 @@ class GLSLRegisterManager {
         return entry.GetName();
     }
 
+    void SetLocalMemory(u64 lmem) {
```
Subv: SetLocalMemorySize. Why is this not part of the constructor?

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -847,6 +882,9 @@ class GLSLGenerator {
         : subroutines(subroutines), program_code(program_code), main_offset(main_offset),
           stage(stage), suffix(suffix) {
         std::memcpy(&header, program_code.data(), sizeof(Tegra::Shader::Header));
+        local_memory_size = (header.common2.shader_local_memory_high_size << 24) |
```
Subv: this can be a function inside the shader header class

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2036,6 +2107,38 @@ class GLSLGenerator {
 
                 break;
             }
+            case OpCode::Id::ST_L: {
+
```
Subv: extra line

FernandoS27:
changes made and clang formatted.

PR #1536, opened 2018-10-20
Implemented Shader dumper for RE

FernandoS27:
Here's a simple shader dumper to use with nvdisasm. It will only dump shaders marked as faulty. It should be used for debugging purposes mostly. In order to use it do the next:

CUDA Toolkit 8.0+ must be installed. https://developer.nvidia.com/cuda-downloads

1. fetch this pr into a separate branch and build it.
2. run yuzu with the game with faulty shaders. 
3. All shaders will be next to Yuzu's executable.
4. dissassemble a shader by using "nvdisasm -b SM52 <shader_name> > myasm.txt"
5. enjoy and happy Reverse Engineering!

Edit to dump specific shaders on certain assert/error, add "faulty =true;" next to it in shader_decompiler.cpp

This pull request has been recreated from another one since it got corrupted.

### DIFF
--- a/src/common/file_util.h
+++ b/src/common/file_util.h
@@ -260,6 +260,14 @@ public:
         return WriteArray(str.c_str(), str.length());
     }
 
+    std::size_t WriteCString(const char* str) {
+        if (!IsOpen()) {
+            return std::numeric_limits<std::size_t>::max();
+        }
+        std::fputs(str, m_file);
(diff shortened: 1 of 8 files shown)


FernandoS27:
Rebased, you can now dump specific shaders by adding them to the dump set in shader_dumper.cpp.

PR #1545, opened 2018-10-20
psm: Add psm service and stub commands 0 and 1

DarkLordZach:
Needed by [LovePotion Homebrew](https://github.com/TurtleP/LovePotion) to get the battery percentage and charger type. Stubbed to 100% and connected to official dock.

Filled in names of PSM functions from switchbrew.

Fixes issues outlined in Ryujinx/Ryujinx#465 -- If the original reporter of that issue (@NoahRoseLedesma) could test this, that would be wonderful.


src/core/hle/service/ptm/psm.h
```diff
@@ -0,0 +1,26 @@
...
+#pragma once
+#include "core/hle/service/service.h"
+
+namespace Service::SM {
+class ServiceManager;
+}
+
+namespace Service::PSM {
+
+class PSM final : public ServiceFramework<PSM> {
```
lioncash: The class can go in the cpp file, since nothing outside of it needs to see it. This'll let you remove the include from the header too.

PR #1551, opened 2018-10-23
Added break types to svcBreak

ogniK5377:
There seems to be more such as type 1, and 2. Unsure what these currently are but when a game hits them we can investigate and add the rest. Info 1 and Info 2 seems to also provide extra information based on the break. Info 1 usually looks like an address and info 2 looks like a size, however, this also can be zeroed out so the info can be different each type? More research needed but this is a start on better-debugging information.

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -374,8 +374,9 @@ static ResultCode ArbitrateUnlock(VAddr mutex_addr) {
     return Mutex::Release(mutex_addr);
 }
 
-enum BreakType : u32 {
+enum class BreakType : u32 {
     Panic = 0,
+    AssertionFailed = 1,
     PreNROLoad = 3,
     PostNROLoad = 4,
     PreNROUnload = 5,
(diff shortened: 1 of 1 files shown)


src/core/hle/kernel/svc.cpp
```diff
@@ -374,22 +374,60 @@ static ResultCode ArbitrateUnlock(VAddr mutex_addr) {
     return Mutex::Release(mutex_addr);
 }
 
+enum BreakType : u32 {
```
lioncash: enum class

src/core/hle/kernel/svc.cpp
```diff
@@ -374,22 +374,60 @@ static ResultCode ArbitrateUnlock(VAddr mutex_addr) {
...
+    Panic = 0,
```
lioncash: `1` is likely for asserts rather than an outright panic.

src/core/hle/kernel/svc.cpp
```diff
@@ -374,22 +374,60 @@ static ResultCode ArbitrateUnlock(VAddr mutex_addr) {
...
     BreakReason break_reason{reason};
-    if (break_reason.signal_debugger) {
+
+    switch (break_reason.break_type) {
+    case BreakType::Panic:
+        LOG_ERROR(Debug_Emulated, "Signalling debugger, PANIC! info1=0x{:016X}, info2=0x{:016X}",
+                  info1, info2);
+        break;
+    case BreakType::PreNROLoad:
+        LOG_ERROR(Debug_Emulated,
```
lioncash: These probably shouldn't be errors, if they were just intended to signal the debugger.

PR #1556, opened 2018-10-23
svc: Improve SleepThread for yield types.

bunnei:
Some improvements to svcSleepThread. This should be right for yield type 0, and is a workaround for yield types -1 and -2 (behind an assert).
The motivation for this change is otherwise svcSleepThread silent is really broken. We should implement proper yield types, but this needs a lot of work for our current scheduler. Until then, it would be helpful to know which games are hitting this.

Partially fixes Super Mario Party.

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -770,12 +770,24 @@ static void ExitThread() {
 
 /// Sleep the current thread
 static void SleepThread(s64 nanoseconds) {
-    LOG_TRACE(Kernel_SVC, "called nanoseconds={}", nanoseconds);
+    LOG_DEBUG(Kernel_SVC, "called nanoseconds={}", nanoseconds);
 
-    // Don't attempt to yield execution if there are no available threads to run,
-    // this way we avoid a useless reschedule to the idle thread.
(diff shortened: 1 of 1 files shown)


src/core/hle/kernel/svc.cpp
```diff
@@ -770,20 +770,35 @@ static void ExitThread() {
...
-    LOG_TRACE(Kernel_SVC, "called nanoseconds={}", nanoseconds);
-
-    // Don't attempt to yield execution if there are no available threads to run,
-    // this way we avoid a useless reschedule to the idle thread.
-    if (nanoseconds == 0 && !Core::System::GetInstance().CurrentScheduler().HaveReadyThreads())
+    LOG_DEBUG(Kernel_SVC, "called nanoseconds={}", nanoseconds);
+
+    if (nanoseconds == 0) {
+        // Yield type 0 is just a yield without load balancing
+        Core::System::GetInstance().CpuCore(0).PrepareReschedule();
```
lioncash: I think this rescheduling of all cores should go into a `PrepareRescheduleAll` function or something, given this is done in other parts of the codebase, to reduce the need to copy 4 lines verbatim.

bunnei:
Superseded by 1732

bunnei:
Reopening and retagging as Canary until #1732 is ready, as I still use this for testing certain games

bunnei:
Re-superceded by #1732

PR #1471, opened 2018-10-11
[WIP] Implemented the LoadNrr and LoadNro functions from the ldr:ro service.

Subv:
The implementation here was based on Atmosphere's implementation of the ldr:ro service. The error condition checks and error codes are still missing, and there's no unloading code at all.

PR #1557, opened 2018-10-23
Implement the LoadNro functions from the ldr:ro service.

bunnei:
Implements LoadNro function using the existing loader code that we have. This is a really simple and somewhat incomplete implementation, but is enough for basic use cases. With this change and #1556, Super Mario Party now boots.

image

Partially supersedes #1471

### DIFF
--- a/src/core/hle/kernel/vm_manager.cpp
+++ b/src/core/hle/kernel/vm_manager.cpp
@@ -143,6 +143,26 @@ ResultVal<VMManager::VMAHandle> VMManager::MapBackingMemory(VAddr target, u8* me
     return MakeResult<VMAHandle>(MergeAdjacent(vma_handle));
 }
 
+ResultVal<VAddr> VMManager::FindFreeRegion(u64 size) const {
+    // Find the first Free VMA.
+    const VAddr base = GetASLRRegionBaseAddress();
+    const VMAHandle vma_handle = std::find_if(vma_map.begin(), vma_map.end(), [&](const auto& vma) {
+        if (vma.second.type != VMAType::Free)
+            return false;
+
+        const VAddr vma_end = vma.second.base + vma.second.size;
+        return vma_end > base && vma_end >= base + size;
+    });
+
+    if (vma_handle == vma_map.end()) {
+        // TODO(Subv): Find the correct error code here.
+        return ResultCode(-1);
+    }
+
+    const VAddr target = std::max(base, vma_handle->second.base);
+    return MakeResult<VAddr>(target);
+}
+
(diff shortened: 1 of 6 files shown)


src/core/loader/nro.cpp
```diff
@@ -127,18 +127,23 @@ static constexpr u32 PageAlignSize(u32 size) {
...
+    NroHeader nro_header{};
+    std::memcpy(&nro_header, data.data(), sizeof(NroHeader));
     if (nro_header.magic != Common::MakeMagic('N', 'R', 'O', '0')) {
         return {};
     }
 
     // Build program image
-    std::vector<u8> program_image = file.ReadBytes(PageAlignSize(nro_header.file_size));
+    std::vector<u8> program_image(PageAlignSize(nro_header.file_size));
+    std::memcpy(program_image.data(), data.data(), program_image.size());
```
lioncash:
```cpp
const size_t page_aligned_size = PageAlignSize(nro_header.file_size);
std::vector<u8> program_image(data.begin(), data.begin() + page_aligned_size);
```

will copy the data into the vector without the need to zero out the backing memory for the vector.

src/core/loader/nro.cpp
```diff
@@ -127,18 +127,23 @@ static constexpr u32 PageAlignSize(u32 size) {
...
     if (program_image.size() != PageAlignSize(nro_header.file_size)) {
```
lioncash: This check should probably be changed to be in terms of the data vector parameter and moved above the vector construction above, otherwise this is always false.

src/core/hle/kernel/vm_manager.cpp
```diff
@@ -143,6 +143,26 @@ ResultVal<VMManager::VMAHandle> VMManager::MapBackingMemory(VAddr target, u8* me
     return MakeResult<VMAHandle>(MergeAdjacent(vma_handle));
 }
 
+ResultVal<VAddr> VMManager::FindFreeRegion(u64 size) const {
+    // Find the first Free VMA.
+    const VAddr base = GetAddressSpaceBaseAddress();
```
lioncash: This should likely be using `GetASLRRegionBaseAddress()`, otherwise this is always zero and doesn't take into account the initial guard memory put at the beginning of the address space.

src/core/loader/nro.cpp
bunnei: Addressed

src/core/loader/nro.cpp
bunnei: Addressed

PR #1579, opened 2018-10-24
service/usb: Update service function tables

lioncash:
Updated based off the information provided by Hexkyz on Switchbrew.

### DIFF
--- a/src/core/hle/service/usb/usb.cpp
+++ b/src/core/hle/service/usb/usb.cpp
@@ -67,15 +67,15 @@ public:
     explicit IClientEpSession() : ServiceFramework{"IClientEpSession"} {
         // clang-format off
         static const FunctionInfo functions[] = {
-            {0, nullptr, "Unknown1"},
-            {1, nullptr, "Unknown2"},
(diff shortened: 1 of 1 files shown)


src/core/hle/service/usb/usb.cpp
```diff
@@ -67,15 +67,15 @@ class IClientEpSession final : public ServiceFramework<IClientEpSession> {
     explicit IClientEpSession() : ServiceFramework{"IClientEpSession"} {
         // clang-format off
         static const FunctionInfo functions[] = {
-            {0, nullptr, "Unknown1"},
-            {1, nullptr, "Unknown2"},
-            {2, nullptr, "Unknown3"},
-            {3, nullptr, "Unknown4"},
+            {0, nullptr, "Open"},
+            {1, nullptr, "Close"},
+            {2, nullptr, "Unknown1"},
```
ogniK5377: It might be easier to track if it was `UnknownCmdNum`, eg `Unknown2`

FearlessTobi: tbf we've done it like it's done in the PR, for years.

PR #1581, opened 2018-10-25
Update MACOSX_DEPLOYMENT_TARGET to 10.14

FreddyFunk:
Enables the usage of further C++ std functions introduced in macOS 10.14.

~~This PR is also required for #1578 for `std::optional::value()`.~~

### DIFF
--- a/.travis/macos/build.sh
+++ b/.travis/macos/build.sh
@@ -2,7 +2,7 @@
 
 set -o pipefail
 
-export MACOSX_DEPLOYMENT_TARGET=10.13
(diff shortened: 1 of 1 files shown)


B3n30:
I suggest reverting that PR since it doesn't what the description says. macOS 10.14 and 10.13 use the same compiler and same libs and thus there is 0 benefit in changing the deployment target. Also travis ci has no version to build with 10.14 atm, so even if there will be a newer lib/compiler available, the CI will still fail.

PR #1582, opened 2018-10-25
W.I.P Implement sRGB support

Tinob:
first iteration to allow testing, feedback is welcome. I'm interested in knowing specially the values of the registers in different games and if rendering is correct.


src/video_core/engines/maxwell_3d.h
```diff
@@ -723,7 +723,11 @@ class Maxwell3D final {
                 StencilOp stencil_back_op_zpass;
                 ComparisonOp stencil_back_func_func;
 
-                INSERT_PADDING_WORDS(0x17);
+                INSERT_PADDING_WORDS(0x4);
+
+                u32 frame_buffer_srgb_disabled;
```
bunnei: nit: `framebuffer_srgb_disabled`

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -416,10 +445,14 @@ struct SurfaceParams {
     }
 
     static PixelFormat PixelFormatFromTextureFormat(Tegra::Texture::TextureFormat format,
-                                                    Tegra::Texture::ComponentType component_type) {
+                                                    Tegra::Texture::ComponentType component_type,
+                                                    bool sRGB) {
```
bunnei: nit: `is_srgb`

src/video_core/renderer_opengl/gl_state.h
```diff
@@ -35,6 +35,10 @@ constexpr TextureUnit ProcTexDiffLUT{9};
 
 class OpenGLState {
 public:
+    struct {
+        bool enabled; // GL_FRAMEBUFFER_SRGB
+    } sRGB;
```
bunnei: nit: `framebuffer_srgb`

src/video_core/textures/texture.h
```diff
@@ -173,6 +173,7 @@ struct TICEntry {
     };
     union {
         BitField<0, 16, u32> width_minus_1;
+        BitField<22, 1, u32> s_rgb;
```
bunnei: nit: `srgb_conversion`, to match https://github.com/envytools/envytools/blob/0d91b8bcef3ceb47ff0b114025d301edb790d472/rnndb/graph/gm200_texture.xml#L322

src/video_core/textures/texture.h
```diff
@@ -227,6 +228,10 @@ struct TICEntry {
         return header_version == TICHeaderVersion::BlockLinear ||
                header_version == TICHeaderVersion::BlockLinearColorKey;
     }
+
+    bool IssRGB() const {
```
bunnei: nit: `IsSrgbConversionEnabled`

src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -77,19 +77,28 @@ struct SurfaceParams {
         ASTC_2D_8X8 = 45,
         ASTC_2D_8X5 = 46,
         ASTC_2D_5X4 = 47,
+        SBGRA8 = 48,
```
bunnei: A better name would probably be post-fixing all of these with `_SRGB`

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -116,7 +116,8 @@ RasterizerOpenGL::RasterizerOpenGL(Core::Frontend::EmuWindow& window, ScreenInfo
     state.Apply();
 
     glEnable(GL_BLEND);
-
+    // Enable sRGB by default
+    glEnable(GL_FRAMEBUFFER_SRGB);
```
bunnei: This shouldn't be necessary right? i.e. let the game turn it on if it uses it

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -417,6 +418,7 @@ void RasterizerOpenGL::ConfigureFramebuffers(bool using_color_fb, bool using_dep
 
     // Bind the framebuffer surfaces
     state.draw.draw_framebuffer = framebuffer.handle;
+    state.sRGB.enabled = (regs.frame_buffer_srgb_disabled == 0);
```
bunnei: This should probably be moved to a `SyncFramebufferSrgbEnabled()` method, like the other state sync functions we have in this class (and called above with the other Sync* methods).

src/video_core/renderer_opengl/gl_rasterizer_cache.h
bunnei: (we use `lowercase_with_underscores` for all variable names)

Tinob:
fixed naming, but still investigating will give more info later.

Tinob:
limit the application of sRGB to surfaces with formats that supports them, also introduce a workaround to support games like SMO and avoid enabling srgb on games that does not need it enabled. feedback is welcome.

src/video_core/renderer_opengl/gl_state.h
```diff
@@ -171,6 +180,7 @@ class OpenGLState {
 
 private:
     static OpenGLState cur_state;
+    static bool s_rgb_used;
```
FernandoS27: Add a comment in the tracking that this is a workaround to an NVIDIA bug and it LGTM as far as no games get affected incorrectly by it.

src/video_core/renderer_opengl/gl_rasterizer.cpp
Tinob: stays in this function because wee need to validate that the formats of the framebuffer supports srgb to really enable it to avoid problems with some nvidia drivers

PR #1586, opened 2018-10-25
file_sys/patch_manager: Remove unnecessary if-statements

FreddyFunk:


src/core/file_sys/patch_manager.cpp
```diff
@@ -53,23 +53,23 @@ PatchManager::~PatchManager() = default;
 VirtualDir PatchManager::PatchExeFS(VirtualDir exefs) const {
     LOG_INFO(Loader, "Patching ExeFS for title_id={:016X}", title_id);
 
-    if (exefs == nullptr)
-        return exefs;
+    if (exefs) {
```
lioncash: It's standard for us to use guard clauses to early exit. This should remain as is. Also the comment that it can be nullptr isn't necessary. The guard clause already did the job of communicating that.

src/core/file_sys/patch_manager.cpp
```diff
@@ -53,23 +53,23 @@ PatchManager::~PatchManager() = default;
...
+        const bool update_status =
+            update->GetStatus() == Loader::ResultStatus::ErrorMissingBKTRBaseRomFS;
 
-    // Game Updates
-    const auto update_tid = GetUpdateTitleID(title_id);
-    const auto update = installed->GetEntry(update_tid, ContentRecordType::Program);
-    if (update != nullptr) {
-        if (update->GetStatus() == Loader::ResultStatus::ErrorMissingBKTRBaseRomFS &&
-            update->GetExeFS() != nullptr) {
+        if (update && update_status && update->GetExeFS()) {
```
DarkLordZach: The `!= nullptr` should be left as it more effectively communicates the intent.

src/core/file_sys/patch_manager.cpp
lioncash: Yes. Because it tells you what the error condition is and any handling code needed to handle said error immediately. This is beneficial when someone is using the function elsewhere and needs to see how it returns. They don't need to read the whole function just to gather what error cases they need to handle. It also reduces the need to nest code in conditionals.

```cpp
void func() {
    if (!some_condition) {
        return;
    }

    /* ... */

    if (!some_other_condition) {
        return;
    }

    /* ... */
}
```

flows nicer than

```cpp
void func() {
    if (some_condition) {
        /* ... */

        if (some_other_condition) {
            /* ... */
        }
    }

    /* ... */
}
```

It's even more useful when dealing with stateful objects that perform behavior in their destructor, like closing a file handle, or releasing a mutex, there's no need to hold onto it for the duration of the whole function if there's an early-exit. See [here](https://en.wikipedia.org/wiki/Guard_(computer_science)) for more info.

FreddyFunk: Thanks a lot; I appriciate your explanation and see your point. Is this something that should be added to the `CONTRIBUTING.md`?

PR #1588, opened 2018-10-25
core: Remove unnessesary nullptr checks

FreddyFunk:
Make if-statements more readable by removing `== nullptr`.

``` C++
if(pointer == nullptr) {
// do something
}
```
equals to
``` C++
if(pointer) {
// do something
}
```


lioncash:
This is incorrect. It's actually equal to `!pointer`. Even so, this will cause warnings from [clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/readability-implicit-bool-conversion.html) about implicit conversions to bool.

PR #1590, opened 2018-10-25
core: Remove unnessesary nullptr checks

FreddyFunk:
Make if-statements more readable by removing `!= nullptr`.

``` C++
if(pointer != nullptr) {
// do something
}
```
equals to
``` C++
if(pointer) {
// do something
}
```


lioncash:
See #1588 for explanation. There's no need to do this sort of thing. Also relying on bool conversion in places like:

```cpp
bool IsThing() {
  return ptr;
}
```

is generally bad, since it can hide bugs. If you need a boolean expression in that case, it's best to explicitly use a boolean expression.

PR #1591, opened 2018-10-26
gl_rasterizer: Implement depth range.

bunnei:


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -119,6 +121,12 @@ void OpenGLState::Apply() const {
         glDepthMask(depth.write_mask);
     }
 
+    // Depth range
+    if (depth.depth_range_near != cur_state.depth.depth_range_near ||
+        depth.depth_range_far != cur_state.depth.depth_range_far) {
+        glDepthRangef(depth.depth_range_near, depth.depth_range_far);
```
ReinUsesLisp: `glDepthRangef` is core in OpenGL 4.1, you have to check for `GL_ARB_ES2_compatibility` or just use `glDepthRange`.

PR #1593, opened 2018-10-26
svc: Implement svcGetInfo command 0xF0000002

lioncash:
This retrieves:

```cpp
if (curr_thread == handle_thread) {
   result = total_thread_ticks + (hardware_tick_count - last_context_switch_ticks);
} else if (curr_thread == handle_thread && sub_id == current_core_index) {
   result = hardware_tick_count - last_context_switch_ticks;
}
```

which is fairly trivial to implement, given mostly everything else is in place already, aside from the tick updating itself

### DIFF
--- a/src/core/hle/kernel/process.h
+++ b/src/core/hle/kernel/process.h
@@ -202,6 +202,16 @@ public:
         return is_64bit_process;
     }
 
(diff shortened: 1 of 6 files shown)


lioncash:
Error isn't related to the code in the PR, it was a web-related error.

PR #1598, opened 2018-10-27
service/filesystem: Implemented DeleteDirectory & DeleteDirectoryRecursive

DeeJayBro:
Implements DeleteDirectory & DeleteDirectoryRecursive Commands into the Service.
With this change, Lego Worlds gets ingame.


src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -329,6 +329,30 @@ class IFileSystem final : public ServiceFramework<IFileSystem> {
         rb.Push(backend.CreateDirectory(name));
     }
 
+    void DeleteDirectory(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+
+        auto file_buffer = ctx.ReadBuffer();
```
lioncash: This and the below variable can be const. Ditto below for the recursive variant.

PR #1605, opened 2018-10-28
Port citra-emu/citra#4382: "configure_general: Add an option to reset defaults"

FearlessTobi:
See citra-emu/citra#4382 for more details.

**Original description:**
I think this change is fairly self-explanatory.

When the button is clicked, the settings file gets deleted and Citra gets closed.

Screenshots:
unbenannt1
unbenannt2

### DIFF
--- a/src/yuzu/configuration/configure_general.cpp
+++ b/src/yuzu/configuration/configure_general.cpp
@@ -2,6 +2,8 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <QMessageBox>
+#include "common/file_util.h"
 #include "core/core.h"
 #include "core/settings.h"
 #include "ui_configure_general.h"
@@ -23,6 +25,9 @@ ConfigureGeneral::ConfigureGeneral(QWidget* parent)
             [] { UISettings::values.is_game_list_reload_pending.exchange(true); });
 
     ui->use_cpu_jit->setEnabled(!Core::System::GetInstance().IsPoweredOn());
+
+    connect(ui->button_reset_defaults, &QPushButton::clicked, this,
+            &ConfigureGeneral::ResetDefaults);
 }
 
 ConfigureGeneral::~ConfigureGeneral() = default;
@@ -33,6 +38,17 @@ void ConfigureGeneral::setConfiguration() {
     ui->toggle_user_on_boot->setChecked(UISettings::values.select_user_on_boot);
     ui->theme_combobox->setCurrentIndex(ui->theme_combobox->findData(UISettings::values.theme));
     ui->use_cpu_jit->setChecked(Settings::values.use_cpu_jit);
+}
+
+void ConfigureGeneral::ResetDefaults() {
+    QMessageBox::StandardButton answer = QMessageBox::question(
+        this, tr("yuzu"),
+        tr("Are you sure you want to <b>reset your settings</b> and close yuzu?"),
+        QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
+    if (answer == QMessageBox::No)
+        return;
+    FileUtil::Delete(FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir) + "qt-config.ini");
+    std::exit(0);
 }
 
 void ConfigureGeneral::applyConfiguration() {
--- a/src/yuzu/configuration/configure_general.h
+++ b/src/yuzu/configuration/configure_general.h
@@ -20,6 +20,7 @@ public:
     explicit ConfigureGeneral(QWidget* parent = nullptr);
     ~ConfigureGeneral() override;
 
+    void ResetDefaults();
     void applyConfiguration();
 
 private:
--- a/src/yuzu/configuration/configure_general.ui
+++ b/src/yuzu/configuration/configure_general.ui
@@ -97,6 +97,13 @@
        </layout>
       </widget>
      </item>
+     <item alignment="Qt::AlignRight">
+      <widget class="QPushButton" name="button_reset_defaults">
+       <property name="text">
+        <string>Reset All Settings</string>
+       </property>
+      </widget>
+     </item>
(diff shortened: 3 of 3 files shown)


src/yuzu/configuration/configure_general.cpp
```diff
@@ -38,6 +43,17 @@ void ConfigureGeneral::PopulateHotkeyList(const HotkeyRegistry& registry) {
...
 
+void ConfigureGeneral::ResetDefaults() {
+    QMessageBox::StandardButton answer = QMessageBox::question(
+        this, tr("yuzu"),
+        tr("Are you sure you want to <b>reset your settings</b> and close yuzu?"),
+        QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
+    if (answer == QMessageBox::No)
+        return;
+    FileUtil::Delete(FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir) + "qt-config.ini");
+    std::exit(0);
```
lioncash: Sorry, but no. Killing the program is something that should absolutely not happen when resetting defaults. Again, sorry, but that's completely unacceptable. This is a very sloppy way to handle the reloading of settings.

bunnei:
@FearlessTobi reminder to finish this up

wwylele:
I want to put my thought here for reference as I silently let this pass in citra. I hate this, but I didn't reject it because of the following thought and arguing with myself:

The need for citra of this option was mainly for those user who screwed up their config (by copy-paste others, modifying without knowledge or stuff), and more importantly, for those who doesn't share this information when asking for support. So we needed a true "reset to factory" button that really reset the config to the point as newly-created. It is also the last resort when we screw up some config load/save code and we want user to be able to do clean test in this unusual situation. This was the hard requirement.

With this in mind, I considered some potentially better solution. An good one was to not exit the program, and to reset all configs in the in-program memory (The Settings object etc.). However, this leaded to some problems to resolve: this only behaves as expect when the code is correct, as in a wrong config-reset code will generates a state other than the factory state. But this button is really useful only when something, such as the code, goes wrong, as I said above. Reusing LoadConfig code for reset is not appropriate either, as it will create a new config file immediately, which isn't the factory state either. I guess there is still a solution to this, but it becomes non-trivial and nobody can say it works as intended by just reading the code (unlike the `std::exit` version).

I convinced myself by making an analogy to the "reset" button on the AC remote control in my home. It is only used for unusual states, and it shuts down the control.

Again, this is just what I thought when I saw the code in citra. I still hate it, but I couldn't find a solid reason to reject it. I'd like to hear a better solution, and also adopt it in citra if applicable.

src/yuzu/configuration/configure_general.cpp
chris062689: For context, this isn't just restoring the configuration to default. 

Because it deletes the config file entirely, and lets it reload from defaults on the next launch, it cleans up old configuration entries that were stale / removed from newer versions. 
The reason why it's killing the application, is because closing "gracefully" would save the configuration we just "deleted" -- in effect not doing what we wanted. 

We wanted a way to restart the configuration with a clean slate -- without old / stale entries.

What would be a good way of achieving this?

FearlessTobi:
This PR still depends on either the original approach being accepted or https://github.com/yuzu-emu/yuzu/pull/2048 getting merged.

FearlessTobi:
Rebased. Does anyone want to voice their opinion now on what approach we should take with this?

lioncash:
My opinion still remains that a configuration option should not kill the application, or require it to be killed in order to properly reset the configuration. The proper fix is to do it in a manner that doesn't require shutting down and restarting the application.

This is a "fix" that doesn't actually address the underlying issue and glosses over it (the underlying config system handling itself).

An AC power analogy doesn't really work here, because it's absolutely a situation we have full control over and *have* a means to fix properly. Non-triviality as an argumentative reason for not handling something properly (*especially* when it's a file handling system we have full control over) is a cop-out in this case. If that were a solid reason, every error case would be dealt with by using `std::exit`, but it's not, that's not how proper error handling works in most cases. It's for unrecoverable execution state; this is not a situation where that is the case. It makes for sloppy UX.

Regardless of the outcome, this is going to be fixed properly one way or another. Either it gets done properly, and it's done. Or it doesn't, and it just gets added to my backlog of things to tackle, then gets done properly (*do not* use this as a reason to not consider the former outcome).

wwylele:
So I tried to come up with some more constructive idea:

First of all, no matter how this is implemented, almost *everything* in the program has to do a reload, including all window and all emulation stuff, because they all depends on some config value.

Without actually shutting down the program, the only place to reload everything is in the `main` function. More specifically, the "Reset" button should trigger a return to the `main` function, then destruct all object and reconstruct them. 

This implementation can be easy to make to prove the correctness if there is zero global state in the entire code base.

So ultimately this boils down to the global state clean up task. Good luck :p

lioncash:
This isn't *necessarily* as painful as it sounds. Obviously, the option shouldn't be available when emulation is actually running. During shutdown, our shutdown functions should be invalidating all state (it's essentially a lifetime bug/issue otherwise, which is unrelated to the configuration handling).

So, it'd only be sufficient to handle the UI side of things when it comes to flushing the configuration values. This can be handled by dispatching a UI signal intended to be handled by receivers that actually need to do anything meaningful (which, for the most part, is just the rest of the configuration menu).

Any outliers to that should be corrected,

bunnei:
@FearlessTobi what is the status of this one?

bunnei:
@FearlessTobi ping again on this one - is this something we can fix now and merge, or should this be closed?

FearlessTobi:
I won't be available at all for the next 3 weeks, I still think though I can fix this eventually.

bunnei:
@FearlessTobi 
image

(but sersiously -- can we get these old PRs mergable or closed if dead? Thanks !)

lioncash:
Closing due to inactivity

PR #1608, opened 2018-10-28
[ns|fsp_srv]: Implement various functions to boot Checkpoint

DarkLordZach:
This includes:
- `fsp_srv` 61: `OpenSaveDataInfoReaderBySaveDataSpaceId` - This returns an object capable of iterating over the currently present saves on the system.
- `fsp_srv` `ISaveDataInfoReader`: The aforementioned iterator object -- I implemented it by iterating over all of the directories in the SaveDataSpace (i.e. /nand/user/save) and filling a vector of SaveDataInfo objects using the directory data.
- `fsp_srv` `ISaveDataInfoReader` 0: `ReadSaveDataInfo` - Reads a game-provided number of entries from the total, resuming where left off between successive calls.
- `ns:am` 400: `GetApplicationControlData` - This returns the raw NACP and icon bytes given the title ID. The data source of this for this implementation is the registration cache (installed titles).

Finally, some basic tests to ensure the backup/restore functionality worked were conducted. Thanks to #1598, it works perfectly.

Screenshot:
checkpoint1

### DIFF
--- a/src/core/file_sys/savedata_factory.h
+++ b/src/core/file_sys/savedata_factory.h
@@ -52,7 +52,7 @@ public:
 
     ResultVal<VirtualDir> Open(SaveDataSpaceId space, SaveDataDescriptor meta);
 
-    VirtualDir GetSaveDataSpaceDirectory(SaveDataSpaceId space);
+    VirtualDir GetSaveDataSpaceDirectory(SaveDataSpaceId space) const;
 
     static std::string GetSaveDataSpaceIdPath(SaveDataSpaceId space);
     static std::string GetFullPath(SaveDataSpaceId space, SaveDataType type, u64 title_id,
--- a/src/core/file_sys/savedata_factory.h
+++ b/src/core/file_sys/savedata_factory.h
@@ -52,6 +52,9 @@ public:
 
     ResultVal<VirtualDir> Open(SaveDataSpaceId space, SaveDataDescriptor meta);
 
+    VirtualDir GetSaveDataSpaceDirectory(SaveDataSpaceId space);
+
+    static std::string GetSaveDataSpaceIdPath(SaveDataSpaceId space);
     static std::string GetFullPath(SaveDataSpaceId space, SaveDataType type, u64 title_id,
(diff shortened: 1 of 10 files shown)


src/core/file_sys/savedata_factory.h
```diff
@@ -52,6 +52,9 @@ class SaveDataFactory {
 
     ResultVal<VirtualDir> Open(SaveDataSpaceId space, SaveDataDescriptor meta);
 
+    VirtualDir GetSaveDataSpaceDirectory(SaveDataSpaceId space);
```
lioncash: This can probably be a const member function.

src/core/hle/service/ns/ns.cpp
```diff
@@ -243,6 +245,65 @@ class IApplicationManagerInterface final : public ServiceFramework<IApplicationM
 
         RegisterHandlers(functions);
     }
+
+    void GetApplicationControlData(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        const auto flag = rp.PopRaw<u64>();
+        LOG_DEBUG(Service_NS, "called with flag={:016X}", flag);
```
lioncash: `common/logging/log.h` needs to be included.

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -451,7 +452,151 @@ class IFileSystem final : public ServiceFramework<IFileSystem> {
...
+        const auto bytes = Common::HexStringToArray<0x8>(str);
+        u64 out{};
+        std::memcpy(&out, bytes.data(), sizeof(u64));
+
+        return Common::swap64(out);
+    }
+
+    static std::array<u8, 0x10> ArraySwap(const std::array<u8, 0x10>& in) {
+        std::array<u8, 0x10> out;
+        for (std::size_t i = 0; i < in.size(); ++i) {
```
lioncash: Can't this just be a `std::reverse` operation?

PR #1610, opened 2018-10-28
renderer_opengl: Enable alpha channel for DXT1 texture format

slashiee:
This fixes textures that showed black for their alphas instead of being see-through.

Before:

After:


FernandoS27:
Nice catch.

PR #1613, opened 2018-10-29
video_core: Move OpenGL specific utils to its renderer

ReinUsesLisp:
`video_core/utils.h` had OpenGL-specific code that depended on `glad/glad.h` (which was not included). Instead of adding it, which also poluted the symbols, a new file for those utils was created.

### DIFF
--- a/src/video_core/renderer_opengl/utils.h
+++ b/src/video_core/renderer_opengl/utils.h
@@ -1,0 +1,15 @@
+// Copyright 2014 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <string>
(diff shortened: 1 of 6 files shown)


src/video_core/renderer_opengl/utils.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#include <string>
+#include <fmt/format.h>
+#include <glad/glad.h>
+#include "common/common_types.h"
+
+namespace OpenGL {
+
+static void LabelGLObject(GLenum identifier, GLuint handle, VAddr addr,
```
lioncash: This should probably be defined in a cpp file, so it doesn't drag in stuff like fmt to whatever is including it.

PR #1614, opened 2018-10-29
video_core: Move surface declarations out of gl_rasterizer_cache

ReinUsesLisp:
Moves `SurfaceParams`' enums, static methods and constexpr functions out of `gl_rasterizer_cache.h` aswell as some other surface related operations defined in `gl_rasterizer_cache.cpp`. These are API agnostic so they make more sense in `video_core` (or a to-be-defined common interface).


bunnei:
Needs a rebase

PR #1615, opened 2018-10-29
configure_system: Contrain profile usernames to 32 characters

lioncash:
Previously, we would let a user enter an unbounded name and then silently truncate away characters that went over the 32-character limit. This is kind of bad from the UX point of view, because we're essentially not doing what the user intended in certain scenarios.

Instead, we clamp it to 32 characters and make that visually apparent in the dialog box to provide a name for a user.

### DIFF
--- a/src/core/hle/service/acc/profile_manager.h
+++ b/src/core/hle/service/acc/profile_manager.h
@@ -57,7 +57,8 @@ struct UUID {
 };
 static_assert(sizeof(UUID) == 16, "UUID is an invalid size!");
 
-using ProfileUsername = std::array<u8, 0x20>;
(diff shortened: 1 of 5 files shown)


src/core/hle/service/acc/profile_manager.h
```diff
@@ -57,7 +57,8 @@ struct UUID {
 };
 static_assert(sizeof(UUID) == 16, "UUID is an invalid size!");
 
-using ProfileUsername = std::array<u8, 0x20>;
+constexpr size_t profile_username_size = 32;
```
bunnei: `std::size_t`

lioncash:
Oh, forgot to comment that I amended the commit

PR #1622, opened 2018-10-30
maxwell_3d: Restructure macro upload to use a single macro code memory.

bunnei:
This changes how we upload macro code. Instead of piping macro code into separate memory for each program, we write it to a single macro memory region. We then execute macros via the specified offset into this region. This fixes an issue where macros could be skipped before. Skipping macros was the cause of distant objects in Super Mario Odyssey not being rendered.

image

### DIFF
--- a/src/video_core/engines/maxwell_3d.h
+++ b/src/video_core/engines/maxwell_3d.h
@@ -475,12 +475,13 @@ public:
                 INSERT_PADDING_WORDS(0x45);
 
                 struct {
-                    INSERT_PADDING_WORDS(1);
+                    u32 upload_address;
                     u32 data;
                     u32 entry;
+                    u32 bind;
                 } macros;
 
-                INSERT_PADDING_WORDS(0x189);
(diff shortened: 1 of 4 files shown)


src/video_core/engines/maxwell_3d.h
```diff
@@ -994,12 +995,28 @@ class Maxwell3D final {
...
+        return macro_memory;
+    }
+
 private:
     void InitializeRegisterDefaults();
 
     VideoCore::RasterizerInterface& rasterizer;
 
-    std::unordered_map<u32, std::vector<u32>> uploaded_macros;
+    static constexpr std::size_t MACRO_COUNT{0x80};
```
Subv: What is this count and where did it come from?

bunnei: This was taken from Ryujinx and appears to be fairly arbitrary. I fixed it to be unbounded and not rely on this count/mask.

PR #1623, opened 2018-10-30
Improve OpenGL state handling

Tinob:
reduce redundant state changes and ensure default state to avoid driver implementation issues.

### DIFF
--- a/src/video_core/renderer_opengl/gl_state.cpp
+++ b/src/video_core/renderer_opengl/gl_state.cpp
@@ -89,7 +89,18 @@ OpenGLState::OpenGLState() {
     point.size = 1;
 }
 
-void OpenGLState::Apply() const {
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -100,59 +111,67 @@ void OpenGLState::Apply() const {
...
         }
     }
-    // Culling
-    if (cull.enabled != cur_state.cull.enabled) {
-        if (cull.enabled) {
-            glEnable(GL_CULL_FACE);
-        } else {
-            glDisable(GL_CULL_FACE);
+    {
+        // Culling
```
lioncash: These should be migrated to their own functions that are called in this one instead of sectioning off the function.

Tinob: done

PR #1630, opened 2018-11-01
memory_manager: Do not MapBufferEx over already in use memory.

bunnei:
- This fixes rendering when changing areas in Super Mario Odyssey.

### DIFF
--- a/src/video_core/memory_manager.cpp
+++ b/src/video_core/memory_manager.cpp
@@ -4,18 +4,21 @@
 
 #include "common/alignment.h"
 #include "common/assert.h"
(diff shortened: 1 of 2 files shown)


EmulationChannel:
@bunnei Oky after the Worlds Rendering rendering still contains some bug visible flaws and menu SMO
anotacao 2018-11-01 084444

FearlessTobi:
@EmulationChannel This is not at all related to this PR and not very helpful as we are obviously already aware of the issues.

ogniK5377:
Various games don't spam assertions anymore 👍

PR #1637, opened 2018-11-03
Improved GPU Caches lookup Speed

FernandoS27:
This PR adds a hash table to the cache in order to improve look-up speed. This is important since 90% of the operations on caches are look-ups.


bunnei:
This breaks clang build

PR #1638, opened 2018-11-03
Implement SetMemoryPermission

FreddyFunk:
~~Rayman Legends: Definitive Edition goes from an instant crash on boot into a fully working menu.~~

~~Rayman Legends: Definitive Edition was crashing on boot before and is now playable.~~

Rayman Legends: Definitive Edition was crashing on boot before and now is playable now without the uniformity check of the memory. To provide these checks some changes to the memory manger may be required (see comment by @lioncash).

Impementation based off suggested code by @lioncash

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -122,6 +122,12 @@ static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
     return RESULT_SUCCESS;
 }
 
+static ResultCode SetMemoryPermission(VAddr addr, u64 size, u32 prot) {
+    LOG_WARNING(Kernel_SVC, "(STUBBED) called, addr=0x{:X}, size=0x{:X}, prot=0x{:X}", addr, size,
+                prot);
+    return RESULT_SUCCESS;
+}
+
 static ResultCode SetMemoryAttribute(VAddr addr, u64 size, u32 state0, u32 state1) {
     LOG_WARNING(Kernel_SVC,
                 "(STUBBED) called, addr=0x{:X}, size=0x{:X}, state0=0x{:X}, state1=0x{:X}", addr,
@@ -1259,7 +1265,7 @@ struct FunctionDef {
 static const FunctionDef SVC_Table[] = {
     {0x00, nullptr, "Unknown"},
     {0x01, SvcWrap<SetHeapSize>, "SetHeapSize"},
-    {0x02, nullptr, "SetMemoryPermission"},
+    {0x02, SvcWrap<SetMemoryPermission>, "SetMemoryPermission"},
     {0x03, SvcWrap<SetMemoryAttribute>, "SetMemoryAttribute"},
     {0x04, SvcWrap<MapMemory>, "MapMemory"},
     {0x05, SvcWrap<UnmapMemory>, "UnmapMemory"},
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -123,9 +123,45 @@ static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
 }
 
 static ResultCode SetMemoryPermission(VAddr addr, u64 size, u32 prot) {
-    LOG_WARNING(Kernel_SVC, "(STUBBED) called, addr=0x{:X}, size=0x{:X}, prot=0x{:X}", addr, size,
-                prot);
-    return RESULT_SUCCESS;
+    LOG_TRACE(Kernel_SVC, "called, addr=0x{:X}, size=0x{:X}, prot=0x{:X}", addr, size, prot);
+
+    if (!Common::Is4KBAligned(addr)) {
+        return ERR_INVALID_ADDRESS;
+    }
+
+    if (size == 0 || !Common::Is4KBAligned(size)) {
+        return ERR_INVALID_SIZE;
+    }
+
+    if (!IsValidAddressRange(addr, size)) {
+        return ERR_INVALID_ADDRESS_STATE;
+    }
+
+    const auto permission = static_cast<MemoryPermission>(prot);
+    if (permission != MemoryPermission::None && permission != MemoryPermission::Read &&
+        permission != MemoryPermission::ReadWrite) {
+        return ERR_INVALID_MEMORY_PERMISSIONS;
+    }
+
+    auto* const current_process = Core::CurrentProcess();
+    auto& vm_manager = current_process->VMManager();
+
+    if (!IsInsideAddressSpace(vm_manager, addr, size)) {
+        return ERR_INVALID_ADDRESS_STATE;
+    }
+
+    const VMManager::VMAHandle iter = vm_manager.FindVMA(addr);
+    if (iter == vm_manager.vma_map.end()) {
+        return ERR_INVALID_ADDRESS_STATE;
+    }
+
+    LOG_WARNING(Kernel_SVC, "Uniformity check on protected memory is not implemented.");
+    // TODO: Performs a uniformity check to make sure only protected memory is changed (it doesn't
+    // make sense to allow changing permissions on kernel memory itself, etc).
+
+    const auto converted_permissions = SharedMemory::ConvertPermissions(permission);
+
+    return vm_manager.ReprotectRange(addr, size, converted_permissions);
 }
 
 static ResultCode SetMemoryAttribute(VAddr addr, u64 size, u32 state0, u32 state1) {
--- a/src/core/hle/kernel/svc_wrap.h
+++ b/src/core/hle/kernel/svc_wrap.h
@@ -121,6 +121,11 @@ void SvcWrap() {
     FuncReturn(func(Param(0), Param(1), Param(2)).raw);
 }
 
+template <ResultCode func(u64, u64, u32)>
+void SvcWrap() {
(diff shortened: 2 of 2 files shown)


DarkLordZach:
Would it be possible to implement this instead of stubbing it? Our memory manager supports permissions, so it shouldn't be too difficult.

lioncash:
An implementation of it would likely be:

```patch
c229d65b399e0df29e0233c1fc7a5dc239468f59
 src/core/hle/kernel/svc.cpp      | 56 +++++++++++++++++++++++++++++++++++++++-
 src/core/hle/kernel/svc_wrap.h   |  5 ++++
 src/core/hle/kernel/vm_manager.h | 30 +++++++++++++++++++++
 3 files changed, 90 insertions(+), 1 deletion(-)

diff --git a/src/core/hle/kernel/svc.cpp b/src/core/hle/kernel/svc.cpp
index c7c579aaf..3dec108f7 100644
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -122,6 +122,60 @@ static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
     return RESULT_SUCCESS;
 }
 
+static ResultCode SetMemoryPermission(VAddr address, u64 size, u32 permission) {
+    LOG_TRACE("called, address=0x{:X}, size={:X}, permission={:X}", address, size, permission);
+
+    if (!Common::Is4KBAligned(address)) {
+        return ERR_INVALID_ADDRESS;
+    }
+
+    if (size == 0 || !Common::Is4KBAligned(size)) {
+        return ERR_INVALID_SIZE;
+    }
+
+    if (address + size <= address) {
+        return ERR_INVALID_ADDRESS_STATE;
+    }
+
+    const auto perms = static_cast<MemoryPermission>(permission);
+    if (perms != MemoryPermission::None && perms != MemoryPermission::Read &&
+        perms != MemoryPermission::ReadWrite) {
+        return ERR_INVALID_MEMORY_PERMISSIONS;
+    }
+
+    auto* const current_process = Core::CurrentProcess();
+    auto& vm_manager = current_process->VMManager();
+
+    if (!IsInsideAddressSpace(vm_manager, address, size)) {
+        return ERR_INVALID_ADDRESS_STATE;
+    }
+
+    const VMManager::VMAHandle iter = vm_manager.FindVMA(address);
+    if (iter == vm_manager.vma_map.end()) {
+        return ERR_INVALID_ADDRESS_STATE;
+    }
+
+    const VirtualMemoryArea& reference_vma = iter->second;
+    const bool is_uniform_range =
+        vm_manager.CheckIfAll(address, address + size, [&reference_vma](const auto& vma) {
+            return reference_vma.meminfo_state == vma.meminfo_state &&
+                   reference_vma.permissions == vma.permissions;
+        });
+
+    if (!is_uniform_range) {
+        return ERR_INVALID_ADDRESS_STATE;
+    }
+
+    // No need to reprotect the range if the permissions are exactly
+    // the same across the whole memory region.
+    const auto converted_permissions = SharedMemory::ConvertPermissions(perms);
+    if (converted_permissions == reference_vma.permissions) {
+        return RESULT_SUCCESS;
+    }
+
+    return vm_manager.ReprotectRange(address, size, converted_permissions);
+}
+
 static ResultCode SetMemoryAttribute(VAddr addr, u64 size, u32 state0, u32 state1) {
     LOG_WARNING(Kernel_SVC,
                 "(STUBBED) called, addr=0x{:X}, size=0x{:X}, state0=0x{:X}, state1=0x{:X}", addr,
@@ -1259,7 +1313,7 @@ struct FunctionDef {
 static const FunctionDef SVC_Table[] = {
     {0x00, nullptr, "Unknown"},
     {0x01, SvcWrap<SetHeapSize>, "SetHeapSize"},
-    {0x02, nullptr, "SetMemoryPermission"},
+    {0x02, SvcWrap<SetMemoryPermission>, "SetMemoryPermission"},
     {0x03, SvcWrap<SetMemoryAttribute>, "SetMemoryAttribute"},
     {0x04, SvcWrap<MapMemory>, "MapMemory"},
     {0x05, SvcWrap<UnmapMemory>, "UnmapMemory"},
diff --git a/src/core/hle/kernel/svc_wrap.h b/src/core/hle/kernel/svc_wrap.h
index b09753c80..233a99fb0 100644
--- a/src/core/hle/kernel/svc_wrap.h
+++ b/src/core/hle/kernel/svc_wrap.h
@@ -121,6 +121,11 @@ void SvcWrap() {
     FuncReturn(func(Param(0), Param(1), Param(2)).raw);
 }
 
+template <ResultCode func(u64, u64, u32)>
+void SvcWrap() {
+    FuncReturn(func(Param(0), Param(1), static_cast<u32>(Param(2))).raw);
+}
+
 template <ResultCode func(u32, u64, u64, u32)>
 void SvcWrap() {
     FuncReturn(
diff --git a/src/core/hle/kernel/vm_manager.h b/src/core/hle/kernel/vm_manager.h
index 2447cbb8f..3aab9b42d 100644
--- a/src/core/hle/kernel/vm_manager.h
+++ b/src/core/hle/kernel/vm_manager.h
@@ -165,6 +165,36 @@ public:
      */
     ResultVal<VAddr> FindFreeRegion(u64 size) const;
 
+    /**
+     * Checks if a given condition holds across all VMAs in an address range.
+     *
+     * @param begin_address The starting address in the range
+     * @param end_address   The ending address within the range.
+     * @param predicate     A unary predicate function accepting a constant
+     *                      VirtualMemoryArea reference.
+     *
+     * @return True if the condition holds across the entire range.
+     *
+     * @note The ending range is calculated as: begin + end - 1, not begin + end
+     */
+    template <typename UnaryPredicate>
+    bool CheckIfAll(VAddr begin_address, VAddr end_address, UnaryPredicate predicate) const {
+        const VAddr target_end = begin_address + end_address - 1;
+
+        VMAHandle vma = FindVMA(begin_address);
+        const VMAHandle end = vma_map.cend();
+
+        while (vma != end && vma->second.base < target_end) {
+            if (!predicate(vma->second)) {
+            }
+
+            vma = std::next(vma);
+        }
+
+        return true;
+    }
+
     /**
      * Maps a memory-mapped IO region at a given address.
      *
```

FreddyFunk:
@lioncash Thanks for the suggestion. I used most of your code, but excluded the uniform check.. I do not understand why it is important to check if the memory range has to have an uniform permission before it will be overwritten with a single new one. This also results in a broke exectution.




Also I stuck to the parameter names providen by SwitchBrew.

src/core/hle/kernel/svc.cpp
```diff
@@ -123,9 +123,40 @@ static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
...
+
+    if (!Common::Is4KBAligned(addr)) {
+        return ERR_INVALID_ADDRESS;
+    }
+
+    if (size == 0 || !Common::Is4KBAligned(size)) {
+        return ERR_INVALID_SIZE;
+    }
+
+    if (addr + size <= addr) {
```
InusualZ: 'addr + size <= addr' am I missing something or this branch is pointless

FreddyFunk: It's an overflow check

lioncash:
You should probably put the fact that you used most of my code verbatim in your commit message btw (i.e. crediting people's work you've used).

Also the reason the kernel performs a uniformity check is to make sure only protected memory is changed (it doesn't make sense to allow changing permissions on kernel memory itself, etc). The check should likely be changed to:

```cpp
const bool is_uniform_range =
        vm_manager.CheckIfAll(address, address + size, [](const auto& vma) {
            return vma.state == MemoryState::Normal || vma.state == MemoryState::CodeMutable;
        });

**Comment by FreddyFunk** [comment]
That's right. Please tell me if you're fine with the way I credited your work.

I committed your second suggestion of the `CheckIfAll` function, but it results in a broke execution as well..

**Comment by lioncash** [comment]
Oh OK. Then this is fine as is. I was more or less just saying toss in a message that's like: "Based off suggested code by <x>" or whatever in the extended description. I don't need to be in the title. It's more just a reference so people know who to ask about specifics in the future if they trace an issue back to the commit.

**Comment by lioncash** [comment]
Err, to clarify by fine as is, I mean fine as is without the permission checks that break execution for the time being.

File: src/core/hle/kernel/svc.cpp [hunk 2]
```diff
@@ -122,6 +122,58 @@ static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
...
```
ogniK5377: `IsValidAddressRange` for clarity on what this actually does

src/core/hle/kernel/svc.cpp
```diff
@@ -122,6 +122,44 @@ static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
     return RESULT_SUCCESS;
 }
 
+static ResultCode SetMemoryPermission(VAddr addr, u64 size, u32 prot) {
+    LOG_TRACE(Kernel_SVC, "(STUBBED) called, addr=0x{:X}, size=0x{:X}, prot=0x{:X}", addr, size,
```
lioncash: The `(STUBBED)` can be removed.

FreddyFunk: Done

FreddyFunk:
Addressed feedback from @lioncash. Are details like the order of checks from the kernel somewhere documented? If so, please provide a source so I could prevent a messy PR like this one in the future.

PR #1640, opened 2018-11-03
game_list: Only reload game list after relevant settings changed

DarkLordZach:
Prevents unnecessary reloads on every configuration operation.


src/yuzu/configuration/configure_gamelist.h
```diff
@@ -20,6 +20,9 @@ class ConfigureGameList : public QWidget {
 
     void applyConfiguration();
 
+public slots:
```
lioncash: This can be private. It's also not necessary to use `slots` in Qt5+ anymore (it was necessary back when everything was string-based in Qt 4 and lower).

src/yuzu/configuration/configure_general.ui
```diff
@@ -66,18 +66,11 @@
      <item>
       <widget class="QGroupBox" name="EmulationGroupBox">
        <property name="title">
-        <string>Emulation</string>
```
lioncash: This and below don't seem related to the change.

FearlessTobi: Why was this change even made in the first place?

DarkLordZach: Sorry, when i was separating this from hid 2 i missed this.
Fixed now

PR #1641, opened 2018-11-04
sm: Implement RegisterService and UnregisterService

DarkLordZach:
These are needed by Edizon to boot. They are used to see if a user is using SX OS, as SX OS registers a custom service called 'tx' and attempting to register a service of the same name lets the application know if it is present.

### DIFF
--- a/src/core/hle/service/sm/sm.cpp
+++ b/src/core/hle/service/sm/sm.cpp
@@ -63,6 +63,17 @@ ResultVal<Kernel::SharedPtr<Kernel::ServerPort>> ServiceManager::RegisterService
     return MakeResult<Kernel::SharedPtr<Kernel::ServerPort>>(std::move(server_port));
 }
 
+ResultCode ServiceManager::UnregisterService(std::string name) {
+    CASCADE_CODE(ValidateServiceName(name));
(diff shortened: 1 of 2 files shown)


src/core/hle/service/sm/sm.cpp
```diff
@@ -127,13 +138,52 @@ void SM::GetService(Kernel::HLERequestContext& ctx) {
...
 
+void SM::RegisterService(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+
+    const auto name_buf = rp.PopRaw<std::array<char, 8>>();
+    const auto end = std::find(name_buf.begin(), name_buf.end(), '\0');
+
+    const std::string name(name_buf.begin(), end);
+
+    const auto unk_bool = static_cast<bool>(rp.PopRaw<u32>());
```
lioncash: The bool determines whether or not the session is considered "lightweight". By that, it has a reduced set of functionality (for instance, a light session cannot send an async request, but regular ones can). Also when light client sessions are destroyed, they won't decrement the reference count of their parent server session. i.e. they can be considered weak handles.

PR #1646, opened 2018-11-05
svcBreak now dumps information from the debug buffer passed

ogniK5377:
info1 and info2 seem to sometimes hold an address to a buffer, this is usually 4 bytes or the size of the int and contains an error code. There are other circumstances where it can be something different so we hexdump these to examine them at a later date.

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -409,11 +409,11 @@ static void Break(u32 reason, u64 info1, u64 info2) {
             // We don't know what's in here so we'll hexdump it
             std::vector<u8> debug_buffer(sz);
             Memory::ReadBlock(addr, debug_buffer.data(), sz);
-            std::string hexdump = "";
-            for (auto i = 0; i < debug_buffer.size(); i++) {
+            std::string hexdump;
+            for (std::size_t i = 0; i < debug_buffer.size(); i++) {
                 hexdump += fmt::format("{:02X} ", debug_buffer[i]);
                 if (i != 0 && i % 16 == 0) {
-                    hexdump += "\n";
+                    hexdump += '\n';
                 }
             }
             LOG_CRITICAL(Debug_Emulated, "debug_buffer=\n{}", hexdump);
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -395,16 +395,42 @@ struct BreakReason {
 /// Break program execution
 static void Break(u32 reason, u64 info1, u64 info2) {
     BreakReason break_reason{reason};
+    bool has_dumped_buffer{};
 
+    const auto handle_debug_buffer = [&](VAddr addr, u64 sz) {
+        if (sz == 0 || addr == 0 || has_dumped_buffer) {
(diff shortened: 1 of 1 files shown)


src/core/hle/kernel/svc.cpp
```diff
@@ -395,16 +395,42 @@ struct BreakReason {
...
+            LOG_CRITICAL(Debug_Emulated, "debug_buffer_err_code={:X}", Memory::Read32(addr));
+        } else {
+            // We don't know what's in here so we'll hexdump it
+            std::vector<u8> debug_buffer(sz);
+            Memory::ReadBlock(addr, debug_buffer.data(), sz);
+            std::string hexdump = "";
+            for (auto i = 0; i < debug_buffer.size(); i++) {
+                hexdump += fmt::format("{:02X} ", debug_buffer[i]);
+                if (i != 0 && i % 16 == 0) {
+                    hexdump += "\n";
```
lioncash: `'\n'`

src/core/hle/kernel/svc.cpp
```diff
@@ -395,16 +395,42 @@ struct BreakReason {
...
+
+        // This typically is an error code so we're going to assume this is the case
+        if (sz == sizeof(u32)) {
+            LOG_CRITICAL(Debug_Emulated, "debug_buffer_err_code={:X}", Memory::Read32(addr));
+        } else {
+            // We don't know what's in here so we'll hexdump it
+            std::vector<u8> debug_buffer(sz);
+            Memory::ReadBlock(addr, debug_buffer.data(), sz);
+            std::string hexdump = "";
+            for (auto i = 0; i < debug_buffer.size(); i++) {
```
lioncash: `size_t i = 0`, otherwise this deduces as an `int` and the loop has a signed/unsigned mismatch in its termination condition.

src/core/hle/kernel/svc.cpp
```diff
@@ -395,16 +395,42 @@ struct BreakReason {
...
+        }
+
+        // This typically is an error code so we're going to assume this is the case
+        if (sz == sizeof(u32)) {
+            LOG_CRITICAL(Debug_Emulated, "debug_buffer_err_code={:X}", Memory::Read32(addr));
+        } else {
+            // We don't know what's in here so we'll hexdump it
+            std::vector<u8> debug_buffer(sz);
+            Memory::ReadBlock(addr, debug_buffer.data(), sz);
+            std::string hexdump = "";
```
lioncash: `= ""` isn't necessary. It'll initialize to empty by default.

src/core/hle/kernel/svc.cpp
```diff
@@ -395,16 +395,42 @@ struct BreakReason {
 /// Break program execution
 static void Break(u32 reason, u64 info1, u64 info2) {
     BreakReason break_reason{reason};
+    bool has_dumped_buffer{};
```
lioncash: This variable is only ever used in the lambda and nothing outside it uses it. Since it dies when it goes out of scope, this variable doesn't actually do anything.

SciresM:
User dying messages were added in 5.0.0, I believe.

PR #1649, opened 2018-11-06
gl_resource_manager: Split implementations in .cpp file.

degasus:
Those implementations are quite costly, so there is no need to inline them to the caller.
Ressource deletion is often a performance bug, so in this way, we support to add breakpoints to them.


src/video_core/renderer_opengl/gl_resource_manager.h
```diff
@@ -5,10 +5,8 @@
 #pragma once
 
 #include <utility>
-#include <glad/glad.h>
```
lioncash: wouldn't this need to remain in the header to provide the GL typedefs like GLuint, etc?

PR #1651, opened 2018-11-06
Stubbed acc:TrySelectUserWithoutInteraction

FreddyFunk:
Gets Shantae: Half- Genie Hero Ultimate Edition from splash screen to menu and ingame.

### DIFF
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -229,6 +229,13 @@ void Module::Interface::IsUserRegistrationRequestPermitted(Kernel::HLERequestCon
     rb.Push(profile_manager->CanSystemRegisterUser());
 }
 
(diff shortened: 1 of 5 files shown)


src/core/hle/service/acc/acc.cpp
```diff
@@ -229,6 +229,13 @@ void Module::Interface::IsUserRegistrationRequestPermitted(Kernel::HLERequestCon
     rb.Push(profile_manager->CanSystemRegisterUser());
 }
 
+void Module::Interface::TrySelectUserWithoutInteraction(Kernel::HLERequestContext& ctx) {
+    LOG_WARNING(Service_ACC, "(STUBBED) called");
```
ogniK5377: By the looks of it, this doesn't need to be labeled as stubbed since the implementation looks to be correct. However, you should add a TODO to check the input parameter and research it more.

PR #1654, opened 2018-11-06
gl_rasterizer: Skip VAO binding if the state is clean.

degasus:
This PR starts with the idea to skip some state lookups if the state is clean. This shall reduce the overhead per draw call a tiny bit.


src/video_core/engines/maxwell_3d.cpp
```diff
@@ -92,8 +92,16 @@ void Maxwell3D::WriteReg(u32 method, u32 value, u32 remaining_params) {
         debug_context->OnEvent(Tegra::DebugContext::Event::MaxwellCommandLoaded, nullptr);
     }
 
+    u32 old = regs.reg_array[method];
     regs.reg_array[method] = value;
 
+    if (value != old) {
```
FreddyFunk: You could avoid the definition of `old`, but `regs.reg_array[method] != value` as check and line 96 in the scope of the if statement

PR #1656, opened 2018-11-07
Ability to switch between docked and undocked mode in-game

ogniK5377:
Started implementation of the AM message queue mainly used in state getters. Added the ability to switch docked mode whilst in-game without stopping emulation. Also removed some things which shouldn't be labeled as stubs as they're implemented correctly

### DIFF
--- a/src/core/hle/service/am/am.h
+++ b/src/core/hle/service/am/am.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <memory>
+#include <queue>
 #include "core/hle/service/service.h"
 
 namespace Kernel {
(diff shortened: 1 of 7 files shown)


src/core/hle/service/am/am.h
```diff
@@ -39,6 +40,31 @@ enum SystemLanguage {
...
+
+    const Kernel::SharedPtr<Kernel::Event>& GetMesssageRecieveEvent() const;
+    const Kernel::SharedPtr<Kernel::Event>& GetOperationModeChangedEvent() const;
+    void PushMessage(AppletMessage msg);
+    AppletMessage PopMessage();
+    std::size_t GetMessageCount() const;
+    void OperationModeChanged();
+
+private:
+    std::queue<AppletMessage> messages{};
```
lioncash: `{}` isn't necessary.

src/yuzu/configuration/configure_general.cpp
```diff
@@ -45,6 +48,27 @@ void ConfigureGeneral::applyConfiguration() {
         ui->theme_combobox->itemData(ui->theme_combobox->currentIndex()).toString();
 
     Settings::values.use_cpu_jit = ui->use_cpu_jit->isChecked();
+    const bool pre_docked_mode = Settings::values.use_docked_mode;
```
lioncash: This and the code below should probably go in its own function to keep all of the behavior together, rather than having it all in a really vaguely named function (and it'd get rid of the need to encase it all in the if statement)

src/yuzu/configuration/configure_general.cpp
```diff
@@ -38,13 +41,38 @@ void ConfigureGeneral::PopulateHotkeyList(const HotkeyRegistry& registry) {
     ui->widget->Populate(registry);
 }
 
+void ConfigureGeneral::CheckIfOperationChanged(bool last_state, bool new_state) {
+    if (last_state != new_state) {
```
lioncash:
```cpp
if (last_state == new_state) {
    return;
}

File: src/yuzu/configuration/configure_general.cpp [hunk 4]
```diff
@@ -38,13 +41,38 @@ void ConfigureGeneral::PopulateHotkeyList(const HotkeyRegistry& registry) {
     ui->widget->Populate(registry);
 }
 
+void ConfigureGeneral::CheckIfOperationChanged(bool last_state, bool new_state) {
```
lioncash: This should probably be a little more specific, since what 'operation' means in the general config menu is kind of... well, general.

PR #1660, opened 2018-11-08
Map more missing opengl states

Tinob:
Add support for anisotropic filtering, lod settings, some fixes to stencil test and some tweaks to reduce state handling redundancy.


ReinUsesLisp:
You should check for anisotropic's `_EXT` extension in SDL2 and Qt. Also, it'd be nice if you can use `ARB_texture_filter_anisotropic` (4.6) when available, that way yuzu doesn't depend on non-core extensions.

Tinob:
@ReinUsesLisp Done

Tinob:
that part is only  for extensions that are requiered for the emulator to work, and none of the extensions added by me are mandatory, if they are present they will be used if not, you will get glitches but it will still work.

bunnei:
@Tinob can you rebase?

Tinob:
rebased

PR #1662, opened 2018-11-08
gl_rasterizer_cache: CopySurface optimization

FreddyFunk:
After profiling yuzu I recognized that in some szenarios quite some time is taken by allocating memory in calls of `CopySurface`. ~~So instead of reallocating memory over and over again a reserved_texture_data_memory in form of a std::vector is used.~~ We do not need to do that.


FreddyFunk:
@lioncash Addressed feedback

PR #1665, opened 2018-11-09
Implement GetClockSnapshot, ToPosixTime & ToPosixTimeWithMyRule

ogniK5377:
Needed by megaman 11.
ToPosixTime & ToPosixTimeWithMyRule  is mainly used by unity games

### DIFF
--- a/src/core/hle/service/time/time.cpp
+++ b/src/core/hle/service/time/time.cpp
@@ -15,6 +15,26 @@
 
 namespace Service::Time {
 
+void PosixToCalendar(u64 posix_time, CalendarTime& calendar_time,
+                     CalendarAdditionalInfo& additional_info, const TimeZoneRule& /*rule*/) {
+    std::time_t time(posix_time);
+    std::tm* tm = std::localtime(&time);
+    if (tm == nullptr) {
+        return;
+    }
+    calendar_time.year = tm->tm_year + 1900;
+    calendar_time.month = tm->tm_mon + 1;
+    calendar_time.day = tm->tm_mday;
+    calendar_time.hour = tm->tm_hour;
+    calendar_time.minute = tm->tm_min;
+    calendar_time.second = tm->tm_sec;
+
+    additional_info.day_of_week = tm->tm_wday;
+    additional_info.day_of_year = tm->tm_yday;
+    std::memcpy(additional_info.name.data(), "UTC", sizeof("UTC"));
+    additional_info.utc_offset = 0;
+}
+
 class ISystemClock final : public ServiceFramework<ISystemClock> {
 public:
     ISystemClock() : ServiceFramework("ISystemClock") {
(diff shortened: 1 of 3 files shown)


src/core/hle/service/time/time.cpp
```diff
@@ -15,6 +15,26 @@
 
 namespace Service::Time {
 
+void PosixToCalendar(u64 posix_time, CalendarTime& calendar_time,
```
lioncash: This can be marked `static`

src/core/hle/service/time/time.cpp
```diff
@@ -207,6 +207,55 @@ void Module::Interface::GetStandardLocalSystemClock(Kernel::HLERequestContext& c
...
+    auto unknown_u8 = rp.PopRaw<u8>();
+
+    ClockSnapshot clock_snapshot{};
+
+    const s64 time_since_epoch{std::chrono::duration_cast<std::chrono::seconds>(
+                                   std::chrono::system_clock::now().time_since_epoch())
+                                   .count()};
+    CalendarTime calendar_time{};
+    std::time_t time(time_since_epoch);
+    std::tm* tm = std::localtime(&time);
```
lioncash: This and `time` can be const.

src/core/hle/service/time/time.h
```diff
@@ -53,6 +53,23 @@ struct SystemClockContext {
...
+    s64_le network_posix_time;
+    CalendarTime system_calendar_time;
+    CalendarTime network_calendar_time;
+    CalendarAdditionalInfo system_calendar_info;
+    CalendarAdditionalInfo network_calendar_info;
+    SteadyClockTimePoint steady_clock_timepoint;
+    LocationName location_name;
+    u8 clock_auto_adjustment_enabled;
+    u8 ipc_u8;
+    INSERT_PADDING_BYTES(2);
```
lioncash: Already an existing issue, but including `common/common_funcs.h` in this header would prevent potential compilation issues in the future if headers in service.h change.

src/core/hle/service/time/time.cpp
```diff
@@ -15,6 +15,40 @@
...
+    calendar_time.minute = tm->tm_min;
+    calendar_time.second = tm->tm_sec;
+
+    additional_info.day_of_week = tm->tm_wday;
+    additional_info.day_of_year = tm->tm_yday;
+    std::memcpy(additional_info.name.data(), "UTC", sizeof("UTC"));
+    additional_info.utc_offset = 0;
+}
+
+u64 CalendarToPosix(const CalendarTime& calendar_time, const TimeZoneRule& /*rule*/) {
```
lioncash: This can just be appended with `[[maybe_unused]]`. A comment should also probably be left indicating why it's unused. This function can also be marked static.

src/core/hle/service/time/time.cpp
```diff
@@ -15,6 +15,40 @@
 
 namespace Service::Time {
 
+static void PosixToCalendar(u64 posix_time, CalendarTime& calendar_time,
+                            CalendarAdditionalInfo& additional_info, const TimeZoneRule& /*rule*/) {
```
lioncash: Ditto about [[maybe_unused]]

src/core/hle/service/time/time.cpp
```diff
@@ -15,6 +15,40 @@
...
+    const std::time_t time(posix_time);
+    const std::tm* tm = std::localtime(&time);
+    if (tm == nullptr) {
+        return;
```
lioncash:
```cpp
if (tm == nullptr) {
    calendar_time = {};
    additional_info = {};
    return;
}
```
otherwise it's possible to use this function and have uninitialized data.

PR #1667, opened 2018-11-10
am: Implement HLE software keyboard applet

DarkLordZach:
Used by Minecraft, Pokken Tournament, Stardew Valley, and some other indie games.
Prompts a dialog when the game requests the swkbd, uses the same validation as the real game.

Big thanks to @misson20000 for his help with understanding applets and hwtest (https://github.com/misson20000/swkbd-demo)

Also adds TransferMemory, as this is used by the swkbd for some of the data transfer.


src/core/hle/service/am/am.cpp
```diff
@@ -596,15 +537,20 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
     }
 
     void Start(Kernel::HLERequestContext& ctx) {
+        ASSERT(applet != nullptr);
+
+        applet->Initialize(storage_stack);
+        interactive_storage_stack.push_back(std::make_shared<IStorage>(applet->Execute()));
```
misson20000: It's called interactive storage because the applet is supposed to be able to write out to interactive storage and read back in over interactive storage while it's still alive.

Swkbd, if text check is enabled, writes what the user typed out over interactive storage out, but actually stays alive and waits for a response over interactive storage in. If the application rejected the text, swkbd will show a message box with the application's message and let the user retype it. If the application accepted it (or if text check was disabled), then swkbd will push the final text over the non-interactive output channel and exit.

Am I missing you pushing the final text over noninteractive storage somewhere? I'm surprised this works if you're not.

src/core/hle/service/am/am.cpp
```diff
@@ -596,15 +537,20 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
         IPC::ResponseBuilder rb{ctx, 2};
         rb.Push(RESULT_SUCCESS);
-
-        LOG_WARNING(Service_AM, "(STUBBED) called");
     }
 
     void PushInData(Kernel::HLERequestContext& ctx) {
         IPC::RequestParser rp{ctx};
-        storage_stack.push(rp.PopIpcInterface<AM::IStorage>());
+        storage_stack.push_back(rp.PopIpcInterface<IStorage>());
```
misson20000: I always thought that the storage channels were queues, not stacks, but I never checked. Are you sure this is a stack?

src/core/hle/service/am/am.h
```diff
@@ -128,6 +128,32 @@ class ICommonStateGetter final : public ServiceFramework<ICommonStateGetter> {
...
+    std::vector<u8> buffer;
+
+    void Open(Kernel::HLERequestContext& ctx);
+
+    friend class IStorageAccessor;
+};
+
+class IStorageAccessor final : public ServiceFramework<IStorageAccessor> {
+public:
+    explicit IStorageAccessor(IStorage& backing);
```
lioncash: The destructor should also be defaulted in the cpp file.

src/core/hle/service/am/applets/applets.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/frontend/applets/software_keyboard.h"
+#include "core/hle/service/am/applets/applets.h"
+
+namespace Service::AM::Applets {
+
+std::shared_ptr<Frontend::SoftwareKeyboardApplet> software_keyboard =
```
lioncash: No global state. It's been stated frequently enough to you that it's actually getting annoying to have to keep pointing this out. Why is this info not sticking?

src/core/hle/service/am/applets/applets.h
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Service::AM {
+
+class IStorage;
+
+namespace Applets {
+
+class Applet {
+public:
+    virtual void Initialize(std::vector<std::shared_ptr<IStorage>> storage);
```
lioncash: This should have a virtual destructor. The constructor and destructor should also both be defaulted in the cpp file.

src/core/hle/service/am/applets/software_keyboard.h
```diff
@@ -0,0 +1,57 @@
...
+    u32_le user_dictionary_offset;
+    u32_le user_dictionary_size;
+    bool text_check;
+    u64_le text_check_callback;
+};
+static_assert(sizeof(KeyboardConfig) == 0x3E0, "KeyboardConfig has incorrect size.");
+
+class SoftwareKeyboard final : public Applet {
+public:
+    void Initialize(std::vector<std::shared_ptr<IStorage>> storage) override;
```
lioncash: Constructor and destructor should also be defaulted in the cpp file.

src/yuzu/applets/software_keyboard.h
```diff
@@ -0,0 +1,62 @@
...
+
+class QDialogButtonBox;
+class QLabel;
+class QLineEdit;
+class QVBoxLayout;
+class QtSoftwareKeyboard;
+
+class QtSoftwareKeyboardValidator final : public QValidator {
+public:
+    explicit QtSoftwareKeyboardValidator(Frontend::SoftwareKeyboardApplet::Parameters parameters);
```
lioncash: Destructor should also be defaulted in the cpp file.

src/yuzu/applets/software_keyboard.h
```diff
@@ -0,0 +1,62 @@
...
+    State validate(QString&, int&) const override;
```
lioncash: Why no parameter names?

src/yuzu/applets/software_keyboard.h
```diff
@@ -0,0 +1,62 @@
...
+
+private:
+    Frontend::SoftwareKeyboardApplet::Parameters parameters;
+};
+
+class QtSoftwareKeyboardDialog final : public QDialog {
+    Q_OBJECT
+
+public:
+    QtSoftwareKeyboardDialog(QWidget* parent,
```
lioncash: Destructor should be defaulted in the cpp file.

src/yuzu/applets/software_keyboard.h
```diff
@@ -0,0 +1,62 @@
...
+
+    friend class QtSoftwareKeyboard;
+};
+
+class QtSoftwareKeyboard final : public QObject, public Frontend::SoftwareKeyboardApplet {
+public:
+    explicit QtSoftwareKeyboard(QWidget& parent);
+    bool GetText(Parameters parameters, std::u16string& text) override;
+
+    ~QtSoftwareKeyboard() {
```
lioncash: This should be defaulted in the cpp file. Also the ordering in the class is:

```cpp
Constructor
Destructor

Member functions
```

src/yuzu/applets/software_keyboard.h
```diff
@@ -0,0 +1,62 @@
...
+        UNREACHABLE();
```
lioncash: This is extremely questionable. Especially because Qt needs to destroy its UI objects when shutting the application down.

src/yuzu/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+void QtSoftwareKeyboardDialog::Submit() {
+    ok = true;
+    text = line_edit->text().toStdU16String();
+    accept();
+}
+
+void QtSoftwareKeyboardDialog::Reject() {
+    ok = false;
+    text = Common::UTF8ToUTF16("");
```
lioncash: There's no need to construct an empty string here. `text.clear();` is sufficient. The string_util header can also be removed.

src/yuzu/applets/software_keyboard.h
```diff
@@ -0,0 +1,62 @@
...
+    QDialogButtonBox* buttons;
+    QLabel* header_label;
+    QLabel* sub_label;
+    QLabel* guide_label;
+    QLineEdit* line_edit;
+    QVBoxLayout* layout;
+
+    Frontend::SoftwareKeyboardApplet::Parameters parameters;
+
+    friend class QtSoftwareKeyboard;
```
lioncash: There's no need to use friend here. Just provide `GetText()` and `IsSubmitted()` to get the state of the `text` and `ok` members.

src/yuzu/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+    connect(buttons, &QDialogButtonBox::accepted, this, &QtSoftwareKeyboardDialog::Submit);
+    connect(buttons, &QDialogButtonBox::rejected, this, &QtSoftwareKeyboardDialog::Reject);
+    layout->addWidget(header_label);
+    layout->addWidget(sub_label);
+    layout->addWidget(guide_label);
+    layout->addWidget(line_edit);
+    layout->addWidget(buttons);
+    setLayout(layout);
+    setWindowTitle("Software Keyboard");
```
lioncash: This should be a translatable string.

src/core/hle/service/am/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    std::u16string text;
+    const auto success = frontend->GetText(parameters, text);
+
+    std::vector<u8> output(SWKBD_OUTPUT_BUFFER_SIZE);
+
+    if (success) {
+        output[0] = 1;
+        std::memcpy(output.data() + 4, text.data(),
+                    std::min<std::size_t>(text.size() * 2, SWKBD_OUTPUT_BUFFER_SIZE - 4));
```
lioncash: `std::size_t` can be omitted here.

src/yuzu/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,108 @@
...
+        return Invalid;
+    if (parameters.disable_space && input.contains(' '))
+        return Invalid;
+    if (parameters.disable_address && input.contains('@'))
+        return Invalid;
+    if (parameters.disable_percent && input.contains('%'))
+        return Invalid;
+    if (parameters.disable_slash && (input.contains('/') || input.contains('\\')))
+        return Invalid;
+    if (parameters.disable_number &&
```
lioncash: This conditional should be braced since the condition travels more than one line.

src/yuzu/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,108 @@
...
+        std::any_of(input.begin(), input.end(), [](QChar c) { return c.isDigit(); }))
+        return Invalid;
+
+    if (parameters.disable_download_code &&
```
lioncash: Ditto

src/core/hle/service/am/am.cpp
```diff
@@ -661,6 +728,31 @@ void ILibraryAppletCreator::CreateStorage(Kernel::HLERequestContext& ctx) {
...
+        Core::System::GetInstance().CurrentProcess()->GetHandleTable().Get<Kernel::SharedMemory>(
+            handle);
+
+    if (shared_mem == nullptr) {
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(ResultCode(-1));
+        return;
+    }
+
+    std::vector<u8> memory(shared_mem->size);
```
lioncash:
```cpp
const auto begin = shared_mem->backing_block.begin() + shared_mem->backing_block_offset;
const auto end = begin + shared_mem->size;
std::vector<u8> memory{begin, end};
```

will do the same thing without the need to initialize a zeroed out container first.

src/core/hle/service/am/am.cpp
```diff
@@ -615,36 +561,157 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
+    LOG_DEBUG(Service_AM, "called");
+}
+
+void IStorageAccessor::Write(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+
+    const u64 offset{rp.Pop<u64>()};
+    const std::vector<u8> data{ctx.ReadBuffer()};
+
+    const auto size = std::min<std::size_t>(data.size(), backing.buffer.size() - offset);
```
lioncash: std::size_t can be omitted here.

src/yuzu/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,108 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <locale>
```
lioncash: This doesn't seem to be used anywhere.

src/yuzu/applets/software_keyboard.h
```diff
@@ -0,0 +1,62 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
```
lioncash: Newline after pragma once

src/core/frontend/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,18 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/logging/backend.h"
```
lioncash: Nothing from logging is used here.

src/core/frontend/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include "common/string_util.h"
+#include "core/frontend/applets/software_keyboard.h"
+
+namespace Frontend {
+bool DefaultSoftwareKeyboardApplet::GetText(Parameters parameters, std::u16string& text) {
+    if (parameters.initial_text.empty())
+        text = Common::UTF8ToUTF16("yuzu");
```
lioncash: This can just be:

```cpp
text = u"yuzu";
```

there's no need to use a conversion function for a literal. The include for string_util can also be removed with this.

src/core/hle/service/am/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,71 @@
...
+    const auto parameters = ConvertToFrontendParameters(config, initial_text);
+
+    std::u16string text;
+    const auto success = frontend->GetText(parameters, text);
+
+    std::vector<u8> output(SWKBD_OUTPUT_BUFFER_SIZE);
+
+    if (success) {
+        output[0] = 1;
+        std::memcpy(output.data() + 4, text.data(),
```
lioncash: `<cstring>` should be included.

src/yuzu/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,108 @@
...
+    if (parameters.disable_space && input.contains(' '))
+        return Invalid;
+    if (parameters.disable_address && input.contains('@'))
+        return Invalid;
+    if (parameters.disable_percent && input.contains('%'))
+        return Invalid;
+    if (parameters.disable_slash && (input.contains('/') || input.contains('\\')))
+        return Invalid;
+    if (parameters.disable_number &&
+        std::any_of(input.begin(), input.end(), [](QChar c) { return c.isDigit(); }))
```
lioncash: `<algorithm>` should be included.

ogniK5377:
Reminder from discord dms,
We seem to get an unmapped read exception at `QtSoftwareKeyboard::GetText`

src/core/hle/service/am/am.cpp
```diff
@@ -542,9 +544,16 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
         state_changed_event->Signal();
-        pop_interactive_out_data_event->Signal();
+
+        if (applet->TransactionComplete()) {
+            storage_stack.push_back(data);
+            pop_out_data_event->Signal();
+        } else {
+            interactive_storage_stack.push_back(data);
+            pop_interactive_out_data_event->Signal();
+        }
```
misson20000: I don't really like this model of having an `Execute` method that's only called when the main application interacts with the applet, and that blocks the main application until the applet is done doing whatever it needs to do. It does make sense though if the intent is to keep the applets synchronous so that behavior is deterministic.

I would recommend you change it so that `applet->Execute()` doesn't return an `IStorage`, and instead, if it wants to push interactive out storage or regular out storage, it can do that by calling some methods, explicitly specifying which channel the output storage goes to rather than having which channel the storage goes to be determined by whether the applet's transaction is complete or not. That way, the applet can also push multiple storages if it needs to.

src/yuzu/main.h
```diff
@@ -95,6 +99,11 @@ class GMainWindow : public QMainWindow {
     // Signal that tells widgets to update icons to use the current theme
     void UpdateThemedIcons();
 
+public slots:
+    bool SoftwareKeyboardGetText(const Core::Frontend::SoftwareKeyboardParameters& parameters,
```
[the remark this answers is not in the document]
DarkLordZach: https://github.com/yuzu-emu/yuzu/pull/1667/files#diff-41128cb4d8543fc3dc355acefa28cd61R210 ?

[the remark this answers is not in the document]
lioncash: Ah, weird. Ctrl+F failed me during the review.

Hexagon12:
Untagged from Canary due to major issues in games.

src/core/hle/service/am/am.cpp
```diff
@@ -618,36 +558,70 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
+        pop_out_data_event = Kernel::Event::Create(kernel, Kernel::ResetType::OneShot,
+                                                   "ILibraryAppletAccessor:PopDataOutEvent");
+        pop_interactive_out_data_event =
+            Kernel::Event::Create(kernel, Kernel::ResetType::OneShot,
+                                  "ILibraryAppletAccessor:PopInteractiveDataOutEvent");
     }
 
 private:
+    void AppletStorageProxyOutData(IStorage storage) {
+        storage_stack.push(std::make_shared<IStorage>(storage));
```
misson20000: "storage stack" is still a strange name for this. They're officially called `nn::am::service::StorageChannel`.

src/core/hle/service/am/am.cpp
```diff
@@ -618,36 +558,70 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
+        ASSERT(applet != nullptr);
+
+        applet->Initialize(storage_stack);
+        while (!storage_stack.empty())
+            storage_stack.pop();
+        while (!interactive_storage_stack.empty())
+            interactive_storage_stack.pop();
+        applet->Execute([this](IStorage storage) { AppletStorageProxyOutData(storage); },
+                        [this](IStorage storage) { AppletStorageProxyOutInteractiveData(storage); },
+                        [this] { state_changed_event->Signal(); });
```
misson20000: What is this callback pattern? Why don't the storage channels + event and all belong to the applet instead of the applet accessor? Perhaps have a class to hold channels + event that both the `ILibraryAppletAccessor` and the applet implementation have shared_ptrs to.

src/core/hle/service/am/am.cpp
```diff
@@ -618,36 +558,70 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
         rb.Push(RESULT_SUCCESS);
 
-        LOG_WARNING(Service_AM, "(STUBBED) called");
+        LOG_DEBUG(Service_AM, "called");
     }
 
     void PushInData(Kernel::HLERequestContext& ctx) {
         IPC::RequestParser rp{ctx};
-        storage_stack.push(rp.PopIpcInterface<AM::IStorage>());
+        storage_stack.push(rp.PopIpcInterface<IStorage>());
```
misson20000: in data and out data are separate storage channels

src/core/hle/service/am/am.cpp
```diff
@@ -657,37 +631,185 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
-    std::stack<std::shared_ptr<AM::IStorage>> storage_stack;
+    void PushInteractiveInData(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        interactive_storage_stack.push(rp.PopIpcInterface<IStorage>());
+
+        ASSERT(applet->IsInitialized());
+        applet->ReceiveInteractiveData(interactive_storage_stack.back());
+        applet->Execute([this](IStorage storage) { AppletStorageProxyOutData(storage); },
+                        [this](IStorage storage) { AppletStorageProxyOutInteractiveData(storage); },
+                        [this] { state_changed_event->Signal(); });
```
misson20000: Same thing here. This callback pattern is strange to me.

src/core/hle/service/am/am.cpp
```diff
@@ -657,37 +631,185 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
+        rb.Push(RESULT_SUCCESS);
+
+        LOG_DEBUG(Service_AM, "called");
+    }
+
+    void PopInteractiveOutData(Kernel::HLERequestContext& ctx) {
+        IPC::ResponseBuilder rb{ctx, 2, 0, 1};
+
+        if (interactive_storage_stack.empty()) {
+            rb.Push(ResultCode(-1));
```
misson20000: Result code for this is `0x480`, and sdk code does check for this result code.

src/core/hle/service/am/am.cpp
```diff
@@ -657,37 +631,185 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
+    LOG_DEBUG(Service_AM, "called");
+}
+
+void IStorageAccessor::Write(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+
+    const u64 offset{rp.Pop<u64>()};
+    const std::vector<u8> data{ctx.ReadBuffer()};
+
+    const auto size = std::min<std::size_t>(data.size(), backing.buffer.size() - offset);
```
misson20000: Out-of-bounds write is an error (code `0x3ee80`). Make sure to check that offset is not negative.

src/core/hle/service/am/am.cpp
```diff
@@ -657,37 +631,185 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
+    LOG_DEBUG(Service_AM, "called, offset={}", offset);
+}
+
+void IStorageAccessor::Read(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+
+    const u64 offset{rp.Pop<u64>()};
+    std::size_t size{ctx.GetWriteBufferSize()};
+
+    size = std::min<std::size_t>(size, backing.buffer.size() - offset);
```
misson20000: Out-of-bounds read is also an error (code `0x3ee80`), check for negative offset, etc.

src/core/hle/service/am/am.cpp
```diff
@@ -657,37 +631,185 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
+    std::queue<std::shared_ptr<IStorage>> storage_stack;
+    std::queue<std::shared_ptr<IStorage>> interactive_storage_stack;
     Kernel::SharedPtr<Kernel::Event> state_changed_event;
+    Kernel::SharedPtr<Kernel::Event> pop_out_data_event;
+    Kernel::SharedPtr<Kernel::Event> pop_interactive_out_data_event;
 };
 
+void IStorage::Open(Kernel::HLERequestContext& ctx) {
+    IPC::ResponseBuilder rb{ctx, 2, 0, 1};
+
```
misson20000: Not sure how important this is for your purposes, but perhaps sure that this is not a transfer memory storage. Error code is `0x3fe80`.

src/core/hle/service/am/am.cpp
```diff
@@ -500,194 +529,265 @@ void ICommonStateGetter::GetPerformanceMode(Kernel::HLERequestContext& ctx) {
...
-        rb.PushIpcInterface<AM::IStorage>(std::move(storage_stack.top()));
+void IStorageAccessor::Read(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
 
-        storage_stack.pop();
+    const u64 offset{rp.Pop<u64>()};
+    const std::size_t size{ctx.GetWriteBufferSize()};
 
-        LOG_DEBUG(Service_AM, "called");
+    if (size > backing.buffer.size() - offset) {
```
misson20000:
```suggestion
    if (size > backing.buffer.size() - offset || offset < 0) {
```

src/core/hle/service/am/applets/applets.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "core/core.h"
+#include "core/frontend/applets/software_keyboard.h"
+#include "core/hle/kernel/event.h"
+#include "core/hle/kernel/server_port.h"
+#include "core/hle/service/am/am.h"
+#include "core/hle/service/am/applets/applets.h"
+
+namespace Service::AM::Applets {
+
+AppletDataBroker::AppletDataBroker() {
```
misson20000: I like this much better

src/core/hle/service/am/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        complete = !config.text_check;
+        final_data = output_main;
+
+        if (complete) {
+            broker->PushNormalDataFromApplet(IStorage{output_main});
+        } else {
+            broker->PushNormalDataFromApplet(IStorage{output_main});
+            broker->PushInteractiveDataFromApplet(IStorage{output_sub});
```
misson20000: Pretty sure this should only push interactive, not normal

src/core/hle/service/am/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+        if (config.utf_8) {
+            const u64 size = text->size() + 8;
+            const auto new_text = Common::UTF16ToUTF8(*text);
+
+            std::memcpy(output_sub.data(), &size, sizeof(u64));
+            std::memcpy(output_sub.data() + 8, new_text.data(),
+                        std::min(new_text.size(), SWKBD_OUTPUT_BUFFER_SIZE - 8));
+
+            output_main[0] = config.text_check;
```
misson20000:
```suggestion
            output_main[0] = 0; // 0: success, 1: cancelled
```
since `output_main` is only used when text check has completed OK

src/core/hle/service/am/applets/software_keyboard.cpp
```diff
@@ -0,0 +1,164 @@
...
+            std::memcpy(output_main.data() + 4, new_text.data(),
+                        std::min(new_text.size(), SWKBD_OUTPUT_BUFFER_SIZE - 4));
+        } else {
+            const u64 size = text->size() * 2 + 8;
+            std::memcpy(output_sub.data(), &size, sizeof(u64));
+            std::memcpy(output_sub.data() + 8, text->data(),
+                        std::min(text->size() * 2, SWKBD_OUTPUT_BUFFER_SIZE - 8));
+
+            output_main[0] = config.text_check;
```
misson20000:
```suggestion
            output_main[0] = 0;
```
same here

PR #1670, opened 2018-11-12
csrng: Add config option to set RNG seed

DarkLordZach:
Two-fold:
- Switch to `std::mt19937` for rng bytes in `csrng` service. This allows determinism given the same seed.
- Add UI options to use a random seed or a set seed. If a seed is not set, `csrng` will use `std::time(nullptr)` to get a seed.

rng1

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -34,6 +34,7 @@
 #include "core/hle/lock.h"
 #include "core/hle/result.h"
 #include "core/hle/service/service.h"
+#include "core/settings.h"
 
 namespace Kernel {
 namespace {
@@ -558,7 +559,7 @@ static ResultCode GetInfo(u64* result, u64 info_id, u64 handle, u64 info_sub_id)
         *result = 0;
         break;
     case GetInfoType::RandomEntropy:
-        *result = 0;
+        *result = Settings::values.rng_seed.value_or(0);
         break;
     case GetInfoType::ASLRRegionBaseAddr:
         *result = vm_manager.GetASLRRegionBaseAddress();
--- a/src/yuzu/configuration/configure_system.cpp
+++ b/src/yuzu/configuration/configure_system.cpp
@@ -140,7 +140,7 @@ ConfigureSystem::ConfigureSystem(QWidget* parent)
     connect(ui->rng_seed_checkbox, &QCheckBox::stateChanged, this, [this](bool checked) {
         ui->rng_seed_edit->setEnabled(checked);
         if (!checked)
-            ui->rng_seed_edit->setText("0000000000000000");
+            ui->rng_seed_edit->setText(QStringLiteral("0000000000000000"));
     });
 
     scene = new QGraphicsScene;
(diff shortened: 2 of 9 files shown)


B3n30:
I remember that the std rng generators are broken on some platforms. You might consider using cryptopp instead

src/yuzu/configuration/configure_system.cpp
```diff
@@ -155,6 +161,11 @@ void ConfigureSystem::setConfiguration() {
 
     PopulateUserList();
     UpdateCurrentUser();
+
+    ui->rng_seed_checkbox->setChecked(Settings::values.rng_seed.has_value());
+    ui->rng_seed_edit->setEnabled(Settings::values.rng_seed.has_value());
+    ui->rng_seed_edit->setText(
+        QString::fromStdString(fmt::format("{:016X}", Settings::values.rng_seed.value_or(0))));
```
lioncash: You don't need to convert between string types here.

```cpp
const QString rng_seed = QString::number(Settings::values.rng_seed.value_or(0), 16).toUpper();
ui->rng_seed_edit->setText(rng_seed);
```

or if the padding is absolutely necessary:

```cpp
const QString rng_seed = QString("%1").arg(Settings::values.rng_seed.value_or(0), 16, QLatin1Char{'0'}).toUpper();
ui->rng_seed_edit->setText(rng_seed);
```

lioncash:
@B3n30 There's no need for that. It's just a basic RNG value. As long as the values are different they're fine. Also mersenne twister is fine on all platforms we support. The only device that has no guarantee of being random is std::random_device (if the platform has no means of providing true random values). Even then, that'd be overkill for this scenario.

src/core/hle/kernel/svc.cpp
```diff
@@ -558,7 +559,7 @@ static ResultCode GetInfo(u64* result, u64 info_id, u64 handle, u64 info_sub_id)
         *result = 0;
         break;
     case GetInfoType::RandomEntropy:
-        *result = 0;
+        *result = Settings::values.rng_seed.value_or(0);
```
lioncash: This is incorrect. It's not retrieving a seed here. It's retrieving a random value from a buffer of values generated from the seed.

i.e.

```cpp
struct Process {
public:
    static constexpr std::size_t random_data_size = 4;

    // ...
    u64 GetRandomValue(std::size_t index) const;

private:
    // ...

    std::array<u64, random_data_size> random_data{};

    // ...
};

// process.cpp

void Process::Initialize(...) {
    std::mt19937 generator(seed);
    std::generate(random_data.begin(), random_data.end(), generator);
}


u64 Process::GetRandomValue(std::size_t index) const {
    ASSERT(index < 4);
    return random_data[index];
} 

// svc.cpp

ResultCode svcGetInfo(u64* result, u64 info_id, u64 handle, u64 info_sub_id) {

    switch (type) {
    // ...

    case GetInfoType::RandomEntropy:
        if (handle != 0) {
            return ERR_INVALID_HANDLE
        }

        if (info_sub_id >= Process::random_data_size) {
            return ERR_INVALID_COMBINATION_KERNEL;
        }

        *result = current_process.GetRandomNumber(info_sub_id);
        return RESULT_SUCCESS;
    }

    // ...
};

PR #1671, opened 2018-11-12
vi: Implement TransactParcel for Disconnect and DetachBuffer

DarkLordZach:
Used by homebrew on exit. According to switchbrew, returns an empty response parcel with one zero in it.

### DIFF
--- a/src/core/hle/service/vi/vi.cpp
+++ b/src/core/hle/service/vi/vi.cpp
@@ -237,6 +237,22 @@ private:
     Data data{};
 };
 
+/// Represents a parcel containing one int '0' as its data
+/// Used by DetachBuffer and Disconnect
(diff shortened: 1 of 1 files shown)


src/core/hle/service/vi/vi.cpp
```diff
@@ -237,6 +237,25 @@ class IGBPConnectResponseParcel : public Parcel {
     Data data{};
 };
 
+/// Represents a parcel containing one int '0' as its data
+/// Used by DetachBuffer and Disconnect
+class IGBPEmptyResponseParcel : public Parcel {
+public:
+    ~IGBPEmptyResponseParcel() override = default;
```
lioncash: This doesn't need to be provided, since it's the basic default constructor.

PR #1675, opened 2018-11-13
Fixed priority switching edge case for handheld

ogniK5377:
We accidently used controller index instead of npad id


src/core/hle/service/hid/controllers/npad.h
```diff
@@ -112,6 +112,7 @@ class Controller_NPad final : public ControllerBase {
     void DisconnectNPad(u32 npad_id);
     LedPattern GetLedPattern(u32 npad_id);
     void SetVibrationEnabled(bool can_vibrate);
+    static std::size_t NPadIdToIndex(u32 npad_id);
```
lioncash: This can just be an internally static function, since it doesn't need anything from the class itself.

PR #1679, opened 2018-11-13
svc: Use proper random entropy generation algorithm

DarkLordZach:
Addresses outlying concerns with #1670.

- Use 32-bit random seed to match kernel behavior
- Use random generator to generate per-core values for svcGetInfo RandomEntropy

### DIFF
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -17,6 +17,7 @@
 #include "core/hle/kernel/thread.h"
 #include "core/hle/kernel/vm_manager.h"
 #include "core/memory.h"
+#include "core/settings.h"
 
 namespace Kernel {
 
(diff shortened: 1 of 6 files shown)


src/core/hle/kernel/process.cpp
```diff
@@ -35,6 +36,11 @@ SharedPtr<Process> Process::Create(KernelCore& kernel, std::string&& name) {
     process->process_id = kernel.CreateNewProcessID();
     process->svc_access_mask.set();
 
+    std::mt19937 rng(Settings::values.rng_seed.value_or(0));
+    std::uniform_int_distribution<u64> distribution;
+    std::generate(process->random_entropy.begin(), process->random_entropy.end(),
```
lioncash:
```cpp
std::generate(process->random_entropy.begin(), process->random_entropy.end(),
              [&] { return distribution(rng); });

File: src/core/hle/kernel/process.h [hunk 2]
```diff
@@ -24,6 +25,8 @@ class ProgramMetadata;
 
 namespace Kernel {
 
+constexpr std::size_t RANDOM_ENTROPY_SIZE = 4;
```
lioncash: This should be a static constexpr member of the KProcess class, since the size of the data only applies to the process. So `KProcess::RANDOM_ENTROPY_SIZE` makes more sense when referring to it.

PR #1682, opened 2018-11-13
hle/audren_u: Implement Get/SetRenderingTimeLimit

lioncash:
These appear to be a basic getter and setter pair based off RE, so these are fairly trivial to implement and get out of the way.

### DIFF
--- a/src/core/hle/service/audio/audren_u.cpp
+++ b/src/core/hle/service/audio/audren_u.cpp
@@ -32,8 +32,8 @@ public:
             {5, &IAudioRenderer::Start, "Start"},
             {6, &IAudioRenderer::Stop, "Stop"},
             {7, &IAudioRenderer::QuerySystemEvent, "QuerySystemEvent"},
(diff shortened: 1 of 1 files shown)


src/core/hle/service/audio/audren_u.cpp
```diff
@@ -110,8 +110,29 @@ class IAudioRenderer final : public ServiceFramework<IAudioRenderer> {
...
+
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(RESULT_SUCCESS);
+
+        LOG_DEBUG(Service_Audio, "called. rendering_time_limit_percent={}",
+                  rendering_time_limit_percent);
+    }
+
+    void GetRenderingTimeLimit(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Service_Audio, "called.");
```
Hexagon12: `LOG_DEBUG(Service_Audio, "called");`

PR #1689, opened 2018-11-14
hid/npad: Add missing break in switch statement within Controller_NPad::OnUpdate

lioncash:


lioncash:
Failure isn't related to this code directly.

PR #1690, opened 2018-11-14
nfp: Correct erroneous sizeof expression within GetTagInfo()

lioncash:
The previous expression would copy sizeof(size_t) amount of bytes (8 on a 64-bit platform) rather than the full 10 bytes comprising the uuid member.

### DIFF
--- a/src/core/hle/service/nfp/nfp.cpp
+++ b/src/core/hle/service/nfp/nfp.cpp
@@ -212,7 +212,7 @@ private:
         IPC::ResponseBuilder rb{ctx, 2};
         auto amiibo = nfp_interface.GetAmiiboBuffer();
         TagInfo tag_info{};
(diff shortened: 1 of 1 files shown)


src/core/hle/service/nfp/nfp.cpp
```diff
@@ -212,7 +212,7 @@ class IUser final : public ServiceFramework<IUser> {
         IPC::ResponseBuilder rb{ctx, 2};
         auto amiibo = nfp_interface.GetAmiiboBuffer();
         TagInfo tag_info{};
-        std::memcpy(tag_info.uuid.data(), amiibo.uuid.data(), sizeof(tag_info.uuid.size()));
+        std::memcpy(tag_info.uuid.data(), amiibo.uuid.data(), sizeof(tag_info.uuid));
```
VPeruS: Why not `tag_info.uuid = amiibo.uuid;`?

lioncash: Good catch. Thanks :>

PR #1693, opened 2018-11-14
Missing ogl states

Tinob:
Add support for more missing functionality in glstate

### DIFF
--- a/src/video_core/engines/maxwell_3d.h
+++ b/src/video_core/engines/maxwell_3d.h
@@ -706,7 +706,9 @@ public:
                 u32 stencil_front_func_mask;
                 u32 stencil_front_mask;
 
-                INSERT_PADDING_WORDS(0x3);
+                INSERT_PADDING_WORDS(0x2);
+
+                u32 frag_color_clamp;
 
                 union {
                     BitField<4, 1, u32> triangle_rast_flip;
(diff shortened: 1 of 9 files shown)


src/video_core/engines/maxwell_3d.h
```diff
@@ -553,7 +553,7 @@ class Maxwell3D final {
                     };
                     float depth_range_near;
                     float depth_range_far;
-                } viewport[NumViewports];
+                } viewports[NumViewports];
```
lioncash: std::array

src/video_core/engines/maxwell_3d.h
```diff
@@ -581,9 +581,10 @@ class Maxwell3D final {
                         BitField<0, 16, u32> min_y;
                         BitField<16, 16, u32> max_y;
                     };
-                } scissor_test;
+                    u32 fill;
+                } scissor_test[NumViewports];
```
lioncash: ditto

src/video_core/engines/maxwell_3d.h
Tinob: that was tyhe original code, but caused compilation error on gcc so changed it in 13e5e15 to []

wwylele: @Tinob You need to move the definition of `ViewportTransform` and `ViewPort` out of the anonymous union/struct to make it work.

Tinob: done

PR #1699, opened 2018-11-15
csrng: Use random integer distribution instead of raw engine

DarkLordZach:
Prevents returning the same value every single call.

### DIFF
--- a/src/core/hle/service/spl/module.cpp
+++ b/src/core/hle/service/spl/module.cpp
@@ -28,8 +28,9 @@ void Module::Interface::GetRandomBytes(Kernel::HLERequestContext& ctx) {
 
     std::size_t size = ctx.GetWriteBufferSize();
 
+    std::uniform_int_distribution<u16> distribution(0, std::numeric_limits<u8>::max());
     std::vector<u8> data(size);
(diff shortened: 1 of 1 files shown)


src/core/hle/service/spl/module.h
```diff
@@ -23,6 +23,7 @@ class Module final {
 
     private:
         std::mt19937 rng;
+        std::uniform_int_distribution<u16> distribution;
```
lioncash: This can just be declared in the function, since it's not a stateful object.

src/core/hle/service/spl/module.cpp
```diff
@@ -29,7 +30,7 @@ void Module::Interface::GetRandomBytes(Kernel::HLERequestContext& ctx) {
     std::size_t size = ctx.GetWriteBufferSize();
 
     std::vector<u8> data(size);
-    std::generate(data.begin(), data.end(), rng);
+    std::generate(data.begin(), data.end(), [&]() { return distribution(rng); });
```
lioncash: You need to cast to u8 here. `()` is also not necessary on lambda functions with no arguments.

PR #1704, opened 2018-11-16
file_sys: Implement open source system archives

DarkLordZach:
This adds:
- A framework for synthesizing system archives and returning them to the game, using a table of functions similar to our HLE impl.
- A fallback to this framework if the user doesn't have the correct archive in NAND
- An implementation of NgWord (806) under this.

This works vastly different from citra's impl, which used externally generated romfses packed into cpp files. This stores the actual contents of the files within the archive (i.e. version.dat for NgWord) and then synthesizes the directory structure and romfs on the fly, resulting in cleaner more updatable code.

Tested in Ultra Street Fighter II, which performs exactly the same with or without it.

### DIFF
--- a/src/core/file_sys/system_archive/ng_word.cpp
+++ b/src/core/file_sys/system_archive/ng_word.cpp
@@ -1,0 +1,42 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <fmt/format.h>
+#include "common/common_types.h"
+#include "core/file_sys/system_archive/ng_word.h"
+#include "core/file_sys/vfs_vector.h"
+
+namespace FileSys::SystemArchive {
+
+namespace NgWord1Data {
+
+constexpr std::size_t NUMBER_WORD_TXT_FILES = 0x10;
+
+// Should this archive replacement mysteriously not work on a future game, consider updating.
+constexpr std::array<u8, 4> VERSION_DAT{0x0, 0x0, 0x0, 0x19}; // 5.1.0 System Version
+
+constexpr std::array<u8, 30> WORD_TXT{
+    0xFE, 0xFF, 0x00, 0x5E, 0x00, 0x76, 0x00, 0x65, 0x00, 0x72, 0x00, 0x79, 0x00, 0x62, 0x00,
+    0x61, 0x00, 0x64, 0x00, 0x77, 0x00, 0x6F, 0x00, 0x72, 0x00, 0x64, 0x00, 0x24, 0x00, 0x0A,
+}; // "^verybadword$" in UTF-16
+
+} // namespace NgWord1Data
+
+VirtualDir NgWord1() {
+    std::vector<VirtualFile> files(NgWord1Data::NUMBER_WORD_TXT_FILES);
+
+    for (std::size_t i = 0; i < NgWord1Data::NUMBER_WORD_TXT_FILES; ++i) {
+        files[i] = std::make_shared<ArrayVfsFile<NgWord1Data::WORD_TXT.size()>>(
+            NgWord1Data::WORD_TXT, fmt::format("{}.txt", i));
+    }
+
+    files.push_back(std::make_shared<ArrayVfsFile<NgWord1Data::WORD_TXT.size()>>(
+        NgWord1Data::WORD_TXT, "common.txt"));
+    files.push_back(std::make_shared<ArrayVfsFile<NgWord1Data::VERSION_DAT.size()>>(
+        NgWord1Data::VERSION_DAT, "version.dat"));
+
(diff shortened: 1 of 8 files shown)


src/core/file_sys/system_archive/ng_word.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "common/common_types.h"
+#include "core/file_sys/system_archive/ng_word.h"
+#include "core/file_sys/vfs_vector.h"
+#include "fmt/format.h"
+
+namespace FileSys::SystemArchive {
+
+namespace NgWord1Data {
+
+constexpr std::array<u8, 4> VERSION_DAT{0x0, 0x0, 0x0, 0x19}; // 5.1.0 System Version
```
lioncash: If this needs to be updated in the future, a comment should be provided.

src/core/file_sys/system_archive/ng_word.cpp
```diff
@@ -0,0 +1,39 @@
...
+    0xFE, 0xFF, 0x00, 0x5E, 0x00, 0x76, 0x00, 0x65, 0x00, 0x72, 0x00, 0x79, 0x00, 0x62, 0x00,
+    0x61, 0x00, 0x64, 0x00, 0x77, 0x00, 0x6F, 0x00, 0x72, 0x00, 0x64, 0x00, 0x24, 0x00, 0x0A,
+}; // "^verybadword$" in UTF-16
+
+} // namespace NgWord1Data
+
+VirtualDir NgWord1() {
+    std::vector<VirtualFile> files;
+
+    for (std::size_t i = 0; i < 16; ++i) {
```
lioncash: magic number. The vector can also be `reserve()`'d ahead of time.

src/core/file_sys/vfs_vector.h
```diff
@@ -8,6 +8,58 @@
 
 namespace FileSys {
 
+// An implementation of VfsFile that is backed by a statically-sized array
+template <std::size_t size>
+class ArrayVfsFile : public VfsFile {
+public:
+    ArrayVfsFile(std::array<u8, size> data, std::string name = "", VirtualDir parent = nullptr)
+        : data(std::move(data)), name(std::move(name)), parent(std::move(parent)) {}
```
lioncash: moving `data` in this case will result in a copy all the time, so it doesn't need to be provided.

legend800:
Once merged, is there still value for having your own system archives? Or can they be removed to save disc space (mine's ~250mb).

src/core/file_sys/system_archive/ng_word.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+constexpr std::array<u8, 30> WORD_TXT{
+    0xFE, 0xFF, 0x00, 0x5E, 0x00, 0x76, 0x00, 0x65, 0x00, 0x72, 0x00, 0x79, 0x00, 0x62, 0x00,
+    0x61, 0x00, 0x64, 0x00, 0x77, 0x00, 0x6F, 0x00, 0x72, 0x00, 0x64, 0x00, 0x24, 0x00, 0x0A,
+}; // "^verybadword$" in UTF-16
+
+} // namespace NgWord1Data
+
+VirtualDir NgWord1() {
+    std::vector<VirtualFile> files;
```
ogniK5377: `std::vector<VirtualFile> files(NgWord1Data::NUMBER_WORD_TXT_FILES);` might work here and removing the push_pack

src/core/file_sys/system_archive/system_archive.cpp
```diff
@@ -0,0 +1,91 @@
...
+    {0x0100000000000821, "PlatformConfigHoag", nullptr},
+    {0x0100000000000822, "ControllerFirmware", nullptr},
+    {0x0100000000000823, "NgWord2", nullptr},
+    {0x0100000000000824, "PlatformConfigIcosaMariko", nullptr},
+    {0x0100000000000825, "ApplicationBlackList", nullptr},
+    {0x0100000000000826, "RebootlessSystemUpdateVersion", nullptr},
+    {0x0100000000000827, "ContentActionTable", nullptr},
+}};
+
+VirtualFile SynthesizeSystemArchive(u64 title_id) {
```
ogniK5377: tid can be const

src/core/file_sys/system_archive/system_archive.h
```diff
@@ -0,0 +1,14 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+#include "core/file_sys/vfs_types.h"
+
+namespace FileSys::SystemArchive {
+
+VirtualFile SynthesizeSystemArchive(u64 title_id);
```
ogniK5377: title id can be const

PR #1705, opened 2018-11-16
Include QT imageformat dependencies with releases

Jcw87:
**Problem:**
Game icons do not show in mingw builds

**Cause:**
Missing libjpeg dll

**Solution:**
This PR (hopefully). I don't currently have a Linux environment set up to fully test this.


Jcw87:
Forgot about appveyor. Hopefully that one works too.

FearlessTobi:
Fixes icons on Appveyor mingw build. 🎉

zhaowenlan1779:
this won't fix travis mingw builds though, I think. as the qt5 ppa we are using there are actually a bit broken.

PR #1713, opened 2018-11-16
Implemented BRA CC conditional and FSET CC Setting

FernandoS27:
This is a common pattern in some games, it should fix some control flow in shaders.


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -3297,6 +3297,10 @@ class GLSLGenerator {
                 regs.SetRegisterToInteger(instr.gpr0, false, 0, predicate + " ? 0xFFFFFFFF : 0", 1,
                                           1);
             }
+            if (instr.generates_cc.Value() != 0) {
+                SetInternalFlag(InternalFlag::ZeroFlag, predicate);
+                LOG_WARNING(HW_GPU, "FSET Control Code is imcomplete");
```
lioncash: incomplete

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -3580,12 +3584,17 @@ class GLSLGenerator {
                 ASSERT_MSG(instr.bra.constant_buffer == 0,
                            "BRA with constant buffers are not implemented");
                 const Tegra::Shader::ControlCode cc = instr.flow_control_code;
+                const u32 target = offset + instr.bra.GetBranchTarget();
                 if (cc != Tegra::Shader::ControlCode::T) {
-                    LOG_CRITICAL(HW_GPU, "BRA Control Code used: {}", static_cast<u32>(cc));
-                    UNREACHABLE();
+                    const std::string control_code = regs.GetControlCode(instr.flow_control_code);
+                    shader.AddLine("if (" + control_code + "){");
```
ReinUsesLisp: `") {"`

Tinob:
im preparing a similar pull but implementing the controlcode handling in a more centralized way you can take a look here https://github.com/Tinob/yuzu/commits/ShaderGenerator

FernandoS27:
> im preparing a similar pull but implementing the controlcode handling in a more centralized way you can take a look here https://github.com/Tinob/yuzu/commits/ShaderGenerator

sadly centralizing it is no good because it's calculation can change depending on the instruction. For the time being since we are only setting zero flag its fine but when we implement the carry flag, it will become hard.

Tinob:
carry is an example of a contro lcode that must be implemented per instruction, but zero and nan can be easily implemented in common code because they work over the result stored in the the register. Please take a lok at my branch ant let me know the problems you found in my implementation.

bunnei:
@FernandoS27 needs a rebase, thanks!

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -3196,6 +3196,10 @@ class GLSLGenerator {
                 regs.SetRegisterToInteger(instr.gpr0, false, 0, predicate + " ? 0xFFFFFFFF : 0", 1,
                                           1);
             }
+            if (instr.generates_cc.Value() != 0) {
+                regs.SetInternalFlag(InternalFlag::ZeroFlag, predicate);
+                LOG_WARNING(HW_GPU, "FSET Control Code is incomplete");
```
ReinUsesLisp: Reminder to use `condition codes` instead of `Control Code` now that the name has changed.

PR #1722, opened 2018-11-17
am: Stub EnableApplicationCrashReport

MysticExile:
This is needed for Youtube to progress, does not yet boot though. Youtube among some other games needs CacheStorage implemented.


lioncash:
Message on the commit should indicate what it's for.

PR #1723, opened 2018-11-17
gl_rasterizer: Skip VB upload if the state is clean.

degasus:
This PR skips reemitting the vertex buffer state if the maxwell registers haven't been touched. But as the vertex buffers obviously depend on the main memory, we need to invalidate this state whenever the main memory may has been touched. This seems to be harder than expected.

### DIFF
--- a/src/video_core/engines/kepler_memory.cpp
+++ b/src/video_core/engines/kepler_memory.cpp
@@ -3,8 +3,10 @@
 // Refer to the license.txt file included.
 
 #include "common/logging/log.h"
+#include "core/core.h"
 #include "core/memory.h"
 #include "video_core/engines/kepler_memory.h"
+#include "video_core/engines/maxwell_3d.h"
 #include "video_core/rasterizer_interface.h"
 
 namespace Tegra::Engines {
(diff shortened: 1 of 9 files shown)


src/video_core/engines/kepler_memory.cpp
```diff
@@ -47,6 +49,7 @@ void KeplerMemory::ProcessData(u32 data) {
     rasterizer.InvalidateRegion(dest_address, sizeof(u32));
 
     Memory::Write32(dest_address, data);
+    Core::System::GetInstance().GPU().Maxwell3D().dirty_flags.OnMemoryWrite();
```
degasus: This happens really often and makes this patch close to superfluous. Is there a better way to check if nothing was written to the vertex buffers? Maybe always uploading them to the buffer_cache, but this will increase the load of the memory helpers a lot :/

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -334,6 +349,7 @@ void Maxwell3D::ProcessCBData(u32 value) {
         memory_manager.GpuToCpuAddress(buffer_address + regs.const_buffer.cb_pos);
 
     Memory::Write32(*address, value);
+    dirty_flags.OnMemoryWrite();
```
degasus: This happens really often and makes this patch close to superfluous. Is there a better way to check if nothing was written to the vertex buffers? Maybe always uploading them to the buffer_cache, but this will increase the load of the memory helpers a lot :/

Or are there some special barriers which just invalidate the vertex fetch cached? We might abuse them here.

Edit: fixed the location within the code...

bunnei:
@degasus Should this really still be RFC?

degasus:
@bunnei I'm not sure. I've got no comment to my question :/
I think this patch is fine, but I hope there is a better solution.

PR #1732, opened 2018-11-19
svc: Implement yield types 0 and -1

DarkLordZach:
This adds YieldNormal and YieldWithLoadBalancing, based upon mesosphere and research. While originally for Super Mario Party, the issue with that game seems to be something else (possibly audio?).

Meanwhile, two new games boot, while other random indie games boot further:
unknown 1
unknown

### DIFF
--- a/src/core/hle/kernel/scheduler.cpp
+++ b/src/core/hle/kernel/scheduler.cpp
@@ -169,6 +169,16 @@ void Scheduler::UnscheduleThread(Thread* thread, u32 priority) {
     ready_queue.remove(priority, thread);
 }
 
+void Scheduler::RescheduleThread(Thread* thread, u32 priority) {
+    std::lock_guard<std::mutex> lock(scheduler_mutex);
+
+    // Thread is not in queue
+    ASSERT(ready_queue.contains(thread) != -1);
+
+    ready_queue.remove(priority, thread);
+    ready_queue.push_back(priority, thread);
+}
+
 void Scheduler::SetThreadPriority(Thread* thread, u32 priority) {
     std::lock_guard<std::mutex> lock(scheduler_mutex);
 
@@ -177,6 +187,14 @@ void Scheduler::SetThreadPriority(Thread* thread, u32 priority) {
         ready_queue.move(thread, thread->GetPriority(), priority);
     else
         ready_queue.prepare(priority);
+}
+
+Thread* Scheduler::GetNextSuggestedThread(u32 core) {
+    std::lock_guard<std::mutex> lock(scheduler_mutex);
+
+    const auto mask = 1 << core;
+    return ready_queue.get_first_filter(
+        [&mask](Thread* thread) { return (thread->GetAffinityMask() & mask) != 0; });
 }
 
 } // namespace Kernel
--- a/src/core/hle/kernel/scheduler.h
+++ b/src/core/hle/kernel/scheduler.h
@@ -48,8 +48,14 @@ public:
     /// Unschedules a thread that was already scheduled
     void UnscheduleThread(Thread* thread, u32 priority);
 
+    /// Moves a thread to the back of the current priority queue
+    void RescheduleThread(Thread* thread, u32 priority);
+
     /// Sets the priority of a thread in the scheduler
     void SetThreadPriority(Thread* thread, u32 priority);
+
+    /// Gets the next suggested thread for load balancing
+    Thread* GetNextSuggestedThread(u32 core);
 
     /// Returns a list of all threads managed by the scheduler
     const std::vector<SharedPtr<Thread>>& GetThreadList() const {
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -962,8 +962,28 @@ static void SleepThread(s64 nanoseconds) {
 
     // Don't attempt to yield execution if there are no available threads to run,
     // this way we avoid a useless reschedule to the idle thread.
-    if (nanoseconds == 0 && !Core::System::GetInstance().CurrentScheduler().HaveReadyThreads())
+    if (!Core::System::GetInstance().CurrentScheduler().HaveReadyThreads())
         return;
+
+    if (nanoseconds <= 0) {
+        switch (nanoseconds) {
+        case 0:
+            GetCurrentThread()->YieldNormal();
+            break;
+        case -1:
+            GetCurrentThread()->YieldWithLoadBalancing();
+            break;
+        case -2:
+            GetCurrentThread()->YieldAndWaitForLoadBalancing();
+            break;
+        default:
+            UNREACHABLE_MSG(
+                "Unimplemented sleep yield type '{:016X}'! Falling back to forced reschedule...",
+                nanoseconds);
+        }
+
+        nanoseconds = 0;
+    }
 
     // Sleep current thread and check for next thread to schedule
     WaitCurrentThread_Sleep();
@@ -971,7 +991,10 @@ static void SleepThread(s64 nanoseconds) {
     // Create an event to wake the thread up after the specified nanosecond delay has passed
     GetCurrentThread()->WakeAfterDelay(nanoseconds);
 
-    Core::System::GetInstance().PrepareReschedule();
+    Core::System::GetInstance().CpuCore(0).PrepareReschedule();
+    Core::System::GetInstance().CpuCore(1).PrepareReschedule();
+    Core::System::GetInstance().CpuCore(2).PrepareReschedule();
+    Core::System::GetInstance().CpuCore(3).PrepareReschedule();
 }
 
 /// Wait process wide key atomic
--- a/src/core/hle/kernel/thread.cpp
+++ b/src/core/hle/kernel/thread.cpp
@@ -388,6 +388,66 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
     return wakeup_callback(reason, std::move(thread), std::move(object), index);
 }
 
+void Thread::YieldNormal() {
+    // Avoid yielding if the thread isn't even running.
+    if (status != ThreadStatus::Running) {
+        return;
+    }
+
+    if (nominal_priority < THREADPRIO_COUNT) {
+        scheduler->RescheduleThread(this, nominal_priority);
+        scheduler->Reschedule();
+    }
+}
+
+void Thread::YieldWithLoadBalancing() {
+    auto priority = nominal_priority;
+    auto core = processor_id;
+
+    // Avoid yielding if the thread isn't even running.
+    if (status != ThreadStatus::Running) {
+        Core::System::GetInstance().CpuCore(processor_id).PrepareReschedule();
+        return;
+    }
+
+    SharedPtr<Thread> next;
+    const auto& threads = scheduler->GetThreadList();
+
+    if (priority < THREADPRIO_COUNT) {
+        // Reschedule thread to end of queue.
+        scheduler->RescheduleThread(this, priority);
+
+        const auto iter = std::find_if(threads.begin(), threads.end(),
+                                       [&priority](const SharedPtr<Thread>& thread) {
+                                           return thread->GetNominalPriority() == priority;
+                                       });
+
+        if (iter != threads.end())
+            next = iter->get();
+    }
+
+    Thread* suggested_thread = nullptr;
+
+    for (int i = 0; i < 4; ++i) {
+        if (i == core)
+            continue;
+
+        const auto res =
+            Core::System::GetInstance().CpuCore(i).Scheduler().GetNextSuggestedThread(core);
+        if (res != nullptr) {
+            suggested_thread = res;
+            break;
+        }
+    }
+
+    if (suggested_thread != nullptr)
+        suggested_thread->ChangeCore(core, suggested_thread->GetAffinityMask());
+}
+
+void Thread::YieldAndWaitForLoadBalancing() {
+    UNIMPLEMENTED_MSG("Wait for load balancing thread yield type is not implemented!");
+}
+
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
 /**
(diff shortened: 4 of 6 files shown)


src/core/hle/kernel/scheduler.cpp
```diff
@@ -179,4 +189,12 @@ void Scheduler::SetThreadPriority(Thread* thread, u32 priority) {
         ready_queue.prepare(priority);
 }
 
+Thread* Scheduler::GetNextSuggestedThread(u32 core) {
+    std::lock_guard<std::mutex> lock(scheduler_mutex);
+
+    const auto mask = 1 << core;
+    return ready_queue.get_first_filter(
+        [&mask](Thread* thread) { return (thread->GetAffinityMask() & mask) != 0; });
```
Subv: Thread const*

src/core/hle/kernel/scheduler.h
```diff
@@ -48,9 +48,15 @@ class Scheduler final {
     /// Unschedules a thread that was already scheduled
     void UnscheduleThread(Thread* thread, u32 priority);
 
+    /// Moves a thread to the back of the current priority queue
+    void RescheduleThread(Thread* thread, u32 priority);
```
Subv: The name of this function is not an indicator of what it actually does, I suggest renaming it and adding more comments to its body

src/core/hle/kernel/scheduler.h
```diff
@@ -48,9 +48,15 @@ class Scheduler final {
...
+
     /// Sets the priority of a thread in the scheduler
     void SetThreadPriority(Thread* thread, u32 priority);
 
+    /// Gets the next suggested thread for load balancing
+    Thread* GetNextSuggestedThread(u32 core);
```
Subv: this can be a const function

src/core/hle/kernel/svc.cpp
```diff
@@ -962,16 +962,39 @@ static void SleepThread(s64 nanoseconds) {
 
     // Don't attempt to yield execution if there are no available threads to run,
     // this way we avoid a useless reschedule to the idle thread.
-    if (nanoseconds == 0 && !Core::System::GetInstance().CurrentScheduler().HaveReadyThreads())
+    if (!Core::System::GetInstance().CurrentScheduler().HaveReadyThreads())
         return;
 
+    if (nanoseconds <= 0) {
+        switch (nanoseconds) {
+        case 0:
```
Subv: use an enum class.

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,66 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
     return wakeup_callback(reason, std::move(thread), std::move(object), index);
 }
 
+void Thread::YieldNormal() {
+    // Avoid yielding if the thread isn't even running.
+    if (status != ThreadStatus::Running) {
+        return;
```
Subv: This should be an assert

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,66 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
...
+    }
+
+    if (nominal_priority < THREADPRIO_COUNT) {
```
Subv: shouldn't this be using the current priority instead of the nominal priority?

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,66 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
...
+        scheduler->Reschedule();
+    }
+}
+
+void Thread::YieldWithLoadBalancing() {
+    auto priority = nominal_priority;
+    auto core = processor_id;
+
+    // Avoid yielding if the thread isn't even running.
+    if (status != ThreadStatus::Running) {
```
Subv: ditto

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,66 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
...
+                                           return thread->GetNominalPriority() == priority;
+                                       });
+
+        if (iter != threads.end())
+            next = iter->get();
+    }
+
+    Thread* suggested_thread = nullptr;
+
+    for (int i = 0; i < 4; ++i) {
```
Subv: don't use magic numbers. You can come up with a better name than `i`

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,66 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
...
+    // Avoid yielding if the thread isn't even running.
+    if (status != ThreadStatus::Running) {
+        Core::System::GetInstance().CpuCore(processor_id).PrepareReschedule();
+        return;
+    }
+
+    SharedPtr<Thread> next;
+    const auto& threads = scheduler->GetThreadList();
+
+    if (priority < THREADPRIO_COUNT) {
```
Subv: These checks shouldn't be necessary. If you want to keep them then make them asserts.

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,66 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
...
+        // Reschedule thread to end of queue.
+        scheduler->RescheduleThread(this, priority);
+
+        const auto iter = std::find_if(threads.begin(), threads.end(),
```
Subv: What is this used for? `next` is an unused variable

src/core/hle/kernel/thread.h
```diff
@@ -370,6 +371,10 @@ class Thread final : public WaitObject {
         return affinity_mask;
     }
 
+    void YieldNormal();
```
Subv: An explanation of what each yield type does should be added as comments. I expect to see a good explanation of each one and how it interacts with other parts of the scheduler (perhaps even adding an example of each), not just a handwavy one-liner.

Subv:
I would like to see an explanation of what each of these new yield types does

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,66 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
     return wakeup_callback(reason, std::move(thread), std::move(object), index);
 }
 
+void Thread::YieldNormal() {
```
lioncash: A more informative name here would be `YieldWithoutLoadBalancing`

jroweboy:
As a notice, users have been reporting somewhat severe performance regressions with this PR. (Narrowed down by its addition in canary 1544 which only adds this pr)

bunnei:
Uncanary tagging. @DarkLordZach I confirmed in my testing that this is the source of user reported perf regressions.

src/core/hle/kernel/scheduler.h
```diff
@@ -49,13 +49,79 @@ class Scheduler final {
...
+     * ready_queue[core=0][prio=0]: ThreadA, ThreadB (affinities not pictured as irrelevant
+     * ready_queue[core=1][prio=0]: ThreadC[affinity=both], ThreadD[affinity=core1only]
+     * Currently Running: ThreadQ on Core 0 || ThreadP on Core 1
+     *
+     * ThreadQ calls YieldWithLoadBalancing
+     *
+     * ThreadQ is moved to the end of ready_queue[core=0][prio=0]:
+     * ready_queue[core=0][prio=0]: ThreadA, ThreadB
+     * ready_queue[core=1][prio=0]: ThreadC[affinity=both], ThreadD[affinity=core1only]
+     * Currently Running: ThreadQ on Core 0 || ThreadP on Core 1
```
Subv: Shouldn't this have Nothing on Core 0 and ThreadQ in core0's queue?

src/core/hle/kernel/scheduler.h
```diff
@@ -49,13 +49,79 @@ class Scheduler final {
...
+     *
+     * A list of suggested threads for each core is compiled
+     * Suggested Threads: {ThreadC on Core 1}
+     * If this were quad core (as the switch is), there could be between 0 and 3 threads in this
```
Subv: Wouldn't there be 4 threads in the suggestion list, given the current core still has some stuff in its queue? I would imagine a higher priority thread in Core0 would still run before a same-priority thread stolen from Core1's queue

src/core/hle/kernel/scheduler.cpp
```diff
@@ -179,4 +190,64 @@ void Scheduler::SetThreadPriority(Thread* thread, u32 priority) {
...
+
+    // Sanity check that the priority is valid
+    ASSERT(priority < THREADPRIO_COUNT);
+
+    // Reschedule thread to end of queue.
+    MoveThreadToBackOfPriorityQueue(thread, priority);
+
+    Thread* suggested_thread = nullptr;
+
+    // Search through all of the cpu cores (except this one) for a suggested thread.
```
Subv: Why is the current core not scanned? What if there's a higher-priority thread waiting to be run in its queue? Or is the suggested thread always stolen even if this scenario occurs?

DarkLordZach: In ref to this comment and the previous, the suggested thread must be from another core (stolen as you put it). It will use threads from the current core iff there are no suitiable candidates on others first.

src/core/hle/kernel/scheduler.h
DarkLordZach: yes, sorry. honest mistake

src/core/hle/kernel/scheduler.cpp
```diff
@@ -179,4 +180,71 @@ void Scheduler::SetThreadPriority(Thread* thread, u32 priority) {
...
+    // Sleep for zero time to be able to force reschedule to different thread
+    WaitCurrentThread_Sleep();
+    GetCurrentThread()->WakeAfterDelay(0);
+
+    Thread* suggested_thread = nullptr;
+
+    // Search through all of the cpu cores (except this one) for a suggested thread.
+    // Take the first non-nullptr one
+    for (unsigned cur_core = 0; cur_core < Core::NUM_CPU_CORES; ++cur_core) {
+        if (cur_core == core)
```
Subv: This condition needs to go away

src/core/hle/kernel/scheduler.cpp
```diff
@@ -179,4 +180,71 @@ void Scheduler::SetThreadPriority(Thread* thread, u32 priority) {
...
+            continue;
+
+        const auto res =
+            Core::System::GetInstance().CpuCore(cur_core).Scheduler().GetNextSuggestedThread(
+                core, priority);
+        if (res != nullptr &&
```
Subv: Please separate the nullptr check into another if.

src/core/hle/kernel/scheduler.cpp
```diff
@@ -179,4 +180,71 @@ void Scheduler::SetThreadPriority(Thread* thread, u32 priority) {
...
+            suggested_thread = res;
+        }
+    }
+
+    // If a suggested thread was found, queue that for this core
+    if (suggested_thread != nullptr)
+        suggested_thread->ChangeCore(core, suggested_thread->GetAffinityMask());
+
+    // Perform actual yielding.
+    Reschedule();
```
Subv: mm, thinking about it again, you shouldn't be calling Reschedule manually, the PrepareReschedule call done in SleepThread will make the cores context-switch as soon as the timeslice ends (which should be immediately if a reschedule is requested).

src/core/hle/kernel/svc.cpp
```diff
@@ -962,14 +962,37 @@ static void SleepThread(s64 nanoseconds) {
...
+        }
+    } else {
+        // Sleep current thread and check for next thread to schedule
+        WaitCurrentThread_Sleep();
+
+        // Create an event to wake the thread up after the specified nanosecond delay has passed
+        GetCurrentThread()->WakeAfterDelay(nanoseconds);
+    }
 
     Core::System::GetInstance().PrepareReschedule();
```
Subv: This should probably be a reschedule of all cores, ie

```
Core::System::GetInstance().CpuCore(0).PrepareReschedule();
Core::System::GetInstance().CpuCore(1).PrepareReschedule();
Core::System::GetInstance().CpuCore(2).PrepareReschedule();
Core::System::GetInstance().CpuCore(3).PrepareReschedule();
```

but in a more stylish manner

src/core/hle/kernel/svc.cpp
```diff
@@ -962,16 +962,41 @@ static void SleepThread(s64 nanoseconds) {
 
     // Don't attempt to yield execution if there are no available threads to run,
     // this way we avoid a useless reschedule to the idle thread.
-    if (nanoseconds == 0 && !Core::System::GetInstance().CurrentScheduler().HaveReadyThreads())
+    if (nanoseconds <= 0 && !Core::System::GetInstance().CurrentScheduler().HaveReadyThreads())
```
Subv: This condition should be removed. I suggest adding a TODO that says there are optimization opportunities here, but they need to be thought over. This check as it currently stands is broken.

src/core/hle/kernel/svc.cpp
```diff
@@ -962,16 +962,41 @@ static void SleepThread(s64 nanoseconds) {
...
+        // Sleep current thread and check for next thread to schedule
+        WaitCurrentThread_Sleep();
 
-    Core::System::GetInstance().PrepareReschedule();
+        // Create an event to wake the thread up after the specified nanosecond delay has passed
+        GetCurrentThread()->WakeAfterDelay(nanoseconds);
+    }
+
+    // Reschedule all CPU cores
+    for (std::size_t i = 0; i < 4; ++i)
```
Subv: NUM_CORES

PR #1734, opened 2018-11-19
kernel/shared_memory: Make data members private, plus minor interface changes

lioncash:
Makes it nicer to reason about the internals of the shared memory instances, by exposing only what is necessary to non-kernel code. In particular, we don't need to directly expose any of the memory-specific data members. Each commit contains rationale to clarify themselves


lioncash:
Failure is just mingw being silly

PR #1735, opened 2018-11-19
Texture decoder: Implemented Tile Width Spacing

FernandoS27:
This fixes textures in Octopath and some indie games.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.h
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.h
@@ -196,9 +196,15 @@ struct SurfaceParams {
 
     /// Checks if surfaces are compatible for caching
     bool IsCompatibleSurface(const SurfaceParams& other) const {
-        return std::tie(pixel_format, type, width, height, target, depth) ==
-               std::tie(other.pixel_format, other.type, other.width, other.height, other.target,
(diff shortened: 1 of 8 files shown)


src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -208,6 +208,7 @@ struct SurfaceParams {
     u32 block_width;
     u32 block_height;
     u32 block_depth;
+    u32 tile_width_spacing;
```
Subv: This should be added to the IsCompatibleSurface check. In fact, block_width/height/depth should also be added there.

FernandoS27: that should be in another PR plus I'm not sure if we really have to do that.

Subv: Of course it's needed, otherwise you run into the risk of reusing an OpenGL surface with different parameters.

bunnei:
@Subv please re-review

FernandoS27:
@Subv your comments have been addressed.

bunnei:
@FernandoS27 needs a rebase

FernandoS27:
@bunnei rebased

PR #1736, opened 2018-11-19
Added predicate comparison LessEqualWithNan

Hexagon12:
Implements predicate comparison 11 (0xB), also known as LessEqualWithNan.

Fixes a crash while trying to go ingame in L.A. Noire.
Fixes trainer battle crashes in Pokemon Let's Go.


bunnei:
@Hexagon12 please address clang format issues

PR #1738, opened 2018-11-19
kernel/resource_limit: Clean up interface

lioncash:
Cleans out the citra/3DS-specific implementation details that don't apply to the Switch. Sets the stage for implementing ResourceLimit instances properly. If there's anything that looks like its missing data-structure-wise, this is intentional, I know about the missing members. I've not included those in this PR. They will be included within the follow-up where they're more relevant.

While we're at it, remove the erroneous checks within CreateThread() and SetThreadPriority(). While these are indeed checked in some capacity, they are not checked via a ResourceLimit instance.

In the process of moving out Citra-specifics, this also replaces the system ResourceLimit instance's values with ones from the Switch.

### DIFF
--- a/src/core/hle/kernel/resource_limit.h
+++ b/src/core/hle/kernel/resource_limit.h
@@ -4,31 +4,25 @@
 
 #pragma once
 
+#include <array>
 #include "common/common_types.h"
 #include "core/hle/kernel/object.h"
+
+union ResultCode;
 
 namespace Kernel {
 
 class KernelCore;
 
-enum class ResourceLimitCategory : u8 {
-    APPLICATION = 0,
-    SYS_APPLET = 1,
-    LIB_APPLET = 2,
-    OTHER = 3
-};
-
 enum class ResourceType {
-    Priority = 0,
-    Commit = 1,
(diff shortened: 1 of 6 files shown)


src/core/hle/kernel/resource_limit.h
```diff
@@ -55,61 +49,51 @@ class ResourceLimit final : public Object {
...
-    s32 max_cpu_time = 0;
+private:
+    explicit ResourceLimit(KernelCore& kernel);
+    ~ResourceLimit() override;
 
-    // TODO(Subv): Increment these in their respective Kernel::T::Create functions, keeping in mind
-    // that APPLICATION resource limits should not be affected by the objects created by service
-    // modules.
+    // TODO(Subv): Increment resource limit current values in their respective Kernel::T::Create
+    // functions, keeping in mind that APPLICATION resource limits should not be affected by the
```
Subv: The second part of this comment is no longer relevant

lioncash: Good catch. I've updated the comment. If you also meant the part about the running applications and service modules, I can remove that too.

PR #1740, opened 2018-11-20
Implemented Shader Unique Identifiers

FernandoS27:
This PR does not add much functionality in its own but is vital for implementing the Disk Shader Cache and Mods that inject shaders.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_cache.cpp
@@ -2,7 +2,9 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <boost/functional/hash.hpp>
 #include "common/assert.h"
+#include "common/hash.h"
 #include "core/core.h"
 #include "core/memory.h"
 #include "video_core/engines/maxwell_3d.h"
@@ -66,14 +68,17 @@ CachedShader::CachedShader(VAddr addr, Maxwell::ShaderProgram program_type)
         // stage here.
         setup.SetProgramB(GetShaderCode(GetShaderAddress(Maxwell::ShaderProgram::VertexB)));
     case Maxwell::ShaderProgram::VertexB:
+        CalculateProperties();
         program_result = GLShader::GenerateVertexShader(setup);
         gl_type = GL_VERTEX_SHADER;
         break;
     case Maxwell::ShaderProgram::Geometry:
+        CalculateProperties();
         program_result = GLShader::GenerateGeometryShader(setup);
         gl_type = GL_GEOMETRY_SHADER;
         break;
     case Maxwell::ShaderProgram::Fragment:
+        CalculateProperties();
         program_result = GLShader::GenerateFragmentShader(setup);
         gl_type = GL_FRAGMENT_SHADER;
         break;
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -139,6 +143,43 @@ GLuint CachedShader::LazyGeometryProgram(OGLProgram& target_program,
...
+    static constexpr size_t SchedPeriod = 4;
+    u32 absolute_offset = offset - main_offset;
+    return (absolute_offset % SchedPeriod) == 0;
+}
+
+static std::size_t CalculateProgramSize(GLShader::ProgramCode program) {
+    const u32 start_offset = 10;
+    u32 offset = start_offset;
+    std::size_t size = start_offset * sizeof(u64);
+    while (true) { // dump until hitting not finding a valid instruction
```
Subv: mm, this is potentially O(Infty). Shaders could keep garbage after their exit instruction.

FernandoS27: Well I can limit it by the size we give to the vector.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -139,6 +143,43 @@ GLuint CachedShader::LazyGeometryProgram(OGLProgram& target_program,
     return target_program.handle;
 };
 
+static bool IsSchedInstruction(u32 offset, u32 main_offset) {
+    // sched instructions appear once every 4 instructions.
+    static constexpr size_t SchedPeriod = 4;
+    u32 absolute_offset = offset - main_offset;
```
jroweboy: const u32

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -139,6 +143,43 @@ GLuint CachedShader::LazyGeometryProgram(OGLProgram& target_program,
...
+        size += 8;
+        offset++;
+    }
+    return size;
+}
+
+void CachedShader::CalculateProperties() {
+    setup.program.real_size = CalculateProgramSize(setup.program.code);
+    setup.program.real_size_b = 0;
+    setup.program.unique_identifier = Common::CityHash64(
```
jroweboy: For one off hashes, Common Hash should be fine, but if you need to combine two hashes, boost::hash_combine is preferable

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -139,6 +143,43 @@ GLuint CachedShader::LazyGeometryProgram(OGLProgram& target_program,
...
+        reinterpret_cast<const char*>(setup.program.code.data()), setup.program.real_size);
+    if (program_type == Maxwell::ShaderProgram::VertexA) {
+        setup.program.real_size_b = CalculateProgramSize(setup.program.code_b);
+        u64 identifier_b = Common::CityHash64(
```
jroweboy: const u64

FernandoS27:
@jroweboy changes made, now uses hash_combine when needed.

jroweboy:
Please fix clang format and i'll merge this <3

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -139,6 +144,46 @@ GLuint CachedShader::LazyGeometryProgram(OGLProgram& target_program,
...
 };
 
+static bool IsSchedInstruction(u32 offset, u32 main_offset) {
+    // sched instructions appear once every 4 instructions.
+    static constexpr size_t SchedPeriod = 4;
+    const u32 absolute_offset = offset - main_offset;
+    return (absolute_offset % SchedPeriod) == 0;
+}
+
+static std::size_t CalculateProgramSize(GLShader::ProgramCode program) {
```
lioncash: `program` should be taken by const reference, otherwise this is creating an unnecessary copy of the entire program code.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -139,6 +144,46 @@ GLuint CachedShader::LazyGeometryProgram(OGLProgram& target_program,
...
+    const u32 start_offset = 10;
+    u32 offset = start_offset;
```
lioncash: `offset` should be a std::size_t` to match the type on the right-hand side of the loop comparison.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -139,6 +144,46 @@ GLuint CachedShader::LazyGeometryProgram(OGLProgram& target_program,
...
+    std::size_t size = start_offset * sizeof(u64);
+    while (offset < program.size()) {
+        const u64 inst = program[offset];
+        if (!IsSchedInstruction(offset, start_offset)) {
+            if (inst == 0) {
```
ReinUsesLisp: Sometimes shaders end without non-zero no-ops, you should check for those too.
Also, shaders are not allowed to end in `nvdisasm` (hence how we have to assume the shader binary actually ends) unless the current instruction is in a 0x18 offset. Here's a dirty example in Ruby:
```ruby
len = 0
while len < raw.length
        if ((len / 8) % 4 != 3)
                len += 8
                next
        end
        low = raw[len + 6]
        high = raw[len + 7]
        if (low == 0xb0 && high == 0x50) || (low == 0 && high == 0)
                break
        end
        len += 8
end
```

src/video_core/renderer_opengl/gl_shader_gen.cpp
```diff
@@ -16,6 +17,8 @@ static constexpr u32 PROGRAM_OFFSET{10};
 ProgramResult GenerateVertexShader(const ShaderSetup& setup) {
     std::string out = "#version 430 core\n";
     out += "#extension GL_ARB_separate_shader_objects : enable\n\n";
+    const std::string id = fmt::format("{:016x}", setup.program.unique_identifier);
+    out += "// Shader Unique Id: VS" + id + "\n\n";
```
ReinUsesLisp: I don't like the idea of writing the unique id to the shader source. If the identifier changes it will invalidate host driver's shader cache.

FernandoS27: depends on your design of the shader cache.

FernandoS27:
Changes made and fixed.

bunnei:
I'm OK with the changes, but we'll canary tag for a day or so just to make sure there are no surprises...

FernandoS27:
@bunnei ping.

PR #1742, opened 2018-11-20
am/applets: Minor cleanup

lioncash:
Mostly just cleans up the comment locations to get rid of wonky wrapping, making it nicer to read vertically. It also cleans up the inclusions and moves the broker instance itself into the Applet instance.

### DIFF
--- a/src/core/hle/service/am/applets/applets.h
+++ b/src/core/hle/service/am/applets/applets.h
@@ -4,13 +4,15 @@
 
 #pragma once
 
-#include <functional>
 #include <memory>
 #include <queue>
 #include "common/swap.h"
-#include "core/hle/kernel/event.h"
 
 union ResultCode;
+
+namespace Kernel {
+class Event;
+}
 
 namespace Service::AM {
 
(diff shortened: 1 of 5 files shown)


misson20000:
Personally, I think `AppletDataBroker` should belong to the `Applet` instead of the `ILibraryAppletAccessor`, since the accessor is supposed to just be accessing the applet, not owning any part of it. In applet services, the storage channels are fields in `IntegratedLibraryApplet`.

lioncash:
Good point! I've updated the commit to move the broker instance into the Applet itself instead of the accessor.

src/core/hle/service/am/applets/applets.h
```diff
@@ -85,8 +103,8 @@ class Applet {
     };
     static_assert(sizeof(CommonArguments) == 0x20, "CommonArguments has incorrect size.");
 
-    CommonArguments common_args;
-    std::shared_ptr<AppletDataBroker> broker;
+    CommonArguments common_args{};
```
DarkLordZach: Is this needed as it's going to be memcpy'd into before use anyway?

lioncash: It provides a deterministic state for the data when constructed (memcpy isn't done at construction time)

DarkLordZach: fair enough

PR #1750, opened 2018-11-21
am: Correct build failure

lioncash:
Corrects a build failure that occurred due to commits being merged that utilized previously public shared memory internals before merging the shared memory change itself.


bunnei:
Oops - thanks

PR #1754, opened 2018-11-21
gl_shader_decompiler: Remove UNREACHABLE when setting RZ

ReinUsesLisp:
The following example sets 1 to `$r20`:
```asm
mov32i 0x0 0xdeadbeef 0xf
mov32i $r20 0x00000001 0xf
iadd $r20 $r20 0x0
```
That implies that writing to `RZ` (0x0 in `envyas` syntax) is a nop (reading from it always returns zero). I didn't test what happens to condition codes in those cases, but that's beyond the scope of `SetRegister`.


FernandoS27:
I was going to change this as well, but CC is actually set if activated.

PR #1756, opened 2018-11-21
gl_shader_decompiler: Fix register overwriting on texture calls

ReinUsesLisp:
Currently `texture*` calls can generate the following (pseudo-)code:
```glsl
{
    vec2 coords = vec2(...);
    gpr1 = texture(tex, coords, gpr2).x;
    gpr2 = texture(tex, coords, gpr2).y;
    gpr3 = texture(tex, coords, gpr2).z;
    gpr4 = texture(tex, coords, gpr2).w;
}
```
`gpr2` is overwrited there. This PR attempts to address that using a temporal variable like `coords`.

This may fix some games, but I couldn't find any noticeable issue being fixed.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -1243,14 +1243,7 @@ private:
         regs.SetRegisterToInteger(dest, true, 0, result, 1, 1);
     }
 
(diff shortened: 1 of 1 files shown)


marcosvitali:
Rodrigo you forget, fixed some texures of mario hat in SMO menu :)

Tinob:
It would be simpler and cleaner to just execute the texture command once, assign it to a vec4 temporal variable and then use that variable as a source to get the components to the registers

PR #1764, opened 2018-11-21
macro_interpreter: Implement AddWithCarry and SubtractWithBorrow.

bunnei:
- Used by Undertale.

### DIFF
--- a/src/video_core/macro_interpreter.cpp
+++ b/src/video_core/macro_interpreter.cpp
@@ -35,6 +35,7 @@ void MacroInterpreter::Reset() {
     // The next parameter index starts at 1, because $r1 already has the value of the first
     // parameter.
     next_parameter_index = 1;
+    carry_flag = false;
 }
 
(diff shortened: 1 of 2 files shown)


src/video_core/macro_interpreter.cpp
```diff
@@ -135,14 +136,28 @@ MacroInterpreter::Opcode MacroInterpreter::GetOpcode(u32 offset) const {
...
+u32 MacroInterpreter::GetALUResult(ALUOperation operation, u32 src_a, u32 src_b) {
     switch (operation) {
-    case ALUOperation::Add:
-        return src_a + src_b;
-    // TODO(Subv): Implement AddWithCarry
-    case ALUOperation::Subtract:
-        return src_a - src_b;
-    // TODO(Subv): Implement SubtractWithBorrow
+    case ALUOperation::Add: {
+        u64 result = static_cast<u64>(src_a) + static_cast<u64>(src_b);
```
lioncash: `result` can be const in all cases added.

PR #1765, opened 2018-11-21
audout_u: Add support for multiple IAudioOut streams.

bunnei:
- Used by Undertale.

### DIFF
--- a/src/core/hle/service/audio/audout_u.cpp
+++ b/src/core/hle/service/audio/audout_u.cpp
@@ -44,8 +44,10 @@ enum class AudioState : u32 {
 
 class IAudioOut final : public ServiceFramework<IAudioOut> {
 public:
-    IAudioOut(AudoutParams audio_params, AudioCore::AudioOut& audio_core)
-        : ServiceFramework("IAudioOut"), audio_core(audio_core), audio_params(audio_params) {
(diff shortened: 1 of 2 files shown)


src/core/hle/service/audio/audout_u.cpp
```diff
@@ -199,7 +202,16 @@ void AudOutU::ListAudioOutsImpl(Kernel::HLERequestContext& ctx) {
 void AudOutU::OpenAudioOutImpl(Kernel::HLERequestContext& ctx) {
     LOG_DEBUG(Service_Audio, "called");
 
-    ctx.WriteBuffer(DefaultDevice);
+    const auto device_name_data{ctx.ReadBuffer()};
+    std::string device_name;
+    if (device_name_data[0] != '\0') {
+        device_name.assign(device_name_data.data(),
```
```cpp
device_name.assign(device_name_data.begin(), device_name_data.end());
```

src/core/hle/service/audio/audout_u.cpp
```diff
@@ -199,7 +202,16 @@ void AudOutU::ListAudioOutsImpl(Kernel::HLERequestContext& ctx) {
...
+                           device_name_data.data() + device_name_data.size());
+    } else {
+        device_name = DefaultDevice.data();
```
```cpp
device_name.assign(DefaultDevice.begin(), DefaultDevice.end());
```

avoids a strlen call

PR #1770, opened 2018-11-22
applets: Add StubApplet and use it as fallback when AppletId is not implemented

DarkLordZach:
This will log all data it receives, log all calls to its methods and push dummy data into both channels on execution.

This should work for most games that regressed with #1667, but some may still svcBreak. This is simply because they expect real data for the results instead of zeroes.

This is intended to be a temporary solution to the problems exposed by #1667.

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -6,8 +6,6 @@
 #include <cinttypes>
 #include <cstring>
 #include <stack>
-#include "applets/applets.h"
-#include "applets/software_keyboard.h"
 #include "audio_core/audio_renderer.h"
 #include "core/core.h"
 #include "core/hle/ipc_helpers.h"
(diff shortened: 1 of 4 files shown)


src/core/hle/service/am/applets/stub_applet.h
```diff
@@ -0,0 +1,31 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
```
lioncash: These standard includes don't seem to be necessary in the header

src/core/hle/service/am/applets/stub_applet.h
```diff
@@ -0,0 +1,31 @@
...
+#include <string>
+#include <vector>
+
+#include "common/common_funcs.h"
```
lioncash: This, `common/swap.h` and `core/hle/service/am/am.h` don't seem to be necessary in the header.

src/core/hle/service/am/am.cpp
```diff
@@ -761,8 +762,9 @@ static std::shared_ptr<Applets::Applet> GetAppletFromId(AppletId id) {
     case AppletId::SoftwareKeyboard:
         return std::make_shared<Applets::SoftwareKeyboard>();
     default:
-        UNREACHABLE_MSG("Unimplemented AppletId [{:08X}]!", static_cast<u32>(id));
-        return nullptr;
+        LOG_CRITICAL(Service_AM, "Unimplemented AppletId [{:08X}]! -- Falling back to stub!",
```
lioncash: This should probably be a warning or error, rather than a critical log.

PR #1772, opened 2018-11-22
[WIP] Missing states v3

Tinob:
Implement support for clear_flags and depth clamp, issues like wrong stencil shadows in smo should be fixed.


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -483,6 +502,27 @@ void OpenGLState::Apply() const {
                          fragment_color_clamp.enabled ? GL_TRUE : GL_FALSE);
         }
     }
+    if (depth_clamp.far_plane != cur_state.depth_clamp.far_plane ||
+        depth_clamp.near_plane != cur_state.depth_clamp.near_plane) {
+        if (GLAD_GL_AMD_depth_clamp_separate) {
```
ReinUsesLisp: It would be nice if you described why a vendor-specific instruction is needed, specially when emulating an NVidia device and that extension is not exposed by NVidia's propietary driver.

ReinUsesLisp: Also, when an extension is used with a fallback (like here, for instance), log it in OpenGL's context initialization.

Tinob: just added a function to log  unsuported extension like this. About the specific case of GLAD_GL_AMD_depth_clamp_separate if you think is necesary i will add a coment about the issue in the code.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -185,6 +186,10 @@ class RasterizerOpenGL : public VideoCore::RasterizerInterface {
     /// Check asserts for alpha testing.
     void CheckAlphaTests();
 
+    /// check for extension that are not strictly required
```
ReinUsesLisp: Caps

src/video_core/renderer_opengl/gl_state.cpp
ReinUsesLisp: I think you should remove the vendor-specific path and add:
```cpp
UNIMPLEMENTED_IF(depth_clamp.far_plane != depth_clamp.near_plane);
```

PR #1780, opened 2018-11-22
configure_input: Add Controller Setup Profiles and simplify input UI

DarkLordZach:
Greatly simplifies the current input UI, while still allowing power users to tweak advanced settings. Adds 'input profiles', which are easy autoconfigurations to make getting started easy and fast. Also has a custom option which brings up the current, full UI.

For the profiles, clicking configure will open the button mapping. For custom, it opens the current UI in a dialog.

Currently there are two profiles:
- `Single Player - Handheld - Undocked` (Default)
- `Single Player - Dual Joycons - Docked`
along with `Custom` which will open the current UI when configured.
From my testing, most booting games will work with one or both of these profiles. If a game doesn't work, file an issue report with the game and we can add a new profile.

Examples:
image

### DIFF
--- a/src/yuzu/configuration/configure_input_simple.cpp
+++ b/src/yuzu/configuration/configure_input_simple.cpp
@@ -1,0 +1,140 @@
+// Copyright 2016 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <array>
+#include <cstring>
+#include <functional>
+#include <tuple>
+
+#include <QDialog>
+
+#include "ui_configure_input_simple.h"
+#include "yuzu/configuration/configure_input.h"
+#include "yuzu/configuration/configure_input_player.h"
+#include "yuzu/configuration/configure_input_simple.h"
+#include "yuzu/ui_settings.h"
+
+namespace {
+
+template <typename Dialog, typename... Args>
+void CallConfigureDialog(ConfigureInputSimple* caller, Args&&... args) {
+    caller->applyConfiguration();
+    Dialog dialog(caller, std::forward<Args>(args)...);
+
+    const auto res = dialog.exec();
+    if (res == QDialog::Accepted) {
+        dialog.applyConfiguration();
+    }
+}
+
+// OnProfileSelect functions should (when applicable):
+// - Set controller types
+// - Set controller enabled
+// - Set docked mode
+// - Set advanced controller config/enabled (i.e. debug, kbd, mouse, touch)
+//
+// OnProfileSelect function should NOT however:
+// - Reset any button mappings
+// - Open any dialogs
+// - Block in any way
+
+constexpr std::size_t HANDHELD_INDEX = 8;
+
+void HandheldOnProfileSelect() {
+    Settings::values.players[HANDHELD_INDEX].connected = true;
+    Settings::values.players[HANDHELD_INDEX].type = Settings::ControllerType::DualJoycon;
+
+    for (std::size_t player = 0; player < HANDHELD_INDEX; ++player) {
+        Settings::values.players[player].connected = false;
+    }
+
+    Settings::values.use_docked_mode = false;
+    Settings::values.keyboard_enabled = false;
+    Settings::values.mouse_enabled = false;
+    Settings::values.debug_pad_enabled = false;
+    Settings::values.touchscreen.enabled = true;
+}
+
+void DualJoyconsDockedOnProfileSelect() {
+    Settings::values.players[0].connected = true;
+    Settings::values.players[0].type = Settings::ControllerType::DualJoycon;
+
+    for (std::size_t player = 1; player <= HANDHELD_INDEX; ++player) {
+        Settings::values.players[player].connected = false;
+    }
+
+    Settings::values.use_docked_mode = true;
+    Settings::values.keyboard_enabled = false;
+    Settings::values.mouse_enabled = false;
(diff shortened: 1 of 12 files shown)


src/yuzu/configuration/configure_input_simple.cpp
```diff
@@ -0,0 +1,130 @@
...
+    Settings::values.touchscreen.enabled = false;
+}
+
+// Name, OnProfileSelect (called when selected in drop down), OnConfigure (called when configure
+// is clicked)
+using InputProfile = std::tuple<const char*, std::function<void(ConfigureInputSimple*)>,
+                                std::function<void(ConfigureInputSimple*)>>;
+
+const std::array<InputProfile, 3> INPUT_PROFILES{{
+    {"Single Player - Handheld - Undocked", HandheldOnProfileSelect,
```
lioncash: These strings should be translatable

src/yuzu/configuration/configure_input_simple.cpp
```diff
@@ -0,0 +1,130 @@
...
+    connect(ui->profile_configure, &QPushButton::pressed, this, &ConfigureInputSimple::OnConfigure);
+
+    this->loadConfiguration();
+}
+
+void ConfigureInputSimple::applyConfiguration() {
+    auto index = ui->profile_combobox->currentIndex();
+    // Make the stored index for "Custom" very large so that if new profiles are added it
+    // doesn't change.
+    if (std::strncmp(std::get<0>(INPUT_PROFILES.at(index)), "Custom", 6) == 0)
```
lioncash: This is a bad way to determine a custom profile. `Custom` should be a translatable string. Don't use user-facing text as a way to determine layout operation. This violates separation of presentation and implementation.

src/yuzu/configuration/configure_input_simple.cpp
```diff
@@ -0,0 +1,130 @@
...
+        index = std::numeric_limits<uint32_t>::max();
```
lioncash: As explained in another comment, this should be an int.

src/yuzu/ui_settings.h
```diff
@@ -58,6 +58,9 @@ struct Values {
     // logging
     bool show_console;
 
+    // Controllers
+    uint32_t profile_index;
```
lioncash: This should be a regular `int`, since that's what Qt's API uses.

src/yuzu/configuration/configure_input_simple.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    UISettings::values.profile_index = index;
+}
+
+void ConfigureInputSimple::loadConfiguration() {
+    const auto index = UISettings::values.profile_index;
+    if (index >= INPUT_PROFILES.size())
+        ui->profile_combobox->setCurrentIndex(INPUT_PROFILES.size() - 1);
```
lioncash: This should be casted to `int`, otherwise this causes truncation warnings.

src/yuzu/configuration/configure_input_simple.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Ui {
+class ConfigureInputSimple;
+}
+
+class ConfigureInputSimple : public QWidget {
+    Q_OBJECT
+
+public:
+    explicit ConfigureInputSimple(QWidget* parent = nullptr);
```
lioncash: Destructor should also go in the cpp file.

src/yuzu/configuration/configure_input_simple.cpp
```diff
@@ -0,0 +1,130 @@
...
+    caller->applyConfiguration();
+    Dialog dialog(caller, std::forward<Args>(args)...);
+
+    const auto res = dialog.exec();
+    if (res == QDialog::Accepted) {
+        dialog.applyConfiguration();
+    }
+}
+
+/**
```
lioncash: Favor regular `//` comments here. This'll cause warnings with Doxygen, since it's not attached to a valid identifier.

src/yuzu/configuration/configure_input_simple.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include <QDialog>
+
+#include "ui_configure_input_simple.h"
+#include "yuzu/configuration/configure_input.h"
+#include "yuzu/configuration/configure_input_player.h"
+#include "yuzu/configuration/configure_input_simple.h"
+#include "yuzu/ui_settings.h"
+
+template <typename Dialog, typename... Args>
+void CallConfigureDialog(ConfigureInputSimple* caller, Args&&... args) {
```
lioncash: This can be wrapped in an anonymous namespace. Ditto for everything up to and including `INPUT_PROFILES`

src/yuzu/configuration/configure_input_simple.h
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
+
+#include <QWidget>
+
+#include "ui_configure_input_simple.h"
```
lioncash: This doesn't seem to be needed in the header file.

bunnei:
@DarkLordZach reminder to address review comments here

bunnei:
@DarkLordZach also - ping to handle hot swapping handheld <-> docked mode, as this worked previously but does not seem to with these changes

PR #1781, opened 2018-11-23
am: Implement HLE profile selector applet

DarkLordZach:
This implements a HLE version of the ProfileSelect (0x10) applet on the switch.
When executed, this applet presents a list of users to the player and returns the selection or the cancel error code.

I plan to make a setting to show this on every game boot optionally, similar to how a real switch does it, but I felt that was out-of-scope for this PR.

Used by: Undertale, Arcade Archives

### DIFF
--- a/src/core/hle/service/am/applets/profile_select.h
+++ b/src/core/hle/service/am/applets/profile_select.h
@@ -1,0 +1,54 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <string>
+#include <vector>
+
+#include "common/common_funcs.h"
+#include "common/swap.h"
+#include "core/hle/service/acc/profile_manager.h"
+#include "core/hle/service/am/am.h"
+#include "core/hle/service/am/applets/applets.h"
+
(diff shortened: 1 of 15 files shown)


src/core/hle/service/am/applets/profile_select.h
```diff
@@ -0,0 +1,54 @@
+// Copyright 2018 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
```
lioncash: Many of these includes aren't necessary in the header.

src/yuzu/applets/profile_select.h
```diff
@@ -0,0 +1,73 @@
...
+class QStandardItem;
+class QStandardItemModel;
+class QTreeView;
+class QVBoxLayout;
+
+class QtProfileSelectionDialog final : public QDialog {
+    Q_OBJECT
+
+public:
+    QtProfileSelectionDialog(QWidget* parent);
```
lioncash: `explicit`

greggameplayer:
It's also needed by Go Vacation but it softlock after choosing our profile

bunnei:
@DarkLordZach can you resolve conflicts? Thanks!

PR #1784, opened 2018-11-23
GPU States: Implement Polygon Offset. This is used in SMO all the time.

marcosvitali:


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -383,6 +383,55 @@ void OpenGLState::ApplyLogicOp() const {
     }
 }
 
+
+void OpenGLState::ApplyPolygonOffset() const {
+
```
lioncash: extraneous newline

PR #1792, opened 2018-11-24
gpu: Rewrite GPU command list processing with DmaPusher class.

bunnei:
Rewrites our GPU command list processing to be a bit closer to how HW works, fixing several bugs along the way. This was based on Nouveau documentation (see https://envytools.readthedocs.io/en/latest/hw/fifo/dma-pusher.html#fifo-dma-pusher for more details).

The motivation for this work was Undertale submitting bogus GPU data - this should now be fixed. I imagine this fixes issues with other games - although this still needs a bit of testing.

This will also make putting GPU command processing on another thread a bit easier, as this more closely reflects a FIFO interface.


src/core/hle/service/nvdrv/devices/nvhost_gpu.cpp
```diff
@@ -146,7 +145,11 @@ u32 nvhost_gpu::SubmitGPFIFO(const std::vector<u8>& input, std::vector<u8>& outp
     std::memcpy(entries.data(), &input[sizeof(IoctlSubmitGpfifo)],
                 params.num_entries * sizeof(Tegra::CommandListHeader));
 
-    Core::System::GetInstance().GPU().ProcessCommandLists(entries);
+    auto& dma_pusher{Core::System::GetInstance().GPU().DmaPusher()};
```
lioncash: This block can likely go in its own function to deduplicate the similar call below

src/video_core/dma_pusher.h
```diff
@@ -0,0 +1,95 @@
...
+ * The DmaPusher class implements DMA submission to FIFOs, providing an area of memory that the
+ * emulated app fills with commands and tells PFIFO to process. The pushbuffers are then assembled
+ * into a "command stream" consisting of 32-bit words that make up "commands".
+ * See https://envytools.readthedocs.io/en/latest/hw/fifo/dma-pusher.html#fifo-dma-pusher for
+ * details on this implementation.
+ */
+class DmaPusher {
+public:
+    explicit DmaPusher(GPU& gpu);
+    ~DmaPusher() = default;
```
lioncash: This should likely go in the cpp file, considering the class contains a non-trivial type.

src/video_core/dma_pusher.cpp
```diff
@@ -0,0 +1,104 @@
...
+// Refer to the license.txt file included.
+
+#include "core/core.h"
+#include "core/memory.h"
+#include "video_core/dma_pusher.h"
+#include "video_core/gpu.h"
+
+namespace Tegra {
+
+DmaPusher::DmaPusher(GPU& gpu) : gpu(gpu), ib_enable(true) {}
```
lioncash: `true` can be assigned directly in the class declaration

src/core/hle/service/nvdrv/devices/nvhost_gpu.cpp
```diff
@@ -146,7 +145,11 @@ u32 nvhost_gpu::SubmitGPFIFO(const std::vector<u8>& input, std::vector<u8>& outp
...
+    for (const auto& entry : entries) {
```
degasus: This loop was quite hot in the old code. Is there a reason why it shall be here instead of in the DmaPusher class?

src/video_core/dma_pusher.h
```diff
@@ -0,0 +1,95 @@
...
+        u32 length_pending;    ///< Large NI command length pending
+        bool non_incrementing; ///< Current command�s NI flag
+    };
+
+    DmaState dma_state{};
+    bool dma_increment_once{};
+
+    GPUVAddr dma_put{}; ///< pushbuffer current end address
+    GPUVAddr dma_get{}; ///< pushbuffer current read address
+    u64 dma_mget{};     ///< main pushbuffer last read address
```
Subv: Why is this one not a GPUVaddr?

src/core/hle/service/nvdrv/devices/nvhost_gpu.cpp
bunnei: I can move it, just curious why it would make a difference?

degasus:
The conflicts with #1723 seems to be resolved well. OnMemoryWrite is called once per SVE call.

src/video_core/dma_pusher.h
```diff
@@ -0,0 +1,95 @@
...
+ * See https://envytools.readthedocs.io/en/latest/hw/fifo/dma-pusher.html#fifo-dma-pusher for
+ * details on this implementation.
+ */
+class DmaPusher {
+public:
+    explicit DmaPusher(GPU& gpu);
+    ~DmaPusher();
+
+    void Push(const CommandListHeader& command_list_header) {
+        dma_pushbuffer.push(command_list_header);
```
degasus: Is there a way to implement this without copying the data? Here, std::queue::push is called about 100k times per frame. What do you think about a queue of vectors, and std::move for passing the vector from the caller?

src/core/hle/service/nvdrv/devices/nvhost_gpu.cpp
bunnei: Done

src/video_core/dma_pusher.h
bunnei: Done

src/video_core/dma_pusher.cpp
bunnei: Done

src/video_core/dma_pusher.h
bunnei: Done

src/video_core/gpu.cpp
```diff
@@ -113,4 +123,52 @@ u32 DepthFormatBytesPerPixel(DepthFormat format) {
     }
 }
 
+enum class BufferMethods {
+    BindObject = 0,
+    CountBufferMethods = 0x40,
+};
+
+MICROPROFILE_DEFINE(ProcessCommandLists, "GPU", "Execute command buffer", MP_RGB(128, 128, 192));
```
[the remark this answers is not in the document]
bunnei: Done

src/core/hle/service/nvdrv/devices/nvhost_gpu.cpp
bunnei: Done

src/video_core/dma_pusher.h
bunnei: Done

PR #1795, opened 2018-11-24
video_core: Minor style changes

ReinUsesLisp:
This removes unused structs and booleans from the rasterizer and adds the required dependency on `ARB_direct_state_access`, aside from two style changes in `gl_rasterizer_cache`.


bunnei:
@ReinUsesLisp needs a force push
@FernandoS27 can you elaborate more - why aren't we ready for DSA?

FernandoS27:
@bunnei from Rodrigo's example it seems we were actualy using DSA at some example. Now I know why yuzu crashed some intel GPUs before.

PR #1796, opened 2018-11-25
video_core: Move morton functions out of gl_rasterizer_cache

ReinUsesLisp:
Since morton functions are API agnostic, this moves them to their own file and removes `video_core/utils.h` since it'd be empty.

It changes some of the style used in those functions.


src/video_core/morton.cpp
```diff
@@ -1,13 +1,188 @@
-// Copyright 2014 Citra Emulator Project
+// Copyright 2018 yuzu Emulator Project
```
ReinUsesLisp: This was not intended. `git` detected it as a file move.

PR #1799, opened 2018-11-25
Make Unimplemented Messages Log Error but not halt execution

FernandoS27:


lioncash:
Disagree with this. If you want execution to continue, do that change locally. The default behavior should not be to continue and ignore. The macros are in a file called `assert.h` for a reason.

This makes it significantly easier to unintentionally overlook unimplemented behavior. Especially if said behavior is not executed frequently enough to be immediately visible in the log output.

FernandoS27:
> Disagree with this. If you want execution to continue, do that change locally. The default behavior should not be to continue and ignore. The macros are in a file called `assert.h` for a reason.
> 
> This makes it significantly easier to unintentionally overlook unimplemented behavior. Especially if said behavior is not executed frequently enough to be immediately visible in the log output.

Well we need a middle ground because before I could look at all the unimplemented stuffs in the shader decompiler, now We can't see anything, nor the user. It's important that those stuffs report to the log but they can continue execution. Maybe a new macro?

wwylele:
I think the existence of `UNIMPLEMENTED` macro itself is inconsistent with other macros. The other three macros `ASSERT`, `DEBUG_ASSERT` and `UNREACHABLE` all carries their *behavioral* meaning in their name (conditional crash, conditional crash debug, and unconditional crash), with additional *semantic* info passed as logging message for `_MSG` macros. However, `UNIMPLEMENTED` carries only semantic info in its name, with no behavioral meaning at all. The unspecified behavior from an API view is the root of this problem: users of the macro interpret the behavior of it in different way, and it actually behaves in an unexpected way to others.

To solve this, `UNIMPLEMENTED` should be simply removed, and replaced with `UNREACHABLE_MSG("Unimplemented")` or `LOG_CRITICAL("Unimplemented")` according to the expected behavior.

lioncash:
UNIMPLEMENTED serves the purpose of separating actual sanitizing asserts from unimplemented features. Quite often, devs need to disable unimplemented feature assertions locally when working on implementations of things, particularly in the graphics code.

If we coalesce both of those under a single macro, then whenever someone needs to blanket-disable unimplemented feature assertions, they end up disabling the sanitizing asserts as well, which is definitely not desirable. Removing `UNIMPLEMENTED` here is a poor solution in that regard.

PR #1801, opened 2018-11-26
Changed logging to be "Log before execution", Added more error logging, all services/svc should now log on some level

ogniK5377:
List of changes in this PR

- All "called" logs have now been moved to the start of the function
- Various function which have error cases now throw LOG_ERROR
- Any services/SVCs which didn't have a log now do

More in depth description of these changes is as follows.

All the calls have been moved before execution to minimize the amount of confusion to what error was caused from what call. We have had multiple instances where we would return an error case and we never hit the log, we also had instances where if we had a HLE issue which caused the emulator to crash, we would not know where the issue came from unless a debugger was attached at the time. With this change we print what svc/service is being called before we do any actual processing/execution.

We had various cases where we would fail silently, this would lead to confusion when trying to track down why certain error codes were thrown and what was exactly broken. Most error cases if not all now throw a LOG_ERROR before returning. This should make it easy to track down why certain games die without warning.

Finally we had a lot of services which didn't log at all such as hwopus DecodeInterleaved and aoc CountAddonContent. We now should be logging for every service and svc as of now.

Further from this pull request we should stick to Log first, Execute after to help minimize confusion with errors and help improve tracking down bugs, issues and crashes.

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -63,6 +63,8 @@ bool IsInsideNewMapRegion(const VMManager& vm, VAddr address, u64 size) {
                                 vm.GetNewMapRegionEndAddress());
 }
 
+const u64 SZ_8GB = 0x200000000;
+
 // Helper function that performs the common sanity checks for svcMapMemory
 // and svcUnmapMemory. This is doable, as both functions perform their sanitizing
 // in the same order.
@@ -75,6 +77,7 @@ ResultCode MapUnmapMemorySanityChecks(const VMManager& vm_manager, VAddr dst_add
 
     if (!Common::Is4KBAligned(src_addr)) {
         LOG_ERROR(Kernel_SVC, "Source address is not aligned to 4KB, 0x{:016X}", src_addr);
+        return ERR_INVALID_SIZE;
     }
 
     if (size == 0) {
@@ -141,12 +144,15 @@ ResultCode MapUnmapMemorySanityChecks(const VMManager& vm_manager, VAddr dst_add
 static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
     LOG_TRACE(Kernel_SVC, "called, heap_size=0x{:X}", heap_size);
 
-    // Size must be a multiple of 0x200000 (2MB) and be equal to or less than 4GB.
-    if ((heap_size & 0xFFFFFFFE001FFFFF) != 0) {
-        LOG_ERROR(
-            Kernel_SVC,
-            "The heap size is not a multiple of 2mb or is greater than 4GB, heap_size=0x{:016X}",
-            heap_size);
+    // Size must be a multiple of 0x200000 (2MB) and be equal to or less than 8GB.
+    if ((heap_size & 0x1FFFFF) != 0) {
+        LOG_ERROR(Kernel_SVC, "The heap size is not a multiple of 2MB, heap_size=0x{:016X}",
+                  heap_size);
+        return ERR_INVALID_SIZE;
+    }
+
+    if (heap_size >= SZ_8GB) {
+        LOG_ERROR(Kernel_SVC, "The heap size is not less than 8GB, heap_size=0x{:016X}", heap_size);
         return ERR_INVALID_SIZE;
     }
 
@@ -1438,15 +1444,17 @@ static ResultCode CreateSharedMemory(Handle* handle, u64 size, u32 local_permiss
                                      u32 remote_permissions) {
     LOG_TRACE(Kernel_SVC, "called, size=0x{:X}, localPerms=0x{:08X}, remotePerms=0x{:08X}", size,
               local_permissions, remote_permissions);
-
-    // Size must be a multiple of 4KB and be less than or equal to
-    // approx. 8 GB (actually (1GB - 512B) * 8)
     if (size == 0) {
         LOG_ERROR(Kernel_SVC, "Size is 0");
+        return ERR_INVALID_SIZE;
     }
-    if ((size & 0xFFFFFFFE00000FFF) != 0) {
-        LOG_ERROR(Kernel_SVC, "Size is not a multiple of 4KB or is greater than 8GB, size={:016X}",
-                  size);
+    if (!Common::Is4KBAligned(size)) {
+        LOG_ERROR(Kernel_SVC, "Size is not aligned to 4KB, 0x{:016X}", size);
+        return ERR_INVALID_SIZE;
+    }
+
+    if (size >= SZ_8GB) {
+        LOG_ERROR(Kernel_SVC, "Size is not less than 8GB, 0x{:016X}", size);
         return ERR_INVALID_SIZE;
     }
 
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -69,37 +69,45 @@ bool IsInsideNewMapRegion(const VMManager& vm, VAddr address, u64 size) {
 ResultCode MapUnmapMemorySanityChecks(const VMManager& vm_manager, VAddr dst_addr, VAddr src_addr,
                                       u64 size) {
     if (!Common::Is4KBAligned(dst_addr) || !Common::Is4KBAligned(src_addr)) {
+        LOG_ERROR(Kernel_SVC, "Invalid address");
         return ERR_INVALID_ADDRESS;
     }
 
     if (size == 0 || !Common::Is4KBAligned(size)) {
+        LOG_ERROR(Kernel_SVC, "Invalid size");
         return ERR_INVALID_SIZE;
     }
 
     if (!IsValidAddressRange(dst_addr, size)) {
+        LOG_ERROR(Kernel_SVC, "size is out of range");
         return ERR_INVALID_ADDRESS_STATE;
     }
 
     if (!IsValidAddressRange(src_addr, size)) {
+        LOG_ERROR(Kernel_SVC, "size is out of range");
         return ERR_INVALID_ADDRESS_STATE;
     }
 
     if (!IsInsideAddressSpace(vm_manager, src_addr, size)) {
+        LOG_ERROR(Kernel_SVC, "Region is out of the address space");
         return ERR_INVALID_ADDRESS_STATE;
     }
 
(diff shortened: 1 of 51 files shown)


src/core/hle/kernel/svc.cpp
```diff
@@ -68,38 +68,68 @@ bool IsInsideNewMapRegion(const VMManager& vm, VAddr address, u64 size) {
...
                                       u64 size) {
-    if (!Common::Is4KBAligned(dst_addr) || !Common::Is4KBAligned(src_addr)) {
+    if (!Common::Is4KBAligned(dst_addr)) {
+        LOG_ERROR(Kernel_SVC, "Destination address is not aligned to 4KB, 0x{:016X}", dst_addr);
         return ERR_INVALID_ADDRESS;
     }
 
-    if (size == 0 || !Common::Is4KBAligned(size)) {
+    if (!Common::Is4KBAligned(src_addr)) {
+        LOG_ERROR(Kernel_SVC, "Source address is not aligned to 4KB, 0x{:016X}", src_addr);
```
DarkLordZach: shouldn't this return ERR_INVALID_SIZE?

src/core/hle/kernel/svc.cpp
```diff
@@ -1285,18 +1441,31 @@ static ResultCode CreateSharedMemory(Handle* handle, u64 size, u32 local_permiss
 
     // Size must be a multiple of 4KB and be less than or equal to
     // approx. 8 GB (actually (1GB - 512B) * 8)
-    if (size == 0 || (size & 0xFFFFFFFE00000FFF) != 0) {
+    if (size == 0) {
+        LOG_ERROR(Kernel_SVC, "Size is 0");
```
DarkLordZach: shouldn't this also return ERR_INVALID_SIZE

src/core/hle/kernel/svc.cpp
```diff
@@ -1285,18 +1441,31 @@ static ResultCode CreateSharedMemory(Handle* handle, u64 size, u32 local_permiss
...
+    }
+    if ((size & 0xFFFFFFFE00000FFF) != 0) {
```
DarkLordZach: Is it possible to use Common::Is4KBAligned and size > 8GB, to make this easier to read?

src/core/hle/service/am/am.cpp
```diff
@@ -566,69 +562,73 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
     }
 
     void PopOutData(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Service_AM, "called");
+
         IPC::ResponseBuilder rb{ctx, 2, 0, 1};
 
         const auto storage = applet->GetBroker().PopNormalDataToGame();
         if (storage == nullptr) {
+            LOG_ERROR(Service_AM, "storage is a nullptr. There is no data in the current channel");
```
DarkLordZach: specify that it is the normal channel

src/core/hle/service/am/am.cpp
```diff
@@ -638,51 +638,51 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
...
     }
 
     void PopInteractiveOutData(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Service_AM, "called");
+
         IPC::ResponseBuilder rb{ctx, 2, 0, 1};
 
         const auto storage = applet->GetBroker().PopInteractiveDataToGame();
         if (storage == nullptr) {
+            LOG_ERROR(Service_AM, "storage is a nullptr. There is no data in the current channel");
```
DarkLordZach: specify that it is the interactive channel

src/core/hle/service/audio/hwopus.cpp
```diff
@@ -46,10 +46,13 @@ class IHardwareOpusDecoderManager final : public ServiceFramework<IHardwareOpusD
 
 private:
     void DecodeInterleaved(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Audio, "called");
+
         u32 consumed = 0;
         u32 sample_count = 0;
         std::vector<opus_int16> samples(ctx.GetWriteBufferSize() / sizeof(opus_int16));
         if (!Decoder_DecodeInterleaved(consumed, sample_count, ctx.ReadBuffer(), samples)) {
+            LOG_ERROR(Audio, "Failed to decode opus data");
```
DarkLordZach: Is it possible to provide any addt'l info to the user about why/how? Or at least differentiate this from other "Failed to decode opus data" logs?

src/core/hle/service/audio/hwopus.cpp
```diff
@@ -63,12 +66,15 @@ class IHardwareOpusDecoderManager final : public ServiceFramework<IHardwareOpusD
...
     void DecodeInterleavedWithPerformance(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Audio, "called");
+
         u32 consumed = 0;
         u32 sample_count = 0;
         u64 performance = 0;
         std::vector<opus_int16> samples(ctx.GetWriteBufferSize() / sizeof(opus_int16));
         if (!Decoder_DecodeInterleaved(consumed, sample_count, ctx.ReadBuffer(), samples,
                                        performance)) {
+            LOG_ERROR(Audio, "Failed to decode opus data");
```
DarkLordZach: ditto

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -62,11 +62,13 @@ class IStorage final : public ServiceFramework<IStorage> {
 
         // Error checking
         if (length < 0) {
+            LOG_ERROR(Service_FS, "Invalid length provided");
```
DarkLordZach: provide length in log

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -62,11 +62,13 @@ class IStorage final : public ServiceFramework<IStorage> {
...
             IPC::ResponseBuilder rb{ctx, 2};
             rb.Push(FileSys::ERROR_INVALID_SIZE);
             return;
         }
         if (offset < 0) {
+            LOG_ERROR(Service_FS, "Invalid offset provided");
```
DarkLordZach: provide offset in log.

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -107,11 +109,13 @@ class IFile final : public ServiceFramework<IFile> {
 
         // Error checking
         if (length < 0) {
+            LOG_ERROR(Service_FS, "Invalid length provided");
```
DarkLordZach: ditto

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -107,11 +109,13 @@ class IFile final : public ServiceFramework<IFile> {
...
             IPC::ResponseBuilder rb{ctx, 2};
             rb.Push(FileSys::ERROR_INVALID_SIZE);
             return;
         }
         if (offset < 0) {
+            LOG_ERROR(Service_FS, "Invalid offset provided");
```
DarkLordZach: ditto

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -138,11 +142,13 @@ class IFile final : public ServiceFramework<IFile> {
 
         // Error checking
         if (length < 0) {
+            LOG_ERROR(Service_FS, "Invalid length provided");
```
DarkLordZach: ditto

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -138,11 +142,13 @@ class IFile final : public ServiceFramework<IFile> {
...
             IPC::ResponseBuilder rb{ctx, 2};
             rb.Push(FileSys::ERROR_INVALID_SIZE);
             return;
         }
         if (offset < 0) {
+            LOG_ERROR(Service_FS, "Invalid offset provided");
```
DarkLordZach: ditto

src/core/hle/kernel/svc.cpp
```diff
@@ -111,8 +144,15 @@ ResultCode MapUnmapMemorySanityChecks(const VMManager& vm_manager, VAddr dst_add
 static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
     LOG_TRACE(Kernel_SVC, "called, heap_size=0x{:X}", heap_size);
 
-    // Size must be a multiple of 0x200000 (2MB) and be equal to or less than 4GB.
-    if ((heap_size & 0xFFFFFFFE001FFFFF) != 0) {
+    // Size must be a multiple of 0x200000 (2MB) and be equal to or less than 8GB.
+    if ((heap_size & 0x1FFFFF) != 0) {
```
lioncash:
```cpp
if ((heap_size % 0x1FFF) != 0)
```

src/core/hle/kernel/svc.cpp
```diff
@@ -63,43 +63,76 @@ bool IsInsideNewMapRegion(const VMManager& vm, VAddr address, u64 size) {
                                 vm.GetNewMapRegionEndAddress());
 }
 
+const u64 SZ_8GB = 0x200000000;
```
lioncash:
```cpp
// 8 GiB
constexpr u64 MAIN_MEMORY_SIZE = 0x200000000;
```

src/core/hle/kernel/svc.cpp
```diff
@@ -145,13 +145,13 @@ static ResultCode SetHeapSize(VAddr* heap_addr, u64 heap_size) {
...
     // Size must be a multiple of 0x200000 (2MB) and be equal to or less than 8GB.
-    if ((heap_size & 0x1FFFFF) != 0) {
+    if ((heap_size % 0x200000) != 0) {
         LOG_ERROR(Kernel_SVC, "The heap size is not a multiple of 2MB, heap_size=0x{:016X}",
                   heap_size);
         return ERR_INVALID_SIZE;
     }
 
-    if (heap_size >= SZ_8GB) {
+    if (heap_size >= 0x200000000) {
```
Nanosync: shouldn't this be referencing MAIN_MEMORY_SIZE?

PR #1803, opened 2018-11-26
kernel: Divide Event into ReadableEvent and WritableEvent

DarkLordZach:
Our current event implementation is not very hardware accurate. Previously, we just had a singular Event class that did everything related to events, but on actual hardware there are two, ReadableEvent and WritableEvent.

From a high-level perspective, these two classes have a very similar relationship compared to that of ServerPort and ClientPort. Sysmodules/services will create an event pair and register the readable event with the kernel. When asked for an event (via and IPC command), sysmodules return the ReadableEvent. Modules can then use the WritableEvent to signal/clear the event.

This addresses part of the issues in #2.


src/core/hle/service/am/am.cpp
```diff
@@ -566,8 +567,8 @@ class ILibraryAppletAccessor final : public ServiceFramework<ILibraryAppletAcces
 
 private:
     void GetAppletStateChangedEvent(Kernel::HLERequestContext& ctx) {
+        applet->GetBroker().SignalStateChanged();
```
Subv: This changes behavior, why?

src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -36,11 +37,9 @@ void SessionRequestHandler::ClientDisconnected(const SharedPtr<ServerSession>& s
     boost::range::remove_erase(connected_sessions, server_session);
 }
 
-SharedPtr<Event> HLERequestContext::SleepClientThread(SharedPtr<Thread> thread,
-                                                      const std::string& reason, u64 timeout,
-                                                      WakeupCallback&& callback,
-                                                      Kernel::SharedPtr<Kernel::Event> event) {
-
+SharedPtr<WritableEvent> HLERequestContext::SleepClientThread(
```
Subv: I'm not sure I like this API. This lets you supply unlinked Readable and Writable events to the waiting machinery.

IMO it'd be better if you just took a WritableEvent and derived the ReadableEvent from it

src/core/hle/service/am/am.cpp
DarkLordZach: This did not change behavior, it just moved the signal to before the getter of the event, as the broker returns a readable event but exposes functions to signal the event. (original loc: https://github.com/yuzu-emu/yuzu/pull/1803/files#diff-c09cae2b75df143eebac772ac5616cf0L570)

src/core/hle/kernel/kernel.cpp
```diff
@@ -17,9 +17,11 @@
 #include "core/hle/kernel/handle_table.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/process.h"
+#include "core/hle/kernel/readable_event.h"
```
lioncash: These includes don't seem to be used.

bunnei:
@DarkLordZach the build failures look real, FYI

bunnei:
Needs a rebase

PR #1811, opened 2018-11-27
yuzu/input: Minor changes

lioncash:
Minor cleanup and organizational changes. It also silences compiler warnings.

### DIFF
--- a/src/yuzu/configuration/configure_input.cpp
+++ b/src/yuzu/configuration/configure_input.cpp
@@ -4,11 +4,9 @@
 
 #include <algorithm>
 #include <memory>
-#include <utility>
-#include <QMenu>
-#include <QMessageBox>
+
 #include <QTimer>
-#include "common/param_package.h"
+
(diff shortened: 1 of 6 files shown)


src/yuzu/configuration/configure_input.cpp
```diff
@@ -4,28 +4,35 @@
...
-#include "ui_configure_mouse_advanced.h"
-#include "ui_configure_touchscreen_advanced.h"
-#include "yuzu/configuration/config.h"
 #include "yuzu/configuration/configure_input.h"
 #include "yuzu/configuration/configure_input_player.h"
 #include "yuzu/configuration/configure_mouse_advanced.h"
 
+namespace {
+template <typename Dialog, typename... Args>
+void CallConfigureDialog(ConfigureInput* parent, Args&&... args) {
```
VPeruS: Maybe use reference for parent to exclude nullptr?

lioncash: Done

PR #1822, opened 2018-11-29
gl_shader_decompiler: Introduce a scoped object and style changes

ReinUsesLisp:
This introduces an object to gain scope that will be automatically returned when it's destroyed (similar to `std::scoped_lock`. It also does some clean up in the code in regards to texture instruction generation. It should use a table, but I'll leave that for later.

Reintroduces the usage of multiple `texture` calls since it was being used in some places but not others. `WriteTexsInstruction` no longer takes `coords` as argument because it sometimes requires more arguments. All these values (`coords` included) have to be declared in the calling function and not the callee.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -1337,15 +1337,7 @@ private:
         regs.SetRegisterToInteger(dest, true, 0, result, 1, 1);
     }
 
-    void WriteTexsInstruction(const Instruction& instr, const std::string& coord,
-                              const std::string& texture) {
-        // Add an extra scope and declare the texture coords inside to prevent
-        // overwriting them in case they are used as outputs of the texs instruction.
-        shader.AddLine('{');
-        ++shader.scope;
-        shader.AddLine(coord);
-        shader.AddLine("vec4 texture_tmp = " + texture + ';');
-
+    void WriteTexsInstruction(const Instruction& instr, const std::string& texture) {
         // TEXS has two destination registers and a swizzle. The first two elements in the swizzle
         // go into gpr0+0 and gpr0+1, and the rest goes into gpr28+0 and gpr28+1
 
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -201,6 +201,34 @@ class ControlFlowAnalyzer {
...
+template <typename T>
+class ShaderScopedScope {
+public:
+    explicit ShaderScopedScope(T& writer, const std::string& begin_expr, std::string end_expr)
+        : writer(writer), end_expr(end_expr) {
+
+        if (begin_expr.empty()) {
+            writer.AddLine('{');
+        } else {
+            writer.AddLine(begin_expr + " {");
```
```cpp
writer.AddLine(begin_expr);
writer.AddLine(" {");
```
(or add member functions that allow appending strings without creating a newline, if you want the brace on the same line) avoids the creation of a temporary string. Avoiding a potential allocation of the same string data.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -201,6 +201,34 @@ class ControlFlowAnalyzer {
     }
 };
 
+template <typename T>
+class ShaderScopedScope {
+public:
+    explicit ShaderScopedScope(T& writer, const std::string& begin_expr, std::string end_expr)
```
lioncash: `begin_expr` can be a `std::string_view` since it's appended immediately on construction.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -201,6 +201,34 @@ class ControlFlowAnalyzer {
...
+        : writer(writer), end_expr(end_expr) {
```
lioncash: end_expr can be std::move'd into the member variable.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -201,6 +201,34 @@ class ControlFlowAnalyzer {
...
+        if (end_expr.empty()) {
+            writer.AddLine('}');
+        } else {
+            writer.AddLine("} " + end_expr + ';');
+        }
+    }
+
+private:
+    T& writer;
+    const std::string end_expr;
```
lioncash: note that making this const will prevent the class from being movable.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -201,6 +201,34 @@ class ControlFlowAnalyzer {
...
+        }
+        ++writer.scope;
+    }
+
+    ~ShaderScopedScope() {
+        --writer.scope;
+        if (end_expr.empty()) {
+            writer.AddLine('}');
+        } else {
+            writer.AddLine("} " + end_expr + ';');
```
lioncash: ditto wrt above and creating temporaries (two in this case instead of one).

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -201,6 +201,34 @@ class ControlFlowAnalyzer {
     }
 };
 
+template <typename T>
+class ShaderScopedScope {
```
lioncash: You likely also want to delete the copy constructor and copy assignment operators:

```cpp
ShaderScopedScope(const ShaderScopedScope&) = delete;

ShaderScopedScope& operator=(const ShaderScopedScope&) = delete;
```

To prevent making copies of the same scope.

PR #1823, opened 2018-11-29
 gl_rasterizer_cache: Fix several surface copy issues.

bunnei:
Several fixes to how the rasterizer cache does surface copies, particularly with non-2D to 2D copies. Fixes some rendering issues with The Legend of Zelda: Breath of the Wild. This might cause issues or performance regressions in other games, so we'll let it soak in Canary a bit - please report back if you notice these.

* Remove BlitSurface and replace with more accurate copy.
* BlitSurface with different texture targets is inherently broken.
* When target is the same, we can just use FastCopySurface.
* Update AccurateCopySurface to flush complete source surface.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -1031,7 +1031,10 @@ void RasterizerCacheOpenGL::AccurateCopySurface(const Surface& src_surface,
                                                 const Surface& dst_surface) {
     const auto& src_params{src_surface->GetSurfaceParams()};
     const auto& dst_params{dst_surface->GetSurfaceParams()};
-    FlushRegion(src_params.addr, dst_params.MemorySize());
+
+    // Flush enough memory for both the source and destination surface
+    FlushRegion(src_params.addr, std::max(src_params.MemorySize(), dst_params.MemorySize()));
+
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1163,7 +1031,7 @@ void RasterizerCacheOpenGL::AccurateCopySurface(const Surface& src_surface,
                                                 const Surface& dst_surface) {
     const auto& src_params{src_surface->GetSurfaceParams()};
     const auto& dst_params{dst_surface->GetSurfaceParams()};
-    FlushRegion(src_params.addr, dst_params.MemorySize());
+    FlushRegion(src_params.addr, src_params.MemorySize());
```
FernandoS27: this may damage cases where the dst texture is bigger than the source such as 3D textures. I suggest you use `std::max(src_params.MemorySize(), dst_params.MemorySize())`

bunnei: There should never be harm in overflushing (other than a slight perf hit), this is effectively just synch'ing up emulated "shared" memory

bunnei: Done

bunnei:
Clang format is unrelated

PR #1824, opened 2018-11-29
gl_rasterizer: Implement a framebuffer cache

ReinUsesLisp:
Instead of reusing the same framebuffer and attaching different images to it, this caches the framebuffer configuration and uses a map-based cache to store and retrieve the framebuffer handles.

I'm not happy with `FramebufferCacheKey`'s layout.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -79,6 +79,26 @@ struct DrawParameters {
     }
 };
 
+struct FramebufferCacheKey {
+    bool is_single_buffer = false;
+    bool stencil_enable = false;
+
+    std::array<GLenum, Maxwell::NumRenderTargets> color_attachments{};
+    std::array<GLuint, Tegra::Engines::Maxwell3D::Regs::NumRenderTargets> colors{};
+    u32 colors_count = 0;
+
+    GLuint zeta = 0;
+
+    auto Tie() const {
+        return std::tie(is_single_buffer, stencil_enable, color_attachments, colors, colors_count,
+                        zeta);
+    }
+
+    bool operator<(const FramebufferCacheKey& rhs) const {
+        return Tie() < rhs.Tie();
+    }
+};
+
 RasterizerOpenGL::RasterizerOpenGL(Core::Frontend::EmuWindow& window, ScreenInfo& info)
     : res_cache{*this}, shader_cache{*this}, emu_window{window}, screen_info{info},
       buffer_cache(*this, STREAM_BUFFER_SIZE) {
@@ -102,9 +122,6 @@ RasterizerOpenGL::RasterizerOpenGL(Core::Frontend::EmuWindow& window, ScreenInfo
     }
 
     OpenGLState::ApplyDefaultState();
-
-    // Create render framebuffer
-    framebuffer.Create();
 
     shader_program_manager = std::make_unique<GLShader::ProgramManager>();
     state.draw.shader_program = 0;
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -487,45 +537,33 @@ void RasterizerOpenGL::ConfigureFramebuffers(OpenGLState& current_state, bool us
...
-                                   depth_surface->Texture().handle, 0);
-            // Clear stencil attachment
-            glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0, 0);
-        }
-    } else {
-        // Clear both depth and stencil attachment
-        glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D, 0,
-                               0);
+        fbkey.zeta = depth_surface->Texture().handle;
+        fbkey.zeta_attachment =
```
degasus: Instead of this GLemum, you can just store the bool `stencil_enable`. This will save 4 bytes (there are 3 padding bytes after is_single_buffer)

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -363,6 +382,39 @@ void RasterizerOpenGL::SetupShaders(GLenum primitive_mode) {
...
+
+    if (fbkey.is_single_buffer) {
+        if (fbkey.single_attachment != GL_NONE) {
+            glFramebufferTexture(GL_DRAW_FRAMEBUFFER, fbkey.single_attachment, fbkey.single_color,
+                                 0);
+        }
+        glDrawBuffer(fbkey.single_attachment);
+    } else {
+        for (std::size_t index = 0; index < Maxwell::NumRenderTargets; ++index) {
+            glFramebufferTexture(GL_DRAW_FRAMEBUFFER,
```
degasus: if `fbkey.colors[index]` is 0, then no texture shall be bound here. So this call might be skip for index 0.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -363,6 +382,39 @@ void RasterizerOpenGL::SetupShaders(GLenum primitive_mode) {
...
+        framebuffer.Create();
+    current_state.draw.draw_framebuffer = framebuffer.handle;
+    current_state.ApplyFramebufferState();
+
+    if (!is_cache_miss)
+        return;
+
+    if (fbkey.is_single_buffer) {
+        if (fbkey.single_attachment != GL_NONE) {
+            glFramebufferTexture(GL_DRAW_FRAMEBUFFER, fbkey.single_attachment, fbkey.single_color,
```
degasus: You can reduce the key size a bit further if you store fbkey.single_color in fbkey.colors[0].

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -79,6 +79,28 @@ struct DrawParameters {
     }
 };
 
+struct FramebufferCacheKey {
+    bool is_single_buffer = false;
+    u32 single_attachment = 0;
```
degasus: GLenum instead of u32?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -363,6 +382,39 @@ void RasterizerOpenGL::SetupShaders(GLenum primitive_mode) {
...
+
+    if (!is_cache_miss)
+        return;
+
+    if (fbkey.is_single_buffer) {
+        if (fbkey.single_attachment != GL_NONE) {
+            glFramebufferTexture(GL_DRAW_FRAMEBUFFER, fbkey.single_attachment, fbkey.single_color,
+                                 0);
+        }
+        glDrawBuffer(fbkey.single_attachment);
```
degasus: If you store `fbkey.colors_count=1` and `fbkey.color_attachments[0] = fbkey.single_attachment`, this function will be identical to the other glDrawBuffers call in the else path.

degasus:
Here, a small present: https://github.com/degasus/yuzu/commit/fbcache

ReinUsesLisp:
The gift was taken.
Originally I thought of the need for a framebuffer invalidation, but afaik the texture cache never removes an entry.

FernandoS27:
I think cache invalidation can be left in a further PR, I highly doubt games recreate the same framebuffer config all the time with different textures.

PR #1827, opened 2018-11-29
gl_rasterizer: Enable clip distances when set in register and in shader

ReinUsesLisp:
Intel's blob driver crops parts of the screen when clip distances are enabled but not set in a shader. This enables them when they are set in the register and in the shader as a workaround for this issue.


bunnei:
Reminder to run clang-format

PR #1829, opened 2018-11-30
 service/set: Implement MakeLanguageCode

lioncash:
This simply converts a provided index into its respective language code. i.e. making a language code from an index.

### DIFF
--- a/src/core/hle/service/set/set.cpp
+++ b/src/core/hle/service/set/set.cpp
@@ -112,7 +112,7 @@ void SET::GetLanguageCode(Kernel::HLERequestContext& ctx) {
 
     IPC::ResponseBuilder rb{ctx, 4};
     rb.Push(RESULT_SUCCESS);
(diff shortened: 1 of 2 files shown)


src/core/hle/service/set/set.cpp
```diff
@@ -67,6 +69,22 @@ void SET::GetAvailableLanguageCodes(Kernel::HLERequestContext& ctx) {
...
+    if (index >= available_language_codes.size()) {
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(ERR_INVALID_LANGUAGE);
+        return;
+    }
+
+
+    IPC::ResponseBuilder rb{ctx, 4};
+    rb.Push(RESULT_SUCCESS);
+    rb.Push(static_cast<u64>(available_language_codes[index]));
```
DarkLordZach: Can't this be PushEnum as well?

lioncash: Yep. Somehow I managed to push an older version of the commit in question that still had the cast

PR #1833, opened 2018-12-01
service/fsp_srv: Implement CleanDirectoryRecursively

lioncash:
This is the same behavior-wise as DeleteDirectoryRecursively, with the
only difference being that it doesn't delete the top level directory in
the hierarchy, so given:

```
dir/
  - some_other_dir/
    - File.txt
  - OtherFile.txt
```

The end result is just:

```
dir/
```

### DIFF
--- a/src/core/file_sys/vfs.cpp
+++ b/src/core/file_sys/vfs.cpp
@@ -453,7 +453,27 @@ std::shared_ptr<VfsFile> ReadOnlyVfsDirectory::CreateFile(std::string_view name)
     return nullptr;
 }
 
+std::shared_ptr<VfsFile> ReadOnlyVfsDirectory::CreateFileAbsolute(std::string_view path) {
+    return nullptr;
+}
+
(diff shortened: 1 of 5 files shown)


src/core/file_sys/vfs.cpp
```diff
@@ -435,6 +457,10 @@ bool ReadOnlyVfsDirectory::DeleteSubdirectory(std::string_view name) {
     return false;
 }
 
+bool ReadOnlyVfsDirectory::CleanSubdirectoryRecursive(std::string_view name) {
```
lioncash: @DarkLordZach btw, should DeleteDirectoryRecursive also be here and be made to return false?

DarkLordZach: yes

lioncash: Added it, along with other missing overrides to the interface.

lioncash:
I don't know about games in particular, but it was verified through RE and a small homebrew test with libnx

PR #1842, opened 2018-12-02
yuzu/configuration: Minor clean-up related changes

lioncash:
Narrows the visible scope of functions, remove redundant uses of the private access specifier, and gets rid of a case where a compilation failure could potentially occur.

### DIFF
--- a/src/yuzu/configuration/configure_audio.h
+++ b/src/yuzu/configuration/configure_audio.h
@@ -21,10 +21,9 @@ public:
     void applyConfiguration();
     void retranslateUi();
 
-public slots:
+private:
     void updateAudioDevices(int sink_index);
 
-private:
     void setConfiguration();
     void setOutputSinkFromSinkID();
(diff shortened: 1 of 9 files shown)


VPeruS:
f3253d0, but it's compilation error in both cases, what is the difference?

lioncash:
@VPeruS It is not. Previously the compiler can inline the destructor code (which involves invoking the destructor of the object within the unique_ptr and freeing its memory). This inlining can occur in places even when the type is not known. Defaulting the destructor in the cpp file prevents that and constrains the destruction logic to always be in a place where the type is complete from the compiler's PoV. See [](https://godbolt.org/z/DbzFfb) for an example

VPeruS:
unique_ptr will resolve destructor at compile time or fail compilation(using static_assert in template class + destructor inlining) so it won't trigger UB. But indeed it harder to not include concrete type into .cpp. Thanks!

PR #1847, opened 2018-12-03
Print backtrace on svcBreak

ogniK5377:
When we get an svcBreak we get a backtrace now

### DIFF
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -141,6 +141,8 @@ public:
 
     /// Prepare core for thread reschedule (if needed to correctly handle state)
     virtual void PrepareReschedule() = 0;
+
+    virtual void LogBacktrace() = 0;
 };
 
 } // namespace Core
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -6,8 +6,6 @@
 
 #include <array>
 #include "common/common_types.h"
-#include "common/logging/log.h"
-#include "core/memory.h"
 
 namespace Kernel {
 enum class VMAPermission : u8;
@@ -144,21 +142,13 @@ public:
     /// Prepare core for thread reschedule (if needed to correctly handle state)
     virtual void PrepareReschedule() = 0;
 
-    void LogBacktrace() {
-        VAddr fp = GetReg(29);
-        VAddr lr = GetReg(30);
-        VAddr sp = GetReg(13);
-        VAddr pc = GetPC();
-        LOG_ERROR(Core_ARM, "Backtrace, sp={:016X}, pc={:016X}", sp, pc);
-        for (;;) {
-            LOG_ERROR(Core_ARM, "{:016X}", lr);
-            if (!fp) {
-                break;
-            }
-            lr = Memory::Read64(fp + 8) - 4;
-            fp = Memory::Read64(fp);
-        }
-    }
+    /// fp (= r29) points to the last frame record.
+    /// Note that this is the frame record for the *previous* frame, not the current one.
+    /// Note we need to subtract 4 from our last read to get the proper address
+    /// Frame records are two words long:
+    /// fp+0 : pointer to previous frame record
+    /// fp+8 : value of lr for frame
+    void LogBacktrace();
 };
 
 } // namespace Core
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -6,6 +6,8 @@
 
 #include <array>
 #include "common/common_types.h"
+#include "common/logging/log.h"
+#include "core/memory.h"
 
 namespace Kernel {
 enum class VMAPermission : u8;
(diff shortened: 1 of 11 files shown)


src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,23 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
...
+    VAddr sp = system.ArmInterface(processor_id).GetReg(13);
+    VAddr pc = system.ArmInterface(processor_id).GetPC();
+    LOG_ERROR(Debug, "Backtrace, sp={:016X}, pc={:016X}", sp, pc);
+    for (std::size_t i = 0; i < 256; i++) {
+        LOG_ERROR(Debug, "{:016X}", lr - 4);
+        if (!fp) {
+            break;
+        }
+        lr = Memory::Read64(fp + 8);
+        fp = Memory::Read64(fp);
```
lioncash: You need to subtract this by 4 to get the correct address.

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,23 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
     return wakeup_callback(reason, std::move(thread), std::move(object), index);
 }
 
+void Thread::LogBacktrace() {
```
lioncash: This should go in the ARM_Interface to avoid the need to use the System object here, especially given the bulk of the behavior uses the ARM interface.

src/core/hle/kernel/thread.cpp
```diff
@@ -388,6 +388,23 @@ bool Thread::InvokeWakeupCallback(ThreadWakeupReason reason, SharedPtr<Thread> t
...
+    auto& system = Core::System::GetInstance();
+    VAddr fp = system.ArmInterface(processor_id).GetReg(29);
+    VAddr lr = system.ArmInterface(processor_id).GetReg(30);
+    VAddr sp = system.ArmInterface(processor_id).GetReg(13);
+    VAddr pc = system.ArmInterface(processor_id).GetPC();
+    LOG_ERROR(Debug, "Backtrace, sp={:016X}, pc={:016X}", sp, pc);
+    for (std::size_t i = 0; i < 256; i++) {
```
lioncash: Magic number

src/core/hle/kernel/thread.h
```diff
@@ -240,6 +240,11 @@ class Thread final : public WaitObject {
         return status == ThreadStatus::WaitSynchAll;
     }
 
+    /**
+     * Logs the backtrace for the current thread
+     */
+    void LogBacktrace();
```
Subv: This function is misleading. When called on a non-running thread it will print the backtrace of its core's current thread instead of the requested one. IMO this should either be fixed, or the function removed.

src/core/arm/unicorn/arm_unicorn.cpp
```diff
@@ -266,6 +267,22 @@ void ARM_Unicorn::ClearExclusiveState() {}
 
 void ARM_Unicorn::ClearInstructionCache() {}
 
+void ARM_Unicorn::LogBacktrace() {
```
Subv: What's up with the code duplication here and in ARM_Dynarmic? Put this in ARM_Interface::LogBacktrace

src/core/hle/kernel/thread.h
lioncash: Agree with removing it. The underlying function should be invoked directly in svc.cpp

src/core/arm/arm_interface.h
```diff
@@ -141,6 +143,22 @@ class ARM_Interface : NonCopyable {
 
     /// Prepare core for thread reschedule (if needed to correctly handle state)
     virtual void PrepareReschedule() = 0;
+
+    void LogBacktrace() {
```
MerryMage: Implementation should probably go in the `.cpp` file.

Please also document the frame record format:
```
// fp (= r29) points to the last frame record.
// Note that this is the frame record for the *previous* frame, not the current one.
// Frame records are two words long:
// fp+0 : pointer to previous frame record
// fp+8 : last value of lr
```

src/core/arm/arm_interface.h
```diff
@@ -141,6 +143,22 @@ class ARM_Interface : NonCopyable {
...
+        VAddr lr = GetReg(30);
+        VAddr sp = GetReg(13);
+        VAddr pc = GetPC();
+        LOG_ERROR(Core_ARM, "Backtrace, sp={:016X}, pc={:016X}", sp, pc);
+        for (;;) {
+            LOG_ERROR(Core_ARM, "{:016X}", lr);
+            if (!fp) {
+                break;
+            }
+            lr = Memory::Read64(fp + 8) - 4;
```
MerryMage: What's the `- 4` for?

bunnei:
@ogniK5377 can you address the above comments so that we can get this merged soon? Thank you.

src/core/arm/arm_interface.h
ogniK5377: We need to subtract 4 to get the proper address

PR #1848, opened 2018-12-03
game_list: Remove a reference of a reference

FreddyFunk:


FearlessTobi:
I'm not sure this is correct. Don't merge yet while I try to find the reason.

FearlessTobi:
Found the reason: https://github.com/citra-emu/citra/pull/4223.

PR #1849, opened 2018-12-03
svc: Implement SetThreadActivity (thread suspension)

encounter:
Implementation of SetThreadActivity.

Adds a new `Paused` thread status, which is essentially `Ready` but with `activity == 1`.
Allows waits to complete before entering `Paused` status.
Permits threads to be paused before started.

Emulator behavior matches console in my test application.

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -851,8 +851,35 @@ static ResultCode GetInfo(u64* result, u64 info_id, u64 handle, u64 info_sub_id)
 }
 
 /// Sets the thread activity
-static ResultCode SetThreadActivity(Handle handle, u32 unknown) {
-    LOG_WARNING(Kernel_SVC, "(STUBBED) called, handle=0x{:08X}, unknown=0x{:08X}", handle, unknown);
+static ResultCode SetThreadActivity(Handle handle, u32 activity) {
+    LOG_DEBUG(Kernel_SVC, "called, handle=0x{:08X}, activity=0x{:08X}", handle, activity);
+    if (activity > static_cast<u32>(ThreadActivity::Paused)) {
+        return ERR_INVALID_ENUM_VALUE;
+    }
+
+    const auto* current_process = Core::CurrentProcess();
+    const SharedPtr<Thread> thread = current_process->GetHandleTable().Get<Thread>(handle);
+    if (!thread) {
+        LOG_ERROR(Kernel_SVC, "Thread handle does not exist, handle=0x{:08X}", handle);
+        return ERR_INVALID_HANDLE;
+    }
+
+    if (thread->GetOwnerProcess() != current_process) {
+        LOG_ERROR(Kernel_SVC,
+                  "The current process does not own the current thread, thread_handle={:08X} "
+                  "thread_pid={}, "
+                  "current_process_pid={}",
+                  handle, thread->GetOwnerProcess()->GetProcessID(),
+                  current_process->GetProcessID());
+        return ERR_INVALID_HANDLE;
+    }
+
+    if (thread == GetCurrentThread()) {
+        LOG_ERROR(Kernel_SVC, "The thread handle specified is the current running thread");
+        return ERR_BUSY;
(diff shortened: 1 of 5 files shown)


Subv:
Could you please upload the test homebrew somewhere? thanks!

src/core/hle/kernel/thread.h
```diff
@@ -60,6 +61,11 @@ enum class ThreadWakeupReason {
     Timeout // The thread was woken up due to a wait timeout.
 };
 
+enum ThreadActivity : u32 {
```
Subv: Please make this an enum class

src/yuzu/debugger/wait_tree.cpp
```diff
@@ -261,6 +264,7 @@ QColor WaitTreeThread::GetColor() const {
     case Kernel::ThreadStatus::Running:
         return QColor(Qt::GlobalColor::darkGreen);
     case Kernel::ThreadStatus::Ready:
+    case Kernel::ThreadStatus::Paused:
```
Subv: IMO this should use a different color

encounter:
Sure, will have time to get that up later today.

src/core/hle/kernel/svc.cpp
```diff
@@ -679,8 +679,24 @@ static ResultCode GetInfo(u64* result, u64 info_id, u64 handle, u64 info_sub_id)
...
+    if (!thread || thread->GetOwnerProcess() != current_process) {
+        return ERR_INVALID_HANDLE;
+    }
+
+    if (thread == GetCurrentThread()) {
+        // TODO(encounter): is this right?
+        return ERR_INVALID_COMBINATION;
+    }
+
+    thread->SetActivity(static_cast<ThreadActivity>(activity));
```
Subv: This call could cause a thread running on another core to become "Paused". You have to invoke the scheduler here (or inside SetActivity), otherwise the thread might continue running with an invalid state, breaking quite a few of the scheduler invariants.

encounter: Sure, I was thinking that may be an issue. So essentially just calling `Cpu::PrepareReschedule` when I update a thread that's currently `ThreadStatus::Running`?

encounter:
I haven't found examples of official applications/games that use it yet. However, [Unity uses bdwgc](https://github.com/ivmai/bdwgc/commit/97182297e6c21021091710a29e1a0ad83ba81ac3) (garbage collector) with the Switch, and I believe they implement `GC_{stop,start}_world` using this.

src/core/hle/kernel/svc.cpp
Subv: Yeah

src/core/hle/kernel/thread.h
```diff
@@ -370,6 +376,12 @@ class Thread final : public WaitObject {
         return affinity_mask;
     }
 
+    void SetActivity(ThreadActivity value);
+
+    ThreadActivity GetActivity() const {
```
lioncash: The getter should be placed before the setter

src/core/hle/kernel/svc.cpp
```diff
@@ -679,8 +679,24 @@ static ResultCode GetInfo(u64* result, u64 info_id, u64 handle, u64 info_sub_id)
...
+
+    const auto* current_process = Core::CurrentProcess();
+    const SharedPtr<Thread> thread = current_process->GetHandleTable().Get<Thread>(handle);
+    if (!thread || thread->GetOwnerProcess() != current_process) {
+        return ERR_INVALID_HANDLE;
+    }
+
+    if (thread == GetCurrentThread()) {
+        // TODO(encounter): is this right?
+        return ERR_INVALID_COMBINATION;
```
lioncash: It is not. This should return `ERR_BUSY`, which would be a new error defined as:

```cpp
constexpr ResultCode ERR_BUSY{ErrorModule::Kernel, 10};
```

in kernel/errors.h, since we don't already have it.

bunnei:
Thanks for the contribution!

PR #1858, opened 2018-12-04
svc: Return success from SleepThread

encounter:
Otherwise returns some bogus value (wonder where 0x3B9ACA00 comes from?)


lioncash:
svcSleepThread doesn't actually return a result in the kernel, so this is incorrect behavior.

encounter:
Hm, okay. I guess the underlying difference is the lack of register clears when returning from SVC calls?

Either way, I'll change libnx instead.

lioncash:
I've also just updated SwitchBrew to rectify the documentation there as well.

PR #1859, opened 2018-12-04
Convert high-frequency LUT arrays from constexpr to static constexpr

heapo:
[before_disassem.txt](https://github.com/yuzu-emu/yuzu/files/2645628/before_disassem.txt)
[after_disassem.txt](https://github.com/yuzu-emu/yuzu/files/2645627/after_disassem.txt)

While debugging, I noticed that msvc is emitting pretty inefficient code in some of the VideoCore::Surface look-up tables (GetCompressionFactor, GetDefaultBlockWidth), etc. MSVC is actually constructing the LUT on the stack upon function invocation (see attached 'before' disassembly), obviously inefficient in many respects. These LUTs are a pretty hot code path in SurfaceParams::InitCacheParameters, and I measure the overhead as about 5% of BOTW's CPU use (on yuzu's main thread).

One way to fix this is to mark the arrays as static constexpr, the lifetime guarantee allows MSVC to compile them into the exe itself. MSVC could theoretically do this on its own even just with constexpr, but its optimizer is apparently having trouble knowing that is safe to do.

'static constexpr' arrays is discussed at the following link and looks like a good approach, but open to any other approach that improves the codegen: https://stackoverflow.com/questions/13865842/does-static-constexpr-variable-make-sense.

With the change these LUT functions are all inlined, and are significantly shrunk. This improves FPS and cpu use on BOTW for me by ~5% (in canary), any additional data points would be appreciated too.


heapo:
Other compilers don't like this, looking..

bunnei:
Nice change, thanks! I'm OK with this - I assume we cannot move these tables to the .CPP file without losing the fast inlining & constexpr-ness?

lioncash:
Looks good to me. Only nitpick is that `static` isn't required on file-scope const/constexpr variables, since they internally link by default

heapo:
Avoiding static variables where possible seems good, updated (and double-checked disassem)

PR #1862, opened 2018-12-05
gl_shader_decompiler: TLDS/TLD4/TLD4S Reworked reflecting the source registers, bugs fixed and modularize.

marcosvitali:
TLD4S: (textureGather) https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/textureGather.xhtml

	Fix DC cases, actually is using vec3 and returning a float but should be use vec2 and return vec4
	Fix DC cases refZ should be use float gr20 and not component integer from immediate
	
TLD4 (textureGather) 

	Fix DC cases, actually is using vec3 and returning a float but should be use vec2 and return vec4
	Fix DC cases refZ should be use float gr20 and not component integer from immediate
	Implement all texture types reflecting teh source registers
	
TLDS (texelFetch) https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/texelFetch.xhtml

	Fix LL cases, Lod is always in gr20 and not in gr20 + offset
	FIX Coord Y when Lod is used is in gr8 + 1 and not in gr20 
	Implement all texture types reflecting the source registers

TEXS 

	Small fix to prevent add ',' in 1D last coord
	
Update the google worksheet with TLDS/TLD4/TLD4S tabs with test cases

https://docs.google.com/spreadsheets/d/1aVzoCYi9QB_2n-6jcIV9J_paDY-FqdZzhfRCap_dkYQ

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -1681,7 +1681,7 @@ private:
         for (size_t i = 0; i < coord_count; ++i) {
             const bool last = (i == (coord_count - 1)) && (coord_count > 1);
             coord += regs.GetRegisterAsFloat(last ? last_coord_register : coord_register + i);
-            if (!last) {
+            if (i < coord_count - 1) {
                 coord += ',';
             }
         }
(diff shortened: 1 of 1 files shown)


bunnei:
@marcosvitali can you rebase? (then I'll tag canary). Thank you!

marcosvitali:
> @marcosvitali can you rebase? (then I'll tag canary). Thank you!

Done.

PR #1871, opened 2018-12-05
yuzu/wait_tree: Pass QString by value and std::move in the initializer list for WaitTreeText

lioncash:
Just a trivial modernization that potentially avoids copying strings in certain scenarios


VPeruS:
When second copy happens?

lioncash:
I meant to phrase that as avoiding single copies in certain scenarios. Amended the description.

VPeruS:
thx, for clarification

PR #1877, opened 2018-12-06
Perf: Avoid (expensive) audio interpolation when sample rates already match

heapo:
Also removes redundant lanczos calculation, which was being calculated twice (surprised the compiler didn't eliminate that).

In BOTW, the vast majority of samples don't need the interpolation, (Interpolation goes from ~4% of CPU time to basically nothing), but if that's an outlier then we should move to a faster sine approximation when interpolation is needed. That's most of the remaining cost.

By code inspection skipping interpolation when sample rates match appears (and sounds) correct, but not very familiar with this logic. At the least a fast-path should be possible for 1.0 ratios.


MerryMage:
> By code inspection skipping interpolation when sample rates match appears (and sounds) correct, but not very familiar with this logic. At the least a fast-path should be possible for 1.0 ratios.

It should be fine. The only difference would be a 5 sample delay which is imperceptable.

PR #1878, opened 2018-12-06
Workaround for BOTW (and others?) deadlock on boot

heapo:
The latest merges into canary result in a deadlock when booting BOTW at the least, looks caused by:

kernel/svc: Correct behavior of svcResetSignal() #1861
https://github.com/yuzu-emu/yuzu/pull/1861

This is a targeted workaround, I'm not sure why this fixes it or if the code is even in error (maybe this reveals another bug), but hopefully useful for merging into canary to unblock others.

### DIFF
--- a/src/core/hle/kernel/readable_event.cpp
+++ b/src/core/hle/kernel/readable_event.cpp
@@ -36,7 +36,8 @@ void ReadableEvent::Clear() {
 
 ResultCode ReadableEvent::Reset() {
     if (!signaled) {
-        return ERR_INVALID_STATE;
+        // Returning ERR_INVALID_STATE here causes downstream deadlocks, unclear why.
+        return RESULT_SUCCESS;
     }
 
     Clear();


lioncash:
The behavior of the function itself matches the behavior of the Kernel, so this is likely indicative of a bug elsewhere.

lioncash:
Ah. I don't know if it'll fix it, but try enclosing the body of `Signal()` with 

```cpp
if (!signaled) {
 // body here
}
```

(the kernel only allows signaling to fire once, and requires a reset to signal again, so this is an inaccuracy on our part). I wonder if threads are being kicked when they shouldn't be due to signal being called twice or more.

heapo:
No that didn't fix it for me - if it helps when returning ERR_INVALID_STATE Signal is called indefinitely in a tight loop, it's from JIT'd code so kind of stuck. With SUCCESS returned it's only called a handful of times before progressing.

heapo:
(Still required for BOTW to boot for me with latest canary merges)

bunnei:
@heapo can you rebase this? We're seeing some weird canary merge issues...

bunnei:
This has been superseded by #1979

PR #1879, opened 2018-12-07
am: Implement GetSaveDataSize and ExtendSaveData using files

DarkLordZach:
Since we don't (and have no plans to) support real IVFC-compressed saves, this was the best solution we could come up with.

This will store the normal and journal sizes in a file in the save directory called `.yuzu_save_size`, sequentially as u64s. When the game calls Get or Extend, this file will be read from or written to. Initially, this is populated using the game's control metadata, or if its not available using a default value of 4GB for each (which is large enough to work).

Tested against Minecraft.

Supersedes #1110.

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -8,6 +8,7 @@
 #include <stack>
 #include "audio_core/audio_renderer.h"
 #include "core/core.h"
+#include "core/file_sys/savedata_factory.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/process.h"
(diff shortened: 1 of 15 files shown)


bunnei:
@DarkLordZach can you please rebase?

src/core/hle/service/am/am.cpp
```diff
@@ -1037,6 +1038,48 @@ void IApplicationFunctions::GetPseudoDeviceId(Kernel::HLERequestContext& ctx) {
...
+
+    // The following value is used upon failure to help the system recover.
+    // Since we always succeed, this should be 0.
+    rb.Push<u64>(0);
+}
+
+void IApplicationFunctions::GetSaveDataSize(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+    const auto type{rp.PopRaw<FileSys::SaveDataType>()};
+    rp.Skip(1, false);
```
lioncash: Looking into it a little more, shouldn't this also pass in the title ID, making calling `Core::CurrentProcess->GetTitleID()` unnecessary?

i.e. it seems like this is a function somewhat like:

```cpp
fn(u64* out_normal_size, u64* out_journal_size, u64 title_id, u8 save_data_type, const u128* user_id);
```

ExtendSaveData similarly also seems to pass in the title ID.

ogniK5377: I believe title id of 0 can be passed as well which means just to get the current TID

bunnei:
@DarkLordZach needs a rebase

greggameplayer:
@DarkLordZach rebase this PR please

PR #1880, opened 2018-12-07
savedata_factory: Add CacheStorage and delete TemporaryStorage on boot

DarkLordZach:
`TemporaryStorage`: Mimics hardware assertions that data is not persisted.
`CacheStorage`: Allows many new games to boot (1-2-Switch, Minecraft, Rocket League, etc). Done at path `/cache/<title_id>`, which is our best guess for how the system actually does it.

### DIFF
--- a/src/core/file_sys/savedata_factory.cpp
+++ b/src/core/file_sys/savedata_factory.cpp
@@ -18,7 +18,11 @@ std::string SaveDataDescriptor::DebugInfo() const {
                        static_cast<u8>(type), title_id, user_id[1], user_id[0], save_id);
 }
 
-SaveDataFactory::SaveDataFactory(VirtualDir save_directory) : dir(std::move(save_directory)) {}
+SaveDataFactory::SaveDataFactory(VirtualDir save_directory) : dir(std::move(save_directory)) {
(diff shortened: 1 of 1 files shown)


src/core/file_sys/savedata_factory.cpp
```diff
@@ -120,6 +124,8 @@ std::string SaveDataFactory::GetFullPath(SaveDataSpaceId space, SaveDataType typ
     case SaveDataType::TemporaryStorage:
         return fmt::format("{}{:016X}/{:016X}{:016X}/{:016X}", out, 0, user_id[1], user_id[0],
                            title_id);
+    case SaveDataType::CacheStorage:
+        return fmt::format("{}cache/{:016X}", out, title_id);
```
lioncash: I could be incorrect, but isn't the cache dir used for the account cache? I'm fairly positive cache storage is put within the save directory as well

DarkLordZach: I agree, for some reason I thought the /save/ was included in out. Fixed

Hexagon12:
The MinGW error in AppVeyor isn't related to this PR. Merging...

PR #1886, opened 2018-12-08
Port citra-emu/citra#4164: "citra_qt, video_core: Screenshot functionality"

FearlessTobi:
See citra-emu/citra#4164 for more details.

Because some stuff is different in yuzu, I needed to make some changes to the PR in order for it to work here. So please pay close attention while reviewing.

### DIFF
--- a/src/video_core/renderer_base.h
+++ b/src/video_core/renderer_base.h
@@ -9,6 +9,7 @@
 #include <optional>
 
 #include "common/common_types.h"
+#include "core/frontend/emu_window.h"
 #include "video_core/gpu.h"
 #include "video_core/rasterizer_interface.h"
 
@@ -21,6 +22,12 @@ namespace VideoCore {
 struct RendererSettings {
     std::atomic_bool use_framelimiter{false};
     std::atomic_bool set_background_color{false};
+
+    // Screenshot
+    std::atomic<bool> screenshot_requested{false};
+    void* screenshot_bits;
+    std::function<void()> screenshot_complete_callback;
+    Layout::FramebufferLayout screenshot_framebuffer_layout;
 };
 
 class RendererBase : NonCopyable {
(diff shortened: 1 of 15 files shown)


src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -138,7 +138,42 @@ void RendererOpenGL::SwapBuffers(
 
         // Load the framebuffer from memory, draw it to the screen, and swap buffers
         LoadFBToScreenInfo(*framebuffer);
-        DrawScreen();
+
+        auto& renderer_settings = Core::System::GetInstance().Renderer().Settings();
```
lioncash: This whole block should be in its own function. Also, is it necessary to access the system here? This is a class deriving from the base renderer class, so you should be able to access the settings instance directly, you don't need accessor functions for it.

src/video_core/video_core.cpp
```diff
@@ -13,4 +15,11 @@ std::unique_ptr<RendererBase> CreateRenderer(Core::Frontend::EmuWindow& emu_wind
     return std::make_unique<OpenGL::RendererOpenGL>(emu_window);
 }
 
+u16 GetResolutionScaleFactor() {
```
lioncash: This should be taking the renderer instance as a parameter to avoid using the system accessor for it here.

src/yuzu/main.cpp
```diff
@@ -334,6 +334,8 @@ void GMainWindow::InitializeHotkeys() {
                                    Qt::ApplicationShortcut);
     hotkey_registry.RegisterHotkey("Main Window", "Load Amiibo", QKeySequence(Qt::Key_F2),
                                    Qt::ApplicationShortcut);
+    hotkey_registry.RegisterHotkey("Main Window", "Capture Screenshot", QKeySequence(tr("CTRL+P")));
```
```cpp
QKeySequence(QKeySequence::Print)
```

makes this not rely on a translatable string.

Hexagon12:
Could this function be used with the dedicated screenshot button in the input settings, or is this out of scope for this PR?

src/video_core/video_core.cpp
```diff
@@ -13,4 +15,10 @@ std::unique_ptr<RendererBase> CreateRenderer(Core::Frontend::EmuWindow& emu_wind
     return std::make_unique<OpenGL::RendererOpenGL>(emu_window);
 }
 
+u16 GetResolutionScaleFactor(RendererBase& renderer) {
```
lioncash: `renderer` can be a const reference.

src/video_core/renderer_base.h
```diff
@@ -57,9 +64,25 @@ class RendererBase : NonCopyable {
         return *rasterizer;
     }
 
+    Core::Frontend::EmuWindow& GetRenderWindow() const {
```
lioncash:
```cpp
Core::Frontend::EmuWindow& GetRenderWindow() {
    return render_window;
}

const Core::Frontend::EmuWindow& GetRenderWindow() const {
    return render_window;
}
```

FearlessTobi:
Reviews addressed.

PR #1889, opened 2018-12-10
applets: Correct usage of SignalStateChanged event

DarkLordZach:
This was causing some games (most notably Pokemon Quest) to softlock due to an event being fired when not supposed to. This also removes a hack wherein we were firing the state changed event when the game retrieves it, which is incorrect.

Pokemon Quest now:
yuzu_pokemon_quest
(credit to BSoD Gaming for screenshot)

### DIFF
--- a/src/core/hle/service/am/applets/applets.cpp
+++ b/src/core/hle/service/am/applets/applets.cpp
@@ -16,11 +16,11 @@ namespace Service::AM::Applets {
 AppletDataBroker::AppletDataBroker() {
     auto& kernel = Core::System::GetInstance().Kernel();
     state_changed_event = Kernel::WritableEvent::CreateEventPair(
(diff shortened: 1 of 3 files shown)


src/core/hle/service/am/applets/applets.cpp
```diff
@@ -16,11 +16,11 @@ namespace Service::AM::Applets {
 AppletDataBroker::AppletDataBroker() {
     auto& kernel = Core::System::GetInstance().Kernel();
     state_changed_event = Kernel::WritableEvent::CreateEventPair(
-        kernel, Kernel::ResetType::OneShot, "ILibraryAppletAccessor:StateChangedEvent");
+        kernel, Kernel::ResetType::Sticky, "ILibraryAppletAccessor:StateChangedEvent");
```
Subv: Who resets this event?

DarkLordZach:
This also fixes an infinite loop of opening selector windows that occurs in Undertale with #1781.

Issue #1778, opened 2018-11-22
Memory leakage in Pokémon Let's Go Eevee


DarthJahus:
Game: Pokémon Let's Go Eevee

Yuzu consumes 9 GB of RAM when the game is started.
It jumps to 11 GB shortly after the save has been loaded.
From there, it slowly rises the amount of used RAM until it completely fills the available space.
The probably to have a soft lock in the game increases drastically at that time; especially if you _switch_ to another program and come back to Yuzu.
Soft locks don't free RAM.

You can see all the steps in the graphics. Max is 15242 MB.

![image](https://user-images.githubusercontent.com/7353366/48907730-9cd4c780-ee68-11e8-8d4c-ecf61ccbc8d6.png)

---

**System specifications**

* Intel Core i3 6100 @ 3.7 GHz
* Gigabyte Aorus RX 570 4 GB
* 16 GB DDR4
* BioStar TB250-BTC motherboard

**Yuzu version**

* Yuzu Canary 1572 (54fe4df)

Hexagon12:
Renamed the issue to reflect more on the issue's description.

Hexagon12:
@wwylele In this case, we already know that this issue is specifically a memory leak and not just some random "consumption".

wwylele:
@Hexagon12 could you give some more explanation? Because I don't see how I can already know it is leak

Luffykun007:
@Hexagon12 is reported since July this problem!

https://github.com/yuzu-emu/yuzu/issues/674

kevinvanrijn:
Still happening on Canary 1613 with both Pokemon: Let's Go, Eevee and Pokemon: Let's Go, Pikachu though my report might also contain a separate issue.

yuzu eats more and more ram until the game eventually hangs:

![screenshot 2018-11-25 22 38 15](https://user-images.githubusercontent.com/6368561/48985004-7aaaa600-f103-11e8-9f2f-d23f3d83698e.png)

^ The memory leak.

---

After running the game twice yuzu looks like this:

![yuzu-crop](https://user-images.githubusercontent.com/6368561/48985008-84340e00-f103-11e8-85ac-e0085c063852.png)

Note that emulation has stopped but yuzu is still using 10 Gigs of RAM.

---

Closing the game before a hang still results in yuzu not releasing any memory.

Other games release memory properly after stopping emulation.

The only changes I've made to the default settings are disabling single window mode and enabling docked joycons.

FlashMau5:
[...]

PR #1890, opened 2018-12-10
Fix Process object leak on emulation stop

jschmer:
The Process object kept itself alive indefinitely because its handle_table contains a SharedMemory object which owns a reference to the same Process object, creating a circular ownership scenario.

Break that up by storing only a non-owning pointer in the SharedMemory object.

This fixes most of the leak reported by #1778, there are still some megabytes lurking around after stopping emulation though.

This is basically my first time looking at yuzu code so I'm totally unfamiliar with the design decision behind having a owning reference in the SharedMemory class. Maybe there's a better solution?


lioncash:
looks good, though the SharedPtr `owner_process` parameter should be converted to a regular pointer as well

src/core/hle/kernel/svc.cpp
```diff
@@ -1488,8 +1488,8 @@ static ResultCode CreateTransferMemory(Handle* handle, VAddr addr, u64 size, u32
 
     auto& kernel = Core::System::GetInstance().Kernel();
     auto& handle_table = Core::CurrentProcess()->GetHandleTable();
-    const auto shared_mem_handle = SharedMemory::Create(
-        kernel, handle_table.Get<Process>(CurrentProcess), size, perms, perms, addr);
+    const auto& process = handle_table.Get<Process>(CurrentProcess);
```
jschmer: Is the process instance in handle_table the same as Core::CurrentProcess()? Then we could remove the round trip through the handle_table here completely.

lioncash: It is indeed, this is likely a remnant from when the handle table was a global variable as well. It can be replaced with `kernel.CurrentProcess()` (Core::CurrentProcess currently forwards to an equivalent of this, and long-term, it'll end up being removed). This can also apply to the other modification a little further down in the file.

jschmer: Changed to using the kernel object to get the current process. I guess this is good to go in now?

PR #1892, opened 2018-12-11
Improve Zero flag implementation

Tinob:
Originally intended this pull to add support for the rest of the Control flags but I will keep every flag on a separate pull.
This pull improve and centralize the set process of the zero flag

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -405,10 +405,17 @@ public:
      */
     void SetRegisterToFloat(const Register& reg, u64 elem, const std::string& value,
                             u64 dest_num_components, u64 value_num_components,
-                            bool is_saturated = false, u64 dest_elem = 0, bool precise = false) {
+                            bool is_saturated = false, bool sets_cc = false, u64 dest_elem = 0,
+                            bool precise = false) {
 
         SetRegister(reg, elem, is_saturated ? "clamp(" + value + ", 0.0, 1.0)" : value,
                     dest_num_components, value_num_components, dest_elem, precise);
+        if (sets_cc) {
+            const std::string zero_condition =
+                "( " + GetRegister(reg, static_cast<u32>(dest_elem)) + " == 0 )";
+            SetInternalFlag(InternalFlag::ZeroFlag, zero_condition);
+            LOG_WARNING(HW_GPU, "Condition codes implementation is incomplete.");
+        }
     }
 
     /**
@@ -425,8 +432,8 @@ public:
     void SetRegisterToInteger(const Register& reg, bool is_signed, u64 elem,
                               const std::string& value, u64 dest_num_components,
                               u64 value_num_components, bool is_saturated = false,
-                              u64 dest_elem = 0, Register::Size size = Register::Size::Word,
-                              bool sets_cc = false) {
+                              bool sets_cc = false, u64 dest_elem = 0,
+                              Register::Size size = Register::Size::Word) {
         UNIMPLEMENTED_IF(is_saturated);
 
         const std::string func{is_signed ? "intBitsToFloat" : "uintBitsToFloat"};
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1297,7 +1305,7 @@ class GLSLGenerator {
         }
 
         if (dest != Tegra::Shader::Register::ZeroIndex) {
-            regs.SetRegisterToInteger(dest, true, 0, result, 1, 1);
+            regs.SetRegisterToInteger(dest, true, 0, result, 1, 1, false, set_cc);
```
heapo: You've tagged WIP (and if the compiler isn't warning) these parameters are using the old SetRegisterToInteger parameter ordering

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1340,7 +1349,7 @@ class GLSLGenerator {
 
         result += ')';
 
-        regs.SetRegisterToInteger(dest, true, 0, result, 1, 1);
+        regs.SetRegisterToInteger(dest, true, 0, result, 1, 1, false, set_cc);
```
heapo: (here too)

heapo:
Nice approach and great to get rid of some of these unimplementeds

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
Tinob: this is using the new paramter layout, set_cc after saturation

heapo: My mistake, the default parameters threw me off

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -404,10 +404,17 @@ class GLSLRegisterManager {
      */
     void SetRegisterToFloat(const Register& reg, u64 elem, const std::string& value,
                             u64 dest_num_components, u64 value_num_components,
-                            bool is_saturated = false, u64 dest_elem = 0, bool precise = false) {
+                            bool is_saturated = false, bool sets_cc = false, u64 dest_elem = 0,
+                            bool precise = false) {
 
         SetRegister(reg, elem, is_saturated ? "clamp(" + value + ", 0.0, 1.0)" : value,
                     dest_num_components, value_num_components, dest_elem, precise);
+        if (sets_cc) {
```
Subv: This will lead to incorrect behavior if RZ is used as the destination register.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -434,7 +441,8 @@ class GLSLRegisterManager {
                     dest_num_components, value_num_components, dest_elem, false);
 
         if (sets_cc) {
```
Subv: Same here

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
Tinob: What do you think is the right behavior here? If Rz is the destination the state should be updated?

Subv: Yes, I've seen some shaders use `I2I.C RZ, R1` or similar, which basically just sets the flags based on R1. This would not work with the implementation in this PR.

Tinob: should be solved now

bunnei:
@Subv ping to review

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -469,10 +492,10 @@ class GLSLRegisterManager {
                 // pack. I couldn't test this on hardware but it shouldn't really matter since most
                 // of the time when a Mrg_* flag is used both components will be mirrored. That
                 // being said, it deserves a test.
-                return "((" + GetRegisterAsInteger(reg, 0, false) +
+                return "uintBitsToFloat((" + GetRegisterAsInteger(reg, 0, false) +
```
Subv: What's up with this seemingly unrelated change?

PR #1894, opened 2018-12-11
process: Store the extra resource memory size

lioncash:
An NPDM stores an amount of extra memory that can be used by a process which is then able to be utilized with svcMapPhysicalMemory and svcUnmapPhysical memory. This is also relevant for svcCreateProcess, given the config parameters passed to it indicate the number of pages that are available for use in this regard.

This is currently not used, but will be made use of in following changes. This just gets the stuff that touches the filesystem code out of the way. At the very least, this amends svcGetInfo to report the proper size, as opposed to always returning zero.

edit: Hold off on merging this for the time being

### DIFF
--- a/src/core/file_sys/program_metadata.cpp
+++ b/src/core/file_sys/program_metadata.cpp
@@ -63,6 +63,10 @@ u32 ProgramMetadata::GetMainThreadStackSize() const {
     return npdm_header.main_stack_size;
 }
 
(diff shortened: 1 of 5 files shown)


ReinUsesLisp:
@ogniK5377 what's the state of this now that we have physical memory?

ogniK5377:
This PR can be closed now since it was implemented with MapPhysicalMemory

PR #1895, opened 2018-12-11
patch_manager: Prevent use of a dangling pointer within PatchRomFS

lioncash:
fmt::format() returns a std::string instance by value, so calling .c_str() on it here is equivalent to doing:

```cpp
auto* ptr = std::string{}.c_str();
```

The data being pointed to isn't guaranteed to actually be valid anymore after that expression ends. Instead, we can just take the string as is, and provide the necessary formatting parameters.


lioncash:
error isn't related to the changes

PR #1896, opened 2018-12-11
[Experimental] shader_decompiler: Improve Accuracy of IPA instruction

FernandoS27:
In this commit we try to do the perspective correction ourselves instead of letting OpenGL does it. This ressembles NVIDIA's hardware as attributes are interpolated linearly or flat and perspective correction is done on the instruction.

This is marked as experimental as we are corrently hacking our glFragCoord.w value.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -955,6 +955,9 @@ private:
         const Tegra::Shader::IpaInterpMode interp_mode =
             declr_input_attribute[attribute].interpolation_mode;
         std::string out;
+        if (stage != Maxwell3D::Regs::ShaderStage::Fragment) {
+            return out;
+        }
         switch (interp_mode) {
         case Tegra::Shader::IpaInterpMode::Flat: {
             out += "flat ";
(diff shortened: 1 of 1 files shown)


Subv:
Are we sure this properly emulates perspective correction?

FernandoS27:
That's the detail, we don't properly emulate glFragCoord.w. In general, according to nouveau devs this is how it should be done.

marcosvitali:
The experiment seems strange for me. You are only changing perspective by noperspective.
While w = 1.0 multiply gr20 in IPA change nothing. Because you are multiplying * 1.
Common basic example: IPA.PASS = 1.0, MUFU = 1/1, IPA ATTR * 1

    /*0010*/                   IPA.PASS R20, a[0x7c], RZ         } = 1
    /*0018*/                   MOV32I R28, 0x3f000000 ;
    /*0028*/               @P0 KIL ;
    /*0030*/                   MUFU.RCP R20, R20 ; = 1/1 
    /*0038*/                   IPA R10, a[0x80], R20 ; new noperspective attr * 1
    /*0048*/                   IPA R11, a[0x84], R20 ;
    /*0050*/                   IPA R4, a[0x88], R20 ;

FernandoS27:
Exactly, the perspective correction is now done on the shader. The problem is in our emulation of the fragcoord value. So, in theory, this improves accuracy but relies on something we don't emulate correctly.

FernandoS27:
I'll redo this in the future under the new shader_ir. I'll close it.

PR #1901, opened 2018-12-13
Fix Service object leak on emulation stop

jschmer:
Services created with the ServiceFramework base class install themselves as HleHandlers with an owning shared_ptr in the ServerPort ServiceFrameworkBase::port member variable, creating a cyclic ownership between ServiceFrameworkBase and the ServerPort, preventing deletion of the service objects.

Fix that by removing the ServiceFrameworkBase::port member because it looks like that was only used to detect multiple attempts at installing a port. Instead store a flag if the port was already installed to achieve the same functionality. The service will then live as long as the corresponding ServerPort/ClientPort pair.

This addresses the remaining leaks not fixed by #1890

### DIFF
--- a/src/core/hle/service/service.cpp
+++ b/src/core/hle/service/service.cpp
@@ -97,29 +97,33 @@ ServiceFrameworkBase::ServiceFrameworkBase(const char* service_name, u32 max_ses
 ServiceFrameworkBase::~ServiceFrameworkBase() = default;
 
 void ServiceFrameworkBase::InstallAsService(SM::ServiceManager& service_manager) {
-    ASSERT(port == nullptr);
-    port = service_manager.RegisterService(service_name, max_sessions).Unwrap();
+    ASSERT(!port_installed);
+
+    auto port = service_manager.RegisterService(service_name, max_sessions).Unwrap();
     port->SetHleHandler(shared_from_this());
+    port_installed = true;
 }
 
 void ServiceFrameworkBase::InstallAsNamedPort() {
-    ASSERT(port == nullptr);
+    ASSERT(!port_installed);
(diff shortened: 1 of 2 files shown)


src/core/hle/service/service.cpp
```diff
@@ -92,34 +92,39 @@ namespace Service {
 
 ServiceFrameworkBase::ServiceFrameworkBase(const char* service_name, u32 max_sessions,
                                            InvokerFn* handler_invoker)
-    : service_name(service_name), max_sessions(max_sessions), handler_invoker(handler_invoker) {}
+    : service_name(service_name), max_sessions(max_sessions), portInstalled(false),
```
lioncash: `portInstalled` can be assigned to in the class body like:

```cpp
class Thing {
public:
    ...

private:
    bool portInstalled = false;
};
```

src/core/hle/service/service.h
```diff
@@ -96,11 +96,9 @@ class ServiceFrameworkBase : public Kernel::SessionRequestHandler {
...
     u32 max_sessions;
 
-    /**
-     * Port where incoming connections will be received. Only created when InstallAsService() or
-     * InstallAsNamedPort() are called.
-     */
-    Kernel::SharedPtr<Kernel::ServerPort> port;
+    /// Flag to store if a port was already create/installed to detect multiple install attempts,
+    /// which is not supported.
+    bool portInstalled;
```
lioncash: our naming convention is `snake_case`.

jschmer: yeah, should have noticed that myself...

bunnei:
Thank you!

PR #1903, opened 2018-12-15
Implement postfactor multiplication/division for fmul instructions

heapo:
This PR implements Maxwell's support for 'postfactor' fmul instructions, where the result of multiplication is further multiplied by up to 2^postfactor (or divided by 2^-postfactor if postfactor is negative). Postfactor has a range of [-3, 3], and this occurs before the saturation op if applicable.

This has a marginal impact on BOTW's atmospheric rendering, might have a greater impact elsewhere. Used nouveau as reference and validated against LibNX opengl shaders.

(e: There's a merge conflict with Tinob's in-review CC improvements FYI)

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -1867,9 +1867,6 @@ private:
                 UNIMPLEMENTED_IF_MSG(instr.fmul.tab5cb8_2 != 0,
                                      "FMUL tab5cb8_2({}) is not implemented",
                                      instr.fmul.tab5cb8_2.Value());
-                UNIMPLEMENTED_IF_MSG(instr.fmul.tab5c68_1 != 0,
-                                     "FMUL tab5cb8_1({}) is not implemented",
-                                     instr.fmul.tab5c68_1.Value());
                 UNIMPLEMENTED_IF_MSG(
                     instr.fmul.tab5c68_0 != 1, "FMUL tab5cb8_0({}) is not implemented",
                     instr.fmul.tab5c68_0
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1879,7 +1876,25 @@ class GLSLGenerator {
 
                 op_b = GetOperandAbsNeg(op_b, false, instr.fmul.negate_b);
 
-                regs.SetRegisterToFloat(instr.gpr0, 0, op_a + " * " + op_b, 1, 1,
+                std::string postfactor_op;
+                if (instr.fmul.postfactor != 0) {
+                    s8 postfactor = static_cast<s8>(instr.fmul.postfactor);
+
+                    // See CodeEmitterGM107::emitPDIV in Nouveau.
```
Subv: Please leave a better comment than this. We shouldn't need to go to any external codebase to understand what this is doing

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1879,7 +1876,25 @@ class GLSLGenerator {
...
+                    if (postfactor >= 4) {
+                        postfactor = 7 - postfactor;
+                    } else {
+                        postfactor = 0 - postfactor;
+                    }
+
+                    if (postfactor > 0) {
+                        postfactor_op = " * " + std::to_string((1 << postfactor));
```
lioncash: the inner set of parentheses aren't necessary. Ditto for the duplicate pair in the else branch as well.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
heapo: Realized this is just a shortcut for plain 1's complement, updated comment

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
heapo: fixed

heapo:
Squashed this down into one commit to be ready for merge

PR #1904, opened 2018-12-15
gl_rasterizer: Implement a more accurate fermi 2D copy.

bunnei:
- This is a blit, use the blit registers.

### DIFF
--- a/src/video_core/engines/fermi_2d.cpp
+++ b/src/video_core/engines/fermi_2d.cpp
@@ -21,7 +21,9 @@ void Fermi2D::CallMethod(const GPU::MethodCall& method_call) {
     regs.reg_array[method_call.method] = method_call.argument;
 
     switch (method_call.method) {
-    case FERMI2D_REG_INDEX(trigger): {
+    // Trigger the surface copy on the last register write. This is blit_src_y, but this is 64-bit,
+    // so trigger on the second 32-bit write.
+    case FERMI2D_REG_INDEX(blit_src_y) + 1: {
         HandleSurfaceCopy();
         break;
     }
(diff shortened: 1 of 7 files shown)


src/video_core/engines/fermi_2d.cpp
```diff
@@ -32,55 +34,23 @@ void Fermi2D::HandleSurfaceCopy() {
...
-    u32 src_bytes_per_pixel = RenderTargetBytesPerPixel(regs.src.format);
-    u32 dst_bytes_per_pixel = RenderTargetBytesPerPixel(regs.dst.format);
-
-    if (!rasterizer.AccelerateSurfaceCopy(regs.src, regs.dst)) {
-        // All copies here update the main memory, so mark all rasterizer states as invalid.
-        Core::System::GetInstance().GPU().Maxwell3D().dirty_flags.OnMemoryWrite();
+    const u32 src_blit_x1{static_cast<u32>(regs.blit_src_x >> 32)};
+    const u32 src_blit_y1{static_cast<u32>(regs.blit_src_y >> 32)};
+    const u32 src_blit_x2{
+        static_cast<u32>(regs.blit_src_x + regs.blit_dst_width * regs.blit_dud_x >> 32)};
```
heapo: Formatting nit (msvc warning), shift has the lowest precedence of all those operators, put parenthesis around the add/muls to increase readability if that's the intended order?

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1008,24 +1008,155 @@ void RasterizerCacheOpenGL::FastLayeredCopySurface(const Surface& src_surface,
...
+                               GL_TEXTURE_2D, 0, 0);
+        glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
+                               dst_surface->Texture().handle, 0);
+
+        buffers = GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT;
+    }
+
+    glBlitFramebuffer(src_rect.left, src_rect.top, src_rect.right, src_rect.bottom, dst_rect.left,
+                      dst_rect.top, dst_rect.right, dst_rect.bottom, buffers,
+                      buffers == GL_COLOR_BUFFER_BIT ? GL_LINEAR : GL_NEAREST);
```
heapo: Since GL_LINEAR is only valid for non-INT formats (acc. to OpenGL reference), maybe add an assert/comment that this function implicitly won't be used with those? (Or othewise change the filter on the fly).

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1008,24 +1008,155 @@ void RasterizerCacheOpenGL::FastLayeredCopySurface(const Surface& src_surface,
...
+                                   GL_TEXTURE_2D, dst_surface->Texture().handle, 0);
+            glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT, GL_TEXTURE_2D,
+                                   0, 0);
+            break;
+        case SurfaceTarget::TextureCubemap:
+            glFramebufferTexture2D(
+                GL_DRAW_FRAMEBUFFER, GL_COLOR_ATTACHMENT0 + dst_attachment,
+                static_cast<GLenum>(GL_TEXTURE_CUBE_MAP_POSITIVE_X + cubemap_face),
+                dst_surface->Texture().handle, 0);
+            glFramebufferTexture2D(
```
heapo: Might be missing something obvious, but any reason to bind D/S stuff when only blitting color?

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -1008,24 +1008,155 @@ void RasterizerCacheOpenGL::FastLayeredCopySurface(const Surface& src_surface,
...
+                GL_READ_FRAMEBUFFER, GL_COLOR_ATTACHMENT0 + src_attachment,
+                static_cast<GLenum>(GL_TEXTURE_CUBE_MAP_POSITIVE_X + cubemap_face),
+                src_surface->Texture().handle, 0);
+            glFramebufferTexture2D(
+                GL_READ_FRAMEBUFFER, GL_DEPTH_STENCIL_ATTACHMENT,
+                static_cast<GLenum>(GL_TEXTURE_CUBE_MAP_POSITIVE_X + cubemap_face), 0, 0);
+            break;
+        case SurfaceTarget::Texture2DArray:
+            glFramebufferTextureLayer(GL_READ_FRAMEBUFFER, GL_COLOR_ATTACHMENT0 + src_attachment,
+                                      src_surface->Texture().handle, 0, 0);
```
heapo: (Missing background on this function, do non-0 layers need to be handled here?)

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
bunnei: Do we know that this will only be used for blitting color?

bunnei: It's binding it to 0 so that no D/S copying occurs. I'm not 100% sure this is correct; it's carry-over from Citra, but seems to be working fine.

src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
bunnei: This has been in canary for a while and as far as I know, we have not seen this GL error being hit. Note, there are plenty of other error conditions that we are not checking for here and elsewhere. I just assume leave it until we get reports of a game hitting this error, then fix it.

bunnei:
Will leave DSA up to a future commit

PR #1907, opened 2018-12-16
kernel/svc: Implement svcSetMemoryAttribute

lioncash:
Following on the tail of #1900, this implements svcSetMemoryAttribute. This allows setting and unsetting the cached memory attribute from user code. At the sane time, this also lays out all the necessities for properly handling attributes in other contexts as well, however, that can be handled in its own separate PR. This just focuses on the service call.

### DIFF
--- a/src/core/hle/kernel/vm_manager.h
+++ b/src/core/hle/kernel/vm_manager.h
@@ -392,6 +392,19 @@ public:
     ///
     MemoryInfo QueryMemory(VAddr address) const;
 
+    /// Sets an attribute across the given address range.
+    ///
+    /// @param address   The starting address
+    /// @param size      The size of the range to set the attribute on.
+    /// @param mask      The attribute mask
+    /// @param attribute The attribute to set across the given address range
+    ///
+    /// @returns RESULT_SUCCESS if successful
+    /// @returns ERR_INVALID_ADDRESS_STATE if the attribute could not be set.
+    ///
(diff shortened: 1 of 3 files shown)


lioncash:
Failure isn't directly related to the changes

src/core/hle/kernel/vm_manager.h
```diff
@@ -435,6 +541,36 @@ class VMManager final {
...
+    /// @param attribute_mask  The memory attribute mask.
+    /// @param attribute       The memory attributes to compare the individual VMA attributes
+    ///                        against, which is done in the form of:
+    ///                        (vma.attributes & attribute_mask) != attribute.
+    /// @param ignore_mask     The memory attributes to ignore during the check.
+    ///
+    /// @return If successful, returns a tuple containing the memory attributes
+    ///         (with ignored bits specified by ignore_mask unset), memory permissions, and
+    ///         memory state across the memory range.
+    ///         @p
```
bunnei: typo?

lioncash: Not a typo. Without an explicit paragraph tag, Doxygen (and by extension the context/mouseover descriptions in Visual Studio) won't consider the text following it to be part of the same `@return` tag, so the description would just cut off. Amended it to just use its own `@returns` tag, which does the same thing in a more straightforward way.

PR #1911, opened 2018-12-17
Allow assertions to not exit on the release build

goedzo:
Very often the release build just exits on an assertion. This allows the release build to just go on instead of failing. Allows a lot of games to be booted at least, without failing on a not implemented error.

### DIFF
--- a/src/common/assert.h
+++ b/src/common/assert.h
@@ -27,6 +27,7 @@ __declspec(noinline, noreturn)
     exit(1); // Keeps GCC's mouth shut about this actually returning
 }
 
+#ifdef _DEBUG
 #define ASSERT(_a_)                                                                                \
     do                                                                                             \
         if (!(_a_)) {                                                                              \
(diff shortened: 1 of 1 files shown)


lioncash:
The point of the regular assertion macros is that they assert at runtime for release builds. This is why DEBUG_ASSERT exists alongside it. Hiding unimplemented or error conditions by default isn't desirable for development.

Either disable it locally when doing self-made builds or use canary, which disables them.

goedzo:
@lioncash thanks for the explanation. Didn't know this. Especially since there already was an existing debug section in the assert.h present. To me it seems that developers are always running this in debug mode anyway, so the assertion will always be hit. This update still logs everything, so very easy to find what is needed to be implemented in the log file.

Maybe it is a good idea to update the documentation to give eplanation about the canary build, and how to build this instead of release and debug. To me this is something new, and did not read any documentation about this in the build documentation on the homepage. Hence, my code update because it was a pain to debug my code without hitting the asserts all the time.

Jcw87:
@goedzo canary works in its own staging repository [here](https://github.com/yuzu-emu/yuzu-canary).

All pull requests that are tagged with canary-merge get automatically merged into canary.
[This one](https://github.com/yuzu-emu/yuzu/pull/615) disables asserts.

PR #1913, opened 2018-12-17
kernel/thread: Set default fpcr

MerryMage:
Somehow I doubt FPCR is 0 on hardware. Unverified.


heapo:
Really generic feedback but could you leave a comment on where this value is from?

PR #1914, opened 2018-12-18
service/am: Unstub GetAppletResourceUserId

lioncash:
This is supposed to return the current process' ID. (0 indicates an invalid ID for both process IDs and ARU IDs).


lioncash:
Build failure isn't a result of the code itself.

bunnei:
Untagging, this broke Undertale. Are you sure this is right? (and that process_id is a good value?)

lioncash:
Yes it's correct behavior. However the way we were handling the process ID was incorrect, we should be starting from PID 81 for userland processes, but were starting at 10 (which is in the range for kernel initial processes). This is fixed in #1925

bunnei:
Thanks. Re-tagging both as canary and will retest later - hopefully the problem is resolved.

PR #1915, opened 2018-12-18
service/sm: Improve debug log for RegisterService

lioncash:
Now it also indicates the name and max session count. This also gives a name to the unknown bool. This indicates if the created port is supposed to be using light handles or regular handles internally. This is passed to the respective svcCreatePort parameter internally.

### DIFF
--- a/src/core/hle/service/sm/sm.cpp
+++ b/src/core/hle/service/sm/sm.cpp
@@ -145,12 +145,13 @@ void SM::RegisterService(Kernel::HLERequestContext& ctx) {
 
     const std::string name(name_buf.begin(), end);
 
-    const auto unk_bool = static_cast<bool>(rp.PopRaw<u32>());
-    const auto session_count = rp.PopRaw<u32>();
+    const auto is_light = static_cast<bool>(rp.PopRaw<u32>());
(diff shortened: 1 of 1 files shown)


src/core/hle/service/sm/sm.cpp
```diff
@@ -145,10 +145,11 @@ void SM::RegisterService(Kernel::HLERequestContext& ctx) {
 
     const std::string name(name_buf.begin(), end);
 
-    const auto unk_bool = static_cast<bool>(rp.PopRaw<u32>());
+    const auto is_light = static_cast<bool>(rp.PopRaw<u32>());
     const auto session_count = rp.PopRaw<u32>();
 
-    LOG_DEBUG(Service_SM, "called with unk_bool={}", unk_bool);
+    LOG_DEBUG(Service_SM, "called with name={}, max_session_count={}, is_light={}", name,
```
heapo: is "max_session_count" intentional in the debug string, or should maybe variable name be changed?

heapo:
Small naming comment but LGTM

src/core/hle/service/sm/sm.cpp
lioncash: Yes, it's intentional. I've updated the variable name.

PR #1916, opened 2018-12-18
service/service: Make FunctionInfo constructor constexpr

lioncash:
The comment about the constexprness of the constructor causing ICEs was made quite a while ago and doesn't seem to apply anymore on MSVC. This allows all of the function lookup-tables to be constexpr as well. Making this to also see if GCC is finally able to handle this.


bunnei:
@lioncash it seems GCC is not able to :(

lioncash:
Unfortunately =/. GCC seems to think a static_cast is actually a reinterpret_cast (which I'm not sure is a compiler bug, or GCC just being overly-restrictive in how it implements the standard). Will close this for the time being.

PR #1603, opened 2018-10-28
Preliminary implementation of LDG (continuation)

bunnei:
This is a continuation of #1323. Since that PR is pretty old, I preferred to open a new one rather than push there.

This is mostly @DarkLordZach's implementation (see #1323 for details), with improvements to performance. The big changes here are:
* Cache uniform locations, as `glGetProgramResourceIndex` is very slow
* Keep global region UBOs long-lived, as constant resource construction/destruction is slow
* Misc. refactoring and cleanup

From my testing, LDG in Super Mario Odyssey now incurs a fairly minimal performance hit, so this should be Canary ready soon. Please test other games!

image

I'm aware that this should probably use SSBOs, but I do not plan on doing that with this change.

PR #1927, opened 2018-12-21
video_core: Replace gl_shader_decompiler with an IR based decompiler

ReinUsesLisp:
This PR aims to solve two issues with one big bullet:
* to have an API agnostic decompiler and
* to be able to do code analysis on shaders (like constant propagation for LDG).

I know this is a pain to review (as it was also to write), so to split the work here's a check list of reviewed instructions (to ease the track):

- [ ] MOV
- [ ] FMUL
- [ ] FADD
- [ ] MUFU
- [ ] FMNMX
- [ ] RRO
- [ ] HADD2 and HMUL2
- [ ] HADD2 and HMUL2 (immediate)
- [ ] MOV32_IMM
- [ ] FMUL32_IMM
- [ ] FADD32_IMM
- [ ] IADD
- [ ] IADD3
- [ ] ISCADD
- [ ] POPC
- [ ] SEL
- [ ] LOP
- [ ] LOP3
- [ ] IMNMX
- [ ] LEA
- [ ] IADD32I
- [ ] LOP32I
- [ ] BFE
- [ ] BFI
- [ ] I2I
- [ ] I2F
- [ ] F2F
- [ ] F2I
- [ ] FFMA
- [ ] FSET
- [ ] FSETP
- [ ] HSET
- [ ] HSETP
- [ ] HFMA2
- [ ] ISET
- [ ] ISETP
- [ ] LD_A
- [ ] LD_C
- [ ] LD_L
- [ ] ST_A
- [ ] ST_L
- [ ] TEX
- [ ] TEXS
- [ ] TLD4
- [ ] TLD4S
- [ ] TXQ
- [ ] TMML
- [ ] TLDS
- [ ] EXIT
- [ ] KIL
- [ ] MOV_SYS
- [ ] BRA
- [ ] SSY and SYNC
- [ ] PBK and BRK
- [ ] IPA
- [ ] OUT
- [ ] ISBERD
- [ ] DEPBAR
- [ ] PSETP
- [ ] PSET
- [ ] R2P
- [ ] SHR
- [ ] SHL
- [ ] VMAD
- [ ] VSETP
- [ ] XMAD

For this to be in a mergeable state I have to implement the remaining and to have a replacement for #1603. There's also some stuff that's wrong at the moment (like array indexes in textures).

Thanks to Ryujinx's IR that served as a base to implement this (there are major differences but without it I couldn't have written it).


src/video_core/shader/shader_ir.h
```diff
@@ -0,0 +1,764 @@
...
+    /// Creates an u32 immediate
+    Node Immediate(u32 value);
+    /// Creates a s32 immediate
+    Node Immediate(s32 value) {
+        return Immediate(static_cast<u32>(value));
+    }
+    /// Creates a f32 immediate
+    Node Immediate(f32 value) {
+        // TODO(Rodrigo): Replace this with bit_cast when C++20 releases
+        return Immediate(*reinterpret_cast<const u32*>(&value));
```
lioncash: You can do:

```cpp
u32 integral;
std::memcpy(&integral, &value, sizeof(u32));
```

to do the same thing in the meantime without invoking undefined behavior.

lioncash:
Note that `assert.h` appears to be included in many headers that don't actually make use of it.

src/video_core/shader/decode.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include "video_core/engines/shader_header.h"
+#include "video_core/shader/shader_ir.h"
+
+namespace VideoCommon::Shader {
+
+using Tegra::Shader::Instruction;
+using Tegra::Shader::OpCode;
+
+/// Merges exit method of two parallel branches.
+constexpr ExitMethod ParallelExit(ExitMethod a, ExitMethod b) {
```
lioncash: This and `IsSchedInstruction` should be placed within an anonymous namespace, otherwise they don't have internal linkage

src/video_core/shader/shader_ir.h
```diff
@@ -0,0 +1,764 @@
...
+
+    void WriteLogicOperation(BasicBlock& bb, Tegra::Shader::Register dest,
+                             Tegra::Shader::LogicOperation logic_op, Node op_a, Node op_b,
+                             Tegra::Shader::PredicateResultMode predicate_mode,
+                             Tegra::Shader::Pred predicate);
+    void WriteLop3Instruction(BasicBlock& bb, Tegra::Shader::Register dest, Node op_a, Node op_b,
+                              Node op_c, Node imm_lut);
+
+    template <typename... T>
+    inline Node Operation(OperationCode code, const T*... operands) {
```
lioncash: You don't need to mark member functions `inline` if they're defined in the class declaration, they're inline by default.

src/video_core/shader/shader_ir.h
```diff
@@ -0,0 +1,764 @@
...
+    }
+
+private:
+    const Node address;
+};
+
+/// Commentary, can be dropped
+class CommentNode final {
+public:
+    explicit CommentNode(const std::string& text) : text{text} {}
```
lioncash:
```suggestion
    explicit CommentNode(std::string text) : text{std::move(text)} {}
```

src/video_core/shader/shader_ir.h
```diff
@@ -0,0 +1,764 @@
...
+
+    const std::string& GetText() const {
+        return text;
+    }
+
+private:
+    const std::string text;
```
lioncash: This should probably be non-const, otherwise it makes the class non-movable, an will always construct a copy, even if `std::move` is used on an instance of it, ditto for all the other nodes above this comment.

src/video_core/shader/shader_ir.h
```diff
@@ -0,0 +1,764 @@
...
+    }
+
+private:
+    const Node address;
+};
+
+/// Global memory node
+class GmemNode final {
+public:
+    explicit GmemNode(Node address) : address{address} {}
```
lioncash: Was this supposed to be `constexpr` as well? (other nodes above have it)

src/video_core/shader/shader_ir.h
```diff
@@ -0,0 +1,764 @@
...
+    explicit Sampler(std::size_t offset, std::size_t index, Tegra::Shader::TextureType type,
+                     bool is_array, bool is_shadow)
+        : offset{offset}, index{index}, type{type}, is_array{is_array}, is_shadow{is_shadow} {}
+
+    std::size_t GetOffset() const {
+        return offset;
+    }
+
+    u32 GetIndex() const {
+        return static_cast<u32>(index);
```
lioncash: any reason why `std::size_t` isn't just the return type?

src/video_core/shader/shader_ir.h
```diff
@@ -0,0 +1,764 @@
...
+        return text;
+    }
+
+private:
+    const std::string text;
+};
+
+class ShaderIR final {
+public:
+    explicit ShaderIR(const ProgramCode& program_code, u32 main_offset)
```
lioncash:
```suggestion
    explicit ShaderIR(ProgramCode program_code, u32 main_offset)
```

src/video_core/shader/shader_ir.h
```diff
@@ -0,0 +1,764 @@
...
+        : program_code{program_code}, main_offset{main_offset} {
```
lioncash:
```suggestion
        : program_code{std::move(program_code)}, main_offset{main_offset} {
```

src/video_core/shader/shader_ir.cpp
```diff
@@ -0,0 +1,408 @@
...
+    bb.push_back(Operation(OperationCode::LogicalAssign, GetInternalFlag(flag), value));
+}
+
+void ShaderIR::SetLocalMemory(BasicBlock& bb, Node address, Node value) {
+    bb.push_back(Operation(OperationCode::Assign, GetLocalMemory(address), value));
+}
+
+/*static*/ OperationCode ShaderIR::SignedToUnsignedCode(OperationCode operation_code,
+                                                        bool is_signed) {
+    if (is_signed) {
```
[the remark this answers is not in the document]
ReinUsesLisp: It's there to avoid having an `if (is_signed)` on the caller.

src/video_core/shader/decode/memory.cpp
```diff
@@ -102,6 +103,44 @@ u32 ShaderIR::DecodeMemory(BasicBlock& bb, u32 pc) {
...
     }
+    case OpCode::Id::TEX: {
+        Tegra::Shader::TextureType texture_type{instr.tex.texture_type};
+        const bool is_array = instr.tex.array != 0;
+        const bool depth_compare = instr.tex.UsesMiscMode(TextureMiscMode::DC);
+        const auto process_mode = instr.tex.GetTextureProcessMode();
+        UNIMPLEMENTED_IF_MSG(instr.tex.UsesMiscMode(TextureMiscMode::AOFFI),
+                             "AOFFI is not implemented");
+
+        if (instr.tex.UsesMiscMode(TextureMiscMode::NODEP)) {
```
heapo: First, awesome approach, obviously lots of potential here.

Honest question, do you think it makes sense to bubble NODEP down into the IR (and then ignore it in the GLSL backend)? You downgraded the UNIMPLEMENTED_IF to a LOG_WARNING so curious what you thought.

(Either way probably fine for a separate PR to tackle all the NODEP stuff).

src/video_core/shader/decode/memory.cpp
```diff
@@ -0,0 +1,618 @@
...
+                    continue;
+                }
+                meta.components_map[dest_elem] = static_cast<u32>(elem);
+                dest[dest_elem] = GetRegister(instr.gpr0.Value() + dest_elem);
+
+                ++dest_elem;
+            }
+            std::generate(dest.begin() + dest_elem, dest.end(), [&]() { return GetRegister(RZ); });
+
+            bb.push_back(Operation(OperationCode::AssignComposite, std::move(meta), texture,
```
heapo: Am I interpreting correctly that this zeroes out the disabled components? That's a change in behavior so wanted to double check if intentional (maybe a good fix..).

src/video_core/shader/decode/memory.cpp
```diff
@@ -0,0 +1,618 @@
...
+          (texture_type == TextureType::TextureCube && !is_array && depth_compare));
+
+    const OperationCode read_method =
+        lod_needed && gl_lod_supported ? OperationCode::F4TextureLod : OperationCode::F4Texture;
+
+    const MetaTexture meta{sampler, static_cast<u32>(coords.size())};
+
+    std::vector<Node> params = std::move(coords);
+
+    if (process_mode != TextureProcessMode::None) {
```
heapo: Quick note to update to the changes made in https://github.com/yuzu-emu/yuzu/pull/1909 re: LOD selection

src/video_core/shader/decode/memory.cpp
```diff
@@ -0,0 +1,618 @@
...
+    if (depth_compare && !is_array && texture_type == TextureType::Texture1D) {
+        coords.push_back(Immediate(0.0f));
+    }
+    if (depth_compare) {
+        // Depth is always stored in the register signaled by gpr20
+        // or in the next register if lod or bias are used
+        const u64 depth_register = instr.gpr20.Value() + (lod_bias_enabled ? 1 : 0);
+        coords.push_back(GetRegister(depth_register));
+    }
+    if (is_array) {
```
heapo: (Also a change from https://github.com/yuzu-emu/yuzu/pull/1909/files), array index belongs before depth comparison value (also GetTexsCode).

src/video_core/shader/decode/memory.cpp
```diff
@@ -0,0 +1,618 @@
...
+        coords.push_back(
+            Operation(OperationCode::ICastFloat, NO_PRECISE, GetRegister(array_register)));
+    }
+    // Fill ignored coordinates
+    while (coords.size() < total_coord_count) {
```
heapo: Nice refactor with ValidateAndGetCoordinateElement. For the special 1D.DC case, it looks like this will add the '0' noop coordinate to the end of params? It should go as the y value, before the depth compare value.

src/video_core/shader/decode/memory.cpp
```diff
@@ -0,0 +1,618 @@
...
+        if (instr.texs.fp32_flag) {
+            WriteTexsInstructionFloat(bb, instr, texture);
+        } else {
+            UNIMPLEMENTED();
+            // WriteTexsInstructionHalfFloat(bb, instr, texture);
+        }
+        break;
+    }
+    case OpCode::Id::TLD4: {
+        ASSERT(instr.tld4.texture_type == Tegra::Shader::TextureType::Texture2D);
```
heapo: Just left-over debugging code?

src/video_core/shader/decode/memory.cpp
```diff
@@ -0,0 +1,618 @@
...
+            break;
+        }
+        default:
+            UNIMPLEMENTED_MSG("Unhandled coordinates number {}", static_cast<u32>(num_coordinates));
+            params.push_back(GetRegister(instr.gpr8));
+            params.push_back(GetRegister(instr.gpr8.Value() + 1));
+            num_coordinates = 2;
+            texture_type = Tegra::Shader::TextureType::Texture2D;
+        }
+        params.push_back(Immediate(static_cast<u32>(instr.tld4.component)));
```
heapo: In existing code this is only done for non-DC case, just checking intentional?

src/video_core/shader/decode/memory.cpp
```diff
@@ -0,0 +1,618 @@
...
+                if (!instr.tex.IsComponentEnabled(elem)) {
+                    // Skip disabled components
+                    continue;
+                }
+                meta.components_map[dest_elem] = static_cast<u32>(elem);
+                dest[dest_elem] = GetRegister(instr.gpr0.Value() + dest_elem);
+
+                ++dest_elem;
+            }
+            std::generate(dest.begin() + dest_elem, dest.end(), [&]() { return GetRegister(RZ); });
```
heapo: As as earlier comment just checking you believe it's correct to zero out the disabled registers

src/video_core/shader/glsl_decompiler.cpp
```diff
@@ -0,0 +1,1398 @@
...
+    std::string F4TextureLod(Operation operation) {
+        std::string expr = GenerateTexture(operation, "textureLod");
+        if (std::get<MetaTexture>(operation.GetMeta()).sampler.IsShadow()) {
+            expr = "vec4(" + expr + ')';
+        }
+        return expr;
+    }
+
+    std::string F4TextureGather(Operation operation) {
+        return GenerateTexture(operation, "textureGather", "int");
```
heapo: For DC gathers, the 'extra' parameter type should be float.

src/video_core/shader/decode/memory.cpp
ReinUsesLisp: It was like that on the original code.

heapo: Ah I see, those lines were removed in https://github.com/yuzu-emu/yuzu/pull/1862, it looks the rest of this function (probably) needs to be rebased on that

src/video_core/shader/decode/memory.cpp
ReinUsesLisp: Just it pointing as unresolved. I have to take a look.

ReinUsesLisp:
All instructions that are implemented in master do so here. I'll be addressing feedback and trying to fix issues that I find (like OS X's build failing).

src/video_core/shader/shader_ir.h
ReinUsesLisp: It was mostly because it's usage was as an `u32` from where the value was being queried. Changed it to `std::size_t` for consistency.

Hexagon12:
Please rebase the PR.

PR #1929, opened 2018-12-21
hid: Fix SetNpadJoyHoldType and improve logging.

bunnei:
With `SetNpadJoyHoldType`, we were popping the applet resource user ID as the hold type. This caused `GetNpadJoyHoldType` to return an invalid value, which broke several games (e.g. Undertale). This worked prior to #1914, as we were always returning an applet resource user iID of 0.

I also took the time here to clean up and improve HID logging, and fixed a few more minor things.


bunnei:
Untagging Canary because this breaks SMO input - however this should be right, so we need to figure out what's up here. @ogniK5377 @DarkLordZach FYI

bunnei:
We can retag this as Canary, as #1945 should fix the bug this exposed..

PR #1931, opened 2018-12-22
mii: Implement MiiManager backend and several mii service commands

DarkLordZach:
**DO NOT MERGE - Awaiting SSBU to become bootable for further testing**

Implements a MiiManager class to serialize/deserialize the Mii database to saves, and perform operations upon it. Provides conversions between MiiInfo (unpacked format -- used by games usually) and MiiStoreData (packed bitfield format -- used for storage).

The format conversion was 'HW tested' by creating 50 miis with one physical difference from default and then comparing the serialized versions. This is also how the default mii was obtained.

This also extracts the Service::Account::UUID class to Common::UUID as other services make use of it to.

### DIFF
--- a/src/core/hle/service/mii/mii.cpp
+++ b/src/core/hle/service/mii/mii.cpp
@@ -4,6 +4,8 @@
 
 #include <memory>
 
+#include <fmt/ostream.h>
+
 #include "common/logging/log.h"
 #include "common/string_util.h"
 #include "core/hle/ipc_helpers.h"
@@ -14,6 +16,10 @@
 #include "core/hle/service/sm/sm.h"
 
 namespace Service::Mii {
+
+constexpr ResultCode ERROR_INVALID_ARGUMENT{ErrorModule::Mii, 1};
+constexpr ResultCode ERROR_CANNOT_FIND_ENTRY{ErrorModule::Mii, 4};
+constexpr ResultCode ERROR_NOT_IN_TEST_MODE{ErrorModule::Mii, 99};
 
 class IDatabaseService final : public ServiceFramework<IDatabaseService> {
 public:
@@ -35,8 +41,8 @@ public:
             {12, &IDatabaseService::Move, "Move"},
             {13, &IDatabaseService::AddOrReplace, "AddOrReplace"},
             {14, &IDatabaseService::Delete, "Delete"},
-            {15, nullptr, "DestroyFile"},
-            {16, nullptr, "DeleteFile"},
+            {15, &IDatabaseService::DestroyFile, "DestroyFile"},
+            {16, &IDatabaseService::DeleteFile, "DeleteFile"},
             {17, &IDatabaseService::Format, "Format"},
             {18, nullptr, "Import"},
             {19, nullptr, "Export"},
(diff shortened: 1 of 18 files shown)


src/core/hle/service/mii/mii.cpp
```diff
@@ -18,35 +20,260 @@ class IDatabaseService final : public ServiceFramework<IDatabaseService> {
...
-            {11, nullptr, "FindIndex"},
-            {12, nullptr, "Move"},
-            {13, nullptr, "AddOrReplace"},
-            {14, nullptr, "Delete"},
+            {11, &IDatabaseService::FindIndex, "FindIndex"},
+            {12, &IDatabaseService::Move, "Move"},
+            {13, &IDatabaseService::AddOrReplace, "AddOrReplace"},
+            {14, &IDatabaseService::Delete, "Delete"},
             {15, nullptr, "DestroyFile"},
             {16, nullptr, "DeleteFile"},
```
lioncash: Looking into this unimplemented function a little more. What this function does is check the `set:sys` settings database by invoking the service's `GetSettingsItemValue` like:

```cpp
//
// In the settings module:
//

ResultCode GetSettingsItemValue(u64* out_read_amount, void* buffer, size_t buffer_size, const char* name, const char* key) {
    if (!name) {
        // Null name buffer
        return 0x19269;
    }

    if (!key) {
        // Null key buffer
        return 0x19469;
    }

    if (!buffer) {
        // Null settings value buffer
        return 0x19A69;
    }

    // Getter also seems to instantiate the service as well if it doesn't exist already
    SomeInternalSharedPtrType<SystemSettingsSharedProxy> proxy = nullptr;
    const ResultCode result = GetSystemSettingsProxy(&proxy);
    if (result != RESULT_SUCCESS) {
        return result;
    }

    // Copy over to hand rolled 72-char length string buffers and make sure they're null terminated.
    char name_buffer[72] = {};
    constexpr std::size_t copy_length = sizeof(name_buffer) -1;
    std::strncpy(name_buffer, name, copy_length);
    name_buffer[copy_length] = '\0';
    
    char key_buffer[72] = {};
    std::strncpy(key_buffer, key, copy_length);
    key_buffer[copy_length] = '\0';

    struct IPCBuffer {
        void* ptr;
        std::size_t size;
    };

    IPCBuffer output_buffer{buffer, buffer_size}
    u64 read_amount = 0;
    const auto read_result = proxy->GetSettingsItemValue(&read_amount, output_buffer, name_buffer, key_buffer);
    if (read_result != RESULT_SUCCESS) {
        return read_result;
    }

    *out_read_amount = read_amount;
    return RESULT_SUCCESS;
}

u64 GetSettingsItemValueWrapper(void* buffer, size_t buffer_size, const char* name, const char* key) {
    u64 read_amount = 0;

    const ResultCode get_result = GetSettingsItemValue(&read_amount, buffer, buffer_size, name, key);
    if (get_result != RESULT_SUCCESS) {
        std::abort(); // Actually uses their own abort function, but descriptively equivalent
    }

    return read_amount;
}

//
// In the mii code (finally)
//

ResultCode DatabaseServiceImpl::DeleteFile() {
    bool is_test_mode = false;
    settings::GetSettingsItemValue(&is_test_mode, sizeof(is_test_mode), "mii", "is_db_test_mode_enabled");

    if (!is_test_mode) {
        // Debug/dev only feature (this is what a consumer variant of the Switch software should hit, afaik)
        return 0x1987E;
    }

    // Actually try to delete the file (note: You cannot return successful deletion if the file
    // doesn't exist already and this is called, i.e. "well, technically it didn't exist, so yeah, it successfully deleted it",
    // it's only guaranteed to return true if the file existed, and was then deleted).
    return this->wrapped_db->DeleteFile();
}

```

src/core/hle/service/mii/mii.cpp
```diff
@@ -18,35 +20,260 @@ class IDatabaseService final : public ServiceFramework<IDatabaseService> {
...
+        offsets[1] += read_size;
+
+        IPC::ResponseBuilder rb{ctx, 3};
+        rb.Push(RESULT_SUCCESS);
+        rb.Push<u32>(read_size);
+    }
+
+    void BuildRandom(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        const auto random_params{rp.PopRaw<RandomParameters>()};
```
lioncash: Assuming the error code `0x27E` is converted to a result code like:

```cpp
constexpr ResultCode ERR_INVALID_ARGUMENT{ErrorModule::Mii, 1};
```
then the following error cases can be handled:

```suggestion
        const auto [unknown1, unknown2, unknown3] = rp.PopRaw<RandomParameters>();
        
        if (unknown1 > 2) {
            IPC::ResponseBuilder rb{ctx, 2};
            rb.Push(ERR_INVALID_ARGUMENT);
            LOG_ERROR(Service_Mii, "Invalid unknown1 value: {}", unknown1);
            return;
        }
        
        if (unknown2 > 1) {
            IPC::ResponseBuilder rb{ctx, 2};
            rb.Push(ERR_INVALID_ARGUMENT);
            LOG_ERROR(Service_Mii, "Invalid unknown2 value: {}", unknown2);
            return;
        }
        
        if (unknown3 > 2) {
            IPC::ResponseBuilder rb{ctx, 2};
            rb.Push(ERR_INVALID_ARGUMENT);
            LOG_ERROR(Service_Mii, "Invalid unknown3 value: {}", unknown3);
            return;
        }
```

src/core/hle/service/mii/mii_manager.cpp
```diff
@@ -0,0 +1,369 @@
...
+        static_cast<u8>(bf.glasses_scale.Value()),
+        static_cast<u8>(bf.glasses_y.Value()),
+        static_cast<u8>(bf.mole_type.Value()),
+        static_cast<u8>(bf.mole_scale.Value()),
+        static_cast<u8>(bf.mole_x.Value()),
+        static_cast<u8>(bf.mole_y.Value()),
+        0x00,
+    };
+}
+MiiStoreData ConvertInfoToStoreData(const MiiInfo& info) {
```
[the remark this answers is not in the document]
DarkLordZach: This was provided for feature pairity, and given that this is not a complete implementation it might be useful in the future. I can remove it if you'd like though

greggameplayer:
@DarkLordZach you should correct conflict with Service::Account::UUID which is with your PR Common::UUID for profile_select applet

greggameplayer:
here is exactly what you have to fix : https://github.com/greggameplayer/yuzu/commit/8e2404c2b8c88f1f2335aeb2980e1fe750ad9c50

Hexagon12:
Please rebase the PR

PR #1933, opened 2018-12-23
file_sys: Implement parser and interpreter for game memory cheats

DarkLordZach:
Implements the cheat format used by Team Xecuter's SX OS, using their reference manual found [here](https://team-xecuter.com/forums/threads/155007-SX-OS-Cheat-Hacker-Guide).

Unlike Citra, which had issues with memory layout and gateway codes in the beginning, due to the Switch's ASLR nature this is not an issue for yuzu, as all cheat codes use memory addresses relative to the executable or heap.

I was concerned that running this interpreter every frame would cause slowdowns, but that's not the case (for simple codes). For more complex/longer codes, I'd imagine it'd affect performance, and to show this I added a microprofile scope for the CheatEngine. For most codes right now, the average time spent in engine is so small compared to other parts of the emu that it's actually negligible.

As for actual cheats to use, I am planning some sort of distribution framework but for now they are available on [the SX OS cheats forum](https://team-xecuter.com/forums/forums/268-Switch-Game-Cheats-Discussion) and can be placed in `yuzu/load/<title_id>/<mod_name>/cheats/<first 8 bytes of build id>.txt`

This does not provide any UI for managing, adding, or searching for cheats nor does it add support for enabling/disabling cheats at a per-cheat level. This PR was getting too large, and those can come soon. However, plenty of Debug-class logging was added for anyone wanting to test out new cheats with yuzu.

As an example, here is Cave Story+ immediately after starting a new game with "max lvl" and "max health" codes active:
image

### DIFF
--- a/src/core/file_sys/cheat_engine.cpp
+++ b/src/core/file_sys/cheat_engine.cpp
@@ -18,7 +18,7 @@
 
 namespace FileSys {
 
-constexpr u64 CHEAT_ENGINE_TICKS = CoreTiming::BASE_CLOCK_RATE / 60;
+constexpr u64 CHEAT_ENGINE_TICKS = Core::Timing::BASE_CLOCK_RATE / 60;
 constexpr u32 KEYPAD_BITMASK = 0x3FFFFFF;
 
 u64 Cheat::Address() const {
@@ -82,7 +82,7 @@ CheatList::CheatList(ProgramSegment master, ProgramSegment standard)
 
 bool CheatList::EvaluateConditional(const Cheat& cheat) const {
     using ComparisonFunction = bool (*)(u64, u64);
-    constexpr ComparisonFunction comparison_functions[] = {
+    constexpr std::array<ComparisonFunction, 6> comparison_functions{
         [](u64 a, u64 b) { return a > b; },  [](u64 a, u64 b) { return a >= b; },
         [](u64 a, u64 b) { return a < b; },  [](u64 a, u64 b) { return a <= b; },
         [](u64 a, u64 b) { return a == b; }, [](u64 a, u64 b) { return a != b; },
@@ -103,7 +103,7 @@ bool CheatList::EvaluateConditional(const Cheat& cheat) const {
         const auto press_state =
             applet_resource
                 ->GetController<Service::HID::Controller_NPad>(Service::HID::HidController::NPad)
-                .GetPressState();
+                .GetAndResetPressState();
         return ((press_state & cheat.KeypadValue()) & KEYPAD_BITMASK) != 0;
     }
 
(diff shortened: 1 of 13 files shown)


src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -682,6 +682,19 @@ void Controller_NPad::ClearAllControllers() {
                   });
 }
 
+u32 Controller_NPad::GetPressState() const {
+    u32 out{};
+
+    for (const auto& controller : buttons) {
+        for (std::size_t i = 0; i < 26; ++i) {
+            if (controller[i] != nullptr && controller[i]->GetStatus())
+                out |= (1 << i);
```
lioncash: `1U` otherwise this can cause sign conversion warnings.

Hexagon12:
Please rebase this PR

greggameplayer:
@DarkLordZach rebase please

src/core/file_sys/cheat_engine.cpp
```diff
@@ -0,0 +1,493 @@
...
+                Common_Filesystem,
+                "Attempted to evaluate input conditional, but applet resource is not initialized!");
+            return false;
+        }
+
+        const auto press_state =
+            applet_resource
+                ->GetController<Service::HID::Controller_NPad>(Service::HID::HidController::NPad)
+                .GetAndResetPressState();
+        return ((press_state & cheat.KeypadValue()) & KEYPAD_BITMASK) != 0;
```
CrazyMax: `const auto cheat_keypad_value = cheat.KeypadValue();
return ((press_state & cheat_keypad_value) & KEYPAD_BITMASK) == cheat_keypad_value;`

> Multiple button values can be combined by OR'ing them together. For example A+B becomes 0000003, and A+B+X+Y becomes 000000f.

DarkLordZach: Are you implying that the keypad conditionals require all of the buttons to be pressed at once? Personally, I'd interpret the way the spec is written to be any of them (logical OR). If you have evidence that the official tx os software behaves as you claim, I'd be more than happy to change/fix.

CrazyMax: This is adding triggering conditionals for pressing several buttons at the same time.

ex.
> [Refill Health (ZL+ZR)]
> 80000300
> 580f0000 026583c0
> 780f0000 00642d3c
> 640f0000 00000000 00000006
> 20000000

This is code for SMO and in this PR it's triggered when you press ZL or ZR buttons but should be  triggered when ZL and ZR pressed at once.

DarkLordZach: Are you certain that on hardware it would require both buttons?

theboy181:
Can we get an update to this? 

You can only have one cheat folder, and you can't disable it when you uncheck it in the properties.

PR #1934, opened 2018-12-23
Fixed shader linking error due to TLDS

ogniK5377:
`coord` should be `coords`


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -1755,7 +1755,7 @@ class GLSLGenerator {
             instr.tlds.GetTextureProcessMode() == Tegra::Shader::TextureProcessMode::LL;
 
         constexpr std::array<const char*, 4> coord_container{
-            {"", "int coord = (", "ivec2 coord = ivec2(", "ivec3 coord = ivec3("}};
+            {"", "int coords = (", "ivec2 coord = ivec2(", "ivec3 coord = ivec3("}};
```
Subv: The name should be corrected for all cases here

PR #1939, opened 2018-12-24
applets: Implement HLE web browser applet (LibAppletOff)

DarkLordZach:
This provides a high-level replacement for the currently missing LibAppletOff in yuzu, which is used to render game-provided manuals and help documents.

Most notably, this fixes a softlock in SMO after the Sand Kingdom when Cappy tries to open the Action Guide. Now, the action guide will actually open.

Details:
- Qt frontend is hidden behind a CMake flag this is by default disabled (`YUZU_USE_QT_WEB_ENGINE`) -- this is because QtWebEngine is quite a hefty dependency and not included by default with Qt. When build without this flag, if the frontend is requested to display a page a warning message informing the user that yuzu was build without QtWebEngine will appear.
- From my research, game manuals use custom JS that is present in Nintendo's browser (called NetFront NX). This is used for everything from input to exiting. To get around this, I wrote a JavaScript shim with @chris062689's help that interacts with QtWebEngine, allowing the C++ side to fire input events and check if the manual has completed.
- When using the SDL frontend (yuzu-cmd), this will log an error and then continue execution. I chose not to simply launch the user's default web browser as this proved problematic in tests (i.e. missized, weird masking, no input, frozen/unresponsive). Most of these bugs can be pinned to the browser window size and the JS shim aforementioned.
- Even though the manual is responsible for closing (usually via 'B'), I added an additional button in the menubar to close, should a user want to end it quickly:
annotation 2018-12-24 163333
- I had to make the Hid service class public and add an accessor for current pad states to Controller_NPad so the frontend could read input via the service manager.
- Since it takes some time to load and set up the browser, a progress dialog will show:
webloading

Finally screenshots:
wa1
wa2
pppt
unknown

Tested against Spla2n, SMO, and PPT and a variety of input methods. Please lmk if it doesn't work for you.


src/core/hle/service/am/applets/web_browser.cpp
```diff
@@ -0,0 +1,184 @@
...
+    filename.clear();
+    status = RESULT_SUCCESS;
+
+    Applet::Initialize();
+
+    const auto web_arg_storage = broker.PopNormalDataToApplet();
+    ASSERT(web_arg_storage != nullptr);
+    const auto& web_arg = web_arg_storage->GetData();
+
+    LOG_CRITICAL(Service_AM, "{}", Common::HexVectorToString(web_arg));
```
lioncash: Critical is not a "Scream louder into the console" logger, it's supposed to be for actual critical errors that threaten the stability of the program.

src/core/hle/service/am/applets/web_browser.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+WebBrowser::~WebBrowser() = default;
+
+void WebBrowser::Initialize() {
+    complete = false;
+    temporary_dir.clear();
+    filename.clear();
+    status = RESULT_SUCCESS;
+
+    Applet::Initialize();
```
lioncash: This should be the first thing executed in the function.

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -682,6 +684,12 @@ void Controller_NPad::ClearAllControllers() {
                   });
 }
 
+u32 Controller_NPad::GetPressState() {
+    const auto res = press_state;
```
lioncash:
```cpp
return std::exchange(press_state, 0);
```

src/core/hle/service/hid/controllers/npad.h
```diff
@@ -124,6 +124,10 @@ class Controller_NPad final : public ControllerBase {
     void ConnectAllDisconnectedControllers();
     void ClearAllControllers();
 
+    // Logical OR for all buttons presses on all controllers
+    // Specifically for cheat engine and other features.
+    u32 GetPressState();
```
lioncash: This should probably have a different name than `Get`, since it doesn't only act as a getter.

src/yuzu/applets/web_browser.h
```diff
@@ -0,0 +1,45 @@
...
+#include <QWebEngineView>
+#include "core/frontend/applets/web_browser.h"
+
+class GMainWindow;
+
+QString GetNXShimInjectionScript();
+
+class NXInputWebEngineView : public QWebEngineView {
+public:
+    NXInputWebEngineView(QWidget* parent = nullptr);
```
lioncash:
```suggestion
    explicit NXInputWebEngineView(QWidget* parent = nullptr);
```

src/core/core.cpp
```diff
@@ -233,6 +237,7 @@ struct System::Impl {
     /// Frontend applets
     std::unique_ptr<Core::Frontend::ProfileSelectApplet> profile_selector;
     std::unique_ptr<Core::Frontend::SoftwareKeyboardApplet> software_keyboard;
+    std::unique_ptr<Core::Frontend::WebBrowserApplet> web_browser;
```
lioncash: This needs to be deallocated in the Shutdown function, otherwise this uses up memory, even after the core has shut down.

CMakeModules/DownloadExternals.cmake
```diff
@@ -8,7 +8,7 @@ set(prefix "${CMAKE_BINARY_DIR}/externals/${lib_name}")
 if (NOT EXISTS "${prefix}")
     message(STATUS "Downloading binaries for ${lib_name}...")
     file(DOWNLOAD
-        https://github.com/yuzu-emu/ext-windows-bin/raw/master/${remote_path}${lib_name}.7z
+        https://github.com/DarkLordZach/ext-windows-bin/raw/master/${remote_path}${lib_name}.7z
```
FearlessTobi: I don't know if that's a good idea... Also is there a reason for the upgrade? If not it should be its own PR.

DarkLordZach: This was to test appveyor against the new binaries i had to create. It has since been force-pushed back to original (now that yuzu-emu/ext-windows-bin#11 has been merged)

bunnei:
Tested on latest canary - lgtm

vuabongro:
Can you make a video guide fix ?

PR #1940, opened 2018-12-24
applets: Implement HLE controller selector applet

DarkLordZach:
This implements the controller selector applet (0x0C).

- For SDL, this will simply pick the best controller layout according to the parameters and log a warning.
- For Qt, this will open a version of the input tab of config with an error display. You must resolve all of these error before continuing.

Example Qt frontend:
unknown 2

The biggest user of this is Snipperclips+, but Undertale has also been shown to use it.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -25,13 +25,13 @@
 #include "core/hle/kernel/process.h"
 #include "core/hle/kernel/scheduler.h"
 #include "core/hle/kernel/thread.h"
-#include "core/hle/service/am/applets/software_keyboard.h"
 #include "core/hle/service/service.h"
 #include "core/hle/service/sm/sm.h"
 #include "core/loader/loader.h"
 #include "core/perf_stats.h"
 #include "core/settings.h"
 #include "core/telemetry_session.h"
+#include "frontend/applets/controller.h"
 #include "frontend/applets/profile_select.h"
 #include "frontend/applets/software_keyboard.h"
 #include "frontend/applets/web_browser.h"
@@ -107,6 +107,8 @@ struct System::Impl {
             virtual_filesystem = std::make_shared<FileSys::RealVfsFilesystem>();
 
         /// Create default implementations of applets if one is not provided.
+        if (controller_applet == nullptr)
+            controller_applet = std::make_unique<Core::Frontend::DefaultControllerApplet>();
         if (profile_selector == nullptr)
             profile_selector = std::make_unique<Core::Frontend::DefaultProfileSelectApplet>();
         if (software_keyboard == nullptr)
@@ -248,6 +250,7 @@ struct System::Impl {
     bool is_powered_on = false;
 
     /// Frontend applets
+    std::unique_ptr<Core::Frontend::ControllerApplet> controller_applet;
     std::unique_ptr<Core::Frontend::ProfileSelectApplet> profile_selector;
     std::unique_ptr<Core::Frontend::SoftwareKeyboardApplet> software_keyboard;
     std::unique_ptr<Core::Frontend::WebBrowserApplet> web_browser;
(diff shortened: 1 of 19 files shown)


src/core/core.cpp
```diff
@@ -231,6 +233,7 @@ struct System::Impl {
     bool is_powered_on = false;
 
     /// Frontend applets
+    std::unique_ptr<Core::Frontend::ControllerApplet> controller_applet;
```
lioncash: This needs to be deallocated in the shutdown function, otherwise this is using up memory even after the core is shutdown.

src/core/frontend/applets/controller.cpp
```diff
@@ -0,0 +1,38 @@
...
+void DefaultControllerApplet::ReconfigureControllers(std::function<void(bool)> completed,
+                                                     ControllerParameters parameters) const {
+    LOG_WARNING(Service_HID, "(STUBBED) called, automatically setting controller types to best fit "
+                             "in configuration, may be incorrect!");
+
+    const auto& npad =
+        Core::System::GetInstance()
+            .ServiceManager()
+            .GetService<Service::HID::Hid>("hid")
+            ->GetAppletResource()
```
misson20000: From my comments on #1939, this really doesn't make much sense to me to be in `GetAppletResource`, but that's also out of the scope of this PR.

src/core/hle/service/am/applets/controller.cpp
```diff
@@ -0,0 +1,96 @@
...
+    ASSERT(controller_config.size() >= sizeof(ControllerConfig));
+    std::memcpy(&config, controller_config.data(), sizeof(ControllerConfig));
+}
+
+bool Controller::TransactionComplete() const {
+    return complete;
+}
+
+ResultCode Controller::GetStatus() const {
+    return complete ? RESULT_SUCCESS : ResultCode(0x2c80);
```
misson20000: Where does this result code come from? Seems like this logic should be somewhere common to all applets

src/core/hle/service/am/applets/controller.cpp
```diff
@@ -0,0 +1,96 @@
...
+    out.player_code = 0;
+    out.selected_npad_id = static_cast<u32>(
+        HID::Controller_NPad::MapSettingsTypeToNPad(Settings::values.players[0].type));
+    out.result = !ok;
+
+    complete = true;
+
+    std::vector<u8> out_data(sizeof(ControllerConfigOutput));
+    std::memcpy(out_data.data(), &out, sizeof(ControllerConfigOutput));
+    broker.PushNormalDataFromApplet(IStorage{out_data});
```
misson20000: This type of thing seems to happen a lot. Perhaps add a `template<typename T>` constructor to `IStorage` to construct from a T instance?

src/yuzu/main.h
```diff
@@ -99,12 +100,19 @@ class GMainWindow : public QMainWindow {
     // Signal that tells widgets to update icons to use the current theme
     void UpdateThemedIcons();
 
+    void ControllerAppletReconfigureFinished(bool ok);
+
     void ProfileSelectorFinishedSelection(std::optional<Service::Account::UUID> uuid);
+
```
misson20000: This still seems strange to be to be in main.h

src/core/hle/service/am/applets/controller.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+void Controller::Initialize() {
+    Applet::Initialize();
+
+    const auto private_config_storage = broker.PopNormalDataToApplet();
+    ASSERT(private_config_storage != nullptr);
+    const auto& private_controller_config = private_config_storage->GetData();
+
+    ASSERT(private_controller_config.size() >= sizeof(ControllerConfigPrivate));
+    std::memcpy(&private_config, private_controller_config.data(), sizeof(ControllerConfigPrivate));
```
misson20000: This can probably all be turned into a template function.

Hexagon12:
Could you please rebase the PR and address the comments?

greggameplayer:
@DarkLordZach rebase please

MysticExile:
@DarkLordZach needs a rebase

ogniK5377:
Requesting rebase and addressing the issues @DarkLordZach

bunnei:
@DarkLordZach what is the status of this one?

DarkLordZach:
Superseded by #4597

PR #1941, opened 2018-12-25
qt: Use ProfileSelectionDialog when selecting user for save data

DarkLordZach:
This allows us to present a much nicer UI to the user over a simple combo box and is made easy with #1781 

Old UI:
image

New UI:
image

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -827,31 +827,25 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
         const std::string nand_dir = FileUtil::GetUserPath(FileUtil::UserPath::NANDDir);
         ASSERT(program_id != 0);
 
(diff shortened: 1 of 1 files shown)


src/yuzu/main.cpp
```diff
@@ -827,31 +827,23 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
-
-        bool ok = false;
-        const auto index_string =
-            QInputDialog::getItem(this, tr("Select User"),
-                                  tr("Please select the user's save data you would like to open."),
-                                  list, Settings::values.current_user, false, &ok);
-        if (!ok)
-            return;
+        u32 index = 0;
+        // Ensure the QtProfileSelectionDialog is destroyed before making a ProfileManager
```
lioncash: This can just be turned into a lambda function that returns the index, then you don't need to worry about the lifetime explicitly.

PR #1942, opened 2018-12-25
qt: Add setting to prompt for user on game boot

DarkLordZach:
Using #1781, this implementation is trivial. This mimics the real switch behavior of asking which user on every game boot, but it is default disabled as that might get inconvenient.

On boot:
image

Setting (in General sub-group):
image

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -687,9 +687,25 @@ bool GMainWindow::LoadROM(const QString& filename) {
     return true;
 }
 
+void GMainWindow::SelectAndSetCurrentUser() {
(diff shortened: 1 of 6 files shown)


src/yuzu/main.cpp
```diff
@@ -691,6 +691,18 @@ void GMainWindow::BootGame(const QString& filename) {
     LOG_INFO(Frontend, "yuzu starting...");
     StoreRecentFile(filename); // Put the filename on top of the list
 
+    if (UISettings::values.select_user_on_boot) {
+        QtProfileSelectionDialog dialog(this);
```
lioncash: The inner portion can go into its own function.

PR #1955, opened 2018-12-28
gpu: Remove PixelFormat G8R8U and G8R8S, as they do not seem to exist.

bunnei:
- Fixes UI rendering issues in The Legend of Zelda: Breath of the Wild.

Supersedes #1809.


src/video_core/surface.cpp
```diff
@@ -198,9 +198,9 @@ PixelFormat PixelFormatFromTextureFormat(Tegra::Texture::TextureFormat format,
     case Tegra::Texture::TextureFormat::G8R8:
         switch (component_type) {
         case Tegra::Texture::ComponentType::UNORM:
-            return PixelFormat::G8R8U;
+            return PixelFormat::RG8U;
```
heapo: Consider a small comment that G8R8 apparently == R8G8? Just to indicate this is intentional.

heapo:
Small commenting comment but looks good to me.

src/video_core/surface.cpp
bunnei: addressed

PR #1957, opened 2018-12-28
file_sys: Provide generic interface for accessing game data

DarkLordZach:
This generifies the current `RegisteredCache` and `RegisteredCacheUnion` into one: `ContentProvider`. This also provides a version of content provider to be used by the frontend and populated with the game list. This allows treating the game list as a full-fledged source of data instead of accessing the current game only through the AppLoader.

Some things that this allows:
- Removing custom code for packed updates
- Displaying the real version numbers instead of `PACKED`, `XCI`, or `NSP` as the version for packed updates
- Storing DLC, Updates, System Archives, and the like in a directory with games without the need to install
- Avoiding a common pattern across FS code where the apploader is queried for info and if it failed the RCache is then queried
- Avoiding another common FS pattern where random accessors for things like manual data, control data, title, etc are added to AppLoader (and subsequently used via the following bullet point)
- Allows future expansion - as an interface it is trivial to add new backends for content

### DIFF
--- a/src/core/file_sys/registered_cache.cpp
+++ b/src/core/file_sys/registered_cache.cpp
@@ -23,19 +23,19 @@ namespace FileSys {
 // The size of blocks to use when vfs raw copying into nand.
 constexpr size_t VFS_RC_LARGE_COPY_BLOCK = 0x400000;
 
-std::string RegisteredCacheEntry::DebugInfo() const {
+std::string ContentProviderEntry::DebugInfo() const {
     return fmt::format("title_id={:016X}, content_type={:02X}", title_id, static_cast<u8>(type));
 }
 
-bool operator<(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs) {
+bool operator<(const ContentProviderEntry& lhs, const ContentProviderEntry& rhs) {
     return (lhs.title_id < rhs.title_id) || (lhs.title_id == rhs.title_id && lhs.type < rhs.type);
 }
 
-bool operator==(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs) {
+bool operator==(const ContentProviderEntry& lhs, const ContentProviderEntry& rhs) {
     return std::tie(lhs.title_id, lhs.type) == std::tie(rhs.title_id, rhs.type);
 }
 
-bool operator!=(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs) {
+bool operator!=(const ContentProviderEntry& lhs, const ContentProviderEntry& rhs) {
     return !operator==(lhs, rhs);
 }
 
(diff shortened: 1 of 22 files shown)


src/core/file_sys/registered_cache.h
```diff
@@ -131,46 +159,70 @@ class RegisteredCache {
...
         std::optional<u64> title_id = {}) const;
 
 private:
-    std::vector<RegisteredCache*> caches;
+    std::map<ContentProviderUnionSlot, ContentProvider*> providers;
+};
+
+class ManualContentProvider : public ContentProvider {
+public:
+    ~ManualContentProvider() override;
```
lioncash: constructor should also go into the cpp file, since the class contains a non-trivial type.

src/core/file_sys/registered_cache.h
```diff
@@ -131,46 +159,70 @@ class RegisteredCache {
...
+    FrontendManual, ///< Frontend-defined game list or similar
+};
 
-    std::vector<RegisteredCacheEntry> ListEntries() const;
-    // If a parameter is not std::nullopt, it will be filtered for from all entries.
-    std::vector<RegisteredCacheEntry> ListEntriesFilter(
+// Combines multiple ContentProvider(s) (i.e. SysNAND, UserNAND, SDMC) into one interface.
+class ContentProviderUnion : public ContentProvider {
+public:
+    ~ContentProviderUnion() override;
```
lioncash: constructor should go in the cpp file as well.

src/core/file_sys/registered_cache.h
```diff
@@ -47,12 +48,46 @@ constexpr u64 GetUpdateTitleID(u64 base_title_id) {
...
 
 // std unique requires operator== to identify duplicates.
-bool operator==(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs);
-bool operator!=(const RegisteredCacheEntry& lhs, const RegisteredCacheEntry& rhs);
+bool operator==(const ContentProviderEntry& lhs, const ContentProviderEntry& rhs);
+bool operator!=(const ContentProviderEntry& lhs, const ContentProviderEntry& rhs);
+
+class ContentProvider {
+public:
+    virtual ~ContentProvider();
```
lioncash: constructor should go in the cpp file as well.

src/core/file_sys/registered_cache.cpp
```diff
@@ -521,108 +516,188 @@ bool RegisteredCache::RawInstallYuzuMeta(const CNMT& cnmt) {
...
+        provider.second->Refresh();
+    }
 }
 
-std::optional<u32> RegisteredCacheUnion::GetEntryVersion(u64 title_id) const {
-    for (const auto& c : caches) {
-        const auto res = c->GetEntryVersion(title_id);
-        if (res)
+bool ContentProviderUnion::HasEntry(u64 title_id, ContentRecordType type) const {
+    for (const auto& provider : providers) {
```
lioncash:
```cpp
return std::any_of(providers.begin(), providers.end(), [title_id, type](const auto& entry) {
    if (entry.second == nullptr) {
        return false;
    }

    return entry.second->HasEntry(title_id, type);
});
```

Hexagon12:
Could you please rebase this PR and address the comments?

PR #1959, opened 2018-12-29
settings: Add support for setting the RTC manually

DarkLordZach:
This PR allows the user to set the current time reported to the game manually via settings. On game boot, the system stores a differential between the requested time and the current time (where negative is a time in the past and if this feature isn't enabled it will be 0). This is done so that the clock still increments, just relative to requested time (without this, random crashes will occur in audio-related code). Then, when the game requests the current time via `time:u`, the system will add the differential to the current time in seconds and use that for the service.

Tested against Cave Story, was successfully able to dis/enable the seasonal palette.

UI:
image

### DIFF
--- a/src/core/hle/service/time/time.cpp
+++ b/src/core/hle/service/time/time.cpp
@@ -16,10 +16,9 @@
 
 namespace Service::Time {
 
-static s64 GetSecondsSinceEpoch() {
+static std::chrono::seconds GetSecondsSinceEpoch() {
     return std::chrono::duration_cast<std::chrono::seconds>(
-               std::chrono::system_clock::now().time_since_epoch())
-               .count() +
+               std::chrono::system_clock::now().time_since_epoch()) +
            Settings::values.custom_rtc_differential;
 }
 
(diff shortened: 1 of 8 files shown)


FearlessTobi:
All UI changes require a screenshot.

src/core/hle/service/time/time.cpp
```diff
@@ -12,9 +12,17 @@
 #include "core/hle/kernel/client_session.h"
 #include "core/hle/service/time/interface.h"
 #include "core/hle/service/time/time.h"
+#include "core/settings.h"
 
 namespace Service::Time {
 
+static s64 GetSecondsSinceEpoch() {
```
lioncash: This should be returning the `std::chrono::seconds` type. Call sites should explictly get the raw value if they need it.

src/core/settings.h
```diff
@@ -350,9 +351,10 @@ struct Values {
     bool use_docked_mode;
     bool enable_nfc;
     std::optional<u32> rng_seed;
-    std::optional<s64> custom_rtc; // Measured in seconds since epoch
-    s64 custom_rtc_differential;   // Set on game boot, reset on stop. Seconds difference between
-                                   // current time and `custom_rtc`
+    std::optional<std::chrono::seconds> custom_rtc; // Measured in seconds since epoch
+    std::chrono::seconds
+        custom_rtc_differential; // Set on game boot, reset on stop. Seconds difference between
```
lioncash: The comments can be placed above the member variable, which avoids wonky formatting like this.

gdelazzari:
This PR is preventing GCC on Linux to build, #1993 #1995
Sorry for the duplicate issue (which I closed)

Hexagon12:
Please resolve the conflict and the Linux build issue.

bunnei:
@DarkLordZach can you fix the conflicts? Thanks!

PR #1323, opened 2018-09-15
[HACK] Preliminary implementation of LDG

DarkLordZach:
Massive thanks to @gdkchan for the impl idea and tons of help in the beginning, @Subv for bug fixes and help at the end, and @bunnei for reminding me how to read from memory.

Works by approximating the value of the final address using the last IADD_C operation and then reading 16kb following that address. Best way to describe it is a 'hackeuristic'

My current understanding of font vertex shaders is that:
`IADD_C ...`
`IADD_C ...`
`<...>`
`LDG [R0], R0`

Where the two IADDs are to calculate upper and lower of the final offset. This works by taking the constbuffer operand of the lower one (high is always 0 is testing, ignored for now) and reading 16kb (max ubo size) from that into a UBO and presenting that to the shader.

Works in:
- SMO
- BotW
- ARMS

Doesn't work in:
- Splatoon 2
- Mario Tennis Aces
- Some parts in game in SMO (background text in map)

Current issues that need to be resolved:
- [ ] Games that render rectangles instead of text (list above)
- [x] Some sort of caching. Currently has rampant RAM usage and tons of `OpenGL using system RAM` errors.
- [x] Profile for performance impact

Overall, when it does work its pretty cool:
botw2
arms
smo2

Not working results in rectangles for characters:
splatoonbooot

PR #1960, opened 2018-12-29
video_core: Implement LDG through heuristics based on IR

ReinUsesLisp:
This PR uses #1927 as a base implementing #1603 that's itself a continuation of #1323, based on gdkchan's approach for LDG.

About the heuristic itself: it takes the LDG low address register and does a desperate recursive search for any constant buffer that operates on it (returning the first one found), including register proxies.

As difference compared to #1603 and #1323, this one:
* Returns global memory entries through shader entries instead of a global state in Maxwell3D;
* uses SSBOs instead of UBOs; and
* it uses a fixed size for SSBOs.

It has the same graphical issues as #1603, but from what I've seen the heuristic is doing fine (it points to the right constant buffer as base); which means that there's an error elsewhere (some LDG addresses are calculated using XMAD). From some debugging, text textures are fine: the problem is in the shader itself.

The giant diff is because it includes #1927, the last two commits are from this PR.

Edit: It's still broken on AMD's proprietary stack.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -34,6 +34,8 @@ using Operation = const OperationNode&;
 enum : u32 { POSITION_VARYING_LOCATION = 0, GENERIC_VARYING_START_LOCATION = 1 };
 constexpr u32 MAX_CONSTBUFFER_ELEMENTS =
     static_cast<u32>(RasterizerOpenGL::MaxConstbufferSize) / (4 * sizeof(float));
+constexpr u32 MAX_GLOBALMEMORY_ELEMENTS =
+    static_cast<u32>(RasterizerOpenGL::MaxGlobalMemorySize) / sizeof(float);
 
 enum class Type { Bool, Bool2, Float, Int, Uint, HalfFloat };
 
@@ -143,6 +145,7 @@ public:
         DeclareInputAttributes();
         DeclareOutputAttributes();
         DeclareConstantBuffers();
+        DeclareGlobalMemory();
         DeclareSamplers();
 
         code.AddLine("void execute_" + suffix + "() {");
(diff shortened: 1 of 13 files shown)


src/video_core/shader/track.cpp
```diff
@@ -0,0 +1,72 @@
...
+        // Reduce the cursor in one to avoid infinite loops when the instruction sets the same
+        // register that it uses as operand
+        const auto [source, new_cursor] = TrackRegister(gpr, code, cursor - 1);
+        if (!source) {
+            return nullptr;
+        }
+        return TrackCbuf(source, code, new_cursor);
+    }
+    if (const auto operation = std::get_if<OperationNode>(tracked)) {
+        for (std::size_t i = 0; i < operation->GetOperandsCount(); ++i) {
```
lioncash: This seems like it can be a std::find_if operation.

src/video_core/shader/track.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <optional>
+#include <tuple>
+#include <variant>
+
+#include "video_core/shader/shader_ir.h"
+
+namespace VideoCommon::Shader {
+
+static std::pair<Node, s64> FindOperation(const BasicBlock& code, s64 cursor,
```
lioncash: Putting this in an anonymous namespace is a better approach than using `static` (makes it impossible to accidentally violate the One-Definition Rule in another translation unit).

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -197,6 +200,11 @@ class GLSLDecompiler final {
             SamplerEntry desc(sampler, stage, GetSampler(sampler));
             entries.samplers.push_back(desc);
         }
+        for (const auto& gmem : ir.GetGlobalMemoryBases()) {
+            GlobalMemoryEntry desc(gmem.cbuf_index, gmem.cbuf_offset, stage,
+                                   GetGlobalMemoryBlock(gmem));
+            entries.global_memory_entries.push_back(desc);
```
lioncash: This can be reduced to:

```cpp
entries.global_memory_entries.emplace_back(gmem.cbuf_index, gmem.cbuf_offset, stage,
                                           GetGlobalMemoryBlock(gmem));
```

which constructs it in the container instead of constructing it out of the container and then copying it in.

Narugakuruga:
When changed language, characters of Traditional Chinese and Simplified Chinese and Japanese are mostly wrong. But before changing language, those characters work 100% correctly.

ReinUsesLisp:
Characters being wrong may point to another issue. It's curious that in SMO (unlike ARMS) characters seem to be always "in-bounds".
If you can, could you highlight a character that's wrong (and what it should have in replacement)? I can't test on Chinese variants (since I wouldn't know if a character is wrong), but I can do something about Japanese.

FearlessTobi:
Also make sure, even if very unlikely, the issue is not caused by our open-source shared font implementation.

Narugakuruga:
It looks like only Chinese characters (Kanji) are wrong, but Japanese Kana looks fine. (But Chinese characters will turn into any characters, including Chinese character, Japanese Kana, and punctuation)

Though I don't know Japanese, I can identify some Kanas word (without Chinese characters), like オプション(Option), データ(Data), モード(Mode), スタイル(Style). 

But both in Japanese and Chinese, Chinese characters are wrong.



And, all of the wrong texts first appeared in [https://github.com/yuzu-emu/yuzu/pull/1603](url).
And they remain in canary 1801. 
All the correct texts first appeared since [https://github.com/yuzu-emu/yuzu/pull/1961](url).

Narugakuruga:
Okay.

src/video_core/shader/track.cpp
```diff
@@ -0,0 +1,76 @@
...
+        if (const auto gpr_target = std::get_if<GprNode>(target)) {
+            if (gpr_target->GetIndex() == tracked->GetIndex()) {
+                return {(*operation)[1], new_cursor};
+            }
+        }
+    }
+    return {};
+}
+
+} // namespace VideoCommon::Shader
```
bunnei: nit: Missing new line

PR #1961, opened 2018-12-29
gl_rasterizer_cache: Texture view if shader samples array but OGL is not

ReinUsesLisp:
This is WIP because it doesn't implement 1D arrays or Cube arrays.

Fixes fonts in PLG and probably other games.
imagen

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.h
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.h
@@ -293,8 +293,29 @@ public:
         return texture;
     }
 
+    const OGLTexture& TextureLayer() {
+        if (params.is_layered) {
+            return Texture();
+        }
+        EnsureTextureView();
+        return texture_view;
+    }
+
     GLenum Target() const {
         return gl_target;
+    }
+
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_opengl/gl_rasterizer_cache.h
```diff
@@ -289,14 +289,29 @@ class CachedSurface final : public RasterizerCacheObject {
         FlushGLBuffer();
     }
 
-    const OGLTexture& Texture() const {
+    const OGLTexture& Texture() {
```
lioncash:
```suggestion
    OGLTexture& Texture() {
        return texture;
    }
    
    const OGLTexture& Texture() const {
        return texture;
    }
```

ReinUsesLisp: Ended up leaving it as `const`, the non-`const` method is not needed. Originally I changed it for consistency with `TextureLayer`.

ReinUsesLisp:
This is no longer WIP. It should be handling 1D, 2D and Cube textures. Also updated the commit description to be more descriptive.

PR #1974, opened 2019-01-02
yuzu: Port translation support from Citra

FearlessTobi:
This PR ports https://github.com/citra-emu/citra/pull/3297, ~~reverts https://github.com/yuzu-emu/yuzu/pull/1367~~ and partially ports https://github.com/citra-emu/citra/pull/4398.

It also restructures the "Game List" tab to be a "UI" tab, like we already have for Citra.
unbenannt

For instructions how to build with translations or how to update them, see https://github.com/citra-emu/citra/pull/3297. 
People that want to help translate can request a role on [Transifex](https://www.transifex.com/yuzu-emulator/yuzu). 

This PR is labeled [WIP] cause I need to verify all changes to the build system work as expected and cause there are for sure still hidden bugs that will only be uncovered when the translation work really begins and languages are added into yuzu.


src/yuzu/configuration/configure_ui.cpp
```diff
@@ -28,29 +30,51 @@ constexpr std::array<const char*, 4> row_text_names{{
 }};
 } // Anonymous namespace
 
-ConfigureGameList::ConfigureGameList(QWidget* parent)
-    : QWidget(parent), ui(new Ui::ConfigureGameList) {
+ConfigureUi::ConfigureUi(QWidget* parent) : QWidget(parent), ui(new Ui::ConfigureUi) {
     ui->setupUi(this);
 
+    ui->language_combobox->addItem(tr("<System>"), QString(""));
+    ui->language_combobox->addItem(tr("English"), QString("en"));
```
lioncash:
```suggestion
    ui->language_combobox->addItem(tr("English"), QStringLiteral("en"));
```

src/yuzu/game_list.cpp
```diff
@@ -88,7 +88,15 @@ bool GameListSearchField::KeyReleaseEater::eventFilter(QObject* obj, QEvent* eve
 }
 
 void GameListSearchField::setFilterResult(int visible, int total) {
-    label_filter_result->setText(tr("%1 of %n result(s)", "", total).arg(visible));
+    QString result_of_text = tr("of");
```
lioncash: No. Do not do this. This was changed #1367 specifically because this form of translating doesn't respect pluralization in every language. This also makes it unnecessarily obtuse for translators to actually translate it properly instead of fragmenting the string. Please revert this. Actually properly fix the underlying issue, as documented on Citra's repo (which is unrelated to the string directly) instead of going for a substantially worse approach.

This is a hard blocker, by the way.

src/yuzu/configuration/configure_ui.cpp
```diff
@@ -28,29 +30,51 @@ constexpr std::array<const char*, 4> row_text_names{{
...
+    QDirIterator it(":/languages", QDirIterator::NoIteratorFlags);
+    while (it.hasNext()) {
+        QString locale = it.next();
+        locale.truncate(locale.lastIndexOf('.'));
+        locale.remove(0, locale.lastIndexOf('/') + 1);
+        QString lang = QLocale::languageToString(QLocale(locale).language());
```
lioncash:
```suggestion
        const QString lang = QLocale::languageToString(QLocale(locale).language());
```

src/yuzu/configuration/configure_ui.cpp
```diff
@@ -28,29 +30,51 @@ constexpr std::array<const char*, 4> row_text_names{{
...
+        ui->language_combobox->addItem(lang, locale);
+    }
+
+    // Unlike other configuration changes, interface language changes need to be reflected on the
+    // interface immediately. This is done by passing a signal to the main window, and then
+    // retranslating when passing back.
+    connect(ui->language_combobox,
+            static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this,
```
lioncash:
```suggestion
            QOverload<int>::of(&QComboBox::currentIndexChanged), this,
```

FearlessTobi:
Rebased and dropped the commit reverting #1367. This will likely break Transifex push. I'll search for a good solution for the problem.

FearlessTobi:
Rebased. This still has the unresolved TX push problem.

FearlessTobi:
Rebased yet again. The new plan is not to use TX auto push for now. 
So this PR should be good to merge if there aren't any new reviews.

src/yuzu/main.h
```diff
@@ -232,6 +235,8 @@ private slots:
     // Whether emulation is currently running in yuzu.
     bool emulation_running = false;
     std::unique_ptr<EmuThread> emu_thread;
+    // The title of the game currently running
+    QString game_title;
```
lioncash: There's no need to make this a class member. You can just leave UpdateWindowTitle(), the way it was and move the UpdateWindowTitle() call from `BootGame` into `LoadROM`

FearlessTobi: When changing the language while a game is running, the game title will disappear with this approach though.

lioncash: Why would we want to allow changing the language when emulation is running? That'll make it more difficult to properly support RTL (a.k.a bidirectional) layouts in the future, particularly with applet dialogs that can appear at any time during the course of emulation.

bunnei:
@FearlessTobi what is the status of this one?

bunnei:
@FearlessTobi ping again - is this something we could try to get merged sooner than later? Thanks.

FearlessTobi:
I won't be available at all for the next 3 weeks, I still think though I can fix this eventually. I also believe @lioncash was working on this, but I don't know his current status.

PR #1979, opened 2019-01-03
Fixed botw deadlock due to incorrect event signal type

ogniK5377:
Upon investigating the issue with #1878, I found that games are the ones who handle the vsync event resetting and not us.


heapo:
+1'ing that this lets us remove the (terrible) hack, botw boots successfully for me

bunnei:
Awesome, thanks! (and thanks heapo for testing)

PR #1983, opened 2019-01-04
Changed UNIMPLEMENTED_MSG -> LOG_CRITICAL for missing services

ogniK5377:
Easier debugging for canary logs

### DIFF
--- a/src/core/hle/service/service.cpp
+++ b/src/core/hle/service/service.cpp
@@ -148,7 +148,7 @@ void ServiceFrameworkBase::ReportUnimplementedFunction(Kernel::HLERequestContext
     }
     buf.push_back('}');
 
-    UNIMPLEMENTED_MSG("Unknown / unimplemented {}", fmt::to_string(buf));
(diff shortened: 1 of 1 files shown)


lioncash:
It would be a better option to update the ignore assertions PR to just change the unimplemented macros to log as well. Unimplemented services should undeniably assert in master, not continue onwards.

ogniK5377:
With a previous PR, i fixed all uninitalized reads, we wouldn’t even have to prevent execution. We could still have assertion messages to show up @lioncash

lioncash:
I don't understand what you're trying to tell me, or why that's relevant. If a service is unimplemented that should be a flat stop no matter what in master. If we're going to impact master, just so canary can hackily continue, then no, I'm totally against this.

FernandoS27:
@lioncash I made a PR that did but you were deeply against it.

lioncash:
As far as I can remember, I was against it being merged into master. I was fine with it being canary tagged though.

PR #1990, opened 2019-01-05
gl_rasterizer_cache: Use GL_STREAM_COPY for PBOs

ReinUsesLisp:
Since the data is doing the path CPU -> GPU -> GPU, `_COPY` is the most approximate hint. Using `GL_STREAM_DRAW` generated a performance warning on Nvidia's stack. Changing this hint removed the warning.


degasus:
Oh no, this copy shall not do a gpu->cpu->gpu movement. Everything shall be done on the gpu. However, I've seen the same warning and I guessed we hit a slow path in the nvidia driver.

PR #2002, opened 2019-01-06
gl_rasterizer: Use DSA for VAOs and buffers

ReinUsesLisp:
Uses DSA (ARB_direct_state_access) for VAOs and buffers except for PBOs due to their bind behaviour it wouldn't make sense to use DSA there.

The plan is to drop as much as `gl_state` as possible to reduce the amount of `Apply()` calls to two or one per draw call. Using DSA also allows a better usage of ARB_multi_bind for textures.

### DIFF
--- a/src/video_core/renderer_opengl/gl_buffer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_buffer_cache.cpp
@@ -14,7 +14,7 @@
 namespace OpenGL {
 
 OGLBufferCache::OGLBufferCache(RasterizerOpenGL& rasterizer, std::size_t size)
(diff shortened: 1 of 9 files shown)


ReinUsesLisp:
It's not GPU's fault, but their vendors deprecating the driver. On mesa even OpenGL 1.4 drivers supports DSA.

ReinUsesLisp:
This has a regression on compatible Intel GPUs. Untagging for now.

ReinUsesLisp:
Pushed a commit to workaround the Intel bug, there's a description in the commit. Re-tagging it.

PR #2012, opened 2019-01-09
[WIP] Asynchronous GPU command processing

bunnei:
This moves GPU command processing to another thread. This works by letting the CPU core (main) thread push GPU commands via the `DmaPusher::Push` interface, while the GPU thread waits for commands to be available. After a push, the main thread signals that commands are available, and the GPU thread wakes up and continues to process these until the command queue is empty.

There are no "emulated" CPU <-> GPU synchronization mechanisms here, so it's possible that this might cause issues. As such, it's behind a user setting. That being said, it seems quite stable, and I do not notice any bugs/regressions with the games that I have.

We do need some synchronization for things to work, of course. Currently, this implementation:
* Waits for all previously queued GPU commands to be processed before calling SwapBuffers
* Waits for all previously queued GPU commands to be processed before letting the main thread invalidate or flush GPU data
* Uses SwapBuffers as a barrier - e.g., wait until SwapBuffers (end of frame) has been handled by the GPU before allowing the main thread to continue to the next frame

Recommend reviewing commit-by-commit.

This is WIP because it needs lots of testing still. There are probably improvements we can also make to make this a little more parallel - the current synchronization is just my first attempt.

### DIFF
--- a/src/video_core/gpu_thread.cpp
+++ b/src/video_core/gpu_thread.cpp
@@ -1,0 +1,115 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/frontend/scope_acquire_window_context.h"
+#include "video_core/gpu.h"
+#include "video_core/gpu_thread.h"
+#include "video_core/renderer_base.h"
+
+namespace {
+static void RunThread(VideoCore::RendererBase& renderer, Tegra::DmaPusher& dma_pusher,
+                      VideoCore::GPUThreadState& state) {
+
+    Core::Frontend::ScopeAcquireWindowContext acquire_context{renderer.GetRenderWindow()};
+
+    while (state.is_running) {
+        bool is_dma_pending{};
+        bool is_swapbuffers_pending{};
+
+        {
+            // Wait for CPU thread to send GPU commands
+            std::unique_lock<std::mutex> lock{state.signal_mutex};
+            state.signal_condition.wait(lock, [&] {
+                return state.is_dma_pending || state.is_swapbuffers_pending || !state.is_running;
+            });
+
+            if (!state.is_running) {
+                return;
+            }
+
+            is_dma_pending = state.is_dma_pending;
+            is_swapbuffers_pending = state.is_swapbuffers_pending;
+
+            if (is_dma_pending) {
+                dma_pusher.QueuePendingCalls();
+                state.is_dma_pending = false;
+            }
+        }
+
+        if (is_dma_pending) {
+            // Process pending DMA pushbuffer commands
+            std::lock_guard<std::recursive_mutex> lock{state.running_mutex};
+            dma_pusher.DispatchCalls();
+        }
+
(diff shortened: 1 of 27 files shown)


src/video_core/gpu_thread.h
```diff
@@ -0,0 +1,56 @@
...
+namespace Tegra {
+struct FramebufferConfig;
+}
+
+namespace VideoCore {
+
+class RendererBase;
+
+struct GPUThreadState final {
+    bool is_running{true};
```
FernandoS27: std::atomic<bool>

src/video_core/gpu_thread.h
```diff
@@ -0,0 +1,56 @@
...
+    bool is_dma_pending{};
```
FernandoS27: ditto

src/video_core/gpu_thread.h
```diff
@@ -0,0 +1,56 @@
...
+    bool is_swapbuffers_pending{};
```
FernandoS27: ditto

src/yuzu/bootmanager.cpp
```diff
@@ -21,7 +21,7 @@
 EmuThread::EmuThread(GRenderWindow* render_window) : render_window(render_window) {}
 
 void EmuThread::run() {
-    if (!Settings::values.use_multi_core) {
+    if (!Settings::values.use_asynchronous_gpu_emulation) {
```
FearlessTobi: With this changed... why is the multicore config setting even kept around at all? Is it still used somewhere?

src/video_core/gpu_thread.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+#include "core/frontend/scope_acquire_window_context.h"
+#include "video_core/gpu.h"
+#include "video_core/gpu_thread.h"
+#include "video_core/renderer_base.h"
+
+namespace {
+static void RunThread(VideoCore::RendererBase& renderer, Tegra::DmaPusher& dma_pusher,
+                      VideoCore::GPUThreadState& state) {
+
```
degasus: MicroProfileOnThreadCreate("GpuThread");

mspass:
commit `19ce7ab` has a 21%~ performance penalty on sonic mania, also adds stuttering between 2D and 3D elements.

WeirdBob:
Hi, I couldn't find why but toggling this feature on and off results in a difference of visual quality, when enabled, the resolution seems lower. 
Changing the resolution factor seems to have no effect.

mspass:
> 
> 
> @mspass Could you find the exact commit that degraded Sonic Mania's performance? The commit you listed doesn't do anything apart disabling changing async GPU while ingame.

canary-1816 work fine, 1817 degrades performance, which was the only different in this PR, and 1819 degrades even further.

Hexagon12:
Untagged from Canary since a lot of users are reporting performance regressions, especially with the latest commits with the invalidation. Also, this PR introduces broken dynamic resolution scaling for games, like Super Mario Odyssey.

PR #2020, opened 2019-01-14
audio_core: remove unnecessary spaces on comments

otaviopace:
:slightly_smiling_face:


simontime:
but why

otaviopace:
Why not? :thinking: 

I couldn't help with the code yet, because C++ and emulators are still hard for me.

PR #2021, opened 2019-01-14
gl_shader_cache: Disk based shader cache

ReinUsesLisp:
This PR has as base #1987 (which has LDG and the shader IR as base itself) - hence the diff. 
b4823fb is the first commit

While the game is running the disk shader cache works like this:
- When a new shader is found and it's not stored, save it in a `transferable` file.
- When a new shader variant is found, add a new usage in the `transferable` file.

When the game is booting:
- Load transferable shaders and decompile them. Delete the file if its version is old, skip it if it's newer. Delete the file if it has an invalid shader hash (checksum).
- Load precompiled cache file - delete it if it has an invalid version hash.
- If a shader variant is precompiled, load it; otherwise build it.
- If the driver rejects it: delete the precompiled cache and build all the following shaders (driver or hardware change).
- After all shaders have been loaded, save all shaders that are not stored in the precompiled cache (`glGetProgramBinary` and friends).

This is not a complete shader cache. Drivers will recompile the shader when the state they were created (or used) with has changed. When a precompiled shader is dumped the shader program was not used by the driver hence the dump is incomplete. A proper solution would be to save in `transferable` all the OpenGL state a shader is used with and use it like that while the game boots. I'm not aware of what state exactly matters and this is driver dependant.

This is tagged as [WIP] because it doesn't properly implement a proper Qt frontend, the window won't even idle while shaders are building - and this takes a long time when there are lots of shaders (~13 minutes on my machine with ~4000 shaders). But it reduces to a few seconds when there's a precompiled cache.

Shaders are built in a single thread.

The precompiled cache does not store GLSL or OpenGL entries, storing these would reduce significantly the time spent decompiling shaders (a step that's not skipped even with a precompiled cache).

It is not recommended to "build" transferable cache libraries for now because the format may change in in the life-cycle of this PR.

Remainder that sharing shader caches of any type is illegal, shaders are game code and that's copyrighted. Sharing transferable cache of free software (most homebrew) is fine.

Thanks to Exzap for originally helping in the design of the cache and @jroweboy for his CMake-based precompiled cache invalidation.


FernandoS27:
It's hard to review since this is based on 2 long PRs, I'll wait for the base PRs to be merged.

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
```diff
@@ -20,15 +20,107 @@
...
+    transferable.insert({id, {}});
+}
+
+void ShaderDiskCacheOpenGL::SaveUsage(const ShaderDiskCacheUsage& usage) {
+    const auto it = transferable.find(usage.unique_identifier);
+    if (it == transferable.end()) {
+        LOG_CRITICAL(Render_OpenGL, "Saving shader usage without storing raw previously");
+        UNREACHABLE();
+    }
+    auto& usages{it->second};
```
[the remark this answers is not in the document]
ReinUsesLisp: `usages` is used with `insert` after the assert.

CMakeModules/GenerateSCMRev.cmake
```diff
@@ -0,0 +1,101 @@
...
+if ($ENV{CI})
+  if ($ENV{TRAVIS})
+    set(BUILD_REPOSITORY $ENV{TRAVIS_REPO_SLUG})
+    set(BUILD_TAG $ENV{TRAVIS_TAG})
+  elseif($ENV{APPVEYOR})
+    set(BUILD_REPOSITORY $ENV{APPVEYOR_REPO_NAME})
+    set(BUILD_TAG $ENV{APPVEYOR_REPO_TAG_NAME})
+  endif()
+  # regex capture the string nightly or canary into CMAKE_MATCH_1
+  string(REGEX MATCH "citra-emu/citra-?(.*)" OUTVAR ${BUILD_REPOSITORY})
```
slashiee:
```suggestion
  string(REGEX MATCH "yuzu-emu/yuzu-?(.*)" OUTVAR ${BUILD_REPOSITORY})
```

ReinUsesLisp: Thanks for the catch! C-c C-v is evil.

FearlessTobi:
Reminder to fix up the title bar problem.

src/video_core/renderer_opengl/gl_shader_cache.h
```diff
@@ -7,38 +7,47 @@
...
-struct BaseBindings {
-    u32 cbuf{};
-    u32 gmem{};
-    u32 sampler{};
-
-    bool operator<(const BaseBindings& rhs) const {
-        return std::tie(cbuf, gmem, sampler) < std::tie(rhs.cbuf, rhs.gmem, rhs.sampler);
-    }
-};
+using PrecompiledPrograms = std::map<ShaderDiskCacheUsage, CachedProgram>;
```
FernandoS27: Do this two require to be ordered all the time? If not, use unordered_map. It's O(1) compared to map that is O(lg_n)

src/video_core/renderer_opengl/gl_shader_cache.h
```diff
@@ -101,13 +112,29 @@ class CachedShader final : public RasterizerCacheObject {
...
+                       const VideoCore::DiskResourceLoadCallback& callback);
 
     /// Gets the current specified shader stage program
     Shader GetStageProgram(Maxwell::ShaderProgram program);
 
 private:
+    std::map<u64, UnspecializedShader> GenerateUnspecializedShaders(
+        const std::atomic_bool& stop_loading, const VideoCore::DiskResourceLoadCallback& callback,
+        const std::vector<ShaderDiskCacheRaw>& raws,
+        const std::map<u64, ShaderDiskCacheDecompiled>& decompiled);
```
FernandoS27: ditto.

src/video_core/renderer_opengl/gl_shader_cache.h
```diff
@@ -65,33 +74,35 @@ class CachedShader final : public RasterizerCacheObject {
...
+    const PrecompiledPrograms& precompiled_programs;
+
+    std::size_t shader_length{};
     GLShader::ShaderEntries entries;
 
     std::string code;
 
-    std::map<BaseBindings, OGLProgram> programs;
+    std::map<BaseBindings, CachedProgram> programs;
     std::map<BaseBindings, GeometryPrograms> geometry_programs;
```
FernandoS27: ditto

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
```diff
@@ -0,0 +1,652 @@
...
+}
+
+} // namespace
+
+ShaderDiskCacheRaw::ShaderDiskCacheRaw(u64 unique_identifier, Maxwell::ShaderProgram program_type,
+                                       u32 program_code_size, u32 program_code_size_b,
+                                       ProgramCode program_code, ProgramCode program_code_b)
+    : unique_identifier{unique_identifier}, program_type{program_type},
+      program_code_size{program_code_size}, program_code_size_b{program_code_size_b},
+      program_code{std::move(program_code)}, program_code_b{std::move(program_code_b)} {}
```
FernandoS27: are this two moves safe? I don't see why can't they use pointers or references instead.

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
```diff
@@ -0,0 +1,652 @@
...
+    ASSERT(usages.find(usage) == usages.end());
+    usages.insert(usage);
+
+    FileUtil::IOFile file = AppendTransferableFile();
+    if (!file.IsOpen())
+        return;
+
+    if (file.WriteObject(TransferableEntryKind::Usage) != 1 || file.WriteObject(usage) != 1) {
+        LOG_ERROR(Render_OpenGL, "Failed to save usage transferable cache entry - removing");
+        file.Close();
```
FernandoS27: this is unnecessary, the wrapper on IOFile guarantees a RAII design, so letting it die, closes the file.

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
```diff
@@ -0,0 +1,652 @@
...
+    }
+
+    FileUtil::IOFile file = AppendPrecompiledFile();
+    if (!file.IsOpen())
+        return;
+
+    if (!SaveDecompiledFile(file, unique_identifier, code, compressed_code, entries)) {
+        LOG_ERROR(Render_OpenGL,
+                  "Failed to save decompiled entry to the precompiled file - removing");
+        file.Close();
```
FernandoS27: ditto.

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
```diff
@@ -0,0 +1,652 @@
...
+
+    if (file.WriteObject(static_cast<u32>(PrecompiledEntryKind::Dump)) != 1 ||
+        file.WriteObject(usage) != 1 || file.WriteObject(static_cast<u32>(binary_format)) != 1 ||
+        file.WriteObject(static_cast<u32>(binary_length)) != 1 ||
+        file.WriteObject(static_cast<u32>(compressed_binary.size())) != 1 ||
+        file.WriteArray(compressed_binary.data(), compressed_binary.size()) !=
+            compressed_binary.size()) {
+        LOG_ERROR(Render_OpenGL, "Failed to save binary program file in shader={:016x} - removing",
+                  usage.unique_identifier);
+        file.Close();
```
FernandoS27: ditto

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
ReinUsesLisp: It will stay alive while `InvalidateTransferable()` is called. That would attempt to delete an opened file.

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
ReinUsesLisp: They are being passed by copy, the move is safe. Caller's `program_code` is a scoped vector, passing by reference to a `std::shared_ptr` wouldn't work here.

src/common/CMakeLists.txt
```diff
@@ -1,42 +1,56 @@
...
-  endif()
-endif()
-configure_file("${CMAKE_CURRENT_SOURCE_DIR}/scm_rev.cpp.in" "${CMAKE_CURRENT_SOURCE_DIR}/scm_rev.cpp" @ONLY)
+# Add a custom command to generate a new shader_cache_version hash when any of the following files change
+# NOTE: This is an approximation of what files affect shader generation, its possible something else
+# could affect the result, but much more unlikely than the following files. Keeping a list of files
+# like this allows for much better caching since it doesn't force the user to recompile binary shaders every update
+set(VIDEO_CORE "${CMAKE_SOURCE_DIR}/src/video_core")
+add_custom_command(OUTPUT scm_rev.cpp
+    COMMAND cmake -DSRC_DIR="${CMAKE_SOURCE_DIR}" -P "${CMAKE_SOURCE_DIR}/CMakeModules/GenerateSCMRev.cmake"
```
FreddyFunk: Please do not rely on Windows PATH for cmake custom command
 
```suggestion
    COMMAND ${CMAKE_COMMAND} -DSRC_DIR="${CMAKE_SOURCE_DIR}" -P "${CMAKE_SOURCE_DIR}/CMakeModules/GenerateSCMRev.cmake"
```

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
```diff
@@ -0,0 +1,652 @@
...
+
+template <typename T>
+std::vector<u8> CompressData(const T* source, std::size_t source_size) {
+    const auto source_size_int = static_cast<int>(source_size);
+    const int max_compressed_size = LZ4_compressBound(source_size_int);
+    std::vector<u8> compressed(max_compressed_size);
+    const int compressed_size = LZ4_compress_default(reinterpret_cast<const char*>(source),
+                                                     reinterpret_cast<char*>(compressed.data()),
+                                                     source_size_int, max_compressed_size);
+    if (compressed_size < 0) {
```
FreddyFunk: return value `0` is also an error. See doc here https://github.com/lz4/lz4/blob/4db65c1d99280a757823914efaa96d2e87f41844/lib/lz4.h#L134

```suggestion
    if (compressed_size <= 0) {
```

src/video_core/renderer_opengl/gl_shader_disk_cache.cpp
```diff
@@ -0,0 +1,652 @@
...
+ShaderCacheVersionHash GetShaderCacheVersionHash() {
+    ShaderCacheVersionHash hash{};
+    const std::size_t length = std::min(std::strlen(Common::g_shader_cache_version), hash.size());
+    std::memcpy(hash.data(), Common::g_shader_cache_version, length);
+    return hash;
+}
+
+template <typename T>
+std::vector<u8> CompressData(const T* source, std::size_t source_size) {
+    const auto source_size_int = static_cast<int>(source_size);
```
FreddyFunk: Not very likely to happen in this scenario, but if we should use this function for other stuff it might be a useful guard
```suggestion
    if (source_size > LZ4_MAX_INPUT_SIZE) {
        // Source size exceeds LZ4 maximum input size
        return {};
    }
    const auto source_size_int = static_cast<int>(source_size);
```

ReinUsesLisp:
The title bar issue should be fixed.

PR #1698, opened 2018-11-15
[WIP] Initial implementation of MapPhysicalMemory

ogniK5377:
This PR adds PersonalMmHeapUsage, UnmapPhysicalMemory, and MapPhysicalMemory. It seems a lot of games can have this disabled which we originally did but a few games seem to require it. Notable that it seemed to have fixed a few unknown unmapped reads in a few games. When MapPhysicalMemory or UnmapPhysicalMemory fails, we need to redo what we've done. This could be improved but for now, we're iterating a vector and restoring what we did.

PR #2024, opened 2019-01-14
[WIP] Some SSBU fixes

BaronKiko:
DO NOT MERGE
This PR doesn't make smash boot, it does give some progress.
This PR isn't meant to be merged, it's meant to be a collection of changes required for smash however they may not be optimal changes. It is also a child of https://github.com/yuzu-emu/yuzu/pull/1698 which is being re-implemented.
At it's current state it should get into an infinite loop, still unknown why.

### DIFF
--- a/src/core/file_sys/program_metadata.cpp
+++ b/src/core/file_sys/program_metadata.cpp
@@ -63,6 +63,10 @@ u32 ProgramMetadata::GetMainThreadStackSize() const {
     return npdm_header.main_stack_size;
 }
 
+u32 ProgramMetadata::GetSystemResourceSize() const {
+    return npdm_header.system_resource_size;
+}
+
 u64 ProgramMetadata::GetTitleID() const {
     return aci_header.title_id;
 }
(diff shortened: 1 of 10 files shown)


BaronKiko:
I have no idea if the sleep change is hardware accurate as I honestly don't have any idea how you would test that nor do I have my switch setup for dev. I simply noticed it got stuck there waiting for int32 max nanoseconds.
I don't think the loop is caused by the parent pr, I think it's a different loop later on. I left it for over 10 mins yesterday and it never progressed. 
I only included the other pr because it's required to test this pr and is consequently the source of my branch. The other mii pr isn't included for a similar reason. I never really intended it to be merged at this point, Once it's showing more progress I will split it up into all the required parts. Till then, as it says, do not merge.

BaronKiko:
> Just going to point out that MapPhysicalMem is implemented wrong

I'm aware but it's the best I have until it has been re-implemented. It's working for what I need here so I'm not exactly sure how it's broken, not that's it's really relevant to me or this pr.

Hexagon12:
@BaronKiko Will you still continue on getting this PR in a mergeable state? If yes, please rebase it.

BaronKiko:
I got distracted. I believe the time less than 0 change was already implemented on the mii pr(?) and the other change is just a stub so I will just close it.

PR #2034, opened 2019-01-18
QT Frontend: Add a Loading screen with progressbar

jroweboy:
With shader caches on the horizon, one requirement is to provide visible
feedback for the progress. The shader cache reportedly takes several
minutes to load for large caches that were invalidated, and as such we
should provide a loading screen with progress.

Adds a loading screen widget that will be shown until the first frame of
the game is swapped. This was chosen in case shader caches are not being
used, several games still take more than a few seconds to launch and
could benefit from a loading screen.

Sorry for clang formatting issues. I'll look more into why VS is doing this to me soon.

UI Screenshot

Video of it in action: https://i.imgur.com/ybQT1fN.gifv

This PR depends on #2025 
Would also like have #2017 merged first as well.


src/yuzu/loading_screen.h
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Loader {
+class AppLoader;
+}
+
+namespace Ui {
+class LoadingScreen;
+}
+
+class QSpacerItem;
```
lioncash: These should be alphabetically ordered.

src/yuzu/loading_screen.h
```diff
@@ -0,0 +1,58 @@
...
+class QByteArray;
+class QBuffer;
+class QProgressBar;
+class QHBoxLayout;
+
+class LoadingScreen : public QWidget {
+    Q_OBJECT
+
+public:
+    explicit LoadingScreen(QWidget* parent = nullptr);
```
lioncash: The destructor should be defaulted within the cpp file as well, since the types the unique_ptr instances are holding are forward declared in this context (and therefore incomplete types). Technically the destructor is able to be inlined currently, and this can cause compilation errors in the future if it's inlined into a context where the types are still incomplete. Defaulting the destructor in the cpp file ensures the destruction logic always has the complete types visible.

src/yuzu/loading_screen.h
```diff
@@ -0,0 +1,58 @@
...
+
+    /// After the loading screen is hidden, the owner of this class can call this to clean up any
+    /// used resources such as the logo and banner.
+    void Clear();
+
+    // In order to use a custom widget with a stylesheet, you need to override the paintEvent
+    // See https://wiki.qt.io/How_to_Change_the_Background_Color_of_QWidget
+    void paintEvent(QPaintEvent* event) override;
+
+public slots:
```
lioncash: This can just be a regular public function. The requirement to be a `slot` is a pre-Qt 5 thing.

jroweboy:
All comments address. Would like to see this and #2038 merged soon so we can get shader caches merged.

PR #2035, opened 2019-01-18
yuzu/configuration: Remove unnecessary inclusions where applicable

lioncash:
Removes a bit of unnecessary headers and gets rid of some direct dependencies on other headers. Reduces the amount of rebuilding necessary if certain headers change in the future.


lioncash:
Error isn't related to the code itself (it's a timeout on appveyor's end).

PR #2038, opened 2019-01-20
Loading progress bar upgrades

jroweboy:
Part two for #2034, this adds more sparkle to the progress bar to make it more useful. Will rebase this pr after #2034 is merged.

**This image was taken with a loading simulator, and is not representative of actual shader loading times**

obligatory ui image

There are two problems with the original single progressbar

1. The shader disk loading is actually broken into two parts: loading raw shaders and loading binary shaders.
2. Newly added shaders are appended to the caches, so on the next launch, loading progress goes quickly until it reaches new shaders at which point it'll dramatically slow down.

With these two problems in mind, I designed the progress bar to be minimal but informative. The progressbar is hidden until shader loading is started, and there are two distinct phases with different color progressbars (the colors used in the yuzu icon). If the progress slows down significantly, then we've reached a point where new shaders are either getting decompiled (step 1) or getting compiled (step 2) and we display an estimated time.

Reviewers:
The first 4 commits are from #2034 as this depends on it. Please just review the last commit.

### DIFF
--- a/src/video_core/rasterizer_interface.h
+++ b/src/video_core/rasterizer_interface.h
@@ -4,12 +4,21 @@
 
 #pragma once
 
+#include <functional>
 #include "common/common_types.h"
 #include "video_core/engines/fermi_2d.h"
 #include "video_core/gpu.h"
 #include "video_core/memory_manager.h"
 
 namespace VideoCore {
+
+enum class LoadCallbackStage {
+    Prepare,
+    Raw,
+    Binary,
+    Complete,
+};
+using DiskResourceLoadCallback = std::function<void(LoadCallbackStage, std::size_t, std::size_t)>;
 
 class RasterizerInterface {
 public:
--- a/src/video_core/rasterizer_interface.h
+++ b/src/video_core/rasterizer_interface.h
@@ -14,8 +14,8 @@ namespace VideoCore {
 
 enum class LoadCallbackStage {
     Prepare,
-    Raw,
-    Binary,
+    Decompile,
+    Build,
     Complete,
 };
 using DiskResourceLoadCallback = std::function<void(LoadCallbackStage, std::size_t, std::size_t)>;
--- a/src/yuzu/loading_screen.cpp
+++ b/src/yuzu/loading_screen.cpp
@@ -28,6 +28,43 @@
 #include <QMovie>
 #endif
 
+constexpr const char* PROGRESSBAR_STYLE_PREPARE = R"(
+QProgressBar {}
+QProgressBar::chunk {})";
+
+constexpr const char* PROGRESSBAR_STYLE_RAW = R"(
+QProgressBar {
+  background-color: black;
+  border: 2px solid white;
+  border-radius: 4px;
+  padding: 2px;
+}
+QProgressBar::chunk {
+  background-color: #0ab9e6;
+})";
+
+constexpr const char* PROGRESSBAR_STYLE_BINARY = R"(
+QProgressBar {
+  background-color: black;
+  border: 2px solid white;
+  border-radius: 4px;
+  padding: 2px;
+}
+QProgressBar::chunk {
+ background-color: #ff3c28;
+})";
+
+constexpr const char* PROGRESSBAR_STYLE_COMPLETE = R"(
+QProgressBar {
+  background-color: black;
+  border: 2px solid white;
+  border-radius: 4px;
+  padding: 2px;
+}
+QProgressBar::chunk {
+  background-color: #ff3c28;
+})";
+
 LoadingScreen::LoadingScreen(QWidget* parent)
     : QWidget(parent), ui(std::make_unique<Ui::LoadingScreen>()),
       previous_stage(VideoCore::LoadCallbackStage::Complete) {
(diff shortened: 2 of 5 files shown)


src/yuzu/loading_screen.cpp
```diff
@@ -25,11 +28,71 @@
 #include <QMovie>
 #endif
 
+/// Static maps defining the differences in text and styling for each stage
+static std::unordered_map<VideoCore::LoadCallbackStage, const char*> stage_translations = {
```
lioncash: `std::unordered_map`, shouldn't be used in a file-scope context, given it has a non-trivial constructor (and also heap allocates).

src/yuzu/loading_screen.cpp
```diff
@@ -25,11 +28,71 @@
...
+    {VideoCore::LoadCallbackStage::Prepare, QT_TRANSLATE_NOOP("LoadingScreen", "Loading...")},
+    {VideoCore::LoadCallbackStage::Raw,
+     QT_TRANSLATE_NOOP("LoadingScreen", "Preparing Shaders %1 / %2")},
+    {VideoCore::LoadCallbackStage::Binary,
+     QT_TRANSLATE_NOOP("LoadingScreen", "Loading Shaders %1 / %2")},
+    {VideoCore::LoadCallbackStage::Complete, QT_TRANSLATE_NOOP("LoadingScreen", "Launching...")},
+};
+
+static std::unordered_map<VideoCore::LoadCallbackStage, const char*> progressbar_style = {
```
lioncash: Ditto

src/yuzu/loading_screen.h
```diff
@@ -43,7 +48,11 @@ class LoadingScreen : public QWidget {
     // See https://wiki.qt.io/How_to_Change_the_Background_Color_of_QWidget
     void paintEvent(QPaintEvent* event) override;
 
-    void OnLoadProgress(std::size_t value, std::size_t total);
+public slots:
```
lioncash: This line isn't needed.

src/yuzu/loading_screen.h
```diff
@@ -43,7 +48,11 @@ class LoadingScreen : public QWidget {
...
+    void OnLoadProgress(VideoCore::LoadCallbackStage stage, std::size_t value, std::size_t total);
```
lioncash: This can technically be private, given it's only used by the LoadingScreen class itself.

jroweboy: Its true, but when we will hook it up with the shader cache, we will need to hook the slot from an external file so I will leave it public.

jroweboy:
Review comments addressed. Reminder that #2034 needs to be merged first and I will rebase this one of #2034.

jroweboy:
Rebased off master and hid the progress bar when its in Prepare state.

src/yuzu/loading_screen.cpp
```diff
@@ -26,10 +29,59 @@
...
+
+    stage_translations = {
+        {VideoCore::LoadCallbackStage::Prepare, tr("Loading...")},
+        {VideoCore::LoadCallbackStage::Raw, tr("Preparing Shaders %1 / %2")},
+        {VideoCore::LoadCallbackStage::Binary, tr("Loading Shaders %1 / %2")},
+        {VideoCore::LoadCallbackStage::Complete, tr("Launching...")},
+    };
+    progressbar_style = {
+        {VideoCore::LoadCallbackStage::Prepare,
+         R"(
```
ReinUsesLisp: IMO this strings would be better as an constexpr outside of the function scope.

jroweboy:
* Moved the progress bar style strings into constexpr strings.
* Changed the estimated time so it won't flash if things speed up
* Renamed (Raw, Binary) stages to (Decompile, Build) to try and be more descriptive.

jroweboy:
* Changed the background of complete stage to blue. Since the complete stage is meant to be used in "indeterminant" aka idle mode, this should create an effect of red and blue chunks alternating over the progressbar.
* Remove the blue border from around the loading screen
* Set minimum size to 1280, 720 for the widget
* Add a fade out to black effect after the first frame is drawn.

src/yuzu/loading_screen.cpp
```diff
@@ -25,11 +30,84 @@
...
+  background-color: black;
+  border: 2px solid white;
+  border-radius: 4px;
+  padding: 2px;
+}
+QProgressBar::chunk {
+ background-color: #ff3c28;
+})";
+
+constexpr const char* PROGRESSBAR_STYLE_COMPLETE = R"(
```
DarkLordZach: For all of these I'm pretty sure you can do `constexpr const char NAME[] = ` instead and it saves space from allocating the pointer.

src/video_core/rasterizer_interface.h
```diff
@@ -4,13 +4,22 @@
...
 #include "video_core/engines/fermi_2d.h"
 #include "video_core/gpu.h"
 #include "video_core/memory_manager.h"
 
 namespace VideoCore {
 
+enum class LoadCallbackStage {
+    Prepare,
+    Decompile,
+    Build,
```
DarkLordZach: `BuildShaders` or similar please

src/video_core/rasterizer_interface.h
```diff
@@ -4,13 +4,22 @@
...
 #include "common/common_types.h"
 #include "video_core/engines/fermi_2d.h"
 #include "video_core/gpu.h"
 #include "video_core/memory_manager.h"
 
 namespace VideoCore {
 
+enum class LoadCallbackStage {
+    Prepare,
+    Decompile,
```
DarkLordZach: `DecompileShaders` or similar please

PR #2048, opened 2019-01-22
Port citra-emu/citra#3924: "citra_qt: Settings (configuration) rework"

FearlessTobi:
See citra-emu/citra#3924.

  
**Original description:**
This is to fix citra-emu/citra#3531.
Design:
1. When a setting does not has a default value, it is stored in the same way as before. (Simple QSettings  entry)
1. When a setting has a default value, it is separated into two entries: `xxxxx\default` and `xxxxx`. If `xxxxx\default` = true, it means that the setting is using its default value, and when the configuration is loaded, the setting will be automatically replaced with the new default value (if any). The `xxxxx` entry stores the setting value, however it is only used when `xxxxx\default` = false. See the `ReadSetting` and `WriteSetting` functions for more detail.
1. In this way the new config file will be compatible with the old file.
1. I did not make any changes to the core/settings part. It remained as-is.

Advantages:
1. We can tell if a user is using the default setting or not and upgrade them to the new defaults when any is changed.
1. Well, I can't think of more.

Concerns:
1. ~~Do we really want to upgrade users to new default settings silently? What if I intentionally set it to some value, and the value happened to be the default value? But~~ I think for a common user it is good enough to use the default settings most of the time.

### DIFF
--- a/src/yuzu/configuration/config.h
+++ b/src/yuzu/configuration/config.h
@@ -42,6 +42,11 @@ private:
     void SaveMouseValues();
     void SaveTouchscreenValues();
 
+    QVariant ReadSetting(const QString& name) const;
+    QVariant ReadSetting(const QString& name, const QVariant& default_value) const;
+    void WriteSetting(const QString& name, const QVariant& value);
+    void WriteSetting(const QString& name, const QVariant& value, const QVariant& default_value);
+
     std::unique_ptr<QSettings> qt_config;
     std::string qt_config_loc;
 };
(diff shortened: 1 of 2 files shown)


src/yuzu/configuration/config.h
```diff
@@ -42,6 +42,11 @@ class Config {
     void SaveMouseValues();
     void SaveTouchscreenValues();
 
+    QVariant ReadSetting(const QString& name);
```
lioncash:
```suggestion
    QVariant ReadSetting(const QString& name) const;
```

src/yuzu/configuration/config.h
```diff
@@ -42,6 +42,11 @@ class Config {
...
+    QVariant ReadSetting(const QString& name, const QVariant& default_value);
```
lioncash:
```suggestion
    QVariant ReadSetting(const QString& name, const QVariant& default_value) const;
```

FearlessTobi:
Rebased. Please try Canary-tagging again @bunnei as this holding up another open PR of mine.

PR #2055, opened 2019-01-25
Asynchronous GPU command processing

bunnei:
This PR has as base #2021 (which has LDG, the shader IR, and disk based shader cache) in order to canary merge - hence the diff. 199ffa2 is the first commit

This is a second attempt at asynchronous GPU emulation. This supersedes #2012, and should fix the issues which that change introduced.

This moves GPU command processing to another thread. This works by letting the CPU core (main) thread push "GPU commands", while the GPU thread waits for commands to be available. After a push, the main thread signals that commands are available, and the GPU thread wakes up and continues to process these until the command queue is empty. The current GPU commands are:
* `SubmitList` - Pushes a GPU command list into the DmaPusher interface
* `SwapBuffers` - Swaps the output framebuffers
* `FlushRegion` - Flushes a cached region
* `InvalidateRegion` - Invalidates a cached region
* `FlushAndInvalidateRegion` - Flushes and invalidates a cached region (one command to reduce threading overhead)

Commands are queued, and executed in the order that they are pushed. Some commands are executed concurrently (e.g. SubmitList), while others must block the CPU thread (e.g. FlushRegion and InvalidateRegion). Some commands can be optionally executed on the CPU thread in the event that they are both blocking AND the GPU command queue is empty (e.g. InvalidateRegion).

There are no "emulated" CPU <-> GPU synchronization mechanisms here, so it's possible that this might cause issues. As such, it's behind a user setting. That being said, it seems quite stable, and I do not notice any bugs/regressions with the games that I have. In lieu of emulated synchronization mechanisms, we use SwapBuffers as a barrier - e.g., wait until SwapBuffers (end of frame) has been handled by the GPU before allowing the main thread to continue to the next frame.

Recommend reviewing commit-by-commit.

### DIFF
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -12,7 +12,7 @@
 #include "video_core/engines/maxwell_3d.h"
 #include "video_core/engines/maxwell_dma.h"
 #include "video_core/gpu.h"
-#include "video_core/rasterizer_interface.h"
+#include "video_core/renderer_base.h"
 
 namespace Tegra {
 
@@ -28,7 +28,8 @@ u32 FramebufferConfig::BytesPerPixel(PixelFormat format) {
     UNREACHABLE();
 }
 
-GPU::GPU(Core::System& system, VideoCore::RasterizerInterface& rasterizer) {
+GPU::GPU(Core::System& system, VideoCore::RendererBase& renderer) : renderer{renderer} {
+    auto& rasterizer{renderer.Rasterizer()};
     memory_manager = std::make_unique<Tegra::MemoryManager>();
     dma_pusher = std::make_unique<Tegra::DmaPusher>(*this);
     maxwell_3d = std::make_unique<Engines::Maxwell3D>(system, rasterizer, *memory_manager);
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -6,14 +6,12 @@
 #include "core/core.h"
 #include "core/core_timing.h"
 #include "core/memory.h"
-#include "core/settings.h"
 #include "video_core/engines/fermi_2d.h"
 #include "video_core/engines/kepler_compute.h"
 #include "video_core/engines/kepler_memory.h"
 #include "video_core/engines/maxwell_3d.h"
 #include "video_core/engines/maxwell_dma.h"
 #include "video_core/gpu.h"
-#include "video_core/gpu_thread.h"
 #include "video_core/renderer_base.h"
 
 namespace Tegra {
@@ -39,10 +37,6 @@ GPU::GPU(Core::System& system, VideoCore::RendererBase& renderer) : renderer{ren
     kepler_compute = std::make_unique<Engines::KeplerCompute>(*memory_manager);
     maxwell_dma = std::make_unique<Engines::MaxwellDMA>(system, rasterizer, *memory_manager);
     kepler_memory = std::make_unique<Engines::KeplerMemory>(system, rasterizer, *memory_manager);
-
-    if (Settings::values.use_asynchronous_gpu_emulation) {
-        gpu_thread = std::make_unique<VideoCommon::GPUThread::ThreadManager>(renderer, *dma_pusher);
-    }
 }
 
 GPU::~GPU() = default;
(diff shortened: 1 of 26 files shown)


src/video_core/gpu.cpp
```diff
@@ -62,6 +69,48 @@ const DmaPusher& GPU::DmaPusher() const {
     return *dma_pusher;
 }
 
+void GPU::PushGPUEntries(Tegra::CommandList&& entries) {
```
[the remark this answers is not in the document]
jroweboy: Conflicting design suggestion (not sure which one is better): Make GPU always be sync, and make a second class AsyncGPU that holds a ThreadManager inside it. ThreadManager will hold the GPU reference, and AsyncGPU simply pushes different commands into the queue, ThreadManager will pop from the queue and simply call the method on GPU. This has the effect that what to do for PushGPUEntries, SwapBuffers, etc will all end up calling the methods on GPU whether its Sync or Async. Similar to the other suggestion, you'll only need to do an `if settings` at the creation point, but now Async will eventually call the methods in the Sync version, keeping the ~~very little~~ logic the same between the two.

Maybe some hybrid of the two is fine 🤷‍♂️ 

Either way this is just some things I was thinking about from a design perspective while i was porting this back to citra ;D

src/video_core/gpu_thread.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include "video_core/gpu.h"
+#include "video_core/gpu_thread.h"
+#include "video_core/renderer_base.h"
+
+namespace VideoCommon::GPUThread {
+
+/// Executes a single GPU thread command
+static void ExecuteCommand(CommandData* command, VideoCore::RendererBase& renderer,
+                           Tegra::DmaPusher& dma_pusher) {
+    if (const auto submit_list = std::get_if<SubmitListCommand>(command)) {
```
jroweboy: Consider using `std::visit` like in the 4th example here. https://en.cppreference.com/w/cpp/utility/variant/visit

Note: I don't really know if theres crazy overheard or something weird like that :P

src/video_core/gpu_thread.h
```diff
@@ -0,0 +1,142 @@
...
+    CommandQueue* pop_queue{&command_queues[1]};
+
+    /// Returns true if the GPU thread should be idle, meaning there are no commands to process
+    bool IsIdle() const {
+        return command_queues[0].empty() && command_queues[1].empty();
+    }
+
+    /// Swaps the write queue (push_queue) and the read queue (pop_queue)
+    void SwapQueues() {
+        CommandQueue* const next_push_queue = pop_queue;
```
jroweboy: can be replaced with `std::swap`

src/video_core/gpu_thread.h
```diff
@@ -0,0 +1,142 @@
...
+    void PushCommand(CommandData&& command_data, bool wait_for_idle, bool allow_on_cpu);
+
+    /// Returns true if this is called by the GPU thread
+    bool IsGpuThread() const {
+        return std::this_thread::get_id() == thread_id;
+    }
+
+private:
+    SynchState state;
+    std::unique_ptr<std::thread> thread;
```
jroweboy: Is there any reason for this to be an `std::unique_ptr<std::thread>`? Could just as easily be a member variable and not a pointer without changing much.

src/yuzu/configuration/configure_graphics.cpp
```diff
@@ -75,6 +75,12 @@ void ConfigureGraphics::setConfiguration() {
     ui->frame_limit->setValue(Settings::values.frame_limit);
     ui->use_disk_shader_cache->setChecked(Settings::values.use_disk_shader_cache);
     ui->use_accurate_gpu_emulation->setChecked(Settings::values.use_accurate_gpu_emulation);
+    ui->use_asynchronous_gpu_emulation->setEnabled(!Core::System::GetInstance().IsPoweredOn());
+    ui->use_asynchronous_gpu_emulation->setChecked(Settings::values.use_asynchronous_gpu_emulation);
+    bg_color = QColor::fromRgbF(Settings::values.bg_red, Settings::values.bg_green,
```
jroweboy: unintended change likely from a messed up merge conflict resolution. (The way this is handled in master is below this `UpdateBackgroundColorButton`) See this for more info https://github.com/yuzu-emu/yuzu/pull/2057/files#diff-2bbbfeb4abc95e27da22871f1c648c0dL79

src/yuzu/configuration/configure_graphics.ui
```diff
@@ -62,6 +62,13 @@
            <string>Use accurate GPU emulation (slow)</string>
           </property>
          </widget>
+        </item>
+         <item>
```
jroweboy: super nitpick: this is indented one space too much

bunnei:
I've updated this thread to bring parity with the Patreon release from last week. That being said, I plan on reworking this a ton before it's ready for merging, so probably not worth reviewing to thoroughly just yet.

src/video_core/gpu_thread.cpp
```diff
@@ -0,0 +1,151 @@
...
+        state.UpdateIdleState();
+        state.push_queue->emplace(command_data);
+    }
+
+    // Signal the GPU thread that commands are pending
+    state.signal_condition.notify_one();
+
+    if (wait_for_idle) {
+        // Wait for the GPU to be idle (all commands to be executed)
+        std::unique_lock<std::mutex> lock{state.idle_mutex};
```
degasus: Please add a microprofile_scope here.

Narugakuruga:
This PR causes Cave Story+ high CPU usage and performance regression.

bunnei:
I've decided to change this to default OFF (due to some games having issues), and merge this in its current form. This will be replaced in an upcoming change, but that still builds off of this initial implementation. Furthermore, this has been blocking other changes from being canary tagged.

PR #2057, opened 2019-01-25
Port citra-emu/citra#4586: "Use QPixmap/QIcon for background color selection button"

FearlessTobi:
See citra-emu/citra#4586 for more details why this change is made.

### DIFF
--- a/src/yuzu/configuration/configure_graphics.cpp
+++ b/src/yuzu/configuration/configure_graphics.cpp
@@ -62,9 +62,7 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
         const QColor new_bg_color = QColorDialog::getColor(bg_color);
         if (!new_bg_color.isValid())
             return;
-        bg_color = new_bg_color;
-        ui->bg_button->setStyleSheet(
-            QString("QPushButton { background-color: %1 }").arg(bg_color.name()));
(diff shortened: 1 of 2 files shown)


src/yuzu/configuration/configure_graphics.cpp
```diff
@@ -63,8 +63,10 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
         if (!new_bg_color.isValid())
             return;
         bg_color = new_bg_color;
-        ui->bg_button->setStyleSheet(
-            QString("QPushButton { background-color: %1 }").arg(bg_color.name()));
+        QPixmap pixmap(ui->bg_button->size());
```
lioncash: This can likely be put into its own function like:

```cpp
void ConfigureGraphics::UpdateBackgroundColorButton(QColor color) {
    bg_color = color;

    QPixmap pixmap(ui->bg_button->size());
    pixmap.fill(bg_color);

    const QIcon color_icon(pixmap);
    ui->bg_button->setIcon(color_icon);
}
```

and then just call it like:

```cpp
UpdateBackgroundColorButton(new_bg_color);
```

here, and like:

```cpp
UpdateBackgroundColorButton(QColor::fromRgbF(Settings::values.bg_red, Settings::values.bg_green,
                                             Settings::values.bg_blue);
```

down below.

PR #2068, opened 2019-01-29
shader_ir: Clean texture management code

ReinUsesLisp:
Previous code relied on GLSL parameter order (something that's always ill-formed on an IR design). This approach passes spatial coordinates through operation nodes and array and depth compare values in the the texture metadata. It still contains an "extra" vector containing generic nodes for bias and component index (for example) which is still a bit ill-formed but it should be better than the previous approach.

This changes messy code so I prefer if it's tested on canary before merging.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -1140,7 +1140,7 @@ private:
                                   Type::HalfFloat);
     }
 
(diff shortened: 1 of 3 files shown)


ReinUsesLisp:
Removing canary tag while I investigate BotW regressions.

ReinUsesLisp:
Adding the canary tag again. Regressions should be fixed (it was a rebase mistake, TXQ registers were stored in temporary registers without setting them to real registers).

PR #2071, opened 2019-01-30
gl_rasterizer: Use DSA for textures and move swizzling to texture state

ReinUsesLisp:
This PR uses DSA to handle textures and framebuffers. Due to ARB limitations texture creation had to be moved to immutable storage (which is good since it avoids unwanted lazy allocations by the driver). This revealed a bug in `SurfaceParams::GetRect()` which returns a different value than the `GetMip*(0)` functions. Fixing that is beyond the scope of this PR.

It moves swizzle state tracking from `gl_state` to the texture, this fixes some edge-case issues like:
1. Bind a texture with swizzle RGBA and set a swizzle of RRRR => gl_state updates the swizzling
2. Bind another texture to the same binding as above with swizzle RGBA and set a swizzle of RRRR => gl_state does **not** update it because its swizzle in that binding didn't change.

(OpenGL's texture swizzling is a texture property, not a property of the state machine as a whole.)

Tested on gen 9 Intel's blob driver and Nvidia's blob driver and both seem to be working fine (there are no driver issues like with VAOs).

This may have conflicts with current canary.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
@@ -18,7 +18,6 @@
 #include "video_core/morton.h"
 #include "video_core/renderer_opengl/gl_rasterizer.h"
 #include "video_core/renderer_opengl/gl_rasterizer_cache.h"
-#include "video_core/renderer_opengl/gl_state.h"
 #include "video_core/renderer_opengl/utils.h"
 #include "video_core/surface.h"
 #include "video_core/textures/astc.h"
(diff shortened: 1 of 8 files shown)


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -906,6 +865,24 @@ void CachedSurface::UploadGLTexture(GLuint read_fb_handle, GLuint draw_fb_handle
...
 
+void CachedSurface::UpdateSwizzle(Tegra::Texture::SwizzleSource swizzle_x,
+                                  Tegra::Texture::SwizzleSource swizzle_y,
+                                  Tegra::Texture::SwizzleSource swizzle_z,
+                                  Tegra::Texture::SwizzleSource swizzle_w) {
+    const GLenum new_x = MaxwellToGL::SwizzleSource(swizzle_x);
+    const GLenum new_y = MaxwellToGL::SwizzleSource(swizzle_y);
+    const GLenum new_z = MaxwellToGL::SwizzleSource(swizzle_z);
+    const GLenum new_w = MaxwellToGL::SwizzleSource(swizzle_w);
+    if (swizzle[0] != new_x || swizzle[1] != new_y || swizzle[2] != new_z || swizzle[3] != new_w) {
```
lioncash: This can be inverted into a guard clause like:

```suggestion
    if (swizzle[0] == new_x && swizzle[1] == new_y && swizzle[2] == new_z && swizzle[3] == new_w) {
        return;
    }
```

src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -462,24 +462,28 @@ void OpenGLState::ApplyPolygonOffset() const {
 }
 
 void OpenGLState::ApplyTextures() const {
+    bool has_delta{};
+    std::size_t first{}, last{};
```
lioncash: would likely be better to declare `last` on its own line.

ReinUsesLisp:
Removed DSA for framebuffers because Nvidia's blob has a bug with them.
https://devtalk.nvidia.com/default/topic/1030494/opengl/bug-amp-amp-spec-violation-checknamedframebufferstatus-returns-gl_framebuffer_incomplete_dimensions_ext-under-gl-4-5-core/

PR #2072, opened 2019-01-30
service: Update function tables

lioncash:
Updates the function tables based off information on SwitchBrew. Gets rid of a swath of unknown names in a few services.


lioncash:
Buildbot error seems unrelated to the code changes

PR #2075, opened 2019-01-30
service/nvflinger: Minor renaming changes

lioncash:
Converts `Get` functions that don't simply "get" something into `Find`. This also makes FindBufferQueueId() a const qualified member function, since it doesn't modify instance state.

This is just a preliminary cleanup before more changes related to the vi service are made following this.


lioncash:
Failure seems to be unrelated to the changes

PR #2082, opened 2019-02-03
Fix TXQ not using the component mask.

FernandoS27:
Simple Fix.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -981,6 +981,10 @@ union Instruction {
             }
             return false;
         }
+
+        bool IsComponentEnabled(std::size_t component) const {
+            return ((1ul << component) & component_mask) != 0;
+        }
     } txq;
 
     union {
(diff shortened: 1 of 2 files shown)


src/video_core/engines/shader_bytecode.h
```diff
@@ -981,6 +981,11 @@ union Instruction {
             }
             return false;
         }
+
+        bool IsComponentEnabled(std::size_t component) const {
+            return ((1ul << component) & component_mask) != 0;
+        }
+
```
ReinUsesLisp: nit: Unneeded new line here.

src/video_core/engines/shader_bytecode.h
```diff
@@ -981,6 +981,10 @@ union Instruction {
             }
             return false;
         }
+
+        bool IsComponentEnabled(std::size_t component) const {
+            return ((1ul << component) & component_mask) != 0;
```
lioncash:
```suggestion
            return ((1ULL << component) & component_mask) != 0;
```

unsigned long is 32-bit on Windows. (It's also better to uppercase literals, since `L` can't be mistaken for a `1` depending on the font used.

PR #2084, opened 2019-02-04
speed up memory page mapping

Annomatg:
split up of type and mapping writes to have better cache optimizing
using std::fill which is much more efficient than doing it byte by byte


lioncash:
I think your rebase failed

Annomatg:
Yeah, it definitely failed. I have to cancel this PR and clean up

ReinUsesLisp:
That's because CI's MinGW build threw a time-out.

PR #2092, opened 2019-02-07
shader/memory: Implement STG and global memory flushing

ReinUsesLisp:
Also adds support for bigger global memory buffers (now it uses hardware's limit). Flushing to guest memory is dangerous so this add a warning when that action is done.

Some game shaders in their global memory operations are being tracked with a bad constant buffer address, fixing that's beyond the scope of this PR.


bunnei:
@ReinUsesLisp can you rebase?

bunnei:
@ReinUsesLisp what's the status of this one?

PR #2093, opened 2019-02-07
Better LZ4 compression utilization for the disk based shader cache and the yuzu build system

FreddyFunk:
This PR addresses three issues in three commits.

1. `video_core` should not be linked with `lz4_static`, since LZ4 is a common general purpose compression algorithm and has nothing directly to do with rendering. The first commit moves the LZ4 compression logic for compression and decompression of transferable and precompiled shader disk chache files to `common/data_compression`. 
2. Use LZ4HC with the highst possible compression level instead of the default LZ4 compression for precompiled shader disk chache files. This results in smaller files. The compressed file sizes can be improved futher, but this requires a redisgn of the chache file structure and exceeds the scope of this PR.
3. `lz4_static` was linked to `core` before, but is now eliminated. Now `common/data_compression` is utilized to decompress nso segments instead.


src/common/data_compression.h
```diff
@@ -0,0 +1,23 @@
...
+
+#pragma once
+
+#include <vector>
+
+#include "common/common_types.h"
+
+namespace Compression {
+
+// Compresses a source memory region with LZ4 and returns the compressed data in an vector. If
```
MysticExile:
```suggestion
// Compresses a source memory region with LZ4 and returns the compressed data in a vector. If
```
nit

src/common/data_compression.h
```diff
@@ -0,0 +1,23 @@
...
+// use_LZ4_high_compression is true, the LZ4 subalgortihmn LZ4HC is used with the highst possible
```
MysticExile:
```suggestion
// use_LZ4_high_compression is true, the LZ4 subalgortihmn LZ4HC is used with the highest possible
```
nit

DarkLordZach:
Please conduct some sort of testing/experimentation to verify that the performance impact is nonexistent or negligible. If it's significant, however, I'm going to have to ask that you make this a config option. While some users may be able to justify a space over speed trade off, others won't/don't.

src/common/data_compression.cpp
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <lz4hc.h>
+
+#include "data_compression.h"
+
+namespace Compression {
```
lioncash:
```suggestion
namespace Common {
```

or 

```suggestion
namespace Common::Compression {
```

src/common/data_compression.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+std::vector<u8> CompressDataLZ4(const u8* source, std::size_t source_size,
+                                bool use_LZ4_high_compression) {
+    if (source_size > LZ4_MAX_INPUT_SIZE) {
```
lioncash: This should probably be an assertion rather than silently returning an empty vector, since it's usually a caller screwing up in this case.

src/common/data_compression.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <lz4hc.h>
+
+#include "data_compression.h"
```
lioncash:
```suggestion
#include "common/data_compression.h"
```

A better naming scheme here would also be `common/lz4_compression.h` (along with changing the cpp file). Making a singular cpp/h file for compression algorithms can get messy over time.

src/common/data_compression.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
```
lioncash: This isn't necessary in a cpp file.

src/common/data_compression.h
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <vector>
+
+#include "common/common_types.h"
+
+namespace Compression {
```
lioncash: See previous comment about the namespace naming.

src/common/data_compression.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <lz4hc.h>
+
+#include "data_compression.h"
+
+namespace Compression {
+
+std::vector<u8> CompressDataLZ4(const u8* source, std::size_t source_size,
+                                bool use_LZ4_high_compression) {
```
lioncash: This should allow specifying the compression level aside from "default" and "max". A helper function like `CompressDataLZ4Max` can be introduced that forwards to:

```cpp
std::vector<u8> CompressDataLZ4Max(const u8* source, std::size_t source_size) {
    return CompressDataLZ4(source, source_size, LZ4HC_CLEVEL_MAX);
}
```

and this if statement:

```cpp
if (use_LZ4_high_compression) {
    compressed_size = LZ4_compress_HC(reinterpret_cast<const char*>(source),
                                      reinterpret_cast<char*>(compressed.data()),
                                      source_size_int, max_compressed_size, LZ4HC_CLEVEL_MAX);
} else {
    compressed_size = LZ4_compress_default(reinterpret_cast<const char*>(source),
                                           reinterpret_cast<char*>(compressed.data()),
                                           source_size_int, max_compressed_size);
}
```

would be changed to:

```cpp
if (compression_level == LZ4HC_CLEVEL_DEFAULT) {
    compressed_size = LZ4_compress_default(reinterpret_cast<const char*>(source),
                                           reinterpret_cast<char*>(compressed.data()),
                                           source_size_int, max_compressed_size);
} else {
    compressed_size = LZ4_compress_HC(reinterpret_cast<const char*>(source),
                                      reinterpret_cast<char*>(compressed.data()),
                                      source_size_int, max_compressed_size, LZ4HC_CLEVEL_MAX);
}
```

FreddyFunk:
@MysticExile @lioncash thanks for the feedback. Should be good now.

@DarkLordZach Decompression speed is not impacted. Compression speed is slower, but since it has to be done only once (and decompression all the time) it sould be fine. Check out [lzbench](https://github.com/inikep/lzbench). This might answer your question for excat numbers.

src/common/lz4_compression.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    compressed.resize(compressed_size);
+
+    return compressed;
+}
+
+std::vector<u8> CompressDataLZ4HC(const u8* source, std::size_t source_size,
+                                  s32 compression_level) {
+    ASSERT_MSG(source_size > LZ4_MAX_INPUT_SIZE, "Source size exceeds LZ4 maximum input size");
```
lioncash:
```suggestion
    ASSERT_MSG(source_size <= LZ4_MAX_INPUT_SIZE, "Source size exceeds LZ4 maximum input size");
```

otherwise this will always assert for sizes within the valid range.

src/common/lz4_compression.cpp
```diff
@@ -0,0 +1,83 @@
...
+        // Compression failed
+        return {};
+    }
+
+    compressed.resize(compressed_size);
+
+    return compressed;
+}
+
+std::vector<u8> CompressDataLZ4HCMAX(const u8* source, std::size_t source_size) {
```
lioncash:
```suggestion
std::vector<u8> CompressDataLZ4HCMax(const u8* source, std::size_t source_size) {
```

Max ain't an abbreviation :p

src/common/lz4_compression.h
```diff
@@ -0,0 +1,24 @@
...
+// Refer to the license.txt file included.
+
+#include <vector>
+
+#include "common/common_types.h"
+
+namespace Common::Compression {
+
+// Compresses a source memory region with LZ4 and returns the compressed data in a vector. If
+// use_LZ4_high_compression is true, the LZ4 subalgortihmn LZ4HC is used with the highest possible
```
lioncash: This is no longer relevant, also:

```suggestion
// use_LZ4_high_compression is true, the LZ4 subalgortihm LZ4HC is used with the highest possible
```

These functions should likely also have doxygen compatible descriptions.

src/common/lz4_compression.cpp
```diff
@@ -0,0 +1,77 @@
...
+    compressed.resize(compressed_size);
+
+    return compressed;
+}
+
+std::vector<u8> CompressDataLZ4HC(const u8* source, std::size_t source_size,
+                                  s32 compression_level) {
+    ASSERT_MSG(source_size <= LZ4_MAX_INPUT_SIZE, "Source size exceeds LZ4 maximum input size");
+
+    compression_level = std::clamp(compression_level, LZ4HC_CLEVEL_MIN, LZ4HC_CLEVEL_MAX);
```
lioncash: `<algorithm>` needs to be included for `std::clamp`.

src/common/lz4_compression.h
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <vector>
+
+#include "common/common_types.h"
+
+namespace Common::Compression {
+
+/*! Compresses a source memory region with LZ4 and returns the compressed data in a vector.
```
lioncash: our doxygen comments use `///` or:

```cpp
/**
 *  Description
 ... etc as necessary
 */
```

and use `@` for beginning keywords, like `@param` and `@return`.

src/common/lz4_compression.h
```diff
@@ -0,0 +1,44 @@
...
+
+\param source the uncompressed source memory region.
+\param source_size the size in bytes of the uncompressed source memory region.
+
+\return the compressed data in a std::vector<u8>.
```
lioncash: `the compressed data` is sufficient here.

src/common/lz4_compression.h
```diff
@@ -0,0 +1,44 @@
...
+/*! Utilizes the LZ4 subalgorithm LZ4HC with the specified compression level. Higher compression
+levels result in a smaller compressed size, but require more CPU time for compression. The
+compression level has almost no impact on decompression speed. Data compressed with LZ4HC can also
+be decompressed with the default LZ4 decompression.
+
+\param source the uncompressed source memory region.
+\param source_size the size in bytes of the uncompressed source memory region.
+\param compression_level the used compression level. Should be between 3 and 12.
+
+\return the compressed data in a std::vector<u8>.
```
lioncash: ditto

src/common/lz4_compression.h
```diff
@@ -0,0 +1,44 @@
...
+*/
+std::vector<u8> CompressDataLZ4HC(const u8* source, std::size_t source_size, s32 compression_level);
+
+/*! Utilizes the LZ4 subalgorithm LZ4HC with the highest possible compression level.
+
+\param source the uncompressed source memory region.
+\param source_size the size in bytes of the uncompressed source memory region.
+
+\return the compressed data in a std::vector<u8>.
```
lioncash: Ditto

src/common/lz4_compression.h
```diff
@@ -0,0 +1,44 @@
...
+*/
+std::vector<u8> CompressDataLZ4HCMax(const u8* source, std::size_t source_size);
+
+std::vector<u8> DecompressDataLZ4(const std::vector<u8>& compressed, std::size_t uncompressed_size);
```
lioncash: This is missing a doxygen comment.

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+#include "common/common_types.h"
+
+namespace Common::Compression {
+
+/**
+* Compresses a source memory region with LZ4 and returns the compressed data in a vector.
+*
+* @param source the uncompressed source memory region.
+* @param source_size the size in bytes of the uncompressed source memory region.
+
```
lioncash: Missing a `*` here

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+// Refer to the license.txt file included.
+
+#include <vector>
+
+#include "common/common_types.h"
+
+namespace Common::Compression {
+
+/**
+* Compresses a source memory region with LZ4 and returns the compressed data in a vector.
```
lioncash: These single `*` lines should be spaced so that the align like:

```cpp
/**
 *
```

not like:

```cpp
/**
*
```

Ditto below where applicable.

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+*
+* @param source the uncompressed source memory region.
+* @param source_size the size in bytes of the uncompressed source memory region.
+
+* @return the compressed data.
+*/
+std::vector<u8> CompressDataLZ4(const u8* source, std::size_t source_size);
+
+/** Utilizes the LZ4 subalgorithm LZ4HC with the specified compression level. Higher compression
```
lioncash: This should go on a line below the opening `/**`

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+* levels result in a smaller compressed size, but require more CPU time for compression. The
+* compression level has almost no impact on decompression speed. Data compressed with LZ4HC can also
+* be decompressed with the default LZ4 decompression.
+
```
lioncash: Missing a `*` on this line.

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+@param source the uncompressed source memory region.
+@param source_size the size in bytes of the uncompressed source memory region.
+@param compression_level the used compression level. Should be between 3 and 12.
+
```
lioncash: Ditto

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+@return the compressed data.
+*/
+std::vector<u8> CompressDataLZ4HC(const u8* source, std::size_t source_size, s32 compression_level);
+
+/** Utilizes the LZ4 subalgorithm LZ4HC with the highest possible compression level.
```
lioncash: This opening description should go below the `/**`

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+
```
lioncash: Missing a `*` on this line.

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+@param source the uncompressed source memory region.
+@param source_size the size in bytes of the uncompressed source memory region.
+
```
lioncash: Ditto

src/common/lz4_compression.h
```diff
@@ -0,0 +1,53 @@
...
+@return the compressed data.
+*/
+std::vector<u8> CompressDataLZ4HCMax(const u8* source, std::size_t source_size);
+
+/**
+* Decompresses a source memory region with LZ4 and returns the uncompressed data in a vector.
+*
+* @param compressed the compressed source memory region.
+* @param uncompressed_size the size in bytes of the uncompressed data.
+
```
lioncash: Missing a `*` on this line.

jroweboy:
@DarkLordZach I think there is a misunderstanding here. As far as I understand it, we do not do any compression for shaders during game play. When the game is running, we only add new entries to the transferrable cache, which is not compressed. Then during the loading screen, we will decompile from transferrable and compress and write to precompiled cache, after which, we also compile that shader and use glGetProgramBinary to compress and write this to the precompiled cache as well.

Because of this, it doesn't affect actual game runtime performance, and only affect the time it takes in the loading screen, and thats something I think is fine. Theres plenty of opportunity to speed up the loading screen by multithreading it, and adding on some extra time to the loading screen for a 30% better compression rate is something very much welcome :)

FreddyFunk:
@jroweboy thanks for clarifying that. I would also suggest to not making a user setting for the compression level.

FreddyFunk:
Check out #2098 for some compression measurements

bunnei:
Sorry for the delay here- this has kind of slipped through the cracks. Let's let this soak in canary a bit, and if there are no issues, we can merge.

Ping for others to review if they want to have a say here.

bunnei:
@FreddyFunk sorry - can you rebase this one more time?

FreddyFunk:
@bunnei rebased

PR #2095, opened 2019-02-08
frontend: Open transferable shader cache for a selected game in the gamelist

FreddyFunk:
This PR adds option to open the transferable shader cache.
image

On every OS the transferable shader cache folder opens. On Windows devices the transferable shader cache file for the selected game gets preselected as well.
image

If no transferable shader cache is present for the selected game, the following warning opens. No file explorer window opens in this case.
image

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1072,7 +1072,7 @@ void GMainWindow::OnTransferableShaderCacheOpenFile(u64 program_id) {
     ASSERT(program_id != 0);
 
     std::string transferable_shader_cache_file_path;
-    const std::string open_target = "Transferable Shader Cache";
+    constexpr char open_target[] = "Transferable Shader Cache";
     const std::string tranferable_shader_cache_folder =
         FileUtil::GetUserPath(FileUtil::UserPath::ShaderDir) + "opengl" + DIR_SEP "transferable";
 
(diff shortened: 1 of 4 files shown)


src/yuzu/main.cpp
```diff
@@ -1066,6 +1068,48 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
     QDesktopServices::openUrl(QUrl::fromLocalFile(qpath));
 }
 
+void GMainWindow::OnTransferableShaderCacheOpenFile(u64 program_id) {
+    ASSERT(program_id != 0);
+
+    std::string transferable_shader_cache_file_path;
```
lioncash: This should be moved closer to its usage. This should also likely be a QString.

src/yuzu/main.cpp
```diff
@@ -1066,6 +1068,48 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+    const std::string open_target = "Transferable Shader Cache";
```
```suggestion
    constexpr char open_target[] = "Transferable Shader Cache";
```

src/yuzu/main.cpp
```diff
@@ -1066,6 +1068,48 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+    const std::string tranferable_shader_cache_folder =
+        FileUtil::GetUserPath(FileUtil::UserPath::ShaderDir) + "opengl" + DIR_SEP "transferable";
+
+    transferable_shader_cache_file_path.append(tranferable_shader_cache_folder);
```
```suggestion
    transferable_shader_cache_file_path += tranferable_shader_cache_folder;
```

etc

PR #2096, opened 2019-02-08
nvdisp_disp0: change drawing message log level from Warning to Trace

FearlessTobi:
This is a leftover from the early yuzu days. 
We shouldn't log every time when we are drawing by default, so let's change the log level to Trace.

As suggested by @Hexagon12 on Discord.


FearlessTobi:
I need to fix clang before we can merge this.

PR #2099, opened 2019-02-09
Implement BGRA8 framebuffer format

greggameplayer:
tested on Lego Ninjago and seems to be good.


greggameplayer:
@DarkLordZach Based on https://github.com/switchbrew/libnx/blob/master/nx/include/switch/display/types.h

ogniK5377:
That’s a list of enums. Not a hardware test

bunnei:
Agree with @FernandoS27

PR #2106, opened 2019-02-11
Update cpu_core_manager.cpp

MalekTaktak:
Fixed for loop with incorrect index for core initializing


jroweboy:
The original code is correct. The index 0-2 for `core_threads` references the host threads 1-3 as the core 0 is emulated directly on the main `emu_thread`. Your changes will put threads 1-2 into core_threads, leaving index 0 uninitialized. Thanks for trying to look into this though 👍

PR #2113, opened 2019-02-12
vulkan: Add dependencies and device abstraction

ReinUsesLisp:
Since #1114 is a very old base to work on top of, the intention is to open individual pull requests for Vulkan sections that can be built and reviewed individually. This is (in my opinion) much better than opening a single PR with +7000 lines of code.

This adds [Vulkan-Headers](https://github.com/KhronosGroup/Vulkan-Headers/) as dependency without requiring any external linkage dependencies or SDK, all Vulkan function pointers can be queried at runtime if `vkGetInstanceProcAddr` is provided (something that's exposed in SDL2, GLFW and Qt).

Aside from that, it adds a device abstraction class which is dead-code at the moment. Description in 4d16376.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_device.cpp
+++ b/src/video_core/renderer_vulkan/vk_device.cpp
@@ -1,0 +1,231 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <map>
+#include <optional>
+#include <set>
+#include <vector>
+#include "common/assert.h"
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/renderer_vulkan/vk_device.h"
+
+namespace Vulkan {
+
+namespace Alternatives {
+
+constexpr std::array<vk::Format, 3> Depth24UnormS8Uint = {
+    vk::Format::eD32SfloatS8Uint, vk::Format::eD16UnormS8Uint, {}};
+constexpr std::array<vk::Format, 3> Depth16UnormS8Uint = {
+    vk::Format::eD24UnormS8Uint, vk::Format::eD32SfloatS8Uint, {}};
+
+} // namespace Alternatives
+
+constexpr const vk::Format* GetFormatAlternatives(vk::Format format) {
+    switch (format) {
+    case vk::Format::eD24UnormS8Uint:
+        return Alternatives::Depth24UnormS8Uint.data();
+    case vk::Format::eD16UnormS8Uint:
+        return Alternatives::Depth16UnormS8Uint.data();
+    default:
+        return nullptr;
+    }
+}
+
+constexpr vk::FormatFeatureFlags GetFormatFeatures(vk::FormatProperties properties,
+                                                   FormatType format_type) {
+    switch (format_type) {
+    case FormatType::Linear:
+        return properties.linearTilingFeatures;
+    case FormatType::Optimal:
+        return properties.optimalTilingFeatures;
+    case FormatType::Buffer:
+        return properties.bufferFeatures;
+    default:
+        return {};
+    }
+}
+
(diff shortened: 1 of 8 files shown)


src/video_core/renderer_vulkan/vk_device.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <optional>
+#include <set>
+#include <vector>
+#include "common/assert.h"
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/renderer_vulkan/vk_device.h"
+
+namespace Vulkan {
+
+static const std::map<vk::Format, std::vector<vk::Format>> FormatAlternatives = {
```
lioncash: You may want to avoid a std::map here, and use a std::array or something else, since this allocates on the heap for the entire duration of the program and can't be compile-time stored in the read-only segment.

src/video_core/renderer_vulkan/vk_device.cpp
```diff
@@ -0,0 +1,216 @@
...
+    SetupProperties(dldi);
+}
+
+VKDevice::~VKDevice() = default;
+
+bool VKDevice::Create(const vk::DispatchLoaderDynamic& dldi, vk::Instance instance) {
+    const auto queue_cis = GetDeviceQueueCreateInfos();
+    vk::PhysicalDeviceFeatures device_features{};
+
+    std::vector<const char*> extensions;
```
lioncash:
```cpp
std::vector<const char*> extensions = {VK_KHR_SWAPCHAIN_EXTENSION_NAME};
```

src/video_core/renderer_vulkan/vk_device.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    return queue_cis;
+}
+
+std::map<vk::Format, vk::FormatProperties> VKDevice::GetFormatProperties(
+    const vk::DispatchLoaderDynamic& dldi, vk::PhysicalDevice physical) {
+    std::map<vk::Format, vk::FormatProperties> format_properties;
+
+    const auto AddFormatQuery = [&format_properties, &dldi, physical](vk::Format format) {
+        format_properties.insert({format, physical.getFormatProperties(format, dldi)});
```
lioncash:
```cpp
format_properties.emplace(format, physical.getFormatProperties(format, dldi));
```

src/video_core/renderer_vulkan/vk_device.cpp
```diff
@@ -0,0 +1,231 @@
...
+    AddFormatQuery(vk::Format::eR5G6B5UnormPack16);
+    AddFormatQuery(vk::Format::eD32Sfloat);
+    AddFormatQuery(vk::Format::eD16UnormS8Uint);
+    AddFormatQuery(vk::Format::eD24UnormS8Uint);
+    AddFormatQuery(vk::Format::eD32SfloatS8Uint);
+
+    return format_properties;
+}
+
+} // namespace Vulkan
```
bunnei: nit: Missing new line

src/video_core/renderer_vulkan/vk_device.h
```diff
@@ -0,0 +1,116 @@
...
+    vk::Queue graphics_queue;           ///< Main graphics queue
+    vk::Queue present_queue;            ///< Main present queue
+    u32 graphics_family{};              ///< Main graphics queue family index
+    u32 present_family{};               ///< Main present queue family index
+    vk::PhysicalDeviceType device_type; ///< Physical device type
+    u64 uniform_buffer_alignment{};     ///< Uniform buffer alignment requeriment
+    std::map<vk::Format, vk::FormatProperties> format_properties; ///< Format properties dictionary
+};
+
+} // namespace Vulkan
```
bunnei: nit: Missing new line

MartinBouteruche:
Does the Vulkan implementation actually work? That would be indeed very exciting

Issue #1226, opened 2018-09-02
Cant put keys in the default user folder


MojoJojoDojo:
If you put the keys (prod.keys and title.keys) in the default (AppData) folder yuzu wont recognize them.
if you create a portable user folder the keys will be detected and games will launch correctly.
I think it's because my default user folder uses non-latin characters but i could be wrong.

tech4me:
Where did you put your keys?

MojoJojoDojo:
keys folder

elemileTLDR:
I seem to have the same issue.  How did you change the keys target folder (or create a portable user one)? Yuzu only allows us to change the sdcard, games and nand folders.

Zenicius:
> I seem to have the same issue. How did you change the keys target folder (or create a portable user one)? Yuzu only allows us to change the sdcard, games and nand folders.
I dont think theres a way to change keys target folder. To be honest i have this problem quite often so i decided to change  my username...

elemileTLDR:
> 
> 
> > I seem to have the same issue. How did you change the keys target folder (or create a portable user one)? Yuzu only allows us to change the sdcard, games and nand folders.
> 
> I dont think theres a way to change keys target folder. To be honest i have this problem quite often so i decided to change my username...

I see. I can't change my username on Windows 8.1 (or I can, but it doesn't change the main user's path name), so I made a new user, with a strictly unicode name, just for Yuzu. It solved the issue, but damn.

Frakatchoo:
Same here! And I confirm, Yuzu is not compatible with user folder containing special characters (mine is "Jérôme" for exemple). The generated shortcuts to the emulator doesn't work, and the app can't access the keys folder. Hope an update for this issue, I think a lot of people can't try the emulator because of this.

Morph1984:
Fixed by #6270

PR #2117, opened 2019-02-13
Add 'Configure Keys Directory' to File Menu

Jungorend:
Addressing Issue #1226, this should allow configuring the Keys directory to a different location if desired.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -405,6 +405,13 @@ void Config::ReadValues() {
                     QString::fromStdString(FileUtil::GetUserPath(FileUtil::UserPath::SDMCDir)))
             .toString()
             .toStdString());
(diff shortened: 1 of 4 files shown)


FearlessTobi:
Sorry but I personally kinda fail to see the point of this. Instead of looking into addressing the original issue, this just adds yet another option to the user interface that's already a bit cluttered.

Jungorend:
The impression I got is it would help to move to a location to avoid non-latin usernames causing issues.

Do you feel it would be better to move all the Configure Directory into a single dedicated Window in the File Menu to avoid clutter, or look into keeping Keys in one place but reading non-latin characters without issue?

lioncash:
It would likely be better to fix the handling for non-Latin file directory paths, since that's an issue by itself.

Jungorend:
Alright.

PR #2118, opened 2019-02-13
shader_decompiler: Improve Accuracy of Attribute Interpolation.

FernandoS27:


src/video_core/engines/shader_header.h
```diff
@@ -103,6 +116,29 @@ struct Header {
...
                 return omap.target & (1 << bit);
             }
+            AttributeUse GetAttributeIndexUse(u32 attribute, u32 index) const {
+                return static_cast<AttributeUse>(
+                    (imap_generic_vector[attribute].raw >> (index * 2)) & 0x03);
+            }
+            AttributeUse GetAttributeUse(u32 attribute) const {
+                AttributeUse result = AttributeUse::Unused;
+                for (u32 i = 0; i < 4; i++) {
+                    auto index = GetAttributeIndexUse(attribute, i);
```
lioncash:
```suggestion
                    const auto index = GetAttributeIndexUse(attribute, i);
```

src/video_core/engines/shader_header.h
```diff
@@ -103,6 +116,29 @@ struct Header {
...
+                    if (index == AttributeUse::Unused) {
+                        continue;
+                    }
+                    if (result == AttributeUse::Unused || result == index) {
+                        result = index;
+                        continue;
+                    } else {
```
lioncash: The else can be omitted here, unindenting the contained code.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -345,8 +334,14 @@ class GLSLDecompiler final {
             if (stage == ShaderStage::Geometry) {
                 attr = "gs_" + attr + "[]";
             }
-            code.AddLine("layout (location = " + std::to_string(idx) + ") " +
-                         GetInputFlags(input_mode) + "in vec4 " + attr + ';');
+            if (stage == ShaderStage::Fragment) {
+                auto input_mode = header.ps.GetAttributeUse(idx - 1);
```
```suggestion
                const auto input_mode = header.ps.GetAttributeUse(idx - 1);
```

src/video_core/engines/shader_header.h
```diff
@@ -103,6 +116,29 @@ struct Header {
...
+                        LOG_CRITICAL(HW_GPU, "Generic Attribute Conflict in Interpolation Mode");
```
ReinUsesLisp: This should be an error IMO

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -288,43 +289,31 @@ class GLSLDecompiler final {
...
-            UNIMPLEMENTED_MSG("Unimplemented IPA sampler mode: {}", static_cast<u32>(sample_mode));
+            LOG_CRITICAL(HW_GPU, "Unused attribute being fetched");
+            UNREACHABLE();
         }
         return out;
     }
 
     void DeclareInputAttributes() {
         const auto& attributes = ir.GetInputAttributes();
+        const auto& header = ir.GetHeader();
```
ReinUsesLisp: The header is already passed by value in the constructor.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -345,8 +334,14 @@ class GLSLDecompiler final {
...
                 attr = "gs_" + attr + "[]";
             }
-            code.AddLine("layout (location = " + std::to_string(idx) + ") " +
-                         GetInputFlags(input_mode) + "in vec4 " + attr + ';');
+            if (stage == ShaderStage::Fragment) {
+                auto input_mode = header.ps.GetAttributeUse(idx - 1);
+                code.AddLine("layout (location = " + std::to_string(idx) + ") " +
+                             GetInputFlags(input_mode) + "in vec4 " + attr + ';');
+            } else {
+                code.AddLine("layout (location = " + std::to_string(idx) + ") " + "in vec4 " +
```
ReinUsesLisp: Since this line is basically the same as above, you could conditionally store `GetInputFlags(input_mode)` in a string and add it to a single `code.AddLine` call.

src/video_core/shader/decode/other.cpp
```diff
@@ -135,7 +135,18 @@ u32 ShaderIR::DecodeOther(NodeBlock& bb, u32 pc) {
                                                 instr.ipa.sample_mode.Value()};
 
         const Node attr = GetInputAttribute(attribute.index, attribute.element, input_mode);
-        const Node value = GetSaturatedFloat(attr, instr.ipa.saturate);
+        Node value = attr;
+        const Tegra::Shader::Attribute::Index index = attribute.index.Value();
+        if (index > Tegra::Shader::Attribute::Index::Attribute_0 &&
```
ReinUsesLisp: This is not including the attribute 0 in the test.

FernandoS27: Attribute 0 is the position, I believe.

ReinUsesLisp: It's not. Look for the `Index::Attribute_0` usages.

src/video_core/engines/shader_header.h
FernandoS27: I'd love it to spam well, I really consider it critical.

src/video_core/shader/decode/other.cpp
FernandoS27: hmn, weird it does not match the vector elements, maybe we have a mistske in the header.

FernandoS27: header is fine, there must be a discrepancy somewhere else.

ReinUsesLisp: What vector elements are you walking about? Generic attributes go from `Attribute_0` (inclusive) and `Attribute_31` (inclusive). It has to be `>=` instead of `>` otherwise you will be skipping the multiplication for the first generic attribute.

FernandoS27: The ones in the header, It seems the direction of Attribute_0 is wrong and starts from position or else the attributes don't match the header.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -288,34 +289,22 @@ class GLSLDecompiler final {
...
 
-    std::string GetInputFlags(const IpaMode& input_mode) {
-        const IpaSampleMode sample_mode = input_mode.sampling_mode;
-        const IpaInterpMode interp_mode = input_mode.interpolation_mode;
+    std::string GetInputFlags(AttributeUse attribute) {
         std::string out;
 
-        switch (interp_mode) {
-        case IpaInterpMode::Flat:
+        switch (attribute) {
```
jroweboy: This switch is missing a case for `AttributeUse::Unused` Its clear you mean for it to fall to the default case, but it can't hurt to be explicit and just do a fallthrough for it to default.

```
case AttributeUse::Unused:
default:
```

PR #2119, opened 2019-02-13
rasterizer_cache_gl: Only do fast layered copy on the same format.

FernandoS27:


lioncash:
Commit message should explain why.

PR #2120, opened 2019-02-13
Fixes Unicode Key File Directories

Jungorend:
Can load directories when on a Unicode file directory.

Code added to convert filepath from UTF8 to UTF16 on Windows, borrowed from FileUtil::Exists where it works appropriately.

### DIFF
--- a/src/core/crypto/key_manager.cpp
+++ b/src/core/crypto/key_manager.cpp
@@ -400,7 +400,7 @@ static bool ValidCryptoRevisionString(std::string_view base, size_t begin, size_
 void KeyManager::LoadFromFile(const std::string& filename, bool is_title_keys) {
 #ifdef _WIN32
     std::string copy(filename);
-    if (copy.size() != 0 && copy.back() == ':')
+    if (!copy.empty() && copy.back() == ':')
         copy += DIR_SEP_CHR;
     std::wstring utf16 = Common::UTF8ToUTF16W(copy);
     std::ifstream file(utf16);
--- a/src/core/crypto/key_manager.cpp
+++ b/src/core/crypto/key_manager.cpp
@@ -398,15 +398,8 @@ static bool ValidCryptoRevisionString(std::string_view base, size_t begin, size_
 }
 
 void KeyManager::LoadFromFile(const std::string& filename, bool is_title_keys) {
-#ifdef _WIN32
-    std::string copy(filename);
-    if (!copy.empty() && copy.back() == ':')
-        copy += DIR_SEP_CHR;
-    std::wstring utf16 = Common::UTF8ToUTF16W(copy);
-    std::ifstream file(utf16);
-#else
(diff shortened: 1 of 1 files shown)


jroweboy:
Could this code instead just use FileUtil to load the file? (Didn't look enough into it to see if this is possible, but i'd prefer to use that instead of `#ifdef`ing here)

src/core/crypto/key_manager.cpp
```diff
@@ -398,7 +398,15 @@ static bool ValidCryptoRevisionString(std::string_view base, size_t begin, size_
 }
 
 void KeyManager::LoadFromFile(const std::string& filename, bool is_title_keys) {
+#ifdef _WIN32
+    std::string copy(filename);
+    if (copy.size() != 0 && copy.back() == ':')
+        copy += DIR_SEP_CHR;
+    std::wstring utf16 = Common::UTF8ToUTF16W(copy).c_str();
```
lioncash:
```suggestion
    std::wstring utf16 = Common::UTF8ToUTF16W(copy);
```

src/core/crypto/key_manager.cpp
```diff
@@ -398,7 +398,15 @@ static bool ValidCryptoRevisionString(std::string_view base, size_t begin, size_
 }
 
 void KeyManager::LoadFromFile(const std::string& filename, bool is_title_keys) {
+#ifdef _WIN32
+    std::string copy(filename);
+    if (copy.size() != 0 && copy.back() == ':')
```
lioncash:
```suggestion
    if (!copy.empty() && copy.back() == ':')
```

Jungorend:
I included the #ifdef statements as I wanted to be sure the check made by Exists was also done on loading. It could also be handled by using FileUtil's OpenFStream if those checks are unnecessary.

zhaowenlan1779:
This should really just use `OpenFStream` instead, which already handles UTF8 to UTF16 conversion on Windows IIRC.

> I included the #ifdef statements as I wanted to be sure the check made by Exists was also done on loading.

I suppose you mean the `:` handling for Windows, right? I don't think we need that here as this is supposed to be a file path.

src/core/crypto/key_manager.cpp
```diff
@@ -398,7 +398,8 @@ static bool ValidCryptoRevisionString(std::string_view base, size_t begin, size_
 }
 
 void KeyManager::LoadFromFile(const std::string& filename, bool is_title_keys) {
-    std::ifstream file(filename);
+    std::ifstream file;
+    OpenFStream<std::ifstream>(file, filename, std::ios_base::in);
```
lioncash: This can be simplified to:

```suggestion
    OpenFStream(file, filename, std::ios_base::in);
```

The type will be auto-deduced by the template.

Jungorend:
I like simple. Applying the recommended changes.

PR #2122, opened 2019-02-14
vk_resource_manager: Implement fence and command buffer allocator

ReinUsesLisp:
`VKResourceManager` is an abstraction layer used to commit fences and command buffers from a pool on Vulkan. This avoids constantly allocating (and deallocating) resources in the hot loop. Details in the individual commits.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_resource_manager.cpp
+++ b/src/video_core/renderer_vulkan/vk_resource_manager.cpp
@@ -1,0 +1,14 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/renderer_vulkan/vk_resource_manager.h"
+
+namespace Vulkan {
+
+VKResource::VKResource() = default;
+
+VKResource::~VKResource() = default;
+
+} // namespace Vulkan
--- a/src/video_core/renderer_vulkan/vk_resource_manager.cpp
+++ b/src/video_core/renderer_vulkan/vk_resource_manager.cpp
@@ -4,11 +4,15 @@
 
 #include <algorithm>
 #include "common/assert.h"
+#include "common/logging/log.h"
 #include "video_core/renderer_vulkan/declarations.h"
 #include "video_core/renderer_vulkan/vk_device.h"
 #include "video_core/renderer_vulkan/vk_resource_manager.h"
 
 namespace Vulkan {
+
+// TODO(Rodrigo): Fine tune these numbers.
+constexpr std::size_t FENCES_GROW_STEP = 0x40;
 
 VKResource::VKResource() = default;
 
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_vulkan/vk_resource_manager.cpp
```diff
@@ -0,0 +1,287 @@
...
+    dev.resetFences({*handle}, dld);
+    is_used = false;
+    return true;
+}
+
+void VKFence::Protect(VKResource* resource) {
+    protected_resources.push_back(resource);
+}
+
+void VKFence::Unprotect(VKResource* resource) {
```
lioncash: unless I'm incorrect here, this can be:

```suggestion
void VKFence::Unprotect(const VKResource* resource) {
```

src/video_core/renderer_vulkan/vk_resource_manager.cpp
```diff
@@ -0,0 +1,287 @@
...
+// TODO(Rodrigo): Fine tune these numbers.
+constexpr std::size_t COMMAND_BUFFER_POOL_SIZE = 0x1000;
+constexpr std::size_t FENCES_GROW_STEP = 0x40;
+
+class CommandBufferPool final : public VKFencedPool {
+public:
+    CommandBufferPool(const VKDevice& device)
+        : VKFencedPool(COMMAND_BUFFER_POOL_SIZE), device{device} {}
+
+    ~CommandBufferPool() = default;
```
lioncash: This is able to be omitted.

src/video_core/renderer_vulkan/vk_resource_manager.h
```diff
@@ -0,0 +1,186 @@
...
+    virtual void Allocate(std::size_t begin, std::size_t end) = 0;
+
+private:
+    /// Manages pool overflow allocating new resources.
+    std::size_t ManageOverflow();
+
+    /// Allocates a new page of resources.
+    void Grow();
+
+    std::size_t grow_step{};       ///< Number of new resources created after an overflow
```
lioncash: Just a consistency thing, but:

```suggestion
    std::size_t grow_step = 0;       ///< Number of new resources created after an overflow
```

to match the variable below it (or change `free_iterator` to use `{}`, either one works).

src/video_core/renderer_vulkan/vk_resource_manager.h
```diff
@@ -0,0 +1,186 @@
...
+class VKFence;
+class VKResourceManager;
+
+class CommandBufferPool;
+
+/// Interface for a Vulkan resource
+class VKResource {
+public:
+    explicit VKResource();
+    virtual ~VKResource() = 0;
```
lioncash: You may want to document why the destructor is pure virtual here, just to make it apparent, since it's a seldom-done thing.

ReinUsesLisp: It was a misinterpretation of mine. I thought that if the class had a pure method, the destructor had to also be pure. Changed it to use a virtual destructor.

PR #2123, opened 2019-02-14
core_timing: De-globalize core_timing facilities

lioncash:
Gets rid of the largest set of mutable global state within the core and makes it a proper class managed by a System instance. This is quite nice, given some of the data structures deglobalized would otherwise allocate on the heap at program start, making program startup ever so slightly slower than it needed to be. This also paves a way for eliminating usages of GetInstance() on the System class in several locations as well. This isn't done within this PR, since it's unrelated code, and would make this change quite large (which is already getting pretty big even without touching that code).

Note that no behavioral changes have been made, and this simply extracts the functionality into a class. This also has the benefit of making dependencies on the core timing functionality explicit within the relevant interfaces.

The bulk of the changes made are within core_timing.cpp/.h and its unit testing file. The rest of the changes are basic modifications to interfaces where necessary.

### DIFF
--- a/src/audio_core/audio_renderer.h
+++ b/src/audio_core/audio_renderer.h
@@ -14,6 +14,10 @@
 #include "common/swap.h"
 #include "core/hle/kernel/object.h"
 
+namespace Core::Timing {
+class CoreTiming;
+}
+
 namespace Kernel {
 class WritableEvent;
 }
@@ -208,7 +212,7 @@ static_assert(sizeof(UpdateDataHeader) == 0x40, "UpdateDataHeader has wrong size
 
 class AudioRenderer {
 public:
-    AudioRenderer(AudioRendererParameter params,
+    AudioRenderer(Core::Timing::CoreTiming& core_timing, AudioRendererParameter params,
                   Kernel::SharedPtr<Kernel::WritableEvent> buffer_event);
     ~AudioRenderer();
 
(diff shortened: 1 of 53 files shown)


src/audio_core/audio_renderer.h
```diff
@@ -208,7 +212,7 @@ static_assert(sizeof(UpdateDataHeader) == 0x40, "UpdateDataHeader has wrong size
 
 class AudioRenderer {
 public:
-    AudioRenderer(AudioRendererParameter params,
+    AudioRenderer(Core::System& system, AudioRendererParameter params,
```
jroweboy: nitpick: Any particular reason to pass a system in here instead of CoreTiming if all System is used for is to get the CoreTiming instance?

src/core/core_timing.cpp
```diff
@@ -8,94 +8,42 @@
...
 
-void UnregisterAllEvents() {
-    ASSERT_MSG(event_queue.empty(), "Cannot unregister events with events pending");
-    event_types.clear();
-}
+CoreTiming::CoreTiming() = default;
+CoreTiming::~CoreTiming() = default;
 
-void Init() {
+void CoreTiming::Initialize() {
```
jroweboy: nitpick: reading whats done in init and shutdown, it seems like it'll be fine to move this into the constructor and the destructor respectively.

lioncash: It's not. Because the system object is currently a global, so this needs to be called to reset and tear-down state.

jroweboy: bonus: this'll get rid of scopedinit in the coretiming test cases

PR #2129, opened 2019-02-16
Correct CNTPCT from using CPU Cycles to using Clock Cycles

FernandoS27:
CNTPCT is the register in ARM64 that takes core of saving the CPU's timer. Currently we were adding ticks measured in CPU Cycles to the CNTPCT, sadly the clock runs at a different frequency marked by CNTFRQ. This PR takes care of correcting that and transforming such ticks into appropriate Clock Cycles.

### DIFF
--- a/src/common/uint128.cpp
+++ b/src/common/uint128.cpp
@@ -1,0 +1,18 @@
+
+namespace Common {
+
+std::pair<u64, u64> udiv128(u128 dividend, u64 divisor) {
+    u64 remainder = dividend[0] % divisor;
+    u64 accum = dividend[0] / divisor;
+    if (dividend[1] == 0)
+        return {accum, remainder};
+    // We ignore dividend[1] / divisor as that overflows
+    u64 first_segment = (dividend[1] % divisor) << 32;
+    accum += (first_segment / divisor) << 32;
+    u64 second_segment = (first_segment % divisor) << 32;
+    accum += (second_segment / divisor);
+    remainder += second_segment % divisor;
+    return {accum, remainder};
+}
+
+} // namespace Common
--- a/src/common/uint128.cpp
+++ b/src/common/uint128.cpp
@@ -3,31 +3,32 @@
 
 #pragma intrinsic(_umul128)
 #endif
+#include <cstring>
 #include "common/uint128.h"
 
 namespace Common {
+
 u128 Multiply64Into128(u64 a, u64 b) {
-#ifdef _MSC_VER
     u128 result;
+#ifdef _MSC_VER
     result[0] = _umul128(a, b, &result[1]);
 #else
     unsigned __int128 tmp = a;
     tmp *= b;
-    u128 result;
     std::memcpy(&result, &tmp, sizeof(u128));
 #endif
     return result;
 }
 
-std::pair<u64, u64> Divide128On64(u128 dividend, u64 divisor) {
+std::pair<u64, u64> Divide128On32(u128 dividend, u32 divisor) {
     u64 remainder = dividend[0] % divisor;
     u64 accum = dividend[0] / divisor;
     if (dividend[1] == 0)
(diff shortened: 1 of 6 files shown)


src/common/uint128.h
```diff
@@ -0,0 +1,13 @@
+#include <array>
```
lioncash: `<array>`, `<cstdint>`, and `<cstring>` aren't necessary here.

src/core/core_timing_util.cpp
```diff
@@ -60,4 +61,10 @@ s64 nsToCycles(u64 ns) {
     return (BASE_CLOCK_RATE * static_cast<s64>(ns)) / 1000000000;
 }
 
+u64 CpuCyclesToClockCycles(u64 ticks) {
+    u128 temporal = Common::Multiply64Into128(ticks, CNTFREQ);
```
lioncash:
```suggestion
    const u128 temporal = Common::Multiply64Into128(ticks, CNTFREQ);
```

src/common/uint128.cpp
```diff
@@ -0,0 +1,40 @@
+#ifdef _MSC_VER
+#include <intrin.h>
+
+#pragma intrinsic(_umul128)
+#endif
+#include "common/uint128.h"
+
```
lioncash: Both of the functions in this file should have unit tests.

src/common/uint128.cpp
```diff
@@ -0,0 +1,40 @@
...
+return result;
+}
+
+std::pair<u64, u64> Divide128On64(u128 dividend, u64 divisor) {
+    u64 remainder = dividend[0] % divisor;
+    u64 accum = dividend[0] / divisor;
+    if (dividend[1] == 0)
+        return {accum, remainder};
+    // We ignore dividend[1] / divisor as that overflows
+    u64 first_segment = (dividend[1] % divisor) << 32;
```
lioncash:
```suggestion
    const u64 first_segment = (dividend[1] % divisor) << 32;
```

src/common/uint128.cpp
```diff
@@ -0,0 +1,40 @@
...
+    accum += (first_segment / divisor) << 32;
+    u64 second_segment = (first_segment % divisor) << 32;
```
lioncash:
```suggestion
    const u64 second_segment = (first_segment % divisor) << 32;
```

src/common/uint128.h
```diff
@@ -0,0 +1,13 @@
...
+#include <cstdint>
+#include <utility>
+#include <cstring>
+#include "common/common_types.h"
+
+namespace Common {
+
```
lioncash: Missing documentation comments for both functions.

src/core/core_timing_util.cpp
```diff
@@ -60,4 +61,10 @@ s64 nsToCycles(u64 ns) {
...
+    std::pair<u64, u64> result = Common::Divide128On64(temporal, BASE_CLOCK_RATE);
```
lioncash:
```suggestion
    return Common::Divide128On64(temporal, BASE_CLOCK_RATE).first;
```

src/core/arm/dynarmic/arm_dynarmic.cpp
```diff
@@ -151,7 +152,7 @@ std::unique_ptr<Dynarmic::A64::Jit> ARM_Dynarmic::MakeJit() const {
     config.tpidr_el0 = &cb->tpidr_el0;
     config.dczid_el0 = 4;
     config.ctr_el0 = 0x8444c004;
-    config.cntfrq_el0 = 19200000; // Value from fusee.
+    config.cntfrq_el0 = Timing::CNTFREQ; // Value from fusee.
```
lioncash: The comment here can likely be removed, since it was relocated to where the constant is defined

PR #2135, opened 2019-02-17
Implemented a lil edge case on RecreateSurface.

FernandoS27:
This implementation is not a hack but fixes a subproblem of a wider problem in the rasterizer cache. This should be merged into canary in liu of a proper implementation.


Hexagon12:
Superseded by #2601.

PR #2138, opened 2019-02-19
vk_memory_manager: Implement memory manager

ReinUsesLisp:
This handles the memory allocations for a device. It allocates chunks of Vulkan memory objects and then sub-allocates them. The data structures and methods are not optimized and a lot of stuff could be done for better memory management, but this is a rabbit hole that's not worth exploring at the moment.

Memory commitments are handled with `std::unique_ptr` and they will be automatically released from the resource pool on destruction.

This does **not** use [VulkanMemoryAllocator](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator). The reason behind that is that we may want to do custom memory management that may or may not be supported by VMA.


src/video_core/renderer_vulkan/vk_memory_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Returns whether this allocation is compatible with the arguments.
+    bool IsCompatible(vk::MemoryPropertyFlags wanted_properties, u32 type_mask) const {
+        return (wanted_properties & properties) != vk::MemoryPropertyFlagBits(0) &&
+               (type_mask & shifted_type) != 0;
+    }
+
+private:
+    static constexpr u32 ShiftType(u32 type) {
+        return 1 << type;
```
lioncash: This should likely be `1U` to avoid the `1 << 31` case with signed values.

src/video_core/renderer_vulkan/vk_memory_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /// A memory allocator, it may return a free region between "start" and "end" with the solicited
+    /// requeriments.
+    inline std::optional<u64> TryFindFreeSection(u64 start, u64 end, u64 size,
```
lioncash: inline isn't necessary here (funcs defined in a class declaration are implicitly inline).

src/video_core/renderer_vulkan/vk_memory_manager.h
```diff
@@ -0,0 +1,91 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
+#include <utility>
+#include <vector>
+#include "common/assert.h"
```
lioncash: This should go in the cpp file.

src/video_core/renderer_vulkan/vk_memory_manager.h
```diff
@@ -0,0 +1,91 @@
+// Copyright 2018 yuzu Emulator Project
```
lioncash: 2019

src/video_core/renderer_vulkan/vk_memory_manager.cpp
```diff
@@ -0,0 +1,254 @@
+// Copyright 2018 yuzu Emulator Project
```
lioncash: 2019

Ghostotakoi:
So we will have vulkan this year?? Awesome!!

PR #2141, opened 2019-02-19
speed up memory page mapping

Annomatg:
another try

use std:fill for nullifying
separate filling of arrays to speed up memory caching

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -71,15 +71,20 @@ static void MapPages(PageTable& page_table, VAddr base, u64 size, u8* memory, Pa
                                  FlushMode::FlushAndInvalidate);
 
     VAddr end = base + size;
-    while (base != end) {
-        ASSERT_MSG(base < page_table.pointers.size(), "out of range mapping at {:016X}", base);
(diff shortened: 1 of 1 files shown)


bunnei:
@Annomatg just curious - did you profile this? If yes - can you give details on how much faster this is or overall how much impact it has? (Otherwise, I'm sure the changes are fine since this has been in canary for days)

Annomatg:
Yes, I profiled it. But this is mostly important for the startup time. It's more than three times faster to clear the whole mapping (128 MB).
I'll check my profiling data and give you a more detailed answer this evening.

Annomatg:
In the startup phase until the first screen (controllers info) appears. The Memory::MapPages total samplecount was reduced from 4,6% to 1,06%.
From Main Menu into the Game:
from 1.03% to 0,35%

so round about 3-4 times faster, when used.

lioncash:
Just a request, but can you put the profiling numbers from your comment into the commit description as well? It seems useful to have those numbers in the commit history in case anyone ever looks up why the code is laid out the way it is (much better than having it on a PR page to get lost in time). Looks fine otherwise 👌

bunnei:
FYI, I added the perf numbers to the commit message on merge

PR #2143, opened 2019-02-21
gl_rasterizer_cache: Create texture views for array discrepancies

ReinUsesLisp:
When a texture is sampled in a shader with a different array mode than the cached state, create a texture view and bind that to the shader instead.


bunnei:
Supersedes #2135

FernandoS27:
@bunnei @ReinUsesLisp Tested it and does not fix the issue.

ReaperOfSouls1909:
i Tested this on my Hd4600 and I didn't have the mipmap corruption issue

alinach120602:
Issue seems to be fixed even on Nvidia (Linux proprietary 396.54.09)

yay

FernandoS27:
It still happens in Canary 1913. Have in mind that it happens after a while of running around, not immediately.

ReaperOfSouls1909:
https://cdn.discordapp.com/attachments/472575665334648833/550430156821364772/unknown.png the issue still happens after a bit

PR #2144, opened 2019-02-21
service/vi: Convert Display and Layer structs into classes

lioncash:
Continuing off of #2142, this converts the Display and Layer structs into classes and puts everything behind interface functions to cleanly isolate behavior for upcoming changes to these structs. #2142 was intended to be the last reorganization PR before functional changes start occurring, but I figured I'd separate this into its own PR beforehand to cut down on the amount of "churn" in later PRs.

This conversion into classes is because future changes will start placing invariants on both classes that need to hold, and so the interface prevents accidentally messing around with those. It also makes it a little more consistent, so we don't have cases where member variables are being directly accessed, and cases where they're not in the future.

### DIFF
--- a/src/core/hle/service/nvflinger/nvflinger.cpp
+++ b/src/core/hle/service/nvflinger/nvflinger.cpp
@@ -28,9 +28,13 @@ namespace Service::NVFlinger {
 constexpr std::size_t SCREEN_REFRESH_RATE = 60;
 constexpr u64 frame_ticks = static_cast<u64>(Core::Timing::BASE_CLOCK_RATE / SCREEN_REFRESH_RATE);
 
-NVFlinger::NVFlinger(Core::Timing::CoreTiming& core_timing)
-    : displays{{0, "Default"}, {1, "External"}, {2, "Edid"}, {3, "Internal"}, {4, "Null"}},
-      core_timing{core_timing} {
+NVFlinger::NVFlinger(Core::Timing::CoreTiming& core_timing) : core_timing{core_timing} {
+    displays.emplace_back(0, "Default");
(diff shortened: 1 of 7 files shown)


src/core/hle/service/nvflinger/nvflinger.cpp
```diff
@@ -28,9 +28,13 @@ namespace Service::NVFlinger {
 constexpr std::size_t SCREEN_REFRESH_RATE = 60;
 constexpr u64 frame_ticks = static_cast<u64>(Core::Timing::BASE_CLOCK_RATE / SCREEN_REFRESH_RATE);
 
-NVFlinger::NVFlinger(Core::Timing::CoreTiming& core_timing)
-    : displays{{0, "Default"}, {1, "External"}, {2, "Edid"}, {3, "Internal"}, {4, "Null"}},
-      core_timing{core_timing} {
+NVFlinger::NVFlinger(Core::Timing::CoreTiming& core_timing) : core_timing{core_timing} {
+    displays.emplace_back(0, "Default");
```
lioncash: In case anyone is wondering, this is necessary, since we delete the copy ctor and copy assignment operator of Display. `std::vector`'s `std::initializer_list` constructor always result in a copy (and since we delete the copy ctor and copy assignment operator, we can't use that). Thus, the emplacing instead.

VPeruS: Why not foreach all emplace_back?

lioncash: That seems overly complicated for something that is unlikely to gain a large amount elements.

PR #2147, opened 2019-02-22
shader_ir: Remove "extras" from the MetaTexture

ReinUsesLisp:
Continuation of #2068. The `extras` field relies on GLSL specific parameter orders, this PR removes it with meaningful fields (such as bias and lod). It also splits memory and texture instructions decoding in different files.

Same as #2068, this touches messy code so I'd prefer if it's tested in canary before getting merged.


bunnei:
@ReinUsesLisp the clang errors look real here - can you take a look?

PR #2152, opened 2019-02-24
 vk_stream_buffer: Implement a stream buffer

ReinUsesLisp:
This manages two kinds of streaming buffers: one for unified memory models and one for dedicated GPUs. The first one skips the copy from the staging buffer to the real buffer, since it creates an unified buffer.

This implementation waits for all fences to finish their operation before "invalidating". This is suboptimal since it should allocate another buffer or start searching from the beginning. There is room for improvement here.

This could also handle AMD's "pinned" memory (a heap with 256 MiB) that seems to be designed for buffer streaming.

### DIFF
--- a/src/video_core/CMakeLists.txt
+++ b/src/video_core/CMakeLists.txt
@@ -111,7 +111,9 @@ if (ENABLE_VULKAN)
         renderer_vulkan/vk_resource_manager.cpp
         renderer_vulkan/vk_resource_manager.h
         renderer_vulkan/vk_scheduler.cpp
-        renderer_vulkan/vk_scheduler.h)
+        renderer_vulkan/vk_scheduler.h
+        renderer_vulkan/vk_stream_buffer.cpp
+        renderer_vulkan/vk_stream_buffer.h)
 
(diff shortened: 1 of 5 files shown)


hrydgard:
For data that is highly cachable (like UBO) or read only once by the GPU, even on discrete/dedicated GPUs there's not much of a win to be had by doing an extra copy to DEVICE_LOCAL memory - it's just fine to write to HOST_VISIBLE|HOST_COHERENT memory from the CPU and read from the same memory on the GPU. (However, avoid reading from such memory using the CPU at all costs).

ReinUsesLisp:
Thanks! I've removed the copy path, using `HOST_VISIBLE | HOST_COHERENT` gave the same results like you've suggested.

hrydgard:
Good, I think simplicity is worth a lot when you bring up a Vulkan backend :)  For a few things like SSBOs it might be worth experimenting with in the future I suppose.

ReinUsesLisp:
On our OpenGL backend we use SSBOs just for global memory operations. We might use them in the future when we run out of UBOs, but that would go through the stream buffer since it's a read only operation.

PR #2154, opened 2019-02-24
Port citra-emu/citra#4647: "citra_qt/main: make SPEED_LIMIT_STEP static constexpr"

FearlessTobi:
See citra-emu/citra#4647 for all the reasoning and discussion on this.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -561,7 +561,10 @@ void GMainWindow::InitializeHotkeys() {
                 Settings::values.use_frame_limit = !Settings::values.use_frame_limit;
                 UpdateStatusBar();
             });
-    constexpr u16 SPEED_LIMIT_STEP = 5;
+    // TODO: Remove this comment/static whenever the next major release of
+    // MSVC occurs and we make it a requirement (see:
+    // https://developercommunity.visualstudio.com/content/problem/93922/constexprs-are-trying-to-be-captured-in-lambda-fun.html)
+    static constexpr u16 SPEED_LIMIT_STEP = 5;
     connect(hotkey_registry.GetHotkey("Main Window", "Increase Speed Limit", this),
             &QShortcut::activated, this, [&] {
                 if (Settings::values.frame_limit < 9999 - SPEED_LIMIT_STEP) {


src/yuzu/main.cpp
```diff
@@ -561,7 +561,9 @@ void GMainWindow::InitializeHotkeys() {
                 Settings::values.use_frame_limit = !Settings::values.use_frame_limit;
                 UpdateStatusBar();
             });
-    constexpr u16 SPEED_LIMIT_STEP = 5;
+    // We use "static" here in order to avoid capturing by lambda due to a MSVC bug, which makes the
```
lioncash: This isn't an MSVC bug. The reason the garbage value occurs is because the lambda variables are capturing by reference not by value.

When the function ends, the lifetime of the constant ends, and it's discarded from the stack, giving it a garbage value. MSVC is performing allowed behavior here.

FearlessTobi: So should I just capture by value and remove the "static" and the comment?

wwylele: @lioncash constexpr variable is supposed not to be implicitly captured if the lambda doesn't have ODR-use. See https://en.cppreference.com/w/cpp/language/lambda

> If a capture list has a capture-default and does not explicitly capture the enclosing object (as this or *this) or an automatic variable, it captures it implicitly if
>  - the body of the lambda odr-uses the variable or the this pointer
>  - or the variable or the this pointer is named in a potentially-evaluated expression within an expression (including when the implicit this-> is added before a use of non-static class member). For this purpose, the operand of typeid is always considered potentially-evaluated. Entities might be implicitly captured even if they are only named within a discarded statement.

The second rule is not related here if I understand it correctly ). For the first rule: https://en.cppreference.com/w/cpp/language/definition#ODR-use

> Informally, an object is odr-used if its value is read (**unless it is a compile time constant**) or written, its address is taken, or a reference is bound to it; a reference is odr-used if it is used and its referent is not known at compile time; and a function is odr-used if a function call to it is made or its address is taken. If an object, a reference or a function is odr-used, its definition must exist somewhere in the program; a violation of that is usually a link-time error.
>Formally, 1) a variable x in a potentially-evaluated expression ex is odr-used unless both of the following are true:
**applying lvalue-to-rvalue conversion to x yields a constant expression that doesn't invoke non-trivial functions**
either x is not an object (that is, x is a reference) or, if x is an object, it is one of the potential results of a larger expression e, where that larger expression is either a discarded-value expression or has the lvalue-to-rvalue conversion applied to it

Here is a simple example showing that GCC correctly replaces constant, while MSVC trying to capture:
https://godbolt.org/z/qKf5we

wwylele: But yes, I agree that slapping `[&]` on lambdas is not a good practice in general. Might still want to change that.

wwylele: Sorry for arguing on small things. I don't want to get another backport to citra which leads to yet another modification there that is going to port here again.

lioncash: Thanks for the info. It'd probably be better to place something like:

```cpp
// TODO: Remove this comment/static whenever the next major release of
// MSVC occurs and we make it a requirement (see: <link to bug>)
```

so we don't have a comment perpetually laying around about old compiler behavior in the future.

FearlessTobi: Addressed.

PR #2160, opened 2019-02-25
audio_core: Resolve compilation warnings

lioncash:
Resolves a few trivial compilation warnings related to truncation, shadowing, and initialization.

### DIFF
--- a/src/audio_core/cubeb_sink.cpp
+++ b/src/audio_core/cubeb_sink.cpp
@@ -98,7 +98,7 @@ private:
     u32 num_channels{};
 
     Common::RingBuffer<s16, 0x10000> queue;
-    std::array<s16, 2> last_frame;
+    std::array<s16, 2> last_frame{};
     std::atomic<bool> should_flush{};
     TimeStretcher time_stretch;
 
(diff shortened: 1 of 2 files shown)


src/audio_core/cubeb_sink.cpp
```diff
@@ -98,7 +98,7 @@ class CubebSinkStream final : public SinkStream {
     u32 num_channels{};
 
     Common::RingBuffer<s16, 0x10000> queue;
-    std::array<s16, 2> last_frame;
+    std::array<s16, 2> last_frame{};
```
bunnei: What's the reason for this change? Isn't std::array already initialized to 0's?

lioncash: std::array by default doesn't initialize primitives. Since `std::array` is defined as an aggregate type, it doesn't have a constructor. The only way elements of a std::array will be initialized is if they have their own constructor that performs some form of initialization. Aside from that, it's possible to request zero-initialization with `{}`.

bunnei: Got it - thanks for the explanation.

PR #2165, opened 2019-02-26
 gl_rasterizer: Unbind textures but don't apply the gl_state 

ReinUsesLisp:
Applying the state to update a framebuffer will also update the state for the textures. Knowing this we can unbind the textures at the end of the draw call without applying it. Also fix a `gl_state` bug were the first index offset was being ignored.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -739,22 +739,10 @@ void RasterizerOpenGL::DrawArrays() {
     state.Apply();
 
     res_cache.SignalPreDrawCall();
-
-    // Execute draw call
     params.DispatchDraw();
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -750,7 +750,6 @@ void RasterizerOpenGL::DrawArrays() {
     for (auto& texture_unit : state.texture_units) {
```
degasus: Just drop the whole for loop, and the comment above it is wrong. There is only one reason why we unbind them: glDeleteTexture must update gl_state.

degasus: By the way, also drop the scissor.enabled = false above.

PR #2170, opened 2019-02-27
core/frontend/emu_window: Make ClipToTouchScreen a const member function

lioncash:
This member function doesn't modify instance state, so it can have the const specifier applied to it.


lioncash:
Failure isn't related to the code itself (it's a timeout).

PR #2175, opened 2019-02-27
Analog Stick deadzone fix

Shedino:
Hi, I know that this is not so high priority and it is just a kind of "brutal" fix, but at least helped me using my controller on some games. It is currently set with a 2% deadzone that seemed reasonable on my hardware and the impact on stick precision should be minimal. There is no scaling logic on remaining range and maybe it can be addressed on a lower level. I'm open to comments/suggestion/etc. Just trying to be useful :)


ExsertWizard:
Would be great to add a slider to set the deadzone percentage right below the binding of analog sticks in the input menu. Not all controllers have the same amount of deadzone so it would be great to add this for everyone so that they can have their preference.

Shedino:
OK. I think I can have a look at it if you think it is necessary

FearlessTobi:
I'm not sure if putting this into the hid service class and core is a good idea. Maybe consider putting this into the input system itself?

Shedino:
> I'm not sure if putting this into the hid service class and core is a good idea. Maybe consider putting this into the input system itself?

do you mean input_common (sdl.cpp)?

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -286,6 +287,18 @@ void Controller_NPad::RequestPadStateUpdate(u32 npad_id) {
     lstick_entry.y = static_cast<s32>(stick_l_y_f * HID_JOYSTICK_MAX);
     rstick_entry.x = static_cast<s32>(stick_r_x_f * HID_JOYSTICK_MAX);
     rstick_entry.y = static_cast<s32>(stick_r_y_f * HID_JOYSTICK_MAX);
+    HandleAnalogDeadZone(&lstick_entry);
+    HandleAnalogDeadZone(&rstick_entry);
+}
+
+void Controller_NPad::HandleAnalogDeadZone(AnalogPosition* analogStick) {
```
DarkLordZach: Use references

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -27,6 +27,7 @@ constexpr u32 MAX_NPAD_ID = 7;
 constexpr std::array<u32, 10> npad_id_list{
     0, 1, 2, 3, 4, 5, 6, 7, NPAD_HANDHELD, NPAD_UNKNOWN,
 };
+constexpr s32 HID_JOYSTICK_DZONE = HID_JOYSTICK_MAX * 0.02;
```
DarkLordZach: HID_JOYSTICK_DEADZONE

src/core/hle/service/hid/controllers/npad.h
```diff
@@ -319,6 +319,7 @@ class Controller_NPad final : public ControllerBase {
     bool IsControllerSupported(NPadControllerType controller) const;
     NPadControllerType DecideBestController(NPadControllerType priority) const;
     void RequestPadStateUpdate(u32 npad_id);
+    void HandleAnalogDeadZone(AnalogPosition *analogStick);
```
DarkLordZach: npad isn't the only HID device (on console) to use joysticks; its simply the only one we have implemented right now.
find a better place for this code.

Shedino:
Thanks for the comments. I feel like it would be cleaner and more practical to add this on the low level joystick code if you agree. I'll perform some changes and will try to update this PR (or open a new one).

Shedino:
Hi all, I think this can be closed considering PR 2187. My changes for the deadzone were obviously similar...
Don't know if you still need some GUI changes. Just let me know if you feel that I can help addressing some issues.

jroweboy:
@Shedino you opened the PR, so you are also free to close the PR if you feel it is sufficiently addressed by the other one. Additionally the other PR doesn't add a UI for the setting, so theres some follow up work that can be done if you would like to look into that.

Shedino:
OK. I'll check the merged PR and possibly come back with some GUI updates.

PR #2185, opened 2019-03-02
Port citra-emu/citra#4630: "Memory: don't lock hle mutex in memory read/write"

FearlessTobi:
See citra-emu/citra#4630 for more details.

**Original description**:
The comment already invalidates itself: neither MMIO nor rasterizer cache belongs to HLE kernel state. This mutex has a too large scope if MMIO or cache is included, which is prone to dead lock when multiple thread acquires these resource at the same time. If necessary, each MMIO component or rasterizer should have their own lock.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -171,9 +171,6 @@ T Read(const VAddr vaddr) {
         return value;
     }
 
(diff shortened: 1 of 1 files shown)


bunnei:
This was part of #2055, but I'll just merge this one and rebase. Thanks!

PR #2190, opened 2019-03-02
core: Remove the global telemetry accessor function

lioncash:
In many cases, the system object reference is nearby, so we can just use that to access the telemetry instance. In the lingering case within compatdb, we can just use the long-form of accessing it.

With all of the usages converted over, we can remove the function that relies on the global variable.


lioncash:
Error is related to a timeout, not this code in particular.

PR #2191, opened 2019-03-02
maxwell_to_vk: Initial implementation

ReinUsesLisp:
This has the same purpose as OpenGL's `maxwell_to_gl`, but its definition is in a `cpp` file instead of the header. It's not intended to be a complete implementation, since I like to add the entries as I find them to avoid obscure bugs.

### DIFF
--- a/src/video_core/renderer_vulkan/maxwell_to_vk.cpp
+++ b/src/video_core/renderer_vulkan/maxwell_to_vk.cpp
@@ -1,0 +1,483 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
+#include "common/common_types.h"
+#include "common/logging/log.h"
+#include "video_core/engines/maxwell_3d.h"
+#include "video_core/renderer_vulkan/declarations.h"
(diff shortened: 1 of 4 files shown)


Ghostotakoi:
The Vulkan is Coming

src/video_core/renderer_vulkan/maxwell_to_vk.cpp
```diff
@@ -0,0 +1,483 @@
...
+    {vk::Format::eD32Sfloat, ComponentType::Float, true}, // Z32F
+    {vk::Format::eD16Unorm, ComponentType::UNorm, true},  // Z16
+
+    // DepthStencil formats
+    {vk::Format::eD24UnormS8Uint, ComponentType::UNorm, true}, // Z24S8
+    {vk::Format::eD24UnormS8Uint, ComponentType::UNorm, true}, // S8Z24 (emulated)
+    {vk::Format::eUndefined, ComponentType::Invalid, false},   // Z32FS8
+}};
+
+constexpr bool IsZetaFormat(PixelFormat pixel_format) {
```
lioncash: This can be made static to make it internally linked.

niko1point0:
Thank you for the hard work. I’m really excited to see Vulkan support

Issue #2157, opened 2019-02-25
I am not able to return to the game super mario odeyssey after the ingame web applet comes on. I control the app with a keyboard key that I never assigned and the directional keys but it does nothing when I press the key it has on the bottom of the screen to exit the page. I end up right clicking with my mouse and restoring the page but that only takes me back to the first web applet menu. Please Help.


MultiCuriousKitty:

ExsertWizard:
...you do realize that you just made the longest title ever and that there's a description text box for all of that...

Zeroksas:
Web applet is not functioning correctly for some users from the beginning. In order to bypass this bug you have to enter web applet in windowed mode and the exit it by mouse clicking exit option in top menu.

MultiCuriousKitty:
Oops sorry ExseertWizard I am new to Github. Thank you Zeroksas it worked and I can now continue gaming.

PR #2196, opened 2019-03-04
web_browser: Add shortcut to Enter key to exit applet

DarkLordZach:
Addresses issues where a user in fullscreen could not exit some web applets without leaving fullscreen. (Specifically #2157)

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -339,6 +339,11 @@ void GMainWindow::WebBrowserOpenPage(std::string_view filename, std::string_view
                 .arg(QString::fromStdString(std::to_string(key_code))));
     };
 
+    QMessageBox::information(
+        this, tr("Exit"),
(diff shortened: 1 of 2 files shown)


src/yuzu/main.cpp
```diff
@@ -339,6 +339,11 @@ void GMainWindow::WebBrowserOpenPage(std::string_view filename, std::string_view
                 .arg(QString::fromStdString(std::to_string(key_code))));
     };
 
+    QMessageBox::information(
```
FearlessTobi: Am I stupid or is this actually displayed every time the user goes into the WebApplet?

DarkLordZach: Personally, given the infrequency of use of the web applet I think this is best -- its not that pesky and people are likely to forget after a while. If you'd like, I have no problem putting it as a callout (shown only once per user).

PR #2197, opened 2019-03-05
core/hle/ipc: Remove unnecessary includes

lioncash:
Removes a few inclusion dependencies from the headers or replaces existing ones with ones that don't indirectly include the required headers.

This allows removing an inclusion of core/memory.h, meaning that if the memory header is ever changed in the future (it likely will), it won't result in rebuilding the entirety of the HLE services (as the IPC headers are used quite ubiquitously throughout the HLE service implementations).

### DIFF
--- a/src/core/hle/ipc.h
+++ b/src/core/hle/ipc.h
@@ -4,10 +4,10 @@
 
 #pragma once
 
+#include "common/bit_field.h"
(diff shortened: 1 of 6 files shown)


lioncash:
Failure is due to a timeout and not the code itself.

PR #2198, opened 2019-03-05
{kernel/thread, video_core/surface}: Remove obsolete TODOs

lioncash:
The TODO in Create() is a TODO carried over from Citra that doesn't apply here. Hexagon also told me that the one in PixelFormatFromRenderTargetFormat() doesn't apply anymore either, so we can get rid of these.


lioncash:
Failure is due to a timeout and not the changes themselves.

PR #2199, opened 2019-03-05
kernel/address_arbiter: Convert the address arbiter into a class

lioncash:
This will be necessary to do further deglobalizing work when it comes to hoisting the memory subsystem into its own class later on. This also allows getting rid of some usages of the global system accessor, moving them further up the chain into the SVC calls themselves, making them much easier to replace in the future, when globals get addressed there.


FernandoS27:
@lioncash It's not compiling in GCC and Clang.

PR #2207, opened 2019-03-07
service/audio/hwopus: Move decoder state to its own class

lioncash:
Moves the non-multistream specific state to its own class. This will be necessary to support the multistream variants of opus decoding, given the overall decoder state parameters differ from the regular non-multistreamed version of the decoder.


lioncash:
Failure is due to appveyor failing to upload. It's not related to any of the code changes.

PR #2209, opened 2019-03-07
video_core/gpu_thread: Silence a -Wreorder warning

lioncash:
Moves the data members to satisfy the order they're declared as in the constructor initializer list. Silences a -Wreorder warning.

This also removes an unimplemented function prototype as well, to prevent accidental usage of it.


lioncash:
Build failure is related to a timeout.

PR #2215, opened 2019-03-09
gl_rasterizer: Encapsulate sampler queries into methods

ReinUsesLisp:
Abstract sampler queries into getter methods to avoid repeating logic across APIs.
It also does some renames like `mip_filter` -> `mipmap_filter` because it was easy to mistype `min_filter`.

### DIFF
--- a/src/video_core/textures/texture.h
+++ b/src/video_core/textures/texture.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <array>
 #include "common/assert.h"
 #include "common/bit_field.h"
 #include "common/common_funcs.h"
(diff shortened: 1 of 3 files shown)


src/video_core/textures/texture.h
```diff
@@ -306,10 +307,40 @@ struct TSCEntry {
...
+        return static_cast<float>(max_lod_clamp) / 256.0f;
+    }
+
+    float GetLodBias() const {
+        // Sign extend the 13-bit value.
+        constexpr u32 mask = 1U << (13 - 1);
+        return static_cast<float>((mip_lod_bias ^ mask) - mask) / 256.0f;
+    }
+
+    std::array<float, 4> GetBorderColor() const {
```
lioncash: This can likely be flattened out into:

```cpp
std::array<float, 4> GetBorderColor() const {
    if (srgb_conversion) {
        return {srgb_border_color_r / 255.0, srgb_border_color_g / 255.0,
                srgb_border_color_b / 255.0, border_color[3]};
    }

    return border_color;
}
```

bunnei:
@ReinUsesLisp this seems to have broken PLG

image

PR #2216, opened 2019-03-09
gl_rasterizer: Use system instance passed from argument

ReinUsesLisp:
OpenGL's rasterizer constructor receives a system instance. Store that in the object and use that instead.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -102,8 +102,9 @@ struct FramebufferCacheKey {
 
 RasterizerOpenGL::RasterizerOpenGL(Core::Frontend::EmuWindow& window, Core::System& system,
                                    ScreenInfo& info)
-    : res_cache{*this}, shader_cache{*this, system}, global_cache{*this}, emu_window{window},
-      screen_info{info}, buffer_cache(*this, STREAM_BUFFER_SIZE) {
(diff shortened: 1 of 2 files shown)


lioncash:
Needs a rebase

bunnei:
Needs a rebase

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -102,8 +102,9 @@ struct FramebufferCacheKey {
 
 RasterizerOpenGL::RasterizerOpenGL(Core::Frontend::EmuWindow& window, Core::System& system,
                                    ScreenInfo& info)
-    : res_cache{*this}, shader_cache{*this, system}, global_cache{*this}, emu_window{window},
-      screen_info{info}, buffer_cache(*this, STREAM_BUFFER_SIZE) {
+    : res_cache{*this}, shader_cache{*this, system}, global_cache{*this},
+      emu_window{window}, system{system}, screen_info{info},
```
bunnei: Wouldn't it make more sense to capture the GPU interface, not System?

ReinUsesLisp: When `system` is passed to this object, the GPU object has not been created, capturing a nullptr.

bunnei: Fair enough. Would be great if we could fix this in the future, as it seems this is an issue in other places...

PR #2221, opened 2019-03-10
set_sys: Implement GetFirmwareVersion(2) for libnx hosversion

DarkLordZach:
This adds: 
- A synthesized/open-source version of system archive 0x09 (SystemVersion), which contains a file that describes the current system version. The variant I implemented will return the following information:
    - `5.1.0-0.0` on platform `YUZU` with long version `YuzuEmulated Firmware for NX 5.1.0-0.0`
    - If using yuzu instead of the official 'NX' platform, becomes an issue, it can easily be changed via string.
- An implementation of GetFirmwareVersion and GetFirmwareVersion2 based on this. On hardware, both read the file aformentioned into state and then copy it into the requested buffer when called (the only difference being that GetFirmwareVersion zeros out two of the fields) -- instead of using the real SystemVersion title if the user has it present, this will always use the synthesized version to ensure consistency and prevent weird bugs (i.e. if user has really old or really new firmware).

I decided to go with 5.1.0 as though yuzu does implement some HLE from 6.0.0+, the most accurate description of yuzu's HLE puts us around 5.1.0. If anyone thinks this is not correct, feel free to ask and I'll change it.

This should address the issue brought up by fincs in Discord, where new libnx homebrew would crash as libnx now uses GetFirmwareVersion* to detect system version.

Tested against homebrew created by Simon in Discord, which produces the following result:
gfv stats

### DIFF
--- a/src/core/file_sys/system_archive/system_version.cpp
+++ b/src/core/file_sys/system_archive/system_version.cpp
@@ -16,15 +16,19 @@ constexpr u8 VERSION_MAJOR = 5;
 constexpr u8 VERSION_MINOR = 1;
 constexpr u8 VERSION_MICRO = 0;
 
-constexpr u8 REVISION_MAJOR = 0;
+constexpr u8 REVISION_MAJOR = 3;
 constexpr u8 REVISION_MINOR = 0;
 
-constexpr char PLATFORM_STRING[] = "YUZU";
-constexpr char VERSION_HASH[] = "";
+constexpr char PLATFORM_STRING[] = "NX";
+constexpr char VERSION_HASH[] = "23f9df53e25709d756e0c76effcb2473bd3447dd";
 constexpr char DISPLAY_VERSION[] = "5.1.0";
-constexpr char DISPLAY_TITLE[] = "YuzuEmulated Firmware for NX 5.1.0-0.0";
+constexpr char DISPLAY_TITLE[] = "NintendoSDK Firmware for NX 5.1.0-3.0";
 
 } // namespace SystemVersionData
+
+std::string GetLongDisplayVersion() {
+    return SystemVersionData::DISPLAY_TITLE;
+}
 
 VirtualDir SystemVersion() {
     VirtualFile file = std::make_shared<VectorVfsFile>(std::vector<u8>(0x100), "file");
@@ -34,13 +38,13 @@ VirtualDir SystemVersion() {
     file->WriteObject(SystemVersionData::REVISION_MAJOR, 4);
     file->WriteObject(SystemVersionData::REVISION_MINOR, 5);
     file->WriteArray(SystemVersionData::PLATFORM_STRING,
-                     std::min<u64>(sizeof(SystemVersionData::PLATFORM_STRING), 0x20ull), 0x8);
+                     std::min<u64>(sizeof(SystemVersionData::PLATFORM_STRING), 0x20ULL), 0x8);
     file->WriteArray(SystemVersionData::VERSION_HASH,
-                     std::min<u64>(sizeof(SystemVersionData::VERSION_HASH), 0x40ull), 0x28);
+                     std::min<u64>(sizeof(SystemVersionData::VERSION_HASH), 0x40ULL), 0x28);
     file->WriteArray(SystemVersionData::DISPLAY_VERSION,
-                     std::min<u64>(sizeof(SystemVersionData::DISPLAY_VERSION), 0x18ull), 0x68);
+                     std::min<u64>(sizeof(SystemVersionData::DISPLAY_VERSION), 0x18ULL), 0x68);
     file->WriteArray(SystemVersionData::DISPLAY_TITLE,
-                     std::min<u64>(sizeof(SystemVersionData::DISPLAY_TITLE), 0x80ull), 0x80);
+                     std::min<u64>(sizeof(SystemVersionData::DISPLAY_TITLE), 0x80ULL), 0x80);
     return std::make_shared<VectorVfsDirectory>(std::vector<VirtualFile>{file},
                                                 std::vector<VirtualDir>{}, "data");
 }
--- a/src/core/file_sys/system_archive/system_version.cpp
+++ b/src/core/file_sys/system_archive/system_version.cpp
@@ -34,13 +34,13 @@ VirtualDir SystemVersion() {
     file->WriteObject(SystemVersionData::REVISION_MAJOR, 4);
     file->WriteObject(SystemVersionData::REVISION_MINOR, 5);
     file->WriteArray(SystemVersionData::PLATFORM_STRING,
-                     std::min(sizeof(SystemVersionData::PLATFORM_STRING), 0x20ull), 0x8);
+                     std::min<u64>(sizeof(SystemVersionData::PLATFORM_STRING), 0x20ull), 0x8);
     file->WriteArray(SystemVersionData::VERSION_HASH,
-                     std::min(sizeof(SystemVersionData::VERSION_HASH), 0x40ull), 0x28);
+                     std::min<u64>(sizeof(SystemVersionData::VERSION_HASH), 0x40ull), 0x28);
     file->WriteArray(SystemVersionData::DISPLAY_VERSION,
-                     std::min(sizeof(SystemVersionData::DISPLAY_VERSION), 0x18ull), 0x68);
+                     std::min<u64>(sizeof(SystemVersionData::DISPLAY_VERSION), 0x18ull), 0x68);
     file->WriteArray(SystemVersionData::DISPLAY_TITLE,
-                     std::min(sizeof(SystemVersionData::DISPLAY_TITLE), 0x80ull), 0x80);
+                     std::min<u64>(sizeof(SystemVersionData::DISPLAY_TITLE), 0x80ull), 0x80);
     return std::make_shared<VectorVfsDirectory>(std::vector<VirtualFile>{file},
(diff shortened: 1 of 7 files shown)


src/core/hle/service/set/set_sys.cpp
```diff
@@ -3,12 +3,71 @@
...
+
+    const auto early_exit_failure = [&ctx](const std::string& desc) {
+        LOG_ERROR(Service_SET, "General failure while attempting to resolve firmware version ({}).",
+                  desc.c_str());
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(ResultCode(-1));
+    };
+
+    if (archive == nullptr) {
+        early_exit_failure("The system version archive couldn't be synthesized.");
```
lioncash: This should likely return an error code of:

```cpp
// Yeah, the FS service typically reports this.
ResultCode(ErrorModule::FS, 3223);
```

to represent failing to mount the system version archive.

src/core/hle/service/set/set_sys.cpp
```diff
@@ -3,12 +3,71 @@
...
+
+    const auto ver_file = archive->GetFile("file");
+    if (ver_file == nullptr) {
+        early_exit_failure("The system version archive didn't contain the file 'file'.");
+        return;
+    }
+
+    auto data = ver_file->ReadAllBytes();
+    if (data.size() != 0x100) {
+        early_exit_failure("The system version file 'file' was not the correct size.");
```
lioncash: This should return:

```cpp
// FS also returns this for reads larger than it can handle
ResultCode(ErrorModule::FS, 3005);
```

To represent an invalid read size.

src/core/hle/service/set/set_sys.cpp
```diff
@@ -3,12 +3,71 @@
...
+    };
+
+    if (archive == nullptr) {
+        early_exit_failure("The system version archive couldn't be synthesized.");
+        return;
+    }
+
+    const auto ver_file = archive->GetFile("file");
+    if (ver_file == nullptr) {
+        early_exit_failure("The system version archive didn't contain the file 'file'.");
```
lioncash: This should return:

```cpp
ResultCode(ErrorModule::FS, 6001);
```

to indicate an invalid name (which the FS service returns for non-existent names)

src/core/hle/service/set/set_sys.cpp
```diff
@@ -3,12 +3,71 @@
...
+        early_exit_failure("The system version file 'file' was not the correct size.");
+        return;
+    }
+
+    // If the command is GetFirmwareVersion (as opposed to GetFirmwareVersion2), hardware will zero
+    // out the REVISION_MAJOR and REVISION_MICRO fields, which are u8s at offsets 0x4 and 0x5,
+    // respectively. This if statement will only execute when the true intended command is
+    // GetFirmwareVersion, and allows removing duplicate code for the implementation of
+    // GetFirmwareVersion2.
+    if (ctx.GetCommand() == 3) {
```
lioncash: It would likely be better to make the main function itself `GetFirmwareVersionImpl`, then implement both of the IPC functions in terms of that. This way you can just have an enum like:

```cpp
enum class GetFirmwareVersionType {
  Version1,
  Version2,
};
```

then make it a parameter like:

```cpp
void GetFilesystemVersionImpl(Kernel::HLERequestContext& ctx, GetFilesystemVersionType type);
```

and check against that. This should allow doing this without depending on the command number.

src/core/hle/service/set/set_sys.cpp
```diff
@@ -3,12 +3,71 @@
...
+    }
+
+    ctx.WriteBuffer(data);
+
+    IPC::ResponseBuilder rb{ctx, 2};
+    rb.Push(RESULT_SUCCESS);
+}
+
+void SET_SYS::GetFirmwareVersion2(Kernel::HLERequestContext& ctx) {
+    LOG_WARNING(Service_SET, "called - delegating to GetFirmwareVersion");
```
lioncash: This should probably be a debug log

src/core/hle/service/set/set_sys.cpp
```diff
@@ -3,12 +3,71 @@
...
+        return;
+    }
+
+    // If the command is GetFirmwareVersion (as opposed to GetFirmwareVersion2), hardware will zero
+    // out the REVISION_MAJOR and REVISION_MICRO fields, which are u8s at offsets 0x4 and 0x5,
+    // respectively. This if statement will only execute when the true intended command is
+    // GetFirmwareVersion, and allows removing duplicate code for the implementation of
+    // GetFirmwareVersion2.
+    if (ctx.GetCommand() == 3) {
+        data[0x4] = 0x0;
```
lioncash: Based off my previous RE work on the system service, this will only zero out the minor revision field, but not the major one.

src/core/hle/service/set/set_sys.cpp
```diff
@@ -11,61 +12,80 @@
 
 namespace Service::Set {
 
-void SET_SYS::GetFirmwareVersion(Kernel::HLERequestContext& ctx) {
-    LOG_DEBUG(Service_SET, "called");
+constexpr u64 SYSTEM_VERSION_FILE_MINOR_REVISION_OFFSET = 0x05;
+
+constexpr ResultCode ERROR_FAILED_MOUNT_ARCHIVE(ErrorModule::FS, 3223);
```
lioncash: These should be placed with the rest of the FS error codes in `file_sys/errors.h`

src/core/hle/service/set/set_sys.cpp
```diff
@@ -11,61 +12,80 @@
...
+constexpr ResultCode ERROR_READ_TOO_LARGE(ErrorModule::FS, 3005);
```
lioncash: `ERROR_OUT_OF_BOUNDS` or `ERROR_OUT_OF_RANGE` is a better name here, since this error code is used for both reads and writes in the FS service.

src/core/hle/service/set/set_sys.cpp
```diff
@@ -11,61 +12,80 @@
...
+constexpr ResultCode ERROR_INVALID_NAME(ErrorModule::FS, 6001);
```
lioncash: This should be named `ERROR_INVALID_ARGUMENT`, because it's a general filesystem error code used to represent any kind of invalid argument, not just names.

src/core/hle/service/set/set_sys.cpp
```diff
@@ -11,61 +12,80 @@
...
+enum class GetFirmwareVersionType {
+    Version1,
+    Version2,
+};
+
+namespace {
+void GetFirmwareVersionImpl(Kernel::HLERequestContext& ctx, GetFirmwareVersionType type) {
+    LOG_WARNING(
+        Service_SET,
+        "called - Using hardcoded firmware version 'YuzuEmulated Firmware for NX 5.1.0-0.0'");
```
lioncash: This probably shouldn't include the hardcode the string into the message itself. It's easier for someone to forget to update this if we ever bump the version on the firmware info (which is in a different cpp file).

src/core/hle/service/set/set_sys.cpp
```diff
@@ -11,61 +12,80 @@
...
+    if (type == GetFirmwareVersionType::Version1) {
+        data[SYSTEM_VERSION_FILE_MINOR_REVISION_OFFSET] = 0;
     }
 
     ctx.WriteBuffer(data);
 
     IPC::ResponseBuilder rb{ctx, 2};
     rb.Push(RESULT_SUCCESS);
 }
+} // namespace
```
lioncash: `// Anonymous namespace`

src/core/file_sys/system_archive/system_version.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+VirtualDir SystemVersion() {
+    VirtualFile file = std::make_shared<VectorVfsFile>(std::vector<u8>(0x100), "file");
+    file->WriteObject(SystemVersionData::VERSION_MAJOR, 0);
+    file->WriteObject(SystemVersionData::VERSION_MINOR, 1);
+    file->WriteObject(SystemVersionData::VERSION_MICRO, 2);
+    file->WriteObject(SystemVersionData::REVISION_MAJOR, 4);
+    file->WriteObject(SystemVersionData::REVISION_MINOR, 5);
+    file->WriteArray(SystemVersionData::PLATFORM_STRING,
+                     std::min<u64>(sizeof(SystemVersionData::PLATFORM_STRING), 0x20ull), 0x8);
```
lioncash: Prefer upper-case for literals (`ULL`), since that's the general convention (and `L` is less likely to be read accidentally as a `1` with different editor fonts)

src/core/hle/service/set/set_sys.cpp
```diff
@@ -2,13 +2,88 @@
...
+#include "core/file_sys/errors.h"
+#include "core/file_sys/system_archive/system_version.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/kernel/client_port.h"
+#include "core/hle/service/filesystem/filesystem.h"
 #include "core/hle/service/set/set_sys.h"
 
 namespace Service::Set {
 
+constexpr u64 SYSTEM_VERSION_FILE_MINOR_REVISION_OFFSET = 0x05;
```
lioncash: Missed these (sorry about that), but this and the enum should be included in the anonymous namespace as well.

DarkLordZach: np

PR #2224, opened 2019-03-11
hwopus: Leverage multistream API for decoding regular Opus packets

lioncash:
After doing a little more reading up on the Opus codec, it turns out that the multistream API that is part of libopus can handle regular packets. Regular packets are just a degenerate case of multistream Opus packets, and all that's necessary is to pass the number of streams as 1 and provide a basic channel mapping, then everything works fine for that case.

This allows us to get rid of the need to use both APIs in the future when implementing multistream variants in a follow-up PR, greatly simplifying the code that needs to be written.

### DIFF
--- a/src/core/hle/service/audio/hwopus.cpp
+++ b/src/core/hle/service/audio/hwopus.cpp
@@ -8,6 +8,7 @@
 #include <vector>
 
 #include <opus.h>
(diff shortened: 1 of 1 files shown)


lioncash:
Failure doesn't seem to be related

PR #2226, opened 2019-03-11
kernel/server_port: Make data members private

lioncash:
With this, all kernel objects finally have all of their data members behind an interface, making it nicer to reason about interactions with other code (as external code no longer has the freedom to totally alter internals and potentially messing up invariants).


lioncash:
Failure is due to a timeout.

PR #2227, opened 2019-03-11
renderer_opengl/gl_global_cache: Add missing override specifiers

lioncash:
Silences compiler warnings. While we're in the area, also convert `ReserveGlobalRegion()` over to using `insert_or_assign()`, since it's a fairly trivial change.


lioncash:
Failure is due to a timeout.

PR #2229, opened 2019-03-12
vk_sampler_cache: Implement a sampler cache

ReinUsesLisp:
Implements a sampler cache that stores Vulkan sampler handles. As future improvement it could receive a fence on sampler queries and destroy samplers that are not being used and have not been used for some time, reducing memory overhead - that being said, samplers should be cheap on most implementations.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_sampler_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_sampler_cache.cpp
@@ -35,7 +35,7 @@ std::size_t SamplerCacheKey::Hash() const {
 }
 
 bool SamplerCacheKey::operator==(const SamplerCacheKey& rhs) const {
-    return std::memcmp(raw.data(), rhs.raw.data(), sizeof(raw)) == 0;
+    return raw == rhs.raw;
 }
 
 VKSamplerCache::VKSamplerCache(const VKDevice& device) : device{device} {}
--- a/src/video_core/renderer_vulkan/vk_sampler_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_sampler_cache.cpp
@@ -1,0 +1,81 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <cstring>
+#include <optional>
+#include <unordered_map>
+
+#include "common/assert.h"
+#include "common/cityhash.h"
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/renderer_vulkan/maxwell_to_vk.h"
+#include "video_core/renderer_vulkan/vk_sampler_cache.h"
+#include "video_core/textures/texture.h"
+
+namespace Vulkan {
+
+static std::optional<vk::BorderColor> TryConvertBorderColor(std::array<float, 4> color) {
+    // TODO(Rodrigo): Manage integer border colors
+    if (color == std::array<float, 4>{0, 0, 0, 0}) {
+        return vk::BorderColor::eFloatTransparentBlack;
+    } else if (color == std::array<float, 4>{0, 0, 0, 1}) {
+        return vk::BorderColor::eFloatOpaqueBlack;
+    } else if (color == std::array<float, 4>{1, 1, 1, 1}) {
+        return vk::BorderColor::eFloatOpaqueWhite;
+    } else {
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_vulkan/vk_sampler_cache.cpp
```diff
@@ -0,0 +1,80 @@
...
+    } else if (color == std::array<float, 4>{1, 1, 1, 1}) {
+        return vk::BorderColor::eFloatOpaqueWhite;
+    } else {
+        return {};
+    }
+}
+
+std::size_t SamplerCacheKey::Hash() const {
+    std::size_t hash = 0;
+    boost::hash_combine(hash, raw);
```
FernandoS27: return Common::CityHash64(reinterpret_cast<char*>(raw.data()), 4);

have in mind that the only legal reinterpret_cast on painters is towards char* or u8*, so there's no undefined behavior. Also don't use hash_combine, it does not calculate hash values, it just mixes data.

src/video_core/renderer_vulkan/vk_sampler_cache.h
```diff
@@ -0,0 +1,56 @@
...
+    }
+};
+
+} // namespace Vulkan
+
+namespace std {
+
+template <>
+struct hash<Vulkan::SamplerCacheKey> {
+    std::size_t operator()(const Vulkan::SamplerCacheKey& k) const {
```
lioncash: This should likely be declared noexcept as well.

src/video_core/renderer_vulkan/vk_sampler_cache.h
```diff
@@ -0,0 +1,56 @@
...
+    }
+};
+
+} // namespace std
+
+namespace Vulkan {
+
+class VKSamplerCache {
+public:
+    VKSamplerCache(const VKDevice& device);
```
lioncash:
```suggestion
    explicit VKSamplerCache(const VKDevice& device);
```

src/video_core/renderer_vulkan/vk_sampler_cache.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include "common/common_types.h"
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/textures/texture.h"
+
+namespace Vulkan {
+
+class VKDevice;
+
+struct SamplerCacheKey : public Tegra::Texture::TSCEntry {
```
lioncash:
```suggestion
struct SamplerCacheKey final : public Tegra::Texture::TSCEntry {
```

src/video_core/renderer_vulkan/vk_sampler_cache.cpp
```diff
@@ -0,0 +1,80 @@
...
+    return *sampler;
+}
+
+UniqueSampler VKSamplerCache::CreateSampler(const Tegra::Texture::TSCEntry& tsc) {
+    const float max_anisotropy = tsc.GetMaxAnisotropy();
+    const bool has_anisotropy = max_anisotropy > 1.0f;
+
+    const auto border_color = tsc.GetBorderColor();
+    const auto vk_border_color = TryConvertBorderColor(border_color);
+    UNIMPLEMENTED_IF_MSG(!vk_border_color, "Unimplemented border color {} {} {} {}",
```
lioncash: `common/assert.h` needs to be included.

src/video_core/renderer_vulkan/vk_sampler_cache.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+std::size_t SamplerCacheKey::Hash() const {
+    static_assert(sizeof(raw) % sizeof(u64) == 0);
+    return static_cast<std::size_t>(
+        Common::CityHash64(reinterpret_cast<const char*>(raw.data()), sizeof(raw) / sizeof(u64)));
+}
+
+bool SamplerCacheKey::operator==(const SamplerCacheKey& rhs) const {
+    return std::memcmp(raw.data(), rhs.raw.data(), sizeof(raw)) == 0;
```
lioncash:
```suggestion
    return raw == rhs.raw;
```

is likely better here, because it's guaranteed to work, even with non-trivial types. The difference in assembly is for the most part, negligible. GCC inlines the memcmp, while clang and MSVC just call memcmp.

PR #2235, opened 2019-03-14
vk_shader_decompiler: Implement a SPIR-V decompiler

ReinUsesLisp:
Adds a SPIR-V decompiler for Vulkan.

This PR adds [sirit](https://github.com/ReinUsesLisp/sirit) as dependency. This is due to SPIR-V being a bytecode IR instead of a human-readable language like GLSL. It feels weird adding to yuzu a dependency that I wrote (with help from @FernandoS27), but I couldn't find at the time of writing this a proper solution for a SPIR-V decompilation backend.

There were three major approaches:
- Emitting (unstandardised) SPIR-V text and compiling it to bytecode SPIR-V with an external tool.
- Emitting Vulkan compatible GLSL and compiling it down to SPIR-V with an external tool.
- Emitting SPIR-V bytecode directly with an external tool. 

I dropped the first option due to not being standardized. The second option is feasible and it leads to the possibility of having a single GLSL decompiler with two paths: one for Vulkan GLSL and one for OpenGL GLSL. The last option is the one implemented in this PR. While emitting GLSL and compiling it down to is a good idea due to code re-usability, emitting SPIR-V directly gives more control on what you are emitting and keeps more critical logic in the hands of the decompiler because there aren't two intermediary compilers (one for `GLSL -> SPIR-V` and `SPIR-V -> Host driver`).

This doesn't aim to be a full implementation, it just adds the most common stuff. Decoding logic is still handled by the shader IR.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -43,3 +43,6 @@
 [submodule "externals/zstd"]
     path = externals/zstd
     url = https://github.com/facebook/zstd
+[submodule "sirit"]
+    path = externals/sirit
+    url = https://github.com/ReinUsesLisp/sirit
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -72,6 +72,11 @@ if (USE_DISCORD_PRESENCE)
     target_include_directories(discord-rpc INTERFACE ./discord-rpc/include)
 endif()
 
+# Sirit
+if (ENABLE_VULKAN)
+    add_subdirectory(sirit)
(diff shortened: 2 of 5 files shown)


ReinUsesLisp:
MinGW build error seems unrelated.

bunnei:
@ReinUsesLisp needs a rebase

ReinUsesLisp:
@bunnei rebased

PR #2237, opened 2019-03-15
gpu: Use host address for caching instead of guest address.

bunnei:
This is the first of several changes that are in the March Patreon build.

With this change, we change our rasterizer caching interface to be based on "CacheAddr", instead of guest CPU address. The latter did not really make a ton of sense (other than being a quick shortcut with our existing CPU memory map...), as the GPU is not aware of this address space.

Ideally, "CacheAddr" would be physical address (the one "true source" of the cached data, which is independent of any emulated virtual address space), but yuzu does not emulate this (...maybe it will in the future). In lieu of this, we use host address - as that is analogous to the "true source" of the data.

This allows us to avoid lots of GPU-to-CPU address lookups (will be more apparent with subsequent changes), which were accounting for 2-3ms per frame in graphically intensive games (BotW, SMO, etc.). 

I'm not necessarily convinced that this is the best approach - since ideally, our impl. would not depend on host state, so I'm open to feedback if folks think this is horrible. But of all ideas I came up with (short of implementing the MMU, and incurring the costs of this...), this seemed like the best. Furthermore, the refactoring here sets us up to more easily swap out the "CacheAddr" abstraction with any other address space.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -67,8 +67,11 @@ static void MapPages(PageTable& page_table, VAddr base, u64 size, u8* memory, Pa
     LOG_DEBUG(HW_Memory, "Mapping {} onto {:016X}-{:016X}", fmt::ptr(memory), base * PAGE_SIZE,
               (base + size) * PAGE_SIZE);
 
-    RasterizerFlushVirtualRegion(base << PAGE_BITS, size * PAGE_SIZE,
-                                 FlushMode::FlushAndInvalidate);
+    // During boot, current_page_table might not be set yet, in which case we need not flush
+    if (current_page_table) {
+        RasterizerFlushVirtualRegion(base << PAGE_BITS, size * PAGE_SIZE,
+                                     FlushMode::FlushAndInvalidate);
+    }
 
     VAddr end = base + size;
     ASSERT_MSG(end <= page_table.pointers.size(), "out of range mapping at {:016X}",
@@ -359,13 +362,13 @@ void RasterizerFlushVirtualRegion(VAddr start, u64 size, FlushMode mode) {
         auto& gpu = system_instance.GPU();
         switch (mode) {
         case FlushMode::Flush:
-            gpu.FlushRegion(overlap_start, overlap_size);
+            gpu.FlushRegion(ToCacheAddr(GetPointer(overlap_start)), overlap_size);
             break;
         case FlushMode::Invalidate:
-            gpu.InvalidateRegion(overlap_start, overlap_size);
+            gpu.InvalidateRegion(ToCacheAddr(GetPointer(overlap_start)), overlap_size);
             break;
         case FlushMode::FlushAndInvalidate:
-            gpu.FlushAndInvalidateRegion(overlap_start, overlap_size);
+            gpu.FlushAndInvalidateRegion(ToCacheAddr(GetPointer(overlap_start)), overlap_size);
             break;
         }
     };
--- a/src/video_core/engines/kepler_memory.cpp
+++ b/src/video_core/engines/kepler_memory.cpp
@@ -9,6 +9,7 @@
 #include "video_core/engines/kepler_memory.h"
 #include "video_core/engines/maxwell_3d.h"
 #include "video_core/rasterizer_interface.h"
+#include "video_core/renderer_base.h"
 
 namespace Tegra::Engines {
 
@@ -48,7 +49,8 @@ void KeplerMemory::ProcessData(u32 data) {
     // We have to invalidate the destination region to evict any outdated surfaces from the cache.
     // We do this before actually writing the new data because the destination address might contain
     // a dirty surface that will have to be written back to memory.
-    Core::System::GetInstance().GPU().InvalidateRegion(*dest_address, sizeof(u32));
+    system.Renderer().Rasterizer().InvalidateRegion(ToCacheAddr(Memory::GetPointer(*dest_address)),
+                                                    sizeof(u32));
 
     Memory::Write32(*dest_address, data);
     system.GPU().Maxwell3D().dirty_flags.OnMemoryWrite();
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -396,7 +396,10 @@ void Maxwell3D::ProcessCBData(u32 value) {
     const auto address = memory_manager.GpuToCpuAddress(buffer_address + regs.const_buffer.cb_pos);
     ASSERT_MSG(address, "Invalid GPU address");
 
-    Memory::Write32(*address, value);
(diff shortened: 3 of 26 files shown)


src/core/memory.cpp
```diff
@@ -67,8 +67,8 @@ static void MapPages(PageTable& page_table, VAddr base, u64 size, u8* memory, Pa
     LOG_DEBUG(HW_Memory, "Mapping {} onto {:016X}-{:016X}", fmt::ptr(memory), base * PAGE_SIZE,
               (base + size) * PAGE_SIZE);
 
-    RasterizerFlushVirtualRegion(base << PAGE_BITS, size * PAGE_SIZE,
-                                 FlushMode::FlushAndInvalidate);
+    // RasterizerFlushVirtualRegion(base << PAGE_BITS, size * PAGE_SIZE,
```
lioncash: Accidentally left in?

src/video_core/dma_pusher.h
```diff
@@ -29,6 +29,11 @@ struct CommandListHeader {
         BitField<41, 1, u64> is_non_main;
         BitField<42, 21, u64> size;
     };
+
+    CommandListHeader& operator=(const CommandListHeader& t) {
```
lioncash: This shouldn't be necessary, now that BitField is trivially copyable in master.

src/video_core/gpu.h
```diff
@@ -11,6 +11,11 @@
 #include "video_core/dma_pusher.h"
 #include "video_core/memory_manager.h"
 
+using CacheAddr = std::uintptr_t;
+constexpr CacheAddr ToCacheAddr(void* host_ptr) {
```
lioncash: This should likely take the pointer as `const void*` to allow const pointers to be used with it as well.

src/core/memory.cpp
bunnei: Leaving this commented out was unintended, will fix

src/video_core/rasterizer_cache.h
```diff
@@ -58,6 +71,8 @@ class RasterizerCacheObject {
     bool is_registered{};      ///< Whether the object is currently registered with the cache
     bool is_dirty{};           ///< Whether the object is dirty (out of sync with guest memory)
     u64 last_modified_ticks{}; ///< When the object was last modified, used for in-order flushing
+    CacheAddr cache_addr{};
```
lioncash: Doc comments (if there's anything particularly special about these members that might not be obvious).

src/video_core/renderer_opengl/gl_buffer_cache.cpp
```diff
@@ -39,17 +45,18 @@ GLintptr OGLBufferCache::UploadMemory(Tegra::GPUVAddr gpu_addr, std::size_t size
     AlignBuffer(alignment);
     const GLintptr uploaded_offset = buffer_offset;
 
-    Memory::ReadBlock(*cpu_addr, buffer_ptr, size);
+    if (host_ptr) {
```
lioncash: This can be inverted into a guard clause like:

```suggestion
    if (!host_ptr) {
        return uplocaded_offset;
    }
```

letting the memcpy sit next to the related buffer_ptr manipulation down below.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -42,9 +42,9 @@ VAddr GetShaderAddress(Maxwell::ShaderProgram program) {
 }
 
 /// Gets the shader program code from memory for the specified address
-ProgramCode GetShaderCode(VAddr addr) {
+ProgramCode GetShaderCode(u8* host_ptr) {
```
lioncash:
```suggestion
ProgramCode GetShaderCode(const u8* host_ptr) {
```

src/video_core/rasterizer_cache.h
```diff
@@ -89,48 +108,60 @@ class RasterizerCache : NonCopyable {
...
-    T TryGet(VAddr addr) const {
+    /// Tries to get an object from the cache with the specified cache address
+    T TryGet(CacheAddr addr) const {
         const auto iter = map_cache.find(addr);
         if (iter != map_cache.end())
             return iter->second;
         return nullptr;
     }
 
+    T TryGet(void* addr) const {
```
lioncash: This can be made:

```suggestion
    T TryGet(const void* addr) const {
```

if `ToCacheAddr`'s parameter is made const.

src/video_core/rasterizer_cache.h
```diff
@@ -12,14 +13,26 @@
 
 #include "common/common_types.h"
 #include "core/settings.h"
+#include "video_core/gpu.h"
 #include "video_core/rasterizer_interface.h"
 
 class RasterizerCacheObject {
 public:
+    explicit RasterizerCacheObject(u8* host_ptr)
```
lioncash: This can be made const

```suggestion
    explicit RasterizerCacheObject(const u8* host_ptr)
```

if ToCacheAddr's parameter is made const.

src/video_core/engines/kepler_memory.cpp
```diff
@@ -48,7 +49,8 @@ void KeplerMemory::ProcessData(u32 data) {
     // We have to invalidate the destination region to evict any outdated surfaces from the cache.
     // We do this before actually writing the new data because the destination address might contain
     // a dirty surface that will have to be written back to memory.
-    Core::System::GetInstance().GPU().InvalidateRegion(*dest_address, sizeof(u32));
+    Core::System::GetInstance().Renderer().Rasterizer().InvalidateRegion(
```
ReinUsesLisp: There's a system instance in this class.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -396,7 +396,10 @@ void Maxwell3D::ProcessCBData(u32 value) {
     const auto address = memory_manager.GpuToCpuAddress(buffer_address + regs.const_buffer.cb_pos);
     ASSERT_MSG(address, "Invalid GPU address");
 
-    Memory::Write32(*address, value);
+    const auto& ptr{Memory::GetPointer(*address)};
```
ReinUsesLisp: Is it intended to store this by reference?

src/video_core/rasterizer_cache.h
```diff
@@ -68,15 +83,19 @@ class RasterizerCache : NonCopyable {
     explicit RasterizerCache(VideoCore::RasterizerInterface& rasterizer) : rasterizer{rasterizer} {}
 
     /// Write any cached resources overlapping the specified region back to memory
-    void FlushRegion(Tegra::GPUVAddr addr, size_t size) {
+    void FlushRegion(CacheAddr addr, size_t size) {
```
ReinUsesLisp: `std::size_t`

PR #2238, opened 2019-03-15
kernel/thread: Amend conditional test and assignment within UpdatePriority()

lioncash:
This condition was checking against the nominal thread priority, whereas the kernel itself checks against the current priority instead. We were also assigning the nominal priority, when we should be assigning
current_priority, which takes priority inheritance into account.

This can lead to the incorrect priority being assigned to a thread.

Given we recursively update the relevant threads, we don't need to go through the whole mutex waiter list. This matches what the kernel does as well (only accessing the first entry within the waiting list) per execution.


FernandoS27:
You should check that we keep the mutex waiting list ordered by priority. I believe we don't.

bunnei:
@lioncash this causes a crash in BotW at boot

PR #2241, opened 2019-03-16
CMakeLists: Remove now-unnecessary GCC special-casing

lioncash:
This issue has since been fixed in newer versions of Boost, so we don't need to worry about this anymore.


lioncash:
Error appears to be unrelated (given this only affects GCC compilers)

PR #2252, opened 2019-03-17
core: Move PageTable struct into Common.

bunnei:
We will reuse this for the GPU page table, hence the move to Common.

### DIFF
--- a/src/common/page_table.cpp
+++ b/src/common/page_table.cpp
@@ -1,0 +1,29 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/page_table.h"
+
+namespace Common {
+
+PageTable::PageTable(std::size_t page_size_in_bits) : page_size_in_bits{page_size_in_bits} {}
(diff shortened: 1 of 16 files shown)


src/common/page_table.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/page_table.h"
+
+namespace Common {
+
+PageTable::PageTable(std::size_t page_size_in_bits) : page_size_in_bits{page_size_in_bits} {}
```
lioncash: This is likely going to break the VMManager without the Resize() call in the constructor like it originally had. This is going to break assumptions when mapping memory

bunnei: That call was redundant as far as I can sell - Resize(..) is called here: https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/kernel/vm_manager.cpp#L80

lioncash: Ah, true. Looks good then.

PR #2253, opened 2019-03-17
Migrate off directly modifying CMAKE_* compilation-related flags directly

lioncash:
Modifying CMAKE_* related flags directly applies those changes to every single CMake target. This includes even the targets we have in the externals directory.

So, if we ever increased our warning levels, or enabled particular ones, or enabled any other compilation setting, then this would apply to externals as well, which is often not desirable.

This makes our compilation flag setup less error prone by only applying our settings to our targets and leaving the externals alone entirely.

This also means we don't end up clobbering any provided flags on the command line either, allowing developers to specifically use the flags they want.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,77 +104,11 @@ endif()
 message(STATUS "Target architecture: ${ARCHITECTURE}")
 
 
-# Configure compilation flags
+# Configure C++ standard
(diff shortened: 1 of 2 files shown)


lioncash:
Travis CI failure is unrelated (Upload failure. The build itself succeeds)

PR #2254, opened 2019-03-17
input_common/sdl_impl: Minor cleanup in SDLState constructor

lioncash:
Just a tiny bit of tidying up.


lioncash:
Error is unrelated

PR #2256, opened 2019-03-17
gpu: Rewrite MemoryManager based on the VMManager implementation.

bunnei:
This rewrites our GPU memory manager to work more like a real page table, rather than just being a wrapper around the CPU memory manager (VMManager class). This is largely based on the VMManager class. This has a few benefits:
* Logically makes more sense, since the GPU has no knowledge of the process CPU address space
* Faster memory access (we do not need to translate GPUAddr to CPU VAddr always)
* Avoids the need to keep track of CPU VAddr in various places of the GPU
* Should make support for non-contiguous regions easier
* Simplifies some things, also allows us to remove a few hacks

Please review commit by commit. This also needs lots of testing.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_buffer_cache.h
+++ b/src/video_core/renderer_vulkan/vk_buffer_cache.h
@@ -68,8 +68,7 @@ public:
 
     /// Uploads data from a guest GPU address. Returns host's buffer offset where it's been
     /// allocated.
-    u64 UploadMemory(Tegra::GPUVAddr gpu_addr, std::size_t size, u64 alignment = 4,
-                     bool cache = true);
+    u64 UploadMemory(GPUVAddr gpu_addr, std::size_t size, u64 alignment = 4, bool cache = true);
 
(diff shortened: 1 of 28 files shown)


src/video_core/renderer_vulkan/vk_buffer_cache.h
```diff
@@ -68,8 +68,7 @@ class VKBufferCache final : public RasterizerCache<std::shared_ptr<CachedBufferE
 
     /// Uploads data from a guest GPU address. Returns host's buffer offset where it's been
     /// allocated.
-    u64 UploadMemory(Tegra::GPUVAddr gpu_addr, std::size_t size, u64 alignment = 4,
-                     bool cache = true);
+    u64 UploadMemory(GPUVAddr gpu_addr, std::size_t size, u64 alignment = 4, bool cache = true);
```
slashiee: this change causes a compilation error in vk_buffer_cache.cpp: `'GPUVAddr': is not a member of 'Tegra'`

PR #2260, opened 2019-03-18
input_common/sdl: Correct return values within GetPollers implementations

lioncash:
Previously they weren't properly returning the containers at all. While we're at it, introduce a type alias to make relevant function declarations and definitions shorter.


lioncash:
Failure is related to a timeout. The changes themselves are fine.

PR #2265, opened 2019-03-20
Replace old Thread Queue for a new Multi Level Queue

FernandoS27:
This replaces the core datatype for managing threads in the Kernel's Scheduler. The new one: should be faster, is iterable, provides more flexibility and is well tested.

### DIFF
--- a/src/common/multi_level_queue.h
+++ b/src/common/multi_level_queue.h
@@ -1,0 +1,329 @@
+// Copyright 2019 TuxSH
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <iterator>
+#include <list>
+
+#include "common/bit_util.h"
+#include "common/common_types.h"
+
+namespace Common {
+
+template <typename T, std::size_t Depth>
+class MultiLevelQueue {
+public:
+    using value_type = T;
+    using reference = value_type&;
+    using const_reference = const value_type&;
+    using pointer = value_type*;
+    using const_pointer = const value_type*;
+
+    using difference_type = typename std::pointer_traits<pointer>::difference_type;
+    using size_type = std::size_t;
+
+    template <bool is_constant>
+    class iterator_impl {
+    public:
+        using iterator_category = std::bidirectional_iterator_tag;
+        using value_type = T;
+        using pointer = std::conditional_t<is_constant, T*, const T*>;
+        using reference = std::conditional_t<is_constant, const T&, T&>;
+        using difference_type = typename std::pointer_traits<pointer>::difference_type;
+
+        friend bool operator==(const iterator_impl& lhs, const iterator_impl& rhs) {
+            return (lhs.IsEnd() && rhs.IsEnd()) || lhs.it == rhs.it;
+        }
+
+        friend bool operator!=(const iterator_impl& lhs, const iterator_impl& rhs) {
+            return !operator==(lhs, rhs);
+        }
+
+        reference operator*() const {
+            return *it;
+        }
+
+        pointer operator->() const {
+            return it.operator->();
+        }
+
+        iterator_impl& operator++() {
+            if (IsEnd()) {
+                return *this;
+            }
+
+            ++it;
+
+            if (it == GetEndItForPrio()) {
+                u64 prios = mlq.used_priorities;
+                prios &= ~((1ULL << (current_priority + 1)) - 1);
+                if (prios == 0) {
+                    current_priority = mlq.depth();
+                } else {
+                    current_priority = CountTrailingZeroes64(prios);
+                    it = GetBeginItForPrio();
+                }
+            }
+            return *this;
+        }
+
+        iterator_impl& operator--() {
+            if (IsEnd()) {
+                if (mlq.used_priorities != 0) {
+                    current_priority = 63 - CountLeadingZeroes64(mlq.used_priorities);
+                    it = GetEndItForPrio();
+                    --it;
+                }
+            } else if (it == GetBeginItForPrio()) {
+                u64 prios = mlq.used_priorities;
+                prios &= (1ULL << current_priority) - 1;
+                if (prios != 0) {
+                    current_priority = CountTrailingZeroes64(prios);
+                    it = GetEndItForPrio();
+                    --it;
+                }
+            } else {
+                --it;
+            }
+            return *this;
+        }
+
+        iterator_impl operator++(int) {
+            const iterator_impl v{*this};
+            ++(*this);
+            return v;
+        }
+
+        iterator_impl operator--(int) {
+            const iterator_impl v{*this};
+            --(*this);
+            return v;
+        }
+
+        // allow implicit const->non-const
+        iterator_impl(const iterator_impl<false>& other)
+            : mlq(other.mlq), it(other.it), current_priority(other.current_priority) {}
+
+        iterator_impl& operator=(const iterator_impl<false>& other) {
+            mlq = other.mlq;
+            it = other.it;
+            current_priority = other.current_priority;
+            return *this;
+        }
+
+        friend class iterator_impl<true>;
+        iterator_impl() = default;
+
+    private:
+        friend class MultiLevelQueue;
+        using container_ref =
+            std::conditional_t<is_constant, const MultiLevelQueue&, MultiLevelQueue&>;
+        using list_iterator = std::conditional_t<is_constant, typename std::list<T>::const_iterator,
+                                                 typename std::list<T>::iterator>;
+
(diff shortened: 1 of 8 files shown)


src/common/multi_level_queue.h
```diff
@@ -0,0 +1,331 @@
...
+        using iterator_category = std::bidirectional_iterator_tag;
+        using value_type = T;
+        using pointer = std::conditional_t<is_constant, T*, const T*>;
+        using reference = std::conditional_t<is_constant, const T&, T&>;
+        using difference_type = typename std::pointer_traits<pointer>::difference_type;
+
+        friend bool operator==(const iterator_impl& lhs, const iterator_impl& rhs) {
+            if (lhs.IsEnd() && rhs.IsEnd())
+                return true;
+            if (lhs.current_priority == rhs.current_priority)
```
lioncash:
```cpp
return std::tie(lhs.current_priority, lhs.it) == std::tie(rhs.current_priority, rhs.it);

File: src/common/multi_level_queue.h [hunk 2]
```diff
@@ -0,0 +1,331 @@
...
+#include <iterator>
+#include <list>
+
+#include "common/bit_util.h"
+#include "common/common_types.h"
+
+namespace Common {
+
+template <typename T, std::size_t Depth>
+class MultiLevelQueue {
```
lioncash: This should likely have a sizable documentation comment above it explaining the class and the underlying design of it to make it easier for the uninitiated to get up to speed with the data structure.

src/common/multi_level_queue.h
```diff
@@ -0,0 +1,331 @@
...
+    const T& back(u32 min_priority = Depth - 1) const {
+        const u32 priority = lowest_priority_set(min_priority); // intended
+        return levels[priority == Depth ? 63 : priority].back();
+    }
+
+private:
+    using const_list_iterator = typename std::list<T>::const_iterator;
+
+    static void ListShiftForward(std::list<T>& list, const std::size_t shift = 1) {
+        // NOTE: May want to consider making this an assertion or something
```
lioncash: This should probably be addressed or be removed, otherwise.

src/core/hle/kernel/scheduler.cpp
```diff
@@ -45,23 +45,27 @@ Thread* Scheduler::PopNextReadyThread() {
     Thread* next = nullptr;
     Thread* thread = GetCurrentThread();
 
+
```
lioncash: Unnecessary newline

src/core/hle/kernel/scheduler.cpp
```diff
@@ -45,23 +45,27 @@ Thread* Scheduler::PopNextReadyThread() {
...
     if (thread && thread->GetStatus() == ThreadStatus::Running) {
+        if (ready_queue.empty())
```
lioncash: This should be braced

src/core/hle/kernel/scheduler.cpp
```diff
@@ -45,23 +45,27 @@ Thread* Scheduler::PopNextReadyThread() {
...
-        next = ready_queue.pop_first_better(thread->GetPriority());
-        if (!next) {
-            // Otherwise just keep going with the current thread
+        next = ready_queue.front();
+        if (next == nullptr || next->GetPriority() >= thread->GetPriority()) {
             next = thread;
         }
     } else {
-        next = ready_queue.pop_first();
+        if (ready_queue.empty())
```
lioncash: Ditto

src/core/hle/kernel/scheduler.cpp
```diff
@@ -45,23 +45,27 @@ Thread* Scheduler::PopNextReadyThread() {
...
+            return nullptr;
+        next = ready_queue.front();
     }
 
     return next;
 }
 
 void Scheduler::SwitchContext(Thread* new_thread) {
-    Thread* const previous_thread = GetCurrentThread();
+    Thread* previous_thread = GetCurrentThread();
```
lioncash: `const` still applies here.

src/core/hle/kernel/scheduler.cpp
```diff
@@ -161,33 +164,35 @@ void Scheduler::ScheduleThread(Thread* thread, u32 priority) {
...
     std::lock_guard<std::mutex> lock(scheduler_mutex);
 
     ASSERT(thread->GetStatus() == ThreadStatus::Ready);
-    ready_queue.remove(priority, thread);
+    ready_queue.remove(thread, priority);
 }
 
 void Scheduler::SetThreadPriority(Thread* thread, u32 priority) {
     std::lock_guard<std::mutex> lock(scheduler_mutex);
+    if (thread->GetPriority() == priority)
```
lioncash: This should be braced

src/core/hle/kernel/scheduler.cpp
```diff
@@ -161,33 +164,35 @@ void Scheduler::ScheduleThread(Thread* thread, u32 priority) {
...
+        return;
 
     // If thread was ready, adjust queues
     if (thread->GetStatus() == ThreadStatus::Ready)
```
lioncash: Ditto

src/core/hle/kernel/scheduler.cpp
```diff
@@ -161,33 +164,35 @@ void Scheduler::ScheduleThread(Thread* thread, u32 priority) {
...
 
 Thread* Scheduler::GetNextSuggestedThread(u32 core, u32 maximum_priority) const {
     std::lock_guard<std::mutex> lock(scheduler_mutex);
 
     const u32 mask = 1U << core;
-    return ready_queue.get_first_filter([mask, maximum_priority](Thread const* thread) {
-        return (thread->GetAffinityMask() & mask) != 0 && thread->GetPriority() < maximum_priority;
-    });
+    for (auto& thread : ready_queue) {
+        if ((thread->GetAffinityMask() & mask) != 0 && thread->GetPriority() < maximum_priority)
```
lioncash: Ditto

src/core/hle/kernel/scheduler.cpp
```diff
@@ -161,33 +164,35 @@ void Scheduler::ScheduleThread(Thread* thread, u32 priority) {
...
 }
 
 Thread* Scheduler::GetNextSuggestedThread(u32 core, u32 maximum_priority) const {
     std::lock_guard<std::mutex> lock(scheduler_mutex);
 
     const u32 mask = 1U << core;
-    return ready_queue.get_first_filter([mask, maximum_priority](Thread const* thread) {
-        return (thread->GetAffinityMask() & mask) != 0 && thread->GetPriority() < maximum_priority;
-    });
+    for (auto& thread : ready_queue) {
```
lioncash:
```suggestion
    for (auto* thread : ready_queue) {
```

src/tests/common/bit_utils.cpp
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <catch2/catch.hpp>
+#include <math.h>
+#include "common/bit_util.h"
+
+namespace Common {
+
+inline u32 CTZ32(u32 value) {
```
lioncash: Why are these functions necessary, you can just test against the direct values in the conditionals below, rather than re-rolling CTZ a different way.

src/tests/common/bit_utils.cpp
```diff
@@ -0,0 +1,42 @@
...
+  u64 count = 0;
+  while (((value >> count) & 0xf) == 0 && count < 64)
+    count += 4;
+  while (((value >> count) & 1) == 0 && count < 64)
+    count++;
+  return count;
+}
+
+
+TEST_CASE("BitUtils", "[common]") {
```
lioncash:
```suggestion
TEST_CASE("BitUtils::CountTrailingZeroes", "[common]") {
```

src/tests/common/multi_level_queue.cpp
```diff
@@ -0,0 +1,55 @@
...
+// Refer to the license.txt file included.
+
+#include <catch2/catch.hpp>
+#include <math.h>
+#include "common/common_types.h"
+#include "common/multi_level_queue.h"
+
+namespace Common {
+
+TEST_CASE("MultiLevelQueue", "[common]") {
```
lioncash: This is... kind of bare. Particularly for a whole data structure. You may want to break this out into several tests, that test against particular aspects of the data structure.

src/common/multi_level_queue.h
FernandoS27: uhm, does that guarantee that the first test is done before the second one always? I can't compare two iterators if they are from different lists.

lioncash: Yes, it does.

PR #2266, opened 2019-03-20
Kernel: Fixes to Arbitration and SignalProcessWideKey Management

FernandoS27:
This fixes a few wrongs in our implementation of wait address, cond vars and miscelleneus. It also creates the new Thread State "WaitCondVar" to mark a difference between threads waiting on condvars and threads waiting on mutexes.

### DIFF
--- a/src/core/hle/kernel/address_arbiter.cpp
+++ b/src/core/hle/kernel/address_arbiter.cpp
@@ -26,7 +26,7 @@ void WakeThreads(const std::vector<SharedPtr<Thread>>& waiting_threads, s32 num_
     // them all.
     std::size_t last = waiting_threads.size();
     if (num_to_wake > 0) {
-        last = num_to_wake;
+        last = std::min(last, static_cast<std::size_t>(num_to_wake));
     }
 
     // Signal the waiting threads.
@@ -90,9 +90,9 @@ ResultCode AddressArbiter::ModifyByWaitingCountAndSignalToAddressIfEqual(VAddr a
     // Determine the modified value depending on the waiting count.
     s32 updated_value;
     if (waiting_threads.empty()) {
-        updated_value = value - 1;
-    } else if (num_to_wake <= 0 || waiting_threads.size() <= static_cast<u32>(num_to_wake)) {
         updated_value = value + 1;
+    } else if (num_to_wake <= 0 || waiting_threads.size() <= static_cast<u32>(num_to_wake)) {
+        updated_value = value - 1;
     } else {
         updated_value = value;
     }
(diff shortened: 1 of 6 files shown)


src/core/hle/kernel/address_arbiter.cpp
```diff
@@ -90,9 +90,9 @@ ResultCode AddressArbiter::ModifyByWaitingCountAndSignalToAddressIfEqual(VAddr a
     // Determine the modified value depending on the waiting count.
     s32 updated_value;
     if (waiting_threads.empty()) {
-        updated_value = value - 1;
-    } else if (num_to_wake <= 0 || waiting_threads.size() <= static_cast<u32>(num_to_wake)) {
         updated_value = value + 1;
+    } else if (num_to_wake <= 0 || waiting_threads.size() <= static_cast<u32>(num_to_wake)) {
+        updated_value = value - 1;
```
FernandoS27: This was fixed according to the idb., the values were inversed.

src/core/hle/kernel/svc.cpp
```diff
@@ -1394,17 +1394,20 @@ static ResultCode SignalProcessWideKey(VAddr condition_variable_addr, s32 target
     // them all.
     std::size_t last = waiting_threads.size();
     if (target != -1)
-        last = target;
+        last = std::min(waiting_threads.size(), static_cast<std::size_t>(target));
 
     // If there are no threads waiting on this condition variable, just exit
-    if (last > waiting_threads.size())
+    if (last == 0)
```
Subv: This will always return success if you call the svc with `target = 0`, is that intended?

src/core/hle/kernel/thread.h
```diff
@@ -51,7 +51,8 @@ enum class ThreadStatus {
     WaitIPC,      ///< Waiting for the reply from an IPC request
     WaitSynchAny, ///< Waiting due to WaitSynch1 or WaitSynchN with wait_all = false
     WaitSynchAll, ///< Waiting due to WaitSynchronizationN with wait_all = true
-    WaitMutex,    ///< Waiting due to an ArbitrateLock/WaitProcessWideKey svc
+    WaitMutex,    ///< Waiting due to an ArbitrateLock svc
+    WaitCondVar,  ///< Waiting due to an WaitProcessWideKey svc
```
Subv: Why is this status needed at all? IMO it complicates some of the logic unnecessarily. You can't wait on a condition variable, you wait on a mutex address and occasionally wake up to check the condvar

FernandoS27: That's false, WaitProcessWideKey Releases all mutexes and makes the thread wait on a condition variable, it keeps the mutex adress since it tries to acquire it on being signaled only (it won't try to acquire it if it wakes up from timeout).

Saying a CondVar thread is waiting for mutex is completely false as A. It's waiting on a condition variable address and B. It relinquishes any owned mutex before entering this wait state. On Signaling, the thread will try to acquire the mutex and thus changes this state.

This state it's require: A. to guarantee such threads don't own special mutex (assertion) and B. distinguish them from those actualy waiting for a mutex on debugging.

src/core/hle/kernel/svc.cpp
FernandoS27: Yes.

PR #2269, opened 2019-03-20
Report out of range memory reads instead of crashing.

FernandoS27:
This PR reports out of range reads instead of letting yuzu crash. It should help with debugging and avoid certain crashes with Async on.

### DIFF
--- a/src/common/page_table.cpp
+++ b/src/common/page_table.cpp
@@ -6,7 +6,9 @@
 
 namespace Common {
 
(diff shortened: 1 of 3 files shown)


lioncash:
Just pasting from Discord, that this should at most be canary only and not actually be put in master.

FernandoS27:
I agree, I'll add DO NOT MERGE to the title.

PR #2270, opened 2019-03-21
kernel/svc: Implement svcGetProcessList and svcGetThreadList

lioncash:
Implements two basic info querying supervisor call functions. One retrieves the list of existing processes in general, while the other retrieves the list of threads under the current process. svcGetThreadList also allows passing in a handle to a kernel debug object, but we currently don't implement these events, so I've added an unimplemented assertion for that case.

Fairly simple as far as querying functions go.


bunnei:
Needs a rebase

lioncash:
DOne

PR #2271, opened 2019-03-21
Name the Main and GPU thread to make debugging easier.

FernandoS27:


jroweboy:
i tried, but failed to resist the bikeshed. i don't like putting yuzu in the name cause you already know its a yuzu thread. but even if you insist to keep it, its `yuzu` not `Yuzu` (sorry!)

src/core/core.cpp
```diff
@@ -88,6 +89,8 @@ struct System::Impl {
     ResultStatus RunLoop(bool tight_loop) {
         status = ResultStatus::Success;
 
+        Common::SetCurrentThreadName("Yuzu Main Thread");
```
bunnei: Can just be "main" IMO

src/video_core/gpu_thread.cpp
```diff
@@ -19,6 +20,7 @@ static void RunThread(VideoCore::RendererBase& renderer, Tegra::DmaPusher& dma_p
 
     // Wait for first GPU command before acquiring the window context
     state.WaitForCommands();
+    Common::SetCurrentThreadName("Yuzu GPU Thread");
```
bunnei: Can just be "gpu" IMO

src/core/core.cpp
FernandoS27: Would confuse with QT main (it's called main thread)

FernandoS27:
> i tried, but failed to resist the bikeshed. i don't like putting yuzu in the name cause you already know its a yuzu thread. but even if you insist to keep it, its `yuzu` not `Yuzu` (sorry!)

I insist as it may confuse with the other threads.

bunnei:
@FernandoS27 needs a rebase

FernandoS27:
Comments addressed.

bunnei:
@FernandoS27 MinGW errors are real

bunnei:
@FernandoS27 reminder to address above comment

FernandoS27:
@bunnei I'll look into this in the future. Closing for the time being.

PR #2274, opened 2019-03-21
core/memory: Remove unnecessary includes

lioncash:
In 93da8e0abfcdcc6e3cb5488a0db12373429f1377, the page table construct was moved to the common library (which utilized these inclusions). Since the move, nothing requires these headers to be included within the memory header.


lioncash:
Failure is unrelated to the changes (it's a timeout)

PR #2277, opened 2019-03-22
Revert "Devirtualize Register/Unregister and use a wrapper instead."

bunnei:
- Fixes graphical issues from transitions in Super Mario Odyssey.


EmulationChannel:
Not Fixed Yuzu Canary 
SharedScreenshot

PR #2281, opened 2019-03-22
kernel/codeset: Make CodeSet's memory data member a regular std::vector

lioncash:
The use of a shared_ptr is an implementation detail of the VMManager itself when mapping memory. Because of that, we shouldn't require all users of the CodeSet to have to allocate the shared_ptr ahead of time. It's intended that CodeSet simply pass in the required direct data, and that the memory manager takes care of it from that point on.

This means we just do the shared pointer allocation in a single place, when loading modules, as opposed to in each loader.

### DIFF
--- a/src/core/hle/kernel/code_set.h
+++ b/src/core/hle/kernel/code_set.h
@@ -5,7 +5,6 @@
 #pragma once
 
 #include <cstddef>
-#include <memory>
 #include <vector>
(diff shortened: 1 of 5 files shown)


bunnei:
@lioncash clang-format error

lioncash:
Amended

PR #2291, opened 2019-03-26
yuzu_tester: Add and implement testing utility for homebrew

DarkLordZach:
This is designed to make unit testing of switch homebrew easier and automatable.

Adds:
- A new frontend (`yuzu_tester`) that will not show a window or emulate input (but otherwise is basically `yuzu_cmd`).
- A service that is only installed for that frontend called `yuzutest` -- this is so the homebrew can communicate with yuzu the results of the test, (NOTE: this is not present in yuzu-qt or yuzu-cmd and can't be used to determine if the app is running on an emulator unless it is running on the test utility.)

When a NRO/NSO that uses the testing service is run on `yuzu-tester`, it will be executed headlessely and when complete the result code from the service will display on screen.

The service also supports passing in a string (for example, to parameterize the test) and string output (for more detailed results/debugging).

In addition to this change, I wrote a library and example for libnx on how to use this system effectively. (https://github.com/DarkLordZach/libyuzutest)

Example of console window when running a test:
image

I eventually want to replace the SDL2-render-to-nowhere with a true null graphics backend, but this works for now.

### DIFF
--- a/src/yuzu_tester/service/yuzutest.cpp
+++ b/src/yuzu_tester/service/yuzutest.cpp
@@ -1,0 +1,104 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <memory>
+#include "common/string_util.h"
+#include "core/hle/ipc_helpers.h"
+#include "core/hle/service/service.h"
+#include "core/hle/service/sm/sm.h"
+#include "yuzu_tester/service/yuzutest.h"
+
+namespace Service::Yuzu {
+
+constexpr u64 SERVICE_VERSION = 1;
+
+class YuzuTest final : public ServiceFramework<YuzuTest> {
+public:
+    explicit YuzuTest(std::string data, std::function<void(u32, std::string)> finish_callback)
+        : ServiceFramework{"yuzutest"}, data(std::move(data)),
+          finish_callback(std::move(finish_callback)) {
+        static const FunctionInfo functions[] = {
+            {0, &YuzuTest::Initialize, "Initialize"},
+            {1, &YuzuTest::GetServiceVersion, "GetServiceVersion"},
+            {2, &YuzuTest::GetData, "GetData"},
+            {3, &YuzuTest::SetResultCode, "SetResultCode"},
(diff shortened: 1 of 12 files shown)


DarkLordZach:
@ebrattr  / @Subv :
This implementation has several advantages to just using svcOutputDebugString (or `lm` sysmodule):
1. This requires no modification of core at all. Hooking into svcOutputDebugString to parse output would require modifications to kernel/system code. Additionally, just setting the log mode to only display svcOutput* isn't feasible either as it shares logging class with other things and asserts/critical cannot be blocked from log. 
2. This allows the test to get information from yuzu in the least convoluted way I could find. The best example of this is test parameterization, which should make it easier to test multiple things quickly without rebuilding.
3. (This is a little weaker but) It also eliminates the need to do any sort of string parsing, as would be required on svcOutputDebugString. With the IPC/service commands we can just pass the u32 result code directly instead of parsing a string for it.

I *did* consider svcOutputDebugString and even `lm` (logging sysmodule), but settled on yuzutest as a custom service because it required less changes to core (making it more resistant to other changes) and because it was easier.

src/yuzu_tester/service/yuzutest.cpp
```diff
@@ -0,0 +1,112 @@
...
+        rb.Push(RESULT_SUCCESS);
+    }
+
+    void ExitProgram(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Frontend, "called");
+
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(RESULT_SUCCESS);
+
+        finish_callback(results);
```
Subv: You can std::move this.

src/yuzu_tester/yuzu.cpp
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    bool finished = false;
+    int return_value = 0;
+    const auto callback = [&finished,
+                           &return_value](std::vector<Service::Yuzu::TestResult> results) {
+        finished = true;
+        return_value = 0;
+
+        const auto len =
```
Subv: I don't get what you're trying to acomplish here, could you please add a comment explaining it?

PR #2293, opened 2019-03-26
core: Remove duplicated account JPEG data structure

DarkLordZach:
~~Pretty self-explanatory, but this is only for things that are constant and completely meaningless in the context of yuzu. Examples include: Size of the emulated SD card, whether or not the ethernet adapter is connected, battery status, etc.~~

~~Eventually, we should work to eliminate most/all of these, but consolidating them into one place seemed like a good idea.~~

~~This also eliminates a duplicate of the backup account JPEG that was in the frontend and hle.~~

Only extracts the account JPEG, which was previously duplicated in frontend and core.


lioncash:
I don't really think it's wise to place everything in one header like this. It makes values and types accessible outside of their intended areas (in the case of types, it ends up pulling in types into files that don't even need them). For example, many of the service-specific things don't need to be exposed outside of that service in particular.

I'd advise against doing this except for the minimal JPEG data, which should have its definition placed within a cpp file and be made extern in a header, to prevent duplicating the whole array into every translation unit that references it.

bunnei:
Reminder to resolve conflicts

PR #2299, opened 2019-03-27
 gl_shader_manager: Remove reliance on a global accessor within MaxwellUniformData::SetFromRegs()

lioncash:
We can just pass in the Maxwell3D instance instead of going through the system class to get at it, getting rid of a hidden dependency. Particularly given the Maxwell3D instance is always available around current callsites.

This also lets us simplify the interface a little bit. Since we pass in the Maxwell3D context now, we only really need to pass the shader stage index value in.

Also performs some minor related cleanup around the code.


lioncash:
Failure is due to a timeout on the MinGW builder.

PR #2300, opened 2019-03-27
shader_cache: Permit a Null Shader in case of a bad host_ptr.

FernandoS27:
This was the behavior we had before the new GPU VM, all it does is return a nullshader code if the host_ptr given is null just like before. It avoids certain crashes and notifies appropriately.

### DIFF
--- a/src/common/assert.h
+++ b/src/common/assert.h
@@ -57,3 +57,21 @@ __declspec(noinline, noreturn)
 
 #define UNIMPLEMENTED_IF(cond) ASSERT_MSG(!(cond), "Unimplemented code!")
 #define UNIMPLEMENTED_IF_MSG(cond, ...) ASSERT_MSG(!(cond), __VA_ARGS__)
+
+// If the assert is ignored, execute _b_
+#define ASSERT_OR_EXECUTE(_a_, _b_)                                                                \
+    do {                                                                                           \
+        ASSERT(_a_);                                                                               \
+        if (!(_a_)) {                                                                              \
+            _b_                                                                                    \
+        }                                                                                          \
+    } while (0)
+
+// If the assert is ignored, execute _b_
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -41,7 +41,12 @@ GPUVAddr GetShaderAddress(Maxwell::ShaderProgram program) {
 /// Gets the shader program code from memory for the specified address
 ProgramCode GetShaderCode(const u8* host_ptr) {
     ProgramCode program_code(VideoCommon::Shader::MAX_PROGRAM_LENGTH);
-    std::memcpy(program_code.data(), host_ptr, program_code.size() * sizeof(u64));
+    if (host_ptr) {
```
bunnei: IMO we could just add an `ASSERT_OR_RETURN_EARLY(host_ptr)` and then merge this. In our production releases (i.e. canary) we should never crash - in fact, it's entirely possible a game will use a bad address somewhere (and just ignore/handle the real error) but we would hard crash here

bunnei:
@FernandoS27 needs clang-format fix

src/common/assert.h
```diff
@@ -57,3 +57,12 @@ __declspec(noinline, noreturn)
 
 #define UNIMPLEMENTED_IF(cond) ASSERT_MSG(!(cond), "Unimplemented code!")
 #define UNIMPLEMENTED_IF_MSG(cond, ...) ASSERT_MSG(!(cond), __VA_ARGS__)
+
+// If the assert is ignored, execute _b_
+#define ASSERT_OR_EXECUTE(_a_, _b_)                                                                \
```
bunnei: Can you add ASSERT_OR_EXECUTE_MSG for completeness?

bunnei:
@FernandoS27 ping on this - also, clang format errors

PR #2301, opened 2019-03-28
core/yuzu: Remove enable_nfc setting

FearlessTobi:
This was initially added to prevent problems from stubbed/not implemented NFC services, but as we never encountered such and as it's only used in a deprecated function anyway, I guess we can just remove it to prevent more clutter of the settings.


FearlessTobi:
Sorry, but can you please tell me a case where this is needed or useful?
I don't see why any user would ever want to turn off NFC which is then only considered in a deprecated function anyway. If you're are a dev and went to play around with NFC - and even there I don't see why you'd want to turn it off - you will very likely be able to recompile yuzu with a local one-line change anyway.

bunnei:
Agree with Tobi - Less settings is almost always better.

PR #2303, opened 2019-03-29
common/thread: Remove unused functions

lioncash:
Many of these functions are carried over from Dolphin (where they aren't used anymore). Given these have no use (and we really shouldn't be screwing around with OS-specific thread scheduler handling from the emulator, these can be removed.

The function for setting the thread name is left, however, since it can have debugging utility usages.

### DIFF
--- a/src/common/thread.cpp
+++ b/src/common/thread.cpp
@@ -27,18 +27,6 @@ namespace Common {
 
 #ifdef _MSC_VER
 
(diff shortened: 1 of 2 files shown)


lioncash:
If we need it in the future, then it can be reintroduced with a rationale for its use. We use a version control system, so it can be brought back in that case.

bunnei:
Agree that we can just add the functions we want if/when we need them

PR #2306, opened 2019-03-30
shader_ir: Implement AOFFI for TEX and TLD4

ReinUsesLisp:
Implements AOFFI arguments for TEX and TLD4. AOFFI maps to the `Offset` GLSL suffix family - e.g. `textureOffset` will use `TEX.AOFFI` while `texture` just uses `TEX`.

This requires light constant folding because in hardware AOFFI coordinates are passed through a packed register while in GLSL these are required to be constant (although Nvidia's driver allows variable coordinates). In case the constant folding fails - fallback to a variable in-shader implementation that will fail on incompatible hardware (this case has not been found).

Arguments are passed through `gpr20`, the first valid argument for these is the bindless pack, this is not handled in this PR.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -34,13 +34,19 @@ using Maxwell = Tegra::Engines::Maxwell3D::Regs;
 using ShaderStage = Tegra::Engines::Maxwell3D::Regs::ShaderStage;
 using Operation = const OperationNode&;
 
+enum class Type { Bool, Bool2, Float, Int, Uint, HalfFloat };
+
+namespace {
+struct TextureAoffi {};
+using TextureArgument = std::pair<Type, Node>;
+using TextureIR = std::variant<TextureAoffi, TextureArgument>;
+} // namespace
+
 enum : u32 { POSITION_VARYING_LOCATION = 0, GENERIC_VARYING_START_LOCATION = 1 };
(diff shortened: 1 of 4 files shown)


src/video_core/shader/decode/texture.cpp
```diff
@@ -463,15 +474,27 @@ Node4 ShaderIR::GetTld4Code(Instruction instr, TextureType texture_type, bool de
...
+
+    u64 parameter_register = instr.gpr20.Value();
+    std::vector<Node> aoffi;
+    if (is_aoffi) {
+        aoffi = GetAoffiCoordinates(GetRegister(parameter_register++), coord_count, true);
+    }
+
+    Node dc{};
+    if (depth_compare) {
+        dc = GetRegister(parameter_register++);
```
ReinUsesLisp: This is not part of AOFFI, but it was found to be missing during the implementation testing.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -34,14 +34,20 @@ using Maxwell = Tegra::Engines::Maxwell3D::Regs;
 using ShaderStage = Tegra::Engines::Maxwell3D::Regs::ShaderStage;
 using Operation = const OperationNode&;
 
+enum class Type { Bool, Bool2, Float, Int, Uint, HalfFloat };
+
+namespace {
```
lioncash: Shouldn't this namespace include everything else local to this translation unit as well?

PR #2315, opened 2019-04-02
shader_ir/decode: Reduce the severity of common assertions

ReinUsesLisp:
Reduce the severity of some checks from asserts to `LOG_DEBUG`, this is mostly because with the current state of shading languages (or shading representations) we can't manage them without a software implementation. Host shader compilers can choose whatever approach the want and OpenGL (and Vulkan) is fine with that.

[SPV_KHR_float_controls](https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_float_controls.html) might help in this regard, but it just exposes this functionality per-shader module instead of per-operation. SPIR-V is capable of exposing these per-operation but that functionality is reserved for kernels, not shaders.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -387,6 +387,13 @@ enum class IpaSampleMode : u64 {
     Offset = 2,
 };
 
+enum class LmemStoreCacheManagement : u64 {
+    Default = 0,
+    CG = 1,
+    CS = 2,
+    WT = 3,
+};
+
 struct IpaMode {
     IpaInterpMode interpolation_mode;
     IpaSampleMode sampling_mode;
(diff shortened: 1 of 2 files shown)


FernandoS27:
@ReinUsesLisp in SPV_float_control if you decorate an instruction with FPRoundingMode that mode is applied, read better in http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_float_controls.html

ReinUsesLisp:
@FernandoS27 It doesn't seem to explicitly mention that the decorations are available per-operation. The spec doesn't mention it either. That being said, I removed changes related to rounding mode since that would require its own PR.

PR #2317, opened 2019-04-02
Implement SyncPoint Register in the GPU.

FernandoS27:
This just implements the register, we still need to implement the appropriate services.

### DIFF
--- a/src/video_core/engines/maxwell_3d.h
+++ b/src/video_core/engines/maxwell_3d.h
@@ -576,7 +576,17 @@ public:
                     u32 bind;
                 } macros;
 
-                INSERT_PADDING_WORDS(0x188);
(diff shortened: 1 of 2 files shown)


src/video_core/engines/maxwell_3d.h
```diff
@@ -576,7 +576,11 @@ class Maxwell3D final {
                     u32 bind;
                 } macros;
 
-                INSERT_PADDING_WORDS(0x188);
+                INSERT_PADDING_WORDS(0x69);
+
+                u32 sync_info;
```
ReinUsesLisp: This can be a BitField definition. That way you can drop the bitwise logic above.

FernandoS27: yeah I was a bit confused with the method argument and stuff.

PR #2319, opened 2019-04-02
gl_framebuffer_cache: Move to its own file and implement invalidation

ReinUsesLisp:
Moves framebuffer cache to its own files and replaces the ordered map with a hashed map.

Implement texture invalidation, on `OGLTexture` constructor it receives a `std::shared_ptr` to the framebuffer cache (or `nullptr`). When releasing `OGLTexture` the framebuffer cache gets signaled that a texture has been destroyed and it looks through the whole cache finding instances where this texture was used and deletes them.

### DIFF
--- a/src/video_core/CMakeLists.txt
+++ b/src/video_core/CMakeLists.txt
@@ -36,6 +36,8 @@ add_library(video_core STATIC
     renderer_base.h
     renderer_opengl/gl_buffer_cache.cpp
     renderer_opengl/gl_buffer_cache.h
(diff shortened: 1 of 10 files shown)


ReinUsesLisp:
cc @degasus since you asked me to implement this when the framebuffer cache was added.

bunnei:
@ReinUsesLisp needs a rebase

ReinUsesLisp:
I prefer to reimplement this after the texture cache is redesigned. Closing for now.

PR #2320, opened 2019-04-03
Optimize CallMethod in Maxwell3D

FernandoS27:
This optimizes slightly the need for checks when the register won't be executed or does not change the dirty state.

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -4,6 +4,7 @@
 
 #include <cinttypes>
 #include <cstring>
+#include <utility>
 #include "common/assert.h"
 #include "core/core.h"
 #include "core/core_timing.h"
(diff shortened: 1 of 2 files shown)


src/video_core/engines/maxwell_3d.cpp
```diff
@@ -24,6 +25,82 @@ Maxwell3D::Maxwell3D(Core::System& system, VideoCore::RasterizerInterface& raste
...
     InitializeRegisterDefaults();
+    InitDirtyRegsMask();
+    InitExecRegsMask();
+}
+
+void Maxwell3D::InitDirtyRegsMask() {
+    std::fill(must_process_dirty.begin(), must_process_dirty.end(), false);
+
+    auto mark_dirty = [&](const u32 start, const u32 end) {
+        for (u32 i = start; i < end; i++) {
```
lioncash: This can be a std::fill operation.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -24,6 +25,82 @@ Maxwell3D::Maxwell3D(Core::System& system, VideoCore::RasterizerInterface& raste
...
       macro_interpreter(*this) {
     InitializeRegisterDefaults();
+    InitDirtyRegsMask();
+    InitExecRegsMask();
+}
+
+void Maxwell3D::InitDirtyRegsMask() {
+    std::fill(must_process_dirty.begin(), must_process_dirty.end(), false);
+
+    auto mark_dirty = [&](const u32 start, const u32 end) {
```
lioncash:
```suggestion
    const auto mark_dirty = [this](const u32 start, const u32 end) {
```

src/video_core/engines/maxwell_3d.h
```diff
@@ -1137,6 +1137,11 @@ class Maxwell3D final {
 private:
     void InitializeRegisterDefaults();
 
+    void InitDirtyRegsMask();
+    void InitExecRegsMask();
+    void CheckDirty(const u32 method);
```
lioncash:
```suggestion
    void CheckDirty(u32 method);
```

value-wise const only has a meaningful use in the definition, not the declaration of a function.

src/video_core/engines/maxwell_3d.h
```diff
@@ -1137,6 +1137,11 @@ class Maxwell3D final {
...
+    void ExecuteMethod(const u32 method, const u32 method_argument);
```
lioncash:
```suggestion
    void ExecuteMethod(u32 method, u32 method_argument);
```

src/video_core/engines/maxwell_3d.h
```diff
@@ -1152,6 +1157,9 @@ class Maxwell3D final {
     /// Parameters that have been submitted to the macro call so far.
     std::vector<u32> macro_params;
 
+    std::array<bool, Regs::NUM_REGS> must_process_dirty;
```
lioncash:
```suggestion
    std::array<bool, Regs::NUM_REGS> must_process_dirty{};
```

otherwise this initializes to an inconsistent state.

These two arrays are also missing a documentation comment.

src/video_core/engines/maxwell_3d.h
```diff
@@ -1152,6 +1157,9 @@ class Maxwell3D final {
...
+    std::array<bool, Regs::NUM_REGS> must_process_exec;
```
lioncash:
```suggestion
    std::array<bool, Regs::NUM_REGS> must_process_exec{};
```

ditto

lioncash:
The commit itself should also contain an explanation of the optimization.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -24,6 +25,82 @@ Maxwell3D::Maxwell3D(Core::System& system, VideoCore::RasterizerInterface& raste
     : memory_manager(memory_manager), system{system}, rasterizer{rasterizer},
       macro_interpreter(*this) {
     InitializeRegisterDefaults();
+    InitDirtyRegsMask();
```
ReinUsesLisp: This is kind of a nit, but since there's already `InitializeRegisterDefaults`, this should be called `InitializeDirtyRegisterMask`.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -24,6 +25,82 @@ Maxwell3D::Maxwell3D(Core::System& system, VideoCore::RasterizerInterface& raste
...
+    InitExecRegsMask();
```
ReinUsesLisp: `InitializeExecRegisterMask`

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -24,6 +25,82 @@ Maxwell3D::Maxwell3D(Core::System& system, VideoCore::RasterizerInterface& raste
...
+}
+
+void Maxwell3D::InitDirtyRegsMask() {
+    std::fill(must_process_dirty.begin(), must_process_dirty.end(), false);
```
ReinUsesLisp: This can be managed in the header with a `{}`.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -24,6 +25,82 @@ Maxwell3D::Maxwell3D(Core::System& system, VideoCore::RasterizerInterface& raste
...
+    end = start + 2 * 32;
+    mark_dirty(start, end);
+
+    start =  MAXWELL3D_REG_INDEX(instanced_arrays);
+    end = start + 32;
+    mark_dirty(start, end);
+}
+
+void Maxwell3D::InitExecRegsMask() {
+    std::fill(must_process_exec.begin(), must_process_exec.end(), false);
```
ReinUsesLisp: This can be handled in the header with a `{}`.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -253,6 +292,57 @@ void Maxwell3D::CallMethod(const GPU::MethodCall& method_call) {
...
+
+    if (regs.reg_array[method] != method_call.argument) {
+        regs.reg_array[method] = method_call.argument;
+        if (must_process_dirty[method]) {
+            CheckDirty(method);
+        }
+    }
+
+    if (must_process_exec[method])
+    {
```
ReinUsesLisp: Run `clang-format`.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -24,6 +25,82 @@ Maxwell3D::Maxwell3D(Core::System& system, VideoCore::RasterizerInterface& raste
...
+
+    auto mark_dirty = [&](const u32 start, const u32 end) {
+        for (u32 i = start; i < end; i++) {
+            must_process_dirty[i] = true;
+        }
+    };
+
+    u32 start = MAXWELL3D_REG_INDEX(rt);
+    u32 end = start + (sizeof(regs.rt[0]) / sizeof(u32)) * Regs::NumRenderTargets;
+    mark_dirty(start, end);
```
ReinUsesLisp: It might be a better idea to define `mark_dirty` as a difference instead of an interval. That way you can drop the `end = start + ...;` statements.

bunnei:
@FernandoS27 clang format errors

bunnei:
@FernandoS27 reminder to address above comment

FernandoS27:
@bunnei I'll refactor and redo this PR later in the future as it will require some redesign. I'll close it until then.

PR #2321, opened 2019-04-03
gl_state: Rework to enable individual applies

ReinUsesLisp:
Implements individual `Apply` methods that change a small subset of the OpenGL state machine. This is the backing functionality of a following commit to remove full `Apply` calls in the places where it's not needed, aswell as de-globalizing `gl_state`.

It also uses generic functions to compare and assign values.

### DIFF
--- a/src/video_core/renderer_opengl/gl_state.cpp
+++ b/src/video_core/renderer_opengl/gl_state.cpp
@@ -10,16 +10,62 @@
 
 namespace OpenGL {
 
-OpenGLState OpenGLState::cur_state;
+using Maxwell = Tegra::Engines::Maxwell3D::Regs;
 
+OpenGLState OpenGLState::cur_state;
 bool OpenGLState::s_rgb_used;
+
+namespace {
+
+template <typename T>
+bool UpdateValue(T& current_value, const T new_value) {
+    const bool changed = current_value != new_value;
+    current_value = new_value;
+    return changed;
+}
+
+template <typename T1, typename T2>
+bool UpdateTie(T1 current_value, const T2 new_value) {
+    const bool changed = current_value != new_value;
+    current_value = new_value;
+    return changed;
+}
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -10,16 +10,62 @@
...
+
+template <typename T>
+inline bool UpdateValue(T& current_value, const T new_value) {
+    const bool changed = current_value != new_value;
+    current_value = new_value;
+    return changed;
+}
+
+template <typename T1, typename T2>
+inline bool UpdateTie(T1 current_value, const T2 new_value) {
```
PZerua: Missing ref in first parameter?
```suggestion
inline bool UpdateTie(T1& current_value, const T2 new_value) {
```

ReinUsesLisp: The reference is passed in the `T1` individual elements, e.g. https://godbolt.org/z/vyAduN
It took me a while to re-realize that (I wrote this some time ago) that. Passing `std::tuple<int>` won't be as reference, `std::tuple<int>&` is illegal in this context and `std::tuple<int&>` (what this template deduces) has the expected behaviour.

PZerua: Really good to know that! Thanks!

lioncash: The reason here is that `tie`, when used in this scenario, actually produces a new object (the tuple of references), so it can't be used as an lvalue reference, since it would be an rvalue in this case.

e.g. The following would work in this scenario by using an rvalue reference.

```cpp
template<typename T1, typename T2>
bool UpdateTie(T1&& current_value, const T2& new_value) {
    return std::exchange(current_value, new_value) != new_value;
}
```

PR #2339, opened 2019-04-05
service/fsp_srv: Update SaveDataInfo and SaveDataDescriptor structs

lioncash:
I realized that I updated the documentation on SwitchBrew a while ago, but never actually updated the structs within yuzu.

This also makes some minor cleanup-related changes involving the structs.


bunnei:
@lioncash clang format

lioncash:
Amended

PR #2348, opened 2019-04-05
Implement Bindless Textures on Shader Decompiler and GL backend

FernandoS27:


src/video_core/engines/maxwell_3d.cpp
```diff
@@ -500,6 +489,22 @@ Texture::FullTextureInfo Maxwell3D::GetStageTexture(Regs::ShaderStage stage,
     return tex_info;
 }
 
+Texture::FullTextureInfo Maxwell3D::GetStageTexture(Regs::ShaderStage stage,
+                                                    std::size_t offset) const {
+    auto& shader = state.shader_stages[static_cast<std::size_t>(stage)];
+    auto& tex_info_buffer = shader.const_buffers[regs.tex_cb_index];
```
lioncash:
```suggestion
    const auto& tex_info_buffer = shader.const_buffers[regs.tex_cb_index];
```

src/video_core/const_buffer_accessor.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <cstring>
+
+#include "common/common_types.h"
+#include "core/core.h"
+#include "video_core/engines/maxwell_3d.h"
+#include "video_core/gpu.h"
+#include "video_core/memory_manager.h"
+
+namespace Tegra {
```
lioncash:
```suggestion
namespace Tegra::ConstBufferAccessor {
```

src/video_core/shader/shader_ir.h
```diff
@@ -195,9 +195,23 @@ enum class ExitMethod {
 
 class Sampler {
 public:
+    // Use this constructor for binded Samplers
```
ReinUsesLisp: `bound` and `samplers`

src/video_core/shader/shader_ir.h
```diff
@@ -219,6 +233,14 @@ class Sampler {
         return is_shadow;
     }
 
+    bool IsBindless() const {
+        return is_bindless;
+    }
+
+    std::pair<u32, u32> GetBindlessCBuf() const {
+        return {offset >> 32, offset & 0x00000000FFFFFFFFULL};
```
ReinUsesLisp: This looks like a `static_cast<u32>(offset)`.

src/video_core/const_buffer_accessor.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace ConstBufferAccessor {
+
+template <typename T>
+T access(Tegra::Engines::Maxwell3D::Regs::ShaderStage stage, u64 const_buffer, u64 offset) {
```
lioncash: Does this really need to be in its own file? This seems like something that would make more sense to be part of the Maxwell3D engine's interface, given it's accessing its shader stages.

In any case, this shouldn't be using the `Core::System::GetInstance()` accessor directly. This should be taking the system object as a parameter.

lioncash:
My bad if any comments overlap with Rodrigo's. My browser window didn't update when his review was submitted.

FernandoS27:
Comments addressed.

PR #2349, opened 2019-04-05
yuzu/debugger/graphics/graphics_surface: General cleanup

lioncash:
Performs a general cleanup to this code, given it hasn't been touched in quite a while.

- Adds the remaining surface formats to the format list.
- Cleans up overload deduction via the use of `qOverload`
- Displays error dialogs to indicate when file I/O errors occur during surface saving.


lioncash:
Note that @FernandoS27 also brought up the idea of simply getting rid of the surface viewer, since Nsight and RenderDoc are generally used instead of the viewer itself.

I'm not opposed to this idea (it'd make deglobalizing the memory-related code slightly easier too). just want to make sure everyone else is cool with getting rid of it, or if anyone has any reason for it to stay.

bunnei:
Kill it! It's a pain to maintain, and I'm also convinced no one uses it.

PR #2353, opened 2019-04-06
yuzu/debugger: Remove graphics surface viewer

lioncash:
This doesn't actually work anymore, and given how long it's been left in that state, it's unlikely anyone actually seriously used it.

Generally it's preferable to use RenderDoc or Nsight to view surfaces.


ReinUsesLisp:
It would also require a non-trivial effort to support non-2D textures.

PR #2358, opened 2019-04-07
gl_shader_cache: Use shared contexts to build shaders in parallel at boot

ReinUsesLisp:
During disk cache driver side shader compilation use shared contexts to invoke from multiple worker threads. As far as I tested, this works only on Nvidia's proprietary driver. Other vendors will on the other contexts but one at a time, there's not much we can do about drivers building shaders one at a time when being invoked from multiple threads. I even tried removing `glGet*`s to avoid halting execution without luck.

This uses #2017 as base. The last commit is the important one.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -98,9 +98,11 @@ struct FramebufferCacheKey {
     }
 };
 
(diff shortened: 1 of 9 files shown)


bunnei:
@ReinUsesLisp xcode errors look real.

bunnei:
@ReinUsesLisp reminder to address above comment

bunnei:
@ReinUsesLisp Sorry - can you rebase this again? Thank you.

PR #2360, opened 2019-04-07
kernel/svc: Deglobalize the supervisor call handlers

lioncash:
Adjusts the interface of the wrappers to take a system reference, which allows accessing a system instance without using the global accessors.

This also allows getting rid of all global accessors within the supervisor call handling code. While this does make the wrappers themselves slightly more noisy, this will be further cleaned up in following changes. This eliminates the global system accessors in the current code while preserving the existing interface, making any replacement changes much less noisy.

This will also make deglobalizing the memory subsystem a little nicer, since the system instance will already be available for utilizing said deglobalized subsystem.


bunnei:
@lioncash needs a rebase

lioncash:
Done

PR #2364, opened 2019-04-07
Kernel: Implement a New Thread Scheduler

FernandoS27:
This PR takes care of re implementing our current scheduler. The new scheduler is meant to be closer to the RE info we have and does a few things in a more appropriate manner. Core differences:

- The new scheduler does not automatically assign threads without an assigned cpu core. Instead it moves them to a suggested queue and resolves them when no scheduled thread can be assigned to the current core.
- All yield types are now implemented and can now be redundant accordingly (before the yielding thread went to sleep, leaving the possibility for a lower priority thread to be scheduled instead).
- Prepare reschedule has been revised and should now be triggered whenever a thread wakes up.
- Some corrections have been made to thread switching and picking threads on the go.

The new scheduler fixes some issues:
- Deadlocks that could occur with hle ipc wait and wait synchronization.
- Better handling of threads on some games, resulting in some less dynarmic usage.
- Unlike the old scheduler, no reads are reported to be running without being set to runs.

Thanks to TuxSH and gdkchan for their work in reversing the switch's kernel.
Thanks to TuxSH' mesosphere and Ryujinx which served as reference to help validate this PR.

PR #2365, opened 2019-04-07
Workaround to Mutex Corruption

FernandoS27:
This a workaround to an unknown bug caused by mutex corruption within game code. The cause is unknown. It requires #2364 .

Here's the full research info:

> well I investigated the softlock issue through a game I thought softlocked derministically , here are the results:
> - First softlocking on condvar mutex is COMPLETELY random. Occurs on different occasions on the booting process of I am Setsuna by an small margin of difference. Logs differ on every run.
> - I have verified that it is the game indeed  the one who is granting mutex ownership to  a paused thread and that the bug is not in our HLE ystem.
> - The paused thread is set to be paused way before the softlock shows an invalid state.
> - The softlock does not occur instantly on an invalid state, the system keeps running for a while.
> 
> My current thoughts on what could be happening:
> - A bug in the Switch's SDK that does not happen  on real hardware.
> - Memory Corruption or bad page mirroring.
> - We were suppose to wake that thread earlier.
> deterministically*

### DIFF
--- a/src/core/hle/kernel/mutex.cpp
+++ b/src/core/hle/kernel/mutex.cpp
@@ -7,6 +7,7 @@
 
 #include "common/assert.h"
 #include "core/core.h"
+#include "core/core_cpu.h"
 #include "core/hle/kernel/errors.h"
 #include "core/hle/kernel/handle_table.h"
 #include "core/hle/kernel/kernel.h"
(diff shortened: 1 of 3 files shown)


bunnei:
Needs a rebase

bunnei:
@FernandoS27 needs a rebase

ogniK5377:
Has any work gone into this to find the cause of the issue?

FernandoS27:
> Has any work gone into this to find the cause of the issue?

It's one of those issues thast happen in game code specially the sdk since it happens in many games. It's hard to know what it may be unless we RE the SDK's condition variable and mutex functions.

FernandoS27:
Fixed correctly in: #3195

PR #2382, opened 2019-04-10
service: Update service function tables

lioncash:
Updates function tables based off information from SwitchBrew.


src/core/hle/service/acc/acc_u1.cpp
```diff
@@ -8,6 +8,7 @@ namespace Service::Account {
 
 ACC_U1::ACC_U1(std::shared_ptr<Module> module, std::shared_ptr<ProfileManager> profile_manager)
     : Module::Interface(std::move(module), std::move(profile_manager), "acc:u1") {
+    // clang-format on
```
mailwl: vice versa (on <--> off)

src/core/hle/service/acc/acc.cpp
```diff
@@ -137,6 +137,7 @@ class IProfile final : public ServiceFramework<IProfile> {
 class IManagerForApplication final : public ServiceFramework<IManagerForApplication> {
 public:
     IManagerForApplication() : ServiceFramework("IManagerForApplication") {
+        // clang-format off
```
VPeruS: Do it have any effect?

lioncash: Yes. Entries added can influence how the table gets formatted according to clang-format. Even if that wasn't the case, I'd still add it to be consistent with how we handle our function tables instead of special-casing them based off whether or not clang-format actually does anything to them.

PR #2383, opened 2019-04-10
gl_shader_decompiler: Disable variable AOFFI on unsupported devices

### LINKED ISSUE #2363: AMD Pokemon regression
After updating yuzu Canary to version 2130, Pokemon: Let's Go, Eevee! is rendering only dialog boxes and menu. Other things like terrain, characters do not render at all. I tried disabling and enabling disk shader cache and asynchronous GPU emulation. Before this update game was working properly only with a text bug.

Here I can see only logo, but terrain with Evee isn't rendered.
pg1

In game dialog boxes are only rendered. I can hear footsteps when I'm moving.
pg1

Character selection seems only to work properly.
pg1

Logs:
[log.zip](https://github.com/yuzu-emu/yuzu/files/3051380/log.zip)

My hardware:
AMD Ryzen 7 2700x
AMD Radeon R9 390

ReinUsesLisp:
Run a test at boot to determine if the hardware is capable of using variable AOFFI (`textureOffset` and friends). In case it's not, disable it on hardware/drivers that don't support it. This should fix #2363 on AMD devices.

This adds a `gl_device` object similar to `vk_device` to store data related to the OpenGL device.

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -5,6 +5,7 @@
 #include <cstddef>
 #include <glad/glad.h>
 
+#include "common/logging/log.h"
 #include "video_core/renderer_opengl/gl_device.h"
 
 namespace OpenGL {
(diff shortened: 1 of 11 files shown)


src/video_core/renderer_opengl/gl_device.cpp
```diff
@@ -0,0 +1,45 @@
...
+    gl_Position = textureOffset(tex, vec2(0), variable_offset);
+}
+)";
+    const GLuint shader{glCreateShaderProgramv(GL_VERTEX_SHADER, 1, &AOFFI_TEST)};
+    GLint link_status{};
+    glGetProgramiv(shader, GL_LINK_STATUS, &link_status);
+    glDeleteProgram(shader);
+
+    const bool supported{link_status == GL_TRUE};
+    LOG_INFO(Render_OpenGL, "Variable AOFFI support: {}", supported ? "Yes" : "No");
```
ReinUsesLisp: I'm not sure if we have a format for these.

FearlessTobi: @chris062689

chris062689: Traditionally they have 
Setting Name: true or false

Example:
[   0.006329] Config <Info> core/settings.cpp:LogSetting:67: Renderer_ShadersAccurateGs: true

ReinUsesLisp:
Addressed.

PR #2390, opened 2019-04-11
game_list: Remove artificially caps to a certain number of directories

FreddyFunk:


jroweboy:
Hey sorry the comment isn't very descriptive, but its not really an artificial limit, its an artifically chosen number to keep from people who select their C drive as the root game folder from bringing their PC to a crawl as qt sets up watchers recursively for all folders on their computer. Its just a cut off point to keep from stupid people doing stupid things, and if you plan to remove it, then you'll need to fix that issue as well.

From the QT docs, we see that file system notification does consume resources, so its not ideal to uncap it completely.
```
The act of monitoring files and directories for modifications consumes system resources. This implies there is a limit to the number of files and directories your process can monitor simultaneously. On all BSD variants, for example, an open file descriptor is required for each monitored file. Some system limits the number of open file descriptors to 256 by default. This means that addPath() and addPaths() will fail if your process tries to add more than 256 files or directories to the file system monitor. Also note that your process may have other file descriptors open in addition to the ones for files being monitored, and these other open descriptors also count in the total. macOS uses a different backend and does not suffer from this issue.
```

src/yuzu/game_list.cpp
```diff
@@ -296,22 +296,19 @@ void GameList::ValidateEntry(const QModelIndex& item) {
...
     }
     // Workaround: Add the watch paths in chunks to allow the gui to refresh
     // This prevents the UI from stalling when a large number of watch paths are added
-    // Also artificially caps the watcher to a certain number of directories
-    constexpr int LIMIT_WATCH_DIRECTORIES = 5000;
-    constexpr int SLICE_SIZE = 25;
-    int len = std::min(watch_list.length(), LIMIT_WATCH_DIRECTORIES);
-    for (int i = 0; i < len; i += SLICE_SIZE) {
-        watcher->addPaths(watch_list.mid(i, i + SLICE_SIZE));
+    constexpr std::size_t slice_size = 25;
```
lioncash: This and the loop induction variable should still be left as `int`. `QStringList`'s length() member function returns an `int`, so these would cause signedness mismatch warnings.

FreddyFunk:
I personally don't think that a hard coded limit is the right way to protect a user from a stupid action like that. 
Since this isn't such an important change, feel free to close this PR, if it doesn't fit into yuzu's style.

ogniK5377:
No user has 5000 games as of right now and jroweboy makes a strong case for keeping the limit in place. If at any point this becomes an issue we can always raise the limit if needed.

PR #2393, opened 2019-04-12
kernel/svc: Implement svcMapProcessCodeMemory/svcUnmapProcessCodeMemory

lioncash:
These are utilized for mapping code modules into memory. Notably, the ldr service would call this in order to map objects into memory. This is what our HLE implementation of said service should actually be calling into to perform the mapping, but our interface isn't currently set up that way to allow for that.

With a few more subsequent modifications to our svc handler following these changes, this will actually be made possible. This'll also have the benefit of allowing HLE'd services to not constantly re-roll their own memory mapping code whenever a service would otherwise invoke an SVC instruction.

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -1257,6 +1257,74 @@ static ResultCode MapProcessCodeMemory(Core::System& system, Handle process_hand
     return vm_manager.MapCodeMemory(dst_address, src_address, size);
 }
 
+ResultCode UnmapProcessCodeMemory(Core::System& system, Handle process_handle, u64 dst_address,
+                                  u64 src_address, u64 size) {
+    LOG_DEBUG(Kernel_SVC,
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/svc.cpp
```diff
@@ -1189,6 +1189,142 @@ static ResultCode QueryMemory(Core::System& system, VAddr memory_info_address,
...
+              "src_address=0x{:016X}, size=0x{:016X}",
+              process_handle, dst_address, src_address, size);
+
+    if (!Common::Is4KBAligned(src_address)) {
+        LOG_ERROR(Kernel_SVC, "src_address is not page-aligned (src_address=0x{:016X}).",
+                  src_address);
+        return ERR_INVALID_ADDRESS;
+    }
+
+    if (!Common::Is4KBAligned(src_address)) {
```
Normmatt: Shouldn't this be dst_address?

lioncash: Indeed. Good catch.

PR #2397, opened 2019-04-14
kernel/thread: Remove unused guest_handle member variable

lioncash:
This member variable is entirely unused. It was only set but never actually utilized. Given that, we can remove it to get rid of noise in the thread interface.


FernandoS27:
To be honest, I use it to debug the Scheduler. In fact a thread's name should use that handle instead of the entry point.

lioncash:
Okay, but that doesn't require it to be an actual member in the thread's interface. We can just do that at creation.

FernandoS27:
Well we should add it to the wait tree. I normaly use the guest handle to check if some SVCs send the correct owner or waiting handle.

lioncash:
It's part of the name, so it'll show up in the wait tree. Also which softlock PR? If it's the mutex one, then it doesn't make use of it, as far as what the diff is showing me.

FernandoS27:
nvm then, go ahead.

PR #2400, opened 2019-04-15
Implement Kepler Memory on both Linear and BlockLinear.

FernandoS27:


src/video_core/engines/kepler_memory.h
```diff
@@ -63,6 +68,18 @@ class KeplerMemory final {
                         return static_cast<GPUVAddr>((static_cast<GPUVAddr>(address_high) << 32) |
                                                      address_low);
                     }
+
+                    u32 BlockWidth() const {
+                        return 1 << block_width.Value();
```
lioncash:
```suggestion
                        return 1U << block_width.Value();
```

otherwise this causes a sign-conversion. Ditto for the other two below.

src/video_core/engines/kepler_memory.cpp
```diff
@@ -10,12 +10,16 @@
...
+#include "video_core/textures/convert.h"
+#include "video_core/textures/decoders.h"
 
 namespace Tegra::Engines {
 
 KeplerMemory::KeplerMemory(Core::System& system, VideoCore::RasterizerInterface& rasterizer,
                            MemoryManager& memory_manager)
-    : system{system}, rasterizer{rasterizer}, memory_manager{memory_manager} {}
+    : system{system}, rasterizer{rasterizer}, memory_manager{memory_manager} {
+    state.inner_buffer.reserve(0x10000);
```
lioncash: You may want to leave a comment about this value, given it looks arbitrary (at least to me, anyways)

PR #2402, opened 2019-04-15
common: Apply the nodiscard attribute where applicable

lioncash:
Given the common library is supposed to be a general set of utilities, we can mark functions with the nodiscard attribute to allow the compiler to create warnings when functions have their return values discarded when they shouldn't be, identifying bugs quicker. This also takes the liberty of applying noexcept to some trivial cases as well (such as the bit arithmetic utilities, etc).

This is only done with the common library, given it's supposed to contain a multitude of functions that don't necessarily change quite frequently (e.g. applying it to other libraries would need to be done on a much more strict case-by-case basis).

I've kept the individual modifications within their own commits to make review slightly nicer, however I can squash these down at the end of review if that's preferable.

### DIFF
--- a/src/common/alignment.h
+++ b/src/common/alignment.h
@@ -8,25 +8,25 @@
 namespace Common {
 
 template <typename T>
-constexpr T AlignUp(T value, std::size_t size) {
+[[nodiscard]] constexpr T AlignUp(T value, std::size_t size) noexcept {
     static_assert(std::is_unsigned_v<T>, "T must be an unsigned value.");
     return static_cast<T>(value + (size - value % size) % size);
 }
 
 template <typename T>
-constexpr T AlignDown(T value, std::size_t size) {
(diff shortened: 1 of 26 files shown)


bunnei:
@lioncash clang-format errors are real.

bunnei:
@lioncash reminder to address above comment

bunnei:
@lioncash ping here

MysticExile:
You might want to rebase this.

lioncash:
This'll first need to have our current code style formatting looked at. Currently clang-format suggests extremely wonky formatting suggestions for some usages.

bunnei:
@lioncash any plan to do anything with this? If as a whole, you do not plan to finish this, are there parts that can be merged? Or should this just be closed? Thanks!

lioncash:
I may as well just close this for the time being. I can revisit this some other time

PR #2410, opened 2019-04-16
 kernel/svc: Reorganize and fix up the initial handling of svcSetThreadCoreMask()

lioncash:
Makes the code much nicer to follow in terms of behavior and control flow. It also fixes a few bugs in the implementation. This initial implementation predated the introduction of the process capability descriptor handling, so these are just a few things that fell through the cracks because of it.

Notably, the thread's owner process shouldn't be accessed in order to retrieve the core mask or ideal core. This should be done through the current running process. The only reason this bug wasn't encountered yet is because we currently only support running one process, and thus every owner process will be the current process.

We also weren't checking against the process' CPU core mask to see if an allowed core is specified or not.

With this out of the way, it'll be less noisy to implement proper handling of the affinity flags internally within the kernel thread instances.

### DIFF
--- a/src/core/hle/kernel/thread.h
+++ b/src/core/hle/kernel/thread.h
@@ -30,12 +30,21 @@ enum ThreadPriority : u32 {
 };
 
 enum ThreadProcessorId : s32 {
-    THREADPROCESSORID_IDEAL = -2, ///< Run thread on the ideal core specified by the process.
-    THREADPROCESSORID_0 = 0,      ///< Run thread on core 0
-    THREADPROCESSORID_1 = 1,      ///< Run thread on core 1
-    THREADPROCESSORID_2 = 2,      ///< Run thread on core 2
-    THREADPROCESSORID_3 = 3,      ///< Run thread on core 3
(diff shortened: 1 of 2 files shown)


src/core/hle/kernel/thread.h
```diff
@@ -30,12 +30,21 @@ enum ThreadPriority : u32 {
...
-    THREADPROCESSORID_2 = 2,      ///< Run thread on core 2
-    THREADPROCESSORID_3 = 3,      ///< Run thread on core 3
-    THREADPROCESSORID_MAX = 4,    ///< Processor ID must be less than this
+    /// Indicates that no particular processor core is preferred.
+    THREADPROCESSORID_DONT_CARE = -1,
+
+    /// Run thread on the ideal core specified by the process.
+    THREADPROCESSORID_IDEAL = -2,
+
+    /// Indicates the the preferred processor ID shouldn't be updated in
```
InusualZ: Duplicated `the`

bunnei:
@lioncash reminder to address above comment

PR #2411, opened 2019-04-16
GPU Manager: Implement ReadBlockUnsafe and WriteBlockUnsafe

FernandoS27:
This functions are mostly used when the memory at hand either is never submitted directly to the host GPU or it's directly being controlled by a Cache.

### DIFF
--- a/src/video_core/memory_manager.h
+++ b/src/video_core/memory_manager.h
@@ -67,6 +67,8 @@ public:
 
     void ReadBlock(GPUVAddr src_addr, void* dest_buffer, std::size_t size) const;
     void WriteBlock(GPUVAddr dest_addr, const void* src_buffer, std::size_t size);
+    void ReadBlockUnsafe(GPUVAddr src_addr, void* dest_buffer, std::size_t size) const;
+    void WriteBlockUnsafe(GPUVAddr dest_addr, const void* src_buffer, std::size_t size);
     void CopyBlock(GPUVAddr dest_addr, GPUVAddr src_addr, std::size_t size);
 
 private:
--- a/src/video_core/memory_manager.h
+++ b/src/video_core/memory_manager.h
@@ -65,11 +65,31 @@ public:
     u8* GetPointer(GPUVAddr addr);
     const u8* GetPointer(GPUVAddr addr) const;
 
-    void ReadBlock(GPUVAddr src_addr, void* dest_buffer, std::size_t size) const;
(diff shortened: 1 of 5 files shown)


src/video_core/memory_manager.h
```diff
@@ -67,6 +67,8 @@ class MemoryManager final {
 
     void ReadBlock(GPUVAddr src_addr, void* dest_buffer, std::size_t size) const;
     void WriteBlock(GPUVAddr dest_addr, const void* src_buffer, std::size_t size);
+    void ReadBlockUnsafe(GPUVAddr src_addr, void* dest_buffer, std::size_t size) const;
```
lioncash: There should probably be a comment by these declarations themselves indicating what makes them unsafe, and particularly *when* you'd use them over the other ones.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -38,13 +38,14 @@ GPUVAddr GetShaderAddress(Maxwell::ShaderProgram program) {
 }
 
 /// Gets the shader program code from memory for the specified address
-ProgramCode GetShaderCode(const u8* host_ptr) {
+ProgramCode GetShaderCode(const GPUVAddr gpu_addr, const u8* host_ptr) {
+    auto& memory_manager{Core::System::GetInstance().GPU().MemoryManager()};
```
lioncash: This should pass in the memory manager instead of using a global accessor.

FernandoS27:
Done.

src/video_core/memory_manager.h
bunnei: +10

src/video_core/memory_manager.h
```diff
@@ -65,9 +65,32 @@ class MemoryManager final {
     u8* GetPointer(GPUVAddr addr);
     const u8* GetPointer(GPUVAddr addr) const;
 
-    void ReadBlock(GPUVAddr src_addr, void* dest_buffer, std::size_t size) const;
-    void WriteBlock(GPUVAddr dest_addr, const void* src_buffer, std::size_t size);
-    void CopyBlock(GPUVAddr dest_addr, GPUVAddr src_addr, std::size_t size);
+    // Returns true if the block is continous in host memory, false otherwise
+    bool IsBlockContinous(const GPUVAddr start, const std::size_t size);
+
+    /*
```
bunnei: Should be `/**`, for docstrings

PR #2418, opened 2019-04-17
es: Implement various ticket accessor commands from IEticketService

DarkLordZach:
This adds:
- Full ticket processing and handling to the KeyManager. Previously we assumed tickets would use a RSA2048 signature scheme but that's not always correct. It now uses fields in a structure instead of offsets into an array.
- Adds support for saving/loading the ETicketExtendedRSAKey, significantly speeding up ticket processing.
- Implements some of the `es` service commands that count, list, and access tickets.
- Adds support to remake a title key and rights ID pair into a common ticket to present to an application.


src/core/crypto/key_manager.cpp
```diff
@@ -135,6 +208,28 @@ void KeyManager::DeriveGeneralPurposeKeys(std::size_t crypto_revision) {
     }
 }
 
+RSAKeyPair<2048> KeyManager::GetETicketRSAKey() {
+    if (eticket_extended_kek == std::array<u8, 576>{} || !HasKey(S128KeyType::ETicketRSAKek))
```
lioncash: This should really have its own function that checks the existing data. This allocates 576 bytes on the stack.

src/core/crypto/key_manager.h
```diff
@@ -30,7 +31,76 @@ constexpr u64 TICKET_FILE_TITLEKEY_OFFSET = 0x180;
...
+    u64 ticket_id;
+    u64 device_id;
+    std::array<u8, 0x10> rights_id;
+    u32 account_id;
+    INSERT_PADDING_BYTES(0x14C);
+};
+static_assert(sizeof(TicketData) == 0x2C0, "TicketData has incorrect size.");
+
+struct Ticket {
+    SignatureType sig_type;
```
lioncash: This should probably be using a std::variant  instead of a tag and an enum.

src/core/crypto/key_manager.h
```diff
@@ -43,6 +113,13 @@ struct RSAKeyPair {
     std::array<u8, 4> exponent;
 };
 
+template <size_t bit_size, size_t byte_size>
+bool operator==(const RSAKeyPair<bit_size, byte_size>& lhs,
```
lioncash: `operator!=` should also be provided.

src/core/crypto/key_manager.cpp
```diff
@@ -135,6 +208,28 @@ void KeyManager::DeriveGeneralPurposeKeys(std::size_t crypto_revision) {
...
+        return {};
+
+    const auto eticket_final = GetKey(S128KeyType::ETicketRSAKek);
+
+    std::vector<u8> extended_iv(0x10);
```
lioncash:
```suggestion
    std::vector<u8> extended_iv(eticket_extended_kek.begin(), eticket_extended_kek.begin() + 0x10);
```

Instead of allocating and then copying.

src/core/crypto/key_manager.cpp
```diff
@@ -135,6 +208,28 @@ void KeyManager::DeriveGeneralPurposeKeys(std::size_t crypto_revision) {
     }
 }
 
+RSAKeyPair<2048> KeyManager::GetETicketRSAKey() {
```
lioncash: This seems like it can be a const member function.

src/core/crypto/key_manager.cpp
```diff
@@ -886,19 +977,40 @@ void KeyManager::DeriveETicket(PartitionDataManager& data) {
...
+            personal_tickets[rights_id] = res[i];
+        }
+
         SetKey(S128KeyType::Titlekey, key, rights_id[1], rights_id[0]);
     }
 }
 
+void KeyManager::SynthesizeTickets() {
+    for (const auto& key : s128_keys) {
+        if (key.first.type == S128KeyType::Titlekey) {
```
lioncash:
```suggestion
        if (key.first.type != S128KeyType::Titlekey) {
            continue;
        }
```

src/core/crypto/key_manager.h
```diff
@@ -30,7 +31,76 @@ constexpr u64 TICKET_FILE_TITLEKEY_OFFSET = 0x180;
...
+            INSERT_PADDING_BYTES(0x40);
+            TicketData data;
+        } ecdsa;
+    };
+
+    TicketData& GetData();
+    const TicketData& GetData() const;
+    u64 GetSize() const;
+
+    static Ticket SynthesizeCommon(Key128 title_key, std::array<u8, 0x10> rights_id);
```
lioncash:
```suggestion
    static Ticket SynthesizeCommon(Key128 title_key, const std::array<u8, 0x10>& rights_id);
```

DarkLordZach:
Build failure unrelated to changes -- seems to be something with deps on msvc

PR #2419, opened 2019-04-17
lr: Add command handler skeletons for Open*LocationResolver

DarkLordZach:
- Implements LR service commands 0-2, which simply take no args and returns an object of the appropriate type.
- Rename LocationResolver to LR to prevent confusion with ILocationResolver.
- Add command names for ILocationResolver, IRegisteredLocationResolver, IAddOnContentLocationResolver.

### DIFF
--- a/src/core/hle/service/ncm/ncm.cpp
+++ b/src/core/hle/service/ncm/ncm.cpp
@@ -45,6 +45,25 @@ private:
     FileSys::StorageId storage;
 };
 
+class IRegisteredLocationResolver final : public ServiceFramework<IRegisteredLocationResolver> {
+public:
+    explicit IRegisteredLocationResolver() : ServiceFramework{"IRegisteredLocationResolver"} {
+        static const FunctionInfo functions[] = {
+            {0, nullptr, "ResolveProgramPath"},
+            {1, nullptr, "RegisterProgramPath"},
+            {2, nullptr, "UnregisterProgramPath"},
+            {3, nullptr, "RedirectProgramPath"},
+            {4, nullptr, "ResolveHtmlDocumentPath"},
+            {5, nullptr, "RegisterHtmlDocumentPath"},
+            {6, nullptr, "UnregisterHtmlDocumentPath"},
+            {7, nullptr, "RedirectHtmlDocumentPath"},
+            {8, nullptr, ""},
+        };
+
+        RegisterHandlers(functions);
+    }
+};
+
 public:
     explicit LocationResolver() : ServiceFramework{"lr"} {
 class LR final : public ServiceFramework<LR> {
@@ -72,6 +91,14 @@ private:
         IPC::ResponseBuilder rb{ctx, 2, 0, 1};
         rb.Push(RESULT_SUCCESS);
         rb.PushIpcInterface(std::make_shared<ILocationResolver>(id));
+    }
+
+    void OpenRegisteredLocationResolver(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Service_NCM, "called");
+
+        IPC::ResponseBuilder rb{ctx, 2, 0, 1};
+        rb.Push(RESULT_SUCCESS);
+        rb.PushIpcInterface(std::make_shared<IRegisteredLocationResolver>());
     }
 
 };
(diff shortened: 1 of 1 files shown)


src/core/hle/service/ncm/ncm.cpp
```diff
@@ -11,8 +11,79 @@
...
+            {7, nullptr, "ResolveApplicationLegalInformationPath"},
+            {8, nullptr, "RedirectApplicationLegalInformationPath"},
+            {9, nullptr, "Refresh"},
+            {10, nullptr, "RedirectProgramPath2"},
+            {11, nullptr, "Refresh2"},
+            {12, nullptr, "DeleteProgramPath"},
+            {13, nullptr, "DeleteApplicationControlPath"},
+            {14, nullptr, "DeleteApplicationHtmlDocumentPath"},
+            {15, nullptr, "DeleteApplicationLegalInformationPath"},
+            {16, nullptr, ""},
```
Hexagon12: *"Unknown1"* and so on to fit our current service command naming conventions.

src/core/hle/service/ncm/ncm.cpp
```diff
@@ -24,6 +95,39 @@ class LocationResolver final : public ServiceFramework<LocationResolver> {
...
     }
+
+private:
+    void OpenLocationResolver(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        const auto id = rp.PopRaw<FileSys::StorageId>();
+
+        LOG_DEBUG(Service_NCM, "called, id={:02X}", static_cast<u8>(id));
+
+        IPC::ResponseBuilder rb{ctx, 2, 0, 1};
```
lioncash: This does more than just create the location resolver. It creates different kinds of location resolvers based off the storage ID. In one case, this involves also calling out to ncm's OpenContentMetaDatabase and ncm's OpenContentStorage

src/core/hle/service/ncm/ncm.cpp
```diff
@@ -24,6 +95,39 @@ class LocationResolver final : public ServiceFramework<LocationResolver> {
...
+        rb.Push(RESULT_SUCCESS);
+        rb.PushIpcInterface(std::make_shared<ILocationResolver>(id));
+    }
+
+    void OpenRegisteredLocationResolver(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Service_NCM, "called");
+
```
lioncash: This is supposed to cache the created instance in a class variable so the allocation is only done once.

i.e.

```cpp
if (registered_location_resolver == nullptr) {
    registered_location_resolver = std::make_shared<IRegisteredLocationResolver>();
}

...

File: src/core/hle/service/ncm/ncm.cpp [hunk 4]
```diff
@@ -24,6 +95,39 @@ class LocationResolver final : public ServiceFramework<LocationResolver> {
...
+        IPC::ResponseBuilder rb{ctx, 2, 0, 1};
+        rb.Push(RESULT_SUCCESS);
+        rb.PushIpcInterface(std::make_shared<IRegisteredLocationResolver>());
+    }
+
+    void OpenAddOnContentLocationResolver(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        const auto id = rp.PopRaw<FileSys::StorageId>();
```
lioncash: This service function doesn't pass in a storage ID. It only has an out parameter.

src/core/hle/service/ncm/ncm.cpp
```diff
@@ -24,6 +95,39 @@ class LocationResolver final : public ServiceFramework<LocationResolver> {
...
+
+        LOG_DEBUG(Service_NCM, "called, id={:02X}", static_cast<u8>(id));
+
+        IPC::ResponseBuilder rb{ctx, 2, 0, 1};
+        rb.Push(RESULT_SUCCESS);
+        rb.PushIpcInterface(std::make_shared<IAddOnContentLocationResolver>(id));
```
lioncash: Ditto with regards to caching the created resolver.

src/core/hle/service/ncm/ncm.cpp
DarkLordZach: Disagree
If no name is provided the command ID/expected header will be used as name on the unimplemented function report, which is much more useful

PR #2420, opened 2019-04-18
service/audctl: Implement GetTargetVolumeMin() and GetTargetVolumeMax()

lioncash:
These two service functions are literally hardcoded to always return these values without any other error checking.

### DIFF
--- a/src/core/hle/service/audio/audctl.cpp
+++ b/src/core/hle/service/audio/audctl.cpp
@@ -2,6 +2,8 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include "common/logging/log.h"
+#include "core/hle/ipc_helpers.h"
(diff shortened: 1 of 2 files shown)


src/core/hle/service/audio/audctl.cpp
```diff
@@ -44,4 +46,28 @@ AudCtl::AudCtl() : ServiceFramework{"audctl"} {
 
 AudCtl::~AudCtl() = default;
 
+void AudCtl::GetTargetVolumeMin(Kernel::HLERequestContext& ctx) {
+    LOG_DEBUG(Audio, "called.");
+
+    // This service function is currently hardcoded on the
```
jroweboy: would be good to at least put the switch OS version this module was RE'd from so we know if something goes wonky here later because a game expects a newer version of the service.

PR #2424, opened 2019-04-20
Allow picking a Compatibility Profile for OpenGL.

FernandoS27:
Allowing the compatibility profile to be selected will ease our work on debugging yuzu as currently some games are fixed under it. We are currently researching the reason behind it.


lioncash:
This should probably have a description in the commit explaining why the option is being introduced

PR #2430, opened 2019-04-23
core: Implement FileSystemController to deglobalize FS services

DarkLordZach:
Previously, the Service::Filesystem namespace was a mess of global functions and variables to store and manage the state for the `fsp-srv` family of services, however as needs grew this namespace started to be used by things all across the core (for example, `am` needs access for save data sizes and ensuresavedata). This puts all of those functions into a class `FileSystemController` which will be managed by the system/core and passed into services that require it. Currently, this includes `pl:u`, `appletAE`, `appletOE`, and of course `fsp-srv`. 

This also adds some things to this class that may seem useless/unnecessary but are part of a later PR and made more sense here, these include:
- Accessors for various NAND and SDMC directories and partitions -- this are used by fsp-srv commands in an upcoming PR
- The `PlaceholderCache` which is responsible for managing the /Contents/placehld directory which stores temporary partially installed game data while it is being installed -- this is used by `ncm` in an upcoming PR
- Accessors for the gamecard and additional functions in XCI for data -- this is used by fsp-srv/IDeviceOperator commands in an upcoming PR
- Some bug fixes I found that were related to this while working, such as:
    - Bug where filenames less than 9 characters in /Contents would hard crash
    - Bug where homebrew with data in the 3 LSB nibbles of the title ID would make patch manager erratic.
    - Some short-circuit logic to patch manager to speed load times marginally
    - Glitch where an invalid title ID would crash patch manager
- Implementation of creation of save data, and the removal of a hack where save data would be auto created upon open. Applications must now use the proper channels (`EnsureSaveData`, `CreateSaveData`) to create it. The current impl in this PR for fsp-srv save data creation is quite sparse and will be improved in a separate PR.

PR #2431, opened 2019-04-23
yuzu: Implement a caching mechanism for the game list

DarkLordZach:
Currently, the game list can drag (for up to 10 seconds) with a large collection of games and/or add-ons. This is because the current algorithm searches the entire list of addons for each game when populating the names for the "Add-Ons" column (which is terribly inefficient and unnecessary). This adds a simple cache that writes the list of addon names and versions to a text file in yuzu_dir/cache/game_list. This will also cache the icons and names of titles, as after add-ons those are the next two biggest time consumers.

I tested this for time difference in two different configurations (one 'light' with about 5 games + 3 updates and one 'heavy' with about 30 games + 20 updates/dlc) In the light the load time went from about 4 seconds to sub 1 second and in the heavy the load time went from about 17 seconds to about 2.

The UI to fully manage this is part of PR #2430.

### DIFF
--- a/src/yuzu/game_list_worker.cpp
+++ b/src/yuzu/game_list_worker.cpp
@@ -9,6 +9,7 @@
 
 #include <QDir>
 #include <QFileInfo>
+#include <QSettings>
 
 #include "common/common_paths.h"
 #include "common/file_util.h"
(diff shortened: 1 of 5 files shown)


src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
 namespace {
+
+template <typename T>
+T GetGameListCachedObject(const std::string& filename, const std::string& ext,
+                          const std::function<T()>& generator);
+
+template <>
+QString GetGameListCachedObject(const std::string& filename, const std::string& ext,
+                                const std::function<QString()>& generator) {
+    if (!UISettings::values.cache_game_list || filename == "0000000000000000")
```
lioncash: This should be braced

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+    const auto& path = FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP + "game_list" +
+                       DIR_SEP + filename + "." + ext;
+
+    FileUtil::CreateFullPath(path);
+
+    if (!FileUtil::Exists(path)) {
+        const auto str = generator();
+
+        std::ofstream stream(path);
+        if (stream)
```
lioncash: Ditto

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+            stream << str.toStdString();
+
+        stream.close();
```
lioncash: This can be omitted. The stream will be closed automatically.

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+        return str;
+    }
+
+    std::ifstream stream(path);
+
+    if (stream) {
+        const std::string out(std::istreambuf_iterator<char>{stream},
+                              std::istreambuf_iterator<char>{});
+        stream.close();
```
lioncash: Ditto

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+T GetGameListCachedObject(const std::string& filename, const std::string& ext,
+                          const std::function<T()>& generator);
+
+template <>
+QString GetGameListCachedObject(const std::string& filename, const std::string& ext,
+                                const std::function<QString()>& generator) {
+    if (!UISettings::values.cache_game_list || filename == "0000000000000000")
+        return generator();
+
+    const auto& path = FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP + "game_list" +
```
lioncash:
```suggestion
    const auto path = FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP + "game_list" +
```

This is holding a completely new string, not a reference.

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+                       DIR_SEP + filename + "." + ext;
```
lioncash:
```suggestion
                       DIR_SEP + filename + '.' + ext;
```

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+}
+
+template <>
+std::pair<std::vector<u8>, std::string> GetGameListCachedObject(
+    const std::string& filename, const std::string& ext,
+    const std::function<std::pair<std::vector<u8>, std::string>()>& generator) {
+    if (!UISettings::values.cache_game_list || filename == "0000000000000000")
+        return generator();
+
+    const auto& path1 = FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP +
```
lioncash:
```suggestion
    const auto path1 = FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP +
```

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+                        "game_list" + DIR_SEP + filename + ".jpeg";
+    const auto& path2 = FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP +
```
lioncash:
```suggestion
    const auto path2 = FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP +
```

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+
+        FileUtil::IOFile file1(path1, "wb");
+        file1.Resize(icon.size());
+        file1.WriteBytes(icon.data(), icon.size());
+
+        std::ofstream stream2(path2, std::ios::out);
+        if (stream2)
+            stream2 << nacp;
+
+        file1.Close();
```
lioncash: These don't need to be explicitly closed.

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+                        "game_list" + DIR_SEP + filename + ".jpeg";
+    const auto& path2 = FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP +
+                        "game_list" + DIR_SEP + filename + ".appname.txt";
+
+    FileUtil::CreateFullPath(path1);
+
+    if (!FileUtil::Exists(path1) || !FileUtil::Exists(path2)) {
+        const auto [icon, nacp] = generator();
+
+        FileUtil::IOFile file1(path1, "wb");
```
lioncash: The file, resizing, and writing, are all operations that can potentially fail.

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+    FileUtil::IOFile file1(path1, "rb");
+    std::ifstream stream2(path2);
+
+    std::vector<u8> vec(file1.GetSize());
+    file1.ReadBytes(vec.data(), vec.size());
+
+    if (stream2 && !vec.empty()) {
+        const std::string out(std::istreambuf_iterator<char>{stream2},
+                              std::istreambuf_iterator<char>{});
+        stream2.close();
```
lioncash: This doesn't need to be explicitly closed.

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+        std::ofstream stream2(path2, std::ios::out);
+        if (stream2)
+            stream2 << nacp;
+
+        file1.Close();
+        stream2.close();
+        return std::make_pair(icon, nacp);
+    }
+
+    FileUtil::IOFile file1(path1, "rb");
```
lioncash: Both this and the following stream can fail to open.

src/yuzu/game_list_worker.cpp
```diff
@@ -30,13 +31,101 @@
...
+}
+
 void GetMetadataFromControlNCA(const FileSys::PatchManager& patch_manager, const FileSys::NCA& nca,
                                std::vector<u8>& icon, std::string& name) {
-    auto [nacp, icon_file] = patch_manager.ParseControlNCA(nca);
-    if (icon_file != nullptr)
-        icon = icon_file->ReadAllBytes();
-    if (nacp != nullptr)
-        name = nacp->GetApplicationName();
+    auto res = GetGameListCachedObject<std::pair<std::vector<u8>, std::string>>(
```
lioncash: This can still use structured bindings, can it not?

DarkLordZach:
Build failure unrelated to changes

lioncash:
Note that you should probably use Qt's facilities over the ones provided by file_util where applicable. You avoid needing to convert between string formats that way and keep interactions consistent.

src/yuzu/main.cpp
```diff
@@ -1396,6 +1396,9 @@ void GMainWindow::OnMenuInstallToNAND() {
                                  tr("The file was successfully installed."));
         game_list->PopulateAsync(UISettings::values.game_directory_path,
                                  UISettings::values.game_directory_deepscan);
+        // Clear the game list cache.
```
lioncash: This doesn't really add any value. If you need to say what you're doing rather than why it needs to be done, it's probably better expressed through putting it in a named function.

PR #2434, opened 2019-04-26
Disable Fast Depth->Color Copies in Turing+ GPUs.

FernandoS27:


src/video_core/renderer_opengl/gl_rasterizer_cache.cpp
```diff
@@ -872,7 +872,7 @@ void CachedSurface::UpdateSwizzle(Tegra::Texture::SwizzleSource swizzle_x,
 }
 
 RasterizerCacheOpenGL::RasterizerCacheOpenGL(RasterizerOpenGL& rasterizer)
-    : RasterizerCache{rasterizer} {
+    : RasterizerCache{rasterizer}, rasterizer{rasterizer} {
```
ReinUsesLisp: Pass the `OpenGL::Device` instance and use that instead of having a getter in the rasterizer.

src/video_core/renderer_opengl/gl_rasterizer.h
```diff
@@ -73,6 +73,10 @@ class RasterizerOpenGL : public VideoCore::RasterizerInterface {
     static_assert(MaxConstbufferSize % sizeof(GLvec4) == 0,
                   "The maximum size of a constbuffer must be a multiple of the size of GLvec4");
 
+    const Device& GetDevice() const {
```
ReinUsesLisp: If the above code is addressed, this would be unused.

src/video_core/renderer_opengl/gl_device.h
```diff
@@ -20,6 +21,10 @@ class Device {
         return has_variable_aoffi;
     }
 
+    bool IsTuringGPU() const {
+        return GLAD_GL_NV_mesh_shader;
```
ReinUsesLisp: This is mixing global state with an instance. IMO it's better to test for `GLAD_GL_NV_mesh_shader` the constructor and store it in a member variable.

ReinUsesLisp: Also add a commentary on why `GL_NV_mesh_shader` is used to test for Turing GPUs and encapsulate it in a static function like `TestVariableAoffi`.

bunnei:
@FernandoS27 can you address remaining comments so we can merge?

src/video_core/renderer_opengl/gl_device.h
ReinUsesLisp: > 
> 
> Also add a commentary on why `GL_NV_mesh_shader` is used to test for Turing GPUs and encapsulate it in a static function like `TestVariableAoffi`.

Changed my mind about this one. Just check for the `GL_NV_mesh_shader` in the constructor.

Hexagon12:
Please rebase the PR so we could tag it again for Canary.

bunnei:
@FernandoS27 ping to rebase

bunnei:
@FernandoS27 ping here

PR #2438, opened 2019-04-29
video_core: Implement texture buffers and SUST

ReinUsesLisp:
~~First of all **do not merge this PR** at the moment to any branch.~~

~~This invalidates the transferable disk shader cache. To reduce the number of invalidations, (in case that it is merged) it has to be merged at the same time as a following PR that also invalidates.~~

This implements texture buffers. The main issue with texture buffers is that they are aliased with 1D textures in shaders, so we can determine from the shader program if a sampler is from a 1D texture or a texture buffer. To solve this problem, all potential texture buffer operations and declarations are behind a preprocessor macro in GLSL code that has to be declared to use a texture buffer for that sampler instead of a 1D texture.

As an example:
```glsl
#ifdef SAMPLER_0_IS_BUFFER
float tmp32 = texelFetch(sampler_0_vs, ...);
#else
float tmp32 = texelFetch(sampler_0_vs, ..., lod); // < not available in texture buffers
#endif
gpr_whatever = tmp32;
```
Then from the rasterizer the samplers that are used as texture buffers are informed to the shader cache as different variants. This is why the disk shader cache has to be invalidated, texture buffer samplers have to be stored.

Texture buffers are emulated with texture buffers, 1D textures are emulated with 1D textures.

About texture cache changes: this uses immutable buffer storage, so now `ARB_buffer_storage` is a required extension. This change is reflected in `gl_stream_buffer`.

The Maxwell instruction `TLD` has been implemented since it is used by nouveau to implement texture fetches, used by the example used for this implementation:
[texture-buffer.zip](https://github.com/yuzu-emu/yuzu/files/3125718/texture-buffer.zip)

### DIFF
--- a/CMakeModules/GenerateSCMRev.cmake
+++ b/CMakeModules/GenerateSCMRev.cmake
@@ -70,6 +70,7 @@ set(HASH_FILES
     "${VIDEO_CORE}/shader/decode/half_set.cpp"
     "${VIDEO_CORE}/shader/decode/half_set_predicate.cpp"
     "${VIDEO_CORE}/shader/decode/hfma2.cpp"
+    "${VIDEO_CORE}/shader/decode/image.cpp"
     "${VIDEO_CORE}/shader/decode/integer_set.cpp"
     "${VIDEO_CORE}/shader/decode/integer_set_predicate.cpp"
     "${VIDEO_CORE}/shader/decode/memory.cpp"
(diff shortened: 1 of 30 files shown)


ReinUsesLisp:
I changed my mind and now this PR includes the shader implementation of `SUST`.

`SUST` is the backend instruction of `imageStore`. I've tested this implementation with a compilation decompilation step.

greggameplayer:
@ReinUsesLisp big regression on The Lego Ninjago Movie Videogame on the 3d models of the menu : 
image

here is the full log : https://gist.github.com/greggameplayer/a243af25157db73f174a07c601a15933

ReinUsesLisp:
@greggameplayer Can you post how the game looks in latest Canary?

greggameplayer:
@ReinUsesLisp in the latest canary the game looks the same so it isn't because of this PR but because of another one

ReinUsesLisp:
This is frozen in favour of a texture cache rewrite.

ReinUsesLisp:
This PR is included in #2601.

PR #2441, opened 2019-04-30
shader: Implement AL2P and ALD.PHYS

ReinUsesLisp:
`AL2P` and `ALD.PHYS` is used to access attributes (and varyings) indirectly, code like this:
```glsl
in vec4 my_attributes[2];
uniform int index;
void main() {
    gl_Position = my_attributes[index];
}
```
Emits:
```
        /*0028*/                   AL2P R1, R0, 0x80;
        /* .. */
        /*0048*/                   ALD.PHYS R1, a[R1];
```

`AL2P` is used to return a physical address with an offset, for example: `AL2P R1, R0, 0x80` stores in `R1` the address of `R0 + 0x80`. This value can then be used to access the attribute indirectly: `ALD.PHYS R1, a[R1]`.
To implement this I made `AL2P` return a table physical address, for example `AL2P R1, R0, 0x80` will exactly return `R0 + 0x80`. This is a fake address since the memory isn't actually there but we can't do better on GLSL without analysing shader's intention.

Then `ALD.PHYS` calls a GLSL defined function that switches across all attributes from its address, returning zero on failure. For now only generic attributes are implemented, so dynamically written `gl_ClipDistance[]` and hand-written shaders (e.g. not aligned values) won't work, this can't be checked at compile time since it's an indirect read, and GLSL lacks traps to signal the error at runtime. Geometry shader physical attributes reads and physical attribute writes (on all stages) are not implemented either.

This also removes the unused attribute input mode tracking.

As a side note, fragment writes are done with branches on nouveau and local memory writes + reads on Nvidia's compiler.

- Used by Donkey Kong Country: Tropical Freeze.

### DIFF
--- a/src/video_core/shader/decode/memory.cpp
+++ b/src/video_core/shader/decode/memory.cpp
@@ -253,7 +253,7 @@ u32 ShaderIR::DecodeMemory(NodeBlock& bb, u32 pc) {
         SetRegister(bb, instr.gpr0, fake_address);
 
         // Signal the shader IR to declare all possible attributes and varyings
-        use_physical_attributes = true;
+        uses_physical_attributes = true;
         break;
     }
     default:
--- a/src/video_core/shader/decode/memory.cpp
+++ b/src/video_core/shader/decode/memory.cpp
@@ -49,6 +49,9 @@ u32 ShaderIR::DecodeMemory(NodeBlock& bb, u32 pc) {
                              "Indirect attribute loads are not supported");
         UNIMPLEMENTED_IF_MSG((instr.attribute.fmt20.immediate.Value() % sizeof(u32)) != 0,
                              "Unaligned attribute loads are not supported");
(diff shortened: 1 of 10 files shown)


src/video_core/shader/decode/memory.cpp
```diff
@@ -12,6 +12,8 @@
 #include "video_core/engines/shader_bytecode.h"
 #include "video_core/shader/shader_ir.h"
 
+#pragma optimize("", off)
```
lioncash: Unintentionally left in?

ReinUsesLisp: Yes.

FernandoS27:
You should add IPA.IDX as well.

ReinUsesLisp:
Have you seen any game using it?

FernandoS27:
Yeah some fragment shaders in donkey kong use them.

ReinUsesLisp:
Added `IPA.IDX`.

PR #2442, opened 2019-05-01
Fix Layered ASTC Textures

FernandoS27:


lioncash:
The commit should have a brief description saying what was wrong that needed to be fixed in particular. People shouldn't need to read code to get a basic overview of a change.

FernandoS27:
done.

PR #2450, opened 2019-05-04
CMakeLists: Explicitly specify -Wall for the non-MSVC case

lioncash:
Ensures that -Wall is always active as a compilation flag regardless of build type.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -37,7 +37,10 @@ if (MSVC)
     set(CMAKE_EXE_LINKER_FLAGS_DEBUG   "/DEBUG /MANIFEST:NO" CACHE STRING "" FORCE)
     set(CMAKE_EXE_LINKER_FLAGS_RELEASE "/DEBUG /MANIFEST:NO /INCREMENTAL:NO /OPT:REF,ICF" CACHE STRING "" FORCE)
 else()
-    add_compile_options("-Wno-attributes")
+    add_compile_options(
+        -Wall
+        -Wno-attributes
+    )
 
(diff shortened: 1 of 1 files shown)


wwylele:
Since I plan to do this for citra as well, I want to some thought on this. I think this should be gated behind a cmake flag, disabled as default, and only turn it on selective CI build. Reasons are:
- warnings can be unstable amomg different compiler version. Right now my local gcc 8 reports more warnings than gcc 7 on CI, and some of them are false positive. It is better to only test this against a frozen compiler version.
- warning as error can discourage fast prototyping. When implementing new features, warning as error can be a distraction and interrupt one's work flow. This depends on one's work style, and people should be able to choose the cmake flag depending on their preference.

lioncash:
> gated behind a cmake flag, disabled as default

I don't agree at all. Warnings are there to catch bugs, they should be opt-out, not opt-in.

>  and some of them are false positive.

-Wall is supposed to enable warnings that aren't false-positive, so this isn't relevant.

> warning as error

This isn't enabling warnings as errors, so this isn't relevant either.

I'm not going to gate warnings behind a flag, especially when said flags quite literally found bugs the moment they were enabled. If one's prototyping involves slipshod warning-heavy code, then that's a personal issue and shouldn't actively be incorporated into the project as a build option. CMake already allows modifying flags on the command line with CMAKE_CXX_FLAGS. If you don't want warnings, then disable them by specifying the necessary flag through that.

wwylele:
@lioncash Sorry I misread the entire thing. You are right.

PR #2452, opened 2019-05-07
Correct possible error on Rasterizer Caches

FernandoS27:
There was a weird bug that could happen if the object died directly and 
the cache address wasn't stored.


src/video_core/rasterizer_cache.h
```diff
@@ -147,8 +147,9 @@ class RasterizerCache : NonCopyable {
 
         object->SetIsRegistered(false);
         rasterizer.UpdatePagesCachedCount(object->GetCpuAddr(), object->GetSizeInBytes(), -1);
+        CacheAddr addr = object->GetCacheAddr();
```
lioncash:
```suggestion
        const CacheAddr addr = object->GetCacheAddr();
```

PR #2455, opened 2019-05-09
configuration/config: Move config loading and saving to functions based off groups

lioncash:
Over time our config values have grown quite numerous in size. Unfortunately it also makes the single functions we have for loading and saving values more error prone.

For example, we were loading the core settings twice when they only should have been loaded once. In another section, a variable was shadowing another variable used to load settings from a completely different section.

Finally, in one other case, there was an extraneous endGroup() call used that didn't need to be done. This was essentially dead code and also a bug waiting to happen.

This separates the section loading code into its own separate functions. This keeps variables only visible to the code that actually needs it, and makes it much easier to visually see the end of each individual configuration group. It also makes it much easier to visually catch bugs during code review.

While we're at it, this also uses QStringLiteral instead of raw string literals, which both avoids constructing a lot of QString instances, but also makes it much easier to disable implicit ASCII to QString and vice-versa in the future via setting QT_NO_CAST_FROM_ASCII and QT_NO_CAST_TO_ASCII as compilation flags.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -230,55 +230,64 @@ void Config::ReadPlayerValues() {
     for (std::size_t p = 0; p < Settings::values.players.size(); ++p) {
         auto& player = Settings::values.players[p];
 
-        player.connected = ReadSetting(QString("player_%1_connected").arg(p), false).toBool();
+        player.connected =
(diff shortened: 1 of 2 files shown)


lioncash:
Failure is related to uploading, not the code itself.

PR #2464, opened 2019-05-11
Fix OpenGL compatibility profile for mesa/Linux

z0z0z:
The  `Use OpenGL compatibility profile` checkbox currently does not work on Linux/mesa, probably due to `QSurfaceFormat::CompatibilityProfile` in upstream Qt not working properly.

So this PR does actually make PLG render correctly on Linux/mesa using AMD, without a user manually setting the environment variable.

Obviously I'm joking by putting this in main(), but I tried putting this in bootmanager.cpp where you actually check the setting from Yuzu, but it doesn't work.

```
    if (Settings::values.use_compatibility_profile) {
        fmt.setProfile(QSurfaceFormat::CompatibilityProfile);
#ifdef __linux__
        setenv("MESA_GL_VERSION_OVERRIDE", "4.5COMPAT", 0);
#endif
    } else {
        fmt.setProfile(QSurfaceFormat::CoreProfile);
    }
```

Your discord's development channel is locked so it's not like I could talk about this there.

Can someone figure out where is best to put this environment variable, and under whatever `#ifdef __linux__` and other conditional stuff you'd like?

Should upstream Qt be fixed? Sure I agree. waitingskeleton.jpg

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2129,6 +2129,11 @@ int main(int argc, char* argv[]) {
     // generating shaders
     setlocale(LC_ALL, "C");
 
+    // Environment variables
+#ifdef __linux__
+    setenv("MESA_GL_VERSION_OVERRIDE", "4.5COMPAT", 0);
+#endif
+
     GMainWindow main_window;
     // After settings have been loaded by GMainWindow, apply the filter
     main_window.show();


FernandoS27:
Send a message to a mod and he'll give you access to #development Now I'm not sure if setenv is safe for portability. I'll have to consult.

src/yuzu/main.cpp
```diff
@@ -2113,6 +2113,9 @@ void GMainWindow::SetDiscordEnabled([[maybe_unused]] bool state) {
 #endif
 
 int main(int argc, char* argv[]) {
+#ifdef __linux__
+    setenv("MESA_GL_VERSION_OVERRIDE", "4.5COMPAT", 1);
```
ReinUsesLisp: What if someone wants to use OpenGL core?

z0z0z:
I actually tried to make this not too ugly with the latest ammend, I put it under where you do setlocale in main().

I also used `0` as the third argument, which means it doesn't override environment variables already set, so if someone wanted to use the core profile, they could set it themselves.

To be honest, I was just joking by making this PR and putting it in main(), but maybe setting environment variables really won't work elsewhere?

src/yuzu/main.cpp
```diff
@@ -2129,6 +2129,11 @@ int main(int argc, char* argv[]) {
     // generating shaders
     setlocale(LC_ALL, "C");
 
+    // Environment variables
+#ifdef __linux__
+    setenv("MESA_GL_VERSION_OVERRIDE", "4.5COMPAT", 0);
```
ReinUsesLisp: Pointing it out again so it doesn't get lost. What if someone wants to use OpenGL core?

z0z0z:
> Pointing it out again so it doesn't get lost. What if someone wants to use OpenGL core from the UI?

That's really the whole problem in the first place.

The OpenGL compatibility profile checkbox in the settings currently doesn't do anything on Linux with mesa (Nvdia too?)

That's why I originally tried setting it in bootmanager.cpp where you check that setting, but setting environment variables in that file doesn't work, it must be too late in the program's execution.

ReinUsesLisp:
It has to be set before the OpenGL window is created.

alinach120602:
Can't you check from the QTConfig and make an ifdev Linux and set the according variable?

z0z0z:
> Can't you check the QTConfig for use_ogl_compat and make an ifdev Linux when true and set the according variable?

I tried that in main, didn't work, I guess because settings aren't loaded yet.

```
#ifdef __linux__
    if (Settings::values.use_compatibility_profile) {
        setenv("MESA_GL_VERSION_OVERRIDE", "4.5COMPAT", 0);
    }
#endif
````

Problem is you need this before the OpenGL window like Rein said, but I think settings aren't loaded yet so you can't check the setting.

I don't necessarily see the problem with using the OpenGL compat profile by default on mesa, because it means that Yuzu is using deprecated OpenGL features that require it, so it should be the default until it doesn't anymore right? As far as I understand, the compat profile doesn't hurt anything it just allows deprecated OpenGL features.

DarkLordZach:
Settings are loaded upon construction of the main window for opengl, so you should be ok.

z0z0z:
> Settings are loaded upon construction of the main window for opengl, so you should be ok.

Setting the environment variable after `GMainWindow main_window;` in main() doesn't work.

Trying to use `if (Settings::values.use_compatibility_profile)` before `GMainWindow main_window;` doesn't work, because settings aren't loaded.

z0z0z:
Fixed in https://github.com/yuzu-emu/yuzu/pull/2476

PR #2467, opened 2019-05-14
video_core/gpu_thread: Remove redundant copy constructor for CommandDataContainer

lioncash:
std::move within a copy constructor (on a data member that isn't mutable) will always result in a copy. Because of that, the behavior of this copy constructor is identical to the one that would be generated automatically by the compiler, so we can remove it.


lioncash:
Failure is related to a timeout on Appveyor

PR #2468, opened 2019-05-14
yuzu: Remove explicit types from locks where applicable

lioncash:
With C++17's deduction guides, the type doesn't need to be explicitly specified within locking primitives anymore. Previously the types were removed from locks, but a few stragglers have crept in over time.


lioncash:
Failure is related to a timeout on Appveyor.

PR #2472, opened 2019-05-14
maxwell_3d: reduce severity of different component formats assert.

FernandoS27:
This was reduced due to happening on most games and at such constant
rate that it affected performance heavily for the end user. In general,
we are well aware of the assert and an implementation is already
planned.


lioncash:
Missing a rationale on why it was reduced. This provides no context.

FernandoS27:
Done.

PR #2474, opened 2019-05-15
CMake: Get Git submodule dependencies via CMake

FreddyFunk:
This simplifies the building process a bit. Now it does not matter if the yuzu git repo was cloned recursively or if the git client did not initialize the git submodules before running cmake.

If this PR gets accepted we could shorten the relating wiki articles.


MerryMage:
If one wants to do this, one should remove the `check_submodules_present` function because this makes that redundant.

FreddyFunk:
Comments addressed and travis linux-mingw enviroment fixed

PR #2476, opened 2019-05-17
yuzu/bootmanager: Explicitly enable deprecated OpenGL features on compat

ReinUsesLisp:
Nvidia's proprietary driver creates a real OpenGL compatibility profile without this option, meanwhile Intel (and probably AMD, I haven't tested it) require that QSurfaceFormat::FormatOption::DeprecatedFunctions is explicitly enabled.

This might supersede #2464 since it should also fix mesa drivers.


z0z0z:
@ReinUsesLisp This works on my RX580/mesa.

Feel free to close my PR whenever.

Issue #2248, opened 2019-03-16
#2054 broke SDL/yuzu-cmd graphics rendering


FearlessTobi:
https://github.com/yuzu-emu/yuzu/pull/2054 broke graphics rendering in SDL/yuzu-cmd.
In all subsequent builds, graphics won't render at all with this Frontend.
This seems too be caused by MakeCurrent/DoneCurrent being called in the wrong places.

The log doesn't show anything interesting.

FearlessTobi:
Fixed by #2477.

PR #2477, opened 2019-05-17
yuzu_cmd: Make OpenGL's context current

ReinUsesLisp:
The SDL2 frontend never bound the OpenGL context, resulting on a white screen and no-ops all over the backend.


Hexagon12:
Does this fix #2248?

PR #2482, opened 2019-05-18
core: Add detailed local reporting feature for development

DarkLordZach:
Pretty much self-explanatory, but when any of the following events occur:
- `svcBreak()` call
- `fatal:u` `ThrowFatal*` call
- Unimplemented applet or service function
- Error applet is invoked
- `prepo:u` `SavePlayReport*` call

yuzu will save a nice JSON formatted report to the log dir if the config option is enabled.
The JSON report includes:
- Backtrace (with symbol names and offsets into modules and module names)
- Processor State (Registers, PC, SP, etc)
- Yuzu Version Information (Build Data and SCM data)
- Timestamp, title ID, and type
- Additional data generated by the specific type of report (i.e. applet ID and data for unimplemented applet, command buffer for unimplemented function, prepo data for prepo)

There is a UI option to enable this, but it will not persists (i.e. it doesn't save). This is because while it may be useful once in a while for a dev to get reports from a user (esp if they can't reproduce a bug), there is *no* reason for a normal user to have this enabled.
Developers can enable this permanently by manually adding the config key to qt-config.ini or sdl2-config.ini. (`reporting_services` in `Debugging` group)

This also adds symbol dumping to backtrace output in logs. I couldn't find a good demangler lib that is compatible with GPL and not a pain to integrate with so it just outputs the mangled name.

I added a function to the Reporter called `SaveUserReport` this is intended for devs during the development process. If this is invoked anywhere in yuzu it will trigger a full data dump to disk with all the data above, which may be useful for seeing what in the game calls what.

### DIFF
--- a/src/core/arm/arm_interface.cpp
+++ b/src/core/arm/arm_interface.cpp
@@ -9,6 +9,7 @@
 #include "common/logging/log.h"
 #include "core/arm/arm_interface.h"
 #include "core/core.h"
+#include "core/loader/loader.h"
 #include "core/memory.h"
 
 namespace Core {
@@ -80,15 +81,17 @@ Symbols GetSymbols(VAddr text_offset) {
         const auto value = Memory::Read64(dynamic_index + 0x8);
         dynamic_index += 0x10;
 
-        if (tag == ELF_DYNAMIC_TAG_NULL)
+        if (tag == ELF_DYNAMIC_TAG_NULL) {
             break;
+        }
 
-        if (tag == ELF_DYNAMIC_TAG_STRTAB)
+        if (tag == ELF_DYNAMIC_TAG_STRTAB) {
             string_table_offset = value;
-        else if (tag == ELF_DYNAMIC_TAG_SYMTAB)
+        } else if (tag == ELF_DYNAMIC_TAG_SYMTAB) {
             symbol_table_offset = value;
-        else if (tag == ELF_DYNAMIC_TAG_SYMENT)
+        } else if (tag == ELF_DYNAMIC_TAG_SYMENT) {
             symbol_entry_size = value;
+        }
     }
 
     if (string_table_offset == 0 || symbol_table_offset == 0 || symbol_entry_size == 0) {
@@ -126,8 +129,10 @@ std::optional<std::string> GetSymbolName(const Symbols& symbols, VAddr func_addr
             return func_address >= symbol.value && func_address < end_address;
         });
 
-    if (iter == symbols.end())
+    if (iter == symbols.end()) {
         return std::nullopt;
+    }
+
     return iter->second;
 }
 
@@ -150,7 +155,12 @@ std::vector<ARM_Interface::BacktraceEntry> ARM_Interface::GetBacktrace() const {
         fp = Memory::Read64(fp);
     }
 
-    const auto& modules{System::GetInstance().GetRegisteredNSOModules()};
+    std::map<VAddr, std::string> modules;
+    auto& loader{System::GetInstance().GetAppLoader()};
+    if (loader.ReadNSOModules(modules) != Loader::ResultStatus::Success) {
+        return {};
+    }
+
     std::map<std::string, Symbols> symbols;
     for (const auto& module : modules) {
         symbols.insert_or_assign(module.second, GetSymbols(module.first));
@@ -158,7 +168,8 @@ std::vector<ARM_Interface::BacktraceEntry> ARM_Interface::GetBacktrace() const {
 
     for (auto& entry : out) {
         VAddr base = 0;
-        for (const auto& module : modules) {
+        for (auto iter = modules.rbegin(); iter != modules.rend(); ++iter) {
+            const auto& module{*iter};
             if (entry.original_address >= module.first) {
                 entry.module = module.second;
                 base = module.first;
@@ -191,7 +202,7 @@ void ARM_Interface::LogBacktrace() const {
     LOG_ERROR(Core_ARM, "Backtrace, sp={:016X}, pc={:016X}", sp, pc);
     LOG_ERROR(Core_ARM, "{:20}{:20}{:20}{:20}{}", "Module Name", "Address", "Original Address",
               "Offset", "Symbol");
-    LOG_ERROR(Core_ARM, "{}", std::string(100, '-'));
+    LOG_ERROR(Core_ARM, "");
 
     const auto backtrace = GetBacktrace();
     for (const auto& entry : backtrace) {
--- a/src/core/arm/arm_interface.cpp
+++ b/src/core/arm/arm_interface.cpp
@@ -2,26 +2,202 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <map>
+#include <optional>
+#include "common/bit_field.h"
 #include "common/common_types.h"
 #include "common/logging/log.h"
 #include "core/arm/arm_interface.h"
+#include "core/core.h"
 #include "core/memory.h"
 
 namespace Core {
-void ARM_Interface::LogBacktrace() const {
-    VAddr fp = GetReg(29);
-    VAddr lr = GetReg(30);
-    const VAddr sp = GetReg(13);
-    const VAddr pc = GetPC();
 
-    LOG_ERROR(Core_ARM, "Backtrace, sp={:016X}, pc={:016X}", sp, pc);
+namespace {
+
+constexpr u64 ELF_DYNAMIC_TAG_NULL = 0;
+constexpr u64 ELF_DYNAMIC_TAG_STRTAB = 5;
+constexpr u64 ELF_DYNAMIC_TAG_SYMTAB = 6;
+constexpr u64 ELF_DYNAMIC_TAG_SYMENT = 11;
+
+enum class ELFSymbolType : u8 {
+    None = 0,
+    Object = 1,
+    Function = 2,
+    Section = 3,
+    File = 4,
+    Common = 5,
+    TLS = 6,
+};
+
+enum class ELFSymbolBinding : u8 {
+    Local = 0,
+    Global = 1,
+    Weak = 2,
+};
+
+enum class ELFSymbolVisibility : u8 {
+    Default = 0,
+    Internal = 1,
+    Hidden = 2,
+    Protected = 3,
+};
+
+struct ELFSymbol {
+    u32 name_index;
+    union {
+        u8 info;
+
+        BitField<0, 4, ELFSymbolType> type;
+        BitField<4, 4, ELFSymbolBinding> binding;
+    };
+    ELFSymbolVisibility visibility;
+    u16 sh_index;
+    u64 value;
+    u64 size;
+};
+static_assert(sizeof(ELFSymbol) == 0x18, "ELFSymbol has incorrect size.");
+
+using Symbols = std::vector<std::pair<ELFSymbol, std::string>>;
+
+Symbols GetSymbols(VAddr text_offset) {
+    const auto mod_offset = text_offset + Memory::Read32(text_offset + 4);
+
+    if (mod_offset < text_offset || (mod_offset & 0b11) != 0 ||
+        Memory::Read32(mod_offset) != Common::MakeMagic('M', 'O', 'D', '0')) {
+        return {};
+    }
+
+    const auto dynamic_offset = Memory::Read32(mod_offset + 0x4) + mod_offset;
+
+    VAddr string_table_offset{};
+    VAddr symbol_table_offset{};
+    u64 symbol_entry_size{};
+
+    VAddr dynamic_index = dynamic_offset;
     while (true) {
-        LOG_ERROR(Core_ARM, "{:016X}", lr);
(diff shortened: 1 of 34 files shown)


src/core/arm/arm_interface.cpp
```diff
@@ -2,26 +2,202 @@
...
+    return out;
 }
+
+void ARM_Interface::LogBacktrace() const {
+    const VAddr sp = GetReg(13);
+    const VAddr pc = GetPC();
+    LOG_ERROR(Core_ARM, "Backtrace, sp={:016X}, pc={:016X}", sp, pc);
+    LOG_ERROR(Core_ARM, "{:20}{:20}{:20}{:20}{}", "Module Name", "Address", "Original Address",
+              "Offset", "Symbol");
+    LOG_ERROR(Core_ARM, "{}", std::string(100, '-'));
```
lioncash: You may want to use a better method of indicating the end of a section than heap allocation 100 characters and tossing it away. This is kind of unnecessarily wasteful.

src/core/core.cpp
```diff
@@ -270,6 +271,9 @@ struct System::Impl {
     /// Telemetry session for this emulation session
     std::unique_ptr<Core::TelemetrySession> telemetry_session;
 
+    std::map<VAddr, std::string, std::greater<>> modules;
```
lioncash: `std::greater<>` can be omitted here, given the key isn't a type that would gain anything from heterogenous lookup.

src/core/core.cpp
```diff
@@ -509,6 +513,18 @@ void System::ClearContentProvider(FileSys::ContentProviderUnionSlot slot) {
     impl->content_provider->ClearSlot(slot);
 }
 
+void System::RegisterNSOModule(std::string name, VAddr start_address) {
```
lioncash: It seems really odd for the system class to manage this directly. This seems like a leaking of responsibilities that should go elsewhere. The system class should only exist to contain state, but avoid performing anything that isn't directly related to the lifecycle of emulation itself.

src/core/core.cpp
```diff
@@ -509,6 +513,18 @@ void System::ClearContentProvider(FileSys::ContentProviderUnionSlot slot) {
...
+    impl->modules.insert_or_assign(start_address, name);
```
lioncash:
```suggestion
    impl->modules.insert_or_assign(start_address, std::move(name));
```

src/core/hle/service/am/applets/general_backend.h
```diff
@@ -34,7 +34,7 @@ class PhotoViewer final : public Applet {
 
 class StubApplet final : public Applet {
 public:
-    StubApplet();
+    StubApplet(AppletId id);
```
lioncash:
```suggestion
    explicit StubApplet(AppletId id);
```

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+            backtrace_out.push_back(fmt::format("{:016X}", entry));
+        }
+        out["backtrace"] = std::move(backtrace_out);
+    }
+
+    return std::move(out);
+}
+
+json GetProcessorStateDataAuto() {
+    const auto* process{Core::CurrentProcess()};
```
lioncash: Please don't add global accessors like this in new code. Ditto below where relevant.

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+using namespace nlohmann;
+
+void SaveToFile(const json& json, const std::string& filename) {
+    FileUtil::CreateFullPath(filename);
+    std::ofstream file(
+        FileUtil::SanitizePath(filename, FileUtil::DirectorySeparator::PlatformDefault));
+    file << std::setw(4) << json << std::endl;
+    file.flush();
```
lioncash: `flush` and `close` aren't necessary. `std::endl` already flushes the stream, and the stream will close automatically when it goes out of scope.

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+std::string GetTimestamp() {
+    const auto time = std::time(nullptr);
+    return fmt::format("{:%FT%H-%M-%S}", *std::localtime(&time));
+}
+
+using namespace nlohmann;
+
+void SaveToFile(const json& json, const std::string& filename) {
+    FileUtil::CreateFullPath(filename);
```
lioncash: This is an operation that can fail.

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+    auto out = json{
+        {"title_id", fmt::format("{:016X}", title_id)},
+        {"result_raw", fmt::format("{:08X}", result.raw)},
+        {"result_module", fmt::format("{:08X}", static_cast<u32>(result.module.Value()))},
+        {"result_description", fmt::format("{:08X}", result.description.Value())},
+        {"timestamp", timestamp},
+    };
+    if (user_id.has_value())
+        out["user_id"] = fmt::format("{:016X}{:016X}", (*user_id)[1], (*user_id)[0]);
+    return std::move(out);
```
lioncash: `std::move()` is a pessimization here and should be removed.

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    if (backtrace.has_value()) {
+        auto backtrace_out = json::array();
+        for (const auto& entry : *backtrace) {
+            backtrace_out.push_back(fmt::format("{:016X}", entry));
+        }
+        out["backtrace"] = std::move(backtrace_out);
+    }
+
+    return std::move(out);
```
lioncash: ditto

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+        out.push_back({
+            {"module", entry.module},
+            {"address", fmt::format("{:016X}", entry.address)},
+            {"original_address", fmt::format("{:016X}", entry.original_address)},
+            {"offset", fmt::format("{:016X}", entry.offset)},
+            {"symbol_name", entry.name},
+        });
+    }
+
+    return std::move(out);
```
lioncash: Ditto

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+json GetFullDataAuto(const std::string& timestamp, u64 title_id) {
+    json out;
+
+    out["yuzu_version"] = GetYuzuVersionData();
+    out["report_common"] = GetReportCommonData(title_id, RESULT_SUCCESS, timestamp);
+    out["processor_state"] = GetProcessorStateDataAuto();
+    out["backtrace"] = GetBacktraceData();
+
+    return std::move(out);
```
lioncash: Ditto

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+        if constexpr (read_value) {
+            std::vector<u8> data(desc.Size());
+            Memory::ReadBlock(desc.Address(), data.data(), desc.Size());
+            entry["data"] = Common::HexVectorToString(data);
+        }
+
+        buffer_out.push_back(std::move(entry));
+    }
+
+    return std::move(buffer_out);
```
lioncash: Ditto

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    out["command_buffer"] = std::move(cmd_buf);
+
+    out["buffer_descriptor_a"] = GetHLEBufferDescriptorData<true>(ctx.BufferDescriptorA());
+    out["buffer_descriptor_b"] = GetHLEBufferDescriptorData<false>(ctx.BufferDescriptorB());
+    out["buffer_descriptor_c"] = GetHLEBufferDescriptorData<false>(ctx.BufferDescriptorC());
+    out["buffer_descriptor_x"] = GetHLEBufferDescriptorData<true>(ctx.BufferDescriptorX());
+
+    return std::move(out);
```
lioncash: Ditto

src/core/reporter.h
```diff
@@ -0,0 +1,53 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <optional>
+#include <vector>
+#include "common/common_types.h"
+#include "core/hle/result.h"
```
lioncash: This can be replaced with a forward declaration.

src/core/hle/service/prepo/prepo.cpp
```diff
@@ -40,8 +48,21 @@ class PlayReport final : public ServiceFramework<PlayReport> {
 
 private:
     void SaveReportWithUserOld(Kernel::HLERequestContext& ctx) {
-        // TODO(ogniK): Do we want to add play report?
-        LOG_WARNING(Service_PREPO, "(STUBBED) called");
+        IPC::RequestParser rp{ctx};
+        const auto user_id = rp.PopRaw<u128>();
+        const auto unk1 = rp.PopRaw<u64>();
```
lioncash: This is a process ID.

src/core/hle/service/am/applets/applets.cpp
```diff
@@ -35,12 +35,27 @@ AppletDataBroker::AppletDataBroker() {
 
 AppletDataBroker::~AppletDataBroker() = default;
 
+AppletDataBroker::RawChannelData AppletDataBroker::PeekDataToAppletForDebug() const {
+    std::vector<std::vector<u8>> out_normal;
+    std::vector<std::vector<u8>> out_interactive;
```
lioncash: This should be moved closest to the loop its actually used at.

src/core/hle/service/am/applets/applets.cpp
```diff
@@ -35,12 +35,27 @@ AppletDataBroker::AppletDataBroker() {
...
+
+    for (const auto& storage : in_channel) {
+        out_normal.push_back(storage->GetData());
+    }
+
+    for (const auto& storage : in_interactive_channel) {
+        out_interactive.push_back(storage->GetData());
+    }
+
+    return {out_normal, out_interactive};
```
lioncash: This should both be std::move'd

src/core/arm/arm_interface.cpp
```diff
@@ -2,26 +2,202 @@
...
+        out.push_back({"", 0, lr, 0});
         if (!fp) {
             break;
         }
         lr = Memory::Read64(fp + 8) - 4;
         fp = Memory::Read64(fp);
     }
+
+    const auto& modules{System::GetInstance().GetRegisteredNSOModules()};
+    std::map<std::string, Symbols> symbols;
```
lioncash: This can have it's size `reserve()`'d ahead of time.

src/core/arm/arm_interface.cpp
```diff
@@ -2,26 +2,202 @@
...
+
+std::optional<std::string> GetSymbolName(const Symbols& symbols, VAddr func_address) {
+    const auto iter =
+        std::find_if(symbols.begin(), symbols.end(), [func_address](const auto& pair) {
+            const auto& [symbol, name] = pair;
+            const auto end_address = symbol.value + symbol.size;
+            return func_address >= symbol.value && func_address < end_address;
+        });
+
+    if (iter == symbols.end())
```
lioncash:
```suggestion
    if (iter == symbols.end()) {
        ...
    }
```

Be consistent with bracing.

src/core/arm/arm_interface.cpp
```diff
@@ -2,26 +2,202 @@
...
+    VAddr symbol_table_offset{};
+    u64 symbol_entry_size{};
+
+    VAddr dynamic_index = dynamic_offset;
+    while (true) {
+        const auto tag = Memory::Read64(dynamic_index);
+        const auto value = Memory::Read64(dynamic_index + 0x8);
+        dynamic_index += 0x10;
+
+        if (tag == ELF_DYNAMIC_TAG_NULL)
```
lioncash: This and below should be braced. Ditto elsewhere where applicable.

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+void Reporter::SaveUnimplementedAppletReport(
+    u32 applet_id, u32 common_args_version, u32 library_version, u32 theme_color,
+    bool startup_sound, u64 system_tick, std::vector<std::vector<u8>> normal_channel,
+    std::vector<std::vector<u8>> interactive_channel) const {
+    if (!IsReportingEnabled())
+        return;
+
+    const auto timestamp{GetTimestamp()};
```
lioncash: Please be consistent with initialization. If you initialize one variable with `=` initialize them all with `=` if possible. It's more readable, and a more expected form of initialization.

src/core/reporter.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    SaveToFile(std::move(out), GetPath("crash_report", title_id, timestamp));
+}
+
+void Reporter::SaveSvcBreakReport(u32 type, bool signal_debugger, u64 info1, u64 info2,
+                                  std::optional<std::vector<u8>> resolved_buffer) const {
+    if (!IsReportingEnabled())
+        return;
+
+    const auto timestamp{GetTimestamp()};
```
lioncash: Ditto

PR #2485, opened 2019-05-18
shader/memory: Implement generic memory stores and loads (ST and LD)

ReinUsesLisp:
These instructions have a similar behaviour to STG and LDG, in fact at the moment of writing this I don't know what in particular makes them different since both write to and read to arbitrary memory addresses. Like STG's name suggests it might just allow writes to global memory meanwhile ST can write to local or other types of memory - that is a guess and it's very unlikely someone does that.

nouveau uses these to implement SSBOs.


bunnei:
@ReinUsesLisp needs a rebase, thanks!

windyboy1704:
Super Mario Odyssey now renders nothing in this PR. Can hear sound though.
Screenshot (243)

PR #2486, opened 2019-05-18
core/kernel/object: Rename ResetType enum members for clarity

lioncash:
Renames the members to more accurately indicate what they signify. "OneShot" and "Sticky" are kind of ambiguous identifiers for the reset types, and can be kind of misleading. Automatic and Manual communicate the kind of reset type in a clearer manner. Either the event is automatically reset, or it isn't and must be manually cleared.

The "OneShot" and "Sticky" terminology is just a hold-over from Citra where the kernel had a third type of event reset type known as "Pulse". Given the Switch kernel only has two forms of event reset types, we don't need to keep the old terminology around anymore.


lioncash:
Ah yeah, the kernel events themselves don't have the concept of a reset event, but the event data structures that wrap the event handles in userland/service code do have that functionality though (auto/manual reset).

PR #2489, opened 2019-05-18
Port citra-emu/citra#4716: "HLE/IPC: HLEContext can memorize the client thread and use it for SleepClientThread"

FearlessTobi:
See citra-emu/citra#4716 for more details.

**Original description**:
This reduces the boilerplate that services have to write out the current thread explicitly. Using current thread instead of client thread is also semantically incorrect, and will be a problem when we implement multicore (at which time there will be multiple current threads)

### DIFF
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -78,7 +78,7 @@ SharedPtr<WritableEvent> HLERequestContext::SleepClientThread(
 
 HLERequestContext::HLERequestContext(SharedPtr<Kernel::ServerSession> server_session,
                                      SharedPtr<Thread> thread)
-    : server_session(std::move(server_session)), thread(thread) {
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -76,8 +76,9 @@ SharedPtr<WritableEvent> HLERequestContext::SleepClientThread(
     return writable_event;
 }
 
-HLERequestContext::HLERequestContext(SharedPtr<Kernel::ServerSession> server_session)
-    : server_session(std::move(server_session)) {
+HLERequestContext::HLERequestContext(SharedPtr<Kernel::ServerSession> server_session,
+                                     SharedPtr<Thread> thread)
+    : server_session(std::move(server_session)), thread(thread) {
```
lioncash:
```suggestion
    : server_session(std::move(server_session)), thread(std::move(thread)) {
```

PR #2491, opened 2019-05-19
Dma_pusher: ASSERT on empty command_list

FernandoS27:
This is a measure to avoid crashes on command list reading as an empty 
command_list is considered a NOP.

### DIFF
--- a/src/video_core/dma_pusher.cpp
+++ b/src/video_core/dma_pusher.cpp
@@ -40,6 +40,13 @@ bool DmaPusher::Step() {
     }
 
     const CommandList& command_list{dma_pushbuffer.front()};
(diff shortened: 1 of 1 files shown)


src/video_core/dma_pusher.cpp
```diff
@@ -40,6 +40,13 @@ bool DmaPusher::Step() {
     }
 
     const CommandList& command_list{dma_pushbuffer.front()};
+    ASSERT_OR_EXECUTE(!command_list.empty(), {
+        // Somehow the command_list is empty, in order to avoid a crash
+        // We ignore it and assume it's size is 0.
```
ReinUsesLisp: `its`

PR #2492, opened 2019-05-19
 yuzu/debugger: Specify string conversions explicitly 

lioncash:
Another set of changes that brings us slightly closer to disabling implicit string conversions in Qt code. This makes all of the string conversions explicit in the debugger, which is relatively lower traffic in terms of changes compared to the rest of the UI code.


lioncash:
Failure is due to a timeout uploading the final executable. The build itself actually completes successfully.

PR #2494, opened 2019-05-19
 gl_shader_decompiler: Add AddLine() overloads with single function that forwards to libfmt

lioncash:
In a lot of places throughout the decompiler, string concatenation via operator+ is used quite heavily. This is usually fine, when not heavily used, but when used extensively, can be a problem. operator+ creates an entirely new heap allocated temporary string and given we perform expressions like:

```cpp
std::string thing = a + b + c + d;
```

this ends up with a lot of unnecessary temporary strings being created and discarded, which kind of thrashes the heap more than we need to. Given we utilize fmt in some AddLine calls, we can make this a part of the ShaderWriter's API. We can make an overload that simply acts as a passthrough to fmt.

This way, whenever things need to be appended to a string, the operation can be done via a single string formatting operation instead of discarding numerous temporary strings. This also has the benefit of making the strings themselves look nicer and makes it easier to spot errors in them. That, and it makes the API nicer to work with.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -57,19 +57,7 @@ public:
         shader_source += text;
     }
 
-    void AddLine(std::string_view text) {
-        AddExpression(text);
-        AddNewLine();
-    }
-
-    void AddLine(char character) {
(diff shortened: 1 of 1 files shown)


bunnei:
@lioncash clang format

lioncash:
Amended

PR #2514, opened 2019-05-25
video_core: Drop OpenGL core in favor of OpenGL compatibility

ReinUsesLisp:
This draft removes OpenGL core and replaces emulated features with native OpenGL compatibility, let me explain why this is a better approach since there's people that look at this profile with taboo.

First of all Maxwell hardware supports OpenGL compatibility features natively, here's a reduced list of compat-only features that it uses:
- Alpha testing
- `GL_QUADS`, `GL_QUAD_STRIPS`, `GL_POLYGON` (yes, quads **are** natively supported on Nvidia)
- `GL_CLAMP`
- Legacy varyings (e.g. `gl_MultiTexCoordN`, `gl_FogCoord`)
- Color sample depth+stencil textures like in hardware. This can't be controlled on OpenGL core but it may be changed on compatibility (I've yet to test it, it seems to be managed by `GL_DEPTH_TEXTURE_MODE`).

None of these features is trivial to emulate with OpenGL core. In contrast, compat is faster on Nvidia hardware since we don't have to inject alpha testing into fragment shaders or manually convert quads, simplifying our code.

I personally want to drop quad transformations to support inter-buffer communication. For example, one shader writes to global memory that is then used as index data for quads; with an OpenGL core approach this requires special a path to flush to host memory (slow). That's more code and potential bugs for something that can straight up avoided with compat.

That being said, this is what we lose if we switch to compatibility profile:
- i965 and nouveau support.
- Distributed RenderDoc binaries won't work on cases where compat features are used, it requires building RenderDoc with forced compat profile on OpenGL.
- OpenGL compatibility profile is a bit slower on some OpenGL implementations.

OpenGL compatibility profile specification can be found [here](https://www.khronos.org/registry/OpenGL/specs/gl/glspec46.compatibility.pdf).

I'm opening this as a draft since I want to discuss it before it gets merged or rejected.


bunnei:
Seems fine to me. My biggest concern is ~10 years from now, if/when compatibility profile is no longer supported by major GPU vendors, this might be a bit of a pain to back-track on. That being said, as long as the project remains active, I'm sure the code will have evolved in tandem.

That being said - Would it make any sense to maintain the core profile code paths to make it easier to port to other platforms? Or will we just plan to use Vulkan here?

ReinUsesLisp:
Vulkan in its current state will require the workarounds core has and more.

FernandoS27:
> Seems fine to me. My biggest concern is ~10 years from now, if/when compatibility profile is no longer supported by major GPU vendors, this might be a bit of a pain to back-track on. That being said, as long as the project remains active, I'm sure the code will have evolved in tandem.
> 
> That being said - Would it make any sense to maintain the core profile code paths to make it easier to port to other platforms? Or will we just plan to use Vulkan here?

The problem though is that OpenGL is dead. Many people doubt there will ever be an OpenGL 4.7 or 5.0

PR #2516, opened 2019-05-25
renderer_opengl/utils: Use a std::string_view with LabelGLObject()

lioncash:
Uses a std::string_view instead of a std::string, given the pointed to string isn't modified and is only used in a formatting operation. This is nice because a few usages directly supply a string literal to the function, allowing these usages to otherwise not heap allocate, unlike with std::string*. It's also nice for the early-return case, as no resources will be unnecessarily created and destroyed within the function. 

While we're at it, we can combine the address formatting into a single formatting call. 

\* - Unless the std::string implementation does small-string optimization and the string fits in the string buffer.

### DIFF
--- a/src/video_core/renderer_opengl/utils.cpp
+++ b/src/video_core/renderer_opengl/utils.cpp
@@ -38,27 +38,27 @@ void BindBuffersRangePushBuffer::Bind() const {
                        sizes.data());
 }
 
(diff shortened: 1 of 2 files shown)


lioncash:
Error doesn't appear to be related to this code.

PR #2519, opened 2019-05-25
loader/nso, core/core_timing_util: Silence sign-comparison warning

lioncash:
Silences a few -Wsign-compare warnings.


lioncash:
(failure isn't due to this code)

PR #2529, opened 2019-05-29
yuzu/bootmanager: Minor interface tidying

lioncash:
Tidies up a bit of the bootmanager code and makes it safer. Particularly, we now enforce that the parent of GRenderWindow is GMainWindow, and get rid of the downcast within the constructor, which makes the interface more type-safe.

We also now log out the destination of a screenshot rather than just saying it was saved. We also check if it actually successfully saved to disk before printing said logging message out. If we were unable to save the screenshot to disk, we now display an error message in the log.

The rest of the changes are primarily cosmetic changes made to make reading it nicer.


ReinUsesLisp:
This needs a rebase.

lioncash:
Done

PR #2533, opened 2019-05-30
memory: Add class to manage and enforce memory freezing

DarkLordZach:
This is pretty simple and is based on Atmosphere's approach to memory freezing. Simply, every frame when the freezer is active, all of the frozen values will be rewritten with the value stored at the beginning of the freezer.

Right now, this is only the class to accomplish this functionality, but I eventually intend to provide support for this with a UI (along with cheats) and hopefully expose this to some sort of scripting API.

### DIFF
--- a/src/core/memory/freezer.cpp
+++ b/src/core/memory/freezer.cpp
@@ -3,18 +3,20 @@
 // Refer to the license.txt file included.
 
 #include "common/assert.h"
+#include "common/logging/log.h"
 #include "core/core.h"
+#include "core/core_timing.h"
 #include "core/core_timing_util.h"
 #include "core/memory.h"
 #include "core/memory/freezer.h"
 
 namespace Memory {
 
-constexpr s64 MEMORY_FREEZER_TICKS = static_cast<s64>(Core::Timing::BASE_CLOCK_RATE / 60);
-
 namespace {
 
-u64 MemoryReadWidth(u8 width, VAddr addr) {
+constexpr s64 MEMORY_FREEZER_TICKS = static_cast<s64>(Core::Timing::BASE_CLOCK_RATE / 60);
+
+u64 MemoryReadWidth(u32 width, VAddr addr) {
     switch (width) {
     case 1:
         return Read8(addr);
@@ -30,7 +32,7 @@ u64 MemoryReadWidth(u8 width, VAddr addr) {
     }
 }
 
-void MemoryWriteWidth(u8 width, VAddr addr, u64 value) {
+void MemoryWriteWidth(u32 width, VAddr addr, u64 value) {
     switch (width) {
     case 1:
         Write8(addr, static_cast<u8>(value));
@@ -73,19 +75,19 @@ void Freezer::SetActive(bool active) {
 }
 
 bool Freezer::IsActive() const {
-    return active.load();
+    return active.load(std::memory_order_relaxed);
 }
 
 void Freezer::Clear() {
-    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
+    std::lock_guard lock{entries_mutex};
 
     LOG_DEBUG(Common_Memory, "Clearing all frozen memory values.");
 
     entries.clear();
 }
 
-u64 Freezer::Freeze(VAddr address, u8 width) {
-    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
+u64 Freezer::Freeze(VAddr address, u32 width) {
+    std::lock_guard lock{entries_mutex};
 
     const auto current_value = MemoryReadWidth(width, address);
     entries.push_back({address, width, current_value});
@@ -98,7 +100,7 @@ u64 Freezer::Freeze(VAddr address, u8 width) {
 }
 
 void Freezer::Unfreeze(VAddr address) {
-    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
+    std::lock_guard lock{entries_mutex};
 
     LOG_DEBUG(Common_Memory, "Unfreezing memory for address={:016X}", address);
 
@@ -108,8 +110,8 @@ void Freezer::Unfreeze(VAddr address) {
         entries.end());
 }
 
-bool Freezer::IsFrozen(VAddr address) {
-    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
+bool Freezer::IsFrozen(VAddr address) const {
+    std::lock_guard lock{entries_mutex};
 
     return std::find_if(entries.begin(), entries.end(), [&address](const Entry& entry) {
                return entry.address == address;
@@ -117,7 +119,7 @@ bool Freezer::IsFrozen(VAddr address) {
 }
 
 void Freezer::SetFrozenValue(VAddr address, u64 value) {
-    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
+    std::lock_guard lock{entries_mutex};
 
     const auto iter = std::find_if(entries.begin(), entries.end(), [&address](const Entry& entry) {
         return entry.address == address;
@@ -135,8 +137,8 @@ void Freezer::SetFrozenValue(VAddr address, u64 value) {
     iter->value = value;
 }
 
-std::optional<Freezer::Entry> Freezer::GetEntry(VAddr address) {
-    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
+std::optional<Freezer::Entry> Freezer::GetEntry(VAddr address) const {
+    std::lock_guard lock{entries_mutex};
 
     const auto iter = std::find_if(entries.begin(), entries.end(), [&address](const Entry& entry) {
         return entry.address == address;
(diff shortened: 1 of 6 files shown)


src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+#include "common/assert.h"
+#include "core/core.h"
+#include "core/core_timing_util.h"
+#include "core/memory.h"
+#include "core/memory/freezer.h"
+
+namespace Memory {
+
+constexpr s64 MEMORY_FREEZER_TICKS = static_cast<s64>(Core::Timing::BASE_CLOCK_RATE / 60);
```
lioncash: This should go in the anonymous namespace as well.

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+namespace {
+
+u64 MemoryReadWidth(u8 width, VAddr addr) {
```
lioncash: Please use a regular-width type like u32 or size_t to indicate width. Using a small size like this is unnecessary and can cause compilation warnings very easily in calling code.

Ditto for `MemoryWriteWidth`.

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+        FillEntryReads();
+        core_timing.ScheduleEvent(MEMORY_FREEZER_TICKS, event);
+        LOG_DEBUG(Common_Memory, "Memory freezer activated!");
+    } else {
+        LOG_DEBUG(Common_Memory, "Memory freezer deactivated!");
+    }
+}
+
+bool Freezer::IsActive() const {
+    return active.load();
```
lioncash: You likely just want:

```suggestion
    return active.load(std::memory_order_relaxed);
```

rather than enforcing the strictest form of atomic ordering.

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+void Freezer::Clear() {
+    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
```
lioncash:
```suggestion
    std::lock_guard lock{entries_mutex};
```

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    LOG_DEBUG(Common_Memory, "Clearing all frozen memory values.");
+
+    entries.clear();
+}
+
+u64 Freezer::Freeze(VAddr address, u8 width) {
+    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
```
lioncash:
```suggestion
    std::lock_guard lock{entries_mutex};
```

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    LOG_DEBUG(Common_Memory,
+              "Freezing memory for address={:016X}, width={:02X}, current_value={:016X}", address,
+              width, current_value);
+
+    return current_value;
+}
+
+void Freezer::Unfreeze(VAddr address) {
+    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
```
lioncash:
```suggestion
    std::lock_guard lock{entries_mutex};
```

... etc for other cases following this.

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    LOG_DEBUG(Common_Memory, "Unfreezing memory for address={:016X}", address);
+
+    entries.erase(
+        std::remove_if(entries.begin(), entries.end(),
+                       [&address](const Entry& entry) { return entry.address == address; }),
+        entries.end());
+}
+
+bool Freezer::IsFrozen(VAddr address) {
```
lioncash: This can be a const member function. The mutex can just be made mutable, which is generally advisable for mutexes, given they generally enforce a threading guarantee and don't directly affect the object they're a part of.

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+        return;
+    }
+
+    LOG_DEBUG(Common_Memory,
+              "Manually overridden freeze value for address={:016X}, width={:02X} to value={:016X}",
+              iter->address, iter->width, value);
+    iter->value = value;
+}
+
+std::optional<Freezer::Entry> Freezer::GetEntry(VAddr address) {
```
lioncash: Ditto

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+    });
+
+    if (iter == entries.end()) {
+        return std::nullopt;
+    }
+
+    return *iter;
+}
+
+std::vector<Freezer::Entry> Freezer::GetEntries() {
```
lioncash: Ditto

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
...
+    std::lock_guard<std::recursive_mutex> lock(entries_mutex);
+
+    return entries;
+}
+
+void Freezer::FrameCallback(u64 userdata, s64 cycles_late) {
+    if (!active.load()) {
```
lioncash:
```suggestion
    if (!IsActive()) {
```

Also are you certain this should be appearing before the mutex being locked?

src/core/memory/freezer.h
```diff
@@ -0,0 +1,58 @@
...
+
+#pragma once
+
+#include <optional>
+#include <vector>
+#include "common/common_types.h"
+#include "core/core_timing.h"
+
+namespace Core {
+class System;
```
lioncash: This doesn't appear to be used anywhere in the header.

src/core/memory/freezer.h
```diff
@@ -0,0 +1,58 @@
...
+// A class that will effectively freeze memory values.
+class Freezer {
+public:
+    struct Entry {
+        VAddr address;
+        u8 width;
+        u64 value;
+    };
+
+    Freezer(Core::Timing::CoreTiming& core_timing);
```
lioncash:
```suggestion
    explicit Freezer(Core::Timing::CoreTiming& core_timing);
```

src/core/memory/freezer.h
```diff
@@ -0,0 +1,58 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <optional>
+#include <vector>
+#include "common/common_types.h"
+#include "core/core_timing.h"
```
lioncash: The types used from this header can be forward declared.

src/core/memory/freezer.h
```diff
@@ -0,0 +1,58 @@
...
+
+    std::vector<Entry> GetEntries();
+
+private:
+    void FrameCallback(u64 userdata, s64 cycles_late);
+    void FillEntryReads();
+
+    std::atomic_bool active{false};
+
+    std::recursive_mutex entries_mutex;
```
lioncash: Are you absolutely certain this needs to be a recursive mutex and not a regular mutex.

src/core/memory/freezer.h
```diff
@@ -0,0 +1,58 @@
...
+#include "common/common_types.h"
+#include "core/core_timing.h"
+
+namespace Core {
+class System;
+} // namespace Core
+
+namespace Memory {
+
+// A class that will effectively freeze memory values.
```
lioncash: The functions themselves should probably contain documentation. Especially if any of them have noteworthy preconditions or postconditions.

src/core/memory/freezer.h
```diff
@@ -0,0 +1,58 @@
...
+
+    Freezer(Core::Timing::CoreTiming& core_timing);
+    ~Freezer();
+
+    void SetActive(bool active);
+    bool IsActive() const;
+
+    void Clear();
+
+    u64 Freeze(VAddr address, u8 width);
```
lioncash: `width` should likely be a regular u32` or something larger than a u16.

src/core/memory/freezer.cpp
```diff
@@ -0,0 +1,186 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/assert.h"
```
lioncash: `common/logging/log.h` also needs to be included.

src/core/memory/freezer.cpp
DarkLordZach: Yes, checking activity/stopping the callback if there is none doesn't read or write the entries vector.

bunnei:
IMO we should decouple code like this from core emulation - so as to not confuse it with something that's actually emulating Switch. Some suggestions:
* Add a detailed docstring that explains the intended use of this
* Put it within a namespace like `Cheats` or something
* Consider moving it within its own folder (thats less general sounding than `memory`)

src/core/tools/freezer.h
```diff
@@ -0,0 +1,75 @@
...
+#include "common/common_types.h"
+
+namespace Core::Timing {
+class CoreTiming;
+struct EventType;
+} // namespace Core::Timing
+
+namespace Tools {
+
+// A class that will effectively freeze memory values.
```
bunnei: nit: Can you just expand this a bit, to cover the fact that this is generally designed for cheats (and an example would help, for future viewers of this code...)

PR #2542, opened 2019-06-03
lbl: Implement brightness and backlight services

DarkLordZach:
This implements most basic commands from the `lbl` service, which is responsible for managing the LCD backlight of the switch. I emulated this functionality via the renderer by multiplying the final color by a value, with 1.0 corresponding to 50% brightness.

I made the setting non-persistent, after a restart of yuzu it will reset to default/normal/50% brightness.

Higher Brightness:
higher

Lower Brightness:
lower

Normal Brightness:
normal

(I had a LayeredFS mod active at the time)

### DIFF
--- a/src/core/hle/service/lbl/lbl.cpp
+++ b/src/core/hle/service/lbl/lbl.cpp
@@ -53,6 +53,21 @@ public:
         RegisterHandlers(functions);
     }
 
+    void LoadFromSettings() {
+        current_brightness = Settings::values.backlight_brightness;
+        current_vr_mode_brightness = Settings::values.backlight_brightness;
+
+        if (auto_brightness_enabled) {
+            return;
+        }
+
+        if (vr_mode_enabled) {
+            Renderer().SetCurrentBrightness(current_vr_mode_brightness);
+        } else {
+            Renderer().SetCurrentBrightness(current_brightness);
+        }
+    }
+
 private:
     void EnableVrMode(Kernel::HLERequestContext& ctx) {
         LOG_DEBUG(Service_LBL, "called");
@@ -80,8 +95,26 @@ private:
         rb.Push(vr_mode_enabled);
     }
 
+    bool auto_brightness_enabled = false;
+    bool dimming_enabled = true;
+
+    f32 current_brightness = GetAutoBrightnessValue();
+    f32 current_vr_mode_brightness = GetAutoBrightnessValue();
+
     bool vr_mode_enabled = false;
 };
+
+void RequestLoadCurrentSetting(SM::ServiceManager& sm) {
+    if (&sm == nullptr) {
+        return;
+    }
+
+    const auto lbl = sm.GetService<LBL>("lbl");
+
+    if (lbl) {
+        lbl->LoadFromSettings();
+    }
+}
 
 void InstallInterfaces(SM::ServiceManager& sm) {
     std::make_shared<LBL>()->InstallAsService(sm);
--- a/src/core/hle/service/lbl/lbl.cpp
+++ b/src/core/hle/service/lbl/lbl.cpp
@@ -10,6 +10,8 @@
 #include "core/hle/service/lbl/lbl.h"
 #include "core/hle/service/service.h"
 #include "core/hle/service/sm/sm.h"
+#include "core/settings.h"
+#include "video_core/renderer_base.h"
 
 namespace Service::LBL {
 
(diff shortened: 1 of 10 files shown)


src/core/hle/service/lbl/lbl.cpp
```diff
@@ -80,9 +283,27 @@ class LBL final : public ServiceFramework<LBL> {
...
+    bool dimming_enabled = true;
+
+    f32 current_brightness = GetAutoBrightnessValue();
+    f32 current_vr_mode_brightness = GetAutoBrightnessValue();
+
     bool vr_mode_enabled = false;
 };
 
+void RequestLoadCurrentSetting(SM::ServiceManager& sm) {
+    if (&sm == nullptr) {
```
lioncash: Please don't do this. Make the calling code explicitly only pass in valid values. A null reference is not well-formed C++. The only way this can actually occur is if a null pointer was previously dereferenced to create the reference value.

src/video_core/renderer_base.h
```diff
@@ -87,6 +91,17 @@ class RendererBase : NonCopyable {
     void RequestScreenshot(void* data, std::function<void()> callback,
                            const Layout::FramebufferLayout& layout);
 
+    // Gets the current brightness, even if it has been changed from the set value. Most of the time
+    // for yuzu this will simply match what was returned, but implementations are free to change the
+    // value in settings.
+    f32 GetCurrentResultantBrightness() const;
+
+    void SetBacklightStatus(bool enabled, u64 fade_transition_time);
```
lioncash: These functions should also have documentation comments. Particularly given `fade_transition_time` isn't really meaningful unless the units the time is in is also given.

These should all arguably be in a different class altogether, an instance of which would live in the renderer, since the renderer should only care about... as the name states, rendering, not directly managing backlight management as well, that's a separate responsibility. In fact, the backlight service delegates backlight calculations to a completely separate object from itself on the actual system.

The boolean parameter should also likely be an enum class. `SetBacklightStatus(false, value);` doesn't really communicate what it's actually toggling at the call site.

src/core/hle/service/lbl/lbl.h
```diff
@@ -10,6 +10,9 @@ class ServiceManager;
 
 namespace Service::LBL {
 
+// Requests the LBL service passed to load brightness values from Settings
+void RequestLoadCurrentSetting(SM::ServiceManager& sm);
```
lioncash: This should be placed below `InstallInterfaces()`

src/video_core/renderer_base.cpp
```diff
@@ -40,4 +40,35 @@ void RendererBase::RequestScreenshot(void* data, std::function<void()> callback,
...
+        renderer_settings.backlight_fade_time = fade_transition_time;
+    }
+}
+
+bool RendererBase::GetBacklightStatus() const {
+    return renderer_settings.current_brightness != 0;
+}
+
+void RendererBase::SetCurrentBrightness(f32 value) {
+    if (value != renderer_settings.current_brightness) {
```
lioncash:
```suggestion
    if (value == renderer_settings.current_brightness) {
        return;
    }
```

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -415,6 +426,29 @@ void RendererOpenGL::CaptureScreenshot() {
     renderer_settings.screenshot_requested = false;
 }
 
+void RendererOpenGL::UpdateBacklight() {
+    constexpr u64 PER_FRAME_FADE_TIME = 1000000000.0f / 60;
```
lioncash: Is the implicit conversion here intended? Note that this will result in `16666667`, rather than `16666666`, if that actually matters.

src/core/hle/service/lbl/lbl.cpp
```diff
@@ -53,13 +55,209 @@ class LBL final : public ServiceFramework<LBL> {
...
+        }
+    }
+
 private:
+    f32 GetAutoBrightnessValue() const {
+        return 0.5f;
+    }
+
+    VideoCore::RendererBase& Renderer() {
+        return Core::System::GetInstance().Renderer();
```
lioncash: Please don't use the global system accessor. Pass the system instance into this class if you need it.

src/core/hle/service/lbl/lbl.cpp
```diff
@@ -53,13 +55,209 @@ class LBL final : public ServiceFramework<LBL> {
...
+        LOG_DEBUG(Service_LBL, "called");
+
+        IPC::ResponseBuilder rb{ctx, 3};
+        rb.Push(RESULT_SUCCESS);
+        rb.Push(Renderer().GetCurrentResultantBrightness());
+    }
+
+    void SwitchBacklightOn(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        const auto fade_time = rp.PopRaw<u64>();
```
lioncash: This is a signed value in the actual service.

src/core/hle/service/lbl/lbl.cpp
```diff
@@ -53,13 +55,209 @@ class LBL final : public ServiceFramework<LBL> {
...
+
+        Renderer().SetBacklightStatus(true, fade_time);
+
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(RESULT_SUCCESS);
+    }
+
+    void SwitchBacklightOff(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        const auto fade_time = rp.PopRaw<u64>();
```
lioncash: Ditto

src/video_core/renderer_base.cpp
```diff
@@ -40,4 +40,35 @@ void RendererBase::RequestScreenshot(void* data, std::function<void()> callback,
     renderer_settings.screenshot_requested = true;
 }
 
+f32 RendererBase::GetCurrentResultantBrightness() const {
+    return renderer_settings.current_brightness / 2.0f;
```
lioncash: I'm fairly sure the calculation for this is much more complex than this in the service itself. AFAICR it uses both VR mode brightness values and normal mode brightness values to derive this.

src/video_core/renderer_base.cpp
```diff
@@ -40,4 +40,35 @@ void RendererBase::RequestScreenshot(void* data, std::function<void()> callback,
...
+        renderer_settings.current_brightness = current_brightness_backup;
+        renderer_settings.backlight_fade_time = fade_transition_time;
+    } else if (!enabled && renderer_settings.current_brightness != 0) {
+        current_brightness_backup = renderer_settings.current_brightness;
+        renderer_settings.current_brightness = 0;
+        renderer_settings.backlight_fade_time = fade_transition_time;
+    }
+}
+
+bool RendererBase::GetBacklightStatus() const {
```
lioncash: This is likely better named as `IsBacklightEnabled()`

bunnei:
@DarkLordZach reminder to address comments here

ogniK5377:
@DarkLordZach Another reminder to address issues

bunnei:
@DarkLordZach please address issues here.

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -416,6 +427,29 @@ void RendererOpenGL::CaptureScreenshot() {
...
+        glUniform4f(uniform_backlight, value, value, value, value);
+        renderer_settings.backlight_fade_time = 0;
+        fade_time_max = 0;
+    } else {
+        if (fade_time_max == 0) {
+            fade_time_max = fade_time;
+            value_max = value;
+        }
+
+        value += (value_max - value) * PER_FRAME_FADE_TIME / fade_time_max;
```
bunnei: This line no longer compiles with -Werror=conversion

bunnei:
Untagging, as this is breaking Linux build due to the above comment

PR #2546, opened 2019-06-05
loader, file_sys: Add support for parsing and loading KIP (Kernel Internal Process) files

DarkLordZach:
KIP files are used by some sysmodules, both homebrew and official. This is intended to be a "first step" towards LLE emulation of services, but all this PR adds is the ability to execute KIPs as if the main application is a KIP. This supports both compressed (with BLZ) and uncompressed KIPs.

This also adds INI file parsing, which is a custom yet simple container Nintendo uses to package up to 0x50 KIPs together. This is used internally with Package 2 (to contain some of the most essential sysmodules, like fsp-srv, sm, ldr, and pm)

I wrote a quick sample homebrew and used `elf2kip` to make a KIP out of it (as an initial PoC):
kip

My KIP if any of you want to try it:
[Hello World KIP File](https://github.com/yuzu-emu/yuzu/files/3255505/hello.zip)

I was also able to launch:
- A homebrew KIP for a custom sysmodule -- there was activity but ofc no graphics (it was xor.play so it wasn't possible to look for a service registration)
- The offical KIP for the `settings` services, v262144 -- this actually tried to register the `set` service and threw fatal because we already have it. If you remove yuzu's HLE of set, this will register it and then wait for requests as intended, a very good sign!

### DIFF
--- a/src/core/file_sys/kernel_executable.cpp
+++ b/src/core/file_sys/kernel_executable.cpp
@@ -34,7 +34,7 @@ bool DecompressBLZ(std::vector<u8>& data) {
         --index;
         auto control = data[index + start_offset];
         for (size_t i = 0; i < 8; ++i) {
-            if ((control & 0x80) > 0) {
+            if (((control << i) & 0x80) > 0) {
                 if (index < 2) {
                     return false;
                 }
@@ -70,9 +70,8 @@ bool DecompressBLZ(std::vector<u8>& data) {
                 data[out_index + start_offset] = data[index + start_offset];
             }
 
-            control <<= 1;
             if (out_index == 0)
-                return true;
+                break;
         }
     }
 
@@ -132,15 +131,15 @@ std::vector<u8> KIP::GetSectionDecompressed(u8 index) const {
 }
 
 bool KIP::Is64Bit() const {
-    return header.flags & 0x8;
+    return (header.flags & 0x8) != 0;
 }
 
 bool KIP::Is39BitAddressSpace() const {
-    return header.flags & 0x10;
+    return (header.flags & 0x10) != 0;
 }
 
 bool KIP::IsService() const {
-    return header.flags & 0x20;
+    return (header.flags & 0x20) != 0;
 }
 
 std::vector<u32> KIP::GetKernelCapabilities() const {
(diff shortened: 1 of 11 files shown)


src/core/file_sys/kernel_executable.cpp
```diff
@@ -0,0 +1,229 @@
...
+    data.resize(compressed_size + additional_size + start_offset);
+
+    std::size_t index = compressed_size - init_index;
+    std::size_t out_index = compressed_size + additional_size;
+
+    while (out_index > 0) {
+        --index;
+        auto control = data[index + start_offset];
+        for (size_t i = 0; i < 8; ++i) {
+            if ((control & 0x80) > 0) {
```
simontime: You can simplify this to `((control << i) & 0x80)`, removing the need for line 73.

src/core/file_sys/kernel_executable.cpp
```diff
@@ -0,0 +1,229 @@
...
+                    return false;
+                }
+                --out_index;
+                --index;
+                data[out_index + start_offset] = data[index + start_offset];
+            }
+
+            control <<= 1;
+            if (out_index == 0)
+                return true;
```
simontime: Breaking here would achieve the same result.

src/core/file_sys/kernel_executable.cpp
```diff
@@ -0,0 +1,229 @@
...
+u64 KIP::GetTitleID() const {
+    return header.title_id;
+}
+
+std::vector<u8> KIP::GetSectionDecompressed(u8 index) const {
+    return decompressed_sections[index];
+}
+
+bool KIP::Is64Bit() const {
+    return header.flags & 0x8;
```
lioncash: These should explicitly check for truthiness, like `(expression) != 0`, otherwise this is an implicit conversion to bool.

src/core/loader/kip.cpp
```diff
@@ -0,0 +1,98 @@
...
+    if (kip == nullptr) {
+        return {ResultStatus::ErrorNullFile, {}};
+    }
+
+    if (kip->GetStatus() != ResultStatus::Success) {
+        return {kip->GetStatus(), {}};
+    }
+
+    const auto address_space =
+        kip->Is64Bit() ? (kip->Is39BitAddressSpace() ? FileSys::ProgramAddressSpaceType::Is39Bit
```
lioncash: This should probably be broken out into a function or a lambda where this can be made way easier to visually parse.

PR #2548, opened 2019-06-05
applets: Implement backend and default frontend for Parental Controls and EShop (ShopN) applets

DarkLordZach:
This does:
- Adds the parental control applet backend based on switchbrew and personal RE verification. This also adds the frontend base class with all of the functions and a default impl which will simply log. I plan to eventually add a Qt frontend if I ever get around to doing the `pctl` service legitimately.
- 'Fixes' web applet processing -- based on data from switchbrew we now properly parse the input data for ShimKind and the TLVs. 
- Adds all offline web applet modes to LibAppletOffline -- previously we just assumed it was manual, but this allows applications to call for legal or system data pages as well and also implements the necessary nuances of each. After this PR, I plan on adding some sort of call into Qt so you can view app legal docs from the game list.
- Adds the LibAppletShop backend for the web browser -- this parses the TLVs and the arguments string and is based on personal RE I did. Right now, this uses a default frontend which just logs data but I considered adding a frontend in Qt to open the shop pages on nintendo.com

Web applet is used by BotW with update 1.1.0 or greater if not all of the DLC is purchased. It can be found on the title screen by going to `Downloadable Content` and then `Nintendo E-Shop`.

### DIFF
--- a/src/core/hle/service/am/applets/web_browser.cpp
+++ b/src/core/hle/service/am/applets/web_browser.cpp
@@ -246,24 +246,25 @@ void WebBrowser::ExecuteInteractive() {
 }
 
 void WebBrowser::Execute() {
-    if (complete)
+    if (complete) {
         return;
+    }
 
     if (status != RESULT_SUCCESS) {
         complete = true;
         return;
     }
 
-    frontend.OpenPage(filename, [this] { UnpackRomFS(); }, [this] { Finalize(); });
+    ExecuteInternal();
 }
 
 void WebBrowser::UnpackRomFS() {
     if (unpacked)
         return;
 
-    ASSERT(manual_romfs != nullptr);
+    ASSERT(offline_romfs != nullptr);
     const auto dir =
-        FileSys::ExtractRomFS(manual_romfs, FileSys::RomFSExtractionType::SingleDiscard);
+        FileSys::ExtractRomFS(offline_romfs, FileSys::RomFSExtractionType::SingleDiscard);
     const auto& vfs{Core::System::GetInstance().GetFilesystem()};
     const auto temp_dir = vfs->CreateDirectory(temporary_dir, FileSys::Mode::ReadWrite);
     FileSys::VfsRawCopyD(dir, temp_dir);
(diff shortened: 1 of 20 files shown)


src/core/hle/service/am/applets/web_browser.h
```diff
@@ -32,15 +39,39 @@ class WebBrowser final : public Applet {
...
-    FileSys::VirtualFile manual_romfs;
+    ShimKind kind;
+    std::map<WebArgTLVType, std::vector<u8>> args;
+
+    FileSys::VirtualFile offline_romfs;
     std::string temporary_dir;
     std::string filename;
+
+    ShopWebTarget shop_web_target;
+    std::map<std::string, std::string> shop_query;
```
lioncash: This can be:

```suggestion
    std::map<std::string, std::string, std::less<>> shop_query;
```

to allow heterogenous lookup. This way lookups like:

```cpp
const auto scene = shop_query.find("scene");
```

won't cause scene to be converted into a std::string before doing the lookup.

src/core/hle/service/am/applets/web_browser.cpp
```diff
@@ -172,17 +275,273 @@ void WebBrowser::UnpackRomFS() {
...
+
+    offline_romfs = GetApplicationRomFS(title_id, type);
+    if (offline_romfs == nullptr) {
+        status = ResultCode(-1);
+        LOG_ERROR(Service_AM, "Failed to find offline data for request!");
+    }
+
+    std::string path_additional_directory;
+    if (source == OfflineWebSource::OfflineHtmlPage) {
+        path_additional_directory = std::string(DIR_SEP) + "html-document";
```
lioncash:
```suggestion
        path_additional_directory = std::string(DIR_SEP).append("html-document");
```

src/core/hle/service/am/applets/web_browser.cpp
```diff
@@ -172,17 +275,273 @@ void WebBrowser::UnpackRomFS() {
...
+    case OfflineWebSource::ApplicationLegalInformation:
+        type = FileSys::ContentRecordType::Legal;
+        break;
+    case OfflineWebSource::SystemDataPage:
+        type = FileSys::ContentRecordType::Data;
+        break;
+    }
+
+    if (title_id == 0) {
+        title_id = Core::System::GetInstance().CurrentProcess()->GetTitleID();
```
lioncash: The global system accessor should be avoided in new code.

src/core/hle/service/am/applets/web_browser.cpp
```diff
@@ -172,17 +275,273 @@ void WebBrowser::UnpackRomFS() {
...
+
+    const auto scene = shop_query.find("scene");
+
+    if (scene == shop_query.end()) {
+        LOG_ERROR(Service_AM, "No scene parameter was passed via shop query!");
+        status = ResultCode(-1);
+        return;
+    }
+
+    const std::map<std::string, ShopWebTarget> target_map{
```
lioncash:
```suggestion
    const std::map<std::string, ShopWebTarget, std::less<>> target_map{
```

to allow heterogenous lookup.

PR #2555, opened 2019-06-06
kernel/process: Decouple TLS handling from threads

lioncash:
Previously each thread instance was holding a block of memory representing a TLS region. An issue with that is:

1. It's not necessary (it's mostly another carry-over from Citra).
2. It over-exposes info that shouldn't really be shared outside of a process instance.

This brings the TLS slot management more in line with how the kernel on the actual system handles it and makes way for handling other aspects related to TLS regions more correctly.

A thread instance is also no longer a requirement for retrieving a TLS region. This is important because a thread isn't the only thing that's assigned a TLS region. Process instances are also assigned one during creation. This will be handled in another PR, as it requires some changes outside of the kernel. This change preserves existing behavior while providing the necessary machinery to work off of.

### DIFF
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <bitset>
 #include <memory>
 #include <random>
 #include "common/alignment.h"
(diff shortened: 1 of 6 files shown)


bunnei:
@lioncash needs a rebase

PR #2563, opened 2019-06-07
gl_shader_cache: Use static constructors for CachedShader initialization 

ReinUsesLisp:
This unifies `CachedShader` constructors into a single method. The two previous constructors had different meanings depending on the arguments that were being passed, this commit addresses that using static public initializers with descriptive names instead.

Shader arguments are now packed into a structure.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_cache.cpp
@@ -221,44 +221,37 @@ std::set<GLenum> GetSupportedFormats() {
 
 } // Anonymous namespace
 
-CachedShader::CachedShader(const Device& device, VAddr cpu_addr, u64 unique_identifier,
-                           Maxwell::ShaderProgram program_type, ShaderDiskCacheOpenGL& disk_cache,
-                           const PrecompiledPrograms& precompiled_programs,
-                           ProgramCode&& program_code, ProgramCode&& program_code_b, u8* host_ptr)
-    : RasterizerCacheObject{host_ptr}, host_ptr{host_ptr}, cpu_addr{cpu_addr},
-      unique_identifier{unique_identifier}, program_type{program_type}, disk_cache{disk_cache},
-      precompiled_programs{precompiled_programs} {
-    const std::size_t code_size{CalculateProgramSize(program_code)};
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_shader_cache.h
```diff
@@ -41,17 +41,27 @@ using Maxwell = Tegra::Engines::Maxwell3D::Regs;
...
+    u64 unique_identifier;
+};
+
 class CachedShader final : public RasterizerCacheObject {
 public:
-    explicit CachedShader(const Device& device, VAddr cpu_addr, u64 unique_identifier,
-                          Maxwell::ShaderProgram program_type, ShaderDiskCacheOpenGL& disk_cache,
-                          const PrecompiledPrograms& precompiled_programs,
-                          ProgramCode&& program_code, ProgramCode&& program_code_b, u8* host_ptr);
+    explicit CachedShader(const ShaderParameters& params, Maxwell::ShaderProgram program_type,
```
DarkLordZach: This should *probably* be private, esp. considering `CreateStageFromCache` just wraps it and from what I can tell only the `CreateStageFrom*` functions should use it as shaders are used as shared_ptrs, not raw object

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -221,44 +221,37 @@ std::set<GLenum> GetSupportedFormats() {
...
+                                           Maxwell::ShaderProgram program_type,
+                                           ProgramCode&& program_code,
+                                           ProgramCode&& program_code_b) {
+    const auto code_size{CalculateProgramSize(program_code)};
+    const auto code_size_b{CalculateProgramSize(program_code_b)};
+    auto result{CreateProgram(params.device, program_type, program_code, program_code_b)};
+    if (result.first.empty()) {
         // TODO(Rodrigo): Unimplemented shader stages hit here, avoid using these for now
-        return;
+        return {};
```
DarkLordZach: Up to you, but should this be logged?

ReinUsesLisp: Logging is done on the callee.

PR #2565, opened 2019-06-07
shader/track: Track indirect buffers

ReinUsesLisp:
nouveau uses LDC instead of MOV Rd, c[n][m] to load global memory pointers into a shader. This commit allows the tracking heuristic to find these.

Tracking would previously fail due to LDC being decoded to an operation on the constant buffer address (it's an indirect read). To fix this issue this commit searches inside the operation for the base address.


ReinUsesLisp:
This commit is wrong, I'm working on a fix for it.

The problem is that it should only accept while tracking those operands that are no-ops (e.g. `IADD R0 RZ`).

ReinUsesLisp:
Addressed. Indirect buffer loads without a variable will get inlined, allowing the tracking code to find it as an immediate.

PR #2574, opened 2019-06-10
arm_dynarmic: Check if jit is nullptr when preparing reschedule

DarkLordZach:
Prevents crash with multiprocess loading.


lioncash:
I feel like this is masking an initialization order issue elsewhere. The JIT really shouldn't be null here.

ogniK5377:
As @lioncash states, the jit shouldn't be null here, however, for the time being, i'll tag it as canary until the issue is resolved. This should not be merged into master

PR #2576, opened 2019-06-10
nsp: Fix various errors with loading and processing of extracted NSPs

DarkLordZach:
Since NSPs actually comprise two formats, one of which contains NCAs and the other contains the raw game executable, the NSP processing code had to be updated as many assumptions were made that make extracted type NSPs have no title ID (even if they have one in main.npdm) and have a Missing NCA error, which is true they don't have NCAs.

It is now possible to properly load these types of NSPs.


DarkLordZach:
Build failure unrelated to PR; cannot restart appveyor job

PR #2577, opened 2019-06-11
file_sys/card_image: Minor cleanup

lioncash:
Just performs a minor readability/deduplication cleanup

### DIFF
--- a/src/core/file_sys/card_image.cpp
+++ b/src/core/file_sys/card_image.cpp
@@ -18,7 +18,12 @@
 
 namespace FileSys {
 
-constexpr std::array<const char*, 0x4> partition_names = {"update", "normal", "secure", "logo"};
+constexpr std::array partition_names{
+    "update",
+    "normal",
+    "secure",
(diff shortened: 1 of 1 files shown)


src/core/file_sys/card_image.cpp
```diff
@@ -169,17 +176,22 @@ VirtualDir XCI::GetParentDirectory() const {
...
     }
 
-    for (const VirtualFile& file : partitions[static_cast<std::size_t>(part)]->GetFiles()) {
-        if (file->GetExtension() != "nca")
+    for (const VirtualFile& file : partition->GetFiles()) {
+        if (file->GetExtension() != "nca") {
             continue;
+        }
         auto nca = std::make_shared<NCA>(file, nullptr, 0, keys);
         // TODO(DarkLordZach): Add proper Rev1+ Support
```
DarkLordZach: While you're here, I think this can be removed.
I have no idea why it's still todo we've had Rev1+ support for months now

PR #2583, opened 2019-06-15
Core_Timing: Make core_timing threadsafe by default.

FernandoS27:
The old implementation had faulty Threadsafe methods (they were threadsafe but it had other issues related to design causing bugs that made missing events possible under certain circunstances). In general, I see no regressions from using a mutex and speed is pretty much the same. If this becomes a concern, it can easily be replaced by a very simple spinlock and fast. This also simplifies everything and keeps event scheduling/unscheduling under one single method.

### DIFF
--- a/src/core/core_timing.cpp
+++ b/src/core/core_timing.cpp
@@ -56,12 +56,12 @@ void CoreTiming::Initialize() {
 }
 
 void CoreTiming::Shutdown() {
-    MoveEvents();
     ClearPendingEvents();
     UnregisterAllEvents();
 }
 
 EventType* CoreTiming::RegisterEvent(const std::string& name, TimedCallback callback) {
+    std::lock_guard guard{inner_mutex};
     // check for existing type with same name.
     // we want event type names to remain unique so that we can use them for serialization.
     ASSERT_MSG(event_types.find(name) == event_types.end(),
@@ -82,6 +82,7 @@ void CoreTiming::UnregisterAllEvents() {
 
 void CoreTiming::ScheduleEvent(s64 cycles_into_future, const EventType* event_type, u64 userdata) {
     ASSERT(event_type != nullptr);
+    std::lock_guard guard{inner_mutex};
     const s64 timeout = GetTicks() + cycles_into_future;
 
     // If this event needs to be scheduled before the next advance(), force one early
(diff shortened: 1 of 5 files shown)


src/core/core_timing.cpp
```diff
@@ -162,19 +151,8 @@ void CoreTiming::ForceExceptionCheck(s64 cycles) {
...
-        std::push_heap(event_queue.begin(), event_queue.end(), std::greater<>());
-    }
-}
-
 void CoreTiming::Advance() {
-    MoveEvents();
-    for (std::pair<const EventType*, u64> ev; unschedule_queue.Pop(ev);) {
-        UnscheduleEvent(ev.first, ev.second);
-    }
+    inner_mutex.lock();
```
lioncash: You can use a unique_lock here, which would let you lock/unlock the contained mutex within the loop and let you remove the manual unlock at the end of the function.

lioncash:
Commit message should also contain a description of the changes so the explanation is in the git history.

src/core/core_timing.h
```diff
@@ -67,7 +68,7 @@ class CoreTiming {
     ///
     EventType* RegisterEvent(const std::string& name, TimedCallback callback);
 
-    /// Unregisters all registered events thus far.
+    /// Unregisters all registered events thus far. Note thread unsafe
```
DarkLordZach: Is there no possible way to make this threadsafe?

FernandoS27: That's suppose to be called on shutdown. It can be a nightmare to make threadsafe.

src/core/core_timing.cpp
```diff
@@ -82,6 +82,7 @@ void CoreTiming::UnregisterAllEvents() {
 
 void CoreTiming::ScheduleEvent(s64 cycles_into_future, const EventType* event_type, u64 userdata) {
```
degasus: This change will call ForceExceptionCheck from other threads. So all functions accessing cycles, downcount and slice_length must be protected as well. e.g. GetTicks and AddTicks

FernandoS27: We don't emulate multicore cpu cores and when we do each will have their own counts and time will be synced together on a barrier.  That's one of the solutions. The other is only adding ticks to the cycle counter on Core0. In general, we can't mix addticks in multicore as it's fully impractical.

degasus: Oh no, I'm talking about calling ScheduleEvent from any thread while the CPU is running.

FernandoS27: Oh I see what you mean. In general, it's not a problem, as ForceExceptionCheck would only trigger sequentially and can only trigger while core timing is doing a callback, it's still fine in that case.

B3n30:
I just noticed that that PR is open. Just wanted to mention that this design is really bad and will for sure cause a ton of issues. My suggestion for a better thread safe impl would be to remove the current (Un)ScheduleEvent and rename the (Un)ScheduleEventThreadsafe to (Un)ScheduleEvent. that way it is safe to add and remove events from any thread while still guarantee that all events are scheduled once and only once

FernandoS27:
the threadsafe versions of the core timing where erronous and faulty.

B3n30:
I‘m not saying they are correct, but it’s the way to Go. From your answer I get you don’t understand what I mean. Scheduling an Event from another Core will now cause the queue to get resorted and this is can happen while the main core is in any State. This will lead to undefiniert behavior.

src/core/core_timing.cpp
degasus: Sorry for the later comment. I think this is a problem here. Yes, ForceExceptionCheck is called sequentially. But the CPU thread will query its internal state (cycles, downcount and slice_length) in an unsynchronized way. So running the CPU while queueing events *is* a race condition.

FernandoS27:
We are droping current Multicore design. In the future only 1 core advances timer/add ticks and manages events, the rest just add into it.

PR #2584, opened 2019-06-16
Impl'd IsUserAccountSwitchLocked, SetAudioOutVolume, GetAudioOutVolume & Partial impl of GetAccumulatedSuspendedTickChangedEvent

ogniK5377:
IPC-100 was changed to InitializeApplicationInfoOld instead of InitializeApplicationInfo. IPC-150 makes an identical call to IPC-100 however does extra processing. They should not have the same name as it's quite confusing to debug.

### DIFF
--- a/src/audio_core/stream.h
+++ b/src/audio_core/stream.h
@@ -61,6 +61,12 @@ public:
     /// Returns a vector of recently released buffers specified by tag
     std::vector<Buffer::Tag> GetTagsAndReleaseBuffers(std::size_t max_count);
 
+    void SetVolume(float volume);
+
+    float GetVolume() const {
+        return game_volume;
+    }
+
     /// Returns true if the stream is currently playing
     bool IsPlaying() const {
         return state == State::Playing;
@@ -103,6 +109,7 @@ private:
     SinkStream& sink_stream;                  ///< Output sink for the stream
     Core::Timing::CoreTiming& core_timing;    ///< Core timing instance.
     std::string name;                         ///< Name of the stream, must be unique
+    float game_volume = 1.0f;                 ///< The volume the game currently has set
 };
 
 using StreamPtr = std::shared_ptr<Stream>;
--- a/src/core/file_sys/control_metadata.cpp
+++ b/src/core/file_sys/control_metadata.cpp
@@ -87,6 +87,10 @@ u64 NACP::GetDefaultJournalSaveSize() const {
     return raw.user_account_save_data_journal_size;
 }
 
+bool NACP::GetUserAccountSwitchLock() const {
+    return raw.user_account_switch_lock == 1;
+}
+
 u32 NACP::GetSupportedLanguages() const {
     return raw.supported_languages;
 }
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -12,13 +12,17 @@
 #include "common/swap.h"
 #include "core/constants.h"
 #include "core/core_timing.h"
+#include "core/file_sys/control_metadata.h"
+#include "core/file_sys/patch_manager.h"
 #include "core/hle/ipc_helpers.h"
+#include "core/hle/kernel/process.h"
 #include "core/hle/service/acc/acc.h"
 #include "core/hle/service/acc/acc_aa.h"
 #include "core/hle/service/acc/acc_su.h"
 #include "core/hle/service/acc/acc_u0.h"
 #include "core/hle/service/acc/acc_u1.h"
 #include "core/hle/service/acc/profile_manager.h"
+#include "core/loader/loader.h"
 
 namespace Service::Account {
 
@@ -213,7 +217,7 @@ void Module::Interface::IsUserRegistrationRequestPermitted(Kernel::HLERequestCon
     rb.Push(profile_manager->CanSystemRegisterUser());
 }
 
-void Module::Interface::InitializeApplicationInfo(Kernel::HLERequestContext& ctx) {
+void Module::Interface::InitializeApplicationInfoOld(Kernel::HLERequestContext& ctx) {
     LOG_WARNING(Service_ACC, "(STUBBED) called");
     IPC::ResponseBuilder rb{ctx, 2};
     rb.Push(RESULT_SUCCESS);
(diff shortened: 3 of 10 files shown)


src/core/hle/service/acc/acc.cpp
```diff
@@ -226,6 +230,29 @@ void Module::Interface::GetBaasAccountManagerForApplication(Kernel::HLERequestCo
     rb.PushIpcInterface<IManagerForApplication>();
 }
 
+void Module::Interface::IsUserAccountSwitchLocked(Kernel::HLERequestContext& ctx) {
+    LOG_DEBUG(Service_ACC, "called");
+    FileSys::NACP nacp;
+    const auto res = Core::System::GetInstance().GetAppLoader().ReadControlData(nacp);
```
lioncash: This system instance should be passed into the Module::Interface instance and used instead of using the global system accessor.

src/core/hle/service/acc/acc.cpp
```diff
@@ -226,6 +230,29 @@ void Module::Interface::GetBaasAccountManagerForApplication(Kernel::HLERequestCo
...
+
+    bool is_locked = false;
+
+    if (res != Loader::ResultStatus::Success) {
+        FileSys::PatchManager pm{Core::CurrentProcess()->GetTitleID()};
```
lioncash: Ditto. This should use the reference to the system instead of a global accessor.

src/core/hle/service/acc/acc.cpp
```diff
@@ -226,6 +230,29 @@ void Module::Interface::GetBaasAccountManagerForApplication(Kernel::HLERequestCo
...
+        auto [nacp_unique, discard] = pm.GetControlMetadata();
```
lioncash: This will cause unused variable warnings on gcc and clang. Don't use structured bindings when only one of the returned variables are used.

src/core/hle/service/audio/audout_u.cpp
```diff
@@ -183,6 +183,25 @@ class IAudioOut final : public ServiceFramework<IAudioOut> {
         rb.Push(static_cast<u32>(stream->GetQueueSize()));
     }
 
+    void SetAudioOutVolume(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        float volume = rp.PopRaw<float>();
```
```suggestion
        const float volume = rp.Pop<float>();
```

src/core/hle/service/audio/audout_u.cpp
```diff
@@ -183,6 +183,25 @@ class IAudioOut final : public ServiceFramework<IAudioOut> {
...
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(RESULT_SUCCESS);
+    }
+
+    void GetAudioOutVolume(Kernel::HLERequestContext& ctx) {
+        LOG_DEBUG(Service_Audio, "called");
+
+        IPC::ResponseBuilder rb{ctx, 3};
+        rb.Push(RESULT_SUCCESS);
+        rb.PushRaw<float>(stream->GetVolume());
```
```suggestion
        rb.Push(stream->GetVolume());
```

src/core/hle/service/acc/acc.cpp
```diff
@@ -226,6 +230,29 @@ void Module::Interface::GetBaasAccountManagerForApplication(Kernel::HLERequestCo
...
+        if (nacp_unique != nullptr) {
+            is_locked = nacp_unique->GetUserAccountSwitchLock();
+        }
+    } else {
+        is_locked = nacp.GetUserAccountSwitchLock();
+    }
+
+    IPC::ResponseBuilder rb{ctx, 3};
+    rb.Push(RESULT_SUCCESS);
+    rb.PushRaw<u8>(is_locked);
```
lioncash:
```suggestion
    rb.Push(is_locked);
```

src/core/hle/service/acc/acc.cpp
```diff
@@ -226,6 +230,29 @@ void Module::Interface::GetBaasAccountManagerForApplication(Kernel::HLERequestCo
     rb.PushIpcInterface<IManagerForApplication>();
 }
 
+void Module::Interface::IsUserAccountSwitchLocked(Kernel::HLERequestContext& ctx) {
+    LOG_DEBUG(Service_ACC, "called");
+    FileSys::NACP nacp;
```
lioncash: This can be moved closer to where it's used.

src/core/hle/service/acc/acc.cpp
```diff
@@ -226,6 +230,29 @@ void Module::Interface::GetBaasAccountManagerForApplication(Kernel::HLERequestCo
...
+
+    if (res != Loader::ResultStatus::Success) {
+        FileSys::PatchManager pm{Core::CurrentProcess()->GetTitleID()};
+        auto [nacp_unique, discard] = pm.GetControlMetadata();
+
+        if (nacp_unique != nullptr) {
+            is_locked = nacp_unique->GetUserAccountSwitchLock();
+        }
+    } else {
+        is_locked = nacp.GetUserAccountSwitchLock();
```
lioncash: Are you sure this does what you want it to? This will always return false because `nacp` isn't supplied with a file instance; it's using the default constructor.

src/audio_core/stream.h
```diff
@@ -103,6 +109,7 @@ class Stream {
     SinkStream& sink_stream;                  ///< Output sink for the stream
     Core::Timing::CoreTiming& core_timing;    ///< Core timing instance.
     std::string name;                         ///< Name of the stream, must be unique
+    float game_volume = 1.0f;                 ///< The volume the game currently has set
```
lioncash: This should be placed up by `sample_rate` and `format`

src/core/hle/service/acc/acc.cpp
```diff
@@ -278,19 +278,22 @@ void Module::Interface::TrySelectUserWithoutInteraction(Kernel::HLERequestContex
...
 Module::Interface::~Interface() = default;
 
 void InstallInterfaces(SM::ServiceManager& service_manager) {
     auto module = std::make_shared<Module>();
     auto profile_manager = std::make_shared<ProfileManager>();
-    std::make_shared<ACC_AA>(module, profile_manager)->InstallAsService(service_manager);
-    std::make_shared<ACC_SU>(module, profile_manager)->InstallAsService(service_manager);
-    std::make_shared<ACC_U0>(module, profile_manager)->InstallAsService(service_manager);
-    std::make_shared<ACC_U1>(module, profile_manager)->InstallAsService(service_manager);
+    Core::System& system = Core::System::GetInstance();
```
lioncash: This should be passed into the InstallInterfaces call. The central Init() call where this InstallInterface function is called already makes a System instance available.

DarkLordZach: @ogniK5377 Just replace the SM::ServiceManager& with Core::System& and use system.ServiceManager() in func.
(sidenote: we should prob do this for all services)

src/core/file_sys/control_metadata.cpp
```diff
@@ -87,6 +87,10 @@ u64 NACP::GetDefaultJournalSaveSize() const {
     return raw.user_account_save_data_journal_size;
 }
 
+bool NACP::GetUserAccountSwitchLock() const {
+    return raw.user_account_switch_lock == 1;
```
DarkLordZach:
```suggestion
    return raw.user_account_switch_lock != 0;
```

src/core/hle/service/acc/acc.cpp
```diff
@@ -226,6 +230,29 @@ void Module::Interface::GetBaasAccountManagerForApplication(Kernel::HLERequestCo
...
+
+    bool is_locked = false;
+
+    if (res != Loader::ResultStatus::Success) {
+        FileSys::PatchManager pm{system.CurrentProcess()->GetTitleID()};
+        auto nacp_unique = pm.GetControlMetadata().first;
+
+        if (nacp_unique != nullptr) {
+            is_locked = nacp_unique->GetUserAccountSwitchLock();
+        }
```
DarkLordZach: Pls log if nacp_unique is nullptr, also is false a suitable default in that case?

ogniK5377: Should be suitable, most games have it set to false

src/audio_core/stream.h
```diff
@@ -94,6 +100,7 @@ class Stream {
 
     u32 sample_rate;                          ///< Sample rate of the stream
     Format format;                            ///< Format of the stream
+    float game_volume = 1.0f;                 ///< The volume the game currently has set
```
DarkLordZach: As lioncash reviewed:
```suggestion
    float volume = 1.0f;                 ///< The volume the game currently has set
```

PR #2587, opened 2019-06-17
gl_rasterizer: Force depth testing when depth mask is enabled

ReinUsesLisp:
This is a workaround for an OpenGL weird design choice where the depth buffer is not changed unless depth testing is enabled.


ReinUsesLisp:
Untagging until I find the cause of the regression. Probably we have to do this when a shader enables gl_FragDepth in the OMAP header.

ReinUsesLisp:
Closing in favour of #2735. This should be revised after those changes.

PR #2590, opened 2019-06-18
service/audio/audren_u: Correct event reset type for the system event

lioncash:
This is actually an auto-reset event in the audio service itself, not a manual one.


FernandoS27:
Hmn I would test this. The current implementation was designed considering it Manual. It may bring bugs.

lioncash:
It works fine. The event currently isn't actually signaled by anything yet. We have the scaffolding for it built-up though.

lioncash:
MingW failure is a timeout, not a compilation error

PR #1634, opened 2018-11-02
hid: Add support for multiplayer and multilayout controllers

DarkLordZach:
The fabled "HID 2" successor to #1444. This would not have been possible without HOS legend @ogniK5377 for his backend work to make this possible!

- Adds full UI support to change the connectivity, type/layouts and buttons for all eight players, the handheld controller, the debug controller, mouse, keyboard, and finally touchscreen.
- Moves the docked mode setting to the input tab as it directly affects whether or not handheld can be enabled
- Automatically rotates the left and right joycons when used in a single-joycon configuration
- Will automatically change your configuration to the closest supported one if the game rejects it
- Avoids reloading the game list on every config change

here are some pics of the UI:
hid2-1
hid2-2
hid-2-3
hid-2-4

PR #2594, opened 2019-06-19
yuzu/configure_input: Add missing space in window title

FearlessTobi:


MysticExile:
I agree with what @Schplee said, one of the main devs should look over this as it’s way to complicated for any normal human to understand the complexity if this PR.

src/yuzu/configuration/configure_input.ui
```diff
@@ -11,7 +11,7 @@
    </rect>
   </property>
   <property name="windowTitle">
-   <string>ConfigureInput</string>
+   <string>Configure Input</string>
```
DarkLordZach: This should be more specific and was originally (in #1780 / #1634 ) intended to be more descriptive. As it stands, it was a holdover from when the name was invisible.

Its really up to you, but `Custom Input Settings` or something akin to that is prob ok.

PR #2607, opened 2019-06-25
glue: Implement arp:w and arp:r services

DarkLordZach:
`glue` is a switch sysmodule, with subservices `arp` used to keep track of an application's control and launch properties. The launch property conveys information about title ID, title version, and the media used to launch the game (SD/NAND/GC). The control data is the same data where the name and version string are stored, along with plenty of additional parameters like DLC, Save Data, BCAT, etc parameters. 

Aptly named `glue`, this module 'glues' together various parts of the system which need to access control data (hint: numerous) and/or launch data. 

This includes the `arp:r` service, which can return the application launch or control properties, cataloged by process ID and title ID, and the `arp:w` service, which manages removal/registration of entries in the DB.

For yuzu, this is managed by a new class `Service::Glue::ARPManager`, owned by `Core::System`, which allows lookup, registration, de/unregistration, and complete reset. For now, the system will clear all entries on service init and register the current process only during load/start. Even though it may seem `InitializeApplicationInfo(Old)` is responsible for this, `account` sysmodule doesn't have access to `arp:w` and only can access `arp:r`, which is read-only. From my investigation, it appears the only service to have access to `arp:w`, and therefore be somehow responsible for registration is/was `ns` prior to 6.0.0 and `am` starting with 6.0.0.

This also adds scaffolding for the `bgtc:t` and `bgtc:sc`, which can be inferred to be `BackGround Task Controller`, but more RE must be done for an implementation of this.

Finally, this moves our current `arp` service scaffolds to `glue` as that is the name of the module.

This was reverse engineered in IDA against the 2.0.0 version of glue for correct functionality and error codes (`glue/errors.h`)


src/core/core.cpp
```diff
@@ -33,12 +34,39 @@
...
+        return FileSys::StorageId::NandSystem;
+    case FileSys::ContentProviderUnionSlot::SDMC:
+        return FileSys::StorageId::SdCard;
+    case FileSys::ContentProviderUnionSlot::FrontendManual:
+        return FileSys::StorageId::Host;
+    default:
+        return FileSys::StorageId::None;
+    }
+
+    UNREACHABLE();
```
mailwl: there is no way to get here (default in switch)

src/core/hle/service/glue/errors.h
```diff
@@ -0,0 +1,17 @@
...
+#pragma once
+
+#include "core/hle/result.h"
+
+namespace Service::Glue {
+
+constexpr ResultCode ERR_OUTPUT_TOO_SMALL{0x3C9D};
+constexpr ResultCode ERR_PROCESS_ID_ZERO{0x3E9D};
+constexpr ResultCode ERR_TITLE_ID_ZERO{0x3E9D};
+constexpr ResultCode ERR_ALREADY_ISSUED{0x549D};
```
lioncash: This is likely used as a general-purpose invalid access code, rather than specifically being about things already being issued.

src/core/hle/service/glue/arp.cpp
```diff
@@ -0,0 +1,285 @@
...
+    IPC::RequestParser rp{ctx};
+    const auto process_id = rp.PopRaw<u64>();
+
+    LOG_DEBUG(Service_ARP, "called, process_id={:016X}", process_id);
+
+    const auto title_id = GetTitleIDForProcessID(system, process_id);
+    if (!title_id.has_value()) {
+        LOG_ERROR(Service_ARP, "Failed to get title ID for process ID!");
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(ERR_NONEXISTENT);
```
lioncash: You likely meant to return here. Ditto for a few other error paths in other service functions.

src/core/file_sys/registered_cache.cpp
```diff
@@ -645,6 +645,20 @@ ContentProviderUnion::ListEntriesFilterOrigin(std::optional<ContentProviderUnion
     return out;
 }
 
+std::optional<ContentProviderUnionSlot> ContentProviderUnion::GetSlotForEntry(
+    u64 title_id, ContentRecordType type) const {
+    for (const auto& [slot, provider] : providers) {
```
lioncash: This seems like it can be a `std::find_if` call.

src/core/hle/service/glue/arp.h
```diff
@@ -0,0 +1,43 @@
...
+#include "core/hle/service/service.h"
+
+namespace Service::Glue {
+
+class ARPManager;
+class IRegistrar;
+
+class ARP_R final : public ServiceFramework<ARP_R> {
+public:
+    explicit ARP_R(const Core::System& system, const ARPManager& manager);
```
lioncash: Core::System needs to be forward-declared in this header.

src/core/hle/service/glue/manager.h
```diff
@@ -0,0 +1,46 @@
...
+    ResultVal<std::vector<u8>> GetControlProperty(u64 title_id) const;
+
+    ResultCode Register(u64 title_id, ApplicationLaunchProperty launch, std::vector<u8> control);
+
+    ResultCode Unregister(u64 title_id);
+
+    void ResetAll();
+
+private:
+    struct MapEntry {
```
lioncash: I believe you can change this into a forward declaration like:

```cpp
class Thing {
private:
    struct Stuff;
    std::map<u64, Stuff> map;
};

// cpp
struct Thing::Stuff {
    // ...
};
```

src/core/hle/service/glue/manager.h
```diff
@@ -0,0 +1,46 @@
+// Copyright 2019 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "core/file_sys/control_metadata.h"
```
lioncash: `<map>`, `<vector>`, `"common/common_types.h"` need to be included.

The romfs_factory include can be converted into a forward declaration.

src/core/hle/service/glue/manager.h
```diff
@@ -0,0 +1,46 @@
...
+    u64 title_id;
+    u32 version;
+    FileSys::StorageId base_game_storage_id;
+    FileSys::StorageId update_storage_id;
+    INSERT_PADDING_BYTES(0x2);
+};
+static_assert(sizeof(ApplicationLaunchProperty) == 0x10,
+              "ApplicationLaunchProperty has incorrect size.");
+
+class ARPManager {
```
lioncash: This and the following public interface should have documentation.

src/core/hle/service/glue/arp.h
DarkLordZach: It is, in the service.h header

src/core/hle/service/glue/manager.cpp
```diff
@@ -0,0 +1,73 @@
...
+    if (title_id == 0) {
+        return ERR_TITLE_ID_ZERO;
+    }
+
+    const auto iter = entries.find(title_id);
+    if (iter == entries.end()) {
+        return ERR_NONEXISTENT;
+    }
+
+    return MakeResult<ApplicationLaunchProperty>(iter->second.launch);
```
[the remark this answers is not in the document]
DarkLordZach: Fails to compile on MSVC, due to the iter->second being const in this context, MSVC seems unable to get rid of the const on its own. Ditto for succeeding comment

PR #2608, opened 2019-06-25
Implement Time::GetSharedMemoryNativeHandle

### LINKED ISSUE #2556: Missing HLE service command 'GetSharedMemoryNativeHandle' in 'time:u'
Stuck on launching screen, speed >5000%, game fps =0
thief

Log file:
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/3262149/yuzu_log.txt)

ogniK5377:
This PR attempts to implement the shared memory provided by GetSharedMemoryNativeHandle. There is still more work to be done however that requires a rehaul of the current time module to handle clock contexts. This PR is mainly to get the basic functionality of the SharedMemory working and allow the use of addition to it whilst things get improved on.

Things to note:
Memory Barriers are used in the SharedMemory and a better solution would need to be done to implement this. Currently in this PR I’m faking the memory barriers as everything is sync and single threaded. They work by incrementing the counter and just populate the two data slots. On data reading, it will read the last added data.

Specific values in the shared memory would need to be updated periodically. This isn't included in this PR since we don't actively do this yet. In a later PR when time is refactored this should be done.

Finally, as we don't handle clock contexts. When time is refactored, we will need to update the shared memory for specific contexts. This PR does this already however since the contexts are all identical and not separated. We're just updating the same values for each context which in this case is empty.

Tiime:SetStandardUserSystemClockAutomaticCorrectionEnabled, Time:IsStandardUserSystemClockAutomaticCorrectionEnabled are also partially implemented in this PR. The reason the implementation is partial is because once again, a lack of clock contexts. This will be improved on in a future PR.

This PR closes #2556

### DIFF
--- a/src/core/hle/service/time/interface.h
+++ b/src/core/hle/service/time/interface.h
@@ -7,10 +7,11 @@
 #include "core/hle/service/time/time.h"
 
 namespace Service::Time {
-
+class SharedMemory;
 class Time final : public Module::Interface {
 public:
-    explicit Time(std::shared_ptr<Module> time, const char* name);
+    explicit Time(std::shared_ptr<Module> time, std::shared_ptr<SharedMemory> shared_memory,
+                  const char* name);
     ~Time() override;
 };
 
--- a/src/core/hle/service/time/interface.h
+++ b/src/core/hle/service/time/interface.h
@@ -7,7 +7,9 @@
 #include "core/hle/service/time/time.h"
 
 namespace Service::Time {
+
 class SharedMemory;
+
 class Time final : public Module::Interface {
 public:
     explicit Time(std::shared_ptr<Module> time, std::shared_ptr<SharedMemory> shared_memory,
--- a/src/core/hle/service/time/time.cpp
+++ b/src/core/hle/service/time/time.cpp
@@ -418,7 +418,8 @@ void InstallInterfaces(Core::System& system) {
 
     std::make_shared<Time>(time, shared_mem, "time:a")->InstallAsService(system.ServiceManager());
     std::make_shared<Time>(time, shared_mem, "time:s")->InstallAsService(system.ServiceManager());
-    std::make_shared<Time>(time, shared_mem, "time:u")->InstallAsService(system.ServiceManager());
+    std::make_shared<Time>(std::move(time), shared_mem, "time:u")
+        ->InstallAsService(system.ServiceManager());
 }
 
 } // namespace Service::Time
--- a/src/core/hle/service/time/time.cpp
+++ b/src/core/hle/service/time/time.cpp
@@ -13,6 +13,7 @@
 #include "core/hle/kernel/client_session.h"
 #include "core/hle/service/time/interface.h"
 #include "core/hle/service/time/time.h"
+#include "core/hle/service/time/time_sharedmemory.h"
 #include "core/settings.h"
 
 namespace Service::Time {
@@ -61,9 +62,18 @@ static u64 CalendarToPosix(const CalendarTime& calendar_time,
     return static_cast<u64>(epoch_time);
 }
 
+enum class ClockContextType {
+    StandardSteady,
+    StandardUserSystem,
+    StandardNetworkSystem,
+    StandardLocalSystem,
+};
+
 class ISystemClock final : public ServiceFramework<ISystemClock> {
 public:
-    ISystemClock() : ServiceFramework("ISystemClock") {
+    ISystemClock(std::shared_ptr<Service::Time::SharedMemory> shared_memory,
+                 ClockContextType clock_type)
+        : ServiceFramework("ISystemClock"), shared_memory(shared_memory), clock_type(clock_type) {
         static const FunctionInfo functions[] = {
             {0, &ISystemClock::GetCurrentTime, "GetCurrentTime"},
             {1, nullptr, "SetCurrentTime"},
@@ -72,6 +82,8 @@ public:
 
         };
         RegisterHandlers(functions);
+
+        UpdateSharedMemoryContext(system_clock_context);
     }
 
 private:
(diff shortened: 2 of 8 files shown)


FernandoS27:
So glad this has a proper implementation. There's nothing worse than stubbing shared memory.

Hexagon12:
The AppVeyor build fail is unrelated to the PR - it timed out

src/core/hle/service/time/time.cpp
```diff
@@ -333,16 +374,51 @@ void Module::Interface::CalculateStandardUserSystemClockDifferenceByUser(
...
+void InstallInterfaces(Core::System& system) {
     auto time = std::make_shared<Module>();
-    std::make_shared<Time>(time, "time:a")->InstallAsService(service_manager);
-    std::make_shared<Time>(time, "time:s")->InstallAsService(service_manager);
-    std::make_shared<Time>(time, "time:u")->InstallAsService(service_manager);
+    auto shared_mem = std::make_shared<SharedMemory>(system);
+
+    std::make_shared<Time>(time, shared_mem, "time:a")->InstallAsService(system.ServiceManager());
+    std::make_shared<Time>(time, shared_mem, "time:s")->InstallAsService(system.ServiceManager());
+    std::make_shared<Time>(time, shared_mem, "time:u")->InstallAsService(system.ServiceManager());
```
DarkLordZach: std::move the shared_mem into the last one (time:u)

src/core/hle/service/time/time_sharedmemory.h
```diff
@@ -0,0 +1,71 @@
...
+            return data[(read_attempt - 1) & 1];
+        }
+    };
+
+    // Shared memory format
+    struct Format {
+        MemoryBarrier<SteadyClockTimePoint> standard_steady_clock_timepoint;
+        MemoryBarrier<SystemClockContext> standard_local_system_clock_context;
+        MemoryBarrier<SystemClockContext> standard_network_system_clock_context;
+        MemoryBarrier<u8> standard_user_system_clock_automatic_correction;
```
DarkLordZach: can you make this a `MemoryBarrier<bool>`? It would make the code cleaner and intent clearer

src/core/hle/service/time/time_sharedmemory.h
```diff
@@ -0,0 +1,71 @@
...
+        u32_le format_version;
+    };
+    static_assert(sizeof(Format) == 0xd8, "Format is an invalid size");
+
+private:
+    const std::size_t SHARED_MEMORY_SIZE = 0x1000;
```
DarkLordZach: move to cpp file and make constexpr

src/core/hle/service/time/time_sharedmemory.cpp
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/core.h"
+#include "core/hle/service/time/time_sharedmemory.h"
+
+namespace Service::Time {
+
+SharedMemory::SharedMemory(Core::System& system) : system(system) {
+    shared_memory_holder = Kernel::SharedMemory::Create(
```
DarkLordZach: Can this operation fail? If so, we need to handle the error esp. considering it would have us write to nullptr if it failed.

src/core/hle/service/time/time_sharedmemory.h
```diff
@@ -0,0 +1,71 @@
...
+    ~SharedMemory();
+
+    // Return the shared memory handle
+    Kernel::SharedPtr<Kernel::SharedMemory> GetSharedMemoryHolder() const;
+
+    // Set memory barriers in shared memory and update them
+    void SetStandardSteadyClockTimepoint(const SteadyClockTimePoint& timepoint);
+    void SetStandardLocalSystemClockContext(const SystemClockContext& context);
+    void SetStandardNetworkSystemClockContext(const SystemClockContext& context);
+    void SetStandardUserSystemClockAutomaticCorrectionEnabled(const bool enabled);
```
DarkLordZach: `const bool` = `bool` for the intent of a declaration. it only affects the definition (you don't need it here).

src/core/hle/service/time/interface.h
```diff
@@ -7,10 +7,11 @@
 #include "core/hle/service/time/time.h"
 
 namespace Service::Time {
-
+class SharedMemory;
```
lioncash: There should be newlines on both sides of the forward declaration so things aren't smushed together.

```suggestion
namespace ServiceTime {

class SharedMemory;

class Time final : public Module::Interface {
// ...
```

src/core/hle/service/time/time.h
```diff
@@ -9,6 +9,7 @@
 #include "core/hle/service/service.h"
 
 namespace Service::Time {
+class SharedMemory;
```
lioncash: Ditto

src/core/hle/service/time/time_sharedmemory.cpp
```diff
@@ -0,0 +1,59 @@
+// Copyright 2018 yuzu emulator team
```
lioncash: 2019

src/core/hle/service/time/time_sharedmemory.h
```diff
@@ -0,0 +1,71 @@
+// Copyright 2018 yuzu emulator team
```
lioncash: 2019

src/core/hle/service/time/time_sharedmemory.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Service::Time {
+
+SharedMemory::SharedMemory(Core::System& system) : system(system) {
+    shared_memory_holder = Kernel::SharedMemory::Create(
+        system.Kernel(), nullptr, SHARED_MEMORY_SIZE, Kernel::MemoryPermission::ReadWrite,
+        Kernel::MemoryPermission::Read, 0, Kernel::MemoryRegion::BASE, "Time:SharedMemory");
+    shared_memory_format = reinterpret_cast<Format*>(shared_memory_holder->GetPointer());
+    shared_memory_format->format_version =
+        14; // Seems static from 1.0.0 -> 8.1.0. Specific games seem to check this value and crash
```
lioncash: This comment should go above this line to make the formatting less wonky.

src/core/hle/service/time/time_sharedmemory.h
```diff
@@ -0,0 +1,71 @@
...
+    SteadyClockTimePoint GetStandardSteadyClockTimepoint() const;
+    SystemClockContext GetStandardLocalSystemClockContext() const;
+    SystemClockContext GetStandardNetworkSystemClockContext() const;
+    bool GetStandardUserSystemClockAutomaticCorrectionEnabled() const;
+
+    // TODO(ogniK): We have to properly simulate memory barriers, how are we going to do this?
+    template <typename T>
+    struct MemoryBarrier {
+        u32_le read_attempt{};
+        T data[2]{};
```
lioncash: std::array

src/core/hle/service/time/time_sharedmemory.h
```diff
@@ -0,0 +1,71 @@
...
+#pragma once
+
+#include "common/common_types.h"
+#include "core/hle/kernel/shared_memory.h"
+#include "core/hle/service/time/time.h"
+
+namespace Service::Time {
+class SharedMemory {
+public:
+    SharedMemory(Core::System& system);
```
lioncash:
```suggestion
    explicit SharedMemory(Core::System& system);
```

src/core/hle/service/time/time_sharedmemory.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "core/core.h"
+#include "core/hle/service/time/time_sharedmemory.h"
+
+namespace Service::Time {
+
+SharedMemory::SharedMemory(Core::System& system) : system(system) {
+    shared_memory_holder = Kernel::SharedMemory::Create(
+        system.Kernel(), nullptr, SHARED_MEMORY_SIZE, Kernel::MemoryPermission::ReadWrite,
+        Kernel::MemoryPermission::Read, 0, Kernel::MemoryRegion::BASE, "Time:SharedMemory");
+    shared_memory_format = reinterpret_cast<Format*>(shared_memory_holder->GetPointer());
```
lioncash: Dereferencing this pointer is undefined behavior, since it breaks strict aliasing rules. You can't arbitrarily type pun like this.

PR #2611, opened 2019-06-27
pm: Implement various pm commands for finding process and title IDs

DarkLordZach:
This implements various commands in `pm:shell`, `pm:dmnt`, and `pm:info` which map between process and title IDs.

- `pm:info GetTitleId(u64 process_id) -> u64 title_id` 
    This takes a process ID and returns the title ID it possesses, if the process ID doesn't correspond to a valid process returns an error code.
- `pm:dmnt GetTitlePid(u64 title_id) -> u64 process_id`
    The inverse of the preceding command, this takes a title ID and returns the process ID if there exists a process with that title ID, otherwise error.
- `pm:dmnt GetApplicationPid() -> u64 process_id` & `pm:shell GetApplicationPid() -> u64 process_id`
    These both return the process ID of the current application, and if no application is running they return 0, which is an invalid process ID.

### DIFF
--- a/src/core/hle/service/pm/pm.cpp
+++ b/src/core/hle/service/pm/pm.cpp
@@ -28,6 +28,17 @@ std::optional<Kernel::SharedPtr<Kernel::Process>> SearchProcessList(
     return *iter;
 }
 
+void GetApplicationPidGeneric(Kernel::HLERequestContext& ctx,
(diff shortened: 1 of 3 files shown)


SciresM:
Worth noting: on < 4.0.0 0 is a valid process ID.

DarkLordZach:
@SciresM That makes sense, I reversed 8.0.1 for this PR.

PR #2650, opened 2019-06-29
mii: Implement IDatabaseService SetInterfaceVersion

### LINKED ISSUE #2619: Missing HLE service command 'SetInterfaceVersion' in 'IDatabaseService' (mii:u)
* Used by Super Mario Maker 2

<meta name="titleids"  content="01009B90006DC000">

[Reporter log](https://github.com/yuzu-emu/yuzu/files/3340474/01009B90006DC000_2019-06-28T21-52-21.txt)

DarkLordZach:
Sets a member variable used to affect the API that games access, and the method used to store data.

Closes #2619

### DIFF
--- a/src/core/hle/service/mii/mii.cpp
+++ b/src/core/hle/service/mii/mii.cpp
@@ -48,7 +48,7 @@ public:
             {19, nullptr, "Export"},
             {20, nullptr, "IsBrokenDatabaseWithClearFlag"},
             {21, &IDatabaseService::GetIndex, "GetIndex"},
-            {22, nullptr, "SetInterfaceVersion"},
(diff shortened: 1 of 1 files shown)


src/core/hle/service/mii/mii.cpp
```diff
@@ -350,8 +350,20 @@ class IDatabaseService final : public ServiceFramework<IDatabaseService> {
         rb.Push(index);
     }
 
+    void SetInterfaceVersion(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        current_interface_version = rp.PopRaw<u32>();
```
ogniK5377: We should assert if we don't support a specific version so we know when games start changing their interface version

PR #2651, opened 2019-06-29
apm: Initial implementation of performance config and boost mode

### LINKED ISSUE #2616: Missing HLE service command 'SetCpuBoostMode' in 'ICommonStateGetter' (AM)
use by Super mario maker 2
unknown

<meta name="titleids"  content="01009B90006DC000">

DarkLordZach:
This implements the getter and setter methods for cpu boost mode in `apm` and `am` services, along with the getters and setters for PerformanceConfig, which is a more granular way of changing speed (and the backend of boost mode).

This also adds the `apm:am` service, which is a rename of `apm` for 8.0.0+, otherwise identical.

For now, this adds the framework to handle setting/getting config and mode and has a function which is called to change CPU speed and for now this simply logs to console.
From my testing, changing emulated clock speed in yuzu will not yield any difference to the user, sometimes even making performance worse.

Closes #2616

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -207,6 +207,8 @@ add_library(core STATIC
     hle/service/aoc/aoc_u.h
     hle/service/apm/apm.cpp
     hle/service/apm/apm.h
+    hle/service/apm/controller.cpp
+    hle/service/apm/controller.h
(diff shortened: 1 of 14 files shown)


DarkLordZach:
Closes #2616

Issue #2610, opened 2019-06-25
Windows mingw builds don't contain the yuzu executable anymore


FearlessTobi:
This is a regression, recently introduced by https://github.com/yuzu-emu/yuzu/pull/2474.

Sample error log: https://travis-ci.org/yuzu-emu/yuzu-canary/jobs/550466063

fluffapparition:
How the hell did this even happen :female_facepalm:

FearlessTobi:
This is fixed.

PR #2653, opened 2019-06-29
Revert "CMake: Get Git submodule dependencies via CMake"

FearlessTobi:
Reverts yuzu-emu/yuzu#2474.

As no one has shown interest in fixing https://github.com/yuzu-emu/yuzu/issues/2610, and it was caused by this PR, I think reverting is the best option for now. 

cc @FreddyFunk


DarkLordZach:
For now this is the best course of action.

PR #2654, opened 2019-06-29
lm: Rewrite logger to use core reporting services

DarkLordZach:
Firstly, this renames command 0 on `ILogger` to `Log`, as opposed to `Initialize`, which was an early unofficial name.

This also now outputs log data to the reporter instead of just outputting them to console. The way LM log packets are structured, there is a head log and tail log which signal the start and end of a stream of messages. All messages in a stream will output to one report, as an array of packets. Additionally, the core will automatically flush any non-tailed log packets on shutdown in case of an unexpected failure.

### DIFF
--- a/src/core/hle/service/lm/lm.cpp
+++ b/src/core/hle/service/lm/lm.cpp
@@ -6,8 +6,10 @@
 #include <string>
 
 #include "common/logging/log.h"
+#include "common/scope_exit.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/service/lm/lm.h"
+#include "core/hle/service/lm/manager.h"
 #include "core/hle/service/service.h"
 #include "core/memory.h"
 
@@ -194,31 +196,30 @@ private:
 
 class LM final : public ServiceFramework<LM> {
 public:
-    explicit LM() : ServiceFramework{"lm"} {
+    explicit LM(Manager& manager) : ServiceFramework{"lm"}, manager(manager) {
+        // clang-format off
         static const FunctionInfo functions[] = {
-            {0x00000000, &LM::OpenLogger, "OpenLogger"},
+            {0, &LM::OpenLogger, "OpenLogger"},
         };
+        // clang-format on
+
         RegisterHandlers(functions);
     }
 
-    /**
-     * LM::OpenLogger service function
-     *  Inputs:
-     *      0: 0x00000000
-     *  Outputs:
-     *      0: ResultCode
-     */
+private:
     void OpenLogger(Kernel::HLERequestContext& ctx) {
         LOG_DEBUG(Service_LM, "called");
 
         IPC::ResponseBuilder rb{ctx, 2, 0, 1};
         rb.Push(RESULT_SUCCESS);
-        rb.PushIpcInterface<ILogger>();
+        rb.PushIpcInterface<ILogger>(manager);
     }
+
+    Manager& manager;
 };
 
-void InstallInterfaces(SM::ServiceManager& service_manager) {
-    std::make_shared<LM>()->InstallAsService(service_manager);
+void InstallInterfaces(Core::System& system) {
+    std::make_shared<LM>(system.GetLogManager())->InstallAsService(system.ServiceManager());
 }
 
 } // namespace Service::LM
--- a/src/core/hle/service/lm/lm.cpp
+++ b/src/core/hle/service/lm/lm.cpp
@@ -17,65 +17,16 @@ namespace Service::LM {
 
 class ILogger final : public ServiceFramework<ILogger> {
 public:
-    ILogger() : ServiceFramework("ILogger") {
+    ILogger(Manager& manager) : ServiceFramework("ILogger"), manager(manager) {
         static const FunctionInfo functions[] = {
-            {0x00000000, &ILogger::Initialize, "Initialize"},
-            {0x00000001, &ILogger::SetDestination, "SetDestination"},
+            {0, &ILogger::Log, "Log"},
+            {1, &ILogger::SetDestination, "SetDestination"},
         };
         RegisterHandlers(functions);
     }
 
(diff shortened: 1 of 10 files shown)


bunnei:
@DarkLordZach needs a rebase

src/core/hle/service/lm/manager.h
```diff
@@ -0,0 +1,106 @@
...
+struct LogMessage {
+    MessageHeader header;
+    FieldMap fields;
+};
+
+std::string FormatField(Field type, const std::vector<u8>& data);
+
+class Manager {
+public:
+    Manager(Core::Reporter& reporter);
```
lioncash:
```suggestion
    explicit Manager(Core::Reporter& reporter);
```

src/core/reporter.cpp
```diff
@@ -350,6 +352,55 @@ void Reporter::SaveErrorReport(u64 title_id, ResultCode result,
...
+                                      std::back_inserter(fields), [](const auto& kv) {
+                                          json out;
+                                          out["type"] = fmt::format("{}", kv.first);
+                                          out["data"] =
+                                              Service::LM::FormatField(kv.first, kv.second);
+                                          return std::move(out);
+                                      });
+
+                       out["fields"] = std::move(fields);
+                       return std::move(out);
```
lioncash: This is a pessimizing move.

src/core/hle/service/lm/manager.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void Manager::Flush() {
+    FinalizeLog();
+}
+
+void Manager::InitializeLog() {
+    current_log.clear();
+
+    LOG_INFO(Service_LM, "Initialized new log session!");
```
lioncash: This doesn't need an exclamation point. There's nothing to really justify the equivalent of yelling over a log here. Ditto for FinalizeLog.

src/core/hle/service/lm/manager.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+std::string FormatField(Field type, const std::vector<u8>& data) {
+    switch (type) {
+    case Field::Skip:
+        return "";
+    case Field::Line:
+        if (data.size() >= sizeof(u32)) {
+            u32 line;
+            std::memcpy(&line, data.data(), sizeof(u32));
+            return fmt::format("{:08X}", line);
```
lioncash: It's generally not beneficial to format a line number as a hex value. Even IDEs don't display line numbers as hex by default (for good reason).

src/core/hle/service/lm/manager.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include "common/string_util.h"
+#include "core/hle/service/lm/manager.h"
+#include "core/reporter.h"
+
+namespace Service::LM {
+
+std::ostream& operator<<(std::ostream& os, DestinationFlag dest) {
+    std::vector<std::string> array;
+    const auto check_single_flag = [dest, &array](DestinationFlag check, std::string name) {
+        if ((static_cast<u32>(check) & static_cast<u32>(dest)) > 0) {
```
lioncash:
```suggestion
        if ((static_cast<u32>(check) & static_cast<u32>(dest)) != 0) {
```

is the idiomatic way to test bit flags.

src/core/reporter.cpp
```diff
@@ -350,6 +352,55 @@ void Reporter::SaveErrorReport(u64 title_id, ResultCode result,
...
+                       out["verbosity"] = fmt::format("{:02X}", message.header.verbosity);
+
+                       auto fields = json::array();
+                       std::transform(message.fields.begin(), message.fields.end(),
+                                      std::back_inserter(fields), [](const auto& kv) {
+                                          json out;
+                                          out["type"] = fmt::format("{}", kv.first);
+                                          out["data"] =
+                                              Service::LM::FormatField(kv.first, kv.second);
+                                          return std::move(out);
```
lioncash: This is a pessimizing move.

src/core/hle/service/lm/lm.cpp
```diff
@@ -85,140 +38,70 @@ class ILogger final : public ServiceFramework<ILogger> {
...
-                break;
-            case MessageHeader::Severity::Critical:
-                LOG_CRITICAL(Debug_Emulated, "{}", log_stream.str());
-                break;
-            }
-        }
+        manager.Log({header, std::move(fields)});
     }
 
-    // This service function is intended to be used as a way to
```
lioncash: This is still correct as far as the service itself is concerned. UART and UART sleeping are pointless to adhere to in an emulated context. You still want to see those logging messages regardless.

src/core/hle/service/lm/lm.cpp
```diff
@@ -85,140 +38,70 @@ class ILogger final : public ServiceFramework<ILogger> {
...
-                function = Memory::ReadCString(addr, length);
-                break;
-            case Field::Module:
-                module = Memory::ReadCString(addr, length);
-                break;
-            case Field::Thread:
-                thread = Memory::ReadCString(addr, length);
-                break;
-            }
+            SCOPE_EXIT({ addr += length; });
```
lioncash: This can just be moved to the end of the scope

MysticExile:
@DarkLordZach needs a rebase

src/core/hle/service/lm/lm.cpp
DarkLordZach: Its because of the continue; and the use of addr later, we would need it twice

src/core/hle/service/lm/lm.cpp
DarkLordZach: Yeah, the reporter will still log regardless of destination. It just includes destination as a field in the JSON.
Comment was removed as it seemed irrelevant/self-explanatory

PR #2660, opened 2019-07-01
file_sys: Ignore DeltaFragment NCAs during installation

bakugo:
Fixes the installation error and warnings when installing a patch NSP without DeltaFragments. Many patch NSPs do not include these even if they're listed in the CNMT, since they are not useful outside of Nintendo's official patch installation process on real consoles, and there is no need to install them when they're present.

Also, I only renamed `ContentRecordType::Patch` since it has the most potential to cause confusion, but it may be a good idea to rename the rest of the values to their more common names as well (Manual -> HtmlDocument, Legal -> LegalInformation). I'm not 100% sure but I believe these names are official.

### DIFF
--- a/src/core/file_sys/nca_metadata.h
+++ b/src/core/file_sys/nca_metadata.h
@@ -37,7 +37,7 @@ enum class ContentRecordType : u8 {
     Control = 3,
     Manual = 4,
     Legal = 5,
-    Patch = 6,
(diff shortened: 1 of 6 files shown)


DarkLordZach:
If you wouldn't mind, could you rename all of the other members as you described?

bakugo:
Done.

PR #2661, opened 2019-07-01
audren: Only manage wave buffers with a size

ogniK5377:
We shouldn't be incrementing if wave buffers are empty. They are considered invalid/unused wave buffers.

This fixes the issue of certain sounds looping when they shouldn't

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -217,13 +217,15 @@ std::vector<s16> AudioRenderer::VoiceState::DequeueSamples(std::size_t sample_co
     if (offset == samples.size()) {
         offset = 0;
 
-        if (!wave_buffer.is_looping) {
+        if (!wave_buffer.is_looping && wave_buffer.buffer_sz) {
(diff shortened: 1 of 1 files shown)


src/audio_core/audio_renderer.cpp
```diff
@@ -217,13 +217,15 @@ std::vector<s16> AudioRenderer::VoiceState::DequeueSamples(std::size_t sample_co
...
             SetWaveIndex(wave_index + 1);
         }
 
-        out_status.wave_buffer_consumed++;
+        if (wave_buffer.buffer_sz) {
+            out_status.wave_buffer_consumed++;
+        }
 
-        if (wave_buffer.end_of_stream) {
+        if (wave_buffer.end_of_stream || !wave_buffer.buffer_sz) {
```
lioncash:
```suggestion
        if (wave_buffer.end_of_stream || wave_buffer.buffer_sz == 0) {
```

communicates the same intent more explicitly, given it's not actually a `bool`, but actually an integral type.

bunnei:
@ogniK5377 can you address the comments here?

PR #2669, opened 2019-07-04
yuzu: Move CPU Jit setting to Debug tab

FearlessTobi:
A normal user shouldn't change this, as it will slow down the emulation and can lead to bugs or crashes. So let's move the setting to Debug, like we do for Citra already.


bunnei:
Can we just move it out of the UI altogether?

FearlessTobi:
Sure.

PR #2674, opened 2019-07-05
core/reporter: Minor changes

lioncash:
Performs minor clean-up/correctness-related changes, some preventing potential compilation issues in the future.


src/core/reporter.cpp
```diff
@@ -34,7 +34,7 @@ std::string GetTimestamp() {
 
 using namespace nlohmann;
 
-void SaveToFile(const json& json, const std::string& filename) {
```
VPeruS: Why not to remove std::move on caller side instead?

PR #2675, opened 2019-07-06
buffer_cache: Implement a generic buffer cache and its OpenGL backend

ReinUsesLisp:
This does:
* Implement an API agnostic buffer cache.
* Implement the OpenGL backend for that buffer cache.
* Support global memory writes to be shared with other buffer types (vertex, index and constants).
* Drop `gl_global_cache`, this is no longer required since all buffers go through the buffer cache.

This does **not**:
* Share buffer memory allocations with texture buffers.
* Implement a ranged buffer cache to support the usage of "buffer views" (buffers used with an offset).
* Invalidate or flush in ranges.

There are two major reasons these changes are needed: to avoid duplication code between graphics backends and to properly support shaders writing to global memory. This is important for transform feedbacks and compute shaders (or any shader writing to global memory).

Buffers marked for destruction are mainly there for Vulkan, these are destroyed after 4 frames.

I tried to make it as performant as possible without losing accuracy, but it might have performance regressions. This requires testing.

The commit history walks through the steps of writing an specific implementation on OpenGL, moving it to an API agnostic class and then implementing that abstract class on OpenGL.


FernandoS27:
LGTM so far. I suggest marking this for canary before merging it. Atleast 1-2 weeks testing time.

src/video_core/renderer_opengl/gl_buffer_cache.cpp
```diff
@@ -2,103 +2,57 @@
...
-        auto entry = std::make_shared<CachedBufferEntry>(
-            *memory_manager.GpuToCpuAddress(gpu_addr), size, uploaded_offset, alignment, host_ptr);
-        Register(entry);
-    }
-
-    return uploaded_offset;
+OGLBuffer OGLBufferCache::CreateBuffer(std::size_t size) {
+    OGLBuffer buffer;
+    buffer.Create();
+    glNamedBufferData(buffer.handle, static_cast<GLsizeiptr>(size), nullptr, GL_DYNAMIC_DRAW);
```
FernandoS27: Why is this GL_DYNAMIC_DRAW ?

ReinUsesLisp: Because it's not wildly updated to be `GL_STREAM_DRAW` and it's updated from to time time so `GL_STATIC_DRAW` might be harmful.

FernandoS27:
It seems to cause a regression in AMD cap kingdom. It should be checked.

PR #2676, opened 2019-07-06
Implement a Host based GPU Clock.

FernandoS27:
The GPU runs at different clockspeeds and paces than the CPU. It requires its own timer in order to keep it at check. Sadly a cycle timer is unfeasible for the GPU since it's pretty hard to accurately estimate the length of time actions like Draw, DispatchCompute and Clear can take (varying very unproportionally)
Using the CPU's clock also has it's disadvantages:

- If the CPU is idle while the GPU is running, the timer will advance unproportionally due to how CPU's idle timing works.

- If the GPU is synced, the timer won't advance until control is given back to the CPU.

For all these reasons, it has been decided to use a host timer for the GPU.

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -9,6 +9,7 @@
 #include "core/core_timing.h"
 #include "video_core/debug_utils/debug_utils.h"
 #include "video_core/engines/maxwell_3d.h"
+#include "video_core/gpu_clock.h"
 #include "video_core/memory_manager.h"
 #include "video_core/rasterizer_interface.h"
 #include "video_core/textures/texture.h"
@@ -330,7 +331,7 @@ void Maxwell3D::ProcessQueryGet() {
             LongQueryResult query_result{};
             query_result.value = result;
             // TODO(Subv): Generate a real GPU timestamp and write it here instead of CoreTiming
-            query_result.timestamp = system.CoreTiming().GetTicks();
+            query_result.timestamp = system.GPU().GetClock().GetNsTime().count();
             memory_manager.WriteBlock(sequence_address, &query_result, sizeof(query_result));
         }
         dirty_flags.OnMemoryWrite();
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -330,8 +330,7 @@ void Maxwell3D::ProcessQueryGet() {
             // wait queues.
             LongQueryResult query_result{};
             query_result.value = result;
-            // TODO(Subv): Generate a real GPU timestamp and write it here instead of CoreTiming
-            query_result.timestamp = system.GPU().GetClock().GetNsTime().count();
+            query_result.timestamp = system.GPU().Clock().GetNsTime().count();
             memory_manager.WriteBlock(sequence_address, &query_result, sizeof(query_result));
         }
         dirty_flags.OnMemoryWrite();
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -296,7 +296,7 @@ void GPU::ProcessSemaphoreTriggerMethod() {
         block.sequence = regs.semaphore_sequence;
         // TODO(Kmather73): Generate a real GPU timestamp and write it here instead of
         // CoreTiming
-        block.timestamp = Core::System::GetInstance().CoreTiming().GetTicks();
+        block.timestamp = clock->GetNsTime().count();
         memory_manager->WriteBlock(regs.semaphore_address.SemaphoreAddress(), &block,
                                    sizeof(block));
     } else {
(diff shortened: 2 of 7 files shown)


src/video_core/gpu_clock.h
```diff
@@ -0,0 +1,55 @@
...
+        Clock::time_point now = Clock::now();
+        return std::chrono::duration_cast<std::chrono::microseconds>(now - start_walltime);
+    }
+
+    std::chrono::milliseconds GetMsTime() const {
+        Clock::time_point now = Clock::now();
+        return std::chrono::duration_cast<std::chrono::milliseconds>(now - start_walltime);
+    }
+
+    u64 GetTicks() const;
```
ReinUsesLisp: This should be moved up, to avoid mixing inlined declarations with cpp definitions.

src/video_core/gpu_clock.h
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    std::chrono::microseconds GetUsTime() const {
+        Clock::time_point now = Clock::now();
+        return std::chrono::duration_cast<std::chrono::microseconds>(now - start_walltime);
+    }
+
+    std::chrono::milliseconds GetMsTime() const {
+        Clock::time_point now = Clock::now();
+        return std::chrono::duration_cast<std::chrono::milliseconds>(now - start_walltime);
```
ReinUsesLisp: `now` should be `const` or inlined inside the arguments:
```cpp
Clock::now() - start_walltime
```

src/video_core/gpu_clock.h
```diff
@@ -0,0 +1,55 @@
...
+    ~GPUClock();
+
+    std::chrono::nanoseconds GetNsTime() const {
+        Clock::time_point now = Clock::now();
+        return std::chrono::duration_cast<std::chrono::nanoseconds>(now - start_walltime);
+    }
+
+    std::chrono::microseconds GetUsTime() const {
+        Clock::time_point now = Clock::now();
+        return std::chrono::duration_cast<std::chrono::microseconds>(now - start_walltime);
```
ReinUsesLisp: Ditto.

src/video_core/gpu_clock.h
```diff
@@ -0,0 +1,55 @@
...
+};
+
+class GPUClock {
+public:
+    GPUClock();
+    ~GPUClock();
+
+    std::chrono::nanoseconds GetNsTime() const {
+        Clock::time_point now = Clock::now();
+        return std::chrono::duration_cast<std::chrono::nanoseconds>(now - start_walltime);
```
ReinUsesLisp: Ditto.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -330,7 +331,7 @@ void Maxwell3D::ProcessQueryGet() {
             LongQueryResult query_result{};
             query_result.value = result;
             // TODO(Subv): Generate a real GPU timestamp and write it here instead of CoreTiming
```
ReinUsesLisp: This commentary can be removed.

src/video_core/gpu.cpp
```diff
@@ -29,7 +30,9 @@ u32 FramebufferConfig::BytesPerPixel(PixelFormat format) {
     UNREACHABLE();
 }
 
-GPU::GPU(Core::System& system, VideoCore::RendererBase& renderer) : renderer{renderer} {
+GPU::GPU(Core::System& system, VideoCore::RendererBase& renderer)
+    : renderer{renderer} {
+    clock = std::make_unique<Tegra::GPUClock>();
```
ReinUsesLisp: Assign this in order of declaration (move it to the bottom).

src/video_core/gpu_clock.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Tegra {
+
+GPUClock::GPUClock() = default;
+GPUClock::~GPUClock() = default;
+
+u64 GPUClock::GetTicks() const {
+    const u64 ns = GetNsTime().count();
+    const u128 middle = Common::Multiply64Into128(ns, gpu_clock);
+    return Common::Divide128On32(middle, 1000000000).first;
```
ReinUsesLisp: Is `1000000000` intended to be the number of nanoseconds in a second? If so, move it to a local constexpr value.

src/video_core/gpu.cpp
lioncash: This can be assigned within the constructor initializer list.

src/video_core/gpu_clock.h
FernandoS27: addressed

src/video_core/gpu_clock.h
FernandoS27: addressed

src/video_core/gpu_clock.h
FernandoS27: addressed

src/video_core/gpu.cpp
lioncash: This still applies

FernandoS27:
I'll see to this later on. Right now is not the correct time.

PR #2685, opened 2019-07-07
loader: Implement various 'information' commands and add ro:1 service

DarkLordZach:
- Implement `DebugMonitor` `GetNsoInfos` -- returns an array of NSOInfo entries which contain the build ID, size, and mapped address of all NSOs for the current application.
- Implement `ProcessManager` `GetProgramInfo` -- Parses the current application's NPDM and copies select fields into a ProgramInfo struct and returns that.
- Implement `ProcessManager` `RegisterTitle` and `UnregisterTitle` -- Upon registry the title's ID and storage are added to an internal list and the index is returned. The index can later be passed to UnregisterTitle to remove the entry.
- Add service registration for `ro:1` service. This creates the same objects as the `ldr:ro` service and was added in 7.0.0.

Researched with IDA on 8.0.1.

### DIFF
--- a/src/core/file_sys/program_metadata.cpp
+++ b/src/core/file_sys/program_metadata.cpp
@@ -48,6 +48,15 @@ Loader::ResultStatus ProgramMetadata::Load(VirtualFile file) {
         return Loader::ResultStatus::ErrorBadKernelCapabilityDescriptors;
     }
 
+    acid_file_access_raw =
+        file->ReadBytes(acid_header.fac_size, npdm_header.acid_offset + acid_header.fac_offset);
+    aci0_file_access_raw =
+        file->ReadBytes(aci_header.fah_size, npdm_header.aci_offset + aci_header.fah_offset);
+    acid_service_access_raw =
+        file->ReadBytes(acid_header.sac_size, npdm_header.acid_offset + acid_header.sac_offset);
+    aci0_service_access_raw =
+        file->ReadBytes(aci_header.sac_size, npdm_header.aci_offset + aci_header.sac_offset);
+
     return Loader::ResultStatus::Success;
 }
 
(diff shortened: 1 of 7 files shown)


src/core/file_sys/program_metadata.cpp
```diff
@@ -48,6 +48,15 @@ Loader::ResultStatus ProgramMetadata::Load(VirtualFile file) {
         return Loader::ResultStatus::ErrorBadKernelCapabilityDescriptors;
     }
 
+    acid_file_access_raw =
```
lioncash: These are all reads that can fail and should be handled accordingly.

src/core/hle/service/ldr/ldr.cpp
```diff
@@ -31,35 +39,241 @@ constexpr ResultCode ERROR_NOT_INITIALIZED{ErrorModule::Loader, 87};
...
+            LOG_ERROR(Service_LDR, "Cannot find process with given process ID!");
+            IPC::ResponseBuilder rb{ctx, 2};
+            rb.Push(ERROR_TITLE_NOT_FOUND);
+            return;
+        }
+
+        const auto& modules = (*iter)->GetModules();
+        const auto nso_infos = GetNSOInfo(modules);
+
+        write_size = std::min<std::size_t>(write_size, nso_infos.size());
```
lioncash:
```suggestion
        write_size = std::min(write_size, nso_infos.size());
```

src/core/hle/service/ldr/ldr.cpp
```diff
@@ -31,35 +39,241 @@ constexpr ResultCode ERROR_NOT_INITIALIZED{ErrorModule::Loader, 87};
...
+
+        ctx.WriteBuffer(nso_infos.data(), write_size * sizeof(NSOInfo));
+
+        IPC::ResponseBuilder rb{ctx, 3};
+        rb.Push(RESULT_SUCCESS);
+        rb.Push<u32>(write_size);
```
lioncash:
```suggestion
        rb.Push(static_cast<u32>(write_size));
```

otherwise this is doing an implicit truncation

Hexagon12:
Untagged from Canary - causes build errors on MSVC

bunnei:
@DarkLordZach reminder rebase and address any remaining comments/issues. Thank you.

lioncash:
Closing due to inactivity

PR #2688, opened 2019-07-07
Added Czech

honzapatCZ:
as title says.
PS: not sure about control_metadata

### DIFF
--- a/src/core/file_sys/control_metadata.cpp
+++ b/src/core/file_sys/control_metadata.cpp
@@ -24,6 +24,7 @@ const std::array<const char*, 15> LANGUAGE_NAMES{{
     "Korean",
     "Taiwanese",
     "Chinese",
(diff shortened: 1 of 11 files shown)


Hexagon12:
If you want to translate yuzu's user interface, please refer to our [Transifex project](https://www.transifex.com/yuzu-emulator/yuzu/).
If you want for all games to be in Czech, you're out of luck. 
That's not how you would add a language to a emulator.

honzapatCZ:
I have done so last month still not in. The option is missing. I thought that will be it.

ogniK5377:
The changes you have made do not reflect how the Nintendo Switches has them. You cannot add new languages, only Nintendo can add new languages to the switch and if games use them then it will work

honzapatCZ:
@ogniK5377 I know. I only wanted to translate the emulator.

Issue #2626, opened 2019-06-28
Missing SVCs 'MapPhysicalMemory' and 'UnmapPhysicalMemory'


Hexagon12:
* Needed for Super Smash Bros. Ultimate
* Needed for Starlink: Battle for Atlas
* Needed for Skyrim
* Needed for Mario + Rabbids (update >1.0.0)
* Needed for ひぐらしのなく頃に 奉 (Higurashi no Naku Koro ni Hō)

<meta name="titleids"  content="01006A800016E000,01002CC003FE6000,01002E500B0A0000,01000A10041EA000,010028D005A3C000,010067300059A000,0100E46003042000,0100F6A00A684000">

ogniK5377:
The actual implementation of this svc is not exactly the issue of it not working.

1. With MapPhysicalMemory, we also need to implement PersonalMnHeapSize which is at the time of writing(IsVirtualAddressMemoryEnabled). This will need to be renamed. The info is pulled from the npdm, under the field System resource size.

2. When implementing this cmd, games which were originally stubbed to 0, will now stop working. Notable examples of games which could be affected are games like Super Mario Odyssey. MapPhysicalMemory should stay in canary for a while whilst we test all previous 3.0.0+ games we have known to work before to make sure we don't have any regressions.

3. TotalPhysicalMemoryAvailableWithoutMmHeap and TotalPhysicalMemoryUsedWithoutMmHeap will also need to be updated as we require to store the personal_mn_heap size.

4. Games tend to start mapping at MapRegionBaseAddr. Currently, I believe this is the main issue to why MapPhysicalMemory games are throwing as what we call internally "Ascii Addr Reads". This will also need to be looked into and checked.

5. Our scheduler could also be a cause for some issues as the addresses which are being mapped aren't consistent as previously discussed in development. However, I believe the issue might still stem from our svcGetInfo returns for a few values as noted above.

6. Our total memory usage also is checked a few times for games and games might map too much or too little depending on what we return, we will need to correct this instead of returning a fixed value.

PR #2690, opened 2019-07-07
Implement MapPhysicalMemory/UnmapPhysicalMemory

SciresM:
This implements svcMapPhysicalMemory/svcUnmapPhysicalMemory for Yuzu, which can be used to map memory at a desired address by games since 3.0.0.

It also properly parses SystemResourceSize from NPDM, and makes information available via svcGetInfo.

This is needed for games like Super Smash Bros. and Diablo 3 -- this PR's implementation does not run into the "ASCII reads" issue mentioned in the comments of #2626, which was caused by the following bugs in Yuzu's memory management that this PR also addresses:
* Yuzu's memory coalescing does not properly merge blocks. This results in a polluted address space/svcQueryMemory results that would be impossible to replicate on hardware, which can lead to game code making the wrong assumptions about memory layout.
  * This implements better merging for AllocatedMemoryBlocks.
* Yuzu's implementation of svcMirrorMemory unprotected the entire virtual memory range containing the range being mirrored. This could lead to games attempting to map data at that unprotected range/attempting to access that range after yuzu improperly unmapped it.
  * This PR fixes it by simply calling ReprotectRange instead of Reprotect.

### DIFF
--- a/src/core/hle/kernel/process.h
+++ b/src/core/hle/kernel/process.h
@@ -168,8 +168,9 @@ public:
         return capabilities.GetPriorityMask();
     }
 
-    u32 IsVirtualMemoryEnabled() const {
-        return is_virtual_address_memory_enabled;
+    /// Gets the amount of secure memory to allocate for memory management.
+    u32 GetSystemResourceSize() const {
+        return system_resource_size;
     }
 
     /// Whether this process is an AArch64 or AArch32 process.
@@ -298,12 +299,16 @@ private:
     /// Title ID corresponding to the process
     u64 program_id = 0;
 
+    /// Specifies additional memory to be reserved for the process's memory management by the
+    /// system. When this is non-zero, secure memory is allocated and used for page table allocation
+    /// instead of using the normal global page tables/memory block management.
+    u32 system_resource_size = 0;
+
     /// Resource limit descriptor for this process
     SharedPtr<ResourceLimit> resource_limit;
 
     /// The ideal CPU core for this process, threads are scheduled on this core by default.
     u8 ideal_core = 0;
-    u32 is_virtual_address_memory_enabled = 0;
 
     /// The Thread Local Storage area is allocated as processes create threads,
     /// each TLS area is 0x200 bytes, so one page (0x1000) is split up in 8 parts, and each part
--- a/src/core/hle/kernel/process.h
+++ b/src/core/hle/kernel/process.h
@@ -173,6 +173,21 @@ public:
         return system_resource_size;
     }
 
+    /// Gets the amount of secure memory currently in use for memory management.
+    u32 GetSystemResourceUsage() const {
+        // On hardware, this returns the amount of system resource memory that has
+        // been used by the kernel. This is problematic for Yuzu to emulate, because
+        // system resource memory is used for page tables -- and yuzu doesn't really
+        // have a way to calculate how much memory is required for page tables for
+        // the current process at any given time.
+        // TODO: Is this even worth implementing? Games may retrieve this value via
+        // an SDK function that gets used + available system resource size for debug
+        // or diagnostic purposes. However, it seems unlikely that a game would make
+        // decisions based on how much system memory is dedicated to its page tables.
+        // Is returning a value other than zero wise?
+        return 0;
+    }
+
     /// Whether this process is an AArch64 or AArch32 process.
     bool Is64BitProcess() const {
         return is_64bit_process;
@@ -197,15 +212,15 @@ public:
     u64 GetTotalPhysicalMemoryAvailable() const;
 
     /// Retrieves the total physical memory available to this process in bytes,
-    /// without the size of the personal heap added to it.
-    u64 GetTotalPhysicalMemoryAvailableWithoutMmHeap() const;
+    /// without the size of the personal system resource heap added to it.
+    u64 GetTotalPhysicalMemoryAvailableWithoutSystemResource() const;
 
     /// Retrieves the total physical memory used by this process in bytes.
     u64 GetTotalPhysicalMemoryUsed() const;
 
     /// Retrieves the total physical memory used by this process in bytes,
-    /// without the size of the personal heap added to it.
-    u64 GetTotalPhysicalMemoryUsedWithoutMmHeap() const;
+    /// without the size of the personal system resource heap added to it.
+    u64 GetTotalPhysicalMemoryUsedWithoutSystemResource() const;
 
     /// Gets the list of all threads created with this process as their owner.
     const std::list<const Thread*>& GetThreadList() const {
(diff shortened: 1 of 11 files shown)


src/core/hle/kernel/process.h
```diff
@@ -298,12 +299,16 @@ class Process final : public WaitObject {
     /// Title ID corresponding to the process
     u64 program_id = 0;
 
+    /// Specifies additional memory to be reserved for the process's memory management by the
+    /// system. When this is non-zero, secure memory is allocated and used for page table allocation
+    /// instead of using the normal global page tables/memory block management.
+    u32 system_resource_size = 0;
```
lioncash: This value should also be subtracted off the returned value in `GetTotalPhysicalMemoryAvailableWithoutMmHeap()` and `GetTotalPhysicalMemoryUsedWithoutMmHeap()`

src/core/hle/kernel/vm_manager.cpp
```diff
@@ -308,6 +309,258 @@ ResultVal<VAddr> VMManager::SetHeapSize(u64 size) {
...
+        }
+
+        // If we already have the desired amount mapped, we're done.
+        if (mapped_size == size) {
+            return RESULT_SUCCESS;
+        }
+    }
+
+    // Check that we can map the memory we want.
+    const auto res_limit = Core::CurrentProcess()->GetResourceLimit();
```
lioncash: Please do not use global accessors here.  The VMManager instance already has a reference to the system object as one of its members. Use that instead.

src/core/hle/kernel/svc.cpp
```diff
@@ -946,6 +960,86 @@ static ResultCode GetInfo(Core::System& system, u64* result, u64 info_id, u64 ha
...
+        LOG_ERROR(Kernel_SVC, "Size is zero");
+        return ERR_INVALID_SIZE;
+    }
+
+    if (!(addr < addr + size)) {
+        LOG_ERROR(Kernel_SVC, "Size causes 64-bit overflow of address");
+        return ERR_INVALID_MEMORY_RANGE;
+    }
+
+    auto* const current_process = Core::CurrentProcess();
```
lioncash: This should be using the `system` parameter to retrieve the current process instead of using a global accessor function.

src/core/hle/kernel/svc.cpp
```diff
@@ -946,6 +960,86 @@ static ResultCode GetInfo(Core::System& system, u64* result, u64 info_id, u64 ha
...
```
lioncash: Ditto

lioncash:
The commit message should also contain the extended description in the PR, so it's stored in the repository's history directly. Especially given it resolves a significant issue.

src/core/hle/kernel/vm_manager.h
```diff
@@ -450,6 +451,34 @@ class VMManager final {
...
     ResultVal<VAddr> SetHeapSize(u64 size);
 
+    /// Maps memory at a given address.
+    ///
+    /// @param addr The virtual address to map memory at.
+    /// @param size The amount of memory to map.
+    ///
+    /// @note The destination address must lie within the Map region.
+    ///
+    /// @note This function requires SystemResourceSize is non-zero,
```
lioncash:
```suggestion
    /// @note This function requires that SystemResourceSize be non-zero,
```

Ditto below for UnmapPhysicalMemory too.

SciresM:
Force-pushed a commit message fix, and a commit that addresses review commentary other than 

```
These functions seem quite... cluttered and can be a lot to keep track of. I have a commit in a branch here: lioncash@61e1c6a that tries to factor out some of it. Maybe that could be used as a visual guide or something to make this slightly nicer to grok.
```

SciresM:
@lioncash I added the helpers from your branch -- I agree they're definitely cleaner.

I also removed the separate mapping + reprotecting logic in MapPhysicalMemory -- the Horizon Kernel does them separately to simplify unmapping on error, but there's no actual need for yuzu to do so and it simplifies logic to map as read-write instead of as None and reprotecting.

src/core/hle/kernel/vm_manager.cpp
```diff
@@ -604,6 +765,57 @@ VMManager::VMAIter VMManager::MergeAdjacent(VMAIter iter) {
...
+                                       right.backing_block->begin() + right.offset + right.size);
+        } else {
+            // Slow case: make a new memory block for left and right.
+            auto new_memory = std::make_shared<std::vector<u8>>();
+            new_memory->insert(new_memory->end(), left.backing_block->begin() + left.offset,
+                               left.backing_block->begin() + left.offset + left.size);
+            new_memory->insert(new_memory->end(), right.backing_block->begin() + right.offset,
+                               right.backing_block->begin() + right.offset + right.size);
+            left.backing_block = new_memory;
+            left.offset = 0;
```
FernandoS27: When can this case happen? In the context of mapPhysicalMemory and others?

SciresM: Memory block is mapped, and then part of the memory block is unmapped, or reprotected, or has SetMemoryAttribute called on it.

src/video_core/memory_manager.cpp
```diff
@@ -49,6 +52,12 @@ GPUVAddr MemoryManager::MapBufferEx(VAddr cpu_addr, u64 size) {
     const GPUVAddr gpu_addr{FindFreeRegion(address_space_base, aligned_size)};
 
     MapBackingMemory(gpu_addr, Memory::GetPointer(cpu_addr), aligned_size, cpu_addr);
+    ASSERT(Core::System::GetInstance()
```
lioncash: Again, please don't use the global system accessor. If you need access to the system instance, pass down the reference via the constructor and then make use of it here. The GPU constructor where the memory manager instance is allocated already provides such a reference.

SciresM: Okay.

lioncash: Unresolved, as, while the reference itself was added to the class, it's not actually used in place of the global accessor function.

SciresM: fixed + force-pushed. Sorry about that, hahaha.

src/core/hle/kernel/vm_manager.cpp
```diff
@@ -604,6 +770,57 @@ VMManager::VMAIter VMManager::MergeAdjacent(VMAIter iter) {
...
+        }
+
+        // Page table update is needed, because backing memory changed.
+        left.size += right.size;
+        UpdatePageTableForVMA(left);
+
+        // Update mappings for unicorn.
+        system.ArmInterface(0).UnmapMemory(left.base, left.size);
```
FernandoS27: All these should be removed since unicorn mapping is no longer in master.

SciresM: Done.

FernandoS27:
@ogniK5377 @DarkLordZach request for review.

PR #2693, opened 2019-07-07
shader/half_set_predicate: Implement missing HSETP2 variants

ReinUsesLisp:
Implement `HSETP2_C` and `HSETP2_IMM`. This also addresses a bug in `HSETP2_R` where it was using in Rb the absolute value from Ra.

### DIFF
--- a/src/video_core/shader/decode/half_set_predicate.cpp
+++ b/src/video_core/shader/decode/half_set_predicate.cpp
@@ -51,26 +51,23 @@ u32 ShaderIR::DecodeHalfSetPredicate(NodeBlock& bb, u32 pc) {
         op_b = Immediate(0);
     }
 
-    // We can't use the constant predicate as destination.
-    ASSERT(instr.hsetp2.pred3 != static_cast<u64>(Pred::UnusedIndex));
-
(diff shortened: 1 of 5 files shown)


src/video_core/shader/decode/half_set_predicate.cpp
```diff
@@ -42,9 +58,9 @@ u32 ShaderIR::DecodeHalfSetPredicate(NodeBlock& bb, u32 pc) {
 
     const OperationCode combiner = GetPredicateCombiner(instr.hsetp2.op);
     const OperationCode pair_combiner =
-        instr.hsetp2.h_and ? OperationCode::LogicalAll2 : OperationCode::LogicalAny2;
+        h_and ? OperationCode::LogicalAll2 : OperationCode::LogicalAny2;
```
FernandoS27: OperationCode::LogicalAny2 needs a fix when H_AND is not set, each lane's result is saved in each corresponding predicate.

FernandoS27: Also the first predicate needs to check if it's unused in case it only wants the result of the second lane.

You should check the results of the hardware tests I forwarded.

PR #2695, opened 2019-07-07
gl_shader_decompiler: Implement gl_ViewportIndex and gl_Layer in vertex shaders

ReinUsesLisp:
This commit implements gl_ViewportIndex and gl_Layer in vertex and geometry shaders. In the case it's used in a vertex shader, it requires ARB_shader_viewport_layer_array. This extension is available on AMD and Nvidia devices (mesa and proprietary drivers), but not available on Intel on any platform. At the moment of writing this description I don't know if this is a hardware limitation or a driver limitation.

In the case that ARB_shader_viewport_layer_array is not available, writes to these registers on a vertex shader are ignored, with the appropriate logging.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -78,7 +78,7 @@ union Attribute {
     constexpr explicit Attribute(u64 value) : value(value) {}
 
     enum class Index : u64 {
-        PointSize = 6,
+        LayerViewportPointSize = 6,
         Position = 7,
         Attribute_0 = 8,
         Attribute_31 = 39,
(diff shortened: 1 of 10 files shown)


FernandoS27:
It seems to cause a regression in AMD cap kingdom. It should be checked.

Hexagon12:
Causes a regression with Smash Ultimate - no 3D rendering with a GTX 750 Ti

fluffapparition:
Considering that this causes regression on AMD and NVidia devices it might be smart to either add an option for the user to select a method appropriate for their device or to automatically use one of the two methods based on the graphics device being used. Otherwise, we could do that and allow the user to switch to another if the one selected isnt working, just incase. 

Unless we find a fix for this on all devices, this is what I think would be reasonable

ReinUsesLisp:
> 
> 
> Considering that this causes regression on AMD and NVidia devices it might be smart to either add an option for the user to select a method appropriate for their device or to automatically use one of the two methods based on the graphics device being used. Otherwise, we could do that and allow the user to switch to another if the one selected isnt working, just incase.
> 
> Unless we find a fix for this on all devices, this is what I think would be reasonable

There's no need to do that, a device that's not able to write gl_ViewportIndex and gl_Layer from a vertex shader currently doesn't use it.

I'm currently investigating what causes the regression, but it's likely that it's a bug somewhere else in the code that just got revealed by using different viewports.

ReinUsesLisp:
It should be fixed now, it wasn't that in the end.

We'll still need to enable multiple viewports when geometry shaders are disabled to render properly on games that use `gl_ViewportIndex` (like SSBU).

PR #2696, opened 2019-07-08
gl_rasterizer: Correct clear to be set to framebuffer RT 0

FernandoS27:
Our framebuffer cache always places this RT on the slot 0, not on the rt register.


bunnei:
Actually - I'm not sure this is right, this slightly differs from the fix we merged into Patreon.

FernandoS27:
It's the same. I know it's not proper fix. In general clearing needs a rework. I got a branch in the works that reworks clearing completely.

FernandoS27:
superseded by #2735

PR #2704, opened 2019-07-10
maxwell3d: Implement Conditional Rendering

FernandoS27:
Conditional Rendering takes care of conditionaly clearing or drawing
depending on a set of queries. This PR implements the query checks to
stablish if things can be rendered or not.

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -249,6 +249,10 @@ void Maxwell3D::CallMethod(const GPU::MethodCall& method_call) {
         ProcessQueryGet();
         break;
     }
+    case MAXWELL3D_REG_INDEX(condition.mode): {
+        ProcessQueryCondition();
+        break;
+    }
     case MAXWELL3D_REG_INDEX(sync_info): {
         ProcessSyncPoint();
         break;
@@ -302,6 +306,7 @@ void Maxwell3D::ProcessQueryGet() {
         result = regs.query.query_sequence;
         break;
     default:
+        result = 1;
         UNIMPLEMENTED_MSG("Unimplemented query select type {}",
                           static_cast<u32>(regs.query.query_get.select.Value()));
     }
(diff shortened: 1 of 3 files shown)


src/video_core/engines/maxwell_3d.h
```diff
@@ -89,13 +89,25 @@ class Maxwell3D final {
 
         enum class QuerySelect : u32 {
             Zero = 0,
+            TimeElapsed = 2,
```
Hexagon12: Please list these in order. I've also heard that query mode 26 is also known (transform feedback related), add that too.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -342,6 +347,43 @@ void Maxwell3D::ProcessQueryGet() {
...
+        break;
+    }
+    case Regs::ConditionMode::ResNonZero: {
+        std::array<u32, 2> data;
+        memory_manager.ReadBlockUnsafe(condition_address, &data, sizeof(data));
+        execute_on = data[0] != 0U && data[1] != 0U;
+        break;
+    }
+    case Regs::ConditionMode::Equal: {
+        std::array<u32, 6> data;
```
ReinUsesLisp: If we know what these values mean, this should be passed as a structure.

src/video_core/engines/maxwell_3d.h
ReinUsesLisp: I'm fine with just ordering it. We don't exactly know what query 26 is and if we can stub it or not.

src/video_core/engines/maxwell_3d.cpp
FernandoS27: we kinda don't know them, nouveau uses that but I can have a better look.

FernandoS27:
Ok I've addressed the comments. Changes: comparison now uses a struct instead of an array of data from what I understood they were in nouveau, I ordered the query select types and added query 26 which is related to transform feedback.

PR #2709, opened 2019-07-10
system_archive: Move shared font data to system_archive and fix extended font data

DarkLordZach:
This:
- Moves the shared font raw data to system_archive/data, uses the same namespacing as the rest of the system archives and adds the shared_font.cpp/h system archive synthesizer to make it as uniform as possible with the other OSS data archives.
- PL:U will now load the RomFs(s) of data archives from either the OSS synthesizer or real NAND, eliminating an entire code path.
- Our OSS version of `FontNintendoExtended[2]` was missing icons for controller buttons -- which caused players on SMM2 using OSS archives to not see the Press ZL+ZR and instead only Press +

image


DarkLordZach:
Huge diff is due to moving of binary files stored as arrays -- actual code is probably sub +200

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -161,114 +158,50 @@ PL_U::PL_U() : ServiceFramework("pl:u"), impl{std::make_unique<Impl>()} {
...
         }
+        std::vector<u32> font_data_u32(font_fp->GetSize() / sizeof(u32));
+        font_fp->ReadBytes<u32>(font_data_u32.data(), font_fp->GetSize());
+        // We need to be BigEndian as u32s for the xor encryption
+        std::transform(font_data_u32.begin(), font_data_u32.end(), font_data_u32.begin(),
+                       Common::swap32);
+        FontRegion region{
+            static_cast<u32>(offset + 8),
+            static_cast<u32>((font_data_u32.size() * sizeof(u32)) -
+                             8)}; // Font offset and size do not account for the header
```
lioncash: This comment should probably be relocated. Formatting is pretty weird otherwise.

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -161,114 +158,50 @@ PL_U::PL_U() : ServiceFramework("pl:u"), impl{std::make_unique<Impl>()} {
...
+            LOG_ERROR(Service_NS, "{:016X} has no file \"{}\"! Skipping",
+                      static_cast<u64>(font.first), font.second);
+            continue;
         }
+        std::vector<u32> font_data_u32(font_fp->GetSize() / sizeof(u32));
+        font_fp->ReadBytes<u32>(font_data_u32.data(), font_fp->GetSize());
+        // We need to be BigEndian as u32s for the xor encryption
+        std::transform(font_data_u32.begin(), font_data_u32.end(), font_data_u32.begin(),
+                       Common::swap32);
+        FontRegion region{
```
lioncash: This can be const

ogniK5377:
@DarkLordZach Rebase and address issues please

PR #2710, opened 2019-07-10
am/kernel: Fix various bugs related to applet signalling and software keyboard

DarkLordZach:
Fixes:
- Missing call to MapMemory when creating a TransferMemory -- this would lead to a hard crash when trying to read the transfer memory.
- Using `Automatic` event types for applet data channels -- this fixes softlocks/infinite loops with applet data channels
- Correct a miscounted IPC header in `am`
- Add a missing state changed event signal in interactive applet -- fixes a softlock for some games
- Make wait object wake on any signalled thread, not all -- fixes softlock with applet as sdk code waits on two events that are mutually exculsive. This fixes a bug introduced by #2416 

pkdx
pkmn
smm2
stardew
mc

### DIFF
--- a/src/core/hle/kernel/transfer_memory.cpp
+++ b/src/core/hle/kernel/transfer_memory.cpp
@@ -23,6 +23,8 @@ SharedPtr<TransferMemory> TransferMemory::Create(KernelCore& kernel, VAddr base_
     transfer_memory->owner_permissions = permissions;
     transfer_memory->owner_process = kernel.CurrentProcess();
 
+    transfer_memory->MapMemory(base_address, size, permissions);
+
     return transfer_memory;
 }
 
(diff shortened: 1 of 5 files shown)


src/core/hle/kernel/transfer_memory.cpp
```diff
@@ -23,6 +23,8 @@ SharedPtr<TransferMemory> TransferMemory::Create(KernelCore& kernel, VAddr base_
     transfer_memory->owner_permissions = permissions;
     transfer_memory->owner_process = kernel.CurrentProcess();
 
+    transfer_memory->MapMemory(base_address, size, permissions);
```
lioncash: This will completely prevent svcMapTransferMemory from working properly. Especially if `svcCreateTransferMemory` is executed and followed up with `svcMapTransferMemory`

TuxSH: svcCreateTransferMemory only reprotects the source range with the given perms in the source process (and creates the tmem object).

bunnei:
Just to confirm since it's a bit ambiguous - comments are not addressed here

bunnei:
Superseded by #3366

PR #2716, opened 2019-07-11
service/am: Remove usages of global system accessors

lioncash:
Avoids the use of global accessors, removing the reliance on global state. This also makes dependencies explicit in the interface, as opposed to being hidden.

With this we're much closer to removing the `Core::CurrentProcess` global accessor.

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -56,7 +56,8 @@ struct LaunchParameters {
 };
 static_assert(sizeof(LaunchParameters) == 0x88);
 
-IWindowController::IWindowController() : ServiceFramework("IWindowController") {
+IWindowController::IWindowController(Core::System& system_)
+    : ServiceFramework("IWindowController"), system{system_} {
     // clang-format off
     static const FunctionInfo functions[] = {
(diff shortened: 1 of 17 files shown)


lioncash:
No idea what the failure reason is for the sync stage (unable to see the failure reason). Builds pass fine, however.

src/core/hle/service/am/am.cpp
```diff
@@ -990,7 +989,8 @@ void ILibraryAppletCreator::CreateTransferMemoryStorage(Kernel::HLERequestContex
     rb.PushIpcInterface(std::make_shared<IStorage>(std::move(memory)));
 }
 
-IApplicationFunctions::IApplicationFunctions() : ServiceFramework("IApplicationFunctions") {
+IApplicationFunctions::IApplicationFunctions(Core::System& system_)
```
ReinUsesLisp: Is there a reason why `system` is being passed with an underscore? I thought C++ resolved the naming conflict.

lioncash: -Wshadow will warn in this case, since the parameter would be shadowing the name of the member variable.

ogniK5377:
@lioncash requesting rebase

PR #2717, opened 2019-07-11
Restore memory perms on svcUnmapMemory/UnloadNro

SciresM:
Prior to PR, Yuzu did not restore memory to RW- on unmap of mirrored memory or unloading of NRO.

(In fact, in the NRO case, the memory was unmapped instead of reprotected to --- on Load, so it was actually lost entirely...)

This PR addresses that, and restores memory to RW- as it should.

This fixes a crash in Super Smash Bros when creating a World of Light save for the first time, and possibly other games/circumstances.

### DIFF
--- a/src/core/hle/kernel/svc.cpp
+++ b/src/core/hle/kernel/svc.cpp
@@ -318,7 +318,14 @@ static ResultCode UnmapMemory(Core::System& system, VAddr dst_addr, VAddr src_ad
         return result;
     }
 
-    return vm_manager.UnmapRange(dst_addr, size);
+    const auto unmap_res = vm_manager.UnmapRange(dst_addr, size);
+
+    // Reprotect the source mapping on success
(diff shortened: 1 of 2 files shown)


SciresM:
Also worth noting: Yuzu uses vm_manager.MirrorMemory in ldr.cpp for NRO stuff. 

This doesn't really match semantics for actual ro (see atmosphere source code)...but then since ro isn't a separate process it's kind of weird in the first place.

Still, if it's going to use MirrorMemory, perhaps yuzu should consider an "UnmirrorMemory" function in vm_manager? I don't know what it would look like, but maybe the memory manager could track what memory is aliased?

Also the memory manager should probably be setting the locked attribute, but I'm not 100% on that. The MapCodeMemory implementation looks like it's handling stuff that maybe MirrorMemory should.

All that can be addressed in a future PR after thought.

PR #2725, opened 2019-07-12
"AudioRenderer" thread should have a unique name

ogniK5377:
Creating multiple "AudioRenderer" threads cause the previous thread to be overwritten. The thread will name be renamed to AudioRenderer-InstanceX, where X is the current instance number.

The only game I'm able to test with this fix currently is CTR. This fixes the issue it has and goes in-game however it seems to have no samples playing. Currently unsure if it's related to this fix or if it's a separate issue. Hopefully having a larger sample of tests will conclude if we have an underlying problem with how we're handling these threads or if it's an audio renderer issue in general

### DIFF
--- a/src/audio_core/audio_renderer.h
+++ b/src/audio_core/audio_renderer.h
@@ -215,7 +215,7 @@ static_assert(sizeof(UpdateDataHeader) == 0x40, "UpdateDataHeader has wrong size
 class AudioRenderer {
 public:
     AudioRenderer(Core::Timing::CoreTiming& core_timing, AudioRendererParameter params,
-                  Kernel::SharedPtr<Kernel::WritableEvent> buffer_event, const std::size_t instance_number);
+                  Kernel::SharedPtr<Kernel::WritableEvent> buffer_event, std::size_t instance_number);
     ~AudioRenderer();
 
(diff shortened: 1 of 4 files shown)


ogniK5377:
We can have a total of 2 valid instances if I recall correctly. We store threads in a map by name on yuzu which in turn just deletes the previous existing thread

src/audio_core/audio_renderer.h
```diff
@@ -215,7 +215,7 @@ static_assert(sizeof(UpdateDataHeader) == 0x40, "UpdateDataHeader has wrong size
 class AudioRenderer {
 public:
     AudioRenderer(Core::Timing::CoreTiming& core_timing, AudioRendererParameter params,
-                  Kernel::SharedPtr<Kernel::WritableEvent> buffer_event);
+                  Kernel::SharedPtr<Kernel::WritableEvent> buffer_event, const std::size_t instance_number);
```
lioncash:
```suggestion
                  Kernel::SharedPtr<Kernel::WritableEvent> buffer_event, std::size_t instance_number);
```

value-wise const only has a meaning in the definition, not the declaration

PR #2734, opened 2019-07-15
gl_rasterizer: Implement compute shaders

ReinUsesLisp:
Implements compute shaders using OpenGL and [GL_ARB_compute_variable_group_size](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_compute_variable_group_size.txt) which is supported on Nvidia and AMD devices (both proprietary and free drivers). Implementing compute without this extension would require a shader cache invalidation, which can be done later on. For now devices without the extension (Intel) will skip kernel invocations.

There's still work to be done for compute shaders to be working on Nintendo games. That being said, I've seen reports of games like Brawlout properly emulating kernels.

A list of missing features potentially required to get kernels working on Nintendo games **not** implemented by this PR:
* Buffer "views" (buffers being used with a different base, buffer cache related).
* Improve global memory tracking (we fail to track some shaders in SSBU).
* Implement image bindings. I've got this working but adding it here is beyond the scope of this PR.
* Implement more half float instructions.
* Implement VOTE.VTG

[Here](https://github.com/yuzu-emu/yuzu/files/3390646/compute-2.zip) is a nouveau test using compute shaders, it requires #2565 to work. The expected output is a quad blinking between black and white, this color is written from a kernel.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_cache.cpp
@@ -123,9 +123,6 @@ ProgramType GetProgramType(Maxwell::ShaderProgram program) {
 
 /// Calculates the size of a program stream
 std::size_t CalculateProgramSize(const GLShader::ProgramCode& program) {
-    if (program.empty()) {
-        return 0;
-    }
     constexpr std::size_t start_offset = 10;
     // This is the encoded version of BRA that jumps to itself. All Nvidia
     // shaders end with one.
@@ -299,8 +296,8 @@ CachedShader::CachedShader(const ShaderParameters& params, ProgramType program_t
 
 Shader CachedShader::CreateStageFromMemory(const ShaderParameters& params,
                                            Maxwell::ShaderProgram program_type,
-                                           const ProgramCode& program_code,
-                                           const ProgramCode& program_code_b) {
+                                           ProgramCode&& program_code,
+                                           ProgramCode&& program_code_b) {
     const auto code_size{CalculateProgramSize(program_code)};
     const auto code_size_b{CalculateProgramSize(program_code_b)};
     auto result{
@@ -327,7 +324,7 @@ Shader CachedShader::CreateStageFromCache(const ShaderParameters& params,
 }
 
 Shader CachedShader::CreateKernelFromMemory(const ShaderParameters& params,
-                                            const ProgramCode& code) {
+                                            ProgramCode&& code) {
     auto result{CreateProgram(params.device, ProgramType::Compute, code, {})};
 
     const auto code_size{CalculateProgramSize(code)};
@@ -683,7 +680,7 @@ Shader ShaderCacheOpenGL::GetComputeKernel(GPUVAddr code_addr) {
     }
 
     // No kernel found - create a new one
-    const auto code{GetShaderCode(memory_manager, code_addr, host_ptr)};
+    auto code{GetShaderCode(memory_manager, code_addr, host_ptr)};
     const auto unique_identifier{GetUniqueIdentifier(ProgramType::Compute, code, {})};
     const auto cpu_addr{*memory_manager.GpuToCpuAddress(code_addr)};
     const ShaderParameters params{disk_cache, precompiled_programs, device, cpu_addr,
--- a/src/video_core/renderer_opengl/gl_shader_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_cache.cpp
@@ -323,8 +323,7 @@ Shader CachedShader::CreateStageFromCache(const ShaderParameters& params,
         new CachedShader(params, GetProgramType(program_type), std::move(result)));
 }
 
-Shader CachedShader::CreateKernelFromMemory(const ShaderParameters& params,
-                                            ProgramCode&& code) {
+Shader CachedShader::CreateKernelFromMemory(const ShaderParameters& params, ProgramCode&& code) {
     auto result{CreateProgram(params.device, ProgramType::Compute, code, {})};
 
     const auto code_size{CalculateProgramSize(code)};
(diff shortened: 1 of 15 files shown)


src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -100,8 +102,30 @@ constexpr std::tuple<const char*, const char*, u32> GetPrimitiveDescription(GLen
...
+    case Maxwell::ShaderProgram::Fragment:
+        return ProgramType::Fragment;
+    }
+    UNREACHABLE();
+    return {};
+}
+
 /// Calculates the size of a program stream
 std::size_t CalculateProgramSize(const GLShader::ProgramCode& program) {
+    if (program.empty()) {
```
FernandoS27: Add a comment in which cases a program can be empty.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -261,33 +296,55 @@ CachedShader::CachedShader(const ShaderParameters& params, Maxwell::ShaderProgra
 
 Shader CachedShader::CreateStageFromMemory(const ShaderParameters& params,
                                            Maxwell::ShaderProgram program_type,
-                                           ProgramCode&& program_code,
-                                           ProgramCode&& program_code_b) {
+                                           const ProgramCode& program_code,
+                                           const ProgramCode& program_code_b) {
```
FernandoS27: You are moving this variables, effectively nullifying their original content. By doing this, the move will be ineffective and a copy will be done. You don't need to do r-values though, a simple lvalue can be moved as far as it's not const but it must be stated on the comments of the function parameters.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -261,33 +296,55 @@ CachedShader::CachedShader(const ShaderParameters& params, Maxwell::ShaderProgra
...
 Shader CachedShader::CreateStageFromCache(const ShaderParameters& params,
                                           Maxwell::ShaderProgram program_type,
                                           GLShader::ProgramResult result) {
-    return std::shared_ptr<CachedShader>(new CachedShader(params, program_type, std::move(result)));
+    return std::shared_ptr<CachedShader>(
+        new CachedShader(params, GetProgramType(program_type), std::move(result)));
+}
+
+Shader CachedShader::CreateKernelFromMemory(const ShaderParameters& params,
+                                            const ProgramCode& code) {
```
FernandoS27: ditto about moving const lvalues

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -305,8 +362,11 @@ std::tuple<GLuint, BaseBindings> CachedShader::GetProgramHandle(const ProgramVar
         handle = program->handle;
     }
 
-    auto base_bindings{variant.base_bindings};
-    base_bindings.cbuf += static_cast<u32>(entries.const_buffers.size()) + RESERVED_UBOS;
+    auto base_bindings = variant.base_bindings;
```
FernandoS27: shouldn't this be a reference?

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -780,7 +787,7 @@ class GLSLDecompiler final {
             return value;
         }
         // There's a bug in NVidia's proprietary drivers that makes precise fail on fragment shaders
```
FernandoS27: remove this comment.

src/video_core/renderer_opengl/gl_shader_util.cpp
```diff
@@ -10,21 +10,25 @@
...
-        break;
+        return "geometry";
     case GL_FRAGMENT_SHADER:
-        debug_type = "fragment";
-        break;
+        return "fragment";
+    case GL_COMPUTE_SHADER:
+        return "compute";
     default:
         UNREACHABLE();
```
FernandoS27: This is not UNREACHABLE, tesselation shaders can hit it.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -102,7 +102,7 @@ constexpr std::tuple<const char*, const char*, u32> GetPrimitiveDescription(GLen
     }
 }
 
-constexpr ProgramType GetProgramType(Maxwell::ShaderProgram program) {
+ProgramType GetProgramType(Maxwell::ShaderProgram program) {
```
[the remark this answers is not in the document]
ReinUsesLisp: MinGW was complaining about `UNREACHABLE` not being constexpr.

src/video_core/renderer_opengl/gl_shader_cache.cpp
ReinUsesLisp: This was a rebase mistake, removing.

src/video_core/renderer_opengl/gl_shader_cache.cpp
ReinUsesLisp: Rebase mistake, I will revert it.

src/video_core/renderer_opengl/gl_shader_cache.cpp
ReinUsesLisp: No, because we modify it and return it by value. It's just 16 bytes long.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
ReinUsesLisp: The `precise` hack is still present.

MysticExile:
Fix clang-format please.

FernandoS27:
I believe this PR is already satisfactory. It's well tested and reviewed. I will, however, seek approval of @bunnei for merging it as it's a bit large.

PR #2736, opened 2019-07-15
gl_texture_cache: Implement asynchronous flushing

ReinUsesLisp:
Instead of flushing textures when "the guest wants them" we flush ahead of time when a linear render target is detached. For most cases this means that when the guest requests for texture memory we wait for a fence and swizzle the texture instead of halting the whole OpenGL pipeline and waiting for the GPU to finish. This heuristic might harm performance on games that render to linear textures but never read them from the CPU, I haven't found a case where this is true.

This also moves texture uploads/downloads to OpenGL buffers potentially avoiding a memcpy in the driver.

The whole fast path is disabled on Intel's proprietary driver since it was throwing `GL_OUT_OF_MEMORY` errors with it. It falls back to classic client side buffers for uploads and downloads, with ahead of time flushing disabled. From my testing, it improves performance on Nvidia and AMD GPUs.

Thanks to exzap for explaining me why `glFlush` is required after `glGetTextureImage` calls, the explanation is in the code.

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -4,6 +4,8 @@
 
 #include <array>
 #include <cstddef>
+#include <string_view>
+
 #include <glad/glad.h>
 
 #include "common/logging/log.h"
@@ -23,6 +25,9 @@ T GetInteger(GLenum pname) {
 } // Anonymous namespace
 
 Device::Device() {
+    const std::string_view vendor = reinterpret_cast<const char*>(glGetString(GL_VENDOR));
+    const bool intel_proprietary = vendor == "Intel";
+
     uniform_buffer_alignment = GetInteger<std::size_t>(GL_UNIFORM_BUFFER_OFFSET_ALIGNMENT);
     shader_storage_alignment = GetInteger<std::size_t>(GL_SHADER_STORAGE_BUFFER_OFFSET_ALIGNMENT);
     max_vertex_attributes = GetInteger<u32>(GL_MAX_VERTEX_ATTRIBS);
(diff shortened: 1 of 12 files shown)


FernandoS27:
you still got clang-format issues

src/video_core/renderer_opengl/gl_device.cpp
```diff
@@ -60,6 +67,7 @@ void main() {
 }
 
 bool Device::TestComponentIndexingBug() {
+    return false;
```
FernandoS27: is this a typo or something?

src/video_core/renderer_opengl/gl_texture_cache.h
```diff
@@ -138,6 +141,7 @@ class TextureCacheOpenGL final : public TextureCacheBase {
     OGLFramebuffer src_framebuffer;
     OGLFramebuffer dst_framebuffer;
     std::unordered_map<u32, OGLBuffer> copy_pbo_cache;
+    std::vector<u8> temporary_buffer;
```
FernandoS27: The main idea of the original stagging buffer was to avoid this. This makes a lot of memory for each texture. Please figure a way to fix this.

Hexagon12:
Please rebase

src/video_core/renderer_opengl/gl_device.cpp
ReinUsesLisp: Typo, addressed.

ogniK5377:
Requesting rebase @ReinUsesLisp

Hexagon12:
Requesting a rebase

ReinUsesLisp:
I'll redo this after some other changes are merged, closing for now.

PR #2737, opened 2019-07-15
Shader_Ir: Correct tracking to track from right to left

FernandoS27:
This corrects tracking to always look first for the right most operand rather than the left most operand. In general, this reduces tracking time and obtains the first cbuf faster. This fixes tracking in SSBU's compute shaders.


Hexagon12:
Fixes Xenoblade 2's global memory.

PR #2738, opened 2019-07-16
shader-ir: Minor cleanup-related changes

lioncash:
Performs a few minor cleanup changes, like removing unused headers, clarity renaming, and getting rid of unnecessary atomic reference count increment and decrements.

### DIFF
--- a/src/video_core/shader/decode.cpp
+++ b/src/video_core/shader/decode.cpp
@@ -46,12 +46,12 @@ void ShaderIR::Decode() {
         coverage_end = shader_info.end;
         if (shader_info.decompilable) {
             disable_flow_stack = true;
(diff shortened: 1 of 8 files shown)


FernandoS27:
Why doing std move on shared pointers? that's kinda dangerous.

lioncash:
How so? It's no more dangerous than using std::move on any other type.

FernandoS27:
I don't know, I've had bad experience with them causing memory corruption and troubles when used on shared pointers. I think this should be tested just to make sure. I really don't feel like reverting a PR this big.

FernandoS27:
You need to rebase.

lioncash:
Done

PR #2739, opened 2019-07-16
video_core/control_flow: Minor changes/warning cleanup

lioncash:
Provides a few minor cleanup-related changes that reduce the amount of work done in some scenarios. They also resolve some compilation warnings that would occur. Each commit should be fairly separated, making reviewing them in isolation possible.


lioncash:
Rebased to resolve conflicts occurring after the merging of #2738

PR #2742, opened 2019-07-18
gl_texture_cache: Miscellaneous texture buffer fixes

ReinUsesLisp:
There were some bugs in texture buffers management, these didn't deserve individual PRs so I packed them into one. The explanation is in the individual commits.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -561,7 +561,7 @@ private:
                 case Tegra::Shader::ImageType::Texture1D:
                     return "image1D";
                 case Tegra::Shader::ImageType::TextureBuffer:
-                    return "bufferImage";
+                    return "imageBuffer";
                 case Tegra::Shader::ImageType::Texture1DArray:
                     return "image1DArray";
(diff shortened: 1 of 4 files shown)


FernandoS27:
squash the commits all this can fit a single commit.

FernandoS27:
Untagged, seems to be crashing octopath for some reason.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -561,7 +561,7 @@ class GLSLDecompiler final {
                 case Tegra::Shader::ImageType::Texture1D:
                     return "image1D";
                 case Tegra::Shader::ImageType::TextureBuffer:
-                    return "bufferImage";
+                    return "imageBuffer";
```
Zeroksas: Using "imageBuffer" crashes Octopath Traveler. Changing it to any other String value fixes problem.

FernandoS27: That's because compute shaders don't have image bindings. O ther values just make the shader not to compile.

bunnei:
@ReinUsesLisp can you investigate the Octopath issue so that we can re-tag and get this merged?

PR #2743, opened 2019-07-18
Downgrade and suppress a series of GPU asserts and debug messages.

FernandoS27:


Hexagon12:
The AppVeyor fail is caused by a timeout.

PR #2748, opened 2019-07-19
VM_Manager: Align allocated host physical memory to 256bytes

FernandoS27:
This PR ensures that all backing memory allocated for the Guest CPU is aligned to 256 bytes. This is due to alignment requirements on the GPU which utilizes host physical memory addressing extensively.

### DIFF
--- a/src/common/alignment.h
+++ b/src/common/alignment.h
@@ -3,7 +3,10 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdlib>
 #include <type_traits>
+#include <malloc.h>
+#include <stdlib.h>
 
 namespace Common {
 
@@ -36,5 +39,81 @@ constexpr bool IsWordAligned(T value) {
     static_assert(std::is_unsigned_v<T>, "T must be an unsigned value.");
     return (value & 0b11) == 0;
 }
+
+template <typename T, std::size_t Align = 16>
+class AlignmentAllocator {
+public:
+    typedef T value_type;
+    typedef std::size_t size_type;
+    typedef std::ptrdiff_t difference_type;
+
+    typedef T* pointer;
+    typedef const T* const_pointer;
+
+    typedef T& reference;
+    typedef const T& const_reference;
+
+public:
+    inline AlignmentAllocator() throw() {}
+
+    template <typename T2>
+    inline AlignmentAllocator(const AlignmentAllocator<T2, Align>&) throw() {}
+
+    inline ~AlignmentAllocator() throw() {}
+
+    inline pointer adress(reference r) {
+        return &r;
+    }
+
+    inline const_pointer adress(const_reference r) const {
+        return &r;
+    }
+
+#if (defined _MSC_VER)
+    inline pointer allocate(size_type n) {
+        return (pointer)_aligned_malloc(n * sizeof(value_type), Align);
+    }
+
+    inline void deallocate(pointer p, size_type) {
+        _aligned_free(p);
+    }
+#else
+    inline pointer allocate(size_type n) {
+        return (pointer)std::aligned_alloc(Align, n * sizeof(value_type));
+    }
+
+    inline void deallocate(pointer p, size_type) {
+        std::free(p);
+    }
+#endif
+
+    inline void construct(pointer p, const value_type& wert) {
+        new (p) value_type(wert);
+    }
+
+    inline void destroy(pointer p) {
+        p->~value_type();
+    }
+
+    inline size_type max_size() const throw() {
+        return size_type(-1) / sizeof(value_type);
+    }
+
+    template <typename T2>
+    struct rebind {
+        typedef AlignmentAllocator<T2, Align> other;
+    };
+
+    bool operator!=(const AlignmentAllocator<T, Align>& other) const {
+        return !(*this == other);
+    }
+
+    // Returns true if and only if storage allocated from *this
+    // can be deallocated from other, and vice versa.
+    // Always returns true for stateless allocators.
+    bool operator==(const AlignmentAllocator<T, Align>& other) const {
+        return true;
+    }
+};
 
 } // namespace Common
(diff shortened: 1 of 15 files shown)


src/common/alignment.h
```diff
@@ -3,7 +3,10 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdlib>
 #include <type_traits>
+#include <malloc.h>
+#include <stdlib.h>
```
lioncash: This isn't necessary

src/common/alignment.h
```diff
@@ -37,4 +40,80 @@ constexpr bool IsWordAligned(T value) {
     return (value & 0b11) == 0;
 }
 
+template <typename T, std::size_t Align = 16>
+class AlignmentAllocator {
```
lioncash: Was this taken from somewhere else? This is using quite an old/outdated means of allocating aligned memory. operator new already supports aligned allocations and doesn't need to use OS-specific memory allocation routines. `throw()` is also deprecated since the introduction of `noexcept`

src/common/alignment.h
```diff
@@ -37,4 +40,80 @@ constexpr bool IsWordAligned(T value) {
...
+public:
+    typedef T value_type;
```
lioncash: This should be using `using` for type aliases.

src/common/alignment.h
```diff
@@ -37,4 +40,80 @@ constexpr bool IsWordAligned(T value) {
...
+    typedef std::ptrdiff_t difference_type;
+
+    typedef T* pointer;
+    typedef const T* const_pointer;
+
+    typedef T& reference;
+    typedef const T& const_reference;
+
+public:
+    inline AlignmentAllocator() throw() {}
```
lioncash: `inline isn't necessary for functions defined in the class body.

src/common/alignment.h
```diff
@@ -37,4 +40,80 @@ constexpr bool IsWordAligned(T value) {
...
+
+    template <typename T2>
+    inline AlignmentAllocator(const AlignmentAllocator<T2, Align>&) throw() {}
+
+    inline ~AlignmentAllocator() throw() {}
+
+    inline pointer adress(reference r) {
```
lioncash:
```suggestion
    inline pointer address(reference r) {
```

Ditto below.

Note that you previously told me you tested this interface on Discord. This is proving to be not the case, given it's a central part of the allocator interface (until C++20). Particularly given the implementation itself is incorrect, as described below.

src/common/alignment.h
```diff
@@ -3,7 +3,10 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdlib>
 #include <type_traits>
+#include <malloc.h>
```
lioncash: This isn't a platform-agnostic header, and also isn't necessary.

src/common/alignment.h
```diff
@@ -37,4 +40,80 @@ constexpr bool IsWordAligned(T value) {
...
+
+    inline pointer adress(reference r) {
+        return &r;
+    }
+
+    inline const_pointer adress(const_reference r) const {
+        return &r;
+    }
+
+#if (defined _MSC_VER)
```
lioncash: This if-def isn't necessary, [operator new](https://en.cppreference.com/w/cpp/memory/new/operator_new) supports allocating with a specific alignment. See the overloads accepting `std::aligned_val_t`

src/common/alignment.h
```diff
@@ -37,4 +40,80 @@ constexpr bool IsWordAligned(T value) {
...
+public:
+    inline AlignmentAllocator() throw() {}
+
+    template <typename T2>
+    inline AlignmentAllocator(const AlignmentAllocator<T2, Align>&) throw() {}
+
+    inline ~AlignmentAllocator() throw() {}
+
+    inline pointer adress(reference r) {
+        return &r;
```
lioncash: This doesn't satisfy the requirement that the address of the object be returned, even if `operator&` is overloaded.

You need to use `return std::addressof(r);` for that. Ditto below.

src/core/hle/kernel/physical_memory.h
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/alignment.h"
+
+namespace Kernel {
+
+using PhysicalMemory = std::vector<u8, Common::AlignmentAllocator<u8, 256>>;
```
lioncash: This should contain a rationale for the existence of the type as a comment so others know why it exists.

FernandoS27:
I'll close this and use boost align once I figure how to get it merged.

src/common/alignment.h
```diff
@@ -37,4 +38,63 @@ constexpr bool IsWordAligned(T value) {
...
+    using difference_type = std::ptrdiff_t;
+
+    using pointer = T*;
+    using const_pointer = const T*;
+
+    using reference = T&;
+    using const_reference = const T&;
+
+public:
+    pointer address(reference r) {
```
lioncash: This and the const variant of `address` should be `noexcept`

src/common/alignment.h
```diff
@@ -37,4 +38,63 @@ constexpr bool IsWordAligned(T value) {
...
+    size_type max_size() const noexcept {
+        return size_type(-1) / sizeof(value_type);
+    }
+
+    template <typename T2>
+    struct rebind {
+        typedef AlignmentAllocator<T2, Align> other;
+    };
+
+    bool operator!=(const AlignmentAllocator<T, Align>& other) const {
```
lioncash: This and `operator==` should be noexcept

src/common/alignment.h
```diff
@@ -37,4 +38,63 @@ constexpr bool IsWordAligned(T value) {
...
+        p->~value_type();
+    }
+
+    size_type max_size() const noexcept {
+        return size_type(-1) / sizeof(value_type);
+    }
+
+    template <typename T2>
+    struct rebind {
+        typedef AlignmentAllocator<T2, Align> other;
```
lioncash:
```cpp
using other = AlignmentAllocator<T2, Align>;
```

bunnei:
Discussed via Discord that it's not worth the effort to bring in Boost here (re the comment above)

PR #2753, opened 2019-07-20
Shader_Ir: Implement F16 Variants of F2F, F2I, I2F.

FernandoS27:
This commit takes care of implementing the F16 Variants of the 
conversion instructions and makes sure conversions are done.

### DIFF
--- a/src/video_core/shader/decode/conversion.cpp
+++ b/src/video_core/shader/decode/conversion.cpp
@@ -57,7 +57,7 @@ u32 ShaderIR::DecodeConversion(NodeBlock& bb, u32 pc) {
     case OpCode::Id::I2F_R:
     case OpCode::Id::I2F_C:
     case OpCode::Id::I2F_IMM: {
-        UNIMPLEMENTED_IF(instr.conversion.dst_size != Register::Size::Word);
+        UNIMPLEMENTED_IF(instr.conversion.dst_size == Register::Size::Long);
         UNIMPLEMENTED_IF(instr.conversion.selector);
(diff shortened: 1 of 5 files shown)


src/video_core/shader/decode/conversion.cpp
```diff
@@ -82,14 +82,19 @@ u32 ShaderIR::DecodeConversion(NodeBlock& bb, u32 pc) {
         value = GetOperandAbsNegFloat(value, false, instr.conversion.negate_a);
 
         SetInternalFlagsFromFloat(bb, value, instr.generates_cc);
+
+        if (instr.conversion.dst_size == Register::Size::Short) {
+            value = Operation(OperationCode::HCastFloat, PRECISE, value);
```
ReinUsesLisp: This should go before `SetInternalFlagsFromFloat`.

FernandoS27: We would need to operate `SetInternalFlagsFromFloat` in half floats. Thankfully all possible operations are the same in half floats as floats so it shouldn't matter if we use the float version for getting the flags.

ReinUsesLisp: It does matter since the shader IR operates at a binary level. It's not the same comparing the sign flag for `0x4f00bf00` as float or as a half float pair.
Add an UNIMPLEMENTED in case dst is half float.

FernandoS27: it's not a half float by then, it's a float. The sign will be preserved on conversion.

PR #2759, opened 2019-07-22
gl_rasterizer: Bind images and samplers to compute

ReinUsesLisp:
Implements sampler and image bindings to compute invocations. Images are used in instructions like `SUST`, `SULD` and `SUATOM`. There is some code repetition between `maxwell_3d` and `kepler_compute` that I'd like to address afterwards.

While adding an extra entry to OpenGLState, this splits texture and samplers binding to pass them to OpenGL as they are instead of generating a temporary buffer in the stack and abstracts the algorithm into a separate function.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -801,9 +801,11 @@ void RasterizerOpenGL::DispatchCompute(GPUVAddr code_addr) {
     }
 
     auto kernel = shader_cache.GetComputeKernel(code_addr);
+    ProgramVariant variant;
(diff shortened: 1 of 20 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -29,6 +29,8 @@
 #include "video_core/renderer_opengl/maxwell_to_gl.h"
 #include "video_core/renderer_opengl/renderer_opengl.h"
 
+#pragma optimize("", off)
```
Hexagon12: This shouldn't be present

ReinUsesLisp:
Added new commit to keep track of images written by the shader. These are marked as modified to allow flushing them to guest memory when requested.

PR #2762, opened 2019-07-24
Implement Accelerate DMA for Texture Cache and correct some methods in DMA engine

FernandoS27:
This PR takes care of providing a safe conduct to do DMAs on the GPU if any of the textures is within the Texture Cache. It also corrects some missing things in Linear->Tile Copies and Tiled->Linear Copies as a result DKCTP now displays text and icons correctly.

### DIFF
--- a/src/video_core/rasterizer_interface.h
+++ b/src/video_core/rasterizer_interface.h
@@ -8,6 +8,7 @@
 #include <functional>
 #include "common/common_types.h"
 #include "video_core/engines/fermi_2d.h"
+#include "video_core/engines/maxwell_dma.h"
 #include "video_core/gpu.h"
 
 namespace Tegra {
@@ -68,6 +69,10 @@ public:
                                        const Tegra::Engines::Fermi2D::Config& copy_config) {
         return false;
     }
+
+    virtual void AccelerateDMATexture(const Tegra::Engines::MaxwellDMA::SurfaceConfig& src_config,
+                                      const Tegra::Engines::MaxwellDMA::SurfaceConfig& dst_config,
+                                      const Tegra::Engines::MaxwellDMA::CopyConfig& copy_config) {}
 
     /// Attempt to use a faster method to display the framebuffer to screen
     virtual bool AccelerateDisplay(const Tegra::FramebufferConfig& config, VAddr framebuffer_addr,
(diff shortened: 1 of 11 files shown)


src/video_core/texture_cache/texture_cache.h
```diff
@@ -773,6 +843,67 @@ class TextureCache {
...
+
+    std::pair<TSurface, TView> GetDMASurface(
+        const Tegra::Engines::MaxwellDMA::SurfaceConfig& config, const DMAInfo& info) {
+        SurfaceParams params = SurfaceParams::CreateForDMASurface(config, info.component_type,
+                                                                  info.pixel_format, info.target);
+        const GPUVAddr gpu_addr = config.gpu_addr;
+        return GetSurface(gpu_addr, params, true, false);
+    }
+
+    SurfaceTarget FigureDMATarget(const Tegra::Engines::MaxwellDMA::SurfaceConfig& config) {
```
lioncash:
```suggestion
    SurfaceTarget DetermineDMATarget(const Tegra::Engines::MaxwellDMA::SurfaceConfig& config) {
```

src/video_core/rasterizer_interface.h
```diff
@@ -22,6 +23,13 @@ enum class LoadCallbackStage {
     Build,
     Complete,
 };
+
+enum Caches : u32 {
```
lioncash: `enum class` and provide the relevant operations. This should also be named something like `CacheFlags`

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -811,6 +811,14 @@ void RasterizerOpenGL::FlushAndInvalidateRegion(CacheAddr addr, u64 size) {
     InvalidateRegion(addr, size);
 }
 
+u32 RasterizerOpenGL::IsCacheHit(const GPUVAddr gpu_addr, const std::size_t size) {
```
lioncash: This should return the actual type of the flags and not erase the type into a u32

src/video_core/rasterizer_interface.h
FernandoS27: I can't use enum class as a flag. It does not inherit the Bitwise Or operator.

FernandoS27:
I've learned that accelerate DMA is mostly unfeasible due to how it works and that oit's used to prepare a texture we got no info on. I've learned a few things on how this DMAs work and I'll add the corrections and relaxations to another PR.

PR #2763, opened 2019-07-24
kernel/vm_manager: Minor cleanup

lioncash:
Performs some minor cleanup related to the memory manager. Mostly just performs simplifications and very small performance improvements in the slow path of `MergeAdjacentVMA`. One commit fixes a trivial bug within a failure case for `UnmapPhysicalMemory` as well. This case shouldn't be hit within normal (stable) software emulation, so that's likely why it has gone unnoticed until now.


bunnei:
@lioncash can you rebase this?

lioncash:
Done

PR #2769, opened 2019-07-26
GPU: Flush commands on every dma pusher step.

FernandoS27:
This commit ensures that the host gpu is constantly fed with commands to
work with, while the guest gpu keeps producing the rest of the commands.
This reduces syncing time between host and guest gpu.

### DIFF
--- a/src/video_core/rasterizer_interface.h
+++ b/src/video_core/rasterizer_interface.h
@@ -50,6 +50,9 @@ public:
     /// and invalidated
     virtual void FlushAndInvalidateRegion(CacheAddr addr, u64 size) = 0;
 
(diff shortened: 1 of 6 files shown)


src/video_core/rasterizer_interface.h
```diff
@@ -50,6 +50,8 @@ class RasterizerInterface {
     /// and invalidated
     virtual void FlushAndInvalidateRegion(CacheAddr addr, u64 size) = 0;
 
+    virtual void FlushCommands() = 0;
```
lioncash: Missing a documentation comment like the other interface functions.

FernandoS27: done.

PR #2770, opened 2019-07-28
ci: Fix Azure PR Builds

DarkLordZach:
- Removes caching from PR builds -- in its current form caches count as a build secret, which aren't accessible to forks

- Updates the version of pip used on the merge build agent -- the pip was too old and was having random errors.


chris062689:
@DarkLordZach please add a description of what exactly is the issue and what you are doing to resolve it.

DarkLordZach:
This:
- Removes caching from PR builds -- in its current form caches count as a build secret, which aren't accessible to forks
- Updates the version of pip used on the merge build agent -- the pip was too old and was having random errors.

PR #2772, opened 2019-07-29
yuzu/CMakeLists: Remove qt5_wrap_ui macro usage

lioncash:
We can enable [CMAKE_AUTOUIC](https://cmake.org/cmake/help/latest/variable/CMAKE_AUTOUIC.html) and let CMake take care of handling the UI code generation for targets. Now whenever a UI file is added, it can just be added to the main source file listing and it'll just work.

As part of letting CMake automatically handle the header file parsing, we must not name includes with "ui_*" unless they're related to the output of the Qt UIC compiler (which is quite sensible, in my opinion). Because of this, we need to rename `ui_settings`, given it would conflict with this restriction.


jroweboy:
Does this resolve the issue with recent versions of CMake and VS that trigger a rebuild of all UI files?

lioncash:
Yes. CMake will cache the results of its parsing. Only translation units that need to be recompiled due to modifications will have their necessary files regenerated in order for compilation to succeed. It won't rebuild all of the UI files regardless of which ones actually need to be rebuilt.

FearlessTobi:
Please force push in order to see if the Azure build passes.

PR #2783, opened 2019-08-09
Implement a New LLE Buffer Cache

FernandoS27:
This is the 3rd attempt at making a buffer cache. So far it's almost as fast as V1 but a lot more accurate. Unlike previous buffer caches, this one is LLE(Low Level Emulation). The cache works by mapping guest memory blocks into Host GPU Buffer blocks. It divides those blocks on 2Mb sized pages that represent the corresponding ones in guest memory. This cache is aware of memory writes to/from Host Memory does keeping correct constancy acrosss. It also sends small buffer caches to a faster stream buffer when the page is not modified and the usage of the buffer is read-only. In general, I haven't seen speed regressions at all but a lot of improvements can still be made.

Requires #2748


src/video_core/buffer_cache/buffer_block.h
```diff
@@ -0,0 +1,78 @@
...
+
+namespace VideoCommon {
+
+class BufferBlock {
+public:
+    bool Overlaps(const CacheAddr start, const CacheAddr end) const {
+        return (cache_addr < end) && (cache_addr_end > start);
+    }
+
+    bool IsInside(const CacheAddr other_start, const CacheAddr other_end) {
```
lioncash: This can be a const member function

src/video_core/buffer_cache/buffer_block.h
```diff
@@ -0,0 +1,78 @@
...
+        return (cache_addr <= other_start && other_end <= cache_addr_end);
```
lioncash: Parentheses are unnecessary

src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -0,0 +1,449 @@
...
+            MapInterval& current_map = overlaps[0];
+            if (current_map->IsInside(cache_addr, cache_addr_end)) {
+                return current_map;
+            }
+        }
+        CacheAddr new_start = cache_addr;
+        CacheAddr new_end = cache_addr_end;
+        bool write_inheritance = false;
+        bool modified_inheritance = false;
+        // Calculate new buffer parameters
```
degasus: // Calculate new buffer parameters which covers the full range of all overlapping blocks

src/video_core/renderer_opengl/gl_buffer_cache.cpp
```diff
@@ -7,52 +7,65 @@
...
 
+MICROPROFILE_DEFINE(OpenGL_Buffer_Download, "OpenGL", "Buffer Download", MP_RGB(192, 192, 128));
+
+CachedBufferBlock::CachedBufferBlock(CacheAddr cache_addr, const std::size_t size)
+    : VideoCommon::BufferBlock{cache_addr, size} {
+    gl_buffer.Create();
+    glNamedBufferData(gl_buffer.handle, static_cast<GLsizeiptr>(size), nullptr, GL_DYNAMIC_DRAW);
+}
+
+CachedBufferBlock::~CachedBufferBlock() = default;
```
degasus: The destructor will call glDeleteBuffers(). This is a huge overhead for the driver, so it is a good idea to use a pool for unused buffer blocks.

src/video_core/renderer_opengl/gl_buffer_cache.cpp
```diff
@@ -7,52 +7,65 @@
...
 
 const GLuint* OGLBufferCache::GetEmptyBuffer(std::size_t) {
     static const GLuint null_buffer = 0;
     return &null_buffer;
 }
 
-void OGLBufferCache::UploadBufferData(const OGLBuffer& buffer, std::size_t offset, std::size_t size,
-                                      const u8* data) {
-    glNamedBufferSubData(buffer.handle, static_cast<GLintptr>(offset),
+void OGLBufferCache::UploadBlockData(const Buffer& buffer, std::size_t offset, std::size_t size,
```
degasus: Please keep in mind that the driver will implement glNamedBufferSubData with a memcpy to a staging buffer and a glCopy to the final buffer. So the caller of this function might implement something better if your API just returns the pointer of an internal staging buffer instead. So never use this method if the caller has another copy for the data.

src/core/hle/kernel/physical_memory.h
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Kernel {
+
+// This encapsulation serves 2 purposes:
+// - First, to encapsulate host physical memory under a single type and set an
+// standard for managing it.
+// - Second to ensure all host backing memory used is aligned to 256 bytes due
+// to strict alignment restrictions on GPU memory.
+
+using PhysicalMemory = std::vector<u8, Common::AlignmentAllocator<u8, 256>>;
```
degasus: It seems like most of the usages are within a shared_ptr. Is there a way to combine the shared_ptr with the aligned allocator to avoid the pointer to a pointer overhead?

src/video_core/renderer_opengl/gl_buffer_cache.cpp
FernandoS27: It's weird for a cachedbufferblock to die. Only happens after being merged with another (thus creating a new one and letting the old ones die after a frame). This process normally happens once as after merge you just have a bigger page to map things on.

src/core/hle/kernel/physical_memory.h
FernandoS27: Not sure, this is normally handled by HLE guys. I really don't know why they have a shared pointer to a vector rather to data itself. Not trying to change their design. In general, it's possible to do a make_shared with a u8 block using the alignmentallocator.

src/video_core/buffer_cache/buffer_block.h
fluffapparition: @lioncash better safe then sorry :upside_down:

PR #2784, opened 2019-08-09
shader_ir: Implement shared memory

ReinUsesLisp:
Implement shared memory using SSBOs. Instead of using host shared memory for this, I opted for SSBOs because it let's us alias memory in GLSL to choose between signed and unsigned overloads sets (required for ATOMS).

As a side effect of this approach, we can avoid issues of different hardware having different shared memory size limitations. SSBOs have a much higher size limit compared to shared memory.


ReinUsesLisp:
There is a better way to implement this with `atomicCompSwap` that I need to investigate.

ReinUsesLisp:
Dropped the SSBO approach in favour of host shared memory. As imirkin suggested, we can use `atomicCompSwap` to emulate any operation atomically.

PR #2786, opened 2019-08-10
shader_ir: Implement VOTE on Nvidia drivers

ReinUsesLisp:
Implement VOTE using Nvidia's intrinsics. Documentation about these can be found [here](https://developer.nvidia.com/reading-between-threads-shader-intrinsics). Instead of using portable ARB instructions I opted to use Nvidia intrinsics because these are the closest we have to how Tegra X1 hardware renders.

To stub VOTE on non-Nvidia drivers (including nouveau) this commit simulates a GPU with a warp size of one, returning what is meaningful for the instruction being emulated:

* `anyThreadNV(value)` -> `value`
* `allThreadsNV(value)` -> `value`
* `allThreadsEqualNV(value)` -> `true`  

`ballotARB`, also known as `uint64_t(activeThreadsNV())`, emits
```asm
VOTE.ANY Rd, PT, PT;
```
on nouveau's compiler. This doesn't match exactly to Nvidia's code
```asm
VOTE.ALL Rd, PT, PT;
```
which is emulated with activeThreadsNV() by this commit. In theory this shouldn't really matter since `.ANY`, `.ALL` and `.EQ` affect the predicates (set to PT on those cases) and not the registers.

nouveau tests used to implement these instructions can be found [here](https://github.com/yuzu-emu/yuzu/files/3488456/tests.zip).

### DIFF
--- a/CMakeModules/GenerateSCMRev.cmake
+++ b/CMakeModules/GenerateSCMRev.cmake
@@ -81,6 +81,7 @@ set(HASH_FILES
     "${VIDEO_CORE}/shader/decode/register_set_predicate.cpp"
     "${VIDEO_CORE}/shader/decode/shift.cpp"
     "${VIDEO_CORE}/shader/decode/video.cpp"
+    "${VIDEO_CORE}/shader/decode/warp.cpp"
     "${VIDEO_CORE}/shader/decode/xmad.cpp"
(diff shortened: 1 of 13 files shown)


fluffapparition:
*vore

Issue #2011, opened 2019-01-09
Support RGB565 framebuffers


fincs:
Currently yuzu only seems to implement support for RGBA8888 framebuffers. This breaks the recently released 2.0.0 release of libnx, which moved the software-rendered console implementation to RGB565 because it saves memory bandwidth.

Some testing on hardware revealed [some other formats](https://github.com/switchbrew/libnx/blob/master/nx/source/display/framebuffer.c#L15-L23) which are also supported by the Switch's compositor other than RGBA8888. Attached is a libnx console example that now exhibits broken behavior on yuzu.

[vt52-demo.zip](https://github.com/yuzu-emu/yuzu/files/2741914/vt52-demo.zip)

greggameplayer:
@fincs test this #2016 please

fincs:
Tested canary c5baa21 (which has #2016) merged; and the same issue appears.

greggameplayer:
@fincs the change isn't in the canary for that moment

greggameplayer:
it's BGRA8 which is in canary for now

fincs:
Alright, I'll test when the next canary gets built.

greggameplayer:
@fincs can you tell me other homebrews who use it in order to do some test ? it will be good

fincs:
That's still not right. It should look something like this: 
![image](https://user-images.githubusercontent.com/581494/51150460-afff8480-1866-11e9-941e-057f43f994ae.png)

ReinUsesLisp:
Closing. This has been addressed by #2793.

PR #2793, opened 2019-08-21
 renderer_opengl: Implement RGB565 framebuffer format 

ReinUsesLisp:
Addresses #2011

While we are at it, this removes the optimized path that only supports 4 bytes with the generic one and replaces `std::optional<std::reference_wrapper<const T>>` with `const T*` for FramebufferConfig.

Currently the block height to 16 is hardcoded since this one is not being passed from HLE.

imagen


FernandoS27:
Bye Bye Citra remains.

PR #2797, opened 2019-08-22
Port citra-emu/citra#4877: "citra_qt: on osx chdir to bundle dir to allow detection of user folder"

FearlessTobi:
See citra-emu/citra#4877 for more details.

Yes, we don't currently have an OSX build but the plan was to add it back when there's Vulkan support and compatibility layers support all the features we need. 
As we don't have any negative effects from it, it makes sense porting all MacOS features or workaround from Citra.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2172,7 +2172,10 @@ int main(int argc, char* argv[]) {
     QCoreApplication::setApplicationName(QStringLiteral("yuzu"));
 
 #ifdef __APPLE__
-    std::string bin_path = FileUtil::GetBundleDirectory() + DIR_SEP + "..";
+    // If you start a bundle (binary) on OSX without the Terminal, the working directory is "/".
+    // But since we require the working directory to be the executable path for the location of the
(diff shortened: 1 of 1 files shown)


src/yuzu/main.cpp
```diff
@@ -2168,6 +2171,11 @@ int main(int argc, char* argv[]) {
     QCoreApplication::setOrganizationName(QStringLiteral("yuzu team"));
     QCoreApplication::setApplicationName(QStringLiteral("yuzu"));
 
+#ifdef __APPLE__
+    std::string bin_path = FileUtil::GetBundleDirectory() + DIR_SEP + "..";
```
```suggestion
    const std::string bin_path = FileUtil::GetBundleDirectory() + DIR_SEP + "..";
```

src/yuzu/main.cpp
```diff
@@ -2168,6 +2171,11 @@ int main(int argc, char* argv[]) {
...
+    chdir(bin_path.c_str());
```
lioncash: There should probably be a comment explaining the context for this in code, otherwise it can be non-obvious,

FearlessTobi:
Addressed.

PR #2801, opened 2019-08-24
gl_shader_decompiler: Rework GLSL decompiler type system

ReinUsesLisp:
GLSL decompiler type system was broken. We converted all return values to float except for some cases where returning we couldn't and implicitly broke the rule of returning floats (e.g. for bools or bool pairs).

Instead of doing this introduce class Expression that knows what type a return value has and when a consumer wants to use the string it asks for it with a required type, emitting a runtime error if types are incompatible.

This has the disadvantage that there's more C++ code, but we can emit better GLSL code that's easier to read.

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -14,12 +14,22 @@
 namespace OpenGL {
 
 namespace {
+
 template <typename T>
 T GetInteger(GLenum pname) {
     GLint temporary;
     glGetIntegerv(pname, &temporary);
     return static_cast<T>(temporary);
 }
+
+bool TestProgram(const GLchar* glsl) {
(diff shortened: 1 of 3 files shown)


Hexagon12:
Untagged due to real build errors.

FernandoS27:
Ig nore the canary tag. I'll wait for our half broken CI.

Hexagon12:
Build seems to work fine now. Retagging the PR again...

FernandoS27:
This seems to cause issues in Kepler NVIDIA GPUs.

ReinUsesLisp:
It should be addressed now (good old precise bug on old Nvidia drivers). Next time please add a log or something to be able to reproduce the bug.

FernandoS27:
I rarely add logs as they are useless 99% of the time for driver issues. Good job.

PR #2807, opened 2019-08-26
Port citra-emu/citra#4864: "yuzu/configure: move speed limiter to general"

FearlessTobi:
See citra-emu/citra#4864 for more details.

**Original description**:
The speed limiter being a frame limiter is an implmentation detail and can be changed in the future. What user care about is that it limit the emulation speed in genenral (not just graphics but also audio+input).

Unbenannt


ogniK5377:
@FearlessTobi Requesting rebase please

PR #2810, opened 2019-08-27
maxwell_3d: Avoid moving macro_params

ReinUsesLisp:
Instead of calling `std::move` on `macro_params` to the callee in `Maxwell3D::CallMethod`, pass a memory reference of it. This avoids constantly having `macro_params` without allocations, making its `std::vector<>::push_back` visible in the profiler (~5% from my testing).

Internally this memory reference is consumed with a `std::vector<u32>`-like `std::unique_ptr<u32[]>` but unlike `std::vector<>::resize` it won't zero newly allocated memory.

This turns constant allocation into a `std::memcpy`.

Thanks to @graphitemaster for pointing this out and suggesting a solution.


ReinUsesLisp:
Untagging in favour of #2826 for now.

Issue #2501, opened 2019-05-20
Possible Inaccuracies in MME emulation


FernandoS27:
fincs found a few inaccuracies in our MME emulation by analyzing code submitted by NVN. 

He created the next test:
```    5 to r1
.loop
    1 to mem
    dec r1 to r1
    bza r1 .end
    2 to mem
    b .loop
    *3 to mem
    nop

.end
    exit
    nop
```

which should return "Produces: 1 2 3 1 2 3 1 2 3 1 2 3 1"
Accordingly he points out that https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/macro_interpreter.cpp#L125 should actualy be: `if (opcode.is_exit && !is_delay_slot) {`

@Subv introduced the solution as: #2466 

Sadly this implementation produced infinite loops within the MME Interpreter. The change has been temporally reverted in order to be better investigate the inaccuracies in the MME interpreter and figure why such change causes an infinite loop.

fincs:
The inaccuracy has a simple cause: NVN runs this loop inside its FirmwareCall macro; in order to check when D00 changes. yuzu doesn't emulate firmware calls, so now that the loop is correctly emulated it effectively becomes an infloop.

ReinUsesLisp:
Should we open a separate issue for properly emulating FirmwareCalls? Can this be closed?

ReinUsesLisp:
Closing as the main issue has been addressed.

PR #2824, opened 2019-08-31
Revert "Revert #2466" and stub FirmwareCall 4

ReinUsesLisp:
cc #2501

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -328,6 +328,10 @@ void Maxwell3D::CallMethod(const GPU::MethodCall& method_call) {
         ProcessMacroBind(method_call.argument);
         break;
     }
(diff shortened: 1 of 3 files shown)


FernandoS27:
let's hope this one works.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -418,6 +422,13 @@ void Maxwell3D::ProcessMacroBind(u32 data) {
     macro_positions[regs.macros.entry++] = data;
 }
 
+void Maxwell3D::ProcessFirmwareCall4() {
+    LOG_WARNING(HW_GPU, "(STUBBED) called");
+
+    // For details refer to #2501
```
FernandoS27: The comments in code should be independent of Github's issue/PR tracking. Please explain here why is that.

PR #2827, opened 2019-09-03
AM: Stub IApplicationFunctions::GetGpuErrorDetectedSystemEvent

mailwl:
This function used in new games, line **Wolfenstein Youngblood**

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -1057,6 +1057,7 @@ IApplicationFunctions::IApplicationFunctions() : ServiceFramework("IApplicationF
         {120, nullptr, "ExecuteProgram"},
         {121, nullptr, "ClearUserChannel"},
         {122, nullptr, "UnpopToUserChannel"},
+        {130, &IApplicationFunctions::GetGpuErrorDetectedSystemEvent, "GetGpuErrorDetectedSystemEvent"},
         {500, nullptr, "StartContinuousRecordingFlushForDebug"},
         {1000, nullptr, "CreateMovieMaker"},
         {1001, nullptr, "PrepareForJit"},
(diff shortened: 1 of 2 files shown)


src/core/hle/service/am/am.cpp
```diff
@@ -1285,6 +1290,15 @@ void IApplicationFunctions::GetSaveDataSize(Kernel::HLERequestContext& ctx) {
     rb.Push(size.journal);
 }
 
+void IApplicationFunctions::GetGpuErrorDetectedSystemEvent(Kernel::HLERequestContext& ctx) {
+    LOG_WARNING(Service_AM, "(STUBBED) called");
+    gpu_error_detected_event.writable->Clear();
```
lioncash: Are you absolutely certain that the function clears the event before returning the handle?

mailwl: from switchbrew: When the Event is signaled, official sw will assert.
i tryed to signal event before return, and game immediately crash

wwylele: I think the expected implementation here is to do nothing with the event (neither clearing or signalling)

mailwl: yes, looks like default value is "clear". ok i will get rid it

src/core/hle/service/am/am.cpp
```diff
@@ -1057,13 +1057,18 @@ IApplicationFunctions::IApplicationFunctions() : ServiceFramework("IApplicationF
...
         {1000, nullptr, "CreateMovieMaker"},
         {1001, nullptr, "PrepareForJit"},
     };
     // clang-format on
 
     RegisterHandlers(functions);
+
+    auto& kernel = Core::System::GetInstance().Kernel();
+    gpu_error_detected_event = Kernel::WritableEvent::CreateEventPair(
+        kernel, Kernel::ResetType::Manual, "ISelfController:GpuErrorDetectedSystemEvent");
```
ogniK5377: might help if the event is renamed to `ISelfController:GpuErrorDetectedSystemEvent` to `IApplicationFunctions:GpuErrorDetectedSystemEvent` since the event is residing in `IApplicationFunctions` currently

mailwl: Oh, right, bad copy/paste, fixed

Issue #2817, opened 2019-08-29
Core Timing: Even share of timeslice per Guest CPU Core.


FernandoS27:
Emulation of the CPU occurs during timeslices in order to allow core timing to react to different timing events. In Citra, a single CPU Core ran on a set timeslice of N cycles and execution could be interrupted due to a kernel reschedule or reaching the downcount of N cycles. Furthermore, there's a mechanism called "ForceExceptionCheck" which reduces the downcount to the next timing event on the heap, thus improving accuracy.

In yuzu we have 4 Guest CPU Cores which have to be run concurrently. We adopted Citra's system in the next way: a timeslice is given for a full run on the 4 cores, this timeslice is unique and shared by all the 4 cores. Timing is continuous through each core execution. Whenever execution is interrupted in one core, execution is yield to the next cpu core. Core Timing events and Idling are only handled in Core 0.

**The issue:**
Using this system has been very affective at handling 4 guest CPU Cores within a single Host thread, sadly it has some draw backs:
1. Inbalanced load execution: CPU cores consume different amounts of the timeslice, thus making execution of each ambiguous.
2. Since timeslice is shared and execution is only interrupted on timing events/ kernel reschedules. There's still a chance for a core to take the timeslice hostage and consume the timeslice at full. Say the timeslice is 20000 cycles, Core 0 consumes 5000 and interrupts execution, yielding towards Core 1. Core 1 consumes the remaining cycles ending the execution pass. Core 2 and Core 3 did not execute anything as there's no timeslice left. The run is done again and the same things happen over and over. Eventually, Core 2 and Core 3 starve and a softlock may happen.
3. If Core 0 iddles, all the timeslice is consumed and no other Cores are executed. If Core 0 depends on Core 1,2,3 actions to reactivate, this results in a softlock.

**Previous work and attempts to fix this issue:**
In the past I reworked Core Timing experimentally separating timing handling. This was done the next way:
1. Each Core had a context which stored it's timeslice, cycle count, downcount and timer events.
2. All cores could: idle and advance core timing.
3. Execution pass was divided in 2 levels: first, a pass that goes and concurrently changes cores as they are interrupted. Second, a pass that repeats pass one until every core had consumed it's timeslice.

[...]

PR #2854, opened 2019-09-10
Core Timing: Rework Core Timing to run all cores evenly.

FernandoS27:
This PR adresses issue #2817 All the information on the implementation can be found there.

In accordance to the info on that issue. I went for the singlecore solution, expecting to have a different core timing class for multicore. The solution works by sharing time across cores but giving each core a separate slice of time. This keeps all cores advancing the same amount of ticks per full run (a full run is 40000 ticks total with 10000 ticks for each core). In general, of all the solutions, I've tried, this one seems to be the best approach with the better yields.

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic.cpp
@@ -116,7 +116,7 @@ public:
         num_interpreted_instructions = 0;
     }
     u64 GetTicksRemaining() override {
-        return std::max(parent.system.CoreTiming().GetDowncount(), 0);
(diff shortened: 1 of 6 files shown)


bunnei:
@FernandoS27 some more details would be helpful, such as which of the N solutions listed in the bug you are going for and justification

FernandoS27:
> @FernandoS27 some more details would be helpful, such as which of the N solutions listed in the bug you are going for and justification

Done.

FernandoS27:
I'll reopen this later after I have fully rewritten the core timing tests. Just closing it for the time being.

PR #2855, opened 2019-09-11
shader_ir/warp: Implement SHFL for Nvidia devices

ReinUsesLisp:
Implements SHFL (not `SHFL.VTG`, that's a different instruction) using Nvidia intrinsics.

On non-Nvidia devices SHFL is emulated with as a theoretical device with a warp size of one, having the same behaviour as NX hardware but with a single thread per warp. We won't have to do this on Vulkan with devices with the option of using a subgroup of 32 (gen9 Intel, Vega and Nvidia). That being said, SPIR-V instructions can't query if a thread is out of bounds. The IR is generic enough to handle three cases.

I'm sad about the math used to convert Nvidia's SHFL mask back into GLSL's width. What this mask means is unknown to me at the moment of writing this.

This also introduces the usage of `std::string_view` to avoid code repetition on the GLSL decompiler.

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -566,6 +566,13 @@ enum class ImageAtomicOperation : u64 {
     Exch = 8,
 };
 
+enum class ShuffleOperation : u64 {
+    Idx = 0,  // shuffleNV
+    Up = 1,   // shuffleUpNV
+    Down = 2, // shuffleDownNV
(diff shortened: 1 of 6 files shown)


amayra:
How is the state of Vulkan backend anyway ?

FernandoS27:
@ReinUsesLisp Rebase
@amayra Please use discord to ask such questions. PRs are suppose to only have posts related to the PR.

PR #2859, opened 2019-09-14
hid: Stub SetNpadJoyAssignmentModeSingle and GetNpadHandheldActivationMode

### LINKED ISSUE #2671: Missing HLE service command 'SetNpadJoyAssignmentModeSingle' in 'hid'
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/3357250/yuzu_log.txt)

Used when booting Pokémon Let's Go in docked mode.

Morph1984:
Here, we are leaving both `SetNpadJoyAssignmentModeSingleByDefault` and `SetNpadJoyAssignmentModeSingle` stubbed because we do not currently know fully the differences in implementation between both commands.

This PR also stubs `GetNpadHandheldActivationMode` and reorganizes some of the hid service commands based on the command ID.

This resolves #2671, **however this will not allow you to proceed past the controller selection screen when docked mode is enabled upon launching the game** (see #2865). Also, Pokemon Let's Go reports a `npad_id` of 32 when `SetNpadJoyAssignmentModeSingle` is called which ~~according to @Hexagon12 is incorrect~~ is the npad_id for "Handheld Docked Joycons."

It seems that the cause of Pokemon Let's Go not booting in the canary versions is due to the Joycons Docked option being ticked even though Docked mode is ticked causing it to call`SetNpadJoyAssignmentModeSingle`, which should not occur (according to @ogniK5377) and is a separate bug within the code. This is fixed by #2866.

Additionally, #2876 has fixed the detection of P1->P8 controllers, now if P1 is set to any controller except "none", `SetNpadJoyAssignmentModeSingle` would also be called. This causes the bugs encountered prior to this PR being tagged for canary. Also causes blending of docked mode and handheld controls.

~~HID code will need to be fixed/developed further to get single joycons to be detected which is not within the scope of this PR.~~ Fixed by #2876 

Also, running Pokemon: Let's Go in docked mode but retaining handheld controls (which can be done within yuzu by switching to docked mode while in game) is unintended behavior not true to the console.

When Pokemon: Let's Go is run when docked,

- The player has to use a single joycon or Poke Ball Plus
- Connecting and shaking a second joycon or Poke Ball Plus will initiate 2 player co-op
- Motion controls are **necessary to catch Pokemon**

All these features would require further development of the hid code to work.

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -203,13 +203,13 @@ Hid::Hid(Core::System& system) : ServiceFramework("hid"), system(system) {
         {120, &Hid::SetNpadJoyHoldType, "SetNpadJoyHoldType"},
         {121, &Hid::GetNpadJoyHoldType, "GetNpadJoyHoldType"},
         {122, &Hid::SetNpadJoyAssignmentModeSingleByDefault, "SetNpadJoyAssignmentModeSingleByDefault"},
-        {123, nullptr, "SetNpadJoyAssignmentModeSingleByDefault"},
+        {123, &Hid::SetNpadJoyAssignmentModeSingle, "SetNpadJoyAssignmentModeSingle"},
         {124, &Hid::SetNpadJoyAssignmentModeDual, "SetNpadJoyAssignmentModeDual"},
         {125, &Hid::MergeSingleJoyAsDualJoy, "MergeSingleJoyAsDualJoy"},
         {126, &Hid::StartLrAssignmentMode, "StartLrAssignmentMode"},
         {127, &Hid::StopLrAssignmentMode, "StopLrAssignmentMode"},
         {128, &Hid::SetNpadHandheldActivationMode, "SetNpadHandheldActivationMode"},
-        {129, nullptr, "GetNpadHandheldActivationMode"},
+        {129, &Hid::GetNpadHandheldActivationMode, "GetNpadHandheldActivationMode"},
         {130, &Hid::SwapNpadAssignment, "SwapNpadAssignment"},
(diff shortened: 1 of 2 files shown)


FernandoS27:
Uhm I'm not sure about this. @ogniK5377 Enlight us with your humble opinion.

ogniK5377:
npad_id of 32 is a valid npad id. However, if pokemon is having issues, we should investigate them first before this is merged

Morph1984:
> npad_id of 32 is a valid npad id(npad ids can be a value from 0->7, 16 and 32. I believe 32 might be a system-specific npad id, unsure on the details and 16 is the handheld npad_id?). However, if pokemon is having issues, we should investigate them first before this is merged

So, this is incorrect? (Line 66 of npad.h)

```cpp
enum class NPadAssignments : u32_le {
    Dual = 0,
    Single = 1,
};
```

Joaozin-tech:
@Morph1984 Dock Mode Control Mode Doesn't Work As Expected PR Requires One More Look Before Merging Branch Master
SharedScreenshot9

Morph1984:
@ogniK5377 Apparently, when you launch **any game** while having Single Joycons (Left or Right), games do not respond to any input from them, so this is definitely an issue within the HID code. Also npad_id 32 is handheld (and it seems the reason why PLG crashes in docked mode is because the "Joycons Docked (Handheld)" option is true even if docked mode is enabled)

ogniK5377:
Has there been any changes to this since #2876

Morph1984:
#2876 causes this PR to be necessary for PLG to boot on all modes.

bunnei:
@ogniK5377 can you review and signoff if this is OK to merge?

PR #2868, opened 2019-09-17
maxwell_to_gl: Fix mipmap filtering

ReinUsesLisp:
OpenGL texture filters follow `GL_<texture_filter>_MIPMAP_<mipmap_filter>` but we were using them in the opposite way.


niko1point0:
Will this fix the mipmap bug that occur when using Nouveau in homebrew?

FernandoS27:
@niko1point0 nope, that's entirely different issue.

PR #2869, opened 2019-09-18
shader/image: Implement SULD and fix SUATOM

ReinUsesLisp:
Implements SULD (`imageLoad`) using `GL_EXT_shader_image_load_formatted` when available. While we are at it, fix SUATOM related GLSL build issues.

This removes the image format guessing and always uses `r32ui` when atomic operations are present. Operations that will require signed/unsigned distinction (e.g. `MIN`) will have to use a compare-and-swap loop. The implementation for these has been removed until we find a game using it.


ogniK5377:
Rebase please @ReinUsesLisp

PR #2870, opened 2019-09-18
Implement a MME Draw commands Inliner and correct host instance drawing

FernandoS27:
This PR aims to accumulate draw calls sent through MME(Macro Interpreter) and flush then once either a sync is required or the program has ended. This heavily increases performance in certain games like SMO, PLGE which use instance drawing heavily. In the pokemon mansion in PLGE performance went from 12fps all the way to 50fps. Some kingdoms in SMO got a significant 5fps boost(changes depending on your computer).

Special thanks to fincs, since he discovered how MME's worked to do instance drawing and suggested this method.

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -92,6 +92,10 @@ void Maxwell3D::InitializeRegisterDefaults() {
 
     // Some games (like Super Mario Odyssey) assume that SRGB is enabled.
     regs.framebuffer_srgb = 1;
+    mme_inline[MAXWELL3D_REG_INDEX(draw.vertex_end_gl)] = true;
+    mme_inline[MAXWELL3D_REG_INDEX(draw.vertex_begin_gl)] = true;
+    mme_inline[MAXWELL3D_REG_INDEX(vertex_buffer.count)] = true;
+    mme_inline[MAXWELL3D_REG_INDEX(index_array.count)] = true;
 }
 
 #define DIRTY_REGS_POS(field_name) (offsetof(Maxwell3D::DirtyRegs, field_name))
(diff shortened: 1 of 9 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -778,11 +725,98 @@ void RasterizerOpenGL::DrawArrays() {
     if (texture_cache.TextureBarrier()) {
         glTextureBarrier();
     }
+}
+
+struct DrawParams {
```
ogniK5377: Initialize all members here

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -778,11 +725,98 @@ void RasterizerOpenGL::DrawArrays() {
...
+}
+
+void RasterizerOpenGL::DrawMultiArrays() {
+    DrawPrelude();
 
-    params.DispatchDraw();
+    auto& maxwell3d = system.GPU().Maxwell3D();
+    const auto& regs = maxwell3d.regs;
+    const auto& draw_setup = maxwell3d.mme_draw;
+    DrawParams draw_call;
```
ogniK5377:
```suggestion
    DrawParams draw_call{};
```

src/video_core/renderer_opengl/gl_rasterizer.cpp
FernandoS27: Done.

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -416,6 +423,92 @@ void Maxwell3D::CallMethod(const GPU::MethodCall& method_call) {
...
+    if (mme_inline[method]) {
+        regs.reg_array[method] = method_call.argument;
+        if (method == MAXWELL3D_REG_INDEX(vertex_buffer.count) ||
+            method == MAXWELL3D_REG_INDEX(index_array.count)) {
+            const MMMEDrawMode expected_mode = method == MAXWELL3D_REG_INDEX(vertex_buffer.count)
+                                                   ? MMMEDrawMode::Array
+                                                   : MMMEDrawMode::Indexed;
+            const u32 count = method_call.argument;
+            while (true) {
+                if (mme_draw.current_mode == MMMEDrawMode::Undefined) {
```
ReinUsesLisp: This should go inside a separated function:
```cpp
bool StepInstance(MMEDrawMode expected_mode, u32 count) {
    if (mme_draw.current_mode == MMEDrawMode::Undefined) {
        // ...
        return false;
    }
    if (mme_draw.current_mode == expected_mode) {
        // ...
        return false;
    }
    FlushMMEInlineDraw();
    return true;
}

File: src/video_core/engines/maxwell_3d.h [hunk 4]
```diff
@@ -1263,6 +1269,21 @@ class Maxwell3D final {
         return execute_on;
     }
 
+    enum class MMMEDrawMode : u32 {
```
ReinUsesLisp: Typo for `MMEDrawMode`?

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -778,11 +705,115 @@ void RasterizerOpenGL::DrawArrays() {
...
+};
+
+bool RasterizerOpenGL::DrawBatch(bool is_indexed) {
+    accelerate_draw = is_indexed ? AccelDraw::Indexed : AccelDraw::Arrays;
+
+    MICROPROFILE_SCOPE(OpenGL_Drawing);
+
+    auto& maxwell3d = system.GPU().Maxwell3D();
+    if (!maxwell3d.ShouldExecute()) {
+        return false;
```
ReinUsesLisp: This should return true, since we are "accelerating" (whatever that means nowadays) the draw. I would do two things:
1. Move this check to `Maxwell3D` and remove `ShouldExecute`.
2. Make `DrawBatch` return void. If we ever write a software backend, it would be with the same interface as `RasterizerOpenGL`, it won't be implemented in the caller.

src/video_core/rasterizer_interface.h
```diff
@@ -29,7 +29,10 @@ class RasterizerInterface {
     virtual ~RasterizerInterface() {}
 
     /// Draw the current batch of vertex arrays
-    virtual void DrawArrays() = 0;
+    virtual bool DrawBatch(bool is_indexed) = 0;
+
+    /// Draw the current batch of multiple instasnces of vertex arrays
```
MysticExile: `/// Draw the current batch of multiple instances of vertex arrays`

Nit: Spelling error

PR #2873, opened 2019-09-19
Initial implementation of Ioctl2 & Ioctl3

ogniK5377:
Purpose of Ioctl2 and Ioctl3 is to prevent the passing of raw pointers through ioctls.

For Ioctl2, instead of Nintendo passing a pointer to KickoffPB, they now pass a second buffer containing the command list.

For Ioctl3, Nintendo writes to an output buffer instead of a pointer.

Currently `KickoffPB` and `GetCharacteristics` have been updated as those are the only two observed functions which have been changed as of now. As more ioctls get called, we will need to update them accordingly.

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvdevice.h
+++ b/src/core/hle/service/nvdrv/devices/nvdevice.h
@@ -9,6 +9,7 @@
 #include "common/common_types.h"
 #include "common/swap.h"
 #include "core/hle/service/nvdrv/nvdata.h"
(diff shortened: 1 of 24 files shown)


ogniK5377:
Note, the file is clang formatted. Unsure why it's failing with azure

ogniK5377:
Do not merge until the clang-format issue has been looked into further

ogniK5377:
Matched clang-format from the server, this should be looked at in the future

PR #2876, opened 2019-09-19
AcquireNpadStyleSetUpdateEventHandle should have a separate event for each controller type

ogniK5377:
We shouldn't return the same event for all npads, this will lead to softlocks/input not working on some games. By setting a separate event for each controller, this should fix a few issues related to style updates.


ogniK5377:
Clang format has passed, just github is being drunk

PR #2877, opened 2019-09-20
Added frame_count for REV5 audio renderer

ogniK5377:
Added framecount. Need a game to actually read it to confirm that this is the frame count

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -107,7 +107,7 @@ Stream::State AudioRenderer::GetStreamState() const {
     return stream->GetState();
 }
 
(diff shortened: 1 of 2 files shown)


ogniK5377:
This fixes most 8.0.1 games which break on audren request update

src/audio_core/audio_renderer.cpp
```diff
@@ -107,6 +107,11 @@ Stream::State AudioRenderer::GetStreamState() const {
     return stream->GetState();
 }
 
+constexpr u32 VersionFromRevision(u32_le rev) {
```
lioncash: This needs to be marked static, otherwise this has external linkage.

ogniK5377:
Games from REV5 use this according to audren 8.0.1

PR #2878, opened 2019-09-20
Shader_IR: Implement ICMP.Instruction

FernandoS27:
This PR implements ICMP instruction.

### DIFF
--- a/src/video_core/shader/decode/arithmetic_integer.cpp
+++ b/src/video_core/shader/decode/arithmetic_integer.cpp
@@ -140,11 +140,11 @@ u32 ShaderIR::DecodeArithmeticInteger(NodeBlock& bb, u32 pc) {
     }
     case OpCode::Id::ICMP_CR:
     case OpCode::Id::ICMP_R:
-    case OpCode::Id::ICMP_RC: {
-        UNIMPLEMENTED_IF(instr.icmp.is_signed != 0);
+    case OpCode::Id::ICMP_RC:
+    case OpCode::Id::ICMP_IMM: {
         const Node zero = Immediate(0);
 
-        const auto [op_a, op_b] = [&]() -> std::tuple<Node, Node> {
+        const auto [op_b, test] = [&]() -> std::pair<Node, Node> {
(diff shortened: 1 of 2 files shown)


src/video_core/shader/decode/arithmetic_integer.cpp
```diff
@@ -138,6 +138,32 @@ u32 ShaderIR::DecodeArithmeticInteger(NodeBlock& bb, u32 pc) {
         SetRegister(bb, instr.gpr0, value);
         break;
     }
+    case OpCode::Id::ICMP_CR:
+    case OpCode::Id::ICMP_R:
+    case OpCode::Id::ICMP_RC: {
+        UNIMPLEMENTED_IF(instr.icmp.is_signed != 0);
+        const Node zero = Immediate(0);
+
+        const auto [op_a, op_b] = [&]() -> std::tuple<Node, Node> {
```
lioncash:
```suggestion
        const auto [op_a, op_b] = [&]() -> std::pair<Node, Node> {
```

generally requires less template instantiation work by the compiler.

src/video_core/shader/decode/arithmetic_integer.cpp
```diff
@@ -138,6 +138,35 @@ u32 ShaderIR::DecodeArithmeticInteger(NodeBlock& bb, u32 pc) {
...
+                        GetRegister(instr.gpr39)};
+            case OpCode::Id::ICMP_R:
+                return {GetRegister(instr.gpr20), GetRegister(instr.gpr39)};
+            case OpCode::Id::ICMP_RC:
+                return {GetRegister(instr.gpr39),
+                        GetConstBuffer(instr.cbuf34.index, instr.cbuf34.offset)};
+            case OpCode::Id::ICMP_IMM:
+                return {Immediate(instr.alu.GetSignedImm20_20()), GetRegister(instr.gpr39)};
+            default:
+                UNIMPLEMENTED();
```
ReinUsesLisp: This would be `UNREACHABLE`.

PR #2884, opened 2019-09-21
Remove usage of System::CurrentInterface() from most services

ogniK5377:
Removes use of this global accessor from most of the services. The filesystem, SM and the nvflinger BufferQueue still use this. Removal will be done in the future as they're used throughout the codebase

### DIFF
--- a/src/core/hle/service/hid/controllers/controller_base.cpp
+++ b/src/core/hle/service/hid/controllers/controller_base.cpp
@@ -9,12 +9,12 @@ namespace Service::HID {
 ControllerBase::ControllerBase() = default;
 ControllerBase::~ControllerBase() = default;
 
-void ControllerBase::ActivateController() {
+void ControllerBase::ActivateController(Core::System& system) {
     if (is_activated) {
         OnRelease();
     }
     is_activated = true;
-    OnInit();
+    OnInit(system);
 }
 
 void ControllerBase::DeactivateController() {
--- a/src/core/hle/service/hid/controllers/controller_base.cpp
+++ b/src/core/hle/service/hid/controllers/controller_base.cpp
@@ -6,7 +6,7 @@
 
 namespace Service::HID {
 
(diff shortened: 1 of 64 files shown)


src/core/hle/service/nvflinger/nvflinger.h
```diff
@@ -38,7 +38,7 @@ class BufferQueue;
 
 class NVFlinger final {
 public:
-    explicit NVFlinger(Core::Timing::CoreTiming& core_timing);
+    explicit NVFlinger(Core::Timing::CoreTiming& core_timing, Core::System& system);
```
lioncash: This can be simplified into a single parameter, since the CoreTiming instance is retrieved from the system instance.

src/core/hle/service/hid/controllers/gesture.h
```diff
@@ -12,11 +12,11 @@
 namespace Service::HID {
 class Controller_Gesture final : public ControllerBase {
 public:
-    Controller_Gesture();
+    Controller_Gesture(Core::System& system);
```
lioncash: explicit

src/core/hle/service/hid/controllers/mouse.h
```diff
@@ -14,11 +14,11 @@
 namespace Service::HID {
 class Controller_Mouse final : public ControllerBase {
 public:
-    Controller_Mouse();
+    Controller_Mouse(Core::System& system);
```
lioncash: Ditto (etc, for other constructors)

src/core/hle/service/hid/controllers/controller_base.cpp
```diff
@@ -6,7 +6,7 @@
 
 namespace Service::HID {
 
-ControllerBase::ControllerBase() = default;
+ControllerBase::ControllerBase(Core::System& system) : system(system){};
```
DarkLordZach: Remove semicolon at end

Issue #2715, opened 2019-07-11
Missing HLE service command 'InitializeSevenSixAxisSensor' in 'hid'


slashiee:
Needed for:
* The Legend of Zelda: Breath of the Wild (1.6.0) - [Reporter Log](https://github.com/yuzu-emu/yuzu/files/3380464/01007EF00011E000_2019-07-10T23-07-10.zip)
* Super Mario Odyssey (1.3.0) - [Reporter Log](https://github.com/yuzu-emu/yuzu/files/3380463/0100000000010000_2019-07-10T23-05-42.zip)
* Super Smash Bros. Ultimate (3.1.0) - [Reporter Log](https://github.com/yuzu-emu/yuzu/files/3380460/01006A800016E000_2019-07-10T23-00-47.zip)

PR #2894, opened 2019-09-22
Stub Hid::InitializeSevenSixAxisSensor

CrazyMax:
Issue #2715 

SMO 1.3.0 ingame now


ogniK5377:
This shouldn't be stubbed this way at all. Sampling numbers should be updated and the shared memory will need to have minor updating done to it. Not doing this will cause some games to have random softlocks/input will break etc

PR #2896, opened 2019-09-22
Port citra-emu/citra#4950: "Add FPS to SDL title bar"

FearlessTobi:
Port citra-emu/citra#4950: "Add FPS to SDL title bar"

**Original description**:
Also fix a small issue with incorrect shutdown ordering in SDL.
Previously the system would still be running so the telemetry task
didn't launch and detached_tasks would assert(count == 0)

I'm adding this so that we can visually have people test to see if theres a performance difference between SDL and Qt. We techincally could have them turn on frametime logging, but thats still slightly unreliable since we learned about how SwapBuffers() isn't included in frametime calculations.

### DIFF
--- a/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
+++ b/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
@@ -4,6 +4,8 @@
 
 #include <SDL.h>
 #include "common/logging/log.h"
+#include "common/scm_rev.h"
+#include "core/core.h"
 #include "input_common/keyboard.h"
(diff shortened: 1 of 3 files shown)


bunnei:
@FearlessTobi (and @jroweboy because this is your code) changes look fine - can you take a look at https://cla-assistant.io/yuzu-emu/yuzu when you get a chance so that we can merge this?

bunnei:
Just to confirm - FearlessTobi and jroweboy have both agreed to CLA, so we are good here now. Thanks everyone.

PR #2897, opened 2019-09-22
pl_u: Move open source font archives and fix NAND error

DarkLordZach:
Address issues that caused #2709 to be reverted.


src/core/hle/service/ns/pl_u.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <memory>
+#include "core/hle/kernel/physical_memory.h"
```
lioncash: Nothing in this header makes use of this include. It can be moved into the cpp file.

PR #2899, opened 2019-09-22
 shader/texture: Defer sampler types for TXQ

ReinUsesLisp:
Some instructions don't specify the texture type when they are used (like TXQ), on these cases defer the sampler declaration until an instruction using the same sampler defines it.
* Required for The Legend of Zelda: Link's Awakening

### DIFF
--- a/src/video_core/shader/decode/texture.cpp
+++ b/src/video_core/shader/decode/texture.cpp
@@ -141,7 +141,7 @@ u32 ShaderIR::DecodeTexture(NodeBlock& bb, u32 pc) {
         const Node component = Immediate(static_cast<u32>(instr.tld4s.component));
 
         const auto& sampler =
-            GetSampler(instr.sampler, TextureType::Texture2D, false, depth_compare);
+            GetSampler(instr.sampler, false, TextureType::Texture2D, false, depth_compare);
 
         Node4 values;
         for (u32 element = 0; element < values.size(); ++element) {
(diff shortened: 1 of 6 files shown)


FernandoS27:
This PR causes regressions in Zelda BotW.

ReinUsesLisp:
The regression is due to BotW using `TXQ` and never using it in other way, this will end up using Texture1D as default. The fast fix would be to change it to assume a Texture2D by default (the same behavior , meaning that games using any other texture type here will end up with a bugged texture on OpenGL and a (likely) crash on Vulkan.

The real solution would be to specialize depending on the used textures types, but this would require changing significantly the OpenGL shader cache. For now, I'll change it back to the first solution.

FernandoS27:
> The regression is due to BotW using `TXQ` and never using it in other way, this will end up using Texture1D as default. The fast fix would be to change it to assume a Texture2D by default (the same behavior , meaning that games using any other texture type here will end up with a bugged texture on OpenGL and a (likely) crash on Vulkan.
> 
> The real solution would be to specialize depending on the used textures types, but this would require changing significantly the OpenGL shader cache. For now, I'll fix it with the solution.

for the time being assume it's a Texture2D as that's what the previous code did.

src/video_core/shader/decode/texture.cpp
```diff
@@ -285,48 +279,70 @@ u32 ShaderIR::DecodeTexture(NodeBlock& bb, u32 pc) {
...
 
     // Otherwise create a new mapping for this sampler
+    if (is_type_known) {
+        samplers_with_known_type.insert(offset);
+    }
     const std::size_t next_index = used_samplers.size();
-    const Sampler entry{offset, next_index, type, is_array, is_shadow};
-    return *used_samplers.emplace(entry).first;
+    const Sampler entry(offset, next_index, type, is_array, is_shadow);
+    return used_samplers.emplace(std::make_pair(offset, entry)).first->second;
```
lioncash:
```suggestion
    return used_samplers.emplace(offset, entry).first->second;
```

Should still work

ReinUsesLisp:
A better approach for this can be done by reading the TIC while decoding and invalidating on changes. Closing in favor of that.

PR #2900, opened 2019-09-22
ci: Split mainline pipeline and add support for GitHub releases

DarkLordZach:
Adds:
- Caching for all builds
- Mock build for faster testing
- Templates for universal and github releases
- Split mainline pipeline to two parts, one for main and sub repos


chris062689:
Approvals for @bunnei and @jroweboy  --  merging.

PR #2908, opened 2019-09-24
fsp_srv: Implement GetFileTimeStampRaw

Morph1984:
This implements the GetFileTimeStampRaw command which returns the timestamp properties of a file (created, accessed, modified time).

Needed for some homebrew like Diablo-NX to boot.

Current issue with this implementation:
- Somehow it can't find the file in /sdmc (will investigate and fix soon)

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -367,6 +367,35 @@ u64 GetSize(FILE* f) {
     return size;
 }
 
+FileTimeStampRaw GetTimeStamp(const std::string& filename) {
+
+    FileTimeStampRaw timestamp = FileTimeStampRaw();
+
+    if (!Exists(filename)) {
+        LOG_ERROR(Common_Filesystem, "failed {}: No such file", filename);
+        return timestamp;
+    }
+
+    struct stat buf;
+#ifdef _WIN32
+    if (_wstat64(Common::UTF8ToUTF16W(filename).c_str(), &buf) == 0)
+#else
+    if (stat(filename.c_str(), &buf) == 0)
+#endif
+    {
+        LOG_TRACE(Common_Filesystem, "{}: Created={}, Accessed={}, Modified={}", filename, buf.st_ctime, buf.st_atime, buf.st_mtime);
+
+        timestamp.Created = buf.st_ctime;
+        timestamp.Accessed = buf.st_atime;
+        timestamp.Modified = buf.st_mtime;
+        
+        return timestamp;
+    }
+
+    LOG_ERROR(Common_Filesystem, "Stat failed {}: {}", filename, GetLastErrorMsg());
+    return timestamp;
+}
+
 bool CreateEmptyFile(const std::string& filename) {
     LOG_TRACE(Common_Filesystem, "{}", filename);
 
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -383,12 +383,13 @@ FileTimeStampRaw GetTimeStamp(const std::string& filename) {
     if (stat(filename.c_str(), &buf) == 0)
 #endif
     {
-        LOG_TRACE(Common_Filesystem, "{}: Created={}, Accessed={}, Modified={}", filename, buf.st_ctime, buf.st_atime, buf.st_mtime);
+        LOG_TRACE(Common_Filesystem, "{}: Created={}, Accessed={}, Modified={}", filename,
+                  buf.st_ctime, buf.st_atime, buf.st_mtime);
 
         timestamp.Created = buf.st_ctime;
         timestamp.Accessed = buf.st_atime;
         timestamp.Modified = buf.st_mtime;
-        
+
         return timestamp;
     }
 
--- a/src/common/file_util.h
+++ b/src/common/file_util.h
@@ -46,6 +46,13 @@ struct FSTEntry {
     std::vector<FSTEntry> children;
 };
 
+// Timestamps for GetFileTimeStampRaw
+struct FileTimeStampRaw {
+    u64 Created;
+    u64 Accessed;
+    u64 Modified;
+};
+
 // Returns true if file filename exists
 bool Exists(const std::string& filename);
 
@@ -60,6 +67,9 @@ u64 GetSize(const int fd);
 
 // Overloaded GetSize, accepts FILE*
 u64 GetSize(FILE* f);
+
+// Returns the timestamp of filename
+FileTimeStampRaw GetTimeStamp(const std::string& filename);
 
 // Returns true if successful, or path already exists.
 bool CreateDir(const std::string& filename);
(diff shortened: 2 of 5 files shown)


Hexagon12:
The description should indicate what was added.

src/core/hle/service/filesystem/filesystem.cpp
```diff
@@ -241,6 +241,15 @@ ResultVal<FileSys::EntryType> VfsDirectoryServiceWrapper::GetEntryType(
     return FileSys::ERROR_PATH_NOT_FOUND;
 }
 
+ResultVal<FileUtil::FileTimeStampRaw> VfsDirectoryServiceWrapper::GetFileTimeStampRaw(
+    const std::string& path_) const {
+    std::string path(FileUtil::SanitizePath(path_));
+    auto file = backing->GetFileRelative(path);
+    if (file == nullptr)
+        return FileSys::ERROR_PATH_NOT_FOUND;
+    return MakeResult(FileUtil::GetTimeStamp(path));
```
DarkLordZach: You can't do this....
The methods of VfsDirectoryServiceWrapper are passed a path relative to the root of the directory in that class. (i.e. the folder 'x' in the root of the SD card is `/x`, not `C:\...\...\x`)
However, the `FileUtil::GetTimeStamp` method takes a path on the current real filesystem.

These two paths can only possibly coincide if the SD root was set to system root on a non-windows system, which is really not a common usecase.

So, you will need to handle transforming the SD relative path into the real filesystem path.

But, that means that this still only works for only a fraction of directories accessed (outside of homebrew, SD is really not used in yuzu that much).

You will need to figure out how to get timestamps for files not directly on the OS, we cannot use a partial solution that doesn't even work in a majority of cases.

If you have further questions, please contact me on Discord. I actually might have a solution to this, but it is part of a PR that is much bigger scope and not close to being done.

src/core/hle/service/filesystem/filesystem.cpp
```diff
@@ -241,6 +241,15 @@ ResultVal<FileSys::EntryType> VfsDirectoryServiceWrapper::GetEntryType(
     return FileSys::ERROR_PATH_NOT_FOUND;
 }
 
+ResultVal<FileUtil::FileTimeStampRaw> VfsDirectoryServiceWrapper::GetFileTimeStampRaw(
+    const std::string& path_) const {
+    std::string path(FileUtil::SanitizePath(path_));
+    auto file = backing->GetFileRelative(path);
+    if (file == nullptr)
+        return FileSys::ERROR_PATH_NOT_FOUND;
```
DarkLordZach: is this hw-accurate?

src/common/file_util.cpp
```diff
@@ -367,6 +367,36 @@ u64 GetSize(FILE* f) {
     return size;
 }
 
+FileTimeStampRaw GetTimeStamp(const std::string& filename) {
```
DarkLordZach:
```suggestion
FileTimeStampRaw GetTimestamp(const std::string& filename) {
```

src/common/file_util.cpp
```diff
@@ -367,6 +367,36 @@ u64 GetSize(FILE* f) {
...
+
+    FileTimeStampRaw timestamp = FileTimeStampRaw();
```
DarkLordZach:
```suggestion
    FileTimeStampRaw timestamp;
```

src/core/hle/service/filesystem/filesystem.h
```diff
@@ -225,6 +226,14 @@ class VfsDirectoryServiceWrapper {
      */
     ResultVal<FileSys::EntryType> GetEntryType(const std::string& path) const;
 
+    /**
+     * Gets the creation, last accessed, and last modified timestamps of a File
+     * specified by its path
+     * @param path Path relative to the archive
+     * @return The timestamps of the File. They are expressed as a Unix timestamp
+     */
+    ResultVal<FileUtil::FileTimeStampRaw> GetFileTimeStampRaw(const std::string& path) const;
```
DarkLordZach:
```suggestion
    ResultVal<FileUtil::FileTimeStampRaw> GetFileTimestampRaw(std::string_view path) const;
```

src/core/hle/service/filesystem/filesystem.cpp
```diff
@@ -241,6 +241,15 @@ ResultVal<FileSys::EntryType> VfsDirectoryServiceWrapper::GetEntryType(
     return FileSys::ERROR_PATH_NOT_FOUND;
 }
 
+ResultVal<FileUtil::FileTimeStampRaw> VfsDirectoryServiceWrapper::GetFileTimeStampRaw(
+    const std::string& path_) const {
```
DarkLordZach:
```suggestion
    std::string_view path_) const {
```

src/common/file_util.h
```diff
@@ -61,6 +68,9 @@ u64 GetSize(const int fd);
 // Overloaded GetSize, accepts FILE*
 u64 GetSize(FILE* f);
 
+// Returns the timestamp of filename
+FileTimeStampRaw GetTimeStamp(const std::string& filename);
```
DarkLordZach:
```suggestion
FileTimeStampRaw GetTimestamp(std::string_view filename);
```

src/common/file_util.h
```diff
@@ -46,6 +46,13 @@ struct FSTEntry {
     std::vector<FSTEntry> children;
 };
 
+// Timestamps for GetFileTimeStampRaw
+struct FileTimeStampRaw {
```
DarkLordZach:
```suggestion
struct Timestamp {
```
imo File is redundant as its in the FileUtil namespace
also Raw doesn't really convey any addt'l info either

src/common/file_util.h
lioncash: This should be using `const std::string&` since its interfacing with POSIX functions that expect a null-terminated string (and std::string_view doesn't guarantee null-termination)

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -496,6 +496,29 @@ class IFileSystem final : public ServiceFramework<IFileSystem> {
         rb.Push(size.get_total_size());
     }
 
+    void GetFileTimeStampRaw(Kernel::HLERequestContext& ctx) {
+        const auto file_buffer = ctx.ReadBuffer();
+        const std::string name = Common::StringFromBuffer(file_buffer);
+
+        LOG_DEBUG(Service_FS, "called. file={}", name);
+
+        auto result = backend.GetFileTimeStampRaw(name);
```
lioncash:
```suggestion
        const auto result = backend.GetFileTimeStampRaw(name);
```

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -496,6 +496,29 @@ class IFileSystem final : public ServiceFramework<IFileSystem> {
...
+
+        if (result.Failed()) {
+            IPC::ResponseBuilder rb{ctx, 2};
+            rb.Push(result.Code());
+            return;
+        }
+
+        auto timestamp = result.Unwrap();
```
lioncash:
```suggestion
        const auto timestamp = result.Unwrap();
```

Morph1984:
I'll close this for now. Will be addressing it with another PR.

PR #2910, opened 2019-09-26
Port citra-emu/citra#4930: "Pause when in background"

FearlessTobi:
See citra-emu/citra#4930 for more details.


bunnei:
Purely FYI - @vitor-k has also agreed to CLA, so we are good to go here.

PR #2912, opened 2019-09-27
General fixes to Async GPU

FernandoS27:
In technical terms this PR takes care of correcting:

- Nvdrv Sync events handling.
- Guest vSync Management.
- Framelimiter management to CPU thread only.
- Accurate Fences on vSync.
- Remove hacked fence that forced the CPU to sync with the GPU every 5ms.
- Resolve possible race conditions on CPU Page mapping from the GPU.
- Force the system to wait for the GPU to idle before shutting down.

What it means for the user:
- New games boot with async (FE:3H, Cadence of Hyrule, Pokken Tournament)
- Fixed the issue where games would suddenly slowdown and report wrong framerates.
- Fixed the issue where some games would surpass frame limiting.
- Fixed the issue where dynamic resolution rescaler would trigger in SMO with fullspeed and won't trigger heavily for underspeed games.
- Removed artificial fence which limited async to 9ms. Now Async will limited by the game. This results in performance boosts in many graphic intensive games.

### DIFF
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include "common/assert.h"
+#include "common/microprofile.h"
 #include "core/core.h"
 #include "core/core_timing.h"
 #include "core/memory.h"
@@ -16,6 +17,8 @@
 #include "video_core/renderer_base.h"
 
 namespace Tegra {
+
+MICROPROFILE_DEFINE(GPU_wait, "GPU", "Wait for the GPU", MP_RGB(128, 128, 192));
 
 GPU::GPU(Core::System& system, VideoCore::RendererBase& renderer, bool is_async)
     : system{system}, renderer{renderer}, is_async{is_async} {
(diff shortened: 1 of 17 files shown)


src/video_core/gpu.cpp
```diff
@@ -63,6 +66,16 @@ const DmaPusher& GPU::DmaPusher() const {
     return *dma_pusher;
 }
 
+void GPU::WaitFence(u32 syncpoint_id, u32 value) const {
+    // Synced GPU, is always in sync
+    if (!is_async) {
+        return;
+    }
+    MICROPROFILE_SCOPE(GPU_wait);
+    while (syncpoints[syncpoint_id].load() < value) {
```
lioncash:
```suggestion
    while (syncpoints[syncpoint_id].load(std::memory_order_relaxed) < value) {
```

is sufficient here.

src/video_core/gpu_thread.cpp
```diff
@@ -96,16 +90,15 @@ void ThreadManager::FlushAndInvalidateRegion(CacheAddr addr, u64 size) {
     InvalidateRegion(addr, size);
 }
 
+void ThreadManager::WaitIdle() const {
+    while (state.last_fence > state.signaled_fence.load()) {
```
lioncash:
```suggestion
    while (state.last_fence > state.signaled_fence.load(std::memory_order_relaxed)) {
```

Ditto

PR #2914, opened 2019-09-27
Fermi2D: limit blit area to only available area

FernandoS27:
Normaly OpenGL does not care if the areas exceed the texture regions but other backends such as Vulkan do care about the limits of this areas. This PR crops the areas of the blit in order that they don't surpass the limits of the textures. This should help Vulkan and faulty OpenGL drivers.

### DIFF
--- a/src/video_core/engines/fermi_2d.cpp
+++ b/src/video_core/engines/fermi_2d.cpp
@@ -47,10 +47,20 @@ void Fermi2D::HandleSurfaceCopy() {
         src_blit_x2 = static_cast<u32>((regs.blit_src_x >> 32) + regs.blit_dst_width);
         src_blit_y2 = static_cast<u32>((regs.blit_src_y >> 32) + regs.blit_dst_height);
     }
-    const Common::Rectangle<u32> src_rect{src_blit_x1, src_blit_y1, src_blit_x2, src_blit_y2};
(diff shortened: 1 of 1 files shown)


ReinUsesLisp:
I'm not sure this is the right approach. `src` and `dest` are off by one pixel, e.g.: `[-1,1279]`. It looks more like an issue in the code above these changes.

FernandoS27:
> I'm not sure this is the right approach. `src` and `dest` are off by one pixel, e.g.: `[-1,1279]`. It looks more like an issue in the code above these changes.

I could look at my logical, it could also mean our base equations may be wrong. Some hardware tests could do.

src/video_core/engines/fermi_2d.cpp
```diff
@@ -28,6 +28,13 @@ void Fermi2D::CallMethod(const GPU::MethodCall& method_call) {
     }
 }
 
+std::pair<u32, u32> DelimitLine(u32 src_1, u32 src_2, u32 dst_1, u32 dst_2, u32 src_line) {
```
ReinUsesLisp: This should go in an anonymous namespace, at the top of the source file.

bunnei:
@FernandoS27 is this OK to be merged?

bunnei:
@FernandoS27 ping again

FernandoS27:
@bunnei yeah

PR #2915, opened 2019-09-28
Link to Yuzu's wiki instead of Citra's

vitor-k:
The original Contributing.md had differences, such as some links, so I created a page on yuzu's wiki based on the original file.


ReaperOfSouls1909:
they share somethings to save resources

vitor-k:
Yes, but on the first part, about reporting issues, citra's wiki links to Citra's FAQ and discord, instead of Yuzu's.
Citra's contributing guide had only one change since it moved to the wiki, so I don't believe there would be much back-and-forth between the pages. 
Even then, porting changes to the guide would be easier on the wiki, and the wiki shouldn't take up any "resources".

bunnei:
@vitor-k Could you please take a look a the CLA and agree to it when you get a chance? 
Learn more here: https://github.com/yuzu-emu/yuzu/wiki/Contributor-License-Agreement-Policy
Thank you.

PR #2917, opened 2019-09-28
TextureCache: Add the ability to deduce if two textures are depth on blit.

FernandoS27:
This PR adds another HLE mechanism to the texture cache. It turns out that Fermi2D blits depth formats as color formats, forcing us to do an expensive depth -> color conversion through buffer copies. With this PR, the texture cache will try to deduce if the textures are both depth (or 1 is depth & the other is not known) and will try to correct the formatting, in order to avoid expensive copies.

This work great on ZLA and DKCTF.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -224,8 +224,13 @@ public:
                      const Tegra::Engines::Fermi2D::Regs::Surface& dst_config,
                      const Tegra::Engines::Fermi2D::Config& copy_config) {
         std::lock_guard lock{mutex};
-        std::pair<TSurface, TView> dst_surface = GetFermiSurface(dst_config);
-        std::pair<TSurface, TView> src_surface = GetFermiSurface(src_config);
+        SurfaceParams src_params = SurfaceParams::CreateForFermiCopySurface(src_config);
+        SurfaceParams dst_params = SurfaceParams::CreateForFermiCopySurface(dst_config);
+        const GPUVAddr src_gpu_addr = src_config.Address();
+        const GPUVAddr dst_gpu_addr = dst_config.Address();
+        DeduceBestBlit(src_params, dst_params, src_gpu_addr, dst_gpu_addr);
+        std::pair<TSurface, TView> dst_surface = GetSurface(dst_gpu_addr, dst_params, true, false);
+        std::pair<TSurface, TView> src_surface = GetSurface(src_gpu_addr, src_params, true, false);
         ImageBlit(src_surface.second, dst_surface.second, copy_config);
         dst_surface.first->MarkAsModified(true, Tick());
     }
@@ -355,6 +360,29 @@ private:
         Ignore = 0,
         Flush = 1,
         BufferCopy = 3,
+    };
+
+    enum class DeductionType : u32 {
+        DeductionComplete,
+        DeductionIncomplete,
+        DeductionFailed,
+    };
+
+    struct Deduction {
+        DeductionType type{DeductionType::DeductionFailed};
+        TSurface surface{};
+
+        bool Failed() const {
+            return type == DeductionType::DeductionFailed;
+        }
+
+        bool Incomplete() const {
+            return type == DeductionType::DeductionIncomplete;
+        }
+
+        bool IsDepth() const {
+            return surface->GetSurfaceParams().IsPixelFormatZeta();
+        }
     };
 
     /**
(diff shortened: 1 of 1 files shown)


src/video_core/texture_cache/texture_cache.h
```diff
@@ -691,6 +719,118 @@ class TextureCache {
                               MatchTopologyResult::FullMatch);
     }
 
+    /**
+     * `DeduceSurface` gets the starting address and parameters of a candidate surface and tries
+     * to find a matching surface within the cache that's similar to it. If there are many textures
+     * or the texture found if entirely imcompatible, it will fail. If no texture is found, the
```
MysticExile:
```suggestion
     * or the texture found if entirely incompatible, it will fail. If no texture is found, the
```

src/video_core/texture_cache/texture_cache.h
```diff
@@ -691,6 +719,118 @@ class TextureCache {
...
+            Deduction result{};
+            result.type = DeductionType::DeductionComplete;
+            result.surface = overlaps[0];
+            return result;
+        }
+    }
+
+    /**
+     * `DeduceBestBlit` gets the a source and destination starting address and parameters,
+     * and tries to deduce if they are suppose to be depth textures. If so, their
```
MysticExile:
```suggestion
     * and tries to deduce if they are supposed to be depth textures. If so, their
```

src/video_core/texture_cache/texture_cache.h
```diff
@@ -691,6 +719,118 @@ class TextureCache {
...
+     * parameters are modified and fixed into so.
+     * @param gpu_addr, the starting address of the candidate surface.
+     * @param params, the paremeters on the candidate surface.
```
MysticExile:
```suggestion
     * @param params, the parameters on the candidate surface.
```

src/video_core/texture_cache/texture_cache.h
```diff
@@ -691,6 +719,118 @@ class TextureCache {
...
+            return;
+        }
+
+        const auto inherit_format = ([](SurfaceParams& to, TSurface from) {
+            const SurfaceParams& params = from->GetSurfaceParams();
+            to.pixel_format = params.pixel_format;
+            to.component_type = params.component_type;
+            to.type = params.type;
+        });
+        // Now we got the cases where one or both is Depth and the other is not known)
```
MysticExile:
```suggestion
        // Now we got the cases where one or both is Depth and the other is not known
```
Typo

src/video_core/texture_cache/texture_cache.h
```diff
@@ -691,6 +719,118 @@ class TextureCache {
...
+     **/
+    void DeduceBestBlit(SurfaceParams& src_params, SurfaceParams& dst_params,
+                        const GPUVAddr src_gpu_addr, const GPUVAddr dst_gpu_addr) {
+        auto deduc_src = DeduceSurface(src_gpu_addr, src_params);
+        auto deduc_dst = DeduceSurface(src_gpu_addr, src_params);
+        if (deduc_src.Failed() || deduc_dst.Failed()) {
+            return;
+        }
+
+        const bool imcomplete_src = deduc_src.Imcomplete();
```
MysticExile:
```suggestion
        const bool imcomplete_src = deduc_src.Incomplete();
```

src/video_core/texture_cache/texture_cache.h
```diff
@@ -691,6 +719,118 @@ class TextureCache {
...
+        const bool imcomplete_dst = deduc_dst.Imcomplete();
```
MysticExile:
```suggestion
        const bool imcomplete_dst = deduc_dst.Incomplete();
```
ditto

src/video_core/texture_cache/texture_cache.h
```diff
@@ -691,6 +719,118 @@ class TextureCache {
...
+    /**
+     * `DeduceBestBlit` gets the a source and destination starting address and parameters,
+     * and tries to deduce if they are supposed to be depth textures. If so, their
+     * parameters are modified and fixed into so.
+     * @param gpu_addr, the starting address of the candidate surface.
+     * @param params, the parameters on the candidate surface.
+     **/
+    void DeduceBestBlit(SurfaceParams& src_params, SurfaceParams& dst_params,
+                        const GPUVAddr src_gpu_addr, const GPUVAddr dst_gpu_addr) {
+        auto deduc_src = DeduceSurface(src_gpu_addr, src_params);
```
ReinUsesLisp: `deduced`

PR #2918, opened 2019-09-29
A setting for changing the sensitivity when using "WASD" keys 

Fueramanta:
This would be useful for those people who can't or don't have a controller, since for example, when playing diablo 3 can't move with the WASD keys as an analog would. 

There is an option for the RPCSX3 emulator where you can change the sensitivity from 0.5 to 0.1 for the WASD keys, maybe a similar solution could be found for this awesome emulator as well.

Thanks

### DIFF
--- a/.ci/templates/build-single.yml
+++ b/.ci/templates/build-single.yml
@@ -14,7 +14,7 @@ steps:
     cacheHitVar: CACHE_RESTORED
 - script: chmod a+x ./.ci/scripts/$(ScriptFolder)/exec.sh && ./.ci/scripts/$(ScriptFolder)/exec.sh
   displayName: 'Build'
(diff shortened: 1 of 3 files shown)


FearlessTobi:
If you have no idea how Github works, please don't use it. 
Otherwise, we might consider excluding you from interacting with this repository.

lioncash:
Hi, pull requests aren't meant to be used for feature requests. Please go to the Discord server for stuff like that or the forums.

Fueramanta:
> If you have no idea how Github works, please don't use it.
> Otherwise, we might consider excluding you from interacting with this repository.

If I have to deal with people as rude and not considerable as you, then go ahead, anyway nobody help, only complain. Btw, try to take it easy from now on, if you are like that in real life, you gonna have a tough life kid.
Ok Lioncash, didn't know that, I'm new as can be seen in my profile, thanks for responding kindly unlike others.

PR #2921, opened 2019-09-29
Services::ES fix casting warnings

FreddyFunk:
Some minor compiler warning fixed

### DIFF
--- a/src/core/hle/service/es/es.cpp
+++ b/src/core/hle/service/es/es.cpp
@@ -128,7 +128,7 @@ private:
     void CountCommonTicket(Kernel::HLERequestContext& ctx) {
         LOG_DEBUG(Service_ETicket, "called");
 
-        const auto count = keys.GetCommonTickets().size();
+        const u32 count = static_cast<u32>(keys.GetCommonTickets().size());
 
         IPC::ResponseBuilder rb{ctx, 3};
         rb.Push(RESULT_SUCCESS);
@@ -138,7 +138,7 @@ private:
     void CountPersonalizedTicket(Kernel::HLERequestContext& ctx) {
         LOG_DEBUG(Service_ETicket, "called");
 
-        const auto count = keys.GetPersonalizedTickets().size();
+        const u32 count = static_cast<u32>(keys.GetPersonalizedTickets().size());
 
         IPC::ResponseBuilder rb{ctx, 3};
         rb.Push(RESULT_SUCCESS);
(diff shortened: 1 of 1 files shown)


src/core/hle/service/es/es.cpp
```diff
@@ -128,7 +128,7 @@ class ETicket final : public ServiceFramework<ETicket> {
     void CountCommonTicket(Kernel::HLERequestContext& ctx) {
         LOG_DEBUG(Service_ETicket, "called");
 
-        const auto count = keys.GetCommonTickets().size();
+        const u32 count = static_cast<u32>(keys.GetCommonTickets().size());
```
ReinUsesLisp: No need to specify the type twice, use `auto` here.

src/core/hle/service/es/es.cpp
```diff
@@ -138,7 +138,7 @@ class ETicket final : public ServiceFramework<ETicket> {
     void CountPersonalizedTicket(Kernel::HLERequestContext& ctx) {
         LOG_DEBUG(Service_ETicket, "called");
 
-        const auto count = keys.GetPersonalizedTickets().size();
+        const u32 count = static_cast<u32>(keys.GetPersonalizedTickets().size());
```
ReinUsesLisp: Ditto

src/core/hle/service/es/es.cpp
FreddyFunk: yes, but then we would need to cast a few lines later:
https://github.com/yuzu-emu/yuzu/blob/7b93c4646a63ba0c30be6717edede190f8123008/src/core/hle/service/es/es.cpp#L135 

I do prefer this cast over this one, because we create a variable with the correct type in the first place:
`
rb.Push<u32>(static_cast<u32>(count));
`

src/core/hle/service/es/es.cpp
FreddyFunk: see comment above

src/core/hle/service/es/es.cpp
ReinUsesLisp: I meant using `auto foo = static_cast<u32>(bar);`

bunnei:
Hi @FreddyFunk - Just wanted to let you know that yuzu requires you to agree to our CLA before future pull requests can be merged. Please take a look at https://github.com/yuzu-emu/yuzu/wiki/Contributor-License-Agreement-Policy to learn more. When you are ready, you can click the "CLA not signed yet" button above to satisfy this requirement. Thank you!

FreddyFunk:
@bunnei CLA signed

PR #2928, opened 2019-10-01
maxwell_3d: Add dirty flags for depth bounds values

ReinUsesLisp:
This is useful in Vulkan where we want to update depth bounds without caring if it's enabled or disabled through `vkCmdSetDepthBounds`.


ReinUsesLisp:
Note that `depth_bounds` is used instead of `depth_bounds[0]` and `depth_bounds[1]` because `depth_bounds` is a `std::array` instead of a C array.

PR #2930, opened 2019-10-01
file_sys: Add code to access raw gamecard partitions and lazily load them

DarkLordZach:
This adds:
- Optimization that lazily loads parts of the gamecard
- Bug fix that allows the update part to be loaded -- this is the reason for the previous bullet point.
- Accessors for the raw part0/1 of the gamecard
- Missing implementations for gamecard certificate and IDs
- Accessors for sizes/offset maps in PFS

### DIFF
--- a/src/core/file_sys/card_image.cpp
+++ b/src/core/file_sys/card_image.cpp
@@ -139,6 +139,35 @@ VirtualDir XCI::GetLogoPartition() {
     return GetPartition(XCIPartition::Logo);
 }
 
+VirtualFile XCI::GetPartitionRaw(XCIPartition partition) const {
+    return partitions_raw[static_cast<std::size_t>(partition)];
+}
+
+VirtualFile XCI::GetSecurePartitionRaw() const {
+    return GetPartitionRaw(XCIPartition::Secure);
+}
+
+VirtualFile XCI::GetStoragePartition0() const {
+    return std::make_shared<OffsetVfsFile>(file, update_normal_partition_end, 0, "partition0");
+}
+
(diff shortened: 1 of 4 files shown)


src/core/file_sys/card_image.cpp
```diff
@@ -55,9 +57,7 @@ XCI::XCI(VirtualFile file_)
         const auto partition_idx = static_cast<std::size_t>(partition);
         auto raw = main_hfs.GetFile(partition_names[partition_idx]);
 
-        if (raw != nullptr) {
-            partitions[partition_idx] = std::make_shared<PartitionFilesystem>(std::move(raw));
-        }
+        partitions_raw[static_cast<std::size_t>(partition)] = raw;
```
lioncash:
```suggestion
        partitions_raw[static_cast<std::size_t>(partition)] = std::move(raw);
```

src/core/file_sys/partition_filesystem.cpp
```diff
@@ -65,6 +65,9 @@ PartitionFilesystem::PartitionFilesystem(std::shared_ptr<VfsFile> file) {
         std::string name(
             reinterpret_cast<const char*>(&file_data[strtab_offset + entry.strtab_offset]));
 
+        offsets[name] = content_offset + entry.offset;
```
lioncash:
```suggestion
        offsets.insert_or_assign(name, content_offset + entry.offset);
        sizes.insert_or_assign(name, entry.size);
```

bunnei:
@DarkLordZach code review feedback is minor here - can you address so we can get this merged? Thanks.

PR #2888, opened 2019-09-21
Shader_IR: Implement a full control flow decompiler for the shader IR.

FernandoS27:
This PR is a follow up to the recent Scan Flow Remake which introduced the elimination of the flow stack. In this PR we go an step further and eliminate all branches by transforming them to control flow structures (ifs and do-whiles) through the method specified in **"Taming control flow: A structured approach to eliminating goto statements"** by AM Erosa, LJ Hendren 1994. Furthermore, the glsl and spir-v decompilers have been enhanced to accept & compile AST structured code. In general, there are no performance gains but resulting code is a lot cleaner and a lot easier to process for less advanced driver compilers. Nevertheless, the old decompilation modes are still available in case a higher mode is not possible to use or the user desires (OpenGL backend will remain using NoFlowStack Mode).

The sum up from testing in most games:

- No speed advantages in NVIDIA OpenGL.
- Only OutwardMoves are implemented as no shader was found to require InwardMoves or Lifting.
- The decompiler has been tested in our Vulkan backend and unlike other modes, it does not crash in AMD Vulkan, even giving a nice performance increase touch.
- So far else derivation is implemented but it's disabled due to incompatibility with Spir-v backend and theoretical bugs it may introduce.
- This PR has been designed for easing the development in future Shader Mods as it makes code a lot more readable.
- Using a fully decompilation mode will help drivers which struggle to generate reasonable code.

PR #2944, opened 2019-10-05
video_core/shader: Minor changes

lioncash:
Makes a few minor amendments to the changes introduced within #2888. Notably, it also fixes a bug within the move constructor of ASTManager where all member variables wouldn't be transferred over to the new instance.

This also gets rid of a few instances where unnecessary atomic reference count increments and decrements were being performed due to std::shared_ptr copies.


FernandoS27:
Fix clang format please. Thanks in advance.

PR #2945, opened 2019-10-06
nifm: Only return that there's an internet connection when there's a BCATServer

FernandoS27:
This helps games that need internet for other purposes boot as the rest of our internet infrastructure is incomplete.


bunnei:
Merging this -- I don't think we should have long term PRs that are merged into our build system automatically, it's not very maintainable. I would be OK with it if there was a clear plan to finish internet connection implementation, but there is not -- and our BCAT implementation is also minimally used and not really in active development. Since our incomplete BCAT/network/internet implementation actually breaks games, this seems like a reasonably compromise.

PR #2951, opened 2019-10-06
core: Remove Core::CurrentProcess()

lioncash:
Migrates the HLE service code off the use of directly accessing the global system instance where trivially able to do so. This removes all usages of Core::CurrentProcess from the service code, only 8 occurrences of this function exist elsewhere. There's still quite a bit of "System::GetInstance()" being used, however this was able to replace a few instances.

Keeping around `Core::CurrentProcess` only encourages the use of the global system instance (which will be phased out long-term). Instead, we use the direct system function call directly to remove the appealing but discouraged short-hand.

### DIFF
--- a/src/core/hle/service/filesystem/filesystem.h
+++ b/src/core/hle/service/filesystem/filesystem.h
@@ -10,6 +10,10 @@
 #include "core/file_sys/vfs.h"
 #include "core/hle/result.h"
 
+namespace Core {
+class System;
+}
+
 namespace FileSys {
 class BISFactory;
 class RegisteredCache;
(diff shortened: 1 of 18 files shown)


src/core/hle/service/filesystem/filesystem.h
```diff
@@ -52,7 +56,7 @@ enum class ImageDirectoryId : u32 {
 
 class FileSystemController {
 public:
-    FileSystemController();
+    explicit FileSystemController(Core::System& system_);
```
ogniK5377: Is this intended? Shouldn't it just be `system`

lioncash: Yes. It's variable shadowing otherwise.

PR #2954, opened 2019-10-06
gl_shader_disk_cache: Properly ignore existing cache

ReinUsesLisp:
Previously old entries where appended to the file even if the shader cache was ignored at boot. Address that issue.


ReinUsesLisp:
Do not merge, this has a bug.

ReinUsesLisp:
Addressed. I forgot to enable the cache when it didn't exist or it was deleted.

PR #2956, opened 2019-10-07
ci: Add TAR-XZ compression to MSVC builds

DarkLordZach:
This makes the builds smaller in a low-dependency format.


Ghostotakoi:
Why not use 7zip compression, using zlib.dll instead???

DarkLordZach:
This has already been discussed internally.

PR #2965, opened 2019-10-09
Core Timing: Rework Core Timing to run all cores evenly.

FernandoS27:
This PR addresses issue #2817 All the information on the implementation can be found there.

In accordance to the info on that issue. I went for the singlecore solution, expecting to have a different core timing class for multicore. The solution works by sharing time across cores but giving each core a separate slice of time. This keeps all cores advancing the same amount of ticks per full run (a full run is 40000 ticks total with 10000 ticks for each core). In general, of all the solutions, I've tried, this one seems to be the best approach with the better yields.

Unlike the previous PR, I have remade the tests for this core timing implementation. Only the Threadsafe test has been entirely removed, with the fifo test being merged with the new FairSharing test.


src/core/arm/dynarmic/arm_dynarmic.cpp
```diff
@@ -116,7 +116,7 @@ class ARM_Dynarmic_Callbacks : public Dynarmic::A64::UserCallbacks {
         num_interpreted_instructions = 0;
     }
     u64 GetTicksRemaining() override {
-        return std::max(parent.system.CoreTiming().GetDowncount(), 0);
+        return std::max<s64>(parent.system.CoreTiming().GetDowncount(), 0LL);
```
lioncash:
```suggestion
        return std::max(parent.system.CoreTiming().GetDowncount(), s64{0});
```

src/core/arm/unicorn/arm_unicorn.cpp
```diff
@@ -156,7 +156,7 @@ void ARM_Unicorn::Run() {
     if (GDBStub::IsServerEnabled()) {
         ExecuteInstructions(std::max(4000000, 0));
     } else {
-        ExecuteInstructions(std::max(system.CoreTiming().GetDowncount(), 0));
+        ExecuteInstructions(std::max<s64>(system.CoreTiming().GetDowncount(), 0LL));
```
lioncash: Ditto

src/core/core_timing.cpp
```diff
@@ -38,10 +38,14 @@ CoreTiming::CoreTiming() = default;
 CoreTiming::~CoreTiming() = default;
 
 void CoreTiming::Initialize() {
-    downcount = MAX_SLICE_LENGTH;
+    for (std::size_t core = 0; core < num_cpu_cores; core++) {
```
lioncash:
```suggestion
    downcounts.fill(MAX_SLICE_LENGTH);
    time_slice.fill(MAX_SLICE_LENGTH);
```

src/core/core_timing.cpp
```diff
@@ -141,22 +146,35 @@ void CoreTiming::RemoveEvent(const EventType* event_type) {
...
+}
+
+std::optional<u64> CoreTiming::NextAvailableCore(const s64 needed_ticks) const {
+    const u64 original_context = current_context;
+    u64 next_context = (original_context + 1) % num_cpu_cores;
+    while (next_context != original_context) {
+        if (time_slice[next_context] >= needed_ticks) {
+            return {next_context};
+        } else if (time_slice[next_context] >= 0) {
+            return {};
```
lioncash:
```suggestion
            return std::nullopt;
```

Value initialization isn't preferable here, since it does more [work](https://godbolt.org/z/qX6mh4) than necessary.

src/core/core_timing.cpp
```diff
@@ -141,22 +146,35 @@ void CoreTiming::RemoveEvent(const EventType* event_type) {
...
+        }
+        next_context = (next_context + 1) % num_cpu_cores;
+    }
+    return {};
```
lioncash:
```suggestion
    return std::nullopt;
```

src/core/core_timing.cpp
```diff
@@ -173,24 +191,46 @@ void CoreTiming::Advance() {
...
+        }
+    }
+
+    accumulated_ticks = 0;
+
+    downcounts[current_context] = time_slice[current_context];
+}
+
+void CoreTiming::ResetRun() {
+    for (std::size_t core = 0; core < num_cpu_cores; core++) {
```
lioncash:
```suggestion
    downcounts.fill(MAX_SLICE_LENGTH);
    time_slice.fill(MAX_SLICE_LENGTH);
```

src/core/core_timing.cpp
```diff
@@ -173,24 +191,46 @@ void CoreTiming::Advance() {
...
+        downcounts[core] = MAX_SLICE_LENGTH;
+        time_slice[core] = MAX_SLICE_LENGTH;
+    }
+    current_context = 0;
+    // Still events left (scheduled in the future)
+    if (!event_queue.empty()) {
+        s64 needed_ticks = std::min<s64>(event_queue.front().time - global_timer, MAX_SLICE_LENGTH);
```
lioncash:
```suggestion
        const s64 needed_ticks = std::min<s64>(event_queue.front().time - global_timer, MAX_SLICE_LENGTH);
```

src/core/core_timing.h
```diff
@@ -104,18 +105,36 @@ class CoreTiming {
...
-    int GetDowncount() const;
+    void ResetRun();
+
+    s64 GetDowncount() const;
+
+    void SwitchContext(u64 new_context) {
+        current_context = new_context;
+    }
+
+    bool CurrentContextCanRun() const {
```
lioncash:
```suggestion
    bool CanCurrentContextRun() const {
```

Issue #1659, opened 2018-11-08
Input: Analog Triggers


FreddyFunk:
Analog triggers (from controllers like the XBOX controller) would not work due to a different range of their signals (from 0 to 255 on analog triggers instead of -32768 to 32768 on analog joysticks).

greggameplayer:
there is a sort of same behavior on shoulder from xbox controller

TheSIlentLink:
Same issue here on Ubuntu 18.10 with a 360 Controller

FreddyFunk:
This seems to be an SDL bug:
https://github.com/libretro/RetroArch/pull/7444

MysticExile:
This issue appears to have been fixed! If it is not, please open a new issue!

PR #2968, opened 2019-10-10
yuzu/configure_input_player: Fix input handling for ZL and ZR from controllers with analog triggers

FreddyFunk:
Analog triggers (from controllers like the XBOX controller) would not work due to a different range of their signals (from 0 to 255 on analog triggers instead of -32768 to 32768 on analog joysticks).
The SDL driver misinterprets analog triggers as analog joysticks.

Fixes:
#1659

### DIFF
--- a/src/yuzu/configuration/configure_input_player.cpp
+++ b/src/yuzu/configuration/configure_input_player.cpp
@@ -245,25 +245,24 @@ ConfigureInputPlayer::ConfigureInputPlayer(QWidget* parent, std::size_t player_i
 
         button->setContextMenuPolicy(Qt::CustomContextMenu);
         connect(button, &QPushButton::clicked, [=] {
-            HandleClick(
-                button_map[button_id],
-                [=](Common::ParamPackage params) {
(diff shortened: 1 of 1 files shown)


src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -247,7 +247,22 @@ ConfigureInputPlayer::ConfigureInputPlayer(QWidget* parent, std::size_t player_i
...
+                    // analog triggers as analog joysticks.
+                    // TODO: reinterpret the signal range for analog triggers to map the values
+                    // correctly. This is required for the correct emulation of the analog triggers
+                    // of the GameCube controller.
+                    if (button_id == Settings::NativeButton::ZL ||
+                        button_id == Settings::NativeButton::ZR) {
+                        params.Set("direction", "+");
+                        params.Set("threshold", "0.5");
+                    }
+                    buttons_param[button_id] = params;
```
lioncash:
```suggestion
                    buttons_param[button_id] = std::move(params);
```

PR #2971, opened 2019-10-12
Kernel: Implement a New Thread Scheduler V2

FernandoS27:
This PR is a continuation of the new scheduler (#2364) with significantly more fixes. The new scheduler is meant to be closer to the RE info we have and does a few things in a more appropriate manner. Core differences:

- The new scheduler does not automatically assign threads without an assigned cpu core. Instead it moves them to a suggested queue and resolves them when no scheduled thread can be assigned to the current core.
- All yield types are now implemented and can now be redundant accordingly (before the yielding thread went to sleep, leaving the possibility for a lower priority thread to be scheduled instead).
- Prepare reschedule has been revised and should now be triggered whenever a thread wakes up.
- Some corrections have been made to thread switching and picking threads on the go.

The new scheduler fixes some issues:
- Deadlocks that could occur with hle ipc wait and wait synchronization.
- Better handling of threads on some games, resulting in some less dynarmic usage.
- Unlike the old scheduler, no reads are reported to be running without being set to runs.

Additions, fixes and corrections in V2:
- New Scheduler V2 has been reworked to work with Fair Core Timing.
- All the comments signaled in #2364 have been addressed.
- The new scheduler has been fully checked against the RE twice now.
- Corrected a bug on Yields where candidate threads without assigned core where checked like if they were.
- Redundant yields will now add ticks to core timing in order to avoid yield bombing.
- Thread Preemption mechanism has been added.
- Mutexes and Condition Variable Threads will now have the correct result when waking up.
- Corrected some logic in address arbitering waiting count if equal signal.
- Wait objects correctly remove the thread from their list on timeout or cancel.
- Setting a Ready Thread as paused will correctly change that thread's scheduling.
- Shutting down the kernel will correctly clear the global scheduler.
- Most global accessors from thread.cpp have been removed.
- Added a few asserts to better picture possible kernel bugs.

Thanks to TuxSH and gdkchan for their work in reversing the switch's kernel.
Thanks to TuxSH' mesosphere and Ryujinx which served as reference to help validate this PR.

Depends on #2965

### DIFF
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -12,6 +12,7 @@
 
 #include "core/core.h"
 #include "core/core_timing.h"
+#include "core/core_timing_util.h"
 #include "core/hle/kernel/address_arbiter.h"
 #include "core/hle/kernel/client_port.h"
 #include "core/hle/kernel/handle_table.h"
@@ -96,6 +97,7 @@ struct KernelCore::Impl {
 
         InitializeSystemResourceLimit(kernel);
         InitializeThreads();
+        InitializePreemption();
     }
 
     void Shutdown() {
@@ -111,6 +113,7 @@ struct KernelCore::Impl {
 
         thread_wakeup_callback_handle_table.Clear();
         thread_wakeup_event_type = nullptr;
+        preemption_event = nullptr;
 
         named_ports.clear();
     }
@@ -133,6 +136,18 @@ struct KernelCore::Impl {
             system.CoreTiming().RegisterEvent("ThreadWakeupCallback", ThreadWakeupCallback);
     }
 
+    void InitializePreemption() {
+        preemption_event = system.CoreTiming().RegisterEvent(
+            "PreemptionCallback", [this](u64 userdata, s64 cycles_late) {
+                global_scheduler.PreemptThreads();
+                s64 time_interval = Core::Timing::msToCycles(std::chrono::milliseconds(10));
+                system.CoreTiming().ScheduleEvent(time_interval, preemption_event);
+            });
+
+        s64 time_interval = Core::Timing::msToCycles(std::chrono::milliseconds(10));
+        system.CoreTiming().ScheduleEvent(time_interval, preemption_event);
+    }
+
     std::atomic<u32> next_object_id{0};
     std::atomic<u64> next_kernel_process_id{Process::InitialKIPIDMin};
     std::atomic<u64> next_user_process_id{Process::ProcessIDMin};
@@ -146,6 +161,7 @@ struct KernelCore::Impl {
     SharedPtr<ResourceLimit> system_resource_limit;
 
     Core::Timing::EventType* thread_wakeup_event_type = nullptr;
+    Core::Timing::EventType* preemption_event = nullptr;
     // TODO(yuriks): This can be removed if Thread objects are explicitly pooled in the future,
     // allowing us to simply use a pool index or similar.
     Kernel::HandleTable thread_wakeup_callback_handle_table;
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -116,6 +116,8 @@ struct KernelCore::Impl {
         thread_wakeup_event_type = nullptr;
         preemption_event = nullptr;
 
+        global_scheduler.Shutdown();
+
         named_ports.clear();
     }
 
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -229,6 +229,14 @@ const Kernel::GlobalScheduler& KernelCore::GlobalScheduler() const {
     return impl->global_scheduler;
 }
 
+Core::System& KernelCore::System() {
+    return impl->system;
+}
+
+const Core::System& KernelCore::System() const {
+    return impl->system;
+}
+
 void KernelCore::AddNamedPort(std::string name, SharedPtr<ClientPort> port) {
     impl->named_ports.emplace(std::move(name), std::move(port));
 }
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -15,6 +15,7 @@
 #include "core/core_timing_util.h"
 #include "core/hle/kernel/address_arbiter.h"
 #include "core/hle/kernel/client_port.h"
+#include "core/hle/kernel/errors.h"
 #include "core/hle/kernel/handle_table.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/process.h"
@@ -60,12 +61,8 @@ static void ThreadWakeupCallback(u64 thread_handle, [[maybe_unused]] s64 cycles_
         if (thread->HasWakeupCallback()) {
             resume = thread->InvokeWakeupCallback(ThreadWakeupReason::Timeout, thread, nullptr, 0);
         }
-    }
-
-    if (thread->GetMutexWaitAddress() != 0 || thread->GetCondVarWaitAddress() != 0 ||
-        thread->GetWaitHandle() != 0) {
-        ASSERT(thread->GetStatus() == ThreadStatus::WaitMutex ||
-               thread->GetStatus() == ThreadStatus::WaitCondVar);
+    } else if (thread->GetStatus() == ThreadStatus::WaitMutex ||
+               thread->GetStatus() == ThreadStatus::WaitCondVar) {
         thread->SetMutexWaitAddress(0);
         thread->SetCondVarWaitAddress(0);
         thread->SetWaitHandle(0);
@@ -85,6 +82,10 @@ static void ThreadWakeupCallback(u64 thread_handle, [[maybe_unused]] s64 cycles_
     }
 
     if (resume) {
+        if (thread->GetStatus() == ThreadStatus::WaitCondVar ||
+            thread->GetStatus() == ThreadStatus::WaitArb) {
+            thread->SetWaitSynchronizationResult(RESULT_TIMEOUT);
+        }
         thread->ResumeFromWait();
     }
 }
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -18,6 +18,7 @@
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/process.h"
 #include "core/hle/kernel/resource_limit.h"
+#include "core/hle/kernel/scheduler.h"
 #include "core/hle/kernel/thread.h"
 #include "core/hle/lock.h"
 #include "core/hle/result.h"
@@ -140,6 +141,7 @@ struct KernelCore::Impl {
     // Lists all processes that exist in the current session.
     std::vector<SharedPtr<Process>> process_list;
     Process* current_process = nullptr;
+    Kernel::GlobalScheduler global_scheduler;
 
     SharedPtr<ResourceLimit> system_resource_limit;
 
@@ -201,6 +203,14 @@ const Process* KernelCore::CurrentProcess() const {
 
 const std::vector<SharedPtr<Process>>& KernelCore::GetProcessList() const {
     return impl->process_list;
+}
+
+Kernel::GlobalScheduler& KernelCore::GlobalScheduler() {
+    return impl->global_scheduler;
+}
+
+const Kernel::GlobalScheduler& KernelCore::GlobalScheduler() const {
+    return impl->global_scheduler;
 }
 
 void KernelCore::AddNamedPort(std::string name, SharedPtr<ClientPort> port) {
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -232,14 +232,6 @@ const Kernel::GlobalScheduler& KernelCore::GlobalScheduler() const {
     return impl->global_scheduler;
 }
 
-Core::System& KernelCore::System() {
-    return impl->system;
-}
-
-const Core::System& KernelCore::System() const {
-    return impl->system;
-}
-
 void KernelCore::AddNamedPort(std::string name, SharedPtr<ClientPort> port) {
     impl->named_ports.emplace(std::move(name), std::move(port));
 }
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -89,7 +89,7 @@ static void ThreadWakeupCallback(u64 thread_handle, [[maybe_unused]] s64 cycles_
 }
 
 struct KernelCore::Impl {
-    explicit Impl(Core::System& system) : system{system} {}
+    explicit Impl(Core::System& system) : system{system}, global_scheduler{system} {}
 
     void Initialize(KernelCore& kernel) {
         Shutdown();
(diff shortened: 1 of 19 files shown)


src/core/hle/kernel/kernel.cpp
```diff
@@ -203,6 +224,22 @@ const std::vector<SharedPtr<Process>>& KernelCore::GetProcessList() const {
...
 
+Kernel::GlobalScheduler& KernelCore::GlobalScheduler() {
+    return impl->global_scheduler;
+}
+
+const Kernel::GlobalScheduler& KernelCore::GlobalScheduler() const {
+    return impl->global_scheduler;
+}
+
+Core::System& KernelCore::System() {
```
lioncash: The kernel shouldn't be directly depending on the system. The intent of the kernel interface is so that it can exist without being directly tied to the system instance. This introduces coupling that shouldn't exist at all.

src/core/hle/kernel/scheduler.h
```diff
@@ -20,124 +20,189 @@ namespace Kernel {
...
+    std::array<u32, NUM_CPU_CORES> preemption_priorities = {59, 59, 59, 62};
+
+    /// Lists all thread ids that aren't deleted/etc.
+    std::vector<SharedPtr<Thread>> thread_list;
+    Core::System& system;
+};
+
+class Scheduler final {
+public:
+    explicit Scheduler(Core::System& system, Core::ARM_Interface& cpu_core, const u32 core_id);
```
lioncash:
```suggestion
    explicit Scheduler(Core::System& system, Core::ARM_Interface& cpu_core, u32 core_id);
```

src/core/hle/kernel/thread.cpp
```diff
@@ -408,6 +372,171 @@ void Thread::Sleep(s64 nanoseconds) {
...
+
+void Thread::SetSchedulingStatus(ThreadSchedStatus new_status) {
+    const u32 old_flags = scheduling_state;
+    scheduling_state =
+        (scheduling_state & ThreadSchedMasks::HighMask) | static_cast<u32>(new_status);
+    AdjustSchedulingOnStatus(old_flags);
+}
+
+void Thread::SetCurrentPriority(u32 new_priority) {
+    u32 old_priority = std::exchange(current_priority, new_priority);
```
lioncash:
```suggestion
    const u32 old_priority = std::exchange(current_priority, new_priority);
```

src/core/hle/kernel/thread.cpp
```diff
@@ -408,6 +372,171 @@ void Thread::Sleep(s64 nanoseconds) {
...
+            if (((mask >> core) & 1) != 0) {
+                return core;
+            }
+        }
+        return -1;
+    };
+
+    const bool use_override = affinity_override_count != 0;
+    // The value -3 is "do not change the ideal core".
+    if (new_core == -3) {
```
lioncash: We have [constants](https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/kernel/thread.h#L40) for this in the header. Please use them where applicable in the changed code

src/core/hle/kernel/thread.cpp
```diff
@@ -408,6 +372,171 @@ void Thread::Sleep(s64 nanoseconds) {
...
+        new_core = use_override ? ideal_core_override : ideal_core;
+        if ((new_affinity_mask & (1 << new_core)) == 0) {
```
lioncash:
```suggestion
        if ((new_affinity_mask & (1U << new_core)) == 0) {
```

src/core/hle/kernel/thread.h
```diff
@@ -75,6 +75,26 @@ enum class ThreadActivity : u32 {
...
 };
 
+enum class ThreadSchedStatus : u32 {
+    None = 0,
+    Paused = 1,
+    Runnable = 2,
+    Exited = 3,
+};
+
+enum ThreadSchedFlags : u32 {
```
lioncash: These should be enum classes. Ditto below.

src/core/hle/kernel/thread.h
```diff
@@ -409,6 +471,7 @@ class Thread final : public WaitObject {
 
     u64 total_cpu_time_ticks = 0; ///< Total CPU running ticks.
     u64 last_running_ticks = 0;   ///< CPU tick when thread was last running
+    u64 yield_count = 0;          ///< Number of innecessaries yields occured.
```
lioncash: This comment doesn't actually make sense and should probably be rewritten.

src/core/hle/kernel/thread.cpp
```diff
@@ -408,6 +372,171 @@ void Thread::Sleep(s64 nanoseconds) {
...
+    if (GetSchedulingStatus() != ThreadSchedStatus::Runnable) {
+        return;
+    }
+    auto& scheduler = kernel.System().GlobalScheduler();
+    if (processor_id >= 0) {
+        scheduler.Unschedule(old_priority, processor_id, this);
+    }
+
+    for (u32 core = 0; core < GlobalScheduler::NUM_CPU_CORES; core++) {
+        if (core != processor_id && ((affinity_mask >> core) & 1) != 0) {
```
lioncash: This results in a sign mismatch warning. core is a u32, but processor_id is an s32.

src/core/hle/kernel/thread.cpp
```diff
@@ -408,6 +372,171 @@ void Thread::Sleep(s64 nanoseconds) {
...
+    if (processor_id >= 0) {
+        if (current_thread == this) {
+            scheduler.SchedulePrepend(current_priority, processor_id, this);
+        } else {
+            scheduler.Schedule(current_priority, processor_id, this);
+        }
+    }
+
+    for (u32 core = 0; core < GlobalScheduler::NUM_CPU_CORES; core++) {
+        if (core != processor_id && ((affinity_mask >> core) & 1) != 0) {
```
lioncash: Ditto

src/core/hle/kernel/thread.cpp
```diff
@@ -408,6 +372,171 @@ void Thread::Sleep(s64 nanoseconds) {
...
+void Thread::AdjustSchedulingOnAffinity(u64 old_affinity_mask, s32 old_core) {
+    auto& scheduler = kernel.System().GlobalScheduler();
+    if (GetSchedulingStatus() != ThreadSchedStatus::Runnable ||
+        current_priority >= THREADPRIO_COUNT) {
+        return;
+    }
+
+    for (u32 core = 0; core < GlobalScheduler::NUM_CPU_CORES; core++) {
+        if (((old_affinity_mask >> core) & 1) != 0) {
+            if (core == old_core) {
```
lioncash: Ditto

src/core/hle/kernel/thread.cpp
```diff
@@ -408,6 +372,171 @@ void Thread::Sleep(s64 nanoseconds) {
...
+                scheduler.Unschedule(current_priority, core, this);
+            } else {
+                scheduler.Unsuggest(current_priority, core, this);
+            }
+        }
+    }
+
+    for (u32 core = 0; core < GlobalScheduler::NUM_CPU_CORES; core++) {
+        if (((affinity_mask >> core) & 1) != 0) {
+            if (core == processor_id) {
```
lioncash: Ditto

src/core/hle/kernel/svc.cpp
```diff
@@ -1572,10 +1576,13 @@ static void SleepThread(Core::System& system, s64 nanoseconds) {
         current_thread->Sleep(nanoseconds);
     }
 
-    // Reschedule all CPU cores
-    for (std::size_t i = 0; i < Core::NUM_CPU_CORES; ++i) {
-        system.CpuCore(i).PrepareReschedule();
+    if (redundant) {
+        // If it's redundant, the core is pretty much idle. Some games keep idling
+        // a core while it's doing nothing, we advance timing to avoid costly continuos
```
lioncash:
```suggestion
        // a core while it's doing nothing, we advance timing to avoid costly continuous
```

src/core/hle/kernel/svc.cpp
```diff
@@ -1553,17 +1556,18 @@ static void SleepThread(Core::System& system, s64 nanoseconds) {
 
     auto& scheduler = system.CurrentScheduler();
     auto* const current_thread = scheduler.GetCurrentThread();
+    bool redundant = false;
```
lioncash:
```suggestion
    bool is_redundant = false;
```

src/core/hle/kernel/kernel.h
```diff
@@ -25,6 +25,7 @@ class HandleTable;
 class Process;
 class ResourceLimit;
 class Thread;
+class GlobalScheduler;
```
lioncash: This should be alphabetically ordered.

src/core/hle/kernel/scheduler.h
```diff
@@ -20,124 +20,189 @@ namespace Kernel {
...
-     */
-    Thread* PopNextReadyThread();
+    bool AskForReselectionOrMarkRedundant(Thread* current_thread, Thread* winner);
+
+    static constexpr u32 min_regular_priority = 2;
+    std::array<Common::MultiLevelQueue<Thread*, THREADPRIO_COUNT>, NUM_CPU_CORES> scheduled_queue;
+    std::array<Common::MultiLevelQueue<Thread*, THREADPRIO_COUNT>, NUM_CPU_CORES> suggested_queue;
+    std::atomic<bool> reselection_pending;
+
+    std::array<u32, NUM_CPU_CORES> preemption_priorities = {59, 59, 59, 62};
```
lioncash: This should probably have a comment explaining the why the priorities are the way they are here.

src/core/hle/kernel/scheduler.h
```diff
@@ -20,124 +20,189 @@ namespace Kernel {
...
+    void TryDoContextSwitch();
+
+    void UnloadThread();
+
+    void SelectThreads();
+
+    /// Gets the current running thread
+    Thread* GetCurrentThread() const;
+
+    Thread* GetSelectedThread() const;
```
lioncash: Missing documentation comment.

src/core/hle/kernel/kernel.cpp
FernandoS27: Tell me an easier way to eliminate global accessors across the kernel.

src/core/hle/kernel/scheduler.h
```diff
@@ -152,19 +217,16 @@ class Scheduler final {
...
 
+    Core::System& system;
     Core::ARM_Interface& cpu_core;
     u64 last_context_switch_time = 0;
+    u64 idle_selection_count = 0;
+    const u32 core_id;
 
-    Core::System& system;
-    static std::mutex scheduler_mutex;
+    bool context_switch_pending = false;
```
lioncash:
```suggestion
    bool is_context_switch_pending = false;
```

src/core/hle/kernel/kernel.cpp
ogniK5377: Yeah removing it as a global accessor it a bit out of this PR. However, we should address this in the future since a few things need to still be reworked for it to be removed.

lioncash: That's not how it works. If you have a design issue like this, you get it out of the way first, not after the fact. The kernel should not be depending on the overall system itself in order to function. Even more, a subsystem of the system instance shouldn't be exposing said system instance. That literally doesn't make sense from a design standpoint and indicates misplaced responsibilities.

This mostly stems from the schedulers existing as separate entities from the kernel, which is not how they should be separated. The schedulers themselves are a part of the overall kernel state in Horizon OS, not the encompassing system.

Please don't break the separation of the interfaces. It's in place for a reason (and clearly works, because it's making it obvious when sketchy stuff like this is done).

FernandoS27: Well I'll reverse it but please don't put on me eliminating global accessors when simply the current interface won't let me.

src/core/hle/kernel/thread.h
```diff
@@ -82,19 +82,25 @@ enum class ThreadSchedStatus : u32 {
...
 
-enum ThreadSchedMasks : u32 {
+enum class ThreadSchedMasks : u32 {
     LowMask = 0x000f,
     HighMask = 0xfff0,
     ForcePauseMask = 0x0070,
 };
 
+enum class CoreFlags : s32 {
+    IgnoreIdealCore = -1,
```
lioncash: We already have enum values for these in this exact header. See lines 33-40. They don't need to be duplicated into another enum.

src/core/hle/kernel/thread.cpp
```diff
@@ -444,14 +443,14 @@ void Thread::AdjustSchedulingOnStatus(u32 old_flags) {
...
-    if (static_cast<ThreadSchedStatus>(old_flags & ThreadSchedMasks::LowMask) ==
+    if (static_cast<ThreadSchedStatus>(old_flags & static_cast<u32>(ThreadSchedMasks::LowMask)) ==
         ThreadSchedStatus::Runnable) {
         // In this case the thread was running, now it's pausing/exitting
         if (processor_id >= 0) {
             scheduler.Unschedule(current_priority, processor_id, this);
         }
 
-        for (u32 core = 0; core < GlobalScheduler::NUM_CPU_CORES; core++) {
+        for (s32 core = 0; core < GlobalScheduler::NUM_CPU_CORES; core++) {
```
lioncash: This and the following similar changes, now cause a signed/unsigned conversion warning in the `Suggest()`/`Unsuggest()` calls, since they expect an unsigned input.

src/core/hle/kernel/kernel.cpp
lioncash: When you design a new subsystem to replace an old one. The onus is more-or-less on the creator of that subsystem to address those issues before putting the new system in place. Failure to do that more often than not results in scenarios like this where subsystems overexpose their contained members to make up for the deficiency in the interface elsewhere.

It's clearly an issue that was encountered when the new incarnation of the scheduler was made, given it's in the diff of this pull request.

src/core/hle/kernel/thread.h
```diff
@@ -35,6 +35,9 @@ enum ThreadProcessorId : s32 {
     /// Run thread on the ideal core specified by the process.
     THREADPROCESSORID_IDEAL = -2,
 
+    /// when setting Core and Affiny, keeps the ideal core intact
+    THREADDONTCHANGE_IDEAL = -3,
```
lioncash: THREADPROCESSORID_DONT_UPDATE below this performs the exact same behavior. This isn't necessary.

src/core/hle/kernel/kernel.cpp
FernandoS27: That was design inherited from the old scheduler. I tried reversing it but it was a mess with dependencies to the cpu_cores and arm interfaces. It goes out of the scope of this PR and I'm entirely unwilling on doing such a major refactor.

lioncash: > It goes out of the scope of this PR 

Indeed. It should have gone into a PR before this one.

> I'm entirely unwilling on doing such a major refactor. 

If that's the case then it shouldn't come as a surprise that perpetuation of a shortcoming in a previous system gets called out.

FernandoS27: It's just hell on earth as it touches all the ARMInterface Intitialization code as it now needs to setup in the kernel. Most global accessors in thread.cpp and wait_objects.cpp are due to stopping CPU Cores and accessing Core Timing. Both things are highly tight to the system class, personally I have no idea how to make shortcuts to the CPU cores in the systems to the kernel as it makes no sense from our design structure. Core timing can easily be added in a reference in the kernel as Core Timing represents the Hardware Clock and that's situated in Horizon OS's kernel.

PR #2972, opened 2019-10-12
{bcat, gpu, nvflinger}: Remove trivial usages of the global system accessor

lioncash:
Eliminates the reliance on global state in a few areas. These are relatively trivial usages that are able to be replaced via passing along the system instance as part of function interfaces. In some cases, it's as simple as just using a member variable instead of the accessor.

Brings us a little close to dropping the global singleton.


bunnei:
@lioncash CI failures look real?

PR #2974, opened 2019-10-12
Switching for surfaces in texture cache #2

3ger:
This PR removes the need for cpu page shrink/growth by "swapping" surfaces instead of registering/unregistering them in texture cache.
This should boost the performance in some cases a little. I think with more work on this, performance can be improved further.

Credits to @FernandoS27 for the idea & help.

Resubmit of #2973 - because of cla problems.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -344,21 +344,25 @@ protected:
             UnregisterInnerCache(old_surface);
             old_surface->MarkAsRegistered(false);
             ReserveSurface(old_surface->GetSurfaceParams(), old_surface);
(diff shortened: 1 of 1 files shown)


lioncash:
Commit message itself should probably be more informative. Looks fine other than that.

FernandoS27:
I'll remove this from the next 2 mainline builds as I'm tracking a regression.

theboy181:
Is there a specific game we can test for performance, with this update?

choppymaster:
Why this PR was never merged? The current new texture cache is #3 nah?

bunnei:
Since the merge of #4967, this is no longer needed.

PR #2979, opened 2019-10-15
video_core/macro_interpreter: Make definitions of most private enums/unions hidden

lioncash:
This allows the implementation of these types to change without requiring a rebuild of everything that includes the macro interpreter header.


src/video_core/macro_interpreter.cpp
```diff
@@ -12,6 +12,76 @@ MICROPROFILE_DEFINE(MacroInterp, "GPU", "Execute macro interpreter", MP_RGB(128,
 
 namespace Tegra {
 
+enum class MacroInterpreter::Operation : u32 {
```
ReinUsesLisp: Should these go in an anonymous namespace?

lioncash: Done

PR #2983, opened 2019-10-15
gl_shader_decompiler/vk_shader_decompiler: Resolve implicit fallthrough cases

lioncash:
The `NeverExecute` and `UnusedIndex` entries were currently having their target overwritten and replaced with the default case. This amends that.


FernandoS27:
nice catch

PR #2987, opened 2019-10-16
Texture_Cache: Redo invalid Surfaces handling.

FernandoS27:
This PR aims to redo the full setup of invalid textures in the texture cache. Invalid textures are when the information provided by such is invalid. According to RE, this is as possible as binding a nullptr in a texture bind in OpenGL. Sadly not all invalid textures may actually be invalid but bad fetches from the buffer or other synchronization misses. In general, this gives the texture cache the ability to track them and correct them with a dummy texture that acts as hardware acts when it finds an invalid texture.

### DIFF
--- a/src/video_core/texture_cache/surface_params.h
+++ b/src/video_core/texture_cache/surface_params.h
@@ -45,6 +45,14 @@ public:
     static SurfaceParams CreateForFermiCopySurface(
         const Tegra::Engines::Fermi2D::Regs::Surface& config);
 
+    /// Obtains the texture target from a shader's sampler entry.
+    static VideoCore::Surface::SurfaceTarget ExpectedTarget(
+        const VideoCommon::Shader::Sampler& entry);
+
+    /// Obtains the texture target from a shader's sampler entry.
+    static VideoCore::Surface::SurfaceTarget ExpectedTarget(
+        const VideoCommon::Shader::Image& entry);
+
     std::size_t Hash() const {
         return static_cast<std::size_t>(
             Common::CityHash64(reinterpret_cast<const char*>(this), sizeof(*this)));
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -95,10 +95,16 @@ public:
         std::lock_guard lock{mutex};
         const auto gpu_addr{tic.Address()};
         if (!gpu_addr) {
-            return {};
+            return GetNullSurface(SurfaceParams::ExpectedTarget(entry));
+        }
+
+        const auto host_ptr{system.GPU().MemoryManager().GetPointer(gpu_addr)};
+        const auto cache_addr{ToCacheAddr(host_ptr)};
+        if (!cache_addr) {
+            return GetNullSurface(SurfaceParams::ExpectedTarget(entry));
         }
         const auto params{SurfaceParams::CreateForTexture(format_lookup_table, tic, entry)};
-        const auto [surface, view] = GetSurface(gpu_addr, params, true, false);
+        const auto [surface, view] = GetSurface(gpu_addr, cache_addr, params, true, false);
         if (guard_samplers) {
             sampled_textures.push_back(surface);
         }
@@ -110,10 +116,15 @@ public:
         std::lock_guard lock{mutex};
         const auto gpu_addr{tic.Address()};
         if (!gpu_addr) {
-            return {};
+            return GetNullSurface(SurfaceParams::ExpectedTarget(entry));
+        }
+        const auto host_ptr{system.GPU().MemoryManager().GetPointer(gpu_addr)};
+        const auto cache_addr{ToCacheAddr(host_ptr)};
+        if (!cache_addr) {
+            return GetNullSurface(SurfaceParams::ExpectedTarget(entry));
         }
         const auto params{SurfaceParams::CreateForImage(format_lookup_table, tic, entry)};
-        const auto [surface, view] = GetSurface(gpu_addr, params, true, false);
+        const auto [surface, view] = GetSurface(gpu_addr, cache_addr, params, true, false);
         if (guard_samplers) {
             sampled_textures.push_back(surface);
         }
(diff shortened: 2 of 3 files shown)


src/video_core/texture_cache/surface_params.h
```diff
@@ -41,6 +41,12 @@ class SurfaceParams {
     static SurfaceParams CreateForFermiCopySurface(
         const Tegra::Engines::Fermi2D::Regs::Surface& config);
 
+    static VideoCore::Surface::SurfaceTarget ExpectedTarget(
```
lioncash: These are missing documentation comments

src/video_core/texture_cache/texture_cache.h
```diff
@@ -776,6 +787,44 @@ class TextureCache {
         }
     }
 
+    /**
+     * `GetNullSurface` gets a null surface based on a target texture.
```
lioncash:
```suggestion
     * Gets a null surface based on a target texture.
```

This way you don't need to remember to change the documentation comment if the function name changes.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -776,6 +787,44 @@ class TextureCache {
...
+     * @param target, The target of the null surface.
```
lioncash:
```suggestion
     * @param target The target of the null surface.
```

Otherwise this is an invalid Doxygen reference

src/video_core/texture_cache/texture_cache.h
```diff
@@ -776,6 +787,44 @@ class TextureCache {
...
+     **/
+    TView GetNullSurface(SurfaceTarget target) {
+        const u32 i_target = static_cast<u32>(target);
+        auto it = invalid_cache.find(i_target);
+        if (it != invalid_cache.end()) {
```
lioncash:
```suggestion
        if (const auto it = invalid_cache.find(i_target); it != invalid_cache.end()) {
```

src/video_core/texture_cache/texture_cache.h
```diff
@@ -776,6 +787,44 @@ class TextureCache {
...
+            TSurface surface = it->second;
+            return surface->GetMainView();
```
lioncash:
```suggestion
            return it->second->GetMainView();
```

This way you don't make an unnecessary surface copy.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -776,6 +787,44 @@ class TextureCache {
...
+        params.depth = 1;
+        params.pitch = 4;
+        params.num_levels = 1;
+        params.emulated_levels = 1;
+        params.pixel_format = VideoCore::Surface::PixelFormat::RGBA16F;
+        params.component_type = VideoCore::Surface::ComponentType::Float;
+        params.type = VideoCore::Surface::SurfaceType::ColorTexture;
+        auto surface = CreateSurface(0ULL, params);
+        invalid_memory.resize(surface->GetHostSizeInBytes());
+        std::fill(invalid_memory.begin(), invalid_memory.end(), 0U);
```
lioncash: You can just perform a clear and then a resize(), which will perform the same behavior

src/video_core/texture_cache/texture_cache.h
```diff
@@ -970,6 +1019,9 @@ class TextureCache {
 
     std::vector<TSurface> sampled_textures;
 
+    std::unordered_map<u32, TSurface> invalid_cache;
```
lioncash: These should probably have comments like the previous variables.

lioncash:
The commit itself should likely contain the explanation within the PR comment as well, so it's actually in the repository history.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -615,22 +639,9 @@ class TextureCache {
...
-            new_params.width = 1;
-            new_params.height = 1;
-            new_params.depth = 1;
-            new_params.block_height = 0;
-            new_params.block_depth = 0;
-            return InitializeSurface(gpu_addr, new_params, false);
-        }
+    std::pair<TSurface, TView> GetSurface(const GPUVAddr gpu_addr, const CacheAddr cache_addr,
+                                          const SurfaceParams& params, bool preserve_contents,
+                                          bool is_render) {
```
ReinUsesLisp: Unnecessary space here.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -794,6 +805,42 @@ class TextureCache {
         }
     }
 
+    /**
+     * Gets a null surface based on a target texture.
+     * @param target The target of the null surface.
+     **/
```
ReinUsesLisp: `*/`

src/video_core/texture_cache/texture_cache.h
```diff
@@ -991,6 +1038,11 @@ class TextureCache {
 
     std::vector<TSurface> sampled_textures;
 
+    /// This cache stores null surfaces in order to be used as a placeholder
+    /// for invalid texture calls.
+    std::unordered_map<u32, TSurface> invalid_cache;
+    std::vector<u8> invalid_memory{};
```
ReinUsesLisp: Unnecessary `{}`.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -991,6 +1038,11 @@ class TextureCache {
 
     std::vector<TSurface> sampled_textures;
 
+    /// This cache stores null surfaces in order to be used as a placeholder
+    /// for invalid texture calls.
+    std::unordered_map<u32, TSurface> invalid_cache;
```
ReinUsesLisp: This can be a `std::array`, `SurfaceTarget` is an ordered enumeration.

src/video_core/texture_cache/texture_cache.h
FernandoS27: ?

src/video_core/texture_cache/texture_cache.h
FernandoS27: ? it follows the commenting style. no idea what you mean.

src/video_core/texture_cache/texture_cache.h
FernandoS27: This may change in the future if the expected texture is a depth texture or not, so I prefer keeping the unordered map, as the key may change.

src/video_core/texture_cache/texture_cache.h
lioncash: The trailing part of a doxygen comment is of the form `*/`, not `**/`

src/video_core/texture_cache/texture_cache.h
ReinUsesLisp: I swear there was an empty line here.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -229,8 +254,14 @@ class TextureCache {
         const GPUVAddr src_gpu_addr = src_config.Address();
         const GPUVAddr dst_gpu_addr = dst_config.Address();
         DeduceBestBlit(src_params, dst_params, src_gpu_addr, dst_gpu_addr);
-        std::pair<TSurface, TView> dst_surface = GetSurface(dst_gpu_addr, dst_params, true, false);
-        std::pair<TSurface, TView> src_surface = GetSurface(src_gpu_addr, src_params, true, false);
+        const auto dst_host_ptr{system.GPU().MemoryManager().GetPointer(dst_gpu_addr)};
```
ReinUsesLisp: Put `system.GPU().MemoryManager()` in a temporary to avoid querying it twice.

ReinUsesLisp:
This has soft conflicts with latest master. I'm untagging for now.

src/video_core/texture_cache/texture_cache.h
FernandoS27: oh then we need another PR to change all the comments in the texture cache

PR #2995, opened 2019-10-18
shader_ir/memory: Ignore global memory when tracking fails

ReinUsesLisp:
Ignore global memory operations instead of invoking undefined behaviour when constant buffer tracking fails and we are blasting through asserts, ignore the operation.

In the case of LDG this means filling the destination registers with zeroes; for STG this means ignore the instruction as a whole.

The default behaviour is still to abort execution on failure.

While we are at it, remove the usage of unnecessary temporaries for STG.

### DIFF
--- a/src/video_core/shader/decode/memory.cpp
+++ b/src/video_core/shader/decode/memory.cpp
@@ -166,9 +166,17 @@ u32 ShaderIR::DecodeMemory(NodeBlock& bb, u32 pc) {
         }();
 
         const auto [real_address_base, base_address, descriptor] =
-            TrackAndGetGlobalMemory(bb, instr, false);
+            TrackGlobalMemory(bb, instr, false);
 
         const u32 count = GetUniformTypeElementsCount(type);
+        if (!real_address_base || !base_address) {
+            // Tracking failed, load zeroes.
+            for (u32 i = 0; i < count; ++i) {
+                SetRegister(bb, instr.gpr0.Value() + i, Immediate(0.0f));
+            }
+            break;
+        }
+
         for (u32 i = 0; i < count; ++i) {
             const Node it_offset = Immediate(i * 4);
             const Node real_address =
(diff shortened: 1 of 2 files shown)


src/video_core/shader/decode/memory.cpp
```diff
@@ -301,15 +306,18 @@ u32 ShaderIR::DecodeMemory(NodeBlock& bb, u32 pc) {
...
     const auto addr_register{GetRegister(instr.gmem.gpr)};
     const auto immediate_offset{static_cast<u32>(instr.gmem.offset)};
 
     const auto [base_address, index, offset] =
         TrackCbuf(addr_register, global_code, static_cast<s64>(global_code.size()));
-    ASSERT(base_address != nullptr);
+    if (!base_address) {
+        UNREACHABLE_MSG("Global memory tracking failed");
+        return {nullptr, nullptr, {}};
+    }
```
FernandoS27: Couldn't we use ASSERT_OR_EXECUTE here?

src/video_core/shader/decode/memory.cpp
```diff
@@ -301,15 +306,17 @@ u32 ShaderIR::DecodeMemory(NodeBlock& bb, u32 pc) {
...
+std::tuple<Node, Node, GlobalMemoryBase> ShaderIR::TrackGlobalMemory(NodeBlock& bb,
+                                                                     Instruction instr,
+                                                                     bool is_write) {
     const auto addr_register{GetRegister(instr.gmem.gpr)};
     const auto immediate_offset{static_cast<u32>(instr.gmem.offset)};
 
     const auto [base_address, index, offset] =
         TrackCbuf(addr_register, global_code, static_cast<s64>(global_code.size()));
-    ASSERT(base_address != nullptr);
+    ASSERT_OR_EXECUTE_MSG(base_address,
```
lioncash: This should probably still make the condition explicit with `base_address != nullptr`, otherwise this looks kind of odd.

Issue #2999, opened 2019-10-18
Several games softlock at MountSaveData if its save folder does not exist


Morph1984:
Several games, such as Fire Emblem Warriors, BOTW and potentially many others softlock at MountSaveData when their corresponding save folder does not exist.

Example:

Fire Emblem Warriors (Title ID: 0100F15003E64000)

If this folder doesn't exist, the game will not launch
`yuzu\nand\user\save\0000000000000000\...\0100F15003E64000`

The current workaround is as follows:

Right click on the affected game, click "Open Save Data Location". This will generate the folder with the appropriate title id.

This issue is related to #2963

Update:

Both #2963 and this issue are caused by #2430

fluffapparition:
Generate the save directory automatically on boot.

Hexagon12:
The issue no longer appears on the latest mainline builds. Fixed by #3007.

Issue #2963, opened 2019-10-09
Several games do not boot without the following directories


Morph1984:
Directory structures

```
yuzu
└───nand
    └───user
        └───save
            └───cache (does not exist)
                └───0000000000000000 (does not exist)
```

```
yuzu
└───nand
    └───temp (does not exist)
        └───0000000000000000 (does not exist)
            └───00000000000000000000000000000000 (does not exist)
                └───0000000000000000 (does not exist)
```

The directories above are not generated as of the latest builds (canary 2649 and later)

Currently, the known games that need the cache directory are the following:
- Super Mario Maker 2

Currently, the known games that need the temp directory are the following:
- Bloons TD 5
- NES/SNES Online

For the temp directory, **you have to re-create it on every launch** as yuzu deletes this folder on launch.

As a temporary workaround you have to create the folders
"yuzu\nand\user\save\cache\0000000000000000" and
"yuzu\nand\temp\0000000000000000\00000000000000000000000000000000\0000000000000000"

or, you can download this zip and place the folders to match the above directories 
[nand.zip](https://github.com/yuzu-emu/yuzu/files/3746048/nand.zip)

For windows users (who do not have yuzu installed in portable mode) 
run this command in cmd:

```batch
if exist "%APPDATA%\yuzu" (
    if not exist "%APPDATA%\yuzu\nand\user\save\cache\0000000000000000" (
        mkdir "%APPDATA%\yuzu\nand\user\save\cache\0000000000000000"
    )
)
```

The cause is traced to the PR #2430

windyboy1704:
This issue started since the September Patreon, I cannot get SMM2 boot. Thanks for your help

Hexagon12:
The issue no longer appears on the latest mainline builds. Fixed by #3007.

PR #3005, opened 2019-10-20
filesystem: Re-introduces a workaround to create save data if it does not exist

Morph1984:
**Do not merge**

Reverts a change (highlighted in bold below) introduced in #2430

> Implementation of creation of save data, and **the removal of a hack where save data would be auto created upon open.** Applications must now use the proper channels (EnsureSaveData, CreateSaveData) to create it. The current impl in this PR for fsp-srv save data creation is quite sparse and will be improved in a separate PR.

Some games such as Breath of the Wild and Fire Emblem Warriors do not use proper channels (EnsureSaveData, CreateSaveData) to create save data. ref #2999 

Furthermore, other relevant directories were not created as a result (cache and temp directories for games like Super Mario Maker 2 and Bloons TD 5 respectively) ref #2963 

We need to investigate what hardware does when save data does not exist because some games do not ensure they exist or are created on boot,

@DarkLordZach This has to be investigated on hardware.


DarkLordZach:
Superseded by #3007.

PR #3032, opened 2019-10-27
shader/control_flow: Abstract repeated code chunks in BRX tracking

ReinUsesLisp:
Remove copied and pasted for cycles into a common templated function and minor style changes.

PR #3034, opened 2019-10-27
maxwell_3d: Silence implicit conversion warnings

ReinUsesLisp:
With #3032 and #3033 it becomes trivial to start treating implicit conversions as errors.

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -98,11 +98,10 @@ void Maxwell3D::InitializeRegisterDefaults() {
     mme_inline[MAXWELL3D_REG_INDEX(index_array.count)] = true;
 }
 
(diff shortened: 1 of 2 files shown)


src/video_core/engines/maxwell_3d.cpp
```diff
@@ -98,11 +98,10 @@ void Maxwell3D::InitializeRegisterDefaults() {
     mme_inline[MAXWELL3D_REG_INDEX(index_array.count)] = true;
 }
 
-#define DIRTY_REGS_POS(field_name) (offsetof(Maxwell3D::DirtyRegs, field_name))
+#define DIRTY_REGS_POS(field_name) static_cast<u8>(offsetof(Maxwell3D::DirtyRegs, field_name))
```
lioncash: There should probably be a static assert added to the header file ensuring the size of DirtyRegs stays within the range of a u8, otherwise this can technically silently do the wrong thing.

PR #3036, opened 2019-10-27
AM/ISelfController: Implement SetOperationModeChangedNotification and Stub SetAlbumImageTakenNotificationEnabled

### LINKED ISSUE #2838: Missing HLE service 'SetAlbumImageTakenNotificationEnabled': port='ISelfController'
## Log
[log.zip](https://github.com/yuzu-emu/yuzu/files/3576956/log.zip)

Morph1984:
Implements SetOperationModeChangedNotification and Stubs SetAlbumImageTakenNotificationEnabled

Ref #2838

These are typically set to true on application boot, and determines whether an AppletMessage gets sent when certain actions are performed such as changing between docked and undocked, capturing a screenshot, etc.

Also fixes some spelling errors: "Messsage" and "Recieve"


src/core/hle/service/am/am.h
```diff
@@ -41,6 +41,13 @@ enum SystemLanguage {
     TraditionalChinese = 16,
 };
 
+// Notification flags
+bool operation_mode_changed_notification = false;
```
lioncash: Yeah, no. We aren't about to introduce global state into the core library more than there already is. Especially when there's virtually none of it within the HLE code. Please find another way of performing this behavior.

src/yuzu/main.cpp
```diff
@@ -1904,6 +1904,18 @@ void GMainWindow::OnCaptureScreenshot() {
         }
     }
     OnStartGame();
+    auto& sm{Core::System::GetInstance().ServiceManager()};
```
lioncash: Taking a screenshot of the rendered display via the UI is not the same as taking a screenshot via in-game functions (i.e. this is not the same thing as actually pressing the capture button on the console itself). It's also very questionable to equate the two together without any way of decoupling the two (not all people taking screenshots may want it to actually be reflected as a console screenshot operation; they may just want to capture the screen and nothing more).

Morph1984: Oh, looks like that is the case, I assumed that the screenshot capture in the UI equated to a short press on the capture button but it is a separate functionality of the emulator, Also, it seems that the capture button does not have any functionality in yuzu yet.

Morph1984: Removed and stubbed for now, since screenshot is not implemented

Morph1984: Removed and stubbed for now since screenshot is not implemented

Morph1984:
@lioncash Are there any more concerns that need addressing?

Issue #3031, opened 2019-10-27
Dragon quest builder 2 make yuzu crash since yuzu 15


momala454:
yuzu simply close itself a few seconds after the loading text with a book is displayed on the bottom right of the screen.
No popup or error, but an error is generated on windows event manager

Détecteur d'erreurs 2161512852989644793, type 4
Nom d’événement : APPCRASH
Réponse : Non disponible
ID de CAB : 0

Signature du problème : 
P1 : yuzu.exe
P2 : 0.0.0.0
P3 : 5da4d021
P4 : yuzu.exe
P5 : 0.0.0.0
P6 : 5da4d021
P7 : c0000005
P8 : 00000000002450f4
P9 : 
P10 : 

windows 10 1903
nvidia 1070 driver 440.97, was also crashing with the driver right before this one.
cpu intel 6700k

crash with and without async, i have pro controller configured

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/3775397/yuzu_log.txt)

<meta name="titleids"  content="010042000a986000">

Hexagon12:
Can you get a call stack from a manual build with Visual Studio? The log doesn't seem very useful.

momala454:
sorry i don't have visual studio, why can't you test yourself ?

fluffapparition:
@momala454 It helps to keep testing in the same environment that the problem occured in

momala454:
did any of you tried the game to see if it crashes for you too ?

momala454:
>	yuzu.exe!Core::Crypto::EncryptionLayer::GetSize() Ligne 16	C++
 	yuzu.exe!Service::FileSystem::IFile::GetSize(Kernel::HLERequestContext & ctx) Ligne 240	C++
 	[Cadre en ligne] yuzu.exe!Service::ServiceFrameworkBase::InvokeRequest(Kernel::HLERequestContext &) Ligne 165	C++
 	yuzu.exe!Service::ServiceFrameworkBase::HandleSyncRequest(Kernel::HLERequestContext & context) Ligne 182	C++
 	yuzu.exe!Kernel::ServerSession::HandleDomainSyncRequest(Kernel::HLERequestContext & context) Ligne 127	C++
 	yuzu.exe!Kernel::ServerSession::HandleSyncRequest(boost::intrusive_ptr<Kernel::Thread> thread) Ligne 140	C++
[...]

PR #3043, opened 2019-10-29
Fix crash when file does not exist

momala454:


ogniK5377:
PRs need to be based off master on yuzu-emu, not yuzu-mainline

momala454:
it i

> 
> 
> PRs need to be based off master on yuzu-emu, not yuzu-mainline

it is based on yuzu emu : https://github.com/momala454/yuzu

fluffapparition:
Related to #3031

PR #3046, opened 2019-10-30
gl_state: Miscellaneous clean up

ReinUsesLisp:
Description in individual commits.


src/video_core/renderer_opengl/gl_state.cpp
```diff
@@ -69,147 +70,29 @@ void Enable(GLenum cap, GLuint index, bool enable) {
...
-        item.height = 0;
-        item.depth_range_near = 0.0f;
-        item.depth_range_far = 1.0f;
-        item.scissor.enabled = false;
-        item.scissor.x = 0;
-        item.scissor.y = 0;
-        item.scissor.width = 0;
-        item.scissor.height = 0;
-    }
+    std::fill(std::begin(viewports), std::end(viewports), Viewport{});
```
lioncash:
```suggestion
    viewports.fill(Viewport{});
```
should be possible, since `viewport` is a `std::array`.

As an aside, the inclusion of `<algorithm>` should be left in, despite std::fill not being used after this, since `std::max` is used in another function, so it still resolves an indirect include.

PR #3047, opened 2019-10-30
gl_rasterizer: Emulate viewport flipping with ARB_clip_control

ReinUsesLisp:
Emulates negative y viewports with [ARB_clip_control](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_clip_control.txt). This allows us to more easily emulate pipelines with tessellation and/or geometry shader stages while also avoiding corrupt feedbacks on games with transform feedbacks and negative viewports (since gl_Position.y was being modified).

While we are at it, use `std::string_view` in `yuzu-cmd` for collecting unsupported extension strings.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -257,10 +257,8 @@ void RasterizerOpenGL::SetupShaders(GLenum primitive_mode) {
             continue;
         }
 
-        const std::size_t stage{index == 0 ? 0 : index - 1}; // Stage indices are 0 - 5
-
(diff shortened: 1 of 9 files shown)


Hexagon12:
Untagged from mainline due to major graphical issues in every single game.

Nearly all shaders fail to build for most games, with this as the root cause:
`0(166) : error C1503: undefined variable "y_negate"`

ReinUsesLisp:
I have to investigate what causes this regression. It seems to be related to front facing being flipped when it shouldn't.

ReinUsesLisp:
Addressed. The cause was front facing being hacked to work with shader-based negative viewports.

PR #3065, opened 2019-11-03
common_func: Use std::array for INSERT_PADDING_* macros.

bunnei:
- Zero initialization here is useful for determinism.

### DIFF
--- a/src/common/common_funcs.h
+++ b/src/common/common_funcs.h
@@ -1,10 +1,11 @@
-// Copyright 2013 Dolphin Emulator Project / 2014 Citra Emulator Project
+// Copyright 2019 yuzu emulator team
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
 #pragma once
 
 #include <algorithm>
+#include <array>
 #include <string>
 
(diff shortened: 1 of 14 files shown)


src/common/common_funcs.h
```diff
@@ -19,8 +20,8 @@
 // helper macro to properly align structure members.
 // Calling INSERT_PADDING_BYTES will add a new member variable with a name like "pad121",
 // depending on the current source line to make sure variable names are unique.
-#define INSERT_PADDING_BYTES(num_bytes) u8 CONCAT2(pad, __LINE__)[(num_bytes)]
-#define INSERT_PADDING_WORDS(num_words) u32 CONCAT2(pad, __LINE__)[(num_words)]
+#define INSERT_PADDING_BYTES(num_bytes) std::array<u8, num_bytes> CONCAT2(pad, __LINE__)
```
lioncash:
```suggestion
#define INSERT_PADDING_BYTES(num_bytes) std::array<u8, num_bytes> CONCAT2(pad, __LINE__){}
```
Is necessary for zero-initialization to occur. Ditto for the expression below this one.

PR #3067, opened 2019-11-03
Initialize various struct members

ogniK5377:
This PR doesn't change anything in file_sys or video core. Most of core is complete, however, there might be a few things missing. This is done to make emulation more deterministic and not rely on possible uninitialized data.

For future PR, `INSERT_PADDING_*` should auto initialize and a variant should be added which doesn't auto initialize the value to be used with unions.


lioncash:
A few of these are better off being zeroed out at the point of use rather than being zeroed out automatically, particularly structs that are often memcpy'd into. This can generate quite a lot of redundant memset operations where they aren't needed. e.g.

```cpp
SomeStruct name; // Unnecessarily zeroed out.
std::memcpy(&name, data, sizeof(name));
```

src/core/hle/service/aoc/aoc_u.cpp
```diff
@@ -75,7 +75,7 @@ AOC_U::~AOC_U() = default;
 
 void AOC_U::CountAddOnContent(Kernel::HLERequestContext& ctx) {
     struct Parameters {
-        u64 process_id;
```
lioncash: This is unnecessary, it's only ever popped out of the request parser.

src/core/hle/service/nvdrv/nvdrv.h
```diff
@@ -30,7 +30,7 @@ struct EventInterface {
     // Mask representing currently busy events
     u64 events_mask{};
     // Each kernel event associated to an NV event
-    std::array<Kernel::EventPair, MaxNvEvents> events;
+    std::array<Kernel::EventPair, MaxNvEvents> events{};
```
lioncash: This is unnecessary. Kernel::EventPair contains entities that have a meaningful constructor

PR #3073, opened 2019-11-04
ci: Rename build folder only on non-mainline builds

DarkLordZach:
Adds a build name suffix to the archive folder name when the build name is *not* `mainline`.

### DIFF
--- a/.ci/scripts/linux/upload.sh
+++ b/.ci/scripts/linux/upload.sh
@@ -6,9 +6,15 @@ REV_NAME="yuzu-linux-${GITDATE}-${GITREV}"
 ARCHIVE_NAME="${REV_NAME}.tar.xz"
 COMPRESSION_FLAGS="-cJvf"
 
-mkdir "$REV_NAME"
+if [ "${RELEASE_NAME}" = "mainline" ]; then
+    DIR_NAME="${REV_NAME}_${RELEASE_NAME}"
+else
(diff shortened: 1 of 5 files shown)


.ci/scripts/windows/upload.sh
```diff
@@ -6,8 +6,14 @@ REV_NAME="yuzu-windows-mingw-${GITDATE}-${GITREV}"
 ARCHIVE_NAME="${REV_NAME}.tar.gz"
 COMPRESSION_FLAGS="-czvf"
 
-mkdir "$REV_NAME"
+if [ "${RELEASE_NAME}" == "mainline" ]; then
```
jroweboy: super minor nitpick, but in bash its just one equals (double equals works but isn't standard)

```bash
if [ "${RELEASE_NAME}" = "mainline" ]; then
```

.ci/scripts/linux/upload.sh
```diff
@@ -6,9 +6,15 @@ REV_NAME="yuzu-linux-${GITDATE}-${GITREV}"
 ARCHIVE_NAME="${REV_NAME}.tar.xz"
 COMPRESSION_FLAGS="-cJvf"
 
-mkdir "$REV_NAME"
+if [ "${RELEASE_NAME}" == "mainline" ]; then
```
jroweboy: ditto

PR #3084, opened 2019-11-08
video_core: Treat implicit conversions as errors

ReinUsesLisp:
This blocks future commits from introducing implicit casts in `video_core`'s code base.


bunnei:
Build errors are real

ReinUsesLisp:
On it, I was testing with clang instead of gcc.

ReinUsesLisp:
Fixed.

src/video_core/CMakeLists.txt
```diff
@@ -180,3 +180,9 @@ target_link_libraries(video_core PRIVATE glad)
 if (ENABLE_VULKAN)
     target_link_libraries(video_core PRIVATE sirit)
 endif()
+
+if (MSVC)
+    target_compile_options(video_core PRIVATE /we4267)
+else()
+    target_compile_options(video_core PRIVATE -Werror=conversion -Wno-sign-conversion)
```
lioncash: Looking over this again, I don't feel that hiding all warnings related to sign conversions is a particularly wise thing to do, given this can prevent actual logic bugs from being reported by the compiler.

ReinUsesLisp: Should I remove their errors but enable them as warnings? That's there because `-Werror=conversion` implies `-Werror=sign-conversion`.

lioncash: This would likely be a better approach, yeah.

PR #3085, opened 2019-11-09
yuzu: configure_web: Use Base64 encoded token

bunnei:
yuzu: configure_web: Use Base64 encoded token for simplifying user experience.

This is a change we are making for both citra and yuzu across all clients. It's purely a user-facing/UI change, to make the experience of adding your credential simpler - it's just a simply copy+paste of one string.

See https://github.com/citra-emu/citra/pull/4993 for more details.

### DIFF
--- a/src/yuzu/configuration/configure_web.ui
+++ b/src/yuzu/configuration/configure_web.ui
@@ -55,11 +55,7 @@
            </widget>
           </item>
           <item row="0" column="1" colspan="3">
-           <widget class="QLineEdit" name="edit_username">
-            <property name="maxLength">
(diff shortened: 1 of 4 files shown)


src/yuzu/configuration/configure_web.ui
```diff
@@ -30,7 +30,7 @@
          </widget>
         </item>
         <item>
-         <layout class="QGridLayout" name="gridLayoutYuzuUsername">
+         <layout class="QGridLayout" name="gridLayoutCitraUsername">
```
FearlessTobi: I doubt this was intended?

PR #3086, opened 2019-11-09
 texture_cache: Use a flat table instead of switch for texture format lookups

ReinUsesLisp:
Use a large flat array to look up texture formats. This allows us to properly implement formats with different component types and it should be faster.


src/video_core/texture_cache/format_lookup_table.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+namespace VideoCommon {
+
+using Tegra::Texture::ComponentType;
+using Tegra::Texture::TextureFormat;
+using VideoCore::Surface::PixelFormat;
+
+namespace {
+
+static constexpr auto SNORM = ComponentType::SNORM;
```
lioncash: Things within an anonymous namespace don't need to be static (they already have internal linkage).

src/video_core/texture_cache/format_lookup_table.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    u32 texture_format : 8;
+    u32 pixel_format : 8;
+    u32 red_component : 3;
+    u32 green_component : 3;
+    u32 blue_component : 3;
+    u32 alpha_component : 3;
+    u32 is_srgb : 1;
+};
+static constexpr Table DefinitionTable[] = {
```
lioncash: `std::array`

src/video_core/texture_cache/format_lookup_table.cpp
```diff
@@ -0,0 +1,213 @@
...
+                    ComponentType green_component, ComponentType blue_component,
+                    ComponentType alpha_component, PixelFormat pixel_format)
+        : texture_format{static_cast<u32>(texture_format)},
+          pixel_format{static_cast<u32>(pixel_format)}, red_component{static_cast<u32>(
+                                                            red_component)},
+          green_component{static_cast<u32>(green_component)}, blue_component{static_cast<u32>(
+                                                                  blue_component)},
+          alpha_component{static_cast<u32>(alpha_component)}, is_srgb{is_srgb ? 1U : 0U} {}
+
+    u32 texture_format : 8;
```
lioncash: You'd likely just want to store the elements as is without a bitfield, otherwise any access to the members requires the compiler to generate a shift and mask operation for every individual access, which kind of goes counter-intuitive to using a lookup table to avoid operations.

src/video_core/texture_cache/format_lookup_table.h
```diff
@@ -0,0 +1,51 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+#include <numeric>
```
lioncash: This should likely be `<limits>`

src/video_core/texture_cache/format_lookup_table.cpp
ReinUsesLisp: This one is only used to construct the large lookup table. But considering the number of elements this shouldn't affect the final binary size.

src/video_core/texture_cache/format_lookup_table.cpp
ReinUsesLisp: Can `std::array` deduce its size? Otherwise I would have to:
- Specify the type on each element and use `std::array`'s template deduction.
- Specify the final size in `std::array`.

ReinUsesLisp:
Rebased with #3110 merged. `R16_G16_B16_A16` `UINT`^4 is included.

PR #3089, opened 2019-11-11
Implement stub for IApplicationFunctions::QueryApplicationPlayStatisticsByUid

SciresM:
Needed to boot some newer games (SDK >= ~7.x).


FernandoS27:
hmn, is this stubbed or it's just a boolean service?

SciresM:
@FernandoS27 It's stubbed. The function normally writes entries to an output buffer, and returns an s32 "NumEntriesWritten".

This stub makes it return zero entries.

SciresM:
For even more clarity: This stub is equivalent to the service telling the application that there are no historical play statistics available (i.e., this is the first boot of the application).

lioncash:
The commit message should probably include the tidbit about it being necessary to boot games on SDK >= ~7.x, as well as the clarity comment stating what the stub is doing behaviorally.

lioncash:
Similarly, `QueryApplicationPlayStatistics` can be stubbed along with this for implementation consistency, given it essentially performs the same behavior, with minor differences.

PR #3092, opened 2019-11-12
key_manager: Make use of IOFile in WriteKeyToFile()

lioncash:
This properly handles unicode-based paths on Windows, while opening a raw stream doesn't out-of-the-box.

Prevents file creation from potentially failing on Windows PCs that make use of unicode characters in their save paths (e.g. writing to a user's AppData folder, where the user has a name with non-ASCII characters).


src/core/crypto/key_manager.cpp
```diff
@@ -668,23 +668,27 @@ void KeyManager::WriteKeyToFile(KeyCategory category, std::string_view keyname,
...
-    else if (category == KeyCategory::Console)
+    } else if (category == KeyCategory::Console) {
         filename = "console.keys_autogenerated";
-    const auto add_info_text = !FileUtil::Exists(yuzu_keys_dir + DIR_SEP + filename);
-    FileUtil::CreateFullPath(yuzu_keys_dir + DIR_SEP + filename);
-    std::ofstream file(yuzu_keys_dir + DIR_SEP + filename, std::ios::app);
-    if (!file.is_open())
+    }
+
+    const std::string path = yuzu_keys_dir + DIR_SEP + filename;
```
DarkLordZach:
```suggestion
    const auto path = yuzu_keys_dir + DIR_SEP + filename;
```

PR #3096, opened 2019-11-12
Move source in windows uploads into its own internal zip

jroweboy:
This should prevent path length issues when extracting the build from the installer

This might also cause a new "source" artifact to be uploaded, but that shouldn't matter for github releases at least. We should check that it doesn't break azure releases later.


jroweboy:
Note this is wholly untested because I don't have an azure setup configured for my fork, but due to the fact that live is broken right now, it wouldn't make it worse if we merged it. At worst we just revert this and the #3090 as well and fix it later. At best its all fixed and we move on :)

PR #3098, opened 2019-11-13
gl_shader_cache: Miscellaneous changes to shaders

ReinUsesLisp:
This diff includes various changes that invalidate the shader cache. I opted to do all of this at once to avoid invalidating several times.

The description of each case is in its individual commit. As highlights this PR implements compute shaders for Intel free and proprietary and AMD proprietary drivers (nouveau and radeonsi were already covered) and it fixes build issues due to shared memory access in compute shaders.


bunnei:
@ReinUsesLisp needs a rebase

PR #3114, opened 2019-11-15
Kernel: Correct behavior of Condition Variables to be more similar to real hardware.

FernandoS27:
This commit ensures cond var threads act exactly as they do in the real console. The original implementation uses an RBTree and the behavior of cond var threads is that at the same priority level they act like a FIFO.

### DIFF
--- a/src/core/hle/kernel/process.cpp
+++ b/src/core/hle/kernel/process.cpp
@@ -142,6 +142,52 @@ u64 Process::GetTotalPhysicalMemoryUsedWithoutSystemResource() const {
     return GetTotalPhysicalMemoryUsed() - GetSystemResourceUsage();
 }
 
+void Process::InsertConditionVariableThread(SharedPtr<Thread> thread) {
+    auto it = cond_var_threads.begin();
+    while (it != cond_var_threads.end()) {
+        const SharedPtr<Thread> current_thread = *it;
+        if (current_thread->GetCondVarWaitAddress() < thread->GetCondVarWaitAddress()) {
+            if (current_thread->GetCondVarWaitAddress() == thread->GetCondVarWaitAddress()) {
+                if (current_thread->GetPriority() > thread->GetPriority()) {
+                    cond_var_threads.insert(it, thread);
+                    return;
+                }
+            } else {
+                cond_var_threads.insert(it, thread);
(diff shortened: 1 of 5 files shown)


src/core/hle/kernel/process.cpp
```diff
@@ -142,6 +142,52 @@ u64 Process::GetTotalPhysicalMemoryUsedWithoutSystemResource() const {
...
+                if (current_thread->GetPriority() > thread->GetPriority()) {
+                    cond_var_threads.insert(it, thread);
+                    return;
+                }
+            } else {
+                cond_var_threads.insert(it, thread);
+                return;
+            }
+        }
+        it++;
```
lioncash:
```suggestion
        ++it;
```

src/core/hle/kernel/process.cpp
```diff
@@ -142,6 +142,52 @@ u64 Process::GetTotalPhysicalMemoryUsedWithoutSystemResource() const {
...
+
+void Process::RemoveConditionVariableThread(SharedPtr<Thread> thread) {
+    auto it = cond_var_threads.begin();
+    while (it != cond_var_threads.end()) {
+        const SharedPtr<Thread> current_thread = *it;
+        if (current_thread.get() == thread.get()) {
+            cond_var_threads.erase(it);
+            return;
+        }
+        it++;
```
lioncash:
```suggestion
        ++it;
```

src/core/hle/kernel/process.cpp
```diff
@@ -142,6 +142,52 @@ u64 Process::GetTotalPhysicalMemoryUsedWithoutSystemResource() const {
...
+
+std::vector<SharedPtr<Thread>> Process::GetConditionVariableThreads(const VAddr cond_var_addr) {
+    std::vector<SharedPtr<Thread>> result{};
+    auto it = cond_var_threads.begin();
+    while (it != cond_var_threads.end()) {
+        SharedPtr<Thread> current_thread = *it;
+        if (current_thread->GetCondVarWaitAddress() == cond_var_addr) {
+            result.push_back(current_thread);
+        }
+        it++;
```
lioncash:
```suggestion
        ++it;
```

PR #3126, opened 2019-11-16
Revert "common/bit_field: Silence sign-conversion warnings"

ReinUsesLisp:
#3106 caused expressions like `instr.gpr8.Value() != Register::ZeroIndex` to always evaluate as true even if `gpr8` is `RZ`. We use this to determine if we should write or not to the output register, resulting in all register stores being no-ops (nothing is rendered).

Until it's properly fixed I'm tagging this for mainline to avoid regressions.


lioncash:
I would say merge this in the meantime, since I'm not around to look at this again until Monday. It's better to have a functional master than not.

PR #3130, opened 2019-11-16
Kernel: Correct Cancel Synchronization.

FernandoS27:
This commit corrects the behavior of cancel synchronization when the thread is running/ready and ensures the next wait is cancelled as it's suppose to.

### DIFF
--- a/src/core/hle/kernel/thread.h
+++ b/src/core/hle/kernel/thread.h
@@ -440,6 +440,14 @@ public:
         is_running = value;
     }
 
+    bool IsSyncCancelled() const {
+        return is_sync_cancelled;
+    }
+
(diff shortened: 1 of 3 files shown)


src/core/hle/kernel/thread.h
```diff
@@ -524,6 +532,7 @@ class Thread final : public WaitObject {
 
     u32 scheduling_state = 0;
     bool is_running = false;
+    bool sync_cancelled = false;
```
lioncash:
```suggestion
    bool is_sync_cancelled = false;
```

for consistency

theboy181:
Can you explain using a game as an example what was fixed ?

PR #3154, opened 2019-11-23
kernel: Replace usage of boost::intrusive_ptr with std::shared_ptr for kernel objects.

bunnei:
We were using SharedPtr (boost::instrusive_ptr) from legacy Citra days -- without obvious advantage. We also had mixed usage of SharedPtr and std::shared_ptr -- which is messy and may even lead to ownership/ref counting bugs. Lastly, there are probably some good places where we want weak pointers, which std::shared_ptr supports.

See https://github.com/citra-emu/citra/pull/4710 for more details.

Should largely be a no-op as far as functionality... I suggest a quick merge so we can rebase other things.

### DIFF
--- a/src/core/hle/kernel/client_session.cpp
+++ b/src/core/hle/kernel/client_session.cpp
@@ -19,22 +19,20 @@ ClientSession::~ClientSession() {
 
     // A local reference to the ServerSession is necessary to guarantee it
     // will be kept alive until after ClientDisconnected() returns.
-    SharedPtr<ServerSession> server = parent->server;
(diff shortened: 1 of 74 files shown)


src/core/hle/kernel/client_session.cpp
```diff
@@ -19,22 +19,20 @@ ClientSession::~ClientSession() {
 
     // A local reference to the ServerSession is necessary to guarantee it
```
lioncash: This comment doesn't apply anymore, since the local was removed.

PR #3159, opened 2019-11-25
core_timing: Use better reference tracking for EventType.

bunnei:
- Moves ownership of the event to the caller, ensuring we don't fire events for destroyed objects.
- Removes need for unique names - we won't be using this for save states anyways. Having non-unique names will be useful for when we use core timing to schedule service calls.


lioncash:
Second commit name should probably clarify that it's updating the core_timing tests, but other than that it looks fine to me.

bunnei:
Second commit was a fixup, I just did it via the GitHub. Will squash on merge.

PR #3163, opened 2019-11-26
Revert "gpu_thread: Don't spin wait if there are no GPU commands."

bunnei:
Reverts yuzu-emu/yuzu#3155

PR #3166, opened 2019-11-26
GPU_Async: Make the GPU thread use the Event pattern

FernandoS27:
This commit makes the GPU use the Event pattern for syncing instead of spin-looping while waiting for commands. This should make better use of the GPU's thread time.

### DIFF
--- a/src/video_core/gpu_thread.cpp
+++ b/src/video_core/gpu_thread.cpp
@@ -19,8 +19,7 @@ static void RunThread(VideoCore::RendererBase& renderer, Tegra::DmaPusher& dma_p
     MicroProfileOnThreadCreate("GpuThread");
 
     // Wait for first GPU command before acquiring the window context
-    while (state.queue.Empty())
-        ;
+    state.sync_event.Wait();
 
     // If emulation was stopped during disk shader loading, abort before trying to acquire context
     if (!state.is_running) {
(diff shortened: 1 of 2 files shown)


degasus:
How does this differ from https://github.com/yuzu-emu/yuzu/pull/3163 ?

FernandoS27:
> How does this differ from #3163 ?

It uses an Event Pattern to put the thread on wait and signal it on commands push. This avoids spin-looping.

degasus:
But how does it differ from the Event used within the ThreadsafeQueue?

degasus:
I don't get the point of adding another cond_var. This PR replaces https://github.com/yuzu-emu/yuzu/blob/master/src/common/threadsafe_queue.h#L82 with https://github.com/yuzu-emu/yuzu/blob/master/src/common/thread.h#L27

If this affects the behavior, we should fix the common code istead :(

bunnei:
@degasus finally got around to testing #3173 on its own -- unfortunately, asynch GPU is still broken, without this change.

degasus:
@bunnei I think I was wrong with my proposed fix. Just switching the order was not enough as @CookiePLMonster has shown.

bunnei:
@FernandoS27 should this be closed, or do you plan to do anything else here?

PR #3169, opened 2019-11-27
core/memory: Deglobalize memory management code

lioncash:
Eliminates global state in the main memory management code and places all the necessary behavior within its own class. Now, the memory subsystem doesn't directly rely upon the global system instance. Subsequent follow-up PR(s) will tidy this up further, as I feel this is a good starting point to introduce the code to the codebase.

I've separated the migration process into their own commits so that they can be reviewed in a more digestible manner, each containing any rationale if necessary. Most of the diff is moving the functions that already exist within memory.cpp into the Memory class. Contrary to what the diff stats indicate, 1318 lines of code haven't been introduced.

I also took the time with this migration to completely document the memory API along with any notable preconditions/postconditions that may apply.


bunnei:
@lioncash sorry - can you rebase? Thanks !

lioncash:
@bunnei done

PR #3173, opened 2019-11-27
common: SPSCQueue: Notify after incrementing queue size.

bunnei:
Fixes a race condition.

### DIFF
--- a/src/common/threadsafe_queue.h
+++ b/src/common/threadsafe_queue.h
@@ -46,9 +46,16 @@ public:
         ElementPtr* new_ptr = new ElementPtr();
         write_ptr->next.store(new_ptr, std::memory_order_release);
         write_ptr = new_ptr;
-        cv.notify_one();
 
-        ++size;
+        const size_t previous_size{size++};
+
+        // Acquire the mutex and then immediately release it as a fence.
+        // TODO(bunnei): This can be replaced with C++20 waitable atomics when properly supported.
+        // See discussion on https://github.com/yuzu-emu/yuzu/pull/3173 for details.
+        if (previous_size == 0) {
+            std::lock_guard lock{cv_mutex};
+        }
+        cv.notify_one();
     }
 
     void Pop() {


degasus:
LGTM

CookiePLMonster:
Does it make sense to have this variable atomic? With a condition variable and a mutex the usual way is to protect whatever predicate there is for a condvar. Currently `cv_mutex` is used only so condvar can be put to sleep, which essentially makes it work as some sort of an awaitable flag.

`wait` with a predicate is realized in a means of a while loop and `wait`, so if I am right there is still a window of opportunity for race to occur if `size` gets incremented and cv gets notified **after** predicate returns false but **before** condvar gets put to sleep. If this happens, it'll be asleep (despite wakeup requirements having been met) until the next time it is notified.

bunnei:
@CookiePLMonster maybe I am missing something -- but `size` here is already an atomic?

CookiePLMonster:
> @CookiePLMonster maybe I am missing something -- but `size` here is already an atomic?

Yeah, my point here is that **as far as I can see by just reading this code**, having it atomic does not make it fully bulletproof against race conditions.

This is all pure theory, but consider this example:

`cv.wait(lock, [this]() { return !Empty(); });` is defined by the standard as
```
while ( !(!Empty() )
  cv.wait(lock);
```

without any guarantees about atomicity of check + wait.

Now, with this in mind assume such code flow:
1) `size` is 0`.
2) On thread A, `Empty()` executes and returns true. Thread A will enter the loop.
2) Context switches to thread B.
3) Thread B executes
```
++size;
cv.notify_one();
```
so `size` is now `1`. `cv` gets signaled, but nobody is sleeping on it.
4) Context switches to thread A;
5) Thread A executes `cv.wait(lock);`. It starts waiting for `cv` to be signaled.

As a result. thread A is sleeping even though `size` is non zero - there was a race condition between signaling and sleeping!

Now, if `size` was incremented from under the lock, this race condition is avoided because in the example mentioned above, thread B could not execute `++size` and signal `cv` until **after** thread A has entered the sleeping state.

degasus:
What about
```
int old = size++;
if (!old) {
  std::lock_guard(my_mutex);
  cv.notify_one();
}
```
This makes sure that the polling loop won't deadlock while still maintaining the lock-free performance in the common case (not empty). We can do the same on the polling side:
```
if (!size) {
  std::lock_guard(my_mutex);
  while(!size)
    cv.wait();
}
```

CookiePLMonster:
> What about

No, it's still wrong* as you need to sync on the predicate (in this case, `size`) and not signaling. Ignoring lock-freeness of this approach, the intended way for this to go is to protect the size like so:

```
{
  std::lock_guard(my_mutex);
  size++;
}
cv.notify_one();
```

If I'm not mistaken, you could achieve correct functionality without the need for a mutex by using [C++20 waitable atomics](https://en.cppreference.com/w/cpp/atomic). I never used them however, so I cannot guarantee this is **the** solution.

* while it may work (would need to follow the flow closely), it's just not the way to do it.

CookiePLMonster:
OK, so I had a second look at it and acquiring the mutex after changing `size` is actually not wrong. I would solve it by:

1) Using C++20 atomics - [this paper](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0514r3.pdf) shows it solves pretty much this exact use case.
2) If that's not possible, "fence" the mutex like you did, albeit without protecting the `notify` - acquiring and instantly releasing the mutex should be really enough, assuming you cannot race between `Push` and `Pop` (in which case this might need some more thought):

```cpp
const size_t old = size++;
if (old == 0) {
  std::lock_guard lock(my_mutex);
}
cv.notify_one();
```

bunnei:
@CookiePLMonster nice catch -- we can't use waitable atomics yet, as our existing build infrastructure does not support them. However, adding a fence by acquiring the mutex and then immediately releasing it fixes our problem here.

src/common/threadsafe_queue.h
```diff
@@ -46,9 +46,19 @@ class SPSCQueue {
         ElementPtr* new_ptr = new ElementPtr();
         write_ptr->next.store(new_ptr, std::memory_order_release);
         write_ptr = new_ptr;
-        cv.notify_one();
 
+        const size_t previous_size{size};
         ++size;
```
CookiePLMonster: Very minor nitpick, but I think
`const size_t previous_size{size++};`

would look nicer, and probably also more correct when it comes to how the atomic is used - right now you technically could have another operation on `size` between those two lines.

src/common/threadsafe_queue.h
```diff
@@ -46,9 +46,19 @@ class SPSCQueue {
...
+
+        // Acquire the mutex and then immediately release it as a fence.
+        // TODO(bunnei): This can be replaced with C++20 waitable atomics when properly supported.
+        // See discussion on https://github.com/yuzu-emu/yuzu/pull/3173 for details.
+        if (previous_size == 0) {
+            std::lock_guard lock{cv_mutex};
+            cv.notify_one();
+        } else {
+            cv.notify_one();
+        }
```
CookiePLMonster: You can (and should) move `cv.notify_one();` outside this if clause - otherwise the thread which was awoken might lose some time, because as soon as it's awoken it'll block on a not yet unlocked `cv_mutex`.

See https://en.cppreference.com/w/cpp/thread/condition_variable, where they also recommend not to keep the lock held for notification.

src/common/threadsafe_queue.h
```diff
@@ -46,9 +46,19 @@ class SPSCQueue {
         ElementPtr* new_ptr = new ElementPtr();
         write_ptr->next.store(new_ptr, std::memory_order_release);
         write_ptr = new_ptr;
-        cv.notify_one();
 
+        const size_t previous_size{size};
```
degasus: I think this might be another race condition. previous_size should be fetched in an atomic exchange way. Else the GPU thread might decrease it between this and the ++size line. So I'd use `const size_t previous_size = size++;`

src/common/threadsafe_queue.h
```diff
@@ -46,9 +46,19 @@ class SPSCQueue {
...
         ++size;
+
+        // Acquire the mutex and then immediately release it as a fence.
+        // TODO(bunnei): This can be replaced with C++20 waitable atomics when properly supported.
+        // See discussion on https://github.com/yuzu-emu/yuzu/pull/3173 for details.
+        if (previous_size == 0) {
+            std::lock_guard lock{cv_mutex};
+            cv.notify_one();
+        } else {
+            cv.notify_one();
```
degasus: The else part is not required. If there was a task in the queue, the consumer won't sleep.

src/common/threadsafe_queue.h
degasus: heh, @CookiePLMonster was faster :D

src/common/threadsafe_queue.h
CookiePLMonster: I'd still signal it, it will not cause issues but technically be more correct. That + the note I left about scope of the lock.

PR #3177, opened 2019-11-28
kernel: Implement a more accurate IPC dispatch.

bunnei:
* Properly puts threads to sleep on SendSyncRequest, and wakes them in the future.
* Uses a threadsafe queue for IPC requests, which will be helpful for multicore impl.
* More accurate representation of Session, ClientSession, and ServerSession kernel objects.
* ClientPort() should only be used for named ports.

### DIFF
--- a/src/core/hle/kernel/hle_ipc.h
+++ b/src/core/hle/kernel/hle_ipc.h
@@ -264,6 +264,18 @@ public:
 
     std::string Description() const;
 
+    Thread& GetThread() {
+        return *thread;
+    }
+
+    const Thread& GetThread() const {
+        return *thread;
+    }
+
+    bool IsThreadWaiting() const {
+        return is_thread_waiting;
+    }
+
 private:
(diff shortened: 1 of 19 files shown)


src/core/hle/kernel/hle_ipc.h
```diff
@@ -264,6 +264,14 @@ class HLERequestContext {
 
     std::string Description() const;
 
+    Thread& GetThread() {
```
lioncash: This should likely have a const getter as well

```cpp
const Thread& GetThread() const {
    return *thread;
}
```

To pair with this non-const getter.

src/core/hle/service/sm/sm.h
```diff
@@ -16,6 +16,7 @@
 #include "core/hle/service/service.h"
 
 namespace Kernel {
+class KernelCore;
```
lioncash: This should be situated between the ClientSession and ServerPort forward declaration.

PR #3178, opened 2019-11-28
patch_manager: Adds check for disabled cheats to prevent them from being enabled

### LINKED ISSUE #3176: Unchecking cheats in the properties menu, not working.
When removing a cheat from the properties menu in yuzu the cheat will still get executed. 

image

Morph1984:
A check for disabled cheats was missing from patch_manager, this PR adds the check.

Fixes #3176

### DIFF
--- a/src/core/file_sys/patch_manager.cpp
+++ b/src/core/file_sys/patch_manager.cpp
@@ -76,7 +76,7 @@ VirtualDir PatchManager::PatchExeFS(VirtualDir exefs) const {
 
     const auto& disabled = Settings::values.disabled_addons[title_id];
     const auto update_disabled =
-        std::find(disabled.begin(), disabled.end(), "Update") != disabled.end();
+        std::find(disabled.cbegin(), disabled.cend(), "Update") != disabled.cend();
 
     // Game Updates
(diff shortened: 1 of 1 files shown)


src/core/file_sys/patch_manager.cpp
```diff
@@ -284,12 +284,16 @@ std::vector<Memory::CheatEntry> PatchManager::CreateCheatList(
...
     }
 
+    const auto& disabled = Settings::values.disabled_addons[title_id];
     auto patch_dirs = load_dir->GetSubdirectories();
     std::sort(patch_dirs.begin(), patch_dirs.end(),
               [](const VirtualDir& l, const VirtualDir& r) { return l->GetName() < r->GetName(); });
 
     std::vector<Memory::CheatEntry> out;
     for (const auto& subdir : patch_dirs) {
+        if (std::find(disabled.begin(), disabled.end(), subdir->GetName()) != disabled.end())
```
lioncash:
```suggestion
        if (std::find(disabled.cbegin(), disabled.cend(), subdir->GetName()) != disabled.cend())
```

This should also be braced

PR #3196, opened 2019-12-06
[EA source]: Ignore git-related files in cmake for early access tarballs

jmerdich:
This fixes the builds from early-access source drops on Windows (tested on EA 58). CMake was previously looking for git-related files that were stripped out of the early access builds and failing; check if those exist before reading them.


bunnei:
Thanks !

PR #3201, opened 2019-12-08
kernel/svc: Provide implementations for svcDumpInfo/svcDumpInfoNew

lioncash:
These are fairly trivial to implement, we can just do nothing. This also provides a spot for us to potentially dump out any relevant info in the future (e.g. for debugging purposes with homebrew, etc).


SciresM:
FYI, these names are incorrect.

As of latest firmware, these SVCs are named "SvcKernelDebug" and "SvcChangeKernelTraceState".

lioncash:
Thanks for the info. I've amended the names of the function.

SciresM:
Looks good :)

If relevant, params for svcKernelDebug are a u32 (kernel_debug_type), then three u64 parameters.

PR #3203, opened 2019-12-08
Texture Cache: Add HLE methods for building 3D textures

FernandoS27:
This commit adds a series of HLE methods for handling 3D textures in general. This helps games that generate 3D textures on every frame and may reduce loading times for certain games.

### DIFF
--- a/src/video_core/texture_cache/surface_params.h
+++ b/src/video_core/texture_cache/surface_params.h
@@ -138,13 +138,13 @@ public:
 
     std::size_t GetConvertedMipmapSize(u32 level) const;
 
-    // Get this texture Tegra Block size in guest memory layout
+    /// Get this texture Tegra Block size in guest memory layout
     u32 GetBlockSize() const;
 
-    // Get X, Y sizes of a block
+    /// Get X, Y sizes of a block
     std::pair<u32, u32> GetBlockXY() const;
 
-    // Get the offset in x, y, z coordinates from a memory offset
+    /// Get the offset in x, y, z coordinates from a memory offset
     std::tuple<u32, u32, u32> GetBlockOffsetXYZ(u32 offset) const;
 
     /// Returns the size of a layer in bytes in guest memory.
--- a/src/video_core/texture_cache/surface_params.h
+++ b/src/video_core/texture_cache/surface_params.h
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <utility>
+
 #include "common/alignment.h"
 #include "common/bit_util.h"
 #include "common/cityhash.h"
@@ -135,6 +137,15 @@ public:
     }
 
     std::size_t GetConvertedMipmapSize(u32 level) const;
+
+    // Get this texture Tegra Block size in guest memory layout
+    u32 GetBlockSize() const;
+
+    // Get X, Y sizes of a block
+    std::pair<u32, u32> GetBlockXY() const;
+
+    // Get the offset in x, y, z coordinates from a memory offset
+    std::tuple<u32, u32, u32> GetBlockOffsetXYZ(u32 offset) const;
 
     /// Returns the size of a layer in bytes in guest memory.
     std::size_t GetGuestLayerSize() const {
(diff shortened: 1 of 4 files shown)


src/video_core/texture_cache/texture_cache.h
```diff
@@ -615,6 +615,85 @@ class TextureCache {
...
+            }
+            if (failed) {
+                return std::nullopt;
+            }
+            for (const auto& surface : overlaps) {
+                Unregister(surface);
+            }
+            new_surface->MarkAsModified(modified, Tick());
+            Register(new_surface);
+            return {{new_surface, new_surface->GetMainView()}};
```
ReinUsesLisp: Since this is passed to a std::pair constructor, you should `std::move(new_surface)` to avoid copies.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -687,6 +766,15 @@ class TextureCache {
             }
         }
 
+        // Look if it's a 3D texture
+        if (params.block_depth > 0) {
+            std::optional<std::pair<TSurface, TView>> surface =
```
ReinUsesLisp: `auto` or `std::optional` without templated arguments.
In the context of this function we don't care about the type itself.

src/video_core/texture_cache/surface_params.h
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <utility>
+
```
ReinUsesLisp: Not strictly related to this commit, but `<optional>`, `<tuple>` and `<vector>` should be included.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -615,6 +615,85 @@ class TextureCache {
...
+                    src_params.block_height != params.block_height) {
+                    failed = true;
+                    break;
+                }
+                const u32 offset = static_cast<u32>(surface->GetCacheAddr() - cache_addr);
+                const auto [x, y, z] = params.GetBlockOffsetXYZ(offset);
+                modified |= surface->IsModified();
+                const CopyParams copy_params(0, 0, 0, 0, 0, z, 0, 0, params.width, params.height,
+                                             1);
+                ImageCopy(surface, new_surface, copy_params);
```
ReinUsesLisp: `failed` might be set and then the for break when some copies have been done. Do we want this? An alternative solution would be to copy only after `failed` is known to be false.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -615,6 +615,85 @@ class TextureCache {
...
+                        Unregister(surface);
+                        return InitializeSurface(gpu_addr, params, preserve_contents);
+                    }
+                    return std::nullopt;
+                }
+                if (surface->GetCacheAddr() != cache_addr) {
+                    continue;
+                }
+                const auto struct_result = surface->MatchesStructure(params);
+                if (struct_result == MatchStructureResult::FullMatch) {
```
ReinUsesLisp: `surface->MatchesStructure(params) == MatchStructureResult::FullMatch`.
That way we avoid forced temporary names.

src/video_core/texture_cache/surface_params.h
```diff
@@ -136,6 +138,15 @@ class SurfaceParams {
 
     std::size_t GetConvertedMipmapSize(u32 level) const;
 
+    // Get this texture Tegra Block size in guest memory layout
```
ReinUsesLisp: Doxy starts with `///`

src/video_core/texture_cache/surface_params.h
```diff
@@ -136,6 +138,15 @@ class SurfaceParams {
...
+    u32 GetBlockSize() const;
+
+    // Get X, Y sizes of a block
```
ReinUsesLisp: Ditto

src/video_core/texture_cache/surface_params.h
```diff
@@ -136,6 +138,15 @@ class SurfaceParams {
...
+    std::pair<u32, u32> GetBlockXY() const;
+
+    // Get the offset in x, y, z coordinates from a memory offset
```
ReinUsesLisp: Ditto

src/video_core/texture_cache/surface_params.h
FernandoS27: This guarantees so, some compilers may need it.

src/video_core/texture_cache/texture_cache.h
FernandoS27: It's a bit hard to setup, we do a similar thing on TryReconstructSurface. Since failing is normally very unlikely, it's better to take the hit and fail with some work done than delegating code and making the texture cache more complex.

PR #3227, opened 2019-12-15
delete appveyor config

amilajack:
Doesn't seem like we're using appveyor anymore so it makes sense to delete this


Margen67:
You might want to also delete `.appveyor/UtilityFunctions.ps1`

amilajack:
Done!

bunnei:
Thanks!

PR #3239, opened 2019-12-20
shader/p2r: Implement P2R Pr

ReinUsesLisp:
P2R dumps predicate or condition codes state to a register. This is useful for unit testing.

It's not known to be used in retail games or homebrew.

### DIFF
--- a/src/video_core/shader/decode/register_set_predicate.cpp
+++ b/src/video_core/shader/decode/register_set_predicate.cpp
@@ -13,37 +13,51 @@ namespace VideoCommon::Shader {
 using Tegra::Shader::Instruction;
 using Tegra::Shader::OpCode;
 
+namespace {
+constexpr u64 NUM_PROGRAMMABLE_PREDICATES = 7;
+}
+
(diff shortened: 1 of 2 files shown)


src/video_core/shader/decode/register_set_predicate.cpp
```diff
@@ -13,37 +13,65 @@ namespace VideoCommon::Shader {
...
+
+            const Node code = Operation(OperationCode::LogicalAssign, GetPredicate(pred), value);
+            bb.push_back(Conditional(condition, {code}));
+        }
+        break;
+    }
+    case OpCode::Id::P2R_IMM: {
+        Node value = Immediate(0);
+        for (u64 pred = 0; pred < NUM_PROGRAMMABLE_PREDICATES; ++pred) {
+            Node bit = Operation(OperationCode::Select, GetPredicate(pred), Immediate(1U << pred),
```
FernandoS27: const

ReinUsesLisp: If it's const, it will be copied on `std::move`

PR #3244, opened 2019-12-23
fixed_pipeline_state: Define structure and loaders

ReinUsesLisp:
The intention behind this hashable structure is to describe the state
of fixed function pipeline state that gets compiled to a single graphics
pipeline state object. This is all dynamic state in OpenGL but Vulkan
wants it in an immutable state, even if hardware can edit it freely.

In this commit the structure is defined in an optimized state (it uses
booleans, has paddings and many data entries that can be packed to
single integers). This is intentional as an initial implementation that
is easier to debug, implement and review. It will be optimized in later
stages, or it might change if Vulkan gets more dynamic states.

### DIFF
--- a/src/video_core/renderer_vulkan/fixed_pipeline_state.h
+++ b/src/video_core/renderer_vulkan/fixed_pipeline_state.h
@@ -30,8 +30,13 @@ struct FixedPipelineState {
         u32 stride;
         u32 divisor;
 
-        std::size_t Hash() const;
-        bool operator==(const VertexBinding& rhs) const;
+        std::size_t Hash() const noexcept;
+
+        bool operator==(const VertexBinding& rhs) const noexcept;
+
+        bool operator!=(const VertexBinding& rhs) const noexcept {
+            return !operator==(rhs);
+        }
     };
 
     struct VertexAttribute {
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_vulkan/fixed_pipeline_state.h
```diff
@@ -0,0 +1,231 @@
...
+        constexpr VertexBinding(u32 index, u32 stride, u32 divisor)
+            : index{index}, stride{stride}, divisor{divisor} {}
+        VertexBinding() = default;
+
+        u32 index;
+        u32 stride;
+        u32 divisor;
+
+        std::size_t Hash() const;
+        bool operator==(const VertexBinding& rhs) const;
```
lioncash: These structs should also have `operator!=` for logical symmetry.

src/video_core/renderer_vulkan/fixed_pipeline_state.h
```diff
@@ -0,0 +1,231 @@
...
+    struct VertexBinding {
+        constexpr VertexBinding(u32 index, u32 stride, u32 divisor)
+            : index{index}, stride{stride}, divisor{divisor} {}
+        VertexBinding() = default;
+
+        u32 index;
+        u32 stride;
+        u32 divisor;
+
+        std::size_t Hash() const;
```
lioncash: Shouldn't these hash functions be marked as noexcept?

PR #3248, opened 2019-12-25
vk_image: Add an image object abstraction

ReinUsesLisp:
This object's job is to contain an image and manage its transitions.
Since Nvidia hardware doesn't know what a transition is but Vulkan
requires them anyway, we have to state track image subresources
individually.

To avoid the overhead of tracking each subresource in images with many
subresources (think of cubemap arrays with several mipmaps), this commit
tracks when subresources have diverged. As long as this doesn't happen
we can check the state of the first subresource (that will be shared
with all subresources) and update accordingly.

Image transitions are deferred to the scheduler command buffer.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_image.cpp
+++ b/src/video_core/renderer_vulkan/vk_image.cpp
@@ -1,0 +1,106 @@
+// Copyright 2018 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <memory>
+#include <vector>
+
+#include "common/assert.h"
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/renderer_vulkan/vk_device.h"
+#include "video_core/renderer_vulkan/vk_image.h"
+#include "video_core/renderer_vulkan/vk_scheduler.h"
+
+namespace Vulkan {
+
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_vulkan/vk_image.h
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /// Returns the Vulkan aspect mask.
+    vk::ImageAspectFlags GetAspectMask() const {
+        return aspect_mask;
+    }
+
+private:
+    struct SubrangeState final {
+        vk::AccessFlags access = {};                          ///< Current access bits.
```
bunnei: nit: Equals here is unnecessary

src/video_core/renderer_vulkan/vk_image.h
```diff
@@ -0,0 +1,84 @@
+// Copyright 2018 yuzu Emulator Project
```
bunnei: nit: 2019 ?

src/video_core/renderer_vulkan/vk_image.cpp
```diff
@@ -0,0 +1,106 @@
+// Copyright 2018 yuzu Emulator Project
```
bunnei: nit: 2019 ?

PR #3264, opened 2020-01-01
vk_descriptor_pool: Initial implementation

ReinUsesLisp:
Create a large descriptor pool where we allocate all our descriptors
from. It has to be wide enough to support any pipeline, hence its large
numbers.

If the descritor pool is filled, we allocate more memory at that moment.
This way we can take advantage of permissive drivers like Nvidia's that
allocate more descriptors than what the spec requires.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_descriptor_pool.cpp
+++ b/src/video_core/renderer_vulkan/vk_descriptor_pool.cpp
@@ -1,0 +1,89 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <memory>
+#include <vector>
+
+#include "common/common_types.h"
+#include "video_core/renderer_vulkan/declarations.h"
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_vulkan/vk_descriptor_pool.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include "common/common_types.h"
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/renderer_vulkan/vk_descriptor_pool.h"
+#include "video_core/renderer_vulkan/vk_device.h"
+#include "video_core/renderer_vulkan/vk_resource_manager.h"
+
+namespace Vulkan {
+
+// Prefer small grow rates to avoid saturating the descriptor pool with barely used pipelines.
+static constexpr std::size_t SETS_GROW_RATE = 0x20;
```
lioncash:
```suggestion
constexpr std::size_t SETS_GROW_RATE = 0x20;
```

const/constexpr file-scope variables are internally linked by default.

PR #3270, opened 2020-01-03
Shader_Ir: Miscellaneous features for future implementations

FernandoS27:
This provides a set of features needed for further implementations into Shader Tracking and code deriving such as array samplers. This adds two small features: Determining information from the in use gpu driver and saving bound buffer used to the shader disk cache.

This PR, sadly, invalidates previous shader caches as this changes are needed for future fixes.

### DIFF
--- a/src/video_core/guest_driver.cpp
+++ b/src/video_core/guest_driver.cpp
@@ -2,6 +2,8 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
+
 #include "video_core/guest_driver.h"
 
 namespace VideoCore {
@@ -28,7 +30,7 @@ void GuestDriverProfile::DeduceTextureHandlerSize(std::vector<u32>&& bound_offse
         return;
     }
     texture_handler_size_deduced = true;
-    texture_handler_size = sizeof(u32) * min_val;
+    texture_handler_size = min_texture_handler_size * min_val;
 }
 
 } // namespace VideoCore
--- a/src/video_core/guest_driver.cpp
+++ b/src/video_core/guest_driver.cpp
@@ -3,7 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
-#include <climits>
+#include <limits>
 
 #include "video_core/guest_driver.h"
 
@@ -17,10 +17,9 @@ void GuestDriverProfile::DeduceTextureHandlerSize(std::vector<u32>&& bound_offse
     if (size < 2) {
         return;
     }
-    std::sort(bound_offsets.begin(), bound_offsets.end(),
-              [](const u32& a, const u32& b) { return a < b; });
-    u32 min_val = UINT_MAX;
-    for (std::size_t i = 1; i < size; i++) {
+    std::sort(bound_offsets.begin(), bound_offsets.end(), std::less{});
+    u32 min_val = std::numeric_limits<u32>::max();
+    for (std::size_t i = 1; i < size; ++i) {
         if (bound_offsets[i] == bound_offsets[i - 1]) {
             continue;
         }
--- a/src/video_core/guest_driver.cpp
+++ b/src/video_core/guest_driver.cpp
@@ -1,0 +1,34 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "video_core/guest_driver.h"
+
+namespace VideoCore {
+
+void GuestDriverProfile::DeduceTextureHandlerSize(std::vector<u32>&& bound_offsets) {
+    if (texture_handler_size_deduced) {
+        return;
+    }
+    std::size_t size = bound_offsets.size();
+    if (size < 2) {
+        return;
+    }
+    std::sort(bound_offsets.begin(), bound_offsets.end(),
+              [](const u32& a, const u32& b) { return a < b; });
+    u32 min_val = 0xFFFFFFFF; // set to highest possible 32 bit integer;
+    for (std::size_t i = 1; i < size; i++) {
+        if (bound_offsets[i] == bound_offsets[i - 1]) {
+            continue;
+        }
+        const u32 new_min = bound_offsets[i] - bound_offsets[i - 1];
+        min_val = std::min(min_val, new_min);
+    }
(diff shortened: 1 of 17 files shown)


src/video_core/guest_driver.cpp
```diff
@@ -0,0 +1,36 @@
+// Copyright 2019 yuzu Emulator Project
```
bunnei: 2020?

src/video_core/guest_driver.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "video_core/guest_driver.h"
+
+namespace VideoCore {
+
+void GuestDriverProfile::DeduceTextureHandlerSize(std::vector<u32>&& bound_offsets) {
+    if (texture_handler_size_deduced) {
+        return;
+    }
+    std::size_t size = bound_offsets.size();
```
bunnei: Can be `const`

src/video_core/guest_driver.cpp
```diff
@@ -0,0 +1,36 @@
...
+    if (size < 2) {
+        return;
+    }
+    std::sort(bound_offsets.begin(), bound_offsets.end(),
+              [](const u32& a, const u32& b) { return a < b; });
+    u32 min_val = 0xFFFFFFFF; // set to highest possible 32 bit integer;
```
bunnei: <climits>  ?

src/video_core/guest_driver.h
```diff
@@ -0,0 +1,39 @@
+// Copyright 2019 yuzu Emulator Project
```
bunnei: 2020?

src/video_core/guest_driver.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include "common/common_types.h"
+
+namespace VideoCore {
+
+/**
+ * The GuestDriverProfile class is used to learn about the GPU drivers behavior and collect
+ * information necessary for impossible to avoid HLE methods like shader tracks.
+ */
+class GuestDriverProfile {
```
bunnei: Shouldn't this be HostDriveProfile? (i.e. the users PC is the "host", the emulated hardware (switch) is the "Guest")

src/video_core/shader/const_buffer_locker.h
```diff
@@ -49,6 +52,10 @@ class ConstBufferLocker {
     /// Inserts a bindless sampler key.
     void InsertBindlessSampler(u32 buffer, u32 offset, Tegra::Engines::SamplerDescriptor sampler);
 
+    /// Set the bound buffer for this locker.
+
```
bunnei: Remove extra line

src/video_core/shader/decode.cpp
```diff
@@ -315,4 +315,25 @@ u32 ShaderIR::DecodeInstr(NodeBlock& bb, u32 pc) {
     return pc + 1;
 }
 
+void ShaderIR::PostDecode() {
+    // Deduce texture handler size if needed
+    auto* gpu_driver = locker.AccessGuestDriverProfile();
```
bunnei: if (!gpu_driver) {
    // Log error?
    return;
}

(early-return code is easier to read than many nested if's)

src/video_core/guest_driver.cpp
bunnei: * use climits

src/video_core/shader/decode.cpp
FernandoS27: yeah but on design I don't want early return as different unrelated passes may be run on this function

src/video_core/guest_driver.h
FernandoS27: This gets info on the driver used by the game not the one in the PC. Therefore, it's a guest driver.

src/video_core/guest_driver.cpp
```diff
@@ -0,0 +1,37 @@
...
+void GuestDriverProfile::DeduceTextureHandlerSize(std::vector<u32>&& bound_offsets) {
+    if (texture_handler_size_deduced) {
+        return;
+    }
+    const std::size_t size = bound_offsets.size();
+    if (size < 2) {
+        return;
+    }
+    std::sort(bound_offsets.begin(), bound_offsets.end(),
+              [](const u32& a, const u32& b) { return a < b; });
```
ReinUsesLisp: Use `std::less{}` instead of a lambda.

src/video_core/rasterizer_interface.h
```diff
@@ -78,5 +79,12 @@ class RasterizerInterface {
     /// Initialize disk cached resources for the game being emulated
     virtual void LoadDiskResources(const std::atomic_bool& stop_loading = false,
                                    const DiskResourceLoadCallback& callback = {}) {}
+
+    GuestDriverProfile& AccessGuestDriverProfile() {
```
ReinUsesLisp: Add a documentation commentary here for consistency.

src/video_core/shader/shader_ir.h
```diff
@@ -187,6 +187,7 @@ class ShaderIR final {
     };
 
     void Decode();
+    void PostDecode();
```
ReinUsesLisp: Marking it for removal, read above.

src/video_core/guest_driver.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <vector>
+
+#include "common/common_types.h"
+
+namespace VideoCore {
+
+/**
+ * The GuestDriverProfile class is used to learn about the GPU drivers behavior and collect
+ * information necessary for impossible to avoid HLE methods like shader tracks.
```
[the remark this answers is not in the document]
FernandoS27: You can't solve entscheidung problems.

src/video_core/shader/shader_ir.cpp
```diff
@@ -27,6 +27,7 @@ ShaderIR::ShaderIR(const ProgramCode& program_code, u32 main_offset, CompilerSet
                    ConstBufferLocker& locker)
     : program_code{program_code}, main_offset{main_offset}, settings{settings}, locker{locker} {
     Decode();
+    PostDecode();
```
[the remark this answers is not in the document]
FernandoS27: No, it makes no logic to for the methods in PostDecode to be within Decode as normally they have nothing to do with Decoding a Shader but more to do with learning from it.

PR #3276, opened 2020-01-05
vk_update_descriptor/vk_renderpass_cache: Add pipeline cache dependencies

ReinUsesLisp:
This PR adds two new abstractions required to add the pipeline cache in a future PR. Description in individual commits.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_update_descriptor.h
+++ b/src/video_core/renderer_vulkan/vk_update_descriptor.h
@@ -1,0 +1,86 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <type_traits>
+#include <variant>
+#include <boost/container/static_vector.hpp>
+
+#include "common/common_types.h"
+#include "video_core/renderer_vulkan/declarations.h"
+
+namespace Vulkan {
+
+class VKDevice;
+class VKScheduler;
+
+class DescriptorUpdateEntry {
+public:
+    explicit DescriptorUpdateEntry() : image{} {}
+
+    DescriptorUpdateEntry(vk::DescriptorImageInfo image) : image{image} {}
+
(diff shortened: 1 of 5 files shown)


src/video_core/renderer_vulkan/vk_update_descriptor.h
```diff
@@ -0,0 +1,86 @@
...
+        std::size_t size;
+    };
+    using Variant = std::variant<vk::DescriptorImageInfo, Buffer, vk::BufferView>;
+    // Old gcc versions don't consider this trivially copyable.
+    // static_assert(std::is_trivially_copyable_v<Variant>);
+
+    const VKDevice& device;
+    VKScheduler& scheduler;
+
+    boost::container::static_vector<Variant, 0x400> entries;
```
bunnei: Curious, what is the advantage of using a boost type here over std:: container?

src/video_core/renderer_vulkan/vk_update_descriptor.h
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    vk::ImageLayout* GetLastImageLayout() {
+        return &std::get<vk::DescriptorImageInfo>(entries.back()).imageLayout;
+    }
+
+private:
+    struct Buffer {
+        const vk::Buffer* buffer;
+        u64 offset;
```
bunnei: Can/should these also be `const` to enforce that they are initialized?

src/video_core/renderer_vulkan/vk_update_descriptor.h
ReinUsesLisp: The `std::` equivalent would be a `std::array` plus an iterator pointing to the last member.
`boost::container::static_vector` offers an interface like `std::vector` but on static storage (it avoids a heap allocation and its dereference).
The only drawback I'm aware of is that it's not `trivially-copyable` on `trivially-copyable` objects, since it only copies objects that are currently part of the vector.

src/video_core/renderer_vulkan/vk_update_descriptor.h
ReinUsesLisp: Addressed it by adding a default member initialize `{}`.

PR #3278, opened 2020-01-06
renderer_vulkan: Buffer cache, stream buffer and memory manager changes

ReinUsesLisp:
* Allocate memory in discrete exponentially increasing chunks until the 128 MiB threshold. Allocations larger than that increase linearly by 256 MiB (depending on the required size). This allows to use small allocations for small resources.
* Move memory maps to a RAII abstraction. To optimize for debugging tools (like RenderDoc) users will map/unmap on usage. If this ever becomes a noticeable overhead (from my profiling it doesn't) we can transparently move to persistent memory maps without harming the API, getting optimal performance for both gameplay and debugging.
* Improve messages on exceptional situations.
* Fix typos "requeriments" -> "requirements".
* Small style changes.

Edit: Detailed description about buffer cache and stream buffer changes is in its individual commit.


src/video_core/renderer_vulkan/vk_memory_manager.h
```diff
@@ -78,10 +90,46 @@ class VKMemoryCommitImpl final {
...
+    }
+
+    /// Returns the address of the memory map;
+    operator u8*() const noexcept {
+        return address;
+    }
+
+private:
+    const VKMemoryCommitImpl* commit; ///< Mapped memory commit.
+    u8* address;                      ///< Address to the mapped memory.
```
bunnei: This can be `const`, right?

bunnei:
../src/video_core/renderer_vulkan/vk_stream_buffer.cpp:78:30: error: 'class Vulkan::VKMemoryCommitImpl' has no member named 'GetData'

ReinUsesLisp:
There was a circular dependency issue. I had to add the commits for the buffer cache and stream buffer changes.

src/video_core/renderer_vulkan/vk_memory_manager.h
ReinUsesLisp: Addressed it by adding `{}` on both members. Notice that `commit` is not a constant, it points to a constant object.

We may want to make this object movable in the future so I don't think having const members is a good idea.

PR #3279, opened 2020-01-07
vk_pipeline_cache: Initial implementation

ReinUsesLisp:
This PR adds Vulkan's pipeline cache (not to be confused with a disk pipeline cache). Detailed description about it in the respective commit.

There are chunks of code we can share between OpenGL's shader cache and Vulkan's pipeline cache that are duplicated here. These should be moved to a common helper file in a future contribution.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.h
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.h
@@ -6,22 +6,48 @@
 
 #include <array>
 #include <cstddef>
+#include <memory>
+#include <tuple>
+#include <type_traits>
+#include <unordered_map>
+#include <utility>
 #include <vector>
 
 #include <boost/functional/hash.hpp>
 
 #include "common/common_types.h"
+#include "video_core/engines/const_buffer_engine_interface.h"
 #include "video_core/engines/maxwell_3d.h"
+#include "video_core/rasterizer_cache.h"
 #include "video_core/renderer_vulkan/declarations.h"
 #include "video_core/renderer_vulkan/fixed_pipeline_state.h"
+#include "video_core/renderer_vulkan/vk_graphics_pipeline.h"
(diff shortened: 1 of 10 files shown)


src/video_core/renderer_vulkan/vk_rasterizer.h
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "video_core/rasterizer_interface.h"
+
+namespace Vulkan {
+
+class RasterizerVulkan : public VideoCore::RasterizerInterface {};
```
bunnei: `final` ?

src/video_core/renderer_vulkan/vk_pipeline_cache.h
```diff
@@ -0,0 +1,200 @@
...
+
+class CachedShader;
+using Shader = std::shared_ptr<CachedShader>;
+using Maxwell = Tegra::Engines::Maxwell3D::Regs;
+
+using ProgramCode = std::vector<u64>;
+
+struct GraphicsPipelineCacheKey {
+    FixedPipelineState fixed_state;
+    std::array<GPUVAddr, Maxwell::MaxShaderProgram> shaders;
```
bunnei: Should this be zero'd out?

src/video_core/renderer_vulkan/vk_rasterizer.h
ReinUsesLisp: This is a stub to avoid circular dependencies. The real implementation will be added later. Adding `final` here would require implementing all the pure virtual methods.

src/video_core/renderer_vulkan/vk_pipeline_cache.h
ReinUsesLisp: That'd initialize the structure twice, something I'd like to avoid in the hot loop.

PR #3283, opened 2020-01-08
vk_compute_pass: Add compute passes to emulate missing Vulkan features

ReinUsesLisp:
This currently only supports quad arrays and u8 indices. We'll have to add indexed quads, polygons, quad strips, line loops, primitive restart for these and mixed u8 index support... None of these features are natively supported on Vulkan even when they are natively supported on hardware (with the exception of raw u8 indices that are available as an extension).

In the future we can remove quad arrays with a table written from the CPU, but this was used to bootstrap the other passes helpers and it was left in the code.

The blob code is generated from the "shaders/" directory. Read the instructions there to know how to generate the SPIR-V.

Header dates in the committed files are not a copy and paste typo, this is when they were written.


src/video_core/renderer_vulkan/vk_compute_pass.cpp
```diff
@@ -0,0 +1,337 @@
...
+    0x1c, 0x00, 0x00, 0x00, 0x3d, 0x00, 0x04, 0x00, 0x06, 0x00, 0x00, 0x00, 0x24, 0x00, 0x00, 0x00,
+    0x08, 0x00, 0x00, 0x00, 0x3d, 0x00, 0x04, 0x00, 0x06, 0x00, 0x00, 0x00, 0x25, 0x00, 0x00, 0x00,
+    0x08, 0x00, 0x00, 0x00, 0x41, 0x00, 0x06, 0x00, 0x26, 0x00, 0x00, 0x00, 0x27, 0x00, 0x00, 0x00,
+    0x15, 0x00, 0x00, 0x00, 0x23, 0x00, 0x00, 0x00, 0x25, 0x00, 0x00, 0x00, 0x3d, 0x00, 0x04, 0x00,
+    0x11, 0x00, 0x00, 0x00, 0x28, 0x00, 0x00, 0x00, 0x27, 0x00, 0x00, 0x00, 0x71, 0x00, 0x04, 0x00,
+    0x1e, 0x00, 0x00, 0x00, 0x29, 0x00, 0x00, 0x00, 0x28, 0x00, 0x00, 0x00, 0x41, 0x00, 0x06, 0x00,
+    0x2a, 0x00, 0x00, 0x00, 0x2b, 0x00, 0x00, 0x00, 0x22, 0x00, 0x00, 0x00, 0x23, 0x00, 0x00, 0x00,
+    0x24, 0x00, 0x00, 0x00, 0x3e, 0x00, 0x03, 0x00, 0x2b, 0x00, 0x00, 0x00, 0x29, 0x00, 0x00, 0x00,
+    0xf9, 0x00, 0x02, 0x00, 0x1d, 0x00, 0x00, 0x00, 0xf8, 0x00, 0x02, 0x00, 0x1d, 0x00, 0x00, 0x00,
+    0xfd, 0x00, 0x01, 0x00, 0x38, 0x00, 0x01, 0x00};
```
FernandoS27: Where does this values come from and what are their purpose. Also, can't they be computed on init.

ReinUsesLisp: It's the SPIR-V shader, I'll add a reference in the code (it's explained in the PR and commit description).

PR #3287, opened 2020-01-09
shader_ir/memory: Implement u16 and u8 for STG and LDG

ReinUsesLisp:
Using the same technique we used for u8 on LDG, implement u16.

In the case of STG, load memory and insert the value we want to set
into it with bitfieldInsert. Then set that value.


ReinUsesLisp:
I forgot to mention, in the future we might want to use extensions (GL_NV_gpu_shader5 and Vulkan extensions) here to avoid loading memory.

PR #3297, opened 2020-01-12
GUI: Remove controls input profiles

CJBok:
Remove Controls Input profiles screen.

image
image


theboy181:
Nice cleanup! What was the original intent of the profile selection?

CJBok:
> Thank you for this change. I'd like at some point to talk about how to reintroduce controller profiles in yuzu. Considering this is actively broken in master for many users, I'm fine removing it for the time being. If you are interested in working on controller profiles, lemme know and we can talk more about it on Discord.
> 
> That said, this doesn't seem to handle migrating the users controller configuration properly or removing the settings for this though. Upgrade paths are probably just a pain to support in the long run, so I'm fine not having any sort of upgrade path. We still need to remove `UISettings::values.profile_index` though, so please do that and I'll change this to an approve.

Setting is now removed. Regarding reintroduction of input profiles I'll look you up on discord.

BSoDGamingYT:
"Even though initially the profiles worked on a wide swath of games, many power users started advocating for users to use custom when there was no real reason to. "

The fact that countless people (Many on the yuzu moderation/support team) reported that the profiles didn't work, despite being mapped fully and correctly is the reason "power users" had to tell others to use the custom controller configuration.

"Thus, there is a much better solution that would preserve the ease of use goal of profiles with minimal effort: Adding new profiles as necessary to improve compatibility with newer titles."

Why hasn't this solution been done in the last year? These profiles have been broken for that amount of time and for longer. In fact, the last time I brought this issue up (About 5 months ago), this exact same thing was said. Yet in that time, nothing has been changed, the profiles remain broken and users are constantly reporting problems with their input (Or lack thereof) Resulting, again in support staff having to tell users to use the custom profile anyway.

CJBok:
Different solution required

PR #3300, opened 2020-01-13
vk_texture_cache: Implement generic texture cache on Vulkan

ReinUsesLisp:
It currently ignores PBO linearisations since these should be dropped as soon as possible on OpenGL.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_texture_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_texture_cache.cpp
@@ -1,0 +1,484 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <array>
+#include <cstddef>
+#include <cstring>
+#include <memory>
+#include <variant>
+#include <vector>
+
+#include "common/alignment.h"
+#include "common/assert.h"
+#include "common/common_types.h"
+#include "core/core.h"
+#include "core/memory.h"
+#include "video_core/engines/maxwell_3d.h"
+#include "video_core/morton.h"
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/renderer_vulkan/maxwell_to_vk.h"
+#include "video_core/renderer_vulkan/vk_device.h"
+#include "video_core/renderer_vulkan/vk_memory_manager.h"
+#include "video_core/renderer_vulkan/vk_rasterizer.h"
+#include "video_core/renderer_vulkan/vk_staging_buffer_pool.h"
+#include "video_core/renderer_vulkan/vk_texture_cache.h"
+#include "video_core/surface.h"
+#include "video_core/textures/convert.h"
+
+namespace Vulkan {
+
+using VideoCore::MortonSwizzle;
+using VideoCore::MortonSwizzleMode;
+
+using Tegra::Texture::SwizzleSource;
+using VideoCore::Surface::PixelFormat;
+using VideoCore::Surface::SurfaceCompression;
+using VideoCore::Surface::SurfaceTarget;
+
+namespace {
+
+vk::ImageType SurfaceTargetToImage(SurfaceTarget target) {
+    switch (target) {
+    case SurfaceTarget::Texture1D:
+    case SurfaceTarget::Texture1DArray:
+        return vk::ImageType::e1D;
+    case SurfaceTarget::Texture2D:
+    case SurfaceTarget::Texture2DArray:
+    case SurfaceTarget::TextureCubemap:
+    case SurfaceTarget::TextureCubeArray:
+        return vk::ImageType::e2D;
+    case SurfaceTarget::Texture3D:
+        return vk::ImageType::e3D;
+    }
+    UNREACHABLE_MSG("Unknown texture target={}", static_cast<u32>(target));
+    return {};
+}
+
+vk::ImageAspectFlags PixelFormatToImageAspect(PixelFormat pixel_format) {
+    if (pixel_format < PixelFormat::MaxColorFormat) {
+        return vk::ImageAspectFlagBits::eColor;
+    } else if (pixel_format < PixelFormat::MaxDepthFormat) {
+        return vk::ImageAspectFlagBits::eDepth;
+    } else if (pixel_format < PixelFormat::MaxDepthStencilFormat) {
+        return vk::ImageAspectFlagBits::eDepth | vk::ImageAspectFlagBits::eStencil;
+    } else {
+        UNREACHABLE_MSG("Invalid pixel format={}", static_cast<u32>(pixel_format));
+        return vk::ImageAspectFlagBits::eColor;
+    }
+}
+
+vk::ImageViewType GetImageViewType(SurfaceTarget target) {
+    switch (target) {
+    case SurfaceTarget::Texture1D:
+        return vk::ImageViewType::e1D;
+    case SurfaceTarget::Texture2D:
+        return vk::ImageViewType::e2D;
+    case SurfaceTarget::Texture3D:
+        return vk::ImageViewType::e3D;
+    case SurfaceTarget::Texture1DArray:
+        return vk::ImageViewType::e1DArray;
+    case SurfaceTarget::Texture2DArray:
+        return vk::ImageViewType::e2DArray;
+    case SurfaceTarget::TextureCubemap:
+        return vk::ImageViewType::eCube;
+    case SurfaceTarget::TextureCubeArray:
+        return vk::ImageViewType::eCubeArray;
+    case SurfaceTarget::TextureBuffer:
+        break;
+    }
+    UNREACHABLE();
+    return {};
+}
+
+UniqueBuffer CreateBuffer(const VKDevice& device, const SurfaceParams& params) {
+    // TODO(Rodrigo): Move texture buffer creation to the buffer cache
+    const vk::BufferCreateInfo buffer_ci({}, params.GetHostSizeInBytes(),
+                                         vk::BufferUsageFlagBits::eUniformTexelBuffer |
+                                             vk::BufferUsageFlagBits::eTransferSrc |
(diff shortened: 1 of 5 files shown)


src/video_core/texture_cache/surface_params.h
```diff
@@ -209,6 +209,11 @@ class SurfaceParams {
         return target == VideoCore::Surface::SurfaceTarget::TextureBuffer;
     }
 
+    /// Returns the number of layers in the surface.
+    std::size_t GetNumLayers() const {
+        return is_layered ? depth : 1;
+    }
+
```
FernandoS27: Try not to add Ninja changes into a PR. This should have been in a separate PR.

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -0,0 +1,482 @@
...
+            aspect = vk::ImageAspectFlagBits::eDepth;
+            UNIMPLEMENTED();
+        }
+
+        // Vulkan doesn't seem to understand swizzling of a depth stencil image, use identity
+        swizzle_x = vk::ComponentSwizzle::eR;
+        swizzle_y = vk::ComponentSwizzle::eG;
+        swizzle_z = vk::ComponentSwizzle::eB;
+        swizzle_w = vk::ComponentSwizzle::eA;
+    }
```
FernandoS27: add comments here on why is this necessary.

src/video_core/texture_cache/surface_params.h
ReinUsesLisp: Moved it to a different commit. It's just moving `GetNumLayers` from private to public and adding a commentary on it.

src/video_core/renderer_vulkan/vk_texture_cache.cpp
ReinUsesLisp: Done

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -0,0 +1,484 @@
...
+    for (u32 level = 0; level < params.num_levels; ++level) {
+        scheduler.Record([image = image->GetHandle(), buffer = *buffer.handle,
+                          copy = GetBufferImageCopy(level)](auto cmdbuf, auto& dld) {
+            cmdbuf.copyImageToBuffer(image, vk::ImageLayout::eTransferSrcOptimal, buffer, {copy},
+                                     dld);
+        });
+    }
+    scheduler.Finish();
+
+    // TODO(Rodrigo): Use an intern buffer for staging buffers and avoid this unnecesary memcpy.
```
MysticExile:
```suggestion
    // TODO(Rodrigo): Use an intern buffer for staging buffers and avoid this unnecessary memcpy.
```
nit

src/video_core/renderer_vulkan/vk_texture_cache.h
```diff
@@ -0,0 +1,244 @@
...
+    const vk::BufferView buffer_view;
+    const vk::ImageAspectFlags aspect_mask;
+
+    const VKDevice& device;
+    CachedSurface& surface;
+    const u32 base_layer;
+    const u32 num_layers;
+    const u32 base_level;
+    const u32 num_levels;
+    vk::ImageViewType image_view_type{};
```
bunnei: This can be const as well, right?

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -0,0 +1,484 @@
...
+                             GPUVAddr gpu_addr, const SurfaceParams& params)
+    : SurfaceBase<View>{gpu_addr, params}, system{system}, device{device},
+      resource_manager{resource_manager}, memory_manager{memory_manager}, scheduler{scheduler},
+      staging_pool{staging_pool} {
+    if (params.IsBuffer()) {
+        buffer = CreateBuffer(device, params);
+        commit = memory_manager.Commit(*buffer, false);
+
+        const auto buffer_view_ci = GenerateBufferViewCreateInfo(device, params, *buffer);
+        format = buffer_view_ci.format;
```
bunnei: Should these be moved to the initializer list, and initialized by a helper function? (that way they can be const)

src/video_core/renderer_vulkan/vk_texture_cache.h
```diff
@@ -0,0 +1,244 @@
+// Copyright 2019 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
+#include <tuple>
```
bunnei: unused include

src/video_core/renderer_vulkan/vk_texture_cache.h
```diff
@@ -0,0 +1,244 @@
...
+#include <unordered_map>
+
+#include <boost/functional/hash.hpp>
+#include <boost/icl/interval_map.hpp>
```
bunnei: unused includes

src/video_core/renderer_vulkan/vk_texture_cache.h
```diff
@@ -0,0 +1,244 @@
...
+#include "video_core/rasterizer_cache.h"
+#include "video_core/renderer_vulkan/declarations.h"
+#include "video_core/renderer_vulkan/vk_image.h"
+#include "video_core/renderer_vulkan/vk_memory_manager.h"
+#include "video_core/renderer_vulkan/vk_scheduler.h"
+#include "video_core/surface.h"
+#include "video_core/texture_cache/surface_base.h"
+#include "video_core/texture_cache/texture_cache.h"
+#include "video_core/textures/decoders.h"
+#include "video_core/textures/texture.h"
```
bunnei: there are some unused includes in here

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -0,0 +1,484 @@
...
+                      is_linear](auto cmdbuf, auto& dld) {
+        cmdbuf.blitImage(src_image, vk::ImageLayout::eTransferSrcOptimal, dst_image,
+                         vk::ImageLayout::eTransferDstOptimal, {blit},
+                         is_linear ? vk::Filter::eLinear : vk::Filter::eNearest, dld);
+    });
+}
+
+void VKTextureCache::BufferCopy(Surface& src_surface, Surface& dst_surface) {
+    // Currently unimplemented. PBO copies should be dropped and we should use a render pass to
+    // convert from color to depth and viceversa.
```
bunnei: Should we log a warning here or something?

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -0,0 +1,484 @@
...
+        {aspect, base_level, num_levels, base_layer, num_layers});
+
+    const auto dev = device.GetLogical();
+    image_view = dev.createImageViewUnique(image_view_ci, nullptr, device.GetDispatchLoader());
+    return last_image_view = *image_view;
+}
+
+bool CachedSurfaceView::IsOverlapping(const View& rhs) const {
+    // TODO(Rodrigo): Also test for layer and mip level overlaps.
+    return &surface == &rhs->surface;
```
bunnei: this is just testing if these are the same surface, right? What about partial overlap?

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -0,0 +1,484 @@
...
+void CachedSurface::DecorateSurfaceName() {
+    // TODO(Rodrigo): Add name decorations
+}
+
+View CachedSurface::CreateView(const ViewParams& params) {
+    return CreateViewInner(params, false);
+}
+
+View CachedSurface::CreateViewInner(const ViewParams& params, bool is_proxy) {
+    auto view = std::make_shared<CachedSurfaceView>(device, *this, params, is_proxy);
```
bunnei: nit: `const` ?

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -0,0 +1,484 @@
...
+    switch (params.target) {
+    case SurfaceTarget::TextureCubemap:
+    case SurfaceTarget::TextureCubeArray:
+        flags |= vk::ImageCreateFlagBits::eCubeCompatible;
+        [[fallthrough]];
+    case SurfaceTarget::Texture1D:
+    case SurfaceTarget::Texture1DArray:
+    case SurfaceTarget::Texture2D:
+    case SurfaceTarget::Texture2DArray:
+        extent = vk::Extent3D(params.width, params.height, 1);
```
bunnei: Is 1 right for arrays?

src/video_core/renderer_vulkan/vk_texture_cache.cpp
ReinUsesLisp: The problem is that it's initialized in different ways depending on `params.IsBuffer`. Is there a solution for this without repeating code execution at runtime?

src/video_core/renderer_vulkan/vk_texture_cache.cpp
ReinUsesLisp: Yes, it means that the depth is 1. On Vulkan number of layers is different to depth.

src/video_core/renderer_vulkan/vk_texture_cache.cpp
ReinUsesLisp: I think that's what the `TODO` commentary suggests?

It should test if the view is partially (or fully) overlapping. It's used to determine if a sampled texture is overlapping with the render target, making us change the image layout. Vulkan validation layers will catch if this fails.

I'll rename it to `IsSameSurface` to avoid new callers expecting a fully implemented behaviour.

src/video_core/renderer_vulkan/vk_texture_cache.cpp
ReinUsesLisp: Other variables (like `commit`) are also initialized depending on that.

PR #3301, opened 2020-01-13
video_core: Remove gl_state and use a state tracker based on dirty flags

ReinUsesLisp:
`gl_state` has several issues, including the usage of global variables to work and being expensive.

This PR initially drops dirty flags to simplify its reimplementation. Then it gradually removes `gl_state` and calls OpenGL directly, incrementing the driver overhead while reducing ours.

This includes dropping the VAO cache. As explained [here](https://developer.nvidia.com/sites/default/files/akamai/gamedev/docs/Porting%20Source%20to%20Linux.pdf#page=57) by Nvidia and from my profiling using Nsight, calling state changes without using VAOs proves to be less expensive than just calling the state changes directly like any other OpenGL state. This is how it works on NVN.

After that, this commit gradually reintroduces dirty flags (more on that below) for each OpenGL state change, reducing what was an expensive `gl_state` comparison to a single branch (that can be turned into a cold branch once we switch to C++20).

Dirty flags are handled per backend. This is because OpenGL and Vulkan dynamic states are different and can't be matched in some cases. By separating them, Vulkan stops holding back what we can do on OpenGL.

These flags are implementing using a `std::bitset<>` table. We can see [here](https://godbolt.org/z/wRae7m) that querying bitsets larger than `sizeof(std::size_t) * CHAR_BIT` (or 64 bits on most machines we care about) are properly handled by modern compilers.

Dirty pointers on the other side are implemented by using a `std::array<std::array<u8, NUM_REGS>, N>` where `N` is 2. We use two layers of dirty flags to avoid doing more branches than what we need in the hot loop while also avoiding unnecessary changes when the state changes. For instance for viewports we have a dirty flag for each viewport index, while also a dirty flag for all viewports.

When a state change is unsynchronized with the guest state (e.g. blits or rendering to the screen) dirty flags for this are manually turned on. This requires reading the spec carefully for knowing what changes are required for some operations (like `glBlitFramebuffer` and `glClear`).

This is marked as work in progress because it's missing the following bits:
- [x] Vulkan integration once it's merged into master.
- [ ] Measure and implement performance impact of unnecessary `glBindFramebuffer` calls with a `gl_state`-like approach.
- [ ] Measure and implement state tracking for `glBindVertexBuffer`. After `glNamedBufferSubData` this is our most expensive call according to Nsight on Nvidia's driver.
- [ ] Investigate a way to fix the three-step initialization for the GPU.
- [x] Investigate if some of the unsynchronized changes are needed.

I know this looks like lots of lines for a single PR, but most of it is boilerplate and it's splitted in multiple commits.


ReinUsesLisp:
I'd prefer to look at the remaining issues once it's merged. The framebuffer bindings can be simplified but it requires some unrelated changes.

src/video_core/dirty_flags.h
```diff
@@ -0,0 +1,51 @@
...
+    ColorBuffer5,
+    ColorBuffer6,
+    ColorBuffer7,
+    ZetaBuffer,
+
+    LastCommonEntry,
+};
+
+template <typename Integer>
+inline void FillBlock(Tegra::Engines::Maxwell3D::DirtyState::Table& table, std::size_t begin,
```
lioncash:
```suggestion
void FillBlock(Tegra::Engines::Maxwell3D::DirtyState::Table& table, std::size_t begin,
```

Templates are inline by default unless they're specializations

Ditto for the template following this one.

src/video_core/renderer_vulkan/vk_state_tracker.h
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    bool TouchStencilProperties() {
+        return Exchange(Dirty::StencilProperties, false);
+    }
+
+private:
+    bool Exchange(std::size_t id, bool new_value) const noexcept {
+        auto& flags = system.GPU().Maxwell3D().dirty.flags;
+        const bool is_dirty = flags[id];
```
lioncash: Wouldn't `std::exchange` be usable here?

ReinUsesLisp: I [tried](https://godbolt.org/z/K5ZBfv) to, but `std::bitset` is incompatible with `std::exchange`.

bunnei:
Removing EA tag due to conflict with #3430

bunnei:
@ReinUsesLisp can you please rebase? Thanks!

PR #3309, opened 2020-01-16
GUI/gamelist: add "None" as an option for second row and remove dynamically duplicate row options

Simek:
Currently user can achieve single row entries on the game list by selecting same option twice in configuration panel. Inspired by Citra I have implemented a bit user friendlier game list row configuration:
* second row has now "None" option
* duplicate options are dynamically removed on initial render and on user interaction

I'm not a QT nor C++ expert, I'm sure this could be written better so I'm awaiting reviews and comments ;) 

I have tried few simplifications  (using `curentIndex()` instead of  `currentData()`, avoiding "caching" current selection etc.) but I failed miserably due to some UI glitches or even crashes. Currently committed code, at least for me, working stable and as intended. 

I have left `row1 == row2` condition in `game_list_p.h` for the backward compatibility with older configs after update. It's also explanation why "None" need to be at the end of options array. I have also removed `.isEmpty()` checks because they seems no longer valid with current implementation, correct me please if I'm wrong.

Preview:
yuzugl

### DIFF
--- a/src/yuzu/configuration/configure_gamelist.cpp
+++ b/src/yuzu/configuration/configure_gamelist.cpp
@@ -25,6 +25,7 @@ constexpr std::array row_text_names{
     QT_TR_NOOP("Filetype"),
     QT_TR_NOOP("Title ID"),
     QT_TR_NOOP("Title Name"),
+    QT_TR_NOOP("None"),
 };
 } // Anonymous namespace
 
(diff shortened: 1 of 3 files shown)


bunnei:
LGTM - Thanks!

bunnei:
LGTM - Thanks!

Issue #3017, opened 2019-10-23
Super Mario Odyssey: Stutters when jumping


legend800:
Despite the game running at 60fps, when performing any jumps in the game (sans short hops), there is a momentary pause right before landing. 

Config:
Windows 10 64-bit
GeForce GTX 1080
i7-6700k @ 4.5ghz
16GB RAM

<meta name="titleids"  content="0100000000010000">

ReaperOfSouls1909:
maybe you should add the stubbed service name and a log with it showing

saban1234:
Same issue here and it still persists, very annoying.
i9-9900k, RTX 2080
intel HD audio, or Realtek HD audio
win10 1809
Tell me which kind log would help to pinpoint the problem ?

legend800:
One of the devs recently commented that this is actually an **_audio_** issue, meaning they have the info they need for now. Unfortunately audio is one of the weaker areas of yuzu and gets the least updates, but you never know.

RyuDanuer:
I have it too!

ReaperOfSouls1909:
It’s related to audio and when audio is stripped using a mod the issue is solved so #3266 we need someone to work on all those services.

FearlessTobi:
Fixed by the audio interpolation overhaul.

PR #3310, opened 2020-01-16
audio_core: Switch to a faster interpolation technique

FearlessTobi:
This changes our current, very-time intensive interpolation algorithm to a faster one based on LUTs.

This fixes the so called "audio stutter" in SMO and other games.

**Thanks to Ryujinx for this awesome algorithm**.
(since this is taken from Unlicensed code, it doesn't require attribution)


src/audio_core/algorithm/interpolate.h
```diff
@@ -11,14 +11,134 @@
 
 namespace AudioCore {
 
-struct InterpolationState {
-    static constexpr std::size_t lanczos_taps = 4;
-    static constexpr std::size_t history_size = lanczos_taps * 2 - 1;
+static std::array<short, 512> curve_lut0 = {
```
FearlessTobi: If I put this in the header file, will it be copied around? I'm always not sure about this in C++.

lioncash: These should be placed within the cpp file and be made constexpr.

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +22,40 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
     }
 
-    if (ratio != state.current_ratio) {
-        const double cutoff_frequency = std::min(0.5 / ratio, 0.5 * ratio);
-        state.nyquist = CascadingFilter::LowPass(std::clamp(cutoff_frequency, 0.0, 0.4), 3);
-        state.current_ratio = ratio;
+    std::vector<s16> output;
+    output.resize(input.size() / ratio);
+
+    std::array<short, 512> lut;
```
lioncash:
```suggestion
    const std::array<short, 512>& lut = [ratio] {
        const int step = static_cast<int>(ratio * 0x8000);
        if (step > 0xaaaa) {
            return curve_lut0;
        }
        if (step <= 0x8000) {
            return curve_lut1;
        }
        return curve_lut2;
    }();
```

This way the whole LUT isn't unnecessarily copied entirely each time the function is called.

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +22,40 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
         ratio = 1.0;
     }
 
-    if (ratio != state.current_ratio) {
-        const double cutoff_frequency = std::min(0.5 / ratio, 0.5 * ratio);
-        state.nyquist = CascadingFilter::LowPass(std::clamp(cutoff_frequency, 0.0, 0.4), 3);
-        state.current_ratio = ratio;
+    std::vector<s16> output;
```
lioncash:
```suggestion
    std::vector<s16> output(input.size() / static_cast<std::size_t>(ratio));
```

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +22,40 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
 
-    constexpr std::size_t taps = InterpolationState::lanczos_taps;
-    const std::size_t num_frames = input.size() / 2;
+    int in_offset = 0;
+    for (int out_offset = 0; out_offset < output.size(); out_offset += 2) {
+        int lut_index = (state.frac >> 8) * 4;
 
-    std::vector<s16> output;
-    output.reserve(static_cast<std::size_t>(input.size() / ratio + 4));
+        int l = input[(in_offset + 0) * 2 + 0] * lut[lut_index + 0] +
```
lioncash:
```suggestion
        const int l = input[(in_offset + 0) * 2 + 0] * lut[lut_index + 0] +
```

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +22,40 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
+                input[(in_offset + 1) * 2 + 0] * lut[lut_index + 1] +
+                input[(in_offset + 2) * 2 + 0] * lut[lut_index + 2] +
+                input[(in_offset + 3) * 2 + 0] * lut[lut_index + 3];
+
+        int r = input[(in_offset + 0) * 2 + 1] * lut[lut_index + 0] +
```
lioncash:
```suggestion
        const int r = input[(in_offset + 0) * 2 + 1] * lut[lut_index + 0] +
```

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +22,40 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
-            }
-            output.emplace_back(static_cast<s16>(std::clamp(l, -32768.0, 32767.0)));
-            output.emplace_back(static_cast<s16>(std::clamp(r, -32768.0, 32767.0)));
+        in_offset += new_offset >> 15;
+        state.frac = new_offset & 0x7fff;
 
-            pos += ratio;
-        }
-        pos -= 1.0;
+        output[out_offset + 0] = static_cast<s16>(std::clamp((double)(l >> 15), -32768.0, 32767.0));
```
lioncash:
```suggestion
        output[out_offset + 0] = static_cast<s16>(std::clamp(static_cast<double>(l >> 15), -32768.0, 32767.0));
```

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +22,40 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
+        output[out_offset + 1] = static_cast<s16>(std::clamp((double)(r >> 15), -32768.0, 32767.0));
```
lioncash:
```suggestion
        output[out_offset + 1] = static_cast<s16>(std::clamp(static_cast<double>(r >> 15), -32768.0, 32767.0));
```

src/audio_core/algorithm/interpolate.h
```diff
@@ -11,14 +11,134 @@
...
+    6635,  25131, 1115,  -121,  6442,  25239, 1193,  -115,  6253,  25342, 1274,  -109,  6066,
+    25440, 1357,  -103,  5882,  25533, 1442,  -98,   5701,  25621, 1531,  -92,   5522,  25704,
+    1622,  -87,   5347,  25780, 1716,  -81,   5174,  25852, 1813,  -76,   5004,  25919, 1912,
+    -72,   4837,  25980, 2015,  -67,   4673,  26035, 2120,  -63,   4512,  26085, 2227,  -58,
+    4354,  26130, 2338,  -54,   4199,  26169, 2451,  -50,   4046,  26202, 2568,  -46,   3897,
+    26230, 2688,  -42,   3751,  26253, 2811,  -38,   3608,  26270, 2936,  -34,   3467,  26281,
+    3064,  -32,   3329,  26287, 3195};
+
+struct InterpolationState {
+    int frac = 0;
```
bunnei: nit: fraction? (prefer verbose/descriptive names to abbreviations)

src/audio_core/algorithm/interpolate.h
bunnei: nit: s16 instead of short? (applies here and elsewhere)

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +22,40 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
+                input[(in_offset + 2) * 2 + 1] * lut[lut_index + 2] +
+                input[(in_offset + 3) * 2 + 1] * lut[lut_index + 3];
 
-    double& pos = state.position;
-    auto& h = state.history;
-    for (std::size_t i = 0; i < num_frames; ++i) {
-        std::rotate(h.begin(), h.end() - 1, h.end());
-        h[0][0] = input[i * 2 + 0];
-        h[0][1] = input[i * 2 + 1];
+        int new_offset = state.frac + step;
```
bunnei: nit: Prefer uniform initialization for numerical types to catch narrowing errors

FernandoS27:
I wonder who really created this algorithm. You should get the root of the original algorithm for better documentation.

legend800:
Confirms it fixes:  https://github.com/yuzu-emu/yuzu/issues/3017

FearlessTobi:
Even after googling around a bit, I couldn't find the original source for this algorithm. @FernandoS27

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +148,42 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
+                      input[(in_offset + 1) * 2 + 0] * lut[lut_index + 1] +
+                      input[(in_offset + 2) * 2 + 0] * lut[lut_index + 2] +
+                      input[(in_offset + 3) * 2 + 0] * lut[lut_index + 3];
+
+        const int r = input[(in_offset + 0) * 2 + 1] * lut[lut_index + 0] +
+                      input[(in_offset + 1) * 2 + 1] * lut[lut_index + 1] +
+                      input[(in_offset + 2) * 2 + 1] * lut[lut_index + 2] +
+                      input[(in_offset + 3) * 2 + 1] * lut[lut_index + 3];
+
+        int new_offset{state.fraction + step};
```
lioncash:
```suggestion
        const int new_offset = state.fraction + step;
```

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +148,42 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
-                r += lanczos_calc * h[j][1];
-            }
-            output.emplace_back(static_cast<s16>(std::clamp(l, -32768.0, 32767.0)));
-            output.emplace_back(static_cast<s16>(std::clamp(r, -32768.0, 32767.0)));
-
-            pos += ratio;
+    std::vector<s16> output(input.size() / ratio);
+
+    const int step = static_cast<int>(ratio * 0x8000);
+    const std::array<s16, 512>& lut = [ratio, step] {
```
lioncash:
```suggestion
    const std::array<s16, 512>& lut = [step] {
```

Since ratio isn't used within the lambda.

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +148,42 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
-            for (std::size_t j = 0; j < h.size(); j++) {
-                const double lanczos_calc = Lanczos(taps, pos + j - taps + 1);
-                l += lanczos_calc * h[j][0];
-                r += lanczos_calc * h[j][1];
-            }
-            output.emplace_back(static_cast<s16>(std::clamp(l, -32768.0, 32767.0)));
-            output.emplace_back(static_cast<s16>(std::clamp(r, -32768.0, 32767.0)));
-
-            pos += ratio;
+    std::vector<s16> output(input.size() / ratio);
```
lioncash: This can now be moved down to above `in_offset`, to place it closer to its usage spot.

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -13,13 +13,131 @@
...
 namespace AudioCore {
 
-/// The Lanczos kernel
-static double Lanczos(std::size_t a, double x) {
-    if (x == 0.0)
-        return 1.0;
-    const double px = M_PI * x;
-    return a * std::sin(px) * std::sin(px / a) / (px * px);
-}
+static constexpr std::array<s16, 512> curve_lut0 = {
```
lioncash: `static` is able to be omitted for these lookup tables (`constexpr` and `const` imply static for file-scope variables).

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +148,42 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
+        }
+        if (step <= 0x8000) {
+            return curve_lut1;
         }
-        pos -= 1.0;
+        return curve_lut2;
+    }();
+
+    int in_offset = 0;
+    for (int out_offset = 0; out_offset < output.size(); out_offset += 2) {
```
lioncash:
```suggestion
    for (std::size_t out_offset = 0; out_offset < output.size(); out_offset += 2) {
```

Otherwise this will cause signedness mismatch warnings.

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +148,42 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
+        int lut_index = (state.fraction >> 8) * 4;
```
lioncash:
```suggestion
        const int lut_index = (state.fraction >> 8) * 4;
```

src/audio_core/algorithm/interpolate.cpp
lioncash: `ratio` should also be static_casted to `std::size_t`

src/audio_core/algorithm/interpolate.cpp
FearlessTobi: I only did this bracket thing originally because bunnei said so. Or did I understand him wrong?

src/audio_core/algorithm/interpolate.cpp
FearlessTobi: ratio is a double between 0 and 1 IIRC. Anyway, I tried this and it throws a divide by zero warning when having the cast here.

src/audio_core/algorithm/interpolate.cpp
lioncash: It's inconsistent with how all the variables elsewhere in the function are initialized.

src/audio_core/algorithm/interpolate.cpp
lioncash: If the implicit conversion to double is actually necessary, then the whole expression should be casted to `std::size_t` (i.e. `static_cast<std::size_t>(input.size() / ratio))`), otherwise this will cause warnings.

src/audio_core/algorithm/interpolate.cpp
degasus: The type is wrong. `std::array<std::array<s16,4>,128>>` is better.

src/audio_core/algorithm/interpolate.cpp
bunnei: Just to be clear -- the current code results in `warning C4244: 'argument': conversion from 'double' to 'const unsigned __int64', possible loss of data`

src/audio_core/algorithm/interpolate.h
```diff
@@ -6,19 +6,12 @@
 
 #include <array>
 #include <vector>
-#include "audio_core/algorithm/filter.h"
```
slashiee: You should also remove the files themselves along with their listing in CMakeLIsts.

bylaws:
The algorithm was reverse engineered from nintendo's CPU side audren implementation.

FearlessTobi:
This PR seems to cause an issue in Crash Team Racing, I need to investigate.

src/audio_core/algorithm/interpolate.cpp
```diff
@@ -30,40 +148,42 @@ std::vector<s16> Interpolate(InterpolationState& state, std::vector<s16> input,
...
+                      input[(in_offset + 2) * 2 + 1] * lut[lut_index + 2] +
+                      input[(in_offset + 3) * 2 + 1] * lut[lut_index + 3];
+
+        int new_offset{state.fraction + step};
+
+        in_offset += new_offset >> 15;
+        state.fraction = new_offset & 0x7fff;
+
+        output[out_offset + 0] =
+            static_cast<s16>(std::clamp(static_cast<double>(l >> 15), -32768.0, 32767.0));
```
lioncash: Actually, now that I look at this again, why is this casting `l >> 15` to a double, instead of just doing:

```cpp
output[out_offset + 0] = static_cast<s16>(std::clamp(l >> 15, -32768, 32767));
```
(ditto for `r`, equivalent)

Given it's already in an integral format at this point?

bunnei:
Untagging EA, as I want to test other audio PRs. This is merged into #3345 for EA testing. Please let me know once current issues are resolved and we can adjust accordingly. Thanks!

bunnei:
@FearlessTobi can you address the issues with this PR soon, so we can get this merged? Thank you.

PR #3314, opened 2020-01-17
core/hle/kernel: Simplify PhysicalMemory usages.

degasus:
This pull requests tries to simplify the interface of PhysicalMemory. PhysicalMemory is currently a std::vector, but it shall be replaced with a custom allocator with mirrored mapping support. So it shall be able to be used for fastmem.

### DIFF
--- a/src/core/hle/kernel/vm_manager.cpp
+++ b/src/core/hle/kernel/vm_manager.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include <algorithm>
+#include <cstring>
 #include <iterator>
 #include <utility>
 #include "common/alignment.h"
@@ -269,18 +270,9 @@ ResultVal<VAddr> VMManager::SetHeapSize(u64 size) {
     // If necessary, expand backing vector to cover new heap extents in
     // the case of allocating. Otherwise, shrink the backing memory,
     // if a smaller heap has been requested.
-    const u64 old_heap_size = GetCurrentHeapSize();
-    if (size > old_heap_size) {
-        const u64 alloc_size = size - old_heap_size;
-
-        heap_memory->insert(heap_memory->end(), alloc_size, 0);
-        RefreshMemoryBlockMappings(heap_memory.get());
-    } else if (size < old_heap_size) {
-        heap_memory->resize(size);
-        heap_memory->shrink_to_fit();
-
-        RefreshMemoryBlockMappings(heap_memory.get());
-    }
+    heap_memory->resize(size);
+    heap_memory->shrink_to_fit();
+    RefreshMemoryBlockMappings(heap_memory.get());
 
     heap_end = heap_region_base + size;
     ASSERT(GetCurrentHeapSize() == heap_memory->size());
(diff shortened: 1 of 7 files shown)


src/core/hle/kernel/vm_manager.cpp
```diff
@@ -752,24 +743,18 @@ void VMManager::MergeAdjacentVMA(VirtualMemoryArea& left, const VirtualMemoryAre
...
         (left.backing_block != right.backing_block || left.offset + left.size != right.offset)) {
-        const auto right_begin = right.backing_block->begin() + right.offset;
-        const auto right_end = right_begin + right.size;
 
         // Check if we can save work.
         if (left.offset == 0 && left.size == left.backing_block->size()) {
             // Fast case: left is an entire backing block.
-            left.backing_block->insert(left.backing_block->end(), right_begin, right_end);
+            left.backing_block->resize(left.size + right.size);
+            memcpy(left.backing_block->data() + left.size, right.backing_block->data() + right.offset, right.size);
```
ReinUsesLisp: `std::memcpy`
`memcpy` is not declared in `<cstring>` by the standard (although all implementations declare it).

src/core/hle/kernel/vm_manager.cpp
```diff
@@ -752,24 +743,18 @@ void VMManager::MergeAdjacentVMA(VirtualMemoryArea& left, const VirtualMemoryAre
...
-            const auto left_size = static_cast<std::size_t>(std::distance(left_begin, left_end));
-            const auto right_size = static_cast<std::size_t>(std::distance(right_begin, right_end));
-
             auto new_memory = std::make_shared<PhysicalMemory>();
-            new_memory->reserve(left_size + right_size);
-            new_memory->insert(new_memory->end(), left_begin, left_end);
-            new_memory->insert(new_memory->end(), right_begin, right_end);
+            new_memory->resize(left.size + right.size);
+            memcpy(new_memory->data(), left.backing_block->data() + left.offset, left.size);
+            memcpy(new_memory->data() + left.size, right.backing_block->data() + right.offset, right.size);
```
ReinUsesLisp: Ditto

src/core/loader/elf.cpp
```diff
@@ -335,7 +335,7 @@ Kernel::CodeSet ElfReader::LoadInto(VAddr vaddr) {
             codeset_segment->addr = segment_addr;
             codeset_segment->size = aligned_size;
 
-            memcpy(&program_image[current_image_position], GetSegmentPtr(i), p->p_filesz);
+            memcpy(program_image.data() + current_image_position, GetSegmentPtr(i), p->p_filesz);
```
ReinUsesLisp: Ditto. It looks like the original code had the same assumption.

src/core/loader/nso.cpp
```diff
@@ -96,8 +96,8 @@ std::optional<VAddr> AppLoader_NSO::LoadModule(Kernel::Process& process,
         if (nso_header.IsSegmentCompressed(i)) {
             data = DecompressSegment(data, nso_header.segments[i]);
         }
-        program_image.resize(nso_header.segments[i].location);
-        program_image.insert(program_image.end(), data.begin(), data.end());
+        program_image.resize(nso_header.segments[i].location + data.size());
+        memcpy(program_image.data() + nso_header.segments[i].location, data.data(), data.size());
```
ReinUsesLisp: Ditto

src/core/hle/kernel/vm_manager.cpp
lioncash: `<cstring>` needs to be included.

src/core/loader/kip.cpp
```diff
@@ -76,8 +76,8 @@ AppLoader::LoadResult AppLoader_KIP::Load(Kernel::Process& process) {
         segment.addr = offset;
         segment.offset = offset;
         segment.size = PageAlignSize(static_cast<u32>(data.size()));
-        program_image.resize(offset);
-        program_image.insert(program_image.end(), data.begin(), data.end());
+        program_image.resize(offset + data.size());
+        memcpy(program_image.data() + offset, data.data(), data.size());
```
lioncash: ditto

src/core/memory.h
```diff
@@ -9,6 +9,7 @@
 #include <string>
 #include "common/common_types.h"
 #include "common/memory_hook.h"
+#include "core/hle/kernel/physical_memory.h"
```
lioncash: This I'm kind of concerned about. It seems kind of... odd to include kernel headers in the kernel-agnostic memory code (at least from an external interface point of view). Though I don't really have a good idea on how to avoid it (aside from moving the type alias of course, but it too is a detail of the kernel =/)

degasus: I'm not sure if you prefer this way: https://gist.github.com/degasus/8721beebfef38eb0d2382f836addba64

lioncash: Yeah, being able to forward declare a type is preferable.

lioncash: Though an anonymous namespace probably shouldn't be used within a header. Aside from that it looks fine.

Issue #1898, opened 2018-12-12
No icon in the headbar on Linux


TotalCaesar659:
Linux nightly builds don't have icon in the headbar, but Windows builds do.
Lubuntu 18.04.1
![1](https://user-images.githubusercontent.com/14265316/49895452-80bdb800-fe61-11e8-8e4a-82d4a1788fb4.png)
Windows 7
![2](https://user-images.githubusercontent.com/14265316/49895459-861b0280-fe61-11e8-86f1-15852e7002bd.png)

DistantThunder:
Shows on latest GIT for me on KDE 5 / Plasma :+1: 

![image](https://user-images.githubusercontent.com/3923810/53670345-b1e8a000-3c7a-11e9-882f-f388f45df080.png)

Hexagon12:
Can the OP confirm that the issue has been fixed?

TotalCaesar659:
Nope, same result on https://github.com/yuzu-emu/yuzu/commit/a9521c983bb66c7aa93c38799752079587aec6f1.

PR #3316, opened 2020-01-17
Add headbar icon on Linux

TotalCaesar659:
Fixes https://github.com/yuzu-emu/yuzu/issues/1898.


bunnei:
Thanks! (and wow, that old icon path is from a long time ago...)

PR #3318, opened 2020-01-18
Remove unused CPU Vendor string and telemtry field

jroweboy:
The information is duplicated in the brand string and the telemetry field is unused.

Already worked with flame sage to make this field optional as well.


chris062689:
@jroweboy Before merging please confirm that telemetry is still flowing to the server properly with this field removed.

PR #3319, opened 2020-01-18
audio_renderer: Process audio data on another thread.

bunnei:


degasus:
This patch just added another thread. However many audio backends already have a thread internally with a callback on gaining new data.

What do you think on still handling the emulation on the main thread: Like when to ask the ARM for more samples.

But at the same time, process the data within the callback on audio backend thread. Like resampling.

However, I have to admit that I have no clue about yuzu's audio processing, so my arguments here are more like guesses :/

FernandoS27:
I'm untagging this until the next thing is resolved:

Audio core has a callback that triggers an Event Signal. Currently Events should only be signalled from the main thread (core emulation thread) are you making sure we do the signalling using core timing events like we do in the GPU?

bunnei:
Untagging due to conflict

PR #3324, opened 2020-01-19
Port citra-emu/citra#5037: "CMake: Create thin archives on Linux"

FearlessTobi:
See citra-emu/citra#5037 for more details. As I don't have a Linux machine with a toolchain installed, I'm just gonna wait for CI to finish.

**Original description**:
This significantly reduces unnecessary disk writes and space usage
when building Citra.

libcore.a is now only ~1MB rather than several hundred megabytes.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -77,6 +77,15 @@ else()
             add_compile_options("-static")
         endif()
     endif()
+
+    if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR MINGW)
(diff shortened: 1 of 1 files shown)


degasus:
Nice patch. Works fine here. Saves me almost a GB on building yuzu.

But WTF, yuzu is 500 MB here (-g)
```
du -sh bin/*
473M	bin/tests
499M	bin/yuzu
470M	bin/yuzu-cmd
467M	bin/yuzu-tester
```

PR #3326, opened 2020-01-19
Port citra-emu/citra#5039: "common/logging: don't use regex for path trimming"

FearlessTobi:
See citra-emu/citra#5039 for more details.

The title of this PR is slightly misleading as we already weren't using a regex anymore for yuzu. 
Either way, this unifies and simplifies the implementation of this feature.

**Original description**:
The regex here is run on the emulation thread, and can cause slowdowns when there is excessive logging. Now the source file path trimming should be done at compile time.
godbolt.org/z/3vqPvB

### DIFF
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -120,7 +120,7 @@ private:
             duration_cast<std::chrono::microseconds>(steady_clock::now() - time_origin);
         entry.log_class = log_class;
         entry.log_level = log_level;
-        entry.filename = Common::TrimSourcePath(filename);
+        entry.filename = filename;
         entry.line_num = line_nr;
         entry.function = function;
         entry.message = std::move(message);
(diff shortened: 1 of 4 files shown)


src/common/logging/backend.cpp
```diff
@@ -120,7 +120,7 @@ class Impl {
             duration_cast<std::chrono::microseconds>(steady_clock::now() - time_origin);
         entry.log_class = log_class;
         entry.log_level = log_level;
-        entry.filename = Common::TrimSourcePath(filename);
+        entry.filename = filename;
```
degasus: This will make a copy within the entry. However the filename is AFAIK always a pointer to a constant string. So may you please replace the type of entry.filename from `std::string` to `const char*`?

FearlessTobi:
Updated.

PR #3337, opened 2020-01-21
yuzu: Implement Vulkan frontend

ReinUsesLisp:
There are some minor commits that I forgot to push before this.

Adds a Qt and SDL2 frontend for Vulkan. It also finishes the missing bits on Vulkan initialization.


YoDevil:
What is needed for this to compile? [linux]

ReinUsesLisp:
A Qt version compiled with Vulkan enabled.

PR #3345, opened 2020-01-25
audio: Synchronize event time with framerate.

bunnei:
This is a pretty simple change that continually re-adjusts audio event time to try to compensate for the current framerate. This should result in real-time audio, even when the game is running slower (or faster) than normal.

This is behind a setting on the Audio tab. For now, this is default on (as internal testing has been very positive), even though this is not necessarily accurate. If we find this causes regressions, we can revisit that decision.

I'm also merging existing audio PRs, so that we can EA test this without conflict. Please ignore those in review.

### DIFF
--- a/src/audio_core/stream.cpp
+++ b/src/audio_core/stream.cpp
@@ -11,8 +11,10 @@
 #include "audio_core/stream.h"
 #include "common/assert.h"
 #include "common/logging/log.h"
+#include "core/core.h"
 #include "core/core_timing.h"
 #include "core/core_timing_util.h"
+#include "core/perf_stats.h"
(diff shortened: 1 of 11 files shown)


EddyHg80:
I was wondering, will this ever come to Citra too?
It would be awesome to unlock the framerate in 3ds games without having distorted audio.
(great work by the way, for everything)

FernandoS27:
In prometheus, audio is real time by default, this shouldn't be merged.

FearlessTobi:
This seems to cause a regression in CTR.

bunnei:
There isn't much point to this anymore, closing.

PR #3346, opened 2020-01-25
bsd: Stub several more functions.

bunnei:
Required for Little Town Hero to boot further. These stubs should be OK, as BSD issues are fairly easy to track down. Furthermore, we are not planning on properly implementing BSD in the near term.

### DIFF
--- a/src/core/hle/service/sockets/bsd.cpp
+++ b/src/core/hle/service/sockets/bsd.cpp
@@ -42,7 +42,47 @@ void BSD::Socket(Kernel::HLERequestContext& ctx) {
     rb.Push<u32>(0); // bsd errno
 }
 
+void BSD::Select(Kernel::HLERequestContext& ctx) {
+    LOG_WARNING(Service, "(STUBBED) called");
+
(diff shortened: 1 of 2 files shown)


bunnei:
@DarkLordZach we shouldn't have long term branches tagged EA/mainline, it's a maintenance burden. I understand your point, but it seems more in the spirit of "all stubs are bad". In this case:
* We have no immediate plans to implement BSD. Because BSD is mostly used for (optional) networking features, it's not high priority.
* To your point about disincentive-izing a proper implementation -- what exactly is our incentive right now to dump time here? If it's booting games where stubs suffice, that is not much incentive. If it's proper network functionality, that's an incentive on its own, regardless of the stubs.
* Various games svcBreak because these functions are not implemented (where stubbing suffices).
* We already partially stub BSD.
* BSD stubs are not really introducing any technical debt -- They are isolated (i.e. they do not bleed into other modules) and they are unlikely to cause difficult to track down bugs. I think once we try to debug networking issues, we will know to start with the stubbed BSD service.

PR #3348, opened 2020-01-25
LOG: Implementation of Level Stubbed

CJBok:
- Stubbed is currently implement on a higher level than Critical.
- Added a filter list ignored_levels (Using a list in case of future expansions).
- Replaced all LOG_* call containing (STUBBED) to LOG_Stubbed.
- Added config setting "log_ignore_stubbed_services".
- Added GUI setting Debug -> Logging -> Ignore stubbed services.

image

Added 2 logs comparing Ignore stubbed services enabled and disabled.
[yuzu_log_ignore_stubbed.txt](https://github.com/yuzu-emu/yuzu/files/4113029/yuzu_log_ignore_stubbed.txt)
[yuzu_log_stubbed.txt](https://github.com/yuzu-emu/yuzu/files/4113030/yuzu_log_stubbed.txt)

### DIFF
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -289,6 +289,7 @@ const char* GetLevelName(Level log_level) {
         LVL(Warning);
         LVL(Error);
         LVL(Critical);
+        LVL(Stubbed);
     case Level::Count:
         break;
     }
--- a/src/common/logging/filter.cpp
+++ b/src/common/logging/filter.cpp
@@ -92,7 +92,18 @@ void Filter::ParseFilterString(std::string_view filter_view) {
     }
 }
 
+void Filter::ClearIgnoredLevels() {
+    ignored_levels.clear();
+}
+
+void Filter::AddIgnoredLevel(Level level) {
+    ignored_levels.push_back(level);
+}
+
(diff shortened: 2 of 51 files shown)


FearlessTobi:
Why does this need a seperate setting? Can't it be part of the log level system?

CJBok:
> Why does this need a seperate setting? Can't it be part of the log level system?

I couldn't come up with a proper position between which levels to put Stubbed. Because putting it too high could filter out other useful information and putting it too low could result in another log spam. Hence the intoduction of ignored levels. Ignored levels doesn't take the levels hiearchy into consideration meaning  you can filter out certain levels without affecting the lower levels.

The setting itself is a bool as there is currently only 1 use-case for it, but can be changed to a list in the future for potentional other use-cases.

lioncash:
> as this masks the issue.

Agreed. A warning is a proper category for a stub anyways, given it makes it obvious it's not ideal and may cause unexpected behavior in games when they show up (complete with yellow-colored text, as is typical for cautionary messages).

A separate category (in light blue) detracts from that and presents them as legitimate messages that can be safely ignored in most cases (they shouldn't) and makes them seem non-cautionary (they aren't).

CJBok:
Log spam it is!

PR #3251, opened 2019-12-28
GUI: Togglable graphics settings buttons in status bar

CJBok:
image

The buttons are not meant for the casual gamer, but more for the supporters / feedback providers whom test a multitude of games with each yuzu update to spot any fixes / bugs between the 2 render APIs.

- When users provide feedback about graphical issues supported by a screenshot (discord) some most frequently asked questions are then already answered inside the screenshot;
- Interactivity is disabled while ingame;
- #ifdef HAS_VULKAN checks will determine if the API button is interactable or not. (PR will also work on EA builds);

PR #3360, opened 2020-01-28
GUI: Togglable graphics settings buttons in status bar

CJBok:
See https://github.com/yuzu-emu/yuzu/pull/3251 for reference.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1910,8 +1910,8 @@ void GMainWindow::OnConfigure() {
     dock_status_button->setChecked(Settings::values.use_docked_mode);
     async_status_button->setChecked(Settings::values.use_asynchronous_gpu_emulation);
 #ifdef HAS_VULKAN
-    renderer_status_button->setChecked(
-        Settings::values.renderer_backend == Settings::RendererBackend::Vulkan);
+    renderer_status_button->setChecked(Settings::values.renderer_backend ==
+                                       Settings::RendererBackend::Vulkan);
 #endif
 }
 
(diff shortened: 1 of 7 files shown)


src/yuzu/main.cpp
```diff
@@ -476,8 +475,73 @@ void GMainWindow::InitializeWidgets() {
...
+    // Setup Dock button
+    dock_status_button = new QPushButton();
+    dock_status_button->setObjectName(QStringLiteral("TogglableStatusBarButton"));
+    dock_status_button->setFocusPolicy(Qt::NoFocus);
+    connect(dock_status_button, &QPushButton::clicked, [&] {
+        Settings::values.use_docked_mode = !Settings::values.use_docked_mode;
+        dock_status_button->setChecked(Settings::values.use_docked_mode);
+        OnDockedModeChanged(!Settings::values.use_docked_mode, Settings::values.use_docked_mode);
+    });
+    dock_status_button->setText(tr("DOCK"));
```
Morph1984: Are these buttons meant to be translatable? They may look bad with other languages, (also not sure how you would translate "OpenGL" or "Vulkan")

CJBok: Technically it shouldn't cause any problems as the buttons will increase in width depending on the character length.
Also I do not know exactly how the translation system will work, I assume the current text will become an identifier / default and hopefully comes with a non embedded language files solution.

Morph1984: I just checked the other labels on the status bar, seems like they too have tr(). So I guess it would be ok. It's also possible that these strings would be left out of the translation if deemed uneccesary.

bunnei:
Thanks!

PR #3361, opened 2020-01-29
yuzu: delete macOS preprocessor directives and code

Zeroksas:
MacOS is no longer supported on Yuzu. This pr removes macOS specific code.


goldenx86:
It's not that yuzu has no more support, it's that it can't be supported _for now_.

PR #3362, opened 2020-01-30
gl_rasterizer: Fix instanced draw arrays

ReinUsesLisp:
`glDrawArrays` was being used when the draw had a base instance specified. This commit removes the draw parameters abstraction and fixes the mentioned issue.

After this, we can start looking into using a simpler interface for draws.


BSoDGamingYT:
Fixes the missing flower rendering in SMO Wooded Kingdom
https://streamable.com/gmjqe

PR #3363, opened 2020-01-30
kernel/physical_core: Make use of std::unique_ptr instead of std::shared_ptr

lioncash:
shared_ptr was used in 2d1984c due to a misunderstanding of how the language generates move constructors and move assignment operators.

If a destructor is user-provided, then the compiler won't generate the move constructor and move assignment operators by default--they must be explicitly opted into.

The reason for the compilation errors is due to the fact that the language will fall back to attempting to use the copy constructor/copy assignment operators if the respective move constructor or move assignment operator is unavailable.

Given that we explicitly opt into them now, the the move constructor and move assignment operators will be generated as expected.

While we're at it, we can do a little bit of tidying up.

### DIFF
--- a/src/core/cpu_manager.cpp
+++ b/src/core/cpu_manager.cpp
@@ -2,14 +2,12 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include "common/assert.h"
 #include "core/arm/exclusive_monitor.h"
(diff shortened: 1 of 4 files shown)


FernandoS27:
Interesting so all I had to do was remove the copy constructor/assignment and define move semantics. Thank you very much for looking into it. You are a live saver!

PR #3151, opened 2019-11-22
system_archive: Fix Korean and Chinese fonts

FearlessTobi:
Fixes Korean fonts when using Open-source system archives.

Fixes https://github.com/yuzu-emu/yuzu/issues/3101.

PR #3365, opened 2020-01-31
Revert "system_archive: Fix Korean and Chinese fonts"

bunnei:
Reverts yuzu-emu/yuzu#3151

Currently is breaking CI due to error:

```##[error]src\core\file_sys\system_archive\data\font_chinese_simplified.cpp(694533,0): Error C1060: compiler is out of heap space```

@FearlessTobi as FYI -- we may need to ship this as a separate binary, break it up into multiple header files, OR update to VS2019 (?)


FearlessTobi:
Just for reference, @DarkLordZach said we can rework this system after the new VFS is merged.

PR #3366, opened 2020-01-31
applets: Fixes for software keyboard and transfer memory.

bunnei:
Supersedes #2710. Fixes:
* Properly reserve and reset TransferMemory region -- this would lead to a hard crash when trying to read the transfer memory.
* Clear events on PopOutData and PopInteractiveOutData.
* Correct a miscounted IPC header in am.
* Add a missing state changed event signal in interactive applet -- fixes a softlock for some games.
* Make wait object wake on any signalled thread, not all -- fixes softlock with applet as sdk code waits on two events that are mutually exculsive. This fixes a bug introduced by #2416.
* Various cleanup.

Thanks to @DarkLordZach for some of the fixes here.

Will tag this one as mainline, to not regress Pokemon in the main release channel.

Fixes pre-entered text in software keyboard:
image

### DIFF
--- a/src/core/hle/service/am/am.h
+++ b/src/core/hle/service/am/am.h
@@ -12,7 +12,8 @@
 
 namespace Kernel {
 class KernelCore;
-}
+class TransferMemory;
+} // namespace Kernel
 
 namespace Service::NVFlinger {
 class NVFlinger;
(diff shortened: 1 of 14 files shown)


src/core/hle/service/am/am.h
```diff
@@ -188,19 +189,35 @@ class ICommonStateGetter final : public ServiceFramework<ICommonStateGetter> {
     std::shared_ptr<AppletMessageQueue> msg_queue;
 };
 
+class IStorageImpl {
```
lioncash: This class should likely have a virtual destructor just to prevent potential undefined behavior in the future related to destruction.

bunnei: Addressed

PR #3377, opened 2020-02-04
vk_rasterizer: Use noexcept methods of std::bitset

ReinUsesLisp:
Removes bounds checking from "texceptions" instances.


goldenx86:
texceptions is illegal use of english, blinkhawk.

PR #3385, opened 2020-02-07
frontend: Add support to batch install files to NAND

### LINKED ISSUE #3383: [Request] Option to batch install files to NAND
For mass-installing DLCs from games like Smash Ultimate, it can quickly become a chore having to install every DLC individually.

RPCS3 recently added a feature like this, along with it's own UI to sort it out and make sure everything is in the correct order.

Morph1984:
Closes #3383

Install Files to NAND

Install Files

Install Files Confirmation

Install Files Confirmation Dark

This extends the implementation of the `Install File to NAND` option to accept and install multiple files in a directory. Furthermore, a confirmation dialog has been added to confirm the installation of said files. This implementation is inspired by RPCS3's recent support for batch PKG install https://github.com/RPCS3/rpcs3/pull/7187.

A new option `Overwrite Existing Files` has been added to this confirmation dialog, allowing the user to choose whether or not to overwrite these files if they exist in the NAND. By default, this option is unticked, which means that already existing files will not be overwritten, and the user will not be prompted about overwriting files. If this option is ticked, it will overwrite all files in the selection.

At the end of installation a "Install Results" dialog will show containing the results of the installation.

### DIFF
--- a/src/yuzu/install_dialog.cpp
+++ b/src/yuzu/install_dialog.cpp
@@ -1,0 +1,72 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <QCheckBox>
+#include <QDialogButtonBox>
+#include <QFileInfo>
+#include <QHBoxLayout>
+#include <QLabel>
+#include <QListWidget>
+#include <QVBoxLayout>
+#include "yuzu/install_dialog.h"
+#include "yuzu/uisettings.h"
+
+InstallDialog::InstallDialog(QWidget* parent, const QStringList& files) : QDialog(parent) {
+    file_list = new QListWidget(this);
+
+    for (const QString& file : files) {
+        QListWidgetItem* item = new QListWidgetItem(QFileInfo(file).fileName(), file_list);
+        item->setData(Qt::UserRole, file);
+        item->setFlags(item->flags() | Qt::ItemIsUserCheckable);
+        item->setCheckState(Qt::Checked);
+    }
+
+    file_list->setMinimumWidth((file_list->sizeHintForColumn(0) * 6) / 5);
+
+    vbox_layout = new QVBoxLayout;
+
+    hbox_layout = new QHBoxLayout;
+
+    description = new QLabel(tr("Please confirm these are the files you wish to install."));
+
+    overwrite_files = new QCheckBox(tr("Overwrite Existing Files"));
+    overwrite_files->setCheckState(Qt::Unchecked);
+
+    buttons = new QDialogButtonBox;
+    buttons->addButton(QDialogButtonBox::Cancel);
+    buttons->addButton(tr("Install"), QDialogButtonBox::AcceptRole);
+
+    connect(buttons, &QDialogButtonBox::accepted, this, &InstallDialog::accept);
+    connect(buttons, &QDialogButtonBox::rejected, this, &InstallDialog::reject);
+
+    hbox_layout->addWidget(overwrite_files);
+    hbox_layout->addWidget(buttons);
+
+    vbox_layout->addWidget(description);
+    vbox_layout->addWidget(file_list);
+    vbox_layout->addLayout(hbox_layout);
+
+    setLayout(vbox_layout);
+    setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+    setWindowTitle(tr("Install Files to NAND"));
+}
+
+InstallDialog::~InstallDialog() = default;
+
+QStringList InstallDialog::GetFilenames() const {
+    QStringList filenames;
+
+    for (int i = 0; i < file_list->count(); ++i) {
+        const QListWidgetItem* item = file_list->item(i);
+        if (item->checkState() == Qt::Checked) {
+            filenames.append(item->data(Qt::UserRole).toString());
+        }
+    }
+
+    return filenames;
+}
+
+bool InstallDialog::ShouldOverwriteFiles() const {
+    return overwrite_files->isChecked();
+}
--- a/src/yuzu/install_dialog.cpp
+++ b/src/yuzu/install_dialog.cpp
@@ -22,7 +22,7 @@ InstallDialog::InstallDialog(QWidget* parent, const QStringList& files) : QDialo
         item->setCheckState(Qt::Checked);
     }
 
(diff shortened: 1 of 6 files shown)


src/yuzu/main.cpp
```diff
@@ -1497,7 +1508,7 @@ void GMainWindow::OnMenuInstallToNAND() {
 
     const auto success = [this]() {
         QMessageBox::information(this, tr("Successfully Installed"),
-                                 tr("The file was successfully installed."));
+                                 tr("The files were successfully installed."));
```
lioncash: This is likely better formatted as:

```cpp
tr("%n file(s) successfully installed", "", count)
```

Where `count` is the number of successfully installed files. This way the translation will automatically adapt whether or not the translation is plural based off the number of successfully installed files (i.e. we won't end up using the plural if only one file is installed).

src/yuzu/main.cpp
```diff
@@ -1518,120 +1530,126 @@ void GMainWindow::OnMenuInstallToNAND() {
...
-            nsp = std::make_shared<FileSys::NSP>(
-                vfs->OpenFile(filename.toStdString(), FileSys::Mode::Read));
-            if (nsp->IsExtractedType())
-                failed();
-        } else {
-            const auto xci = std::make_shared<FileSys::XCI>(
-                vfs->OpenFile(filename.toStdString(), FileSys::Mode::Read));
-            nsp = xci->GetSecurePartitionNSP();
-        }
+    for (const QString& filename : filenames) {
```
lioncash: The code this loop is around should likely be extracted to its own function. This is getting particularly complicated to follow now.

src/yuzu/install_dialog.h
```diff
@@ -0,0 +1,33 @@
...
+
+class InstallDialog : public QDialog {
+    Q_OBJECT
+
+public:
+    explicit InstallDialog(QWidget* parent, const QStringList& files);
+    ~InstallDialog() override;
+
+    QStringList GetFilenames() const;
+    bool OverwriteFiles() const;
```
lioncash:
```suggestion
    bool ShouldOverwriteFiles() const;
```

Since this is only checking if the files should be overwritten, it isn't actually overwriting them.

src/yuzu/install_dialog.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    hboxlayout->addWidget(overwrite_files);
+    hboxlayout->addWidget(buttons);
+
+    vboxlayout->addWidget(description);
+    vboxlayout->addWidget(file_list);
+    vboxlayout->addLayout(hboxlayout);
+
+    setLayout(vboxlayout);
+    setWindowTitle(tr("Install Files to NAND"));
```
lioncash: `setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);`

Should likely be added before this line to disable the `?` button that appears on dialogs by default (given we disable them everywhere else).

Morph1984:
> * If one title fails, should the remaining be attempted or cancelled

Currently, if one title fails, the remaining are cancelled

> * if cancel is clicked, is that for the current title only or not

Is it for the `overwrite` dialog?

```cpp
    const auto overwrite = [this]() {
        return QMessageBox::question(this, tr("Failed to Install"),
                                     tr("The file you are attempting to install already exists "
                                        "in the cache. Would you like to overwrite it?")) ==
               QMessageBox::Yes;
    };
```

If so, the cancel button is only for the current title

> * what if a massive amount of titles are selected

How many titles qualifies as a massive amount?

Morph1984:
> This should:
> 
> * not have overwrite as a checkbox, and instead present it as a yes/no only if necessary.

My intent of this overwrite checkbox is actually for it to **not** overwrite files if they already exist in the NAND, if this checkbox is ticked, it will present dialog boxes for each file that already exist in NAND (I originally intended for it to default to all files, but I am keeping it as-is for now).

If the files already exist in NAND, users likely do not want to overwrite them, and this also saves on clicking "No" on every overwrite dialog, and saves on disk write cycles too.

> * not double check for confirmation and just install all selected (the confirmation is redundant the user already picked them)

As it is currently, it is very cumbersome to individually select files to install, you would have to CTRL + Click each file.
In order to avoid this, most users would either drag and highlight the files they want to install, or click CTRL+A to select all files.
This is why I think this confirmation dialog is necessary - There may be some files in that list that users may not want to install, or they may have second thoughts about the files they want to install.
This is to reduce the chances of mistakes occurring, since currently, they cannot be undone easily, and the only way to revert this is to delete the entire registered folder and start from scratch.


> * provide some sort of progress as to which title is having installed / how many left (rn it will just show progress of current title)

I am planning on adding this one.

Morph1984:
An alternative solution for the overwrite checkbox would be to have an "Apply for all" checkbox in the overwrite dialog

Something like this
image

Kewlan:
> * not double check for confirmation and just install all selected (the confirmation is redundant the user already picked them)

Agreed. I don't know how others do it but I have all the updates in one folder and always use Ctrl+A.

> ...
> This is to reduce the chances of mistakes occurring, since currently, they cannot be undone easily, and the only way to revert this is to delete the entire registered folder and start from scratch.

Having to delete the registered folder is somewhat common anyway since some updates can't be installed on top of old ones, so I think this is a separate issue that should be solved another time.

> An alternative solution for the overwrite checkbox would be to have an "Apply for all" checkbox in the overwrite dialog, and also add additional information regarding the file

I like this. It should also say which file it is, if possible.

> > * If one title fails, should the remaining be attempted or cancelled
> 
> Currently, if one title fails, the remaining are cancelled

I say it should try install all of them, and if at least one has failed it lists which ones.

Also this is how the progress window looks for me:
image
Along with no green bar, the window sometimes grow horizontally when the remaining count goes down.

Morph1984:
@Kewlan I'll update you on the status of this one, with the upcoming VFS rewrite, there will be massive changes to the way we install updates and DLC in the future, these changes will necessitate the removal of the "Install File(s) to NAND" feature" to simplify code.

This feature is said to return some time in the future (although without batch functionality), for accuracy's sake due to the development of a better method of installing updates and DLC (and yes, a feature to easily remove updates is in the works too)

Morph1984:
Before I forget, that white window is an artifact of the updates/dlc installing too fast, there isn't a good way of preventing it as far as I know.

PR #3388, opened 2020-02-08
hle: services: Use std::shared_ptr instead of copy by value.

bunnei:


lioncash:
Commit should probably explain why the change was necessary. I don't have any context for what this resolves

bunnei:
Addressing on check-in

PR #3401, opened 2020-02-12
Set of refactors for Kernel Synchronization and Hardware Constants

FernandoS27:
This PR aims to refactor kernel synchronization mechanisms so they are more similar to the most common used RE rather than citra's old kernel.

It also centralizes all hardware constants into a single place.

### DIFF
--- a/src/core/hardware_properties.h
+++ b/src/core/hardware_properties.h
@@ -8,14 +8,6 @@
 
 namespace Core {
 
-union EmuThreadHandle {
-    u64 raw;
-    struct {
-        u32 host_handle;
-        u32 guest_handle;
-    };
-};
-
 namespace Hardware {
 
 // The below clock rate is based on Switch's clockspeed being widely known as 1.020GHz
@@ -23,6 +15,29 @@ namespace Hardware {
 constexpr u64 BASE_CLOCK_RATE = 1019215872; // Switch cpu frequency is 1020MHz un/docked
 constexpr u64 CNTFREQ = 19200000;           // Switch's hardware clock speed
 constexpr u32 NUM_CPU_CORES = 4;            // Number of CPU Cores
+
 } // namespace Hardware
+
+struct EmuThreadHandle {
+    u32 host_handle;
+    u32 guest_handle;
+
+    u64 GetRaw() const {
+        return (static_cast<u64>(host_handle) << 32) | guest_handle;
+    }
+
+    bool operator==(const EmuThreadHandle& rhs) const {
+        return std::tie(host_handle, guest_handle) == std::tie(rhs.host_handle, rhs.guest_handle);
+    }
+
+    bool operator!=(const EmuThreadHandle& rhs) const {
+        return !operator==(rhs);
+    }
+
+    static constexpr EmuThreadHandle InvalidHandle() {
+        constexpr u32 invalid_handle = 0xFFFFFFFF;
+        return {invalid_handle, invalid_handle};
+    }
+};
 
 } // namespace Core
--- a/src/core/hardware_properties.h
+++ b/src/core/hardware_properties.h
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <tuple>
+
 #include "common/common_types.h"
 
 namespace Core {
--- a/src/core/hardware_properties.h
+++ b/src/core/hardware_properties.h
@@ -1,0 +1,28 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+
+namespace Core {
+
+union EmuThreadHandle {
+    u64 raw;
+    struct {
+        u32 host_handle;
+        u32 guest_handle;
+    };
+};
+
+namespace Hardware {
+
+// The below clock rate is based on Switch's clockspeed being widely known as 1.020GHz
+// The exact value used is of course unverified.
+constexpr u64 BASE_CLOCK_RATE = 1019215872; // Switch cpu frequency is 1020MHz un/docked
+constexpr u64 CNTFREQ = 19200000;           // Switch's hardware clock speed
+constexpr u32 NUM_CPU_CORES = 4;            // Number of CPU Cores
+} // namespace Hardware
+
+} // namespace Core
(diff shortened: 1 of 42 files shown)


src/core/hle/kernel/synchronization.cpp
```diff
@@ -0,0 +1,86 @@
...
+        u32 index = static_cast<s32>(std::distance(sync_objects.begin(), itr));
+        return {RESULT_SUCCESS, index};
+    }
+
+    // No objects were ready to be acquired, prepare to suspend the thread.
+
+    // If a timeout value of 0 was provided, just return the Timeout error code instead of
+    // suspending the thread.
+    if (nano_seconds == 0) {
+        return {RESULT_TIMEOUT, 0};
```
lioncash: We have a [constant](https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/kernel/handle_table.h#L18) to represent an invalid handle, rather than a magic `0` value. Ditto for occurrences below.

src/core/hle/kernel/synchronization.h
```diff
@@ -0,0 +1,34 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <memory>
```
lioncash: `<vector>` needs to be included.

src/core/hle/kernel/synchronization.cpp
```diff
@@ -0,0 +1,86 @@
...
+    thread->SetWaitSynchronizationOutput(static_cast<u32>(index));
+    return true;
+};
+
+Synchronization::Synchronization(Core::System& system) : system{system} {}
+
+void Synchronization::SignalObject(SynchronizationObject& obj) const {
+    if (obj.IsSignaled()) {
+        obj.WakeupAllWaitingThreads();
+    };
```
lioncash:
```suggestion
    }
```

The semicolon isn't necessary here.

src/core/hle/kernel/synchronization.cpp
```diff
@@ -0,0 +1,86 @@
...
+    if (reason == ThreadWakeupReason::Timeout) {
+        thread->SetWaitSynchronizationResult(RESULT_TIMEOUT);
+        return true;
+    }
+
+    ASSERT(reason == ThreadWakeupReason::Signal);
+    thread->SetWaitSynchronizationResult(RESULT_SUCCESS);
+    thread->SetWaitSynchronizationOutput(static_cast<u32>(index));
+    return true;
+};
```
lioncash:
```suggestion
}
```

The semicolon isn't necessary here.

src/core/hardware_properties.h
```diff
@@ -0,0 +1,28 @@
...
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+
+namespace Core {
+
+union EmuThreadHandle {
+    u64 raw;
```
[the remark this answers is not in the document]
FernandoS27: uhm this is not good news, we use this style a lot on the GPU registers.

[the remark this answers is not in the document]
lioncash: It's valid in that case because all entries within the union are the same type.

e.g.

``cpp
union Regs {
    struct {
        u32 a;
        u32 b;
    };
    u32 data[2]; // or std::array<u32, 2>
};
```

which the standard explicitly allows, since its a common initial sequence.

This case and the GPU structs aren't comparable. Again, I'd suggest using a dedicated function for assigning and retrieving the combined values if u64 is absolutely necessary in this case.

File: src/core/hardware_properties.h [hunk 6]
```diff
@@ -8,21 +8,39 @@
...
 constexpr u32 NUM_CPU_CORES = 4;            // Number of CPU Cores
+
 } // namespace Hardware
 
+struct EmuThreadHandle {
+    u32 host_handle;
+    u32 guest_handle;
+
+    u64 GetRaw() const {
+        return (static_cast<u64>(host_handle) << 32ULL) | static_cast<u64>(guest_handle);
```
lioncash:
```suggestion
        return (static_cast<u64>(host_handle) << 32) | guest_handle;
```

src/core/hardware_properties.h
```diff
@@ -8,21 +8,39 @@
...
+    }
+
+    bool operator==(const EmuThreadHandle& rhs) const {
+        std::tie(host_handle, guest_handle) == std::tie(rhs.host_handle, rhs.guest_handle);
```
lioncash:
```suggestion
        return std::tie(host_handle, guest_handle) == std::tie(rhs.host_handle, rhs.guest_handle);
```

src/core/hardware_properties.h
```diff
@@ -8,21 +8,39 @@
...
+    }
+
+    bool operator!=(const EmuThreadHandle& rhs) const {
+        return !operator==(rhs);
+    }
+
+    static EmuThreadHandle InvalidHandle() {
+        constexpr u32 invalid_handle = ~(0U);
```
lioncash:
```suggestion
        constexpr u32 invalid_handle = 0xFFFFFFFF;
```

src/core/hardware_properties.h
```diff
@@ -8,21 +8,39 @@
...
+        EmuThreadHandle result{};
```
lioncash:
```suggestion
        return {invalid_handle, invalid_handle};
```

src/core/hardware_properties.h
```diff
@@ -8,21 +8,39 @@
...
+
+    bool operator==(const EmuThreadHandle& rhs) const {
+        std::tie(host_handle, guest_handle) == std::tie(rhs.host_handle, rhs.guest_handle);
+    }
+
+    bool operator!=(const EmuThreadHandle& rhs) const {
+        return !operator==(rhs);
+    }
+
+    static EmuThreadHandle InvalidHandle() {
```
lioncash:
```suggestion
    static constexpr EmuThreadHandle InvalidHandle() {
```

src/core/hardware_properties.h
```diff
@@ -8,21 +8,36 @@
...
+struct EmuThreadHandle {
+    u32 host_handle;
+    u32 guest_handle;
+
+    u64 GetRaw() const {
+        return (static_cast<u64>(host_handle) << 32) | guest_handle;
+    }
+
+    bool operator==(const EmuThreadHandle& rhs) const {
+        return std::tie(host_handle, guest_handle) == std::tie(rhs.host_handle, rhs.guest_handle);
```
lioncash: Forgot to mention this in the previous review, but `<tuple>` needs to be included for `std::tie`

FernandoS27: Done.

PR #3395, opened 2020-02-10
GPU: Refactor queries implementation and correct GPU Clock.

FernandoS27:
This PR is based on a series of findings by fincs on his project to build an NVN-like API for homebrew. Special thanks to him.

This PR aims to more accurately address queries and time in the GPU.

PR #3409, opened 2020-02-13
query_cache: Implement a query cache and query 21 (samples passed)

ReinUsesLisp:
Queries are a method to write information from the GPU to GPU and CPU visible memory. The most common types of queries are counters, timestamps and fences.

Khronos' APIs (OpenGL and Vulkan) use ranged queries with the following model:
```
BeginQuery();
Draw();
EndQuery();
value = AskValue();
```
Meanwhile NVN and hardware "queries" do this:
```
ResetCounter();
Draw();
value = AskValue();
```
To emulate this behaviour we add up previous queries to the current ones on request.

Asking for a query value is a costly operation, not because of the action of reading the value itself but because they are a blocking call. It doesn't have to wait for all GPU operations, only until the ranged query was ended. To minimize this cost this PR delays the query up it's read by the CPU.

This is done through a generic implementation in the `VideoCommon` namespace and then implemented on both Vulkan an OpenGL. Vulkan now requires `VK_EXT_host_query_reset`, this is widely available and now part of Vulkan 1.2.

Many games use `GL_SAMPLES_PASSED`, but I've only seen Splatoon 2 using it for visible effects.

- Allows to swim on ink on Splatoon 2 (only works with async off).
- Properly implements conditional rendering on Super Mario Odyssey (it shouldn't have visible effects).

This will be rebased after #3395 is merged.

### DIFF
--- a/src/video_core/query_cache.h
+++ b/src/video_core/query_cache.h
@@ -1,0 +1,323 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <algorithm>
+#include <array>
+#include <cstring>
+#include <iterator>
+#include <memory>
+#include <optional>
+#include <unordered_map>
+#include <vector>
+
+#include "common/assert.h"
+#include "core/core.h"
+#include "video_core/engines/maxwell_3d.h"
+#include "video_core/gpu.h"
+#include "video_core/memory_manager.h"
+#include "video_core/rasterizer_interface.h"
+
+namespace VideoCommon {
+
+template <class QueryCache, class HostCounter>
+class CounterStreamBase {
+public:
+    explicit CounterStreamBase(QueryCache& cache, VideoCore::QueryType type)
+        : cache{cache}, type{type} {}
+
+    /// Updates the state of the stream, enabling or disabling as needed.
+    void Update(bool enabled) {
+        if (enabled) {
+            Enable();
+        } else {
+            Disable();
+        }
+    }
+
+    /// Resets the stream to zero. It doesn't disable the query after resetting.
+    void Reset() {
+        if (current) {
+            current->EndQuery();
+
+            // Immediately start a new query to avoid disabling its state.
+            current = cache.Counter(nullptr, type);
+        }
+        last = nullptr;
+    }
+
+    /// Returns the current counter slicing as needed.
+    std::shared_ptr<HostCounter> Current() {
+        if (!current) {
+            return nullptr;
+        }
+        current->EndQuery();
+        last = std::move(current);
+        current = cache.Counter(last, type);
+        return last;
+    }
+
+    /// Returns true when the counter stream is enabled.
+    bool IsEnabled() const {
+        return static_cast<bool>(current);
+    }
+
+private:
+    /// Enables the stream.
+    void Enable() {
+        if (current) {
+            return;
+        }
+        current = cache.Counter(last, type);
+    }
+
+    // Disables the stream.
+    void Disable() {
+        if (current) {
+            current->EndQuery();
+        }
+        last = std::exchange(current, nullptr);
+    }
+
+    QueryCache& cache;
+    const VideoCore::QueryType type;
+
+    std::shared_ptr<HostCounter> current;
+    std::shared_ptr<HostCounter> last;
+};
+
+template <class QueryCache, class CachedQuery, class CounterStream, class HostCounter>
+class QueryCacheBase {
+public:
+    explicit QueryCacheBase(Core::System& system, VideoCore::RasterizerInterface& rasterizer)
+        : system{system}, rasterizer{rasterizer}, streams{{CounterStream{
+                                                      static_cast<QueryCache&>(*this),
+                                                      VideoCore::QueryType::SamplesPassed}}} {}
+
+    void InvalidateRegion(CacheAddr addr, std::size_t size) {
+        FlushAndRemoveRegion(addr, size);
+    }
+
+    void FlushRegion(CacheAddr addr, std::size_t size) {
+        FlushAndRemoveRegion(addr, size);
+    }
+
+    /**
+     * Records a query in GPU mapped memory, potentially marked with a timestamp.
+     * @param gpu_addr  GPU address to flush to when the mapped memory is read.
+     * @param type      Query type, e.g. SamplesPassed.
+     * @param timestamp Timestamp, when empty the flushed query is assumed to be short.
+     */
+    void Query(GPUVAddr gpu_addr, VideoCore::QueryType type, std::optional<u64> timestamp) {
+        auto& memory_manager = system.GPU().MemoryManager();
+        const auto host_ptr = memory_manager.GetPointer(gpu_addr);
+
+        CachedQuery* query = TryGet(ToCacheAddr(host_ptr));
+        if (!query) {
+            const auto cpu_addr = memory_manager.GpuToCpuAddress(gpu_addr);
+            ASSERT_OR_EXECUTE(cpu_addr, return;);
+
+            query = Register(type, *cpu_addr, host_ptr, timestamp.has_value());
+        }
+
+        query->BindCounter(Stream(type).Current(), timestamp);
+    }
+
+    /// Updates counters from GPU state. Expected to be called once per draw, clear or dispatch.
+    void UpdateCounters() {
+        const auto& regs = system.GPU().Maxwell3D().regs;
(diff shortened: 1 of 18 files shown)


src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+        }
+        auto& contents = it->second;
+        const auto found =
+            std::find_if(std::begin(contents), std::end(contents),
+                         [addr](auto& query) { return query.GetCacheAddr() == addr; });
+        return found != std::end(contents) ? &*found : nullptr;
+    }
+
+    static constexpr std::uintptr_t PAGE_SIZE = 4096;
+    static constexpr int PAGE_SHIFT = 12;
```
lioncash: You may want to make this unsigned, since it's only ever used with unsigned values.

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+    std::array<CounterStream, VideoCore::NumQueryTypes> streams;
+};
+
+template <class QueryCache, class HostCounter>
+class HostCounterBase {
+public:
+    explicit HostCounterBase(std::shared_ptr<HostCounter> dependency_)
+        : dependency{std::move(dependency_)}, depth{dependency ? (dependency->Depth() + 1) : 0} {
+        // Avoid nesting too many dependencies to avoid a stack overflow when these are deleted.
+        static constexpr u64 depth_threshold = 96;
```
lioncash:
```suggestion
        constexpr u64 depth_threshold = 96;
```

Given it's just a local primitive value, static isn't really necessary here.

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+            return *result;
+        }
+
+        u64 value = BlockingQuery() + base_result;
+        if (dependency) {
+            value += dependency->Query();
+            dependency = nullptr;
+        }
+
+        return *(result = value);
```
lioncash: You may want to separate the assignment and the return here, since this looks pretty odd at first glance to most.

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+    u64 base_result = 0;                     ///< Equivalent to nested dependencies value.
+};
+
+template <class HostCounter>
+class CachedQueryBase {
+public:
+    explicit CachedQueryBase(VAddr cpu_addr, u8* host_ptr)
+        : cpu_addr{cpu_addr}, host_ptr{host_ptr} {}
+
+    CachedQueryBase(CachedQueryBase&& rhs) noexcept
```
lioncash:
```suggestion
    CachedQueryBase(CachedQueryBase&& rhs) noexcept = default;
```

Performs the same behavior

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+        : cpu_addr{rhs.cpu_addr}, host_ptr{rhs.host_ptr}, counter{std::move(rhs.counter)},
+          timestamp{rhs.timestamp} {}
+
+    CachedQueryBase(const CachedQueryBase&) = delete;
+
+    CachedQueryBase& operator=(CachedQueryBase&& rhs) noexcept {
```
lioncash:
```suggestion
    CachedQueryBase& operator=(CachedQueryBase&& rhs) noexcept = default;
```

Ditto

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+class CachedQueryBase {
+public:
+    explicit CachedQueryBase(VAddr cpu_addr, u8* host_ptr)
+        : cpu_addr{cpu_addr}, host_ptr{host_ptr} {}
+
+    CachedQueryBase(CachedQueryBase&& rhs) noexcept
+        : cpu_addr{rhs.cpu_addr}, host_ptr{rhs.host_ptr}, counter{std::move(rhs.counter)},
+          timestamp{rhs.timestamp} {}
+
+    CachedQueryBase(const CachedQueryBase&) = delete;
```
lioncash: Copy assignment should likely also be deleted.

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+    CachedQueryBase& operator=(CachedQueryBase&& rhs) noexcept {
+        cpu_addr = rhs.cpu_addr;
+        host_ptr = rhs.host_ptr;
+        counter = std::move(rhs.counter);
+        timestamp = rhs.timestamp;
+        return *this;
+    }
+
+    /// Flushes the query to guest memory.
+    virtual void Flush() {
```
lioncash: Given this is intended to be used polymorphically, it'd probably be best to default the virtual destructor just to be safe. Ditto for the previous classes.

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+
+    /// Returns a new host counter.
+    std::shared_ptr<HostCounter> Counter(std::shared_ptr<HostCounter> dependency,
+                                         VideoCore::QueryType type) {
+        return std::make_shared<HostCounter>(static_cast<QueryCache&>(*this), std::move(dependency),
+                                             type);
+    }
+
+    /// Returns the counter stream of the specified type.
+    CounterStream& Stream(VideoCore::QueryType type) {
```
lioncash: This can also have a const equivalent:

```cpp
const CounterStream& Stream(VideoCore::QueryType type) const {
    return streams[static_cast<std::size_t>(type)];
}
```

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+        }
+        current->EndQuery();
+        last = std::move(current);
+        current = cache.Counter(last, type);
+        return last;
+    }
+
+    /// Returns true when the counter stream is enabled.
+    bool IsEnabled() const {
+        return static_cast<bool>(current);
```
lioncash:
```suggestion
        return current != nullptr;
```

src/video_core/renderer_opengl/gl_query_cache.h
```diff
@@ -0,0 +1,76 @@
...
+    OGLQuery query;
+};
+
+class CachedQuery final : public VideoCommon::CachedQueryBase<HostCounter> {
+public:
+    explicit CachedQuery(QueryCache& cache, VideoCore::QueryType type, VAddr cpu_addr,
+                         u8* host_ptr);
+    CachedQuery(CachedQuery&& rhs) noexcept;
+
+    CachedQuery& operator=(CachedQuery&& rhs) noexcept;
```
lioncash: The copy constructor and copy assignment operators should also be defaulted or deleted to make their behavior explicit in the interface.

src/video_core/query_cache.h
```diff
@@ -0,0 +1,359 @@
...
+
+    CacheAddr GetCacheAddr() const noexcept {
+        return ToCacheAddr(host_ptr);
+    }
+
+    u64 SizeInBytes() const noexcept {
+        return SizeInBytes(timestamp.has_value());
+    }
+
+    static u64 SizeInBytes(bool with_timestamp) {
```
lioncash: This can be `noexcept` as well as constexpr if that matters

src/video_core/renderer_vulkan/vk_query_cache.cpp
```diff
@@ -0,0 +1,122 @@
...
+    query_pools[static_cast<std::size_t>(type)].Reserve(query);
+}
+
+HostCounter::HostCounter(VKQueryCache& cache, std::shared_ptr<HostCounter> dependency,
+                         VideoCore::QueryType type)
+    : VideoCommon::HostCounterBase<VKQueryCache, HostCounter>{std::move(dependency)}, cache{cache},
+      type{type}, query{cache.AllocateQuery(type)}, ticks{cache.Scheduler().Ticks()} {
+    const auto dev = cache.Device().GetLogical();
+    cache.Scheduler().Record([dev, query = query](vk::CommandBuffer cmdbuf, auto& dld) {
+        dev.resetQueryPoolEXT(query.first, query.second, 1, dld);
```
asLody: cmd.resetQueryPool(query.first, query.second, 1, dld);

asLody:
You should minimize the use of extensions, why not use cmd.resetQueryPool?

src/video_core/renderer_vulkan/vk_query_cache.cpp
ReinUsesLisp: That was my original intension, but resetting a query pool requires being outside of a renderpass. IMO it's not worth ending a renderpass just to reset a query. Using the extension allows us to reset it from the CPU without worrying about our current scope.

PR #3404, opened 2020-02-12
GUI: add option aspect ratio

raven02:
Add new option aspect ratio 
Untitled

PR #3412, opened 2020-02-14
GUI: Add aspect ratio dropdown

### LINKED ISSUE #2427: [Request] Stretch to full screen
Override aspect ratio and stretches the image to the full display area

Morph1984:
Supersedes #3404 and closes #2427, credit to @Margen67 for discovering the relevant line modification. Special thanks to @raven02 for the initial PR(s).

image

16:9 (Default)
yuzu_1uQtTbDNxB

21:9
M34DwDy5oL

Stretch to Window
Pkm5xQU1lo

### DIFF
--- a/src/core/frontend/framebuffer_layout.cpp
+++ b/src/core/frontend/framebuffer_layout.cpp
@@ -27,22 +27,9 @@ FramebufferLayout DefaultFrameLayout(u32 width, u32 height) {
     // so just calculate them both even if the other isn't showing.
     FramebufferLayout res{width, height};
 
-    const auto window_aspect_ratio = static_cast<float>(height) / width;
-    float emulation_aspect_ratio;
-
-    switch (static_cast<Aspect>(Settings::values.aspect_ratio)) {
-    case Aspect::AspectDefault:
-        emulation_aspect_ratio = static_cast<float>(ScreenUndocked::Height) / ScreenUndocked::Width;
-        break;
-    case Aspect::Aspect21by9:
-        emulation_aspect_ratio = 9.f / 21;
-        break;
-    case Aspect::AspectStretch:
-        emulation_aspect_ratio = window_aspect_ratio;
-        break;
-    default:
-        emulation_aspect_ratio = static_cast<float>(ScreenUndocked::Height) / ScreenUndocked::Width;
-    }
+    const float window_aspect_ratio = static_cast<float>(height) / width;
+    float emulation_aspect_ratio = EmulationAspectRatio(
+        static_cast<Aspect>(Settings::values.aspect_ratio), window_aspect_ratio);
 
     const Common::Rectangle<u32> screen_window_area{0, 0, width, height};
     Common::Rectangle<u32> screen = MaxRectangle(screen_window_area, emulation_aspect_ratio);
@@ -69,6 +56,19 @@ FramebufferLayout FrameLayoutFromResolutionScale(u32 res_scale) {
     }
 
     return DefaultFrameLayout(width, height);
+}
+
+float EmulationAspectRatio(Aspect aspect, float window_aspect_ratio) {
+    switch (aspect) {
+    case Aspect::Default:
+        return static_cast<float>(ScreenUndocked::Height) / ScreenUndocked::Width;
+    case Aspect::Aspect21by9:
+        return 9.0f / 21.0f;
+    case Aspect::StretchToWindow:
+        return window_aspect_ratio;
+    default:
+        return static_cast<float>(ScreenUndocked::Height) / ScreenUndocked::Width;
+    }
 }
 
 } // namespace Layout
--- a/src/core/frontend/framebuffer_layout.cpp
+++ b/src/core/frontend/framebuffer_layout.cpp
@@ -30,17 +30,17 @@ FramebufferLayout DefaultFrameLayout(u32 width, u32 height) {
     const auto window_aspect_ratio = static_cast<float>(height) / width;
     float emulation_aspect_ratio;
 
-    switch (Settings::values.aspect_ratio) {
-    case 0: // 16:9 (Default)
+    switch (static_cast<Aspect>(Settings::values.aspect_ratio)) {
+    case Aspect::AspectDefault:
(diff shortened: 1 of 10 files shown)


src/core/frontend/framebuffer_layout.cpp
```diff
@@ -27,9 +27,22 @@ FramebufferLayout DefaultFrameLayout(u32 width, u32 height) {
     // so just calculate them both even if the other isn't showing.
     FramebufferLayout res{width, height};
 
-    const float emulation_aspect_ratio{static_cast<float>(ScreenUndocked::Height) /
-                                       ScreenUndocked::Width};
     const auto window_aspect_ratio = static_cast<float>(height) / width;
+    float emulation_aspect_ratio;
+
+    switch (Settings::values.aspect_ratio) {
+    case 0: // 16:9 (Default)
```
DarkLordZach: please use an enumeration instead of comments&magic numbers
you can do something like:
`namespace EnumName {
    enum Impl {
        Values, ...
    }
}
`

This code snippet combines the name prefixing of `enum class`(es) with the int-cast-ability of raw enums.

src/core/frontend/framebuffer_layout.cpp
```diff
@@ -27,9 +27,22 @@ FramebufferLayout DefaultFrameLayout(u32 width, u32 height) {
     // so just calculate them both even if the other isn't showing.
     FramebufferLayout res{width, height};
 
-    const float emulation_aspect_ratio{static_cast<float>(ScreenUndocked::Height) /
-                                       ScreenUndocked::Width};
     const auto window_aspect_ratio = static_cast<float>(height) / width;
+    float emulation_aspect_ratio;
```
lioncash: This block can be extracted to its own function and be called here.

src/core/frontend/framebuffer_layout.cpp
```diff
@@ -27,9 +27,22 @@ FramebufferLayout DefaultFrameLayout(u32 width, u32 height) {
...
+
+    switch (static_cast<Aspect>(Settings::values.aspect_ratio)) {
+    case Aspect::AspectDefault:
+        emulation_aspect_ratio = static_cast<float>(ScreenUndocked::Height) / ScreenUndocked::Width;
+        break;
+    case Aspect::Aspect21by9:
+        emulation_aspect_ratio = 9.f / 21;
```
lioncash:
```suggestion
        emulation_aspect_ratio = 9.0f / 21.0f;
```

src/core/frontend/framebuffer_layout.h
```diff
@@ -18,6 +18,12 @@ enum ScreenDocked : u32 {
     HeightDocked = 1080,
 };
 
+enum class Aspect {
+    AspectDefault,
```
lioncash:
```suggestion
    Default,
```

The `Aspect` prefix is only really necessary for `21by9`, since C++ doesn't allow identifiers to begin with a number.

src/core/frontend/framebuffer_layout.h
```diff
@@ -18,6 +18,12 @@ enum ScreenDocked : u32 {
...
+    Aspect21by9,
+    AspectStretch,
```
lioncash:
```suggestion
    StretchToWindow,
```

src/core/frontend/framebuffer_layout.cpp
```diff
@@ -27,22 +27,9 @@ FramebufferLayout DefaultFrameLayout(u32 width, u32 height) {
...
-        emulation_aspect_ratio = 9.f / 21;
-        break;
-    case Aspect::AspectStretch:
-        emulation_aspect_ratio = window_aspect_ratio;
-        break;
-    default:
-        emulation_aspect_ratio = static_cast<float>(ScreenUndocked::Height) / ScreenUndocked::Width;
-    }
+    const float window_aspect_ratio = static_cast<float>(height) / width;
+    float emulation_aspect_ratio = EmulationAspectRatio(
```
lioncash:
```suggestion
    const float emulation_aspect_ratio = EmulationAspectRatio(
```

DarkLordZach: Also should be auto -- EmulationAspectRatio returns float

src/core/frontend/framebuffer_layout.h
```diff
@@ -18,6 +18,12 @@ enum ScreenDocked : u32 {
     HeightDocked = 1080,
 };
 
+enum class Aspect {
```
DarkLordZach: Rename to AspectRatio for readability

src/core/frontend/framebuffer_layout.h
```diff
@@ -18,6 +18,12 @@ enum ScreenDocked : u32 {
...
+    Default,
+    Aspect21by9,
```
DarkLordZach: `R21_9` or something similar --  dont need to use words

src/core/frontend/framebuffer_layout.cpp
lioncash: There is no difference here, either one works. Using the type directly is arguably better for direct readability, since you don't need to check what the function is actually returning to understand what behavior the code here is intended to have.

Auto is good for cases when the type can be known directly within the same scope, or for overly-long type names (such as iterators). Using it everywhere possible just because you can isn't advisable

src/core/frontend/framebuffer_layout.cpp
```diff
@@ -28,8 +28,8 @@ FramebufferLayout DefaultFrameLayout(u32 width, u32 height) {
     FramebufferLayout res{width, height};
 
     const float window_aspect_ratio = static_cast<float>(height) / width;
-    float emulation_aspect_ratio = EmulationAspectRatio(
-        static_cast<Aspect>(Settings::values.aspect_ratio), window_aspect_ratio);
+    const float emulation_aspect_ratio = EmulationAspectRatio(
```
DarkLordZach:
```suggestion
    const auto emulation_aspect_ratio = EmulationAspectRatio(
```

PR #3413, opened 2020-02-14
frontend: Set minimum window size to 640x360 instead of 1280x720

Morph1984:
Previously the minimum window size was 1280x720, causing some users with monitors/screens with a height of less than 800 pixels (taskbar hidden) or 840 pixels (taskbar visible) have the status bar cut off **and not be able to resize the window below 720p**. This amends the issue by reducing the minimum window size to 640x360.

To those who are confused what this does, and does not do:

1. It **does not** cause the window to run at 360p by default
2. This is just allows users to resize the window below 720p to a minimum of 360p where previously you were not able to,

Before (1280x720):

yuzu_jdIhxNuuO6
yuzu_CgCDWG6dlG

After (640x360):

yuzu_lPRg3ZiXT7
yuzu_djYBOBkb4J

### DIFF
--- a/src/core/frontend/emu_window.cpp
+++ b/src/core/frontend/emu_window.cpp
@@ -46,7 +46,7 @@ private:
 EmuWindow::EmuWindow() {
     // TODO: Find a better place to set this.
     config.min_client_area_size =
-        std::make_pair(Layout::ScreenUndocked::Width / 2, Layout::ScreenUndocked::Height / 2);
+        std::make_pair(Layout::MinimumSize::MinimumWidth, Layout::MinimumSize::MinimumHeight);
     active_config = config;
     touch_state = std::make_shared<TouchState>();
     Input::RegisterFactory<Input::TouchDevice>("emu_window", touch_state);
--- a/src/core/frontend/emu_window.cpp
+++ b/src/core/frontend/emu_window.cpp
@@ -46,7 +46,7 @@ private:
 EmuWindow::EmuWindow() {
     // TODO: Find a better place to set this.
     config.min_client_area_size =
-        std::make_pair(Layout::MinimumSize::MinimumWidth, Layout::MinimumSize::MinimumHeight);
+        std::make_pair(Layout::MinimumSize::Width, Layout::MinimumSize::Height);
     active_config = config;
     touch_state = std::make_shared<TouchState>();
     Input::RegisterFactory<Input::TouchDevice>("emu_window", touch_state);
--- a/src/core/frontend/emu_window.cpp
+++ b/src/core/frontend/emu_window.cpp
@@ -46,7 +46,7 @@ private:
 EmuWindow::EmuWindow() {
     // TODO: Find a better place to set this.
     config.min_client_area_size =
-        std::make_pair(Layout::ScreenUndocked::Width, Layout::ScreenUndocked::Height);
+        std::make_pair(Layout::ScreenUndocked::Width / 2, Layout::ScreenUndocked::Height / 2);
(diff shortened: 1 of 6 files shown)


Ghostotakoi:
i think the default resolution should remain 720p but in the graphics tab the user can have the option to switch to 360p if want to.

Morph1984:
@DarkLordZach @Ghostotakoi The default window size *should* still remain at 720p, this just allows users to resize the window below 720p. It would not default to launch with this smaller window if that's your concern.

DarkLordZach:
Yes, but for a normal user resizing the window below 720p is a degraded experience.
This should only be an emergency failsafe in the case of really wonky screen resolutions to prevent the status bar from being cut off.

Morph1984:
Switch hardware is able to output 480p to TV, so we should allow users to resize the window below the handheld LCD resolution.

Morph1984:
A normal user wont resize the window below 720p, they will keep it as is. Given that they are unlikely to resize the window there is no harm in having the option to resize the window.

DarkLordZach:
`A normal user wont resize the window below 720p, they will keep it as is.`
   You can't make this claim. Some users will absolutely do this.
`Given that they are unlikely to resize the window there is no harm in having the option to resize the window.`
    There is, a degraded user experience.
`Besides, most users will play with a maximized window or fullscreen for the best experience.`
    Again, you can't really make this claim.

src/core/frontend/emu_window.cpp
```diff
@@ -46,7 +46,7 @@ class EmuWindow::TouchState : public Input::Factory<Input::TouchDevice>,
 EmuWindow::EmuWindow() {
     // TODO: Find a better place to set this.
     config.min_client_area_size =
-        std::make_pair(Layout::ScreenUndocked::Width, Layout::ScreenUndocked::Height);
+        std::make_pair(Layout::ScreenUndocked::Width / 2, Layout::ScreenUndocked::Height / 2);
```
bunnei: These should be constants defined in src/core/frontend/ somewhere

src/yuzu/loading_screen.cpp
```diff
@@ -73,7 +73,7 @@ LoadingScreen::LoadingScreen(QWidget* parent)
     : QWidget(parent), ui(std::make_unique<Ui::LoadingScreen>()),
       previous_stage(VideoCore::LoadCallbackStage::Complete) {
     ui->setupUi(this);
-    setMinimumSize(1280, 720);
+    setMinimumSize(640, 360);
```
bunnei: These should be (the same) constants defined in src/core/frontend/ somewhere

bunnei:
To add one to what we discussed via Discord -- I do tend to agree that it's a bit annoying for any users on 1080p (or above). For example, I resize my window all the time, but then often want it back to the native/accurate 720p display resolution (handheld mode) -- which it's super convenient to just resize the window to the minimum. In this sense, it's also useful for 1:1 pixel emulation accuracy. 

With your change, it does not seem like there is a way to restore the native/accurate size if I accidentally (or intentionally) resize the window. I'd be OK with this change if we add settings (probably to View menu) to reset the window to 720p (handheld native size) and maybe 1080p (docked native size). I think this is an acceptable solution, especially since most other popular emulators also allow arbitrary window sizes.

Morph1984:
Latest commit adds an option to reset the window size

image

Depending on the aspect ratio selected, the window will resize to the following resolutions:

Default / Stretch to Window = 1280x720
4:3 = 960x720
21:9 = 1680x720

PR #3429, opened 2020-02-17
Fix CMake build errors for certain SDL2 installations

### LINKED ISSUE #3426: Build fails due to missing SDL.h
Building fails after commit https://github.com/yuzu-emu/yuzu/commit/68043dd2337feab0b0ccc9868cd535b2254335ea, even though I have SDL2 installed, and the find_package(SDL2 REQUIRED) in `CMakeLists.txt:159` sets `SDL2_FOUND` to `1`

`gcc` build, tried it on a new clone, fails on every commit after the one mentioned above, same error.

brianclinkenbeard:
Previous commit led to a regression for Arch Linux users. This patch should fix that issue. See [here](https://discourse.libsdl.org/t/arch-linux-cmake-find-package-sdl2-required-passes-but-doesnt-find-anything/24226/2) and [here](https://github.com/tsoding/nothing/issues/923#issuecomment-506370650) for more information. Fixes #3426.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,8 +157,14 @@ if (ENABLE_SDL2)
         target_include_directories(SDL2 INTERFACE "${SDL2_INCLUDE_DIR}")
     else()
         find_package(SDL2 REQUIRED)
-        include_directories(${SDL2_INCLUDE_DIRS})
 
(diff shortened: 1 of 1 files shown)


bunnei:
Thank you!

PR #3431, opened 2020-02-18
InputCommon: analog_from_button get direction implementation

CJBok:


bunnei:
Thanks for the quick fix here -- tagging as EA so we can get more testing.

Issue #3048, opened 2019-10-30
Segmentation fault on Linux


Sunderland93:
Both binaries from Download section and builded from source are not even start. Yuzu from Download section:

```

Starting program: /media/DATA/yuzu-linux-20191026-b1065b017/yuzu 

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7fe08ed in _dl_relocate_object (scope=0x7ffff7ffe4f0, reloc_mode=<optimized out>, 
    consider_profiling=consider_profiling@entry=0) at dl-reloc.c:231
231     dl-reloc.c: Нет такого файла или каталога.
```

Yuzu builded from source code:

```
Starting program: /usr/bin/yuzu 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
__GI___pthread_rwlock_wrlock (rwlock=0x0) at pthread_rwlock_wrlock.c:27
27      pthread_rwlock_wrlock.c: Нет такого файла или каталога.

```
My OS is Debian 10, but same error in Ubuntu 18.04

baconwaifu:
seems to be `glibc` related. are your headers the same version as the system libraries? the only reason I can think of for this to occur with a source-built version is a mismatch between the headers and installed libraries.

altp0:
I have the same problem, in Debian 10 too, but only get the second messages for both provided binaries (3e66f3100) and binary built myself.

And I had to add `#include <boost/range/iterator_range.hpp>` in src/video_core/rasterizer_accelerated.h in order to make build work. Else it says ‘make_iterator_range’ is not a member of ‘boost’.

Is yuzu incompatible with Debian 10, as I read it was only tested for Ubuntu 18.04?

```
(gdb) run
Starting program: /home/user/apps/yuzu/build/bin/yuzu
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
__GI___pthread_rwlock_wrlock (rwlock=0x0) at pthread_rwlock_wrlock.c:27
27      pthread_rwlock_wrlock.c: No such file or directory.
(gdb) bt
#0  __GI___pthread_rwlock_wrlock (rwlock=0x0) at pthread_rwlock_wrlock.c:27
[...]

PR #3432, opened 2020-02-18
Update httplib to 0.5.5

brianclinkenbeard:
Some linux users (#3048) report segfaults using the mainline builds that seem to trace back to older versions of httplib being [incompatible with OpenSSL 1.1](https://github.com/yhirose/cpp-httplib/pull/243). For me, this only occurred on the provided binaries and not on built images, so without a DevOps build, I can't confirm that this fixes the issue.

It's worth noting the latest version of cpp-httplib is 0.5.5, however v0.5 introduces breaking changes that prevented yuzu from building. Anyone more familiar is obviously welcome to rectify the code that uses httplib to work with they latest version if they want.

### DIFF
--- a/externals/httplib/README.md
+++ b/externals/httplib/README.md
@@ -1,4 +1,4 @@
-From https://github.com/yhirose/cpp-httplib/commit/d9479bc0b12e8a1e8bce2d34da4feeef488581f3
+From https://github.com/yhirose/cpp-httplib/tree/v0.4.2
 
 MIT License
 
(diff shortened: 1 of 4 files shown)


brianclinkenbeard:
Testing the artifacts, it looks like this doesn't solve the issue so I will do some more digging.

bunnei:
@brianclinkenbeard does the update to 0.5.5 resolve this issue?

brianclinkenbeard:
@bunnei Sadly no. The issue is probably with glibc. You can close this or merge it if you want, the latest version should offer greater compatibility and might fix some bugs.

bunnei:
Hitting CI errors with this: https://gist.github.com/bunnei/9c09b2853e8a3b5c7c66777d496192ca

Do you mind taking a look? Thanks!

brianclinkenbeard:
Does this work?

bunnei:
Looks good -- thanks!

PR #3433, opened 2020-02-18
renderer_vulkan: Add the rest of case for TryConvertBorderColor

namkazt:
Follow this implement on RPCS3: https://github.com/RPCS3/rpcs3/blob/90f4023cb8682ed6d5ca2feab4c91c5024697f0c/rpcs3/Emu/RSX/VK/VKFormats.cpp#L101

### DIFF
--- a/src/video_core/renderer_vulkan/vk_sampler_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_sampler_cache.cpp
@@ -26,6 +26,13 @@ static std::optional<vk::BorderColor> TryConvertBorderColor(std::array<float, 4>
         if (color[0] + color[1] + color[2] > 1.35f) {
             // If color elements are brighter than roughly 0.5 average, use white border
             return vk::BorderColor::eFloatOpaqueWhite;
+        }
+        if (color[3] > 0.5f) {
+            return vk::BorderColor::eFloatOpaqueBlack;
+        }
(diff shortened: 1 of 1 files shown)


FernandoS27:
@ReinUsesLisp

src/video_core/renderer_vulkan/vk_sampler_cache.cpp
```diff
@@ -23,7 +23,16 @@ static std::optional<vk::BorderColor> TryConvertBorderColor(std::array<float, 4>
     } else if (color == std::array<float, 4>{1, 1, 1, 1}) {
         return vk::BorderColor::eFloatOpaqueWhite;
     } else {
-        return {};
+        if (color[0] + color[1] + color[2] > 1.35f) {
```
lioncash: This can be unindented as follows:

```suggestion
        if (color[0] + color[1] + color[2] > 1.35f) {
            // If color elements are brighter than roughly 0.5 average, use white border
            return vk::BorderColor::eFloatOpaqueWhite;
        }
        if (color[3] > 0.5f) {
            return vk::BorderColor::eFloatOpaqueBlack;
        }
        return vk::BorderColor::eFloatTransparentBlack;
```

namkazt: yeah, right :p

Issue #3289, opened 2020-01-09
Kirby Star Allies crash upon rendering the title screen using Vulkan


scoofz:
Hello there, 

Kirby star allies works with Opengl but completely crashes on the game menu using Vulkan, I have an AMD card so i'd really appreciate Vulkan support on this :) . I have no idea how to provide a log, so if you can point me in the right direction, I'd gladly provide one. 

Thank you

Hexagon12:
This is an issue tracker for developers and not a support forum.
Please use our Discord for support requests.

scoofz:
And? 

I'm not asking for support, I know what to look for in terms of compatibility, other users reported the same thing on YT and reddit, I'm reporting a problem that anyone can easily reproduce that's all. 

I was asking for direction for logs in case you wanted one. 

I also don't demand for the game to work, I know I'm a patreon for an experimental emulator, I don't expect it to work perfectly with every game...

Yvar192:
This tells you how to get a log: https://yuzu-emu.org/help/reference/log-files/

And yeah, opening issues for actual issues is what they're supposed for, aren't they

scoofz:
Thank you Yvar192 

So here is the full log of the crash : https://pastebin.com/KmFYGhEi

We can see at the end when Yuzu crashes : 

[ 163.929131] Debug <Critical> video_core\shader\decode\conversion.cpp:DecodeConversion:93: Assertion Failed! Unimplemented code!
[ 163.929885] Render.Vulkan <Info> video_core\renderer_vulkan\vk_pipeline_cache.cpp:GetGraphicsPipeline:224: Compile 0x76A9E15588B44898
[ 163.931094] Service.Audio <Warning> core\hle\service\audio\audren_u.cpp:RequestUpdateImpl:95: (STUBBED) called
[ 163.932146] Service.Audio <Warning> core\hle\service\audio\audren_u.cpp:RequestUpdateImpl:95: (STUBBED) called
[ 163.935612] Debug <Critical> video_core\renderer_vulkan\maxwell_to_vk.cpp:VertexFormat:398: Assertion Failed! Unimplemented vertex format of type=4 and size=1

Hexagon12:
I'm sorry for closing this issue. I see that we currently have other Vulkan related issues open.

Reopening this issue...

theboy181:
Close?

ReinUsesLisp:
Fixed by #3435. Closing.

PR #3434, opened 2020-02-18
vk_shader: Implement ImageLoad

namkazt:
Used by Kirby Star Allies.
Issue: #3289 

Thanks @ReinUsesLisp for helping me out.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_device.cpp
+++ b/src/video_core/renderer_vulkan/vk_device.cpp
@@ -106,6 +106,8 @@ bool VKDevice::Create(const vk::DispatchLoaderDynamic& dldi, vk::Instance instan
     features.tessellationShader = true;
     features.fragmentStoresAndAtomics = true;
     features.shaderImageGatherExtended = true;
+    features.shaderStorageImageReadWithoutFormat =
+        is_shader_storage_img_read_without_format_supported;
     features.shaderStorageImageWriteWithoutFormat = true;
     features.textureCompressionASTC_LDR = is_optimal_astc_supported;
 
@@ -457,6 +459,8 @@ void VKDevice::SetupFamilies(const vk::DispatchLoaderDynamic& dldi, vk::SurfaceK
 
 void VKDevice::SetupFeatures(const vk::DispatchLoaderDynamic& dldi) {
     const auto supported_features{physical.getFeatures(dldi)};
+    is_shader_storage_img_read_without_format_supported =
+        supported_features.shaderStorageImageReadWithoutFormat;
     is_optimal_astc_supported = IsOptimalAstcSupported(supported_features, dldi);
 }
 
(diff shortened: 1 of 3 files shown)


theboy181:
any other games?

namkazt:
> any other games?

I'm not sure. Only test with kirby

lioncash:
Commit message itself should probably indicate that Kirby Star Allies uses it, so that information is stored in the repository

namkazt:
> Commit message itself should probably indicate that Kirby Star Allies uses it, so that information is stored in the repository

@lioncash anyway how can i change the commit message in Github.

lioncash:
Use `git commit --amend` and then alter the message, then force push.

namkazt:
> Use `git commit --amend` and then alter the message, then force push.

ok so it must be use on local. Thanks.

ReinUsesLisp:
@theboy181 Xenoblade 2 also uses SULD, but it uses a variant that's not currently implemented (SULD.D instead of SULD.P).

namkazt:
> Aside from that, we also need to stub this call (by returning `Immediate(0.0f)`) when the device is not capable of loading images without an explicit format. In other words, if the Vulkan feature `shaderStorageImageReadWithoutFormat` is false. Feel free to ask me on Discord or here if you need help loading this in `vk_device`.

Added check for `shaderStorageImageReadWithoutFormat`

src/video_core/renderer_vulkan/vk_shader_decompiler.cpp
```diff
@@ -274,6 +274,7 @@ class SPIRVDecompiler final : public Sirit::Module {
         AddCapability(spv::Capability::ImageBuffer);
         AddCapability(spv::Capability::ImageGatherExtended);
         AddCapability(spv::Capability::SampledBuffer);
+        AddCapability(spv::Capability::StorageImageReadWithoutFormat);
```
ReinUsesLisp: Only add this one when the extension is available.

src/video_core/renderer_vulkan/vk_device.cpp
```diff
@@ -530,6 +534,7 @@ std::unordered_map<vk::Format, vk::FormatProperties> VKDevice::GetFormatProperti
                                         vk::Format::eBc6HUfloatBlock,
                                         vk::Format::eBc6HSfloatBlock,
                                         vk::Format::eBc1RgbaSrgbBlock,
+                                        vk::Format::eBc2SrgbBlock,
```
ReinUsesLisp: Is this a left over from another branch?

ReinUsesLisp:
Just one nit, other than that, it looks good to me.

src/video_core/renderer_vulkan/vk_device.cpp
namkazt: oh right

PR #3438, opened 2020-02-19
video_core: memory_manager: Flush/invalidate asynchronously when possible.

bunnei:
`Invalidate` (on memory write) and `FlushAndInvalidate` (on unmap memory) can be asynchronous.

This replaces #3386, which needed to be reverted due to a regression with Dead Cells (because the `Flush` still needs to be synchronous).

Results in a minor performance improvement, as measured by @Morph1984 .

### DIFF
--- a/src/video_core/memory_manager.cpp
+++ b/src/video_core/memory_manager.cpp
@@ -9,6 +9,7 @@
 #include "core/hle/kernel/process.h"
 #include "core/hle/kernel/vm_manager.h"
 #include "core/memory.h"
+#include "video_core/gpu.h"
 #include "video_core/memory_manager.h"
 #include "video_core/rasterizer_interface.h"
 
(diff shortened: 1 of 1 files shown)


src/video_core/memory_manager.cpp
```diff
@@ -292,7 +297,8 @@ void MemoryManager::WriteBlock(GPUVAddr dest_addr, const void* src_buffer, const
         switch (page_table.attributes[page_index]) {
         case Common::PageType::Memory: {
             u8* dest_ptr{page_table.pointers[page_index] + page_offset};
-            rasterizer.InvalidateRegion(ToCacheAddr(dest_ptr), copy_amount);
+            // Invalidate through the GPU interface, to be asynchronous if possible.
+            system.GPU().InvalidateRegion(ToCacheAddr(dest_ptr), copy_amount);
```
FernandoS27: This should always be sync.

PR #3443, opened 2020-02-22
audio_core: interpolate: Pad input with zeros, the algorithm reads past on final frame.

bunnei:
- Fixes audio crackling in Crash Team Racing Nitro-Fueled.
- Addresses followup issues from #3310.


bunnei:
Superseded by #3448

PR #3458, opened 2020-02-24
cubeb_sink: Don't discard other channels when performing downmixing

FearlessTobi:
Previously, when performing downmixing, we would discard all channels except the left and right one.
This implementation respects them when mixing down to Stereo.
It is taken from this document: http://www.atsc.org/wp-content/uploads/2015/03/A52-201212-17.pdf.

Fixes Luigis Mansion 3 cutscene and Bayonetta audio + Sonic Forces audio.
[**Before**](https://streamable.com/o5a3k) and [**After**](https://streamable.com/uf5f6) (thanks to @slashiee).

### DIFF
--- a/src/audio_core/cubeb_sink.cpp
+++ b/src/audio_core/cubeb_sink.cpp
@@ -8,6 +8,7 @@
 #include "audio_core/cubeb_sink.h"
 #include "audio_core/stream.h"
 #include "audio_core/time_stretch.h"
+#include "common/assert.h"
 #include "common/logging/log.h"
 #include "common/ring_buffer.h"
 #include "core/settings.h"
@@ -65,12 +66,25 @@ public:
     void EnqueueSamples(u32 source_num_channels, const std::vector<s16>& samples) override {
         if (source_num_channels > num_channels) {
             // Downsample 6 channels to 2
+            ASSERT_MSG(source_num_channels == 6, "Channel count must be 6");
+
             std::vector<s16> buf;
             buf.reserve(samples.size() * num_channels / source_num_channels);
             for (std::size_t i = 0; i < samples.size(); i += source_num_channels) {
-                for (std::size_t ch = 0; ch < num_channels; ch++) {
-                    buf.push_back(samples[i + ch]);
-                }
+                // Downmixing implementation taken from the ATSC standard
+                const s16 left{samples[i + 0]};
+                const s16 right{samples[i + 1]};
+                const s16 center{samples[i + 2]};
+                const s16 surround_left{samples[i + 4]};
+                const s16 surround_right{samples[i + 5]};
+                // Not used in the ATSC reference implementation
+                [[maybe_unused]] const s16 low_frequency_effects { samples[i + 3] };
+
+                constexpr s32 clev{707}; // center mixing level coefficient
+                constexpr s32 slev{707}; // surround mixing level coefficient
+
+                buf.push_back(left + (clev * center / 1000) + (slev * surround_left / 1000));
+                buf.push_back(right + (clev * center / 1000) + (slev * surround_right / 1000));
             }
             queue.Push(buf);
             return;


src/audio_core/cubeb_sink.cpp
```diff
@@ -65,12 +66,24 @@ class CubebSinkStream final : public SinkStream {
...
+                const s16 right = samples[i + 1];
+                const s16 center = samples[i + 2];
+                [[maybe_unused]] const s16 low_frequency_effects = samples[i + 3];
+                const s16 sourround_left = samples[i + 4];
+                const s16 sourround_right = samples[i + 5];
+
+                const float clev = 0.707; // center mixing level coefficient
+                const float slev = 0.707; // surround mixing level coefficient
+
+                buf.push_back(1.0 * left + clev * center + slev * sourround_left);
```
bunnei: `buf.push_back(static_cast<s16>(left + (707 * static_cast<s32>(center) / 1000) + (707 * static_cast<s32>(sourround_left ) / 1000)); ?`

(seems like we can use fixed point math here, should be faster)

src/audio_core/cubeb_sink.cpp
```diff
@@ -65,12 +66,24 @@ class CubebSinkStream final : public SinkStream {
...
             for (std::size_t i = 0; i < samples.size(); i += source_num_channels) {
-                for (std::size_t ch = 0; ch < num_channels; ch++) {
-                    buf.push_back(samples[i + ch]);
-                }
+                // Downmixing implementation taken from the ASTC standard
+                const s16 left = samples[i + 0];
+                const s16 right = samples[i + 1];
+                const s16 center = samples[i + 2];
+                [[maybe_unused]] const s16 low_frequency_effects = samples[i + 3];
+                const s16 sourround_left = samples[i + 4];
```
bunnei: `surround_left` ?

src/audio_core/cubeb_sink.cpp
```diff
@@ -65,12 +66,24 @@ class CubebSinkStream final : public SinkStream {
...
+            ASSERT_MSG(source_num_channels == 6, "Channel count must be 6");
+
             std::vector<s16> buf;
             buf.reserve(samples.size() * num_channels / source_num_channels);
             for (std::size_t i = 0; i < samples.size(); i += source_num_channels) {
-                for (std::size_t ch = 0; ch < num_channels; ch++) {
-                    buf.push_back(samples[i + ch]);
-                }
+                // Downmixing implementation taken from the ASTC standard
+                const s16 left = samples[i + 0];
```
bunnei: Prefer uniform initialization with numerics to prevent narrowing mistakes

src/audio_core/cubeb_sink.cpp
```diff
@@ -65,12 +66,24 @@ class CubebSinkStream final : public SinkStream {
...
+                const s16 center = samples[i + 2];
+                [[maybe_unused]] const s16 low_frequency_effects = samples[i + 3];
+                const s16 sourround_left = samples[i + 4];
+                const s16 sourround_right = samples[i + 5];
+
+                const float clev = 0.707; // center mixing level coefficient
+                const float slev = 0.707; // surround mixing level coefficient
+
+                buf.push_back(1.0 * left + clev * center + slev * sourround_left);
+                buf.push_back(1.0 * right + clev * center + slev * sourround_right);
```
bunnei: ditto

src/audio_core/cubeb_sink.cpp
bunnei: (feel free to leave those defined as constants, and use s32 to prevent the inner casts)

src/audio_core/cubeb_sink.cpp
```diff
@@ -65,12 +66,24 @@ class CubebSinkStream final : public SinkStream {
...
             buf.reserve(samples.size() * num_channels / source_num_channels);
             for (std::size_t i = 0; i < samples.size(); i += source_num_channels) {
-                for (std::size_t ch = 0; ch < num_channels; ch++) {
-                    buf.push_back(samples[i + ch]);
-                }
+                // Downmixing implementation taken from the ASTC standard
+                const s16 left = samples[i + 0];
+                const s16 right = samples[i + 1];
+                const s16 center = samples[i + 2];
+                [[maybe_unused]] const s16 low_frequency_effects = samples[i + 3];
```
lioncash: You may want to leave a comment here to explain why this is unused.

bunnei:
@FearlessTobi ping to make the changes here -- they are fairly minor. Thanks!

src/audio_core/cubeb_sink.cpp
```diff
@@ -65,12 +66,25 @@ class CubebSinkStream final : public SinkStream {
...
             // Downsample 6 channels to 2
+            ASSERT_MSG(source_num_channels == 6, "Channel count must be 6");
+
             std::vector<s16> buf;
             buf.reserve(samples.size() * num_channels / source_num_channels);
             for (std::size_t i = 0; i < samples.size(); i += source_num_channels) {
-                for (std::size_t ch = 0; ch < num_channels; ch++) {
-                    buf.push_back(samples[i + ch]);
-                }
+                // Downmixing implementation taken from the ASTC standard
```
Morph1984: ATSC

PR #3469, opened 2020-02-28
shader_decode: Fix LD, LDG when track constant buffer

namkazt:
It fixed opcode LD, LDG on Pokemon Sword that can't find the constant buffer. Not sure if it helps any on visual.


lioncash:
Commit title should probably be improved. I don't understand what it's actually trying to say.

namkazt:
> Commit title should probably be improved. I don't understand what it's actually trying to say.

Sorry. My english is not good :(. I will try to find better words for that commit message

lioncash:
Looks nicer! The commit itself should probably have its title updated as well.

PR #3471, opened 2020-02-29
core/qt/sdl: Remove dangerous API URL setting

DarkLordZach:
This setting allows users to change the URL of the backend core API (responsible for gamedb, telemetry, testcases, amongst others). This is extemely dangerous to have as a configuration setting because a malicious actor could distribute custom configuration packages with a custom API URL, and use this setup to collect user tokens, which can be used to impersonate users on the core API, which is catastrophic.

This flaw was recently discovered in a backend meeting.

### DIFF
--- a/src/core/settings.h
+++ b/src/core/settings.h
@@ -468,7 +468,6 @@ struct Values {
 
     // WebService
     bool enable_telemetry;
-    std::string web_api_url;
     std::string yuzu_username;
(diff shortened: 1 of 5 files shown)


bunnei:
Closing this PR because it is ancient. I don't have a strong opinion on this change, someone can reopen this if they do.

PR #3473, opened 2020-02-29
gl_shader_cache: Rework shader cache and store texture arrays

ReinUsesLisp:
Instead of pre-specializing shaders and then post-specializing them, drop the later and only "specialize" the shader while decoding it. This reduces the complexity of the shader cache significantly.

While we are changing code that invalidates the disk shader cache, do some maintenance to the code related to it.

* This invalidates the disk shader cache.
* This fixes reflections in LM3 when the shader cache is enabled.


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -56,6 +57,25 @@ using TextureIR = std::variant<TextureOffset, TextureDerivates, TextureArgument>
 constexpr u32 MAX_CONSTBUFFER_ELEMENTS =
     static_cast<u32>(Maxwell::MaxConstBufferSize) / (4 * sizeof(float));
 
+std::string_view CommonDeclarations = R"(#define ftoi floatBitsToInt
```
lioncash: This can be constexpr

src/video_core/shader/registry.h
```diff
@@ -82,22 +105,33 @@ class ConstBufferLocker {
         return bound_buffer;
     }
 
+    /// Returns compute information from this shader
+    const GraphicsInfo& GetGraphicsInfo() const {
```
lioncash: These should probably be defined in the cpp file, or have `common/assert.h` included.

ReinUsesLisp: The commentary is also wrong.

BSoDGamingYT:
This PR also fixed graphical corruption when moving room to room in Luigi's Mansion 3 on OpenGL

Issue #2852, opened 2019-09-10
Action Guide in Super Mario Odyssey is Broken


ReaperOfSouls1909:
![image](https://user-images.githubusercontent.com/18043258/64643329-7deaf600-d3e6-11e9-909b-2bfb8a4af39f.png)
image of Softlock after action guide is called to open.

if any other info is required give me 2 hours

<meta name="titleids"  content="0100000000010000">

DarkLordZach:
Can you track down which canary started this?

Miguejun:
Same problem in latest versions of yuzu

MysticExile:
This issue appears to have been (mostly) fixed with #3477 if the softlock still occurs please open a new issue, preferably with a log too.

ConBil16:
Hey! I am currently experiencing this problem and I don't know how to do the steps described in #3477 below I've attached my log although I am not entirely sure what that mean.
[yuzu_log.txt.old.txt](https://github.com/yuzu-emu/yuzu/files/6120503/yuzu_log.txt.old.txt)

PR #3477, opened 2020-03-01
core/web_browser: Allow WebApplet to exit gracefully when an error occurs

FearlessTobi:
Currently, yuzu just freezes when an error occurs while Initializing the WebApplet.
From a user perspective, this obviously isn't great as the game just softlocks.
With this change, yuzu will call the Finalize method, so to the game it seems like as the user just exited the WebApplet normally.

This works around https://github.com/yuzu-emu/yuzu/issues/2852.

### DIFF
--- a/src/core/hle/service/am/applets/web_browser.cpp
+++ b/src/core/hle/service/am/applets/web_browser.cpp
@@ -254,6 +254,12 @@ void WebBrowser::Execute() {
 
     if (status != RESULT_SUCCESS) {
         complete = true;
+
+        // This is a workaround in order not to softlock yuzu when an error happens during the
+        // webapplet init. In order to avoid an svcBreak, the status is set to RESULT_SUCCESS
+        Finalize();
+        status = RESULT_SUCCESS;
+
         return;
     }
 


goldenx86:
DLZ, what other alternative do we have?
This bug affects many.

BSoDGamingYT:
> This is not hardware accurate behavior. If the browser fails to init on hardware, it wont just continue execution as if it succeeded.

While this is true, is having a fallback like this not desirable?

ReaperOfSouls1909:
This issue effects me every time I can’t get around it unless I use a save so adding this will be a great idea until the issue is fixed if it can be fixed

src/core/hle/service/am/applets/web_browser.cpp
```diff
@@ -254,6 +254,9 @@ void WebBrowser::Execute() {
 
     if (status != RESULT_SUCCESS) {
         complete = true;
+        Finalize();
+        // Set the status to RESULT_SUCCESS, so the application doesn't svcBreak when returning
```
bunnei: Can you update the comment to reflect that this is a workaround?

bunnei:
@FearlessTobi reminder to address comments

src/core/hle/service/am/applets/web_browser.cpp
FearlessTobi: Comment addressed.

PR #3478, opened 2020-03-02
Refactoring to boot A32 games

bunnei:
**NOTE: No 32-bit games will run with these changes alone!**

These are the very initial changes to boot A32 games. With these changes we:
* Remove error checks on loading 32-bit games
* Refactor ARM interface to support 32-bit and 64-bit JIT backends
* Implement a 32-bit ARM interface, based on Citra's armv6 Dynarmic backend
* Implement several 32-bit SVCs

With these changes, Mario Kart 8 Deluxe will load, boot, and initialize a few services. We quickly get blocked on unimplemented CPU instructions.

With regards to missing ARM functionality -- you'll need to implement the following instructions in Dynarmic for these changes to work: `LDAB`, `LDAEX`, `STLB`, `STLEX`. After this, armv7 VFP needs to be implemented -- the first instruction we are blocked on is `VMOV`.

Credit to @FearlessTobi for several changes here.

### DIFF
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -25,7 +25,20 @@ public:
     explicit ARM_Interface(System& system_) : system{system_} {}
     virtual ~ARM_Interface() = default;
 
-    struct ThreadContext {
+    struct ThreadContext32 {
+        std::array<u32, 16> cpu_registers;
+        u32 cpsr;
+        std::array<u8, 4> padding;
+        std::array<u64, 32> fprs;
+        u32 fpscr;
+        u32 fpexc;
+        u32 tpidr;
+    };
+    // Internally within the kernel, it expects the AArch32 version of the
+    // thread context to be 344 bytes in size.
+    static_assert(sizeof(ThreadContext32) == 0x158);
+
+    struct ThreadContext64 {
         std::array<u64, 31> cpu_registers;
         u64 sp;
         u64 pc;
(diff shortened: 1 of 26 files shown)


src/core/arm/dynarmic/arm_dynarmic_cp15.h
```diff
@@ -0,0 +1,152 @@
...
+
+#pragma once
+
+#include <memory>
+#include <optional>
+
+#include <dynarmic/A32/coprocessor.h>
+#include "common/common_types.h"
+
+enum CP15Register {
```
lioncash:
```suggestion
enum class CP15Register {
```

src/core/arm/dynarmic/arm_dynarmic_cp15.h
```diff
@@ -0,0 +1,152 @@
...
+    // Not an actual register.
+    // All registers should be defined above this.
+    CP15_REGISTER_COUNT,
+};
+
+class DynarmicCP15 final : public Dynarmic::A32::Coprocessor {
+public:
+    using CoprocReg = Dynarmic::A32::CoprocReg;
+
+    DynarmicCP15(u32* cp15) : CP15(cp15){};
```
lioncash:
```suggestion
    explicit DynarmicCP15(u32* cp15) : CP15(cp15){};
```

src/core/arm/arm_interface.h
```diff
@@ -38,7 +38,13 @@ class ARM_Interface : NonCopyable {
     };
     // Internally within the kernel, it expects the AArch64 version of the
     // thread context to be 800 bytes in size.
-    static_assert(sizeof(ThreadContext) == 0x320);
+    static_assert(sizeof(ThreadContext64) == 0x320);
+
+    struct ThreadContext32 {
```
lioncash: I have a feeling this struct is larger than this (e.g. it probably also includes the fpcr and fpsr as well, otherwise floating point status registers wouldn't be able to be restored) when restoring contexts.

PR #3489, opened 2020-03-09
shader: implement SULD.D bits32/64

namkazt:
bits 32 is done . bits64 SINT is not finish yet.
Fixed kirby black water

Before:
2020-04-05 19_59_09-Window
After:
2020-04-06 12_59_20-Window

@ReinUsesLisp please check on this.

### DIFF
--- a/src/video_core/shader/decode/image.cpp
+++ b/src/video_core/shader/decode/image.cpp
@@ -80,10 +80,34 @@ u32 ShaderIR::DecodeImage(NodeBlock& bb, u32 pc) {
 
             switch (instr.suldst.GetStoreDataLayout()) {
             case StoreType::Bits32: {
-                MetaImage meta{image, {}, {}};
-                Node value = Operation(OperationCode::ImageLoad, meta, GetCoordinates(type));
-                SetTemporary(bb, 0, std::move(value));
-                SetRegister(bb, instr.gpr0.Value(), GetTemporary(0));
+                Node value{};
+                for (s32 i = 3; i >= 0; i--) {
+                    MetaImage meta{image, {}, i};
+                    Node element_value =
+                        Operation(OperationCode::ImageLoad, meta, GetCoordinates(type));
+
+                    const Node comp = GetPredicateComparisonFloat(PredCondition::GreaterEqual,
+                                                                  element_value, Immediate(1.0f));
+                    const Node mul =
+                        Operation(OperationCode::Select, comp, Immediate(1.f), Immediate(255.f));
+
+                    Node element = Operation(OperationCode::FMul, NO_PRECISE, element_value, mul);
+                    element = SignedOperation(OperationCode::ICastFloat, true, NO_PRECISE,
+                                              std::move(element));
+                    element = Operation(OperationCode::ULogicalShiftLeft, std::move(element),
+                                        Immediate(8 * i));
+                    if (i == 3) {
+                        //(namkazt) for now i'm force it to 0 at alpha component if color is in
+                        // range (0-255)
+                        value = Operation(OperationCode::Select, comp, Immediate(0),
+                                          std::move(element));
+                    } else {
+                        value = Operation(OperationCode::UBitwiseOr, value,
+                                          Operation(OperationCode::Select, comp,
+                                                    std::move(element_value), std::move(element)));
+                    }
+                }
+                SetRegister(bb, instr.gpr0.Value(), std::move(value));
                 break;
             }
             default:
--- a/src/video_core/shader/decode/image.cpp
+++ b/src/video_core/shader/decode/image.cpp
@@ -18,8 +18,8 @@ namespace VideoCommon::Shader {
 
 using Tegra::Shader::Instruction;
 using Tegra::Shader::OpCode;
-using Tegra::Shader::StoreType;
 using Tegra::Shader::PredCondition;
+using Tegra::Shader::StoreType;
 
 namespace {
 std::size_t GetImageTypeNumCoordinates(Tegra::Shader::ImageType image_type) {
--- a/src/video_core/shader/decode/image.cpp
+++ b/src/video_core/shader/decode/image.cpp
@@ -18,6 +18,7 @@ namespace VideoCommon::Shader {
 
 using Tegra::Shader::Instruction;
 using Tegra::Shader::OpCode;
+using Tegra::Shader::StoreType;
 
 namespace {
 std::size_t GetImageTypeNumCoordinates(Tegra::Shader::ImageType image_type) {
--- a/src/video_core/shader/decode/image.cpp
+++ b/src/video_core/shader/decode/image.cpp
@@ -53,7 +53,6 @@ u32 ShaderIR::DecodeImage(NodeBlock& bb, u32 pc) {
 
     switch (opcode->get().GetId()) {
     case OpCode::Id::SULD: {
-        UNIMPLEMENTED_IF(instr.suldst.mode != Tegra::Shader::SurfaceDataMode::P);
         UNIMPLEMENTED_IF(instr.suldst.out_of_bounds_store !=
                          Tegra::Shader::OutOfBoundsStore::Ignore);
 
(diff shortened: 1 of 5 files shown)


src/video_core/shader/decode/image.cpp
```diff
@@ -19,6 +19,7 @@ namespace VideoCommon::Shader {
 using Tegra::Shader::Instruction;
 using Tegra::Shader::OpCode;
 using Tegra::Shader::StoreType;
+using Tegra::Shader::PredCondition;
```
lioncash: This should be placed above `StoreType`, since we sort them alphabetically

src/video_core/shader/decode/image.cpp
```diff
@@ -62,17 +63,58 @@ u32 ShaderIR::DecodeImage(NodeBlock& bb, u32 pc) {
...
+                    const Node mul =
+                        Operation(OperationCode::Select, comp, Immediate(1.f), Immediate(255.f));
+
+                    Node element = Operation(OperationCode::FMul, NO_PRECISE, element_value, mul);
+                    element = SignedOperation(OperationCode::ICastFloat, true, NO_PRECISE,
+                                              std::move(element));
+                    element = Operation(OperationCode::ULogicalShiftLeft, std::move(element),
+                                        Immediate(8 * i));
+                    if (i == 3) {
+                        //(namkazt) for now i'm force it to 0 at alpha component if color is in
```
lioncash: This should probably explain why it's being forced to zero for the time being so others have context

src/video_core/shader/decode/image.cpp
```diff
@@ -62,17 +63,58 @@ u32 ShaderIR::DecodeImage(NodeBlock& bb, u32 pc) {
...
+                SetRegister(bb, instr.gpr0.Value() + i, GetTemporary(i));
+            }
+        } else if (instr.suldst.mode == Tegra::Shader::SurfaceDataMode::D_BA) {
+            UNIMPLEMENTED_IF(instr.suldst.GetStoreDataLayout() != StoreType::Bits32);
+
+            switch (instr.suldst.GetStoreDataLayout()) {
+            case StoreType::Bits32: {
+                Node value{};
+                for (s32 i = 3; i >= 0; i--) {
+                    MetaImage meta{image, {}, i};
```
lioncash:
```suggestion
                    MetaImage meta{image, {}, static_cast<u32>(i)};
```

Otherwise this will cause sign-conversion warnings.

src/video_core/shader/decode/image.cpp
namkazt: are you sure ? i got clang format at that point.
2020-03-12 01_11_42-Pipelines - Run 20200311 4 logs

bunnei:
@ReinUsesLisp can you take a look at this one? Thanks!

namkazt:
> @ReinUsesLisp can you take a look at this one? Thanks!

I think it should not be merged at the current time.

ReinUsesLisp:
Yes, this needs to query the format from the current Like namkazt said, it shouldn't be merged as is.

namkazt:
@ReinUsesLisp @bunnei  please review. It is possible to merge now.

src/video_core/shader/decode/image.cpp
```diff
@@ -62,17 +295,93 @@ u32 ShaderIR::DecodeImage(NodeBlock& bb, u32 pc) {
...
+                Node value = Operation(OperationCode::ImageLoad, meta, GetCoordinates(type));
+                SetTemporary(bb, indexer++, std::move(value));
+            }
+            for (u32 i = 0; i < indexer; ++i) {
+                SetRegister(bb, instr.gpr0.Value() + i, GetTemporary(i));
+            }
+        } else if (instr.suldst.mode == Tegra::Shader::SurfaceDataMode::D_BA) {
+            UNIMPLEMENTED_IF(instr.suldst.GetStoreDataLayout() != StoreType::Bits32);
+
+            const auto maxwell3d = &Core::System::GetInstance().GPU().Maxwell3D();
```
ReinUsesLisp: No, we can't do this. This breaks on offline decodes and it breaks when the shader cache is being built. This has to be queried from the registry and stored in disk.

src/video_core/shader/decode/image.cpp
```diff
@@ -10,16 +10,250 @@
...
+    case TextureFormat::R16_G16_B16_A16:
+    case TextureFormat::R32_G32_B32_A32:
+    case TextureFormat::R32_G32_B32:
+    case TextureFormat::R32_G32:
+    case TextureFormat::R16_G16:
+    case TextureFormat::R32:
+    case TextureFormat::R16:
+    case TextureFormat::R8:
+    case TextureFormat::R1:
+        if (0 == component) {
```
lioncash: Our convention is `thing == <constant>`, not `<constant> == thing`. Ditto for all occurrences below.

src/video_core/shader/decode/image.cpp
lioncash: The global system accessor should also not be used in new code.

bunnei:
@namkazt can you address the followup comments? Thanks!

namkazt:
> @namkazt can you address the followup comments? Thanks!

@bunnei will do this weekend. ;)

namkazt:
@bunnei it is ready now!

src/video_core/shader/decode/image.cpp
```diff
@@ -271,6 +271,40 @@ std::size_t GetImageTypeNumCoordinates(Tegra::Shader::ImageType image_type) {
 }
 } // Anonymous namespace
 
+Node ShaderIR::GetComponentValue(ComponentType component_type, u32 component_size,
```
lioncash, line 274: This should likely return a `std::pair<Node, bool>` instead of using an out parameter.

src/video_core/shader/shader_ir.h
```diff
@@ -312,6 +312,10 @@ class ShaderIR final {
     /// Conditionally saturates a half float pair
     Node GetSaturatedHalfFloat(Node value, bool saturate = true);
 
+    /// Get image component value by type and size
+    Node GetComponentValue(Tegra::Texture::ComponentType component_type, u32 component_size,
+                           const Node original_value, bool* is_signed);
```
lioncash, line 317:
```suggestion
                           Node original_value, bool* is_signed);
```

src/video_core/shader/decode/image.cpp
namkazt: okey

src/video_core/shader/decode/image.cpp
```diff
@@ -13,13 +13,247 @@
...
+}
+
+u32 GetComponentSize(TextureFormat format, std::size_t component) {
+    switch (format) {
+    case TextureFormat::R32_G32_B32_A32:
+        return 32;
+    case TextureFormat::R16_G16_B16_A16:
+        return 16;
+    case TextureFormat::R32_G32_B32:
+        return (component == 0 || component == 1 || component == 2) ? 32 : 0;
```
ReinUsesLisp, line 144:
```suggestion
        return component <= 2 ? 32 : 0;
```

src/video_core/shader/decode/image.cpp
```diff
@@ -13,13 +13,247 @@
...
+    case TextureFormat::R32_G32:
+        return (component == 0 || component == 1) ? 32 : 0;
```
ReinUsesLisp, line 146:
```suggestion
        return component <= 1 ? 32 : 0;
```

src/video_core/shader/decode/image.cpp
```diff
@@ -13,13 +13,247 @@
...
+    case TextureFormat::R16_G16:
+        return (component == 0 || component == 1) ? 16 : 0;
```
ReinUsesLisp, line 148:
```suggestion
        return component <= 1 ? 16 : 0;
```

src/video_core/shader/decode/image.cpp
```diff
@@ -13,13 +13,247 @@
...
+    case TextureFormat::R32:
+        return (component == 0) ? 32 : 0;
```
ReinUsesLisp, line 150:
```suggestion
        return component == 0 ? 32 : 0;
```

src/video_core/shader/decode/image.cpp
```diff
@@ -13,13 +13,247 @@
...
+    case TextureFormat::R16:
+        return (component == 0) ? 16 : 0;
```
ReinUsesLisp, line 152:
```suggestion
        return component == 0 ? 16 : 0;
```

src/video_core/shader/decode/image.cpp
```diff
@@ -13,13 +13,247 @@
...
+    case TextureFormat::R8:
+        return (component == 0) ? 8 : 0;
```
ReinUsesLisp, line 154:
```suggestion
        return component == 0 ? 8 : 0;
```

src/video_core/shader/decode/image.cpp
```diff
@@ -13,13 +13,247 @@
...
+    case TextureFormat::R1:
+        return (component == 0) ? 1 : 0;
```
ReinUsesLisp, line 156:
```suggestion
        return component == 0 ? 1 : 0;
```

src/video_core/shader/decode/image.cpp
```diff
@@ -37,6 +271,39 @@ std::size_t GetImageTypeNumCoordinates(Tegra::Shader::ImageType image_type) {
...
+        return {SignedOperation(OperationCode::ICastFloat, is_signed, std::move(cnv_value)),
+                is_signed};
+    }
+    case ComponentType::UINT: // range [0, (1 << component_size) - 1]
+        return {original_value, false};
+    case ComponentType::FLOAT:
+        if (component_size == 16) {
+            return {Operation(OperationCode::HCastFloat, original_value), true};
+        } else {
+            return {original_value, true};
```
ReinUsesLisp, line 299:
```suggestion
            return {std::move(original_value), true};
```
`original_value` has to be explicitly moved here (unlike with normal return values) because it's being used in a constructor. `std::make_pair(original_value, true)` will copy, `std::make_pair(std::move(original_value), true)` won't.

src/video_core/shader/decode/image.cpp
```diff
@@ -37,6 +271,39 @@ std::size_t GetImageTypeNumCoordinates(Tegra::Shader::ImageType image_type) {
...
+        }
+    default:
+        UNIMPLEMENTED_MSG("Unimplement component type={}", component_type);
+        return {original_value, true};
```
ReinUsesLisp, line 303:
```suggestion
        return {std::move(original_value), true};
```

PR #3490, opened 2020-03-09
video_core: Initial implementation of transform feedbacks

ReinUsesLisp:
This PR implements transform feedbacks for its most common usage. This won't work if games do anything of the following:
- Feedback more than one draw.
- Draw feedbacks (`glDrawTransformFeedback`).
- Use transform feedback queries in a meaningful way.
- Stream to multiple buffers in a geometry shader.
- Output a different topology than what `glDraw*` emits.
- Use a non-zero output buffer on Vulkan.

This implementation works for POKKEN, Donkey Kong Tropical Freeze and Xenoblade Chronicles to some extent.

This PR includes #3473, the first commit is 59c1475.

### DIFF
--- a/src/video_core/engines/maxwell_3d.h
+++ b/src/video_core/engines/maxwell_3d.h
@@ -634,6 +634,11 @@ public:
             u32 address_low;
             s32 buffer_size;
             s32 buffer_offset;
+
+            GPUVAddr Address() const {
+                return static_cast<GPUVAddr>((static_cast<GPUVAddr>(address_high) << 32) |
+                                             address_low);
+            }
         };
         static_assert(sizeof(TransformFeedbackBinding) == 32);
 
@@ -650,6 +655,10 @@ public:
                 return true;
             }
             return shader_config[index].enable != 0;
+        }
+
+        bool IsShaderConfigEnabled(Regs::ShaderProgram type) const {
+            return IsShaderConfigEnabled(static_cast<std::size_t>(type));
         }
 
         union {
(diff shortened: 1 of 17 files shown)


src/video_core/engines/maxwell_3d.h
```diff
@@ -621,6 +622,27 @@ class Maxwell3D final : public ConstBufferEngineInterface {
             float depth_range_far;
         };
 
+        struct alignas(32) TransformFeedbackBinding {
```
lioncash: You may want to document in a comment why the alignments are necessary.

src/video_core/renderer_vulkan/vk_shader_decompiler.cpp
```diff
@@ -760,13 +806,45 @@ class SPIRVDecompiler final : public Sirit::Module {
...
+                Decorate(id, spv::Decoration::XfbStride, static_cast<u32>(tfb->stride));
+                Decorate(id, spv::Decoration::Offset, static_cast<u32>(tfb->offset));
+            }
+
+            element += static_cast<u8>(num_components);
+        }
+    }
+
+    std::optional<VaryingTFB> GetTransformFeedbackInfo(Attribute::Index index, u8 element = 0) {
+        const u8 location = static_cast<u8>(index) * 4 + element;
```
lioncash:
```suggestion
        const u8 location = static_cast<u8>(static_cast<u32>(index) * 4 + element);
```

Otherwise this'll cause a truncation and signedness warning (operations on types `< sizeof(int)` promote to an `int`)

bunnei:
Untagging, due to compilation issue with type_traits

PR #3494, opened 2020-03-11
gl_shader_manager: Fix interaction between graphics and compute

ReinUsesLisp:
After a compute shader was set to the pipeline, no graphics shader was
invoked again. To address this use glUseProgram to bind compute shaders
(without state tracking) and call glUseProgram(0) when transitioning out
of it back to the graphics pipeline.

- Fixes ZLA on OpenGL.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -565,7 +565,7 @@ void RasterizerOpenGL::Draw(bool is_indexed, bool is_instanced) {
     bind_ubo_pushbuffer.Bind();
     bind_ssbo_pushbuffer.Bind();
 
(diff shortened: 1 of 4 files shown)


roger6325:
In the yuzu Early Access build 20200312-3d7983e95 (265), the ZLA start screen graphical still glitches both on the OpenGL and Vulkan (Device: Radon Pro 555) 
PS. Mainline 191 is okay on both API

ReinUsesLisp:
EA 266 seems to be fine here on Nvidia. If the problem persists, please open an issue. There might be some weird interaction between programs and program pipelines.

roger6325:
The issue is same as [https://github.com/yuzu-emu/yuzu/issues/3495](url)
Graphical glitches after Yuzu 192 update #3495

Issue #3493, opened 2020-03-10
Fails to build with Clang 10


jbeich:
Not a recent regression unlike https://github.com/yuzu-emu/yuzu/issues/3407. Removing `static_cast` seems to help.

```c++
$ CC=clang10 CXX=clang++10 cmake -GNinja /path/to/yuzu
$ ninja
[...]
In file included from src/video_core/video_core.cpp:11:
In file included from src/./video_core/renderer_base.h:12:
In file included from src/./core/frontend/emu_window.h:11:
src/./core/frontend/framebuffer_layout.h:41:54: error: arithmetic between floating-point type 'float' and enumeration type 'Layout::ScreenUndocked' [-Werror,-Wenum-float-conversion]
        return static_cast<float>(screen.GetWidth()) / ScreenUndocked::Width;
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~
```

vitor-k:
Can you check if the error is fixed?

jbeich:
I confirm, builds fine after e891ff9a0cbd5466fff302ff73862f3a73440efd.

PR #3496, opened 2020-03-11
framebuffer_layout.h: drop the use of enum for screen dimensions

vitor-k:
There's no reason to use enum here.
May address #3493.


bunnei:
Changes look fine, although please address clang format issues. Thanks!

PR #3500, opened 2020-03-12
texture_cache: Report incompatible textures as black

ReinUsesLisp:
Some games bind incompatible texture types to certain types.
For example Astral Chain binds a 2D texture with 1 layer (non-array) to
a cubemap slot (that's how it's used in the shader). After testing this
in hardware, the expected "undefined behavior" is to report all pixels
as black.

We already have a path for reporting black textures in the texture
cache. When textures types are incompatible, this commit binds these
kind of textures. This is done on the API agnostic texture cache so no
extra code has to be inserted on OpenGL or Vulkan.

As a side effect, this fixes invalidations of ASTC textures on Astral
Chain. This happened because yuzu detected a cube texture and forced
6 faces, generating a texture larger than what the TIC reported.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -104,6 +104,11 @@ public:
         if (!cache_addr) {
             return GetNullSurface(SurfaceParams::ExpectedTarget(entry));
         }
+
+        if (!IsTypeCompatible(tic.texture_type, entry)) {
+            return GetNullSurface(SurfaceParams::ExpectedTarget(entry));
+        }
+
         const auto params{SurfaceParams::CreateForTexture(format_lookup_table, tic, entry)};
         const auto [surface, view] = GetSurface(gpu_addr, cache_addr, params, true, false);
         if (guard_samplers) {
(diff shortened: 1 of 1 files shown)


src/video_core/texture_cache/texture_cache.h
```diff
@@ -1082,6 +1089,34 @@ class TextureCache {
...
+        case Tegra::Texture::TextureType::Texture1D:
+        case Tegra::Texture::TextureType::Texture1DArray:
+            return shader_type == Tegra::Shader::TextureType::Texture1D;
+        case Tegra::Texture::TextureType::Texture1DBuffer:
+            // TODO(Rodrigo): Assume as valid for now
+            return true;
+        case Tegra::Texture::TextureType::Texture2D:
+        case Tegra::Texture::TextureType::Texture2DArray:
+        case Tegra::Texture::TextureType::Texture2DNoMipmap:
+            return shader_type == Tegra::Shader::TextureType::Texture2D;
```
FernandoS27: Separate Texture2DArray and check it against cube and cube array. 2d-array is compatible with those two.

PR #3504, opened 2020-03-13
audio_renderer: passthrough 6 channel data

makigumo:
Seems to play fine in Megaman 11 intro

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -291,8 +291,9 @@ void AudioRenderer::VoiceState::RefreshBuffer(Memory::Memory& memory) {
             samples[index * 2 + 1] = new_samples[index];
         }
         break;
-    case 2: {
-        // 2 channel is played as is
+    case 2:
+    case 6: {
+        // 2 and 6 channel is played as is
         samples = std::move(new_samples);
         break;
     }


FearlessTobi:
I'm not sure this is correct. I'd rather have ogniK or someone well versed check on this.

Its-Rei:
Yuzu crashes with this PR with Fire Emblem Three Houses every time there is a cutscene.
Steps to replicate the issue with the PR:

- boot fe3h
- select `New Game`
- follow the on-screen prompts for difficulty etc.
- a progress bar shows up
- yuzu crashes right as the progress bar is filled and the cutscene would start playing

Here's a log after the crash:
[EA_351.txt](https://github.com/yuzu-emu/yuzu/files/4466772/EA_351.txt)

The `[  90.554738] Debug <Critical> audio_core\audio_renderer.cpp:GetMixVolumeIndex:366: Assertion Failed! Unimplemented channel_count=6` lines still show up without the PR but yuzu does not crash anymore since this PR was untagged.

makigumo:
Works for me on mainline with this patch. Progress bar finishes, black cut scene starts with sound. This is with v1.0.0 of the game [010055D009F78000].

Its-Rei:
I tested it with the same title id [010055D009F78000] but with the newest update v1.2.0 and wave 4 DLC. It crashes consistently every single time with EA 351 where this PR was included and does never crash with EA 352 where it was disabled (tried both 5+ times).

BSoD also tried this on his PC, also with version 1.2.0 and had the same results. It's definitely this PR that causes the issue.

makigumo:
Very curiously 1.2.0 also works for me. I can't test with EA.
I'm attaching my build, so you can test it yourself.

[Release.zip](https://github.com/yuzu-emu/yuzu/files/4466916/Release.zip)

Its-Rei:
I do not get the crash with the build you posted but I definitely got the crash when the PR was in EA. It's reproducible as well as it crashes every time I try it with the last build that had the PR and does not crash every time  I try EA 352, when it was removed, or later. I've had multiple users confirm that the issue is gone for them as well since the PR was untagged.

makigumo:
Thanks for testing. I think it is a combination of things causing the crash. Too bad it is not clear which other things are changed/merged in the EA builds whose interactions are causing the crash.

PR #3505, opened 2020-03-13
shader_decode: implement XMAD mode CSfu

namkazt:
Follow this:

```c++
if (p_a==0 || p_b==0) break;
//v & 0x80000000 -> as_twos_complement(v) < 0
if (p_a & 0x80000000) c -= 65536;
if (p_b & 0x80000000) c -= 65536;
```

ref: [link](https://envytools.readthedocs.io/en/latest/hw/graph/maxwell/cuda/int.html)

### DIFF
--- a/src/video_core/shader/decode/xmad.cpp
+++ b/src/video_core/shader/decode/xmad.cpp
@@ -100,8 +100,7 @@ u32 ShaderIR::DecodeXmad(NodeBlock& bb, u32 pc) {
                                                               op_a, Immediate(0));
             const Node comp_b = GetPredicateComparisonInteger(PredCondition::Equal, is_signed_b,
                                                               op_b, Immediate(0));
-            const Node comp =
-                Operation(OperationCode::LogicalOr, std::move(comp_a), std::move(comp_b));
+            const Node comp = Operation(OperationCode::LogicalOr, comp_a, comp_b);
 
             const Node comp_minus_a = GetPredicateComparisonInteger(
                 PredCondition::NotEqual, is_signed_a,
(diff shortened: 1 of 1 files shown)


src/video_core/shader/decode/xmad.cpp
```diff
@@ -63,15 +64,18 @@ u32 ShaderIR::DecodeXmad(NodeBlock& bb, u32 pc) {
...
+                           instr.xmad.high_a ? Immediate(16) : Immediate(0), Immediate(16));
 
     const Node original_b = op_b;
-    op_b = BitfieldExtract(op_b, is_high_b ? 16 : 0, 16);
+    op_b = SignedOperation(OperationCode::IBitfieldExtract, is_signed_b, std::move(op_b),
+                           is_high_b ? Immediate(16) : Immediate(0), Immediate(16));
 
-    // TODO(Rodrigo): Use an appropiate sign for this operation
-    Node product = Operation(OperationCode::IMul, NO_PRECISE, op_a, op_b);
+    // we already check sign_a and sign_b is difference or not before so just use one in here.
```
FearlessTobi: `We already checked if sign_a and sign_b is different before, so just use one in here`?

src/video_core/shader/decode/xmad.cpp
```diff
@@ -88,12 +92,40 @@ u32 ShaderIR::DecodeXmad(NodeBlock& bb, u32 pc) {
...
+        }
+        case Tegra::Shader::XmadMode::CSfu: {
+            const Node comp_a = GetPredicateComparisonInteger(PredCondition::Equal, is_signed_a,
+                                                              op_a, Immediate(0));
+            const Node comp_b = GetPredicateComparisonInteger(PredCondition::Equal, is_signed_b,
+                                                              op_b, Immediate(0));
+            const Node comp = Operation(OperationCode::LogicalOr, comp_a, comp_b);
+
+            const Node comp_minus_a = GetPredicateComparisonInteger(
+                PredCondition::NotEqual, is_signed_a,
```
pendingchaos: I'm not familiar at all with this IR, but couldn't this be a single (signed) LessThan comparison instead of a NotEqual comparison and IBitwiseAnd?

ReinUsesLisp: It should be possible. What the IR doesn't support 16 or 8 bit integers (yet), so those have to be emulated. 32 bit integers should be fine.

bunnei:
LGTM. Thank you!

PR #3506, opened 2020-03-13
shader_decode: Implement partial ATOM/ATOMS instr

namkazt:
This PR implements all operations for ATOM/ATOMS but `Increment `and `Decrement`.
for now, it only supports S32 and U32 type.

Note: still have some issue with OpenGL

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -2307,6 +2307,8 @@ private:
         ~Func() = delete;
 
         static constexpr std::string_view Add = "Add";
+        static constexpr std::string_view Min = "Min";
+        static constexpr std::string_view Max = "Max";
         static constexpr std::string_view And = "And";
         static constexpr std::string_view Or = "Or";
         static constexpr std::string_view Xor = "Xor";
@@ -2457,7 +2459,21 @@ private:
         &GLSLDecompiler::AtomicImage<Func::Xor>,
         &GLSLDecompiler::AtomicImage<Func::Exchange>,
 
+        &GLSLDecompiler::Atomic<Func::Exchange, Type::Uint>,
         &GLSLDecompiler::Atomic<Func::Add, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::Min, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::Max, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::And, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::Or, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::Xor, Type::Uint>,
+
+        &GLSLDecompiler::Atomic<Func::Exchange, Type::Int>,
+        &GLSLDecompiler::Atomic<Func::Add, Type::Int>,
+        &GLSLDecompiler::Atomic<Func::Min, Type::Int>,
+        &GLSLDecompiler::Atomic<Func::Max, Type::Int>,
+        &GLSLDecompiler::Atomic<Func::And, Type::Int>,
+        &GLSLDecompiler::Atomic<Func::Or, Type::Int>,
+        &GLSLDecompiler::Atomic<Func::Xor, Type::Int>,
 
         &GLSLDecompiler::Branch,
         &GLSLDecompiler::BranchIndirect,
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -2114,6 +2114,10 @@ private:
 
     template <const std::string_view& opname, Type type>
     Expression Atomic(Operation operation) {
+        if ((opname == Func::Min || opname == Func::Max) && type == Type::Int) {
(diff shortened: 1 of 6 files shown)


src/video_core/shader/decode/memory.cpp
```diff
@@ -375,25 +378,87 @@ u32 ShaderIR::DecodeMemory(NodeBlock& bb, u32 pc) {
...
+        const bool is_signed =
+            instr.atoms.type == AtomicType::S32 || instr.atoms.type == AtomicType::S64;
         Node gmem = MakeNode<GmemNode>(real_address, base_address, descriptor);
-        Node value = Operation(OperationCode::AtomicAdd, std::move(gmem), GetRegister(instr.gpr20));
+        Node data = GetRegister(instr.gpr20);
+
+        Node value = [&]() {
+            switch (instr.atoms.operation) {
+            case AtomicOp::Add:
+                return SignedOperation(OperationCode::AtomicIAdd, is_signed, std::move(gmem),
```
ReinUsesLisp: All of this can be on a separate switch case inside a function in an anonymous namespace to avoid code repetition.

src/video_core/shader/decode/memory.cpp
```diff
@@ -375,25 +378,87 @@ u32 ShaderIR::DecodeMemory(NodeBlock& bb, u32 pc) {
...
         address = Operation(OperationCode::IAdd, std::move(address), Immediate(offset));
 
         Node memory = GetSharedMemory(std::move(address));
         Node data = GetRegister(instr.gpr20);
 
-        Node value = Operation(OperationCode::AtomicAdd, std::move(memory), std::move(data));
+        Node value = [&]() {
+            switch (instr.atoms.operation) {
+            case AtomicOp::Add:
+                return SignedOperation(OperationCode::AtomicIAdd, is_signed, std::move(memory),
```
ReinUsesLisp: Ditto

src/video_core/shader/node.h
```diff
@@ -162,7 +162,21 @@ enum class OperationCode {
     AtomicImageXor,      /// (MetaImage, int[N] coords) -> void
     AtomicImageExchange, /// (MetaImage, int[N] coords) -> void
 
-    AtomicAdd, /// (memory, {u}int) -> {u}int
+    AtomicUExchange, /// (memory, uint) -> uint
```
ReinUsesLisp: These functions don't have SignedOperation cases defined.
They can be found in `SignedToUnsignedCode` inside `node_helper.cpp`.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2227,7 +2229,21 @@ class GLSLDecompiler final {
...
         &GLSLDecompiler::Atomic<Func::Add, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::Min, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::Max, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::And, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::Or, Type::Uint>,
+        &GLSLDecompiler::Atomic<Func::Xor, Type::Uint>,
+
+        &GLSLDecompiler::Atomic<Func::Exchange, Type::Int>,
+        &GLSLDecompiler::Atomic<Func::Add, Type::Int>,
+        &GLSLDecompiler::Atomic<Func::Min, Type::Int>,
```
ReinUsesLisp: This implementation for min/max on GLSL is wrong. You'll have to spin to properly implement signed max/min operations atomically.

bunnei:
@namkazt do you plan to address the feedback here?

namkazt:
> @namkazt do you plan to address the feedback here?

@bunnei yeb!, i'm just a bit busy this time. sorry about that. gonna address it in few days or this weekend.

src/video_core/shader/node.h
namkazt: addressed

src/video_core/shader/decode/memory.cpp
namkazt: addressed

namkazt:
@ReinUsesLisp please check.

src/video_core/shader/decode/memory.cpp
```diff
@@ -28,6 +27,34 @@ using Tegra::Shader::StoreType;
 
 namespace {
 
+Node getAtomOperation(AtomicOp op, bool is_signed, Node memory, Node data) {
```
ReinUsesLisp:
```suggestion
Node GetAtomOperation(AtomicOp op, bool is_signed, Node memory, Node data) {
```

src/video_core/shader/decode/memory.cpp
```diff
@@ -28,6 +27,34 @@ using Tegra::Shader::StoreType;
...
+    switch (op) {
+    case AtomicOp::Add:
+        return SignedOperation(OperationCode::AtomicIAdd, is_signed, std::move(memory),
```
ReinUsesLisp: These calls can be reduced to something like this:
```cpp
const OperationCode operation_code = [op] {
    switch (op) {
    case AtomicOp::Add:
        return OperationCode::AtomicIAdd;
    case AtomicOp::Min:
        return OperationCode::AtomicIMin;
    // ...
    }
}();
return SignedOperation(operation_code, is_signed, std::move(memory), std::move(data));

PR #3508, opened 2020-03-13
PageTable: move backing addresses to a children class as the CPU page table does not need them.

FernandoS27:
This PR aims to reduce the memory usage in the CPU page table by moving GPU specific parameters into a child class. This saves 1Gb of Memory for most games.

### DIFF
--- a/src/common/page_table.h
+++ b/src/common/page_table.h
@@ -76,9 +76,20 @@ struct PageTable {
      */
     std::vector<PageType> attributes;
 
-    std::vector<u64> backing_addr;
-
     const std::size_t page_size_in_bits{};
+};
+
+/**
+ * A more advanced Page Table with the ability to save a backing address when using it
+ * depends on another MMU.
+ */
+struct BackingPageTable : PageTable {
+    explicit BackingPageTable(std::size_t page_size_in_bits);
+    ~BackingPageTable();
+
+    void Resize(std::size_t address_space_width_in_bits);
+
+    std::vector<u64> backing_addr;
 };
 
 } // namespace Common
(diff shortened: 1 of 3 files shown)


src/common/page_table.h
```diff
@@ -76,9 +77,17 @@ struct PageTable {
      */
     std::vector<PageType> attributes;
 
-    std::vector<u64> backing_addr;
-
     const std::size_t page_size_in_bits{};
 };
 
+class BackingPageTable : public PageTable {
```
lioncash: This should have a documentation comment like `PageTable` does.

src/common/page_table.h
```diff
@@ -76,9 +77,17 @@ struct PageTable {
...
+public:
+    explicit BackingPageTable(std::size_t page_size_in_bits);
+    ~BackingPageTable();
+
+    void Resize(std::size_t address_space_width_in_bits);
+
+    std::vector<u64> backing_addr;
```
lioncash: This should be private if `BackingPageTable` is a class and not a struct.

src/common/page_table.h
```diff
@@ -46,7 +46,8 @@ struct SpecialRegion {
  * A (reasonably) fast way of allowing switchable and remappable process address spaces. It loosely
  * mimics the way a real CPU page table works.
  */
-struct PageTable {
+class PageTable {
```
lioncash: This should remain a struct if all the data members are public, since no invariants are being enforced.

src/common/page_table.h
```diff
@@ -76,9 +76,20 @@ struct PageTable {
...
 
+/**
+ * A more advanced Page Table with the ability to save a backing address when using it
+ * depends on another MMU.
+ */
+struct BackingPageTable : PageTable {
+    explicit BackingPageTable(std::size_t page_size_in_bits);
+    ~BackingPageTable();
+
+    void Resize(std::size_t address_space_width_in_bits);
```
degasus: override

FernandoS27: That made me learn that it does not work on structs. If I add it, I get an error.

PR #3513, opened 2020-03-14
 video_core: Use native ASTC when available 

ReinUsesLisp:
This uses native ASTC bypassing our decoder when it is reported by the host driver. It improves boot times on some games.

Nvidia on OpenGL reports ASTC as available, but it decodes it behind the scenes. Their decoder is a bit faster than ours, although it seems to be implemented on the CPU (no compute magic).


src/video_core/texture_cache/surface_params.h
```diff
@@ -67,16 +65,14 @@ class SurfaceParams {
...
-            }
-        } else {
-            host_size_in_bytes = GetInnerMemorySize(true, false, false);
+    std::size_t GetHostSizeInBytes(bool is_converted) const {
+        if (!is_converted) {
+            return GetInnerMemorySize(true, false, false);
+        }
+        // ASTC is uncompressed in software, in emulated as RGBA8
+        std::size_t host_size_in_bytes = 0;
+        for (u32 level = 0; level < num_levels; ++level) {
```
lioncash: Should this loop be placed into its own function? This is currently duplicated within `GetHostMipmapLevelOffset` it seems

ReinUsesLisp:
It seems to cause regressions on AMD OpenGL. Untagging for now.

jduncanator:
The Mesa drivers also expose a software ASTC decoder (so all recent Mesa-based drivers report that ASTC is available), however it seems tremendously broken.

When passed certain textures from AC:NH, the Mesa decoder will either decode complete junk, or attempt to allocate massive buffers (one specific texture causes Mesa to attempt to allocate 63GB), which cause texture uploads to fail with GL_OUT_OF_MEMORY.

I haven't tested these specific drivers to determine if they return `true` when queried for `INTERNALFORMAT_SUPPORTED`, however from a brief discussion with @ReinUsesLisp, it seems that there may be a way to differentiate hardware ASTC support and software ASTC support by use of some other fields.

For now, Ryujinx plans on introducing a hardware blacklist for software ASTC textures, however a "built-in" way to be able to differentiate hardware and driver provided ASTC support would be much preferred.

ReinUsesLisp:
I've addressed the issues, re-tagging.

PR #3520, opened 2020-03-16
gl_shader_decompiler: Implement legacy varyings

ReinUsesLisp:
Legacy varyings are special attributes carried over in hardware from
the OpenGL 1 and OpenGL 2 days. These were generally used instead of the
generic attributes we use today. They are deprecated or removed from
most APIs, but Nvidia still ships them in hardware.

To implement these, this commit maps them 1:1 to OpenGL compatibility.

- Used by Xash3D.


bunnei:
Untagging because we are fairly confident this regressed SMO

ReinUsesLisp:
Addressed.

Issue #3528, opened 2020-03-17
Animal Crossing: New Horizons doesn't start


ANIMEFANATIC91:
I'm using Yuzu 191
The emulator gets stuck in launching screen, no matter how I configure it.

Hexagon12:
The game has not been officially released yet. Please wait for the official release.

TheMysticLynx:
The game has been officially released.

burstpup:
Re-open the issue for the release of the game? I've got my copy, lol

Evernow:
@Hexagon12 Re-open, the game has been released.

Issue #2621, opened 2019-06-27
Missing implementation of TransactionId type 2 'SetBufferCount' in VI service 


GDhex12:
* Needed for Super Mario Maker 2

<meta name="titleids"  content="01009B90006DC000">

hottestchilipepper:
Hi, I'm new on yuzu.
I'm struggling to implement 'SetBufferCount', just now at getting informations about this.
I see that I should define 'IGBPSetBufferCountRequestParcel' and 'IGBPSetBufferCountResponseParcel'.
Then at the 'TransactParcel' function I will implement the case of 'SetBufferCount' transaction.
However, at this link [https://switchbrew.org/wiki/Nvnflinger_services#TransactParcel](url) and any other place, there are no information about structs of 'IGBPSetBufferCountRequestParcel' and 'IGBPSetBufferCountResponseParcel'.
Do I have to just test and see values of memory? or are there some docs about these?

hottestchilipepper:
I will modify 'buffer_queue' to implement this one

ZiggyDev:
> I will modify 'buffer_queue' to implement this one

It should be the same as [Android's implementation](https://github.com/shashlik/old-android-frameworks-native/blob/master/libs/gui/IGraphicBufferProducer.cpp#L73), try reading those docs to see if you can find something useful.

Anyway, I can confirm this is an issue with Animal Crossing: New Horizons after locally building with [PR#3531](https://github.com/yuzu-emu/yuzu/pull/3531).

hottestchilipepper:
> Anyway, I can confirm this is an issue with Animal Crossing: New Horizons after locally building with [PR#3531](https://github.com/yuzu-emu/yuzu/pull/3531).
I confirmed it while I tried to play it.
Thank you for your help.
I will implement this.

choppymaster:
Is this the IGraphicBufferProducer: 2 command?

choppymaster:
Needed for: 
* Animal crossing: new horizons 
* 
* Crayon Shin-chan The Storm Called Flaming Kasukabe Runner! 
* Ring fit Adventure 
* Woodle tree 2
* seabed
* wenjia 
* Prison princess
* heaven dust
* tools up
* doom 64
* Manticore - Galaxy on Fire
* blades of time

german77:
This is already implemented. https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/service/nvflinger/buffer_queue_producer.cpp#L63

PR #3539, opened 2020-03-21
Implementing SetBufferCount in BufferQueue

hottestchilipepper:
I just have implemented the SetBufferCount function in BufferQueue.
Described at #2621, and related #3528
Please check this.
It's not an actual implement, because of leak of information how to get a new IGBPBuffer from GPU...
But it seems it's not a problem yet.

Unfortunately, there is another problem to play Animal Crossing: New Horizons.
I will check it.


FearlessTobi:
If this implementation doesn't really do anything, and other games using this function already work without the stub, why should we merge it at all?

bunnei:
Closing a PR that has not been updated in almost a year.

PR #3529, opened 2020-03-17
maxwell_3d: Implement MME shadow RAM

blueskythlikesclouds:
Implementation was based on this Ryujinx PR: https://github.com/Ryujinx/Ryujinx/pull/987
Makes Hatsune Miku: Project DIVA MEGA39's render correctly.

PR #3542, opened 2020-03-22
Implement MME shadow RAM

namkazt:
This implemented based on #3529. I hope it working ok

Thank @fincs to explain it to me again.

Screenshot from @BSoDGamingYT
Capture1
Capture2


ReinUsesLisp:
The default shadow state must match the default state initialized in `InitializeRegisterDefaults`. This can be done by adding a `shadow_state = regs;` after `regs` is initialized in that same function.

namkazt:
> The default shadow state must match the default state initialized in `InitializeRegisterDefaults`. This can be done by adding a `shadow_state = regs;` after `regs` is initialized in that same function.

addressed

PR #3545, opened 2020-03-23
audio_core: Accept Audren REV8

FearlessTobi:
According to Ryujinx, REV8 only added changes on Performance buffer and Wavebuffer DSP command generation.
As we don't support any of those, we can just increment the revision number for now.

This is required by Animal Crossing New Horizons.

### DIFF
--- a/src/audio_core/audio_renderer.h
+++ b/src/audio_core/audio_renderer.h
@@ -189,7 +189,7 @@ struct UpdateDataHeader {
     UpdateDataHeader() {}
 
     explicit UpdateDataHeader(const AudioRendererParameter& config) {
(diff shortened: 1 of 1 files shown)


jduncanator:
Just a heads up, we bumped from [REV7 to REV8](https://github.com/Ryujinx/Ryujinx/pull/993/files#diff-547dfa5ab13f2251c0768f004a136fb4L6), I noticed you're going from REV4 to REV8. Whilst it might work fine, the only changes we observed being made were relative to REV7, and not REV4.

I'm not sure how many changes were made between REV4 and REV8, but I suspect more than just unsupported features were changed.

PR #3549, opened 2020-03-24
account: stub ListQualifiedUsers

makigumo:
Allows to start:
* _Contra: Rogue Corps_
* _Bulletstorm: Duke of Switch Edition_

### DIFF
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -340,6 +340,13 @@ void Module::Interface::InitializeApplicationInfoRestricted(Kernel::HLERequestCo
     rb.Push(InitializeApplicationInfoBase(pid));
 }
 
+void Module::Interface::ListQualifiedUsers(Kernel::HLERequestContext& ctx) {
+    LOG_WARNING(Service_ACC, "(STUBBED) called");
+    ctx.WriteBuffer(profile_manager->GetAllUsers());
+    IPC::ResponseBuilder rb{ctx, 2};
+    rb.Push(RESULT_SUCCESS);
+}
+
 ResultCode Module::Interface::InitializeApplicationInfoBase(u64 process_id) {
     if (application_info) {
         LOG_ERROR(Service_ACC, "Application already initialized");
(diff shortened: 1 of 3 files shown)


src/core/hle/service/acc/acc.cpp
```diff
@@ -28,6 +28,7 @@
 #include "core/hle/service/glue/manager.h"
 #include "core/hle/service/sm/sm.h"
 #include "core/loader/loader.h"
+using namespace Common;
```
FearlessTobi: Don't do that.

src/core/hle/service/acc/acc.cpp
```diff
@@ -340,6 +341,13 @@ void Module::Interface::InitializeApplicationInfoRestricted(Kernel::HLERequestCo
     rb.Push(InitializeApplicationInfoBase(pid));
 }
 
+void Module::Interface::ListQualifiedUsers(Kernel::HLERequestContext& ctx) {
+    LOG_DEBUG(Service_ACC, "(STUBBED) called");
```
FearlessTobi: LOG_WARNING

makigumo:
Thanks, fixed.

This is basically guesswork from the workings of the other functions. I haven't really gotten far in my RE efforts yet :(

src/core/hle/service/acc/acc.cpp
```diff
@@ -340,6 +340,13 @@ void Module::Interface::InitializeApplicationInfoRestricted(Kernel::HLERequestCo
     rb.Push(InitializeApplicationInfoBase(pid));
 }
 
+void Module::Interface::ListQualifiedUsers(Kernel::HLERequestContext& ctx) {
+    LOG_WARNING(Service_ACC, "(STUBBED) called");
+    ctx.WriteBuffer(profile_manager->GetAllUsers());
```
lioncash: I'm quite sure this isn't correct based off what I remember from RE a while ago. Qualified users can differ from the total list of users if I'm not mistaken.

CJBok: This is purely a guess, but I think ListQualifiedUsers might be related to Parental Control.
I know that Contra Rogue Corps which is Rated M and uses this function at start of the game.

theboy181: What is the Parental Control function called ?

CJBok: It is a configuration setting on the Switch itself. To be more precise, it is possibly related to Parental Control -> Restrictred Software where you can set the minimum age of which a profile is not allowed to play. So possibly the function contains a minimum age parameter unless this information is already given in a different function.

image

PR #3552, opened 2020-03-25
Refactor Context management (Fixes renderdoc on opengl issues)

jroweboy:
Changes the GraphicsContext to be managed by the GPU core. This eliminates the need for the frontends to fool around with tricky MakeCurrent/DoneCurrent calls that are dependent on the settings (such as async gpu option).

This also refactors out the need to use QWidget::fromWindowContainer as that caused issues with focus and input handling. Now we use a regular QWidget and just access the native windowHandle() directly.

Another change is removing the debug tool setting in FrameMailbox. Instead of trying to block the frontend until a new frame is ready, the core will now take over presentation and draw directly to the window if the renderer detects that its hooked by NSight or RenderDoc

Lastly, since it was in the way, I removed ScopeAcquireWindowContext and replaced it with a simple subclass in GraphicsContext that achieves the same result

The first commit is just https://github.com/yuzu-emu/yuzu/pull/3548 which I needed for local development.


src/core/core.cpp
```diff
@@ -186,7 +181,6 @@ struct System::Impl {
 
     ResultStatus Load(System& system, Frontend::EmuWindow& emu_window,
                       const std::string& filepath) {
-        Core::Frontend::ScopeAcquireContext acquire_context{emu_window};
```
ReinUsesLisp: Unnecessary newline here.

src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -327,8 +326,8 @@ void ShaderCacheOpenGL::LoadDiskCache(const std::atomic_bool& stop_loading,
 
     const auto worker = [&](Core::Frontend::GraphicsContext* context, std::size_t begin,
                             std::size_t end) {
-        context->MakeCurrent();
-        SCOPE_EXIT({ return context->DoneCurrent(); });
+        const auto scope = context->Acquire();
+        const auto supported_formats = GetSupportedFormats();
```
ReinUsesLisp: Why was this added?

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -317,47 +309,19 @@ class FrameMailbox {
...
-    }
 };
 
-RendererOpenGL::RendererOpenGL(Core::Frontend::EmuWindow& emu_window, Core::System& system)
-    : VideoCore::RendererBase{emu_window}, emu_window{emu_window}, system{system},
-      frame_mailbox{std::make_unique<FrameMailbox>()} {}
+RendererOpenGL::RendererOpenGL(Core::Frontend::EmuWindow& emu_window, Core::System& system,
+                               Core::Frontend::GraphicsContext& context)
+    : VideoCore::RendererBase{emu_window}, emu_window{emu_window}, system{system}, frame_mailbox{},
+      has_debug_tool{HasDebugTool()}, context{context} {}
```
ReinUsesLisp: `has_debug_tool` should be specified after `context`, this will silence a `-Wswitch` warning

src/video_core/renderer_opengl/gl_shader_cache.cpp
jroweboy: I can try moving it back. I moved it inside the thread because the CPU GL Context isn't current until the GPU is started, but now I require that the frontend starts the GPU, so I suppose I can switch that around again.

src/core/frontend/emu_window.h
```diff
@@ -13,19 +13,39 @@
...
-    virtual void MakeCurrent() = 0;
+    virtual void MakeCurrent() {}
 
     /// Releases (dunno if this is the "right" word) the context from the caller thread
-    virtual void DoneCurrent() = 0;
+    virtual void DoneCurrent() {}
+
+    class Scoped {
+    public:
+        Scoped(GraphicsContext& context_) : context(context_) {
```
lioncash:
```suggestion
        explicit Scoped(GraphicsContext& context_) : context(context_) {
```

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -7,9 +7,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <memory>
-
```
lioncash: These should remain separated

src/video_core/video_core.cpp
```diff
@@ -30,13 +30,23 @@ std::unique_ptr<RendererBase> CreateRenderer(Core::Frontend::EmuWindow& emu_wind
...
-    if (Settings::values.use_asynchronous_gpu_emulation) {
-        return std::make_unique<VideoCommon::GPUAsynch>(system, system.Renderer());
+namespace VideoCore {
+
+std::unique_ptr<Tegra::GPU> CreateGPU(Core::Frontend::EmuWindow& emu_window, Core::System& system) {
+    auto context = emu_window.CreateSharedContext();
+    const auto scope = context->Acquire();
+    auto renderer = CreateRenderer(emu_window, system, *context);
+    if (!renderer->Init()) {
+        return {};
```
lioncash:
```suggestion
        return nullptr;
```

src/video_core/video_core.cpp
```diff
@@ -30,13 +30,23 @@ std::unique_ptr<RendererBase> CreateRenderer(Core::Frontend::EmuWindow& emu_wind
         return nullptr;
     }
 }
+} // namespace
```
lioncash:
```suggestion
} // Anonymous namespace
```

src/yuzu/bootmanager.cpp
```diff
@@ -111,162 +97,195 @@ void EmuThread::run() {
...
+        }
         context->makeCurrent(surface);
     }
 
     void DoneCurrent() override {
         context->doneCurrent();
+        is_current = false;
+    }
+
+    QOpenGLContext* GetShareContext() const {
```
lioncash: This should likely have both variants to preserve the constness of the underlying type.

```cpp
QOpenGLContext* GetShareContext() {
    return context.get();
}

const QOpenGLContext* GetShareContext() const {
    return context.get();
}
```
```

File: src/yuzu/bootmanager.h [hunk 9]
```diff
@@ -165,6 +154,8 @@ class GRenderWindow : public QWidget, public Core::Frontend::EmuWindow {
 
     void CaptureScreenshot(u32 res_scale, const QString& screenshot_path);
 
+    std::pair<u32, u32> ScaleTouch(const QPointF pos) const;
```
lioncash:
```suggestion
    std::pair<u32, u32> ScaleTouch(const QPointF& pos) const;
```

src/yuzu/bootmanager.h
```diff
@@ -87,14 +84,8 @@ class EmuThread final : public QThread {
...
     std::atomic_bool stop_run{false};
-    std::mutex running_mutex;
-    std::condition_variable running_cv;
-
-    /// Only used in asynchronous GPU mode
-    std::unique_ptr<Core::Frontend::GraphicsContext> shared_context;
-
-    /// This is shared_context in asynchronous GPU mode, core_context in synchronous GPU mode
-    Core::Frontend::GraphicsContext& context;
+    std::mutex running_mutex = {};
```
lioncash: `= {}` can be omitted for this and `running_cv`. They already have default constructors.

src/yuzu/bootmanager.h
```diff
@@ -7,23 +7,20 @@
 #include <atomic>
 #include <condition_variable>
 #include <mutex>
-
```
lioncash: These should remain separated

src/yuzu/bootmanager.cpp
```diff
@@ -111,162 +97,195 @@ void EmuThread::run() {
...
+        const auto [x, y] = render_window->ScaleTouch(pos);
+        render_window->TouchMoved(x, y);
+        InputCommon::GetMotionEmu()->Tilt(pos.x(), pos.y());
     }
 
-    ~OpenGLWindow() override {
-        context->doneCurrent();
+    void mouseReleaseEvent(QMouseEvent* event) override {
+        if (event->source() == Qt::MouseEventSynthesizedBySystem)
+            return; // touch input is handled in TouchEndEvent
```
lioncash: This should be braced and have the comment go above the return.

src/yuzu/bootmanager.cpp
```diff
@@ -111,162 +97,195 @@ void EmuThread::run() {
...
+
+        if (event->button() == Qt::LeftButton)
```
lioncash: This and the `else` should be braced for consistency.

src/yuzu/bootmanager.cpp
```diff
@@ -111,162 +97,195 @@ void EmuThread::run() {
...
+            const auto [x, y] = render_window->ScaleTouch(pos);
+            render_window->TouchPressed(x, y);
+        } else if (event->button() == Qt::RightButton) {
+            InputCommon::GetMotionEmu()->BeginTilt(pos.x(), pos.y());
+        }
+    }
+
+    void mouseMoveEvent(QMouseEvent* event) override {
+        if (event->source() == Qt::MouseEventSynthesizedBySystem)
+            return; // touch input is handled in TouchUpdateEvent
```
lioncash: Ditto

src/yuzu/bootmanager.cpp
```diff
@@ -111,162 +97,195 @@ void EmuThread::run() {
...
+        InputCommon::GetKeyboard()->ReleaseKey(event->key());
+    }
 
-        context->setShareContext(shared_context);
-        context->setScreen(this->screen());
-        context->setFormat(format);
-        context->create();
+    void mousePressEvent(QMouseEvent* event) override {
+        if (event->source() == Qt::MouseEventSynthesizedBySystem)
+            return; // touch input is handled in TouchBeginEvent
```
lioncash: Ditto

src/yuzu/bootmanager.cpp
```diff
@@ -111,162 +97,195 @@ void EmuThread::run() {
...
+        setAttribute(Qt::WA_NativeWindow);
+        setAttribute(Qt::WA_PaintOnScreen);
     }
 
-    void exposeEvent(QExposeEvent* event) override {
-        QWindow::requestUpdate();
-        QWindow::exposeEvent(event);
+    virtual ~RenderWidget() = default;
+
+    virtual void Present() {}
```
lioncash: Interface member functions should have documentation to state what their purpose is.

src/yuzu/bootmanager.cpp
```diff
@@ -111,162 +97,195 @@ void EmuThread::run() {
...
-        case QEvent::DragResponse:
-        case QEvent::DragEnter:
-        case QEvent::DragLeave:
-        case QEvent::DragMove:
-            GetMainWindow()->AcceptDropEvent(static_cast<QDropEvent*>(event));
-            return true;
-        default:
-            return QWindow::event(event);
-        }
+    RenderWidget(GRenderWindow* parent) : QWidget(parent), render_window(parent) {
```
lioncash:
```suggestion
    explicit RenderWidget(GRenderWindow* parent) : QWidget(parent), render_window(parent) {
```

PR #3566, opened 2020-03-27
renderer_vulkan/wrapper: Add a Vulkan wrapper (part 1 of 2)

ReinUsesLisp:
The intention behind a Vulkan wrapper is to drop Vulkan-Hpp.

The issues with Vulkan-Hpp are:
- Regular breaks of the API.
- Copy constructors that do the same as the aggregates (fixed recently)
- External dynamic dispatch that is hard to remove
- Alias KHR handles with non-KHR handles making it impossible to use
smart handles on Vulkan 1.0 instances with extensions that were included
on Vulkan 1.1.
- Dynamic dispatchers silently change size depending on preprocessor
definitions. Different files will have different dispatch definitions,
generating all kinds of hard to debug memory issues.

In other words, Vulkan-Hpp is not "production ready" for our needs and
this wrapper aims to replace it without losing RAII and exception
safety. Not being able to update our Vulkan headers due to Vulkan-Hpp
API breakages sometimes forces us to stay behind in recent features.

The wrapper is not fully added in this pull request, although the second part
is mostly boilerplate.

### DIFF
--- a/src/video_core/renderer_vulkan/wrapper.h
+++ b/src/video_core/renderer_vulkan/wrapper.h
@@ -422,4 +422,85 @@ private:
     }
 };
 
+/// Array of a pool allocation.
+/// Analogue to std::vector
+template <typename AllocationType, typename PoolType>
+class PoolAllocations {
+public:
+    /// Construct an empty allocation.
+    PoolAllocations() = default;
+
+    /// Construct an allocation. Errors are reported through IsOutOfPoolMemory().
+    explicit PoolAllocations(std::unique_ptr<AllocationType[]> allocations, std::size_t num,
+                             VkDevice device, PoolType pool, const DeviceDispatch& dld) noexcept
+        : allocations{std::move(allocations)}, num{num}, device{device}, pool{pool}, dld{&dld} {}
+
+    /// Copying Vulkan allocations is not supported and will never be.
+    PoolAllocations(const PoolAllocations&) = delete;
+    PoolAllocations& operator=(const PoolAllocations&) = delete;
+
+    /// Construct an allocation transfering ownership from another allocation.
+    PoolAllocations(PoolAllocations&& rhs) noexcept
+        : allocations{std::move(rhs.allocations)}, num{rhs.num}, device{rhs.device}, pool{rhs.pool},
+          dld{rhs.dld} {}
+
+    /// Assign an allocation transfering ownership from another allocation.
+    /// Releases any previously held allocation.
+    PoolAllocations& operator=(PoolAllocations&& rhs) noexcept {
+        Release();
+        allocations = std::move(rhs.allocations);
+        num = rhs.num;
+        device = rhs.device;
+        pool = rhs.pool;
+        dld = rhs.dld;
+        return *this;
+    }
+
+    /// Destroys any held allocation.
+    ~PoolAllocations() {
+        Release();
+    }
+
+    /// Returns the number of allocations.
+    std::size_t size() const noexcept {
+        return num;
+    }
+
+    /// Returns a pointer to the array of allocations.
+    AllocationType const* data() const noexcept {
+        return allocations.get();
+    }
+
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_vulkan/wrapper.h
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    /// Returns a reference to the element in the passed index.
+    /// @pre: index < size()
+    constexpr const T& operator[](std::size_t index) const noexcept {
+        return ptr[index];
+    }
+
+    /// Returns an iterator to the beginning of the span.
+    constexpr const T* begin() const noexcept {
```
lioncash: This should also provide `cbegin` and `cend` accessors.

src/video_core/renderer_vulkan/wrapper.h
```diff
@@ -0,0 +1,524 @@
...
+    /// This is inteded for subranges.
+    constexpr Span(const T* ptr, std::size_t num) noexcept : ptr{ptr}, num{num} {}
+
+    /// Returns the data pointer by the span.
+    constexpr const T* data() const noexcept {
+        return ptr;
+    }
+
+    /// Returns the number of elements in the span.
+    constexpr u32 size() const noexcept {
```
lioncash: May want to indicate in the comment why you're using `u32` instead of `size_t` for the size type.

src/video_core/renderer_vulkan/wrapper.h
```diff
@@ -0,0 +1,524 @@
...
+namespace Vulkan::vk {
+
+/**
+ * Span for Vulkan arrays.
+ * Based on std::span but optimized for array access instead of iterators.
+ * Size returns uint32_t instead of size_t to ease interaction with Vulkan functions.
+ */
+template <typename T>
+class Span {
+public:
```
lioncash: may want to include the standard type definitions like `value_type`, etc, to allow type inspections in generic code.

src/video_core/renderer_vulkan/wrapper.h
```diff
@@ -0,0 +1,524 @@
...
+        return &handle;
+    }
+
+    /// Returns the held Vulkan handle.
+    Type operator*() const noexcept {
+        return handle;
+    }
+
+    /// Returns true when there's a held object.
+    operator bool() const noexcept {
```
lioncash:
```suggestion
    explicit operator bool() const noexcept {
```

src/video_core/renderer_vulkan/wrapper.h
```diff
@@ -0,0 +1,524 @@
...
+    /// Destroys any held object.
+    void reset() noexcept {
+        Release();
+        handle = nullptr;
+    }
+
+    /// Returns the address of the held object.
+    /// Intended for Vulkan structures that expect a pointer to an array.
+    const Type* address() const noexcept {
+        return &handle;
```
lioncash:
```suggestion
        return std::addressof(handle);
```

is generally advisable for generic types, since it'll always work, even if the underlying type has an overloaded unary `operator&`.

src/video_core/renderer_vulkan/wrapper.h
```diff
@@ -0,0 +1,524 @@
...
```
lioncash: Ditto

src/video_core/renderer_vulkan/wrapper.h
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    /// Returns the held Vulkan handle.
+    Type operator*() const noexcept {
+        return handle;
+    }
+
+    /// Returns true when there's a held object.
+    operator bool() const noexcept {
```
lioncash:
```suggestion
    explicit operator bool() const noexcept {
```

ReinUsesLisp:
While this is approved, don't merge it until I fix the specified issues.

ReinUsesLisp:
Addressed

Issue #2720, opened 2019-07-11
Missing HLE service command 'GetAlbumFileListByAruid' in 'caps:u'


GDhex12:
* Needed for Super Smash Bros. Ultimate (after trying to start a World of Light battle)

<meta name="titleids"  content="01006A800016E000">

shukenmg:
What is needed to solve this?

german77:
Stubbed in #3571

PR #3571, opened 2020-03-27
capsrv: Split Capture services into individual files and stub GetAlbumContentsFileListForApplication

Morph1984:
Splits caps into caps_a, caps_c, caps_u, caps_sc, caps_ss, and caps_su.

Stubs GetAlbumContentsFileListForApplication (#2720) as it is used in Super Smash Bros. Ultimate World of Light. The crash following this is unrelated. This command is also used when taking a snapshot. However, it calls SaveScreenShotEx0 in caps:su next and crashes due to it also being unimplemented.

### DIFF
--- a/src/core/hle/service/caps/caps_u.cpp
+++ b/src/core/hle/service/caps/caps_u.cpp
@@ -1,0 +1,53 @@
+// Copyright 2020 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/hle/service/caps/caps_u.h"
+
+namespace Service::Capture {
+
+class IAlbumAccessorApplicationSession final
+    : public ServiceFramework<IAlbumAccessorApplicationSession> {
+public:
+    explicit IAlbumAccessorApplicationSession()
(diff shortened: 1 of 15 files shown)


src/core/hle/service/caps/caps_u.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+CAPS_U::~CAPS_U() = default;
+
+void CAPS_U::GetAlbumContentsFileListForApplication(Kernel::HLERequestContext& ctx) {
+    // Takes a type-0x6 output buffer containing an array of ApplicationAlbumFileEntry, a PID, an
+    // u8 ContentType, two s64s, and an u64 AppletResourceUserId. Returns an output u64 for total
+    // output entries (which is copied to a s32 by official SW).
+    IPC::RequestParser rp{ctx};
+    auto application_album_entries = rp.PopRaw<std::array<u8, 0x30>>();
```
lioncash: All these variables should be marked const. 

Ones that aren't used should be marked `[[maybe_unused]]` as well

bunnei:
LGTM, thanks!

PR #3572, opened 2020-03-27
Port citra-emu/citra#5127: "common: Port some changes from dolphin"

FearlessTobi:
See citra-emu/citra#5127 for more details and the list of commits.

**Original description**:
Ports parts of dolphin-emu/dolphin#5572, dolphin-emu/dolphin#4611 and dolphin-emu/dolphin#5632
Mostly fixes a few warnings.

### DIFF
--- a/src/common/thread.cpp
+++ b/src/common/thread.cpp
@@ -28,11 +28,8 @@ namespace Common {
 #ifdef _MSC_VER
 
 // Sets the debugger-visible name of the current thread.
-// Uses undocumented (actually, it is now documented) trick.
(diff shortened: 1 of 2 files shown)


src/common/thread.cpp
```diff
@@ -47,7 +44,7 @@ void SetCurrentThreadName(const char* name) {
 
     info.dwType = 0x1000;
     info.szName = name;
-    info.dwThreadID = -1; // dwThreadID;
+    info.dwThreadID = static_cast<DWORD>(-1);
```
lioncash:
```cpp
std::numeric_limits<DWORD>::max();
```

FearlessTobi: Fixed.

PR #3578, opened 2020-03-28
shader/video: Partially implement VMNMX

ReinUsesLisp:
Implements the common usages for VMNMX. Inputs with a different size
than 32 bits are not supported and sign mismatches aren't supported
either.

VMNMX works as follows:
It grabs Ra and Rb and applies a maximum/minimum on them (this is
defined by .MX), having in mind the input sign. This result can then be
saturated. After the intermediate result is calculated, it applies
another operation on it using Rc. These operations are merges,
accumulations or another min/max pass.

This instruction is useful for implementing with a more flexible approach GCN's
min3 and max3 instructions (for instance).

- Used by ACNH. Thanks to Ryujinx for pointing out this game uses VMNMX.

### DIFF
--- a/src/video_core/shader/decode/video.cpp
+++ b/src/video_core/shader/decode/video.cpp
@@ -10,15 +10,23 @@
 
 namespace VideoCommon::Shader {
 
+using std::move;
 using Tegra::Shader::Instruction;
 using Tegra::Shader::OpCode;
 using Tegra::Shader::Pred;
 using Tegra::Shader::VideoType;
 using Tegra::Shader::VmadShr;
+using Tegra::Shader::VmnmxOperation;
+using Tegra::Shader::VmnmxType;
 
 u32 ShaderIR::DecodeVideo(NodeBlock& bb, u32 pc) {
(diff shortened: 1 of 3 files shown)


src/video_core/shader/decode/video.cpp
```diff
@@ -10,16 +10,71 @@
...
 using Tegra::Shader::VmadShr;
+using Tegra::Shader::VmnmxOperation;
+using Tegra::Shader::VmnmxType;
 
 u32 ShaderIR::DecodeVideo(NodeBlock& bb, u32 pc) {
     const Instruction instr = {program_code[pc]};
     const auto opcode = OpCode::Decode(instr);
 
+    if (opcode->get().GetId() == OpCode::Id::VMNMX) {
+        // VMNMX is different to the other video instructions
```
lioncash: Thoughts on moving the contents of this block to its own `ExecuteVMNMX()` function, so that the decode function is nicer to read?

Ideally VSETP and VMAD would be in their own as well, but these are pre-existing so I don't really mind them not being touched.

PR #3594, opened 2020-04-02
yuzu: Drop SDL2 and Qt frontend Vulkan requirements

ReinUsesLisp:
This PR takes Dolphin's DynamicLibrary. The implementation is from Stenzek.
It's slightly modified to properly support move semantics.

Create Vulkan instances and surfaces from the Vulkan backend instead of the frontend. This removes the requirement for QVulkanWindow on Qt, allowing CI to build Linux executables with Vulkan support.
It also simplifies the frontend code, simplifying port efforts.

Thanks to @jroweboy for the initial implementation, this is based on that.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -106,6 +106,8 @@ add_library(common STATIC
     common_funcs.h
     common_paths.h
     common_types.h
+    dynamic_library.cpp
+    dynamic_library.h
     file_util.cpp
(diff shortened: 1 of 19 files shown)


ReinUsesLisp:
Disabled gcc Vulkan in CI again due to missing Qt requirements.

bunnei:
@ReinUsesLisp can you fix the conflicts?

PR #3603, opened 2020-04-04
Port citra-emu/citra#5123: "SDL: Disable hidapi drivers due to compatibility problems with certain controllers"

FearlessTobi:
See citra-emu/citra#5123 for more details.

**Original description**:
The change of SDL2 from 2.0.8 to 2.0.10 broke compatibility with some controllers that reported as Switch and Xbox One controllers.
Having SDL_HINT_JOYSTICK_HIDAPI set to "0" should get us the same behavior as before the change, as all of the reported problems stem from joysticks that have a hidapi driver implementation.

From some investigation, the current drivers try to write to the controller, for setting rumble, stick calibration, LED behavior and other similar tasks. 
Most or all of the users that had this problem and identified their controllers had non-official versions of the controllers. From the available information some of these controllers don't have some of the features the driver tries to use, and the driver fails to init the joystick in those cases.

Another reported issue caused by the hidapi drivers is #5118. In this case the driver works but always writes (0,0,80) to the lightbar, and there is no way to change the driver behavior.

Looking ahead at the proposed 2.0.12 source code, the compatibility problems should be solved by the changes to the driver to differentiate input-only controllers, but the PS4 lightbar behavior stays the same.
Closes citra-emu/citra#5118, but the issue might come back if the hidapi drivers are re-enabled.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -508,6 +508,13 @@ SDLState::SDLState() {
     if (SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1") == SDL_FALSE) {
         LOG_ERROR(Input, "Failed to set hint for background events with: {}", SDL_GetError());
     }
+// these hints are only defined on sdl2.0.9 or higher
+#if SDL_VERSION_ATLEAST(2, 0, 9)
+#if !SDL_VERSION_ATLEAST(2, 0, 12)
+    // There are also hints to toggle the individual drivers if needed.
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI, "0");
+#endif
+#endif
 
     SDL_AddEventWatch(&SDLEventWatcher, this);
 


bunnei:
Now I remember why I untagged this. @FearlessTobi 
```
63>D:\a\1\s\src\input_common\sdl\sdl_impl.cpp(16,10): fatal error C1083: Cannot open include file: 'SDL.h': No such file or directory [d:\a\1\s\build\src\input_common\input_common.vcxproj]
    63>Done Building Project "d:\a\1\s\build\src\input_common\input_common.vcxproj" (default targets) -- FAILED.
```

FearlessTobi:
Note to self: https://github.com/citra-emu/citra/pull/5179 needs to be also ported here

FearlessTobi:
This isn't really needed anymore as we are already on version 2.0.12 now.

FearlessTobi:
I just learned this is still needed for some Linux distros with outdated packages.
Sorry for the notification spam.

v1993:
I would suggest using [`SDL_GetVersion`](https://wiki.libsdl.org/SDL_GetVersion) instead of macros. Macros are expanded at build time while `SDL_GetVersion` tests SDL version used at runtime (it's very relevant on Linux where SDL is shipped via packaging system). This allows to use HIDAPI drivers with just updating SDL2 library to supported version and without having to rebuild Yuzu. This also ensures that building on machine with "bad" SDL won't cause all users of build to have HIDAPI disabled.

bunnei (approved):

bunnei:
This PR has been open almost a year, and since it is a port of Citra, we can be fairly confident that the code in question is fine. If someone would like to submit a follow-on to use `SDL-GetVersion`, please do so. But for now, I am going to merge to cleanup old PRs.

v1993:
@bunnei Since then, this problem have been fixed in SDL and updated version is shipped in all major Linux distros now, not to mention official AppImage. I suggest to revert this PR since it doesn't really affect anyone anymore (or at least implement run-time version check I've suggested).

PR #3606, opened 2020-04-05
service/vi: Partially implement BufferQueue disconnect

ReinUsesLisp:
Thanks to fincs for guiding me in this.
This is not based on reverse engineering. It's purely based on making [deko_examples](https://github.com/switchbrew/switch-examples/tree/master/graphics/deko3d/deko_examples). I'm open for feedback on this, as I don't have experience doing HLE.

`Disconnect` and `DetachBuffer` were no-ops. This commit implements `Disconnect`.

While we are at it, change the code to use a `switch` case.

### DIFF
--- a/src/core/hle/service/nvflinger/buffer_queue.cpp
+++ b/src/core/hle/service/nvflinger/buffer_queue.cpp
@@ -103,6 +103,13 @@ void BufferQueue::ReleaseBuffer(u32 slot) {
     buffer_wait_event.writable->Signal();
 }
 
+void BufferQueue::Disconnect() {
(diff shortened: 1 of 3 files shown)


FearlessTobi:
While I can't say anything about the code itself, maybe this should print a STUBBED log?
Might help debugging in the future if there are inaccuracies between your code and the actual implementation.

ReinUsesLisp:
The original code was stubbed and it didn't print anything. I can add it though.

FernandoS27:
uhm, mingw is failing.

PR #3610, opened 2020-04-05
Refactor all the GPU Caches to use VAddr for cache addressing

FernandoS27:
This refactors aims to match the future changes of Project Epimetheus. This should make GPU caching a lot safer. Sadly it's not a perfect solution as this means GPU caching is tied to a process. Of all the options, this was the best so far but still not perfect and with other issues. A perfect solution would a page table for device memory and have process address be converted into device addresses.

The old solution used physical address but that was made under the premise that physical memory blocks were coalasce. It turns out the VMM should only coalasce virtual memory blocks and not physical. Thus rendering our current approach with flaws.

### DIFF
--- a/src/core/memory.h
+++ b/src/core/memory.h
@@ -294,6 +294,9 @@ public:
     void ReadBlock(const Kernel::Process& process, VAddr src_addr, void* dest_buffer,
                    std::size_t size);
 
+    void ReadBlockUnsafe(const Kernel::Process& process, VAddr src_addr, void* dest_buffer,
+                         std::size_t size);
+
     /**
      * Reads a contiguous block of bytes from the current process' address space.
      *
@@ -311,6 +314,8 @@ public:
      *       current process' address space.
      */
     void ReadBlock(VAddr src_addr, void* dest_buffer, std::size_t size);
+
+    void ReadBlockUnsafe(VAddr src_addr, void* dest_buffer, std::size_t size);
 
     /**
      * Writes a range of bytes into a given process' address space at the specified
@@ -335,6 +340,9 @@ public:
     void WriteBlock(const Kernel::Process& process, VAddr dest_addr, const void* src_buffer,
                     std::size_t size);
 
+    void WriteBlockUnsafe(const Kernel::Process& process, VAddr dest_addr, const void* src_buffer,
+                          std::size_t size);
+
     /**
      * Writes a range of bytes into the current process' address space at the specified
      * virtual address.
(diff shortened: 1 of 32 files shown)


src/core/memory.h
```diff
@@ -294,6 +294,9 @@ class Memory {
     void ReadBlock(const Kernel::Process& process, VAddr src_addr, void* dest_buffer,
                    std::size_t size);
 
+    void ReadBlockUnsafe(const Kernel::Process& process, VAddr src_addr, void* dest_buffer,
```
lioncash, line 297: All of these member functions should have a documentation comment and explain why they're unsafe (and *when* the unsafe ones should be used over the safe one)

src/video_core/memory_manager.h
```diff
@@ -97,6 +97,11 @@ class MemoryManager final {
     void WriteBlockUnsafe(GPUVAddr dest_addr, const void* src_buffer, std::size_t size);
     void CopyBlockUnsafe(GPUVAddr dest_addr, GPUVAddr src_addr, std::size_t size);
 
+    static bool IsGranularRange(GPUVAddr gpu_addr, std::size_t size) {
```
lioncash, line 100: This should have a documentation comment

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -495,20 +495,27 @@ void RasterizerVulkan::Query(GPUVAddr gpu_addr, VideoCore::QueryType type,
 
 void RasterizerVulkan::FlushAll() {}
 
-void RasterizerVulkan::FlushRegion(CacheAddr addr, u64 size) {
+void RasterizerVulkan::FlushRegion(VAddr addr, u64 size) {
+    if (!addr || !size) {
```
lioncash, line 499:
```suggestion
    if (addr == 0 || size == 0) {
```

Since these are integral values, not bools.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -495,20 +495,27 @@ void RasterizerVulkan::Query(GPUVAddr gpu_addr, VideoCore::QueryType type,
...
+    }
+    CacheAddr cache_addr = ToCacheAddr(system.Memory().GetPointer(addr));
     texture_cache.FlushRegion(addr, size);
     buffer_cache.FlushRegion(addr, size);
     query_cache.FlushRegion(addr, size);
 }
 
-void RasterizerVulkan::InvalidateRegion(CacheAddr addr, u64 size) {
+void RasterizerVulkan::InvalidateRegion(VAddr addr, u64 size) {
+    if (!addr || !size) {
```
lioncash, line 509: Ditto

PR #3612, opened 2020-04-06
shader/memory: Implement RED.E.ADD and minor changes to ATOM

ReinUsesLisp:
Implements a reduction operation. It's an atomic operation that doesn't
return a value.

This commit introduces another primitive because some shading languages
might have a primitive for reduction operations.


lioncash:
Does this PR still have issues of some sort? (don't know why the EA tag was added and then removed)

PR #3619, opened 2020-04-07
shader/conversion: Implement I2I sign extension, saturation and selection

ReinUsesLisp:
Reimplements I2I adding sign extension, saturation (clamp source value
to the destination), selection and destination sizes that are not 32
bits wide.

It doesn't implement CC yet.


ReinUsesLisp:
These [these](https://github.com/yuzu-emu/yuzu/files/4442241/i2i.tests.zip) were used for the implementation.

PR #3621, opened 2020-04-07
Log version and about section version fix

SilverBeamx:
I noticed some time ago that the "human-readable" yuzu version was missing from the log and about section. This PR aims to fix that by "hacking together" the required parameter directly into GenerateSCMRev.cmake instead of modifying every single CI to pass the correct argument to cmake.
I tested this against Mainline, i hope it works in the Patreon repo too, as i couldn't test it being that it isn't public.

### DIFF
--- a/CMakeModules/GenerateSCMRev.cmake
+++ b/CMakeModules/GenerateSCMRev.cmake
@@ -48,6 +48,25 @@ if (BUILD_REPOSITORY)
   endif()
 endif()
 
+# "Hack": Generate BUILD_FULLNAME from the Git desc
+if (GIT_DESC)
+  # regex capture the characters before the first "-" into CMAKE_MATCH_1
+  string(REGEX MATCH "^([a-z]*)-.*" OUTVAR ${GIT_DESC})
+  if ("${CMAKE_MATCH_COUNT}" GREATER 0)
+    # capitalize the first letter of the repo name.
+    string(SUBSTRING ${CMAKE_MATCH_1} 0 1 FIRST_LETTER)
+    string(SUBSTRING ${CMAKE_MATCH_1} 1 -1 REMAINDER)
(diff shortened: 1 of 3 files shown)


SilverBeamx:
I followed that very closely. The window title is built by formatting the string "yuzu {5} ", supplied via cmake command. As such the cpp files only know the title bar format and the build version, but not the build repo. This is hardcoded somewhere and passed to cmake based on the repo you are building for. If you want i can recycle the same format in the log and about section, but i don't think this would be the right course of action. Almost every cmake parameter is unset/unused apart from the title bar formats and the build number, as such i had limited options and had to derive the build repo from the github description, that is supplied automatically.

(Sorry for the wall of text)

DarkLordZach:
For all intents and purposes, `Common::g_title_bar_format_idle` is the version of yuzu. I just double-checked on CI and it evaluates to either `yuzu ###` or `yuzu Early Access ###`. This is plenty sufficient for everywhere. Just make sure to have a fallback based on git rev if the variable doesn't exist (like what setWindowTitle does)

SilverBeamx:
Ok, i'll modify my PR to reflect that. Also, if i drop the changes in the cmake, i don't think a fallback based on the git rev would be useful. This fallback would only be invoked if yuzu is built without the CI, and as such the first section of the git-rev would be empty (which is what i used to determine the repo). Would a simple "yuzu " fallback string be acceptable?

DarkLordZach:
Probably add something like 'development' or similar to indicate that it was built not on CI, but that's up to you.

SilverBeamx:
Addressed feedback.

DarkLordZach:
1. We use `snake_case` not `camelCase` for variable names
2. Fix the clang formatting errors.

Then I'd be happy to merge.

SilverBeamx:
You're absolutely right regarding the snake case, sorry. Force of habit.

Addressed feedback.

PR #3630, opened 2020-04-09
common/file_util: Allow access to files on network shares

benru:
On Windows, network shares use paths like \\\\server\share\file which were being broken by FileUtil::SanitizePath() removing double slashes.

Changed the code in SanitizePath to permit a double-backslash if it occurs at the start of a filepath (on Windows only).


goldenx86:
I hope no one tries this with a 10mbps network.

bunnei:
I see no harm in allowing it; Games work fine for me on mapped network drives

PR #3634, opened 2020-04-10
Texture Cache: Only do buffer copies on accurate GPU.

FernandoS27:
This is a simple optimization as Buffer Copies are mostly used for texture recycling. They are, however, useful when games abuse undefined behavior but most 3D APIs forbid it.


lioncash:
Commit message should definitely include the rationale for this one before merging though.

bunnei:
I'll revise the commit message on merge

PR #3635, opened 2020-04-10
Buffer queue: Correct behavior of free buffer.

FernandoS27:
This corrects the behavior of free buffer after witnessing it in an unrelated hardware test. I haven't found any games affected by it but in name of better accuracy we'll correct such behavior.


src/core/hle/service/nvflinger/buffer_queue.cpp
```diff
@@ -28,23 +28,45 @@ void BufferQueue::SetPreallocatedBuffer(u32 slot, const IGBPBuffer& igbp_buffer)
...
+
+            // Make sure that the parameters match.
+            return buffer.igbp_buffer.width == width && buffer.igbp_buffer.height == height;
+        });
+
+        if (itr != queue.end()) {
+            free_buffers.erase(f_itr);
+            break;
+        }
+        f_itr++;
```
ReinUsesLisp, line 68:
```suggestion
        ++f_itr;
```

PR #3637, opened 2020-04-11
Port citra-emu/citra#5094: "yuzu: Option to hide mouse on inactivity"

FearlessTobi:
See citra-emu/citra#5094 for more details.

**Original description**:
Add the option to hide the mouse cursor when a defined amount of time (currently ~~3s~~ 2.5s)
Adresses/closes citra-emu/citra#4687.

Mouse clicks now count as activity and the setting can be changed while emulation is running.

After reading the comments and some deliberation, I decided to not implement in this PR neither hotkeys nor changing the mouse timeout. Both are rather niche features and could be easily implemented by someone passionate about it, if they decide the current amount of control over the mouse hiding is not enough.

UI change:
hide-mouse

### DIFF
--- a/src/yuzu/bootmanager.cpp
+++ b/src/yuzu/bootmanager.cpp
@@ -292,6 +292,8 @@ GRenderWindow::GRenderWindow(GMainWindow* parent_, EmuThread* emu_thread_)
     setLayout(layout);
     InputCommon::Init();
 
+    this->setMouseTracking(true);
+
     connect(this, &GRenderWindow::FirstFrameDisplayed, parent_, &GMainWindow::OnLoadComplete);
 }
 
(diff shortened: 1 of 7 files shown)


goldenx86:
Yes please.

src/yuzu/bootmanager.cpp
```diff
@@ -351,6 +353,7 @@ void GRenderWindow::mousePressEvent(QMouseEvent* event) {
     } else if (event->button() == Qt::RightButton) {
         InputCommon::GetMotionEmu()->BeginTilt(pos.x(), pos.y());
     }
+    QWidget::mouseMoveEvent(event);
```
lioncash, line 356: Is this supposed to be `QWidget::mousePressEvent`? If not, it should probably be documented that this is intentional or provide a reason why it's needed, since this seems like a typo when calling the base class member function.

vitor-k: > Is this supposed to be QWidget::mousePressEvent?
Err, yes, it is.
Copy/paste error.

bunnei:
@FearlessTobi can you address the comment above?

FearlessTobi:
Fixed.

Issue #3647, opened 2020-04-12
Super Mario Odyssey - Regression, display error in Wooded Kingdom since mainline 223


Its-Rei:
The lighting/mist in Wooded Kingdom displays in moving horizontal lines since mainline 223.

Both OpenGL and Vulkan, async and without async are affected. Early Access as well.

Mainline 222
![222](https://user-images.githubusercontent.com/61669744/79071683-51c98780-7cdd-11ea-8ddd-30468de0bf8b.png)

Mainline 223
![223](https://user-images.githubusercontent.com/61669744/79071689-59892c00-7cdd-11ea-8c18-4274286c5a33.png)

FernandoS27:
The issue has been identified, a fix will be deployed soon.

FernandoS27:
Fixed by #3649.

PR #3649, opened 2020-04-12
Texture Cache: Read current data when flushing a 3D segment.

FernandoS27:
This PR corrects flushing of 3D segments when data of other segments is mixed, this aims to preserve the data in place.


FernandoS27:
Closes #3647.

src/video_core/texture_cache/surface_base.cpp
```diff
@@ -248,8 +248,14 @@ void SurfaceBaseImpl::FlushBuffer(Tegra::MemoryManager& memory_manager,
 
     // Use an extra temporal buffer
     auto& tmp_buffer = staging_cache.GetBuffer(1);
+    /// Special case for 3D Texture Segments
```
lioncash, line 251:
```suggestion
    // Special case for 3D Texture Segments
```

Otherwise this causes a warning with doxygen (and clang with -Wdocumentation).

PR #3653, opened 2020-04-13
 renderer_vulkan: Integrate Nvidia Nsight Aftermath on Windows

ReinUsesLisp:
Adds optional support for Nsight Aftermath. It is enabled through
ENABLE_NSIGHT_AFTERMATH in cmake. A path to the SDK has to be provided
by the environment variable NSIGHT_AFTERMATH_SDK.

Nsight Aftermath allows an application to generate "minidumps" of the
GPU state when a device loss happens. By analysing these on Nsight we
can know what a game was doing and why it triggered a device loss.

The dump is generated inside %APPDATA%\yuzu\log\gpucrash and this
directory is deleted every time a new instance is initialized with
Nsight enabled.

To enable it on yuzu there has a to be a driver and device capable of
running Nsight Aftermath on Vulkan. That means only Turing based GPUs
on the latest stable driver, beta drivers won't work for now.

It is manually enabled in Configuration>Debug>Enable Graphics Debugging
because when using all debugging capabilities there is a runtime cost.

### DIFF
--- a/src/video_core/renderer_vulkan/nsight_aftermath_tracker.h
+++ b/src/video_core/renderer_vulkan/nsight_aftermath_tracker.h
@@ -1,0 +1,87 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <mutex>
+#include <string>
+#include <vector>
+
+#define VK_NO_PROTOTYPES
+#include <vulkan/vulkan.h>
+
+#ifdef HAS_NSIGHT_AFTERMATH
+#include <GFSDK_Aftermath_Defines.h>
+#include <GFSDK_Aftermath_GpuCrashDump.h>
+#include <GFSDK_Aftermath_GpuCrashDumpDecoding.h>
+#endif
+
+#include "common/common_types.h"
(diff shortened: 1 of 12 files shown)


ReinUsesLisp:
Nsight Aftermath is disabled by default because it'd require shipping proprietary files.

src/video_core/renderer_vulkan/nsight_aftermath_tracker.h
```diff
@@ -0,0 +1,83 @@
...
+#include "common/dynamic_library.h"
+
+namespace Vulkan {
+
+class NsightAftermathTracker {
+public:
+    NsightAftermathTracker();
+    ~NsightAftermathTracker();
+
+    NsightAftermathTracker(const NsightAftermathTracker&) = delete;
```
lioncash, line 30: The move assignment and constructor should likely be explicitly defaulted or deleted if the copy variants are specified to make the semantics explicit. (specifying a destructor will prevent the move variants from being generated unless they're explicitly defaulted).

ReinUsesLisp: I've explicitly deleted the move constructor. Initialization takes a `this` as user data, it's probably a bad idea moving it.

PR #3656, opened 2020-04-14
gl_shader_cache: Use CompileDepth::FullDecompile on GLSL

ReinUsesLisp:
From my testing on a Splatoon 2 shader that takes 3800ms on average to
compile changing to FullDecompile reduces it to 900ms on average.

The shader decoder will automatically fallback to a more naive method if
it can't use full decompile.


goldenx86:
This will be fun to test in FE3H.

FernandoS27:
Well the only reason I kept this as a debug feature was because it didn't seem to have any benefits compared to the traditional method, since it improves shader compiling times, I think it's great to have it set.

PR #3659, opened 2020-04-14
service: time: Implement CalculateStandardUserSystemClockDifferenceByUser.

bunnei:


src/core/hle/service/time/time.cpp
```diff
@@ -308,6 +308,29 @@ void Module::Interface::GetClockSnapshotFromSystemClockContext(Kernel::HLEReques
...
 
+void Module::Interface::CalculateStandardUserSystemClockDifferenceByUser(
+    Kernel::HLERequestContext& ctx) {
+    LOG_DEBUG(Service_Time, "called");
+
+    IPC::RequestParser rp{ctx};
+    const auto snapshot_a = rp.PopRaw<Clock::ClockSnapshot>();
+    const auto snapshot_b = rp.PopRaw<Clock::ClockSnapshot>();
+
+    Clock::TimeSpanType time_span_type{Clock::TimeSpanType::FromSeconds(
```
lioncash, line 319: Isn't this also able to be auto? Given the function is called `FromSeconds` and is qualified by the class type, it's pretty reasonable to assume the return type is obvious.

PR #3662, opened 2020-04-14
gl_rasterizer: Implement constant vertex attributes

ReinUsesLisp:
Credits go to gdkchan from Ryujinx for finding constant attributes are
used in retail games.

This commit does not implement them on Vulkan.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -140,8 +140,8 @@ void RasterizerOpenGL::SetupVertexFormat() {
         const auto attrib = gpu.regs.vertex_attrib_format[index];
         const auto gl_index = static_cast<GLuint>(index);
 
-        // Ignore invalid attributes.
-        if (!attrib.IsValid()) {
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -140,8 +140,8 @@ void RasterizerOpenGL::SetupVertexFormat() {
         const auto attrib = gpu.regs.vertex_attrib_format[index];
         const auto gl_index = static_cast<GLuint>(index);
 
-        // Ignore invalid attributes.
-        if (!attrib.IsValid()) {
+        // Disable constant attributes.
+        if (attrib.IsConstant()) {
```
FernandoS27, line 144: Why not check for both IsValid or IsConstant

ReinUsesLisp: Because if it's not valid it will signal an error, and that behavior is not tested.
We needed `IsValid` before because we were not using `IsConstant` (NVN's default for disabled attributes)

PR #3665, opened 2020-04-15
FS: Improve emulation of device saves

bunnei:
This is some minor improvements for emulation of device saves. This:
- Fixes Animal Crossing: New Horizons saving.
- Makes it so "Open Save Data Location" option goes to the device save folder where appropriate.
- Stubs the FileSystem MultiCommitManager. Since this is not strictly necessary for the only game we've seen use this (ACNH), we should implement this with the ongoing VFS rewrite. @DarkLordZach as FYI.

### DIFF
--- a/src/core/file_sys/control_metadata.cpp
+++ b/src/core/file_sys/control_metadata.cpp
@@ -95,6 +95,10 @@ u32 NACP::GetSupportedLanguages() const {
     return raw.supported_languages;
 }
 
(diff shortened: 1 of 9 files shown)


FernandoS27:
Moved to mainline from EA as 1 week has passed (this PR was first on the private repo). Since it should wait until new VFS, we are not going to merge this.

PR #3667, opened 2020-04-15
vk_blit_screen: Initialize all members in VkPipelineViewportStateCreateInfo

ReinUsesLisp:
When the dynamic state is specified, pViewports and pScissors are
ignored, quoting the specification:

  pViewports is a pointer to an array of VkViewport structures, defining
  the viewport transforms. If the viewport state is dynamic, this member
  is ignored.

That said, AMD's proprietary driver itself seem to read it regardless of
what the specification says.


goldenx86:
Being AMD user is suffering.

PR #3668, opened 2020-04-15
maxwell_to_vk: Add uint16 vertex formats

ReinUsesLisp:
- Used by Xenoblade 2.


lioncash:
Last commit should probably mention the game that uses it, so it's stored in the repository,

PR #3675, opened 2020-04-15
externals: Use shared libraries if possible

degasus:
This is mostly done by pkgconfig.
I've focused on the larger and more stable libraries.

Also the OpenSSL dependency is moved to httplib.

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -86,7 +86,10 @@ if (ENABLE_WEB_SERVICE)
     set(LIBRESSL_SKIP_INSTALL ON CACHE BOOL "")
     add_subdirectory(libressl EXCLUDE_FROM_ALL)
     target_include_directories(ssl INTERFACE ./libressl/include)
-	target_compile_definitions(ssl PRIVATE -DHAVE_INET_NTOP)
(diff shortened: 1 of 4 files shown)


externals/CMakeLists.txt
```diff
@@ -30,13 +45,28 @@ add_subdirectory(glad)
...
+    add_subdirectory(lz4/contrib/cmake_unofficial EXCLUDE_FROM_ALL)
+    target_include_directories(lz4_static INTERFACE ./lz4/lib)
+endif()
 
 # mbedtls
-add_subdirectory(mbedtls EXCLUDE_FROM_ALL)
-target_include_directories(mbedtls PUBLIC ./mbedtls/include)
+# mbedtls_cipher_cmac seems to be missing on many builds. So don't use the shared library
+# find_package(MBEDTLS 2.12.0)
+if (MBEDTLS_FOUND)
```
degasus, line 61: This is dead code right now. It seems like at least the archlinux library is shipped without CMAC support.

Do you want me to debug more why this issue exists at all, shall I delete this handler here, or shall I keep it as it is?

degasus: I've dropped it.

PR #3678, opened 2020-04-16
guest_driver: Take std::vector by reference in DeduceTextureHandlerSize()

lioncash:
There's no need to take the whole vector by value, considering it only reads the data like a view. A constant reference is sufficient here.


lioncash:
Never mind, turns out I can't read (missed the std::sort)

PR #3681, opened 2020-04-16
decoder/image: Fix incorrect G24R8 component sizes in GetComponentSize()

lioncash:
The components' sizes were mismatched. This corrects that.

### DIFF
--- a/src/video_core/shader/decode/image.cpp
+++ b/src/video_core/shader/decode/image.cpp
@@ -201,10 +201,10 @@ u32 GetComponentSize(TextureFormat format, std::size_t component) {
         return 0;
     case TextureFormat::G24R8:
         if (component == 0) {
-            return 8;
+            return 24;
         }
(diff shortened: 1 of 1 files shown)


FernandoS27:
Component 0 is Red and component 1 is Green. I believe they were correct.

lioncash:
Would this not make its use in [GetComponentType()](https://github.com/yuzu-emu/yuzu/blob/24620bc4ea9ca59a757b7f07ca912f6645c5b8ef/src/video_core/shader/decode/image.cpp#L111) incorrect? Its case label is returning `g_type` for component 0 rather than r_type.

ReinUsesLisp:
Never mind, I should test this in hardware.

bunnei:
Added the `do-not-merge` tag, since it sounds like @ReinUsesLisp plans to verify this on HW first

PR #3593, opened 2020-04-02
[WIP] [Not Yet For Review] MacOS support through MoltenVK

jroweboy:
This branch is based off Rodrigo's WIP vulkan-hpp removal, vulkan 1.0 support, and my macOS specific tweaks. This is still suuuuper experimental.

This uses the vulkan renderer with moltenVK and its kinda hacked together atm. I doubt this even builds on anyone elses machine without massaging it to get it running, but if someone wants to pick this up, I figure its best to get this out in the open. This needs a lot of work before its ready for use, but I'll upload a WIP `.app` of the build here whenever I get a chance. USE AT YOUR OWN RISK. Also its self signed, so it's going to be annoying to get it to run on your mac i bet. 

Still needs a lot of work before its ready, so don't bother reviewing yet until the needed PRs are merged.

Build link:
[yuzu.app.zip](https://github.com/yuzu-emu/yuzu/files/4418521/yuzu.app.zip)

Screenshots:

https://i.imgur.com/MmsLfn9.jpg

https://i.imgur.com/Uy5hJPt.jpg

https://i.imgur.com/C7VLzdx.jpg

PR #3692, opened 2020-04-16
MoltenVK/MacOS support

VHRanger:
This pull request builds on  #3593 and rebases with master since the required PRs have been closed. It also adds normal command line build support and targets for OSX 10.14.

All of the credit for the OSX compatibility goes to @jroweboy 

Still missing: SDL2 frontend on OSX.


VHRanger:
Will resubmit once cleaner

PR #3708, opened 2020-04-17
Fix compilation error

gxcreator:
maxwell_to_gl.h:186:9: error: this statement may fall through

### DIFF
--- a/src/video_core/renderer_opengl/maxwell_to_gl.h
+++ b/src/video_core/renderer_opengl/maxwell_to_gl.h
@@ -193,6 +193,7 @@ inline GLenum TextureFilterMode(Tegra::Texture::TextureFilter filter_mode,
         }
         break;
     }
+    break;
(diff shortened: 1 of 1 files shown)


lioncash:
Weren't these compilation errors already resolved in: https://github.com/yuzu-emu/yuzu/pull/3706?

gxcreator:
There are also breaks needed on higher level. At least on mingw it seems to be required to avoid fall through errors.

lioncash:
Our buildbots make use of mingw for the Windows builds and they seem to build yuzu just fine, so I'm not sure that's the case.

gxcreator:
Just tried with mingw on ubuntu and it builds well. Iit looks like different behaviour between my ubuntu's mingw(7.3) and MSYS2's mingw(8.3). I believe 8.3 have added more checks.
Which gcc version do you use?

gxcreator:
Seems to be fixed on newer gcc versions

PR #3732, opened 2020-04-19
vulkan: Remove unnecessary includes

lioncash:
Reduces some header churn and reduces rebuilds when some header internals change.

While we're at it we can also resolve a missing include in buffer_cache.


bunnei:
@lioncash can you rebase when you get a chance? Thanks!

lioncash:
@bunnei done

Issue #3731, opened 2020-04-19
Initialisation order mismatch in vk_rasterizer


H27CK:
Build failure. Initialisation order mismatch in `video_core/renderer_vulkan/vk_rasterizer.h` and `video_core/renderer_vulkan/vk_rasterizer.cpp`. From `vk_rasterizer.cpp`:

```
uint8_pass(device, scheduler, descriptor_pool, staging_pool, update_descriptor_queue),
quad_indexed_pass(device, scheduler, descriptor_pool, staging_pool, update_descriptor_queue),
```
From `src/video_core/renderer_vulkan/vk_rasterizer.h`:
```
QuadIndexedPass quad_indexed_pass;
Uint8Pass uint8_pass;
```
Ordering mismatch causes failure due to `-Werror=reorder`. Consistent ordering resolves issue, i.e.
```
Uint8Pass uint8_pass;
QuadIndexedPass quad_indexed_pass;
```
in `src/video_core/renderer_vulkan/vk_rasterizer.h`.

PR #3733, opened 2020-04-19
Initialize quad_indexed_pass before uint8_pass

ambasta:
Fixes Werror=reorder in gcc

The members are delcared out of order in the headers


ambasta:
Also fixes #3731

Gustav0ar:
This PR fixes building mainline with gcc in linux for me.

PR #3734, opened 2020-04-20
decode/arithmetic_half: Fix HADD2 and HMUL2 absolute and negation bits

ReinUsesLisp:
The encoding for negation and absolute value was wrong.
Extracting is now done manually. Similar instructions having different
encodings is the rule, not the exception. To keep sanity and readability
I preferred to extract the desired bit manually.

This is implemented against nxas:
https://github.com/ReinUsesLisp/nxas/blob/8dbc38995711cc12206aa370145a3a02665fd989/table.h#L68

That is itself tested against nvdisasm (Nvidia's official disassembler).

### DIFF
--- a/src/video_core/shader/decode/arithmetic_half.cpp
+++ b/src/video_core/shader/decode/arithmetic_half.cpp
@@ -19,22 +19,46 @@ u32 ShaderIR::DecodeArithmeticHalf(NodeBlock& bb, u32 pc) {
     const Instruction instr = {program_code[pc]};
     const auto opcode = OpCode::Decode(instr);
 
-    if (opcode->get().GetId() == OpCode::Id::HADD2_C ||
-        opcode->get().GetId() == OpCode::Id::HADD2_R) {
+    bool negate_a = false;
+    bool negate_b = false;
+    bool absolute_a = false;
+    bool absolute_b = false;
+
+    switch (opcode->get().GetId()) {
+    case OpCode::Id::HADD2_R:
         if (instr.alu_half.ftz == 0) {
             LOG_DEBUG(HW_GPU, "{} without FTZ is not implemented", opcode->get().GetName());
         }
+        negate_a = ((instr.value >> 43) & 1) != 0;
+        negate_b = ((instr.value >> 31) & 1) != 0;
+        absolute_a = ((instr.value >> 44) & 1) != 0;
+        absolute_b = ((instr.value >> 30) & 1) != 0;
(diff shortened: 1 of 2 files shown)


src/video_core/shader/decode/arithmetic_half.cpp
```diff
@@ -19,22 +19,46 @@ u32 ShaderIR::DecodeArithmeticHalf(NodeBlock& bb, u32 pc) {
...
             LOG_DEBUG(HW_GPU, "{} without FTZ is not implemented", opcode->get().GetName());
         }
+        negate_a = (instr.value >> 43) & 1;
+        negate_b = (instr.value >> 31) & 1;
+        absolute_a = (instr.value >> 44) & 1;
+        absolute_b = (instr.value >> 30) & 1;
+        break;
+    case OpCode::Id::HADD2_C:
+        if (instr.alu_half.ftz == 0) {
+            LOG_DEBUG(HW_GPU, "{} without FTZ is not implemented", opcode->get().GetName());
```
ogniK5377, line 39: might be good to mark as a warning instead

FernandoS27: nah it's fine that way. It's something that doesn't really affect emulation at all.

src/video_core/shader/decode/arithmetic_half.cpp
```diff
@@ -19,22 +19,46 @@ u32 ShaderIR::DecodeArithmeticHalf(NodeBlock& bb, u32 pc) {
...
+    bool negate_b = false;
+    bool absolute_a = false;
+    bool absolute_b = false;
+
+    switch (opcode->get().GetId()) {
+    case OpCode::Id::HADD2_R:
         if (instr.alu_half.ftz == 0) {
             LOG_DEBUG(HW_GPU, "{} without FTZ is not implemented", opcode->get().GetName());
         }
+        negate_a = (instr.value >> 43) & 1;
```
lioncash, line 32: These should be wrapped like so:

```suggestion
        negate_a = ((instr.value >> 43) & 1) != 0;
```

otherwise these can cause truncation warnings, since it's implicitly truncating down to bool.

src/video_core/shader/decode/arithmetic_half.cpp
ReinUsesLisp: It affects emulation, but we can't implement it without using CUDA or OpenCL.

PR #3735, opened 2020-04-20
Replace externals with Conan

jroweboy:
# THIS NEEDS TESTING BY BOTH DEVELOPERS AND ALL OF THE CI TARGETS BEFORE TAGGING OR MERGING

Replaces many externals with conan for downloading dependencies.

This uses the experimental conan cmake_find_package_multi generator to generate modern PackageConfig.cmake files for each of the dependencies. This is setup such that conan install is called through cmake directly if the dep is missing, otherwise it will onnly load the conan search path in order to find the dependencies. Included with this change are several custom FindPackage.cmake files that are written to find the library through PkgConfig if thats available. These files generate targets that are compatible with the ones that the generated conan Config.cmake files create, so there shouldn't be any issue using system libraries. 

In many cases, the conan generated cmake files will work just fine, but there are a few exceptions. Qt5 being the primary example. For Qt5, we use conan to find the path to the root dir, and then use find package to load the official Qt5Config.cmake. For SDL2, the generated one works fine, but we needed the additional variables that the script generates elsewhere, so it was easier to manually check for that dependency.


CMakeLists.txt
```diff
@@ -115,62 +113,165 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+        libzip:with_openssl=False
+        libzip:enable_windows_crypto=False
+    )
+    # Add the bincrafters remote
+    conan_add_remote(NAME bincrafters INDEX 1
+                    URL https://api.bintray.com/conan/bincrafters/public-conan)
+    conan_cmake_run(REQUIRES ${CONAN_REQUIRED_LIBS}
+                    OPTIONS ${CONAN_LIB_OPTIONS}
+                    BUILD missing
+                    CONFIGURATION_TYPES "Release;Debug"
```
degasus, line 226: CONFIGURATION_TYPES must not be set on linux

externals/CMakeLists.txt
```diff
@@ -106,12 +52,12 @@ if (ENABLE_VULKAN)
...
 endif()
 
-# zlib
-find_package(ZLIB 1.2.11)
-if (NOT ZLIB_FOUND)
-    message(STATUS "zlib 1.2.11 or newer not found, falling back to externals")
-    add_subdirectory(zlib EXCLUDE_FROM_ALL)
-    set(ZLIB_LIBRARIES z)
+# pkgconfig -- it is used to find shared libraries without cmake modules on linux systems
+find_package(PkgConfig)
```
degasus, line 56: I think PkgConfig can be dropped here completely

CMakeLists.txt
```diff
@@ -115,62 +113,165 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+set_target_properties(Boost::Boost PROPERTIES IMPORTED_GLOBAL TRUE)
+set_target_properties(nlohmann_json::nlohmann_json PROPERTIES IMPORTED_GLOBAL TRUE)
+set_target_properties(OpenSSL::OpenSSL PROPERTIES IMPORTED_GLOBAL TRUE)
+if (NOT TARGET Boost::boost)
+    add_library(Boost::boost ALIAS Boost::Boost)
+endif()
+if (NOT TARGET boost)
+    add_library(boost ALIAS Boost::Boost)
+endif()
+if (NOT TARGET json-headers)
```
degasus, line 263: I think we should change all the importer to use the full names Boost::Boost and nlohmann_json::nlohmann_json

externals/find-modules/FindOpus.cmake
```diff
@@ -0,0 +1,42 @@
+
```
degasus, line 1: Case senstive issue in filename: Findopus.cmake

externals/find-modules/FindZSTD.cmake
```diff
@@ -0,0 +1,41 @@
+
```
degasus, line 1: Case senstive issue: Findzstd.cmake

CMakeLists.txt
```diff
@@ -115,62 +113,165 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+    # Cmake has a *serious* lack of 2D array or associative array...
+    # Capitalization matters here. We need the naming to match the generated paths from Conan
+    set(REQUIRED_LIBS
+    #    Cmake Pkg Prefix  Version     Conan Pkg
+        "Boost             1.72        boost/1.72.0"
+        "Catch2            2.11        catch2/2.11.0"
+        "fmt               6.2         fmt/6.2.0"
+        "OpenSSL           1.1         openssl/1.1.1f"
+    # required to fix a dep conflict between qt and sdl2
+        "libiconv          1.16        libiconv/1.16"
```
degasus, line 145: The file is called "Iconv" instead of "libiconv"

CMakeLists.txt
```diff
@@ -115,62 +113,165 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+        string(REGEX REPLACE "[ \t\r\n]+" ";" PACKAGE_SPLIT ${PACKAGE})
+        list(GET PACKAGE_SPLIT 0 PACKAGE_PREFIX)
+        list(GET PACKAGE_SPLIT 1 PACKAGE_VERSION)
+        list(GET PACKAGE_SPLIT 2 PACKAGE_CONAN)
+
+        if (NOT ${PACKAGE_PREFIX}_FOUND)
+            if (FN_USE_CONAN)
+                find_package(${PACKAGE_PREFIX} ${PACKAGE_VERSION} CONFIG REQUIRED QUIET)
+            else()
+                find_package(${PACKAGE_PREFIX} ${PACKAGE_VERSION} QUIET)
```
degasus, line 169: I'd prefer not using quiet here.

CMakeLists.txt
degasus: iconv.h belongs to the glibc, so it is not really a library which can be searched. My suggestion is to move this line to the windows only section

externals/find-modules/FindZSTD.cmake
```diff
@@ -0,0 +1,41 @@
...
+find_package(PkgConfig QUIET)
+pkg_check_modules(PC_zstd QUIET zstd)
+
+find_path(zstd_INCLUDE_DIR
+  NAMES zstd.h
+  PATHS ${PC_zstd_INCLUDE_DIRS}
+)
+find_library(zstd_LIBRARY
+  NAMES zstd_static
```
degasus, line 10: No _static at least on linux

externals/find-modules/FindZSTD.cmake
```diff
@@ -0,0 +1,41 @@
+
+find_package(PkgConfig QUIET)
+pkg_check_modules(PC_zstd QUIET zstd)
```
degasus, line 3: The pkgconfig file is called libzstd

CMakeLists.txt
jroweboy: The generated conan BoostConfig.cmake exports `Boost::Boost`, the upstream BoostConfig.cmake exports `Boost::boost` and dynarmic is searching for the target `boost`. So to get compatibility between the generated conan config file and upstream I just aliased one to the other and use the upstream `Boost::boost` everywhere

CMakeLists.txt
```diff
@@ -115,63 +113,192 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+        include(${CMAKE_BINARY_DIR}/sdl2Config.cmake)
+        list(APPEND CMAKE_MODULE_PATH "${CONAN_SDL2_ROOT_RELEASE}>")
+        list(APPEND CMAKE_PREFIX_PATH "${CONAN_SDL2_ROOT_RELEASE}>")
+    endif()
+    find_package(SDL2 2.0.12)
+    if (TARGET sdl2::sdl2)
+        # If its loading SDL2 from the conan generated cmake, then we don't need to do anything
+    elseif(SDL2_FOUND)
+        # If its found through the system find_package then we want to make a compatible target from the sources
+        add_library(sdl2::sdl2 INTERFACE)
```
degasus, line 207: Seems like cmake does not like to overwrite this symbol.
```
CMake Error at CMakeLists.txt:207 (add_library):
  The target name "sdl2::sdl2" is reserved or not valid for certain CMake
  features, such as generator expressions, and may result in undefined
  behavior.
```

CMakeLists.txt
```diff
@@ -115,63 +113,192 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+endif()
+# find SDL2 exports a bunch of variables that are needed, so its easier to do this outside of the yuzu_find_package
+if(ENABLE_SDL2)
+    if(EXISTS ${CMAKE_BINARY_DIR}/sdl2Config.cmake)
+        include(${CMAKE_BINARY_DIR}/sdl2Config.cmake)
+        list(APPEND CMAKE_MODULE_PATH "${CONAN_SDL2_ROOT_RELEASE}>")
+        list(APPEND CMAKE_PREFIX_PATH "${CONAN_SDL2_ROOT_RELEASE}>")
+    endif()
+    find_package(SDL2 2.0.12)
+    if (TARGET sdl2::sdl2)
```
degasus, line 203: This seems to be a case senstive issue: SDL2::SDL2

CMakeLists.txt
```diff
@@ -115,63 +113,192 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+        list(APPEND CMAKE_PREFIX_PATH "${CONAN_SDL2_ROOT_RELEASE}")
+        find_package(SDL2 2.0.12 REQUIRED)
+    endif()
+
+endif()
+
+# Reexport some targets that are named differently when using the upstream CmakeConfig vs the generated Conan config
+set_target_properties(Boost::Boost PROPERTIES IMPORTED_GLOBAL TRUE)
+set_target_properties(OpenSSL::OpenSSL PROPERTIES IMPORTED_GLOBAL TRUE)
+if (NOT TARGET Boost::boost)
```
degasus, line 284: The official cmake variable is "Boost::headers" now, it was "Boost::boost" before. Where have you seen "Boost::Boost"?
I'd try to use Boost::boost everywhere

CMakeLists.txt
```diff
@@ -115,63 +113,192 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+        list(APPEND CMAKE_MODULE_PATH "${CONAN_SDL2_ROOT_RELEASE}")
+        list(APPEND CMAKE_PREFIX_PATH "${CONAN_SDL2_ROOT_RELEASE}")
+        find_package(SDL2 2.0.12 REQUIRED)
+    endif()
+
+endif()
+
+# Reexport some targets that are named differently when using the upstream CmakeConfig vs the generated Conan config
+set_target_properties(Boost::Boost PROPERTIES IMPORTED_GLOBAL TRUE)
+set_target_properties(OpenSSL::OpenSSL PROPERTIES IMPORTED_GLOBAL TRUE)
```
degasus, line 283: Why do you need to change thos properties? IMO this fits better in the FindXXX handler. Also OpenSSL::OpenSSL isn't defined here.

src/core/CMakeLists.txt
```diff
@@ -606,11 +606,11 @@ endif()
 create_target_directory_groups(core)
 
 target_link_libraries(core PUBLIC common PRIVATE audio_core video_core)
-target_link_libraries(core PUBLIC Boost::boost PRIVATE fmt json-headers mbedtls opus unicorn)
+target_link_libraries(core PUBLIC Boost::boost PRIVATE fmt::fmt nlohmann_json::nlohmann_json mbedtls Opus::Opus unicorn)
```
degasus, line 609: Case senstive error: "opus::opus"

CMakeLists.txt
```diff
@@ -115,63 +113,192 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+if (CONAN_REQUIRED_LIBS)
+    message(STATUS "Packages ${CONAN_REQUIRED_LIBS} not found!")
+    # Use Conan to fetch the libraries that aren't found
+    # Download conan.cmake automatically, you can also just copy the conan.cmake file
+    if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
+    message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+    file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/v0.15/conan.cmake"
+                    "${CMAKE_BINARY_DIR}/conan.cmake")
+    endif()
+    include(${CMAKE_BINARY_DIR}/conan.cmake)
```
degasus, line 226: I get this error here:
```
CMake Error at build2/conan.cmake:82 (message):
  Please specify in command line CMAKE_BUILD_TYPE
  (-DCMAKE_BUILD_TYPE=Release)
```
Shall we just pick the script form dynarmic which selects Release by default?

CMakeLists.txt
```diff
@@ -115,63 +113,192 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+        "zstd              1.4         zstd/1.4.4"
+    )
+
+    foreach(PACKAGE ${REQUIRED_LIBS})
+        string(REGEX REPLACE "[ \t\r\n]+" ";" PACKAGE_SPLIT ${PACKAGE})
+        list(GET PACKAGE_SPLIT 0 PACKAGE_PREFIX)
+        list(GET PACKAGE_SPLIT 1 PACKAGE_VERSION)
+        list(GET PACKAGE_SPLIT 2 PACKAGE_CONAN)
+
+        if (NOT ${PACKAGE_PREFIX}_FOUND)
```
degasus, line 164: I have no clue why, but on the second run with FN_USE_CONAN, all packages are checked again. This is an issue here as OpenSSL does not provide the "config" way of find_package.

jroweboy: > all packages are checked again
They *SHOULD* only be checked IFF they are not found. I wonder if the syntax i did above is wrong for checking if the packages found or not.

degasus: Yeah, I just don't understand why it is not working. The block below works fine, so I assume the syntax is fine. Is it because of the scope being in a function here?

degasus:
LZ4 has changed their numbering schema ( https://github.com/yuzu-emu/yuzu/issues/3790#issuecomment-620877325 for reference), so I assume you will an explicit version check like this one: https://github.com/Kitware/VTK/blob/master/CMake/FindLZ4.cmake#L10

ogniK5377:
Relaying from DMs onto here but:

Building on the conan branch seems to hang on "Target architecture: x86_64" when trying to configure with cmake.

ogniK5377:
Update, hang is fixed. Was a cmake issue and was fixed with an updated version of cmake.

A secondary issue is that qt is trying to build from source and is failing to locate a pre-existing build

CMakeLists.txt
```diff
@@ -115,63 +113,194 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+    file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/v0.15/conan.cmake"
+                    "${CMAKE_BINARY_DIR}/conan.cmake")
+    endif()
+    include(${CMAKE_BINARY_DIR}/conan.cmake)
+
+    set(CONAN_LIB_OPTIONS
+        libzip:with_openssl=False
+        libzip:enable_windows_crypto=False
+    )
+    conan_check(VERSION 1.24.0 REQUIRED)
```
DarkLordZach, line 233: Same reason as before

CMakeLists.txt
```diff
@@ -115,63 +113,194 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+    # Add the bincrafters remote
+    conan_add_remote(NAME bincrafters INDEX 1
```
DarkLordZach, line 235: Don't specify an index, it isn't required and it increases the likelihood of a conflict if a user already has remotes.
It is fine with just name/url

CMakeLists.txt
```diff
@@ -115,63 +113,194 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+        conan_cmake_run(REQUIRES ${CONAN_REQUIRED_LIBS}
+                        "libiconv/1.16"
+                        OPTIONS ${CONAN_LIB_OPTIONS}
+                        BUILD missing
+                        CONFIGURATION_TYPES "Release;Debug"
+                        GENERATORS cmake_multi cmake_find_package_multi)
+        include(${CMAKE_BINARY_DIR}/conanbuildinfo_multi.cmake)
+    else()
+        conan_cmake_run(REQUIRES ${CONAN_REQUIRED_LIBS}
+            "libiconv/1.16"
```
DarkLordZach, line 250: match formatting with other side of if statement

src/common/CMakeLists.txt
```diff
@@ -172,5 +172,5 @@ endif()
 
 create_target_directory_groups(common)
 
-target_link_libraries(common PUBLIC Boost::boost fmt microprofile)
-target_link_libraries(common PRIVATE lz4_static libzstd_static)
+target_link_libraries(common PUBLIC Boost::boost fmt::fmt microprofile)
```
DarkLordZach, line 175: should be Boost::headers

CMakeLists.txt
```diff
@@ -115,63 +113,194 @@ message(STATUS "Target architecture: ${ARCHITECTURE}")
...
+    # Cmake has a *serious* lack of 2D array or associative array...
+    # Capitalization matters here. We need the naming to match the generated paths from Conan
+    set(REQUIRED_LIBS
+    #    Cmake Pkg Prefix  Version     Conan Pkg
+        "Boost             1.72        boost/1.72.0"
+        "Catch2            2.11        catch2/2.11.0"
+        "fmt               6.2         fmt/6.2.0"
+        "OpenSSL           1.1         openssl/1.1.1f"
+    # can't use until https://github.com/bincrafters/community/issues/1173
+        #"libzip            1.5         libzip/1.5.2@bincrafters/stable"
```
DarkLordZach, line 151: Either fix the todo or remove FindLIBZIP.cmake, it doesn't make sense to have unused code around

jroweboy: I'd rather keep a note there to remind people to check up on it.

src/common/CMakeLists.txt
jroweboy: This is not actually whats happening here. This `Boost::boost` is from conan's generated `BoostConfig.cmake` file, and its "correct" as it is.

jroweboy:
The `FindPackage.cmake` names are intentional to make the generated conan `PackageConfig.cmake` files for the record. That way i can call find package on the conan name and find both.

src/common/CMakeLists.txt
DarkLordZach: The built-into-cmake FindBoost.cmake file provides a target called Boost::headers, which only includes the header-only libraries from boost. We should be using that here because nothing in common requires more than that.

CMakeLists.txt
jroweboy: We use this `FindLibzip.cmake` file in `externals/CMakeLists.txt` when we call `find_package(Libzip)` and fallback to externals if its not found.

src/common/CMakeLists.txt
jroweboy: Yes, but the built-into-cmake `FindBoost.cmake` is not being used if conan is providing boost. Conan provides a `cmake_find_package_multi` generator that creates a `boostConfig.cmake` file that `find_package` picks up in order to create the `Boost::Boost` target. This is then aliased to `Boost::boost` for compatibility with the built-into-cmake created `Boost::boost` target. 

So we do not always have a `Boost::headers` target. Why doesn't conan re-export the upstream `boostConfig.cmake`? I think their reasoning is pretty weak, but its "a feature not a bug" https://github.com/conan-io/conan-center-index/wiki/FAQ#why-are-cmake-findconfig-files-and-pkg-config-files-not-packaged (See here for more info https://github.com/conan-io/conan-center-index/pull/1392#issuecomment-615300617 )

PR #3738, opened 2020-04-20
audio_renderer: Implement 6 channel downmixing

ogniK5377:
Adds 6 channel support by downmixing to 2 channels. Nintendo does this in their DSP firmware however the downmix coefficients are different depending on the internal state. This will be updated in a later PR.


Morph1984:
Supersedes #3504

FernandoS27:
question why is channel 4 not being used in the formula?

ogniK5377:
Refer to #3757

PR #3742, opened 2020-04-20
Optimize GPU Command Lists and Introduce Fast GPU Time Option

FernandoS27:
This PR optimizes Command list methods with multiple arguments such as Const Buffer Data Upload. This helps considerably a few games like Diablo 3 which abuse buffer inlines. It also ports a hack from Ryujinx called Fast GPU Time: https://github.com/Ryujinx/Ryujinx/pull/1139 by gdkchan credits to him for this hack as it avoids activating game's dynamic resolution. The hack is enabled/disabled through a separate option in Graphics Advance. It is, however, still recommended to use dynamic resolution disabling mods.

### DIFF
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -209,16 +209,31 @@ void GPU::CallMethod(const MethodCall& method_call) {
 
     ASSERT(method_call.subchannel < bound_engines.size());
 
-    if (ExecuteMethodOnEngine(method_call)) {
+    if (ExecuteMethodOnEngine(method_call.method)) {
         CallEngineMethod(method_call);
     } else {
         CallPullerMethod(method_call);
     }
 }
 
-bool GPU::ExecuteMethodOnEngine(const MethodCall& method_call) {
-    const auto method = static_cast<BufferMethods>(method_call.method);
-    return method >= BufferMethods::NonPullerMethods;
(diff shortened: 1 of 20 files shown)


src/video_core/gpu.cpp
```diff
@@ -132,7 +133,10 @@ u64 GPU::GetTicks() const {
     constexpr u64 gpu_ticks_den = 625;
 
     const u64 cpu_ticks = system.CoreTiming().GetTicks();
```
degasus, line 135: Unrelated to this PR, but accesses CoreTiming.GetTicks from the GPU thread is highly unstable. This logic should be moved to the CPU thread instead.

FernandoS27: That would generate some terrible rendezvous but you are right it's an issue. I don't, however, know the right solution.

src/video_core/gpu.cpp
```diff
@@ -132,7 +133,10 @@ u64 GPU::GetTicks() const {
...
-    const u64 nanoseconds = Core::Timing::CyclesToNs(cpu_ticks).count();
+    u64 nanoseconds = Core::Timing::CyclesToNs(cpu_ticks).count();
+    if (Settings::values.use_fast_gpu_time) {
+        nanoseconds /= 256;
```
H27CK, line 138: Is it worth assessing the performance impact of this value?

PR #3748, opened 2020-04-21
Unbreak -Werror=reorder with Clang

jbeich:
Regressed by #3694 . See [error log](https://github.com/yuzu-emu/yuzu/files/4506950/yuzu-qt5-s20200420.log).


jbeich:
Sorry, I didn't notice #3733.

PR #3753, opened 2020-04-21
{gl,vk}_rasterizer: Add lazy default buffer maker and use it for empty buffers

ReinUsesLisp:
On NVN buffers can be enabled but have no size. According to deko3d and
the behavior we see in Animal Crossing: New Horizons these buffers get
the address of 0x1000 and limit themselves to 0xfff.

Implement buffers without a size by binding a null buffer to OpenGL
without a side.

Introduce a default buffer getter that lazily constructs an empty
buffer. This is intended to match OpenGL's buffer 0.

Use this for disabled vertex and uniform buffers.

While we are at it, include vertex buffer usages for staging buffers to
silence validation errors.

- Fixes asserts on Animal Crossing: New Horizons
- Fixes high memory consumption on Animal Crossing: New Horizons.

### DIFF
--- a/src/video_core/engines/maxwell_3d.h
+++ b/src/video_core/engines/maxwell_3d.h
@@ -1259,7 +1259,8 @@ public:
 
                     GPUVAddr LimitAddress() const {
                         return static_cast<GPUVAddr>((static_cast<GPUVAddr>(limit_high) << 32) |
-                                                     limit_low);
+                                                     limit_low) +
+                               1;
                     }
                 } vertex_array_limit[NumVertexArrays];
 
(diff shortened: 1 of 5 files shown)


src/video_core/engines/maxwell_3d.h
```diff
@@ -1259,7 +1259,8 @@ class Maxwell3D final : public ConstBufferEngineInterface {
 
                     GPUVAddr LimitAddress() const {
                         return static_cast<GPUVAddr>((static_cast<GPUVAddr>(limit_high) << 32) |
-                                                     limit_low);
+                                                     limit_low) +
+                               1;
```
lioncash, line 1263: clang format is wild

ReinUsesLisp: Should we replace `limit_low` and `limit_high` with a single `GPUVAddr`? It's not like we have to handle big endian vs. little endian on yuzu

PR #3321, opened 2020-01-18
audio_renderer: Add a voice mixing system

FearlessTobi:
This implements a rudimentary mixing system for voices in audren. 
So far it only supports Stereo mixing, but Sorround isn't implemented anywhere anyway.

Improvements by this can be seen in quite a few games, including Super Mario Odyssey.
[Before](https://streamable.com/6uk6v) and [After](https://streamable.com/dzp39) (thanks to BsoD Gaming).

This has been extensively hw-tested with a modified version of the `audren-simple` program.
[Version 1](https://cdn.discordapp.com/attachments/576428479726223370/668132374667722782/audren-simple-volumeTobiHex.nro) / [Version 2](https://cdn.discordapp.com/attachments/576428479726223370/668132377360596992/audren-simpleSide.nro)

The only new issue known with this is the menu music in Disgaea 5 not playing anymore, but I think I just uncovered a deeper lying bug here as there are many other audren issues there.

PR #3757, opened 2020-04-22
audio_renderer: Better voice mixing and 6 channel downmixing

ogniK5377:
Supersedes #3738 and #3321

Expands on the initial work done by tobi to fix issues in other games for the voice mixer. This is still preliminary mixing as we don't have dedicated mix buffers yet. 6 channel downmixing was merged into this PR since it needed to be changed either way for voice mixing.

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -37,9 +37,11 @@ public:
     }
 
     void SetWaveIndex(std::size_t index);
-    std::vector<s16> DequeueSamples(std::size_t sample_count, Core::Memory::Memory& memory);
+    std::vector<s16> DequeueSamples(std::size_t sample_count, Core::Memory::Memory& memory,
+                                    std::array<VoiceResourceInformation*, 6> voice_resources);
     void UpdateState();
-    void RefreshBuffer(Core::Memory::Memory& memory);
+    void RefreshBuffer(Core::Memory::Memory& memory,
+                       std::array<VoiceResourceInformation*, 6> voice_resources);
 
 private:
     bool is_in_use{};
@@ -79,7 +81,7 @@ AudioRenderer::AudioRenderer(Core::Timing::CoreTiming& core_timing, Core::Memory
                              std::shared_ptr<Kernel::WritableEvent> buffer_event,
                              std::size_t instance_number)
     : worker_params{params}, buffer_event{buffer_event}, voices(params.voice_count),
-      effects(params.effect_count), memory{memory_} {
+      voice_resources(params.voice_count), effects(params.effect_count), memory{memory_} {
     behavior_info.SetUserRevision(params.revision);
     audio_out = std::make_unique<AudioCore::AudioOut>();
     stream = audio_out->OpenStream(core_timing, STREAM_SAMPLE_RATE, STREAM_NUM_CHANNELS,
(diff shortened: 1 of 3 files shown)


src/audio_core/audio_renderer.cpp
```diff
@@ -37,9 +37,11 @@ class AudioRenderer::VoiceState {
     }
 
     void SetWaveIndex(std::size_t index);
-    std::vector<s16> DequeueSamples(std::size_t sample_count, Core::Memory::Memory& memory);
+    std::vector<s16> DequeueSamples(std::size_t sample_count, Core::Memory::Memory& memory,
+                                    std::array<VoiceResourceInformation*, 6> voice_resources);
     void UpdateState();
-    void RefreshBuffer(Core::Memory::Memory& memory);
+    void RefreshBuffer(Core::Memory::Memory& memory,
+                       std::array<VoiceResourceInformation*, 6> voice_resources);
```
lioncash, line 44: 1. 

You may wish to make this a type alias inside the VoiceState class, given it's repeated in several areas. This should also be passed as a const reference.

src/audio_core/audio_renderer.cpp
```diff
@@ -271,7 +280,8 @@ void AudioRenderer::VoiceState::UpdateState() {
     is_in_use = info.is_in_use;
 }
 
-void AudioRenderer::VoiceState::RefreshBuffer(Core::Memory::Memory& memory) {
+void AudioRenderer::VoiceState::RefreshBuffer(
+    Core::Memory::Memory& memory, std::array<VoiceResourceInformation*, 6> voice_resources) {
```
lioncash, line 284: Ditto about passing it as a const reference.

src/audio_core/audio_renderer.cpp
```diff
@@ -37,9 +37,11 @@ class AudioRenderer::VoiceState {
     }
 
     void SetWaveIndex(std::size_t index);
-    std::vector<s16> DequeueSamples(std::size_t sample_count, Core::Memory::Memory& memory);
+    std::vector<s16> DequeueSamples(std::size_t sample_count, Core::Memory::Memory& memory,
+                                    std::array<VoiceResourceInformation*, 6> voice_resources);
```
lioncash, line 41: const reference.

legend800:
In https://github.com/yuzu-emu/yuzu/pull/3321 bgm was broken on Ultra Street Fighter 2. With this pr, bgm is back but very loud out of right speaker, while sfx are fairly quiet. So a few mixing issues remain there. Sounds fine on master.

ogniK5377:
> In #3321 bgm was broken on Ultra Street Fighter 2. With this pr, bgm is back but very loud out of right speaker, while sfx are fairly quiet. So a few mixing issues remain there. Sounds fine on master.

Should be fixed now, I had a little bit of a logic error. It sounds fine with the latest commit

legend800:
Sounds perfect now! Thanks for the quick fix.

legend800:
Sound is Mario Tennis not correct now. Bgm is barely audible with normal sfx volume. No issues prior.

FernandoS27:
Since I don't know if this can be merged. I'm moving it to mainline.

ogniK5377:
Shouldn't be merged yet as it causes regressions in some games(Shovel knight)

callmenemo491:
"yuzu verify build" fails with:
_The pipeline is not valid. Could not get the latest source version for repository yuzu-emu/yuzu hosted on https://github.com/ using ref 818726c1d865be4b14969687302c9d9d4e5b3346. GitHub reported the error, "Bad credentials"_

PR #3761, opened 2020-04-22
configure_input_player: Use slider to edit modifier scale

Kewlan:
Use the existing deadzone slider to edit the modifier scale when keyboard is being used for stick controls. The slider is now always shown.

I also changed the shown value to show percent instead of decimals. I think that makes more sense to a regular user.

image

The setText at line 543 and 552 are needed because QSlider::valueChanged isn't called when the sliders are set to 0 and the configure window is opened.

### DIFF
--- a/src/yuzu/configuration/configure_input_player.cpp
+++ b/src/yuzu/configuration/configure_input_player.cpp
@@ -56,7 +56,6 @@ static void SetAnalogButton(const Common::ParamPackage& input_param,
     if (analog_param.Get("engine", "") != "analog_from_button") {
         analog_param = {
             {"engine", "analog_from_button"},
-            {"modifier_scale", "0.5"},
         };
     }
     analog_param.Set(button_name, input_param.Serialize());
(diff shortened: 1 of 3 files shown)


src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -517,20 +526,29 @@ void ConfigureInputPlayer::UpdateButtonLabels() {
...
+
+        if (param.Has("engine")) {
+            if (param.Get("engine", "") == "sdl") {
+                if (!param.Has("deadzone")) {
+                    param.Set("deadzone", 0.1f);
+                }
+
+                analog_stick_slider->setValue(static_cast<int>(param.Get("deadzone", 0.1f) * 100));
+                if (analog_stick_slider->value() == 0)
+                    analog_stick_slider_label->setText(tr("Deadzone: 0%"));
```
Morph1984, line 541: This should be braced

src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -517,20 +526,29 @@ void ConfigureInputPlayer::UpdateButtonLabels() {
...
+            } else {
+                if (!param.Has("modifier_scale")) {
+                    param.Set("modifier_scale", 0.5f);
+                }
+
+                analog_stick_slider->setValue(
+                    static_cast<int>(param.Get("modifier_scale", 0.5f) * 100));
+                if (analog_stick_slider->value() == 0)
+                    analog_stick_slider_label->setText(tr("Modifier Scale: 0%"));
```
Morph1984, line 550: Ditto

Morph1984:
Can you provide screenshots of what this functionality looks like?

bunnei:
Looks good, thank you!

Issue #3746, opened 2020-04-20
yuzu-cmd.exe fails due to 'fmt::v6::format_error'


H27CK:
Mainline MinGW-w64 build terminates at runtime (before game launch) with:
```
terminate called after throwing an instance of 'fmt::v6::format_error'
  what():  invalid type specifier
```
Offending line is
```
const auto title = fmt::format(
  "yuzu {} | {}-{} | FPS: {:.0f} ({:.0%})", Common::g_build_fullname,
  Common::g_scm_branch, Common::g_scm_desc, results.game_fps, results.emulation_speed);
```
in `yuzu_cmd/emu_window/emu_window_sdl2.cpp`, where the type specifier for `results.emulation_speed` is erroneous.
```
const auto title = fmt::format(
  "yuzu {} | {}-{} | FPS: {:.0f}", Common::g_build_fullname,
  Common::g_scm_branch, Common::g_scm_desc, results.game_fps);
```
allows game to launch as expected.

PR #3764, opened 2020-04-22
Fix format error in yuzu-cmd performance statistic

H27CK:
This fixes a format error (that otherwise causes `yuzu-cmd` to fail) to allow the reporting of emulation speed statistic in the window title of `yuzu-cmd`. Resolves https://github.com/yuzu-emu/yuzu/issues/3746.

### DIFF
--- a/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
+++ b/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
@@ -182,8 +182,13 @@ void EmuWindow_SDL2::PollEvents() {
     if (current_time > last_time + 2000) {
         const auto results = Core::System::GetInstance().GetAndResetPerfStats();
         const auto title = fmt::format(
-            "yuzu {} | {}-{} | FPS: {:.0f} ({:.0%})", Common::g_build_fullname,
-            Common::g_scm_branch, Common::g_scm_desc, results.game_fps, results.emulation_speed);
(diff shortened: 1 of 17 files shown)


src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
```diff
@@ -182,8 +182,8 @@ void EmuWindow_SDL2::PollEvents() {
     if (current_time > last_time + 2000) {
         const auto results = Core::System::GetInstance().GetAndResetPerfStats();
         const auto title = fmt::format(
-            "yuzu {} | {}-{} | FPS: {:.0f} ({:.0%})", Common::g_build_fullname,
-            Common::g_scm_branch, Common::g_scm_desc, results.game_fps, results.emulation_speed);
+            "yuzu {} | {}-{} | FPS: {:.0f} ({:.0f}%)", Common::g_build_fullname,
```
ReinUsesLisp, line 185: Shouldn't we use `{:.0}` instead of `{:.0f}`? `results.emulation_speed` is a double.

H27CK: Using `{:.0}` seems to result in truncation, i.e. I only see values that are multiples of 10.

H27CK:
I'm going to resubmit this. I've made a mess of this one.

PR #3766, opened 2020-04-23
vk_renderpass_cache: Pack renderpass cache key and unify keys

ReinUsesLisp:
Continues what #3718 started for renderpasses. This allows us to call Common::CityHash and std::memcmp only once for GraphicsPipelineCacheKey.

Framebuffer creation was ignoring the number of color attachments, this caused validation errors. No known issues were caused by this.


ReinUsesLisp:
This seems to cause issues on Intel's proprietary driver on Breath of the Wild, investigating.

ReinUsesLisp:
Addressed.

PR #3768, opened 2020-04-23
Fix format error in performance statistics

H27CK:
Resubmission: this fixes a format error (that otherwise causes `yuzu-cmd` to fail) to allow the reporting of emulation speed statistic in the window title of `yuzu-cmd`. Resolves https://github.com/yuzu-emu/yuzu/issues/3746.


H27CK:
@lioncash @ReinUsesLisp resubmitted without making a mess while squashing commits. Hope that's OK now.

H27CK:
@lioncash does this PR need any more attention?

PR #3771, opened 2020-04-23
Dump RomFS command to include Updates

benru:
Patch the RomFS with the selected updates before dumping. Previously the resulting RomFS only contained data from the original title.

To dump the RomFS without updates the user can disable the update under Properties before choosing Dump RomFS.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1304,7 +1304,9 @@ void GMainWindow::OnGameListDumpRomFS(u64 program_id, const std::string& game_pa
     FileSys::VirtualFile romfs;
 
     if (*romfs_title_id == program_id) {
-        romfs = file;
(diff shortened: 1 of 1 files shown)


src/yuzu/main.cpp
```diff
@@ -1304,7 +1304,9 @@ void GMainWindow::OnGameListDumpRomFS(u64 program_id, const std::string& game_pa
     FileSys::VirtualFile romfs;
 
     if (*romfs_title_id == program_id) {
-        romfs = file;
+        u64 ivfc_offset = loader->ReadRomFSIVFCOffset();
```
lioncash, line 1307:
```suggestion
        const u64 ivfc_offset = loader->ReadRomFSIVFCOffset();
```

bunnei:
LGTM, Thanks!

Issue #3782, opened 2020-04-24
Compile error on Jetson Nano (aarch64)


tkircher:
In physical_core.cpp when ARCHITECTURE_x86_64 is false:

/testing/yuzu/src/core/hle/kernel/physical_core.cpp: In constructor ‘Kernel::PhysicalCore::PhysicalCore(Core::System&, std::size_t, Core::ExclusiveMonitor&)’:
/testing/yuzu/src/core/hle/kernel/physical_core.cpp:30:63: error: cannot convert ‘std::shared_ptr<Core::ARM_Unicorn>’ to ‘Core::ARM_Interface*’ in assignment
     arm_interface = std::make_shared<Core::ARM_Unicorn>(system);

A new Core::ARM_Unicorn is instantiated with std::make_shared<T> but assigned to a Core::ARM_Interface* which isn't a shared_ptr.

tkircher:
Incidentally, instantiating a new object with

arm_interface = new Core::ARM_Unicorn(system);

makes it build successfully, but who knows what chaos that will bring. It did actually launch after it finished building.
![yuzu_jetson_nano](https://user-images.githubusercontent.com/19326823/80176439-0fd5f680-85a5-11ea-8c22-1d4e6c9f161a.jpeg)

tkircher:
Just saw the patch that @lioncash posted. Going to close this.

AuroraRAS:
I hope to know more about your experiment.
Due to the similar structure to Switch, jetson may be the most suitable hardware platform for yuzu.

jesuslg123:
> Just saw the patch that @lioncash posted. Going to close this.

Hi! Could you share you experience and how to compile it for nano please?

tkircher:
@AuroraRAS I definitely wish it were possible to run Switch games natively, considering that the only real difference in hardware is that the Switch has more CUDA cores.

@jesuslg123 I didn't get it to run, just to compile. I thought there might have already been a way to bypass the ARM emulation to run code natively, but that turned out not to be the case.

FearlessTobi:
I'm not sure if this will help you, but an ARM64 port of dynarmic (our JIT) is in the making.

tkircher:
I think that will be extremely helpful. If not for the Jetson specifically, maybe for faster ARM systems.

bigfatbrowncat:
>I'm not sure if this will help you, but an ARM64 port of dynarmic (our JIT) is in the making.

[...]

PR #3783, opened 2020-04-24
physical_core: Make use of std::make_unique instead of std::make_shared in ctor

lioncash:
Allows this code to compile successfully on non x86-64 architectures.

Fixes issue #3782


lioncash:
Merging, as this fixes outlying platform support, and has been sitting without comment for 5 days.

Issue #3772, opened 2020-04-23
PR 3718 breaks vulkan on Linux (amdgpu)


RealJohnGalt:
On Polaris and linux, PR #3718 causes a black screen, with audio occurring behind it (and fps ramps up as usual with video). Tested radv/aco, radv/llvm, amdvlk, and amdgpu-pro vulkan drivers. All but amdgpu-pro have the black screen, and it just crashes.

Log: https://bpaste.net/RSFA

This commit only builds on linux with either static casting or removing the cflag. So worth noting this behavior happens with either compilation workaround implemented. With this commit reverted there's no problem.

**Specs:**
 - Arch Linux
 - Mesa git
 - Ryzen 3600 @ 4.3ghz
 - rx580 8G

lat9nq:
I tested yuzu with #3718 merged using the `-Werror=conversion` compilation workaround. It seems at least GCN 1.0 is not affected on the amdgpu/radv llvm driver, as I am able to boot SSBU and get into a match with minor mesh artifacts.

My specs are:
Manjaro Linux 20.0
Kernel 5.4.32-1-MANJARO
host-passthrough 8 core (AMD Zen+ CPU extensions)
8 GB RAM (Host is DDR4 3200 MHz @ 3400 MHz)
R5 340 2GB

PR #3784, opened 2020-04-24
shader/memory_util: Deduplicate code

ReinUsesLisp:
Deduplicate code shared between vk_pipeline_cache and gl_shader_cache as
well as shader decoder code.

While we are at it, fix a bug in gl_shader_cache where compute shaders
had an start offset of a stage shader.

### DIFF
--- a/src/video_core/shader/memory_util.cpp
+++ b/src/video_core/shader/memory_util.cpp
@@ -1,0 +1,77 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <cstddef>
+
+#include <boost/container_hash/hash.hpp>
+
+#include "common/common_types.h"
+#include "core/core.h"
+#include "video_core/engines/maxwell_3d.h"
+#include "video_core/memory_manager.h"
+#include "video_core/shader/memory_util.h"
+#include "video_core/shader/shader_ir.h"
+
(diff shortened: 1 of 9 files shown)


src/video_core/shader/memory_util.cpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+    // The last instruction is included in the program size
+    return std::min(offset + 1, program.size());
+}
+
+ProgramCode GetShaderCode(Tegra::MemoryManager& memory_manager, GPUVAddr gpu_addr,
+                          const u8* host_ptr, bool is_compute) {
+    ProgramCode code(VideoCommon::Shader::MAX_PROGRAM_LENGTH);
+    ASSERT_OR_EXECUTE(host_ptr != nullptr, {
+        std::fill(code.begin(), code.end(), 0);
```
lioncash, line 62: `ProgramCode` is a `std::vector` automatically, so the construction being performed above already zeroes out its buffer.

src/video_core/shader/memory_util.h
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <cstddef>
+#include <vector>
```
lioncash, line 8: `<vector>` doesn't seem to be used in this header.

RealJohnGalt:
Fixes #3772 caused by #3718

PR #3785, opened 2020-04-24
vi: Don't let uninitialized data pass as a response for SetBufferCount

ogniK5377:
Currently SetBufferCount doesn't write to the out buffer which then contains uninitialized data. This leads to non-zero data which leads to responding with different error codes

### DIFF
--- a/src/core/hle/service/vi/vi.cpp
+++ b/src/core/hle/service/vi/vi.cpp
@@ -267,7 +267,7 @@ protected:
 
 private:
     struct Data {
-        u32_le unk_0;
+        u32_le unk_0{};
     };
 
     Data data{};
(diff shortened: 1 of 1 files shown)


src/core/hle/service/vi/vi.cpp
```diff
@@ -614,6 +614,14 @@ class IHOSBinderDriver final : public ServiceFramework<IHOSBinderDriver> {
             ctx.WriteBuffer(response.Serialize());
             break;
         }
+        case TransactionId::SetBufferCount: {
+            LOG_WARNING(Service_VI, "(STUBBED) called, transaction=SetBufferCount");
+            const auto buffer = ctx.ReadBuffer();
```
lioncash, line 619: You may wish to amend the commit (`git add -u`, then `git commit --amend`), with:

```suggestion
            [[maybe_unused]] const auto buffer = ctx.ReadBuffer();
```

Otherwise this can cause an unused variable warning.

PR #3791, opened 2020-04-24
configuration: Add Restore Default and Clear options to hotkeys

Kewlan:
* This adds Clear and Restore Default for each and all hotkeys:
image

* Shows which action the key sequence is already assigned to, both if you're trying to set a new one or restore one to default:
image
image

* Reorders the hotkeys in config.cpp to actually be in alphabetical order. 

I'm not sure if it's correct to make the key sequence empty, but it works like intended and I'm not getting any errors. It seems Qt deals with that on it's own.

### DIFF
--- a/src/yuzu/configuration/configure_hotkeys.cpp
+++ b/src/yuzu/configuration/configure_hotkeys.cpp
@@ -2,10 +2,12 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <QMenu>
 #include <QMessageBox>
 #include <QStandardItemModel>
 #include "core/settings.h"
 #include "ui_configure_hotkeys.h"
+#include "yuzu/configuration/config.h"
 #include "yuzu/configuration/configure_hotkeys.h"
 #include "yuzu/hotkeys.h"
 #include "yuzu/util/sequence_dialog/sequence_dialog.h"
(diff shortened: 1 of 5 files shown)


src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -82,24 +91,28 @@ void ConfigureHotkeys::Configure(QModelIndex index) {
...
+            this, tr("Conflicting Key Sequence"),
+            tr("The entered key sequence is already assigned to: %1").arg(used_action));
     } else {
         model->setData(index, key_sequence.toString(QKeySequence::NativeText));
     }
 }
 
-bool ConfigureHotkeys::IsUsedKey(QKeySequence key_sequence) const {
-    for (int r = 0; r < model->rowCount(); r++) {
+bool ConfigureHotkeys::IsUsedKey(QKeySequence key_sequence, QString& used_action) const {
```
lioncash, line 105:
```suggestion
std::pair<bool, QString> ConfigureHotkeys::IsUsedKey(QKeySequence key_sequence) const {
```

Then calling sites can just do:

```cpp
const auto [is_used, used_action] = IsUsedKey(sequence);
```

Instead of needing to construct the string separately before calling the function.

bunnei:
Looks good, thank you!

PR #3793, opened 2020-04-25
kernel/physical_core: fix build on non-x86_64 architectures where Dynarmic is not available

zhuowei:


lioncash:
This is resolved in #3783

PR #3799, opened 2020-04-26
shader: Implement P2R CC, IADD Rd.CC and IADD.X

ReinUsesLisp:
Description in the individual commits.

This is a first step on removing the buggy and improperly named `SetInternalFlagsFrom*`. Internal flags should be called "Condition Codes" as described in [NV_gpu_program4](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_gpu_program4.txt), Section 2.X.3.7

### DIFF
--- a/src/video_core/engines/shader_bytecode.h
+++ b/src/video_core/engines/shader_bytecode.h
@@ -814,6 +814,10 @@ union Instruction {
     } alu_integer;
 
     union {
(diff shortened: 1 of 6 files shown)


ReinUsesLisp:
These are the [tests](https://github.com/yuzu-emu/yuzu/files/4534503/nxgputests.zip) that were used for the implementation.

PR #3812, opened 2020-04-28
account: ListQualifiedUsers

### LINKED ISSUE #2844: Missing HLE service command 'ListQualifiedUsers' in 'acc:u'
## log > [log.zip](https://github.com/yuzu-emu/yuzu/files/3581001/log.zip)

<meta name="titleids"  content="01003DD00D658000">

ogniK5377:
Closes #2844

supersedes #3549 as it doesn't have all the interfaces

### DIFF
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -428,6 +428,15 @@ void Module::Interface::GetProfileEditor(Kernel::HLERequestContext& ctx) {
     rb.PushIpcInterface<IProfileEditor>(user_id, *profile_manager);
 }
 
+void Module::Interface::ListQualifiedUsers(Kernel::HLERequestContext& ctx) {
+    LOG_DEBUG(Service_ACC, "called");
+
+    // All users should be qualified
+    ctx.WriteBuffer(profile_manager->GetAllUsers());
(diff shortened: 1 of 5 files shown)


src/core/hle/service/acc/acc.cpp
```diff
@@ -428,6 +428,15 @@ void Module::Interface::GetProfileEditor(Kernel::HLERequestContext& ctx) {
     rb.PushIpcInterface<IProfileEditor>(user_id, *profile_manager);
 }
 
+void Module::Interface::ListQualifiedUsers(Kernel::HLERequestContext& ctx) {
+    LOG_DEBUG(Service_ACC, "called");
+
+    // All users should be qualified
```
lioncash, line 434: This should probably expand on the fact that this is yuzu-specific, given this checks parental control qualifications on the switch if I'm not mistaken.

ogniK5377: I believe it does checks for Nintendo online as well, since we don't implement both we just can assume users are qualified

lioncash: Right, but this should be indicated explicitly in the comment to make it obvious it's known about and not being handled.

ogniK5377: Will add

PR #3818, opened 2020-04-29
Don't fail silently for vi, sm, set and ns services

ogniK5377:
Retail games rarely throw error codes. Actually logging when calls fail helps with debugging/locating more issues in a users logs.


lioncash:
Please squash the latest commit into the previous one. It's directly related to those changes and really doesn't belong as a separate commit, considering the change before it will fundamentally not work without those changes.

PR #3819, opened 2020-04-29
kernel: Don't fail silently

ogniK5377:
Continuation of #3818

### DIFF
--- a/src/core/hle/kernel/process_capability.cpp
+++ b/src/core/hle/kernel/process_capability.cpp
@@ -3,6 +3,7 @@
 // Refer to the license.txt file included.
 
 #include "common/bit_util.h"
+#include "common/logging/log.h"
 #include "core/hle/kernel/errors.h"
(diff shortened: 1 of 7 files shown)


src/core/hle/kernel/process_capability.cpp
```diff
@@ -162,6 +171,7 @@ ResultCode ProcessCapabilities::ParseSingleFlagCapability(u32& set_flags, u32& s
     const u32 flag_length = GetFlagBitOffset(type);
     const u32 set_flag = 1U << flag_length;
     if ((set_flag & set_flags & InitializeOnceMask) != 0) {
+        LOG_ERROR(Kernel, "SetFlags already initialized! set_flags={}", set_flags);
```
lioncash, line 174:
```suggestion
        LOG_ERROR(Kernel, "Attempted to initialize flags that may only be initialized once. set_flags={}", set_flags);
```

PR #3820, opened 2020-04-29
CoreManager: Only run Reschedule() once

Benedani:
It doesn't seem to be necessary to call it 2 times. We verified it in 6 games (SMO, SSBU, Pokemon Sword, SMM2, Tetris 99, ACNH?) using a custom build including this change with no regressions. Feel free to test it on other games, though.

This should be essentially a slight free boost in FPS.

### DIFF
--- a/src/core/core_manager.cpp
+++ b/src/core/core_manager.cpp
@@ -28,6 +28,8 @@ CoreManager::CoreManager(System& system, std::size_t core_index)
 CoreManager::~CoreManager() = default;
 
 void CoreManager::RunLoop(bool tight_loop) {
+    Reschedule();
+
     // If we don't have a currently active thread then don't execute instructions,
     // instead advance to the next event and try to yield to the next thread
     if (Kernel::GetCurrentThread() == nullptr) {
@@ -41,7 +43,6 @@ void CoreManager::RunLoop(bool tight_loop) {
         }
     }
 
-    Reschedule();
     core_timing.Advance();
 }
 
--- a/src/core/core_manager.cpp
+++ b/src/core/core_manager.cpp
@@ -40,7 +40,7 @@ void CoreManager::RunLoop(bool tight_loop) {
             physical_core.Step();
         }
     }
-    
+
     Reschedule();
     core_timing.Advance();
 }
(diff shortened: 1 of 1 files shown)


Benedani:
I honestly don't get what's wrong. I built clang-format, and it made absolutely no changes to the code.

callmenemo491:
it is complainging:
**src/core/core_manager.cpp not compliant to coding style**

it doesn't like the whitespaces on Line 43,  please remove.

Benedani:
So for future reference, I should make sure none of the lines have any whitespace in them?

src/core/core_manager.cpp
```diff
@@ -42,9 +40,9 @@ void CoreManager::RunLoop(bool tight_loop) {
             physical_core.Step();
         }
     }
-    core_timing.Advance();
 
     Reschedule();
+    core_timing.Advance();
```
ogniK5377, line 45: Any reason these were swapped?

Benedani: Originally, it worked without being swapped, but I decided to do so just to be safe. `core_timing.Advance()` is/seems to be what handles events, after all

ogniK5377: Shouldn't functionality remain the same as before?

Benedani: Looked into it a bit further, I think it might handle events for the wrong CPU core if it's done after. Not sure how much of a difference that makes. Tbh this PR is just a result of me messing around with the code

Benedani:
Apparently people are reporting softlocks from the generated build? Strange, it works in my custom build and that works for other people too.

FernandoS27:
Thank you for your contribution, sadly all the cpu management will be remade in a pending PR that will render this pointless. I'll mark it temporally for EA but it won't be merged as to not cause rebase issues.

ogniK5377:
As expected, advancing the core timing after rescheduling causes yielding to fail. Untagging until addressed

ogniK5377:
Upon further investigation, not rescheduling the thread the first time around leads to similar issues

ogniK5377:
Please amend the commit message to better describe the change

Benedani:
> 
> 
> Please amend the commit message to better describe the change



> 
> 
> Please amend the commit message to better describe the change

I'm not sure how to do this on the online interface where I created the commit

Benedani:
I guess I'll create a new, less messy PR then.

Benedani:
Nevermind, it looks like moving Reschedule() before the CPU execution causes a softlock instead. And I don't think this change is worth adding a setting for.

PR #3821, opened 2020-04-29
acc: Fix InitializeApplicationInfo

ogniK5377:
Originally we were popping a u64 and assuming that was the process id. Upon further examination, only a pid is sent, not a u64. As we don't multiprocess support right now, we can ignore the pid and assume it's the current process we're targeting.

Secondly, we don't seem to set a storage ID for a game if we run it from a file. Since we have no way to work out if it's a gamecard or digital game, we can assume it's a digital game


lioncash:
Commit should probably have the `acc:` prefix in it. LGTM otherwise.

ogniK5377:
amended

PR #3822, opened 2020-04-29
acc: Return a unique value per account for GetAccountId

ogniK5377:
Currently we return 1 for a users Nintendo ID. By returning just the lower part of the users UUID, we can have a unique account for each user

### DIFF
--- a/src/common/uuid.h
+++ b/src/common/uuid.h
@@ -40,6 +40,11 @@ struct UUID {
         uuid = INVALID_UUID;
     }
 
+    // TODO(ogniK): Properly generate a Nintendo ID
+    constexpr u64 GetNintendoID() const {
+        return uuid[0];
+    }
+
     std::string Format() const;
     std::string FormatSwitch() const;
 };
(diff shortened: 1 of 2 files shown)


src/common/uuid.h
```diff
@@ -40,6 +40,11 @@ struct UUID {
         uuid = INVALID_UUID;
     }
 
+    // TODO(ogniK): Properly generate a Nintendo ID
+    constexpr u64 GetNintendoID() {
```
lioncash, line 44:
```suggestion
    constexpr u64 GetNintendoID() const {
```

src/core/hle/service/acc/acc.cpp
```diff
@@ -228,7 +228,8 @@ class IProfileEditor final : public IProfileCommon {
 
 class IManagerForApplication final : public ServiceFramework<IManagerForApplication> {
 public:
-    IManagerForApplication() : ServiceFramework("IManagerForApplication") {
+    IManagerForApplication(Common::UUID user_id)
```
lioncash, line 231:
```suggestion
    explicit IManagerForApplication(Common::UUID user_id)
```

PR #3823, opened 2020-04-29
am: IsVrModeEnabled & SetVrModeEnabled fixes

ogniK5377:
Return the proper state of vr mode for IsVrModeEnabled
We should not return an error for SetVrModeEnabled. When VR Mode is turned on, it signals to lbl to turn vr mode on, not return an error code

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -52,11 +52,6 @@ enum class LaunchParameterKind : u32 {
     AccountPreselectedUser = 2,
 };
 
-enum class VrMode : u8 {
-    Disabled = 0,
-    Enabled = 1,
-};
-
 constexpr u32 LAUNCH_PARAMETER_ACCOUNT_PRESELECTED_USER_MAGIC = 0xC79497CA;
 
 struct LaunchParameterAccountPreselectedUser {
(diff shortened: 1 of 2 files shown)


src/core/hle/service/am/am.cpp
```diff
@@ -685,27 +680,31 @@ void ICommonStateGetter::GetCurrentFocusState(Kernel::HLERequestContext& ctx) {
...
 
-    IPC::ResponseBuilder rb{ctx, 2};
-    if (!is_vr_mode_enabled) {
-        rb.Push(RESULT_SUCCESS);
-    } else {
-        // TODO: Find better error code for this
-        UNIMPLEMENTED_MSG("is_vr_mode_enabled={}", is_vr_mode_enabled);
-        rb.Push(RESULT_UNKNOWN);
+    if (last_vr_mode_state != vr_mode_state) {
+        if (vr_mode_state) {
```
lioncash, line 699:
```cpp
LOG_WARNING(Service_AM, "VR Mode is {}", vr_mode_state ? "on" : "off");
```

ogniK5377:
addressed

lioncash:
Doesn't seem to be the case

PR #3827, opened 2020-04-29
externals: Update cubeb to 616d773

ogniK5377:
6 channel support


lioncash:
Commit message should explain why it's being updated.

PR #3842, opened 2020-04-30
maxwell_to_vk: implement missing signed int formats

makigumo:
One is used by _Dark Souls: Remastered_.

Log message:
```
Debug <Critical> video_core\renderer_vulkan\maxwell_to_vk.cpp:VertexFormat:462: Assertion Failed! Unimplemented vertex format of type=3 and size=15
```


ogniK5377:
Have all the remaining formats been tested? If so which games use them

makigumo:
Don't know the games, that are using the other formats. But all of them appear to be in [maxwell_to_gl.h](https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/renderer_opengl/maxwell_to_gl.h#L53-L76) already.

PR #3875, opened 2020-05-03
settings: Add anisotropic filtering level to the yuzu configuration log

Morph1984:
```
Default = 0,
2x = 1,
4x = 2,
8x = 3,
16x = 4,
```


lioncash:
Commit title should be amended to be more specific

goldenx86:
This would help a lot when giving support.

callmenemo491:
@Morph1984 , i think something similar could be done to include in the yuzu.log the amount of RAM the person has...

Fneufneu:
Why keep this obscur ?
can you make some #define that convert 0 1 2 3 4 to a human reading setting: Default , 2x ....

ogniK5377:
> @Morph1984 , i think something similar could be done to include in the yuzu.log the amount of RAM the person has...

Unrelated to this pull request

PR #3884, opened 2020-05-04
vk_sampler_cache: Use VK_EXT_custom_border_color when available

ReinUsesLisp:
This should fix grass interactions on Breath of the Wild on Vulkan.
It is currently untested against validation layers.

Nvidia's Windows 443.09 beta driver or Linux 440.66.12 is required for
now.


Fneufneu:
and AMD users can't use vulkan anymore ? :(

ReinUsesLisp:
The extension is only used when it's available, it's in the title.

PR #3892, opened 2020-05-06
Port citra-emu/citra#5280: "Small fixes to "Option to hide mouse on inactivity""

FearlessTobi:
See citra-emu/citra#5280 for more details.

**Original description**:
Follow up to citra-emu/citra#5094.
Fixes a bug reported on discord by Dragois, and backports review comments from yuzu-emu/yuzu#3637.
The reported bug was that the menubar would lose sensitivity to the hover interactions. This happens because Qt menubars have mouse tracking true as default, therefore I had operated on false assumptions. 
There are no changes in behaviour besides the bug fix.


theboy181:
Thanks. Was totally going to request this

PR #3904, opened 2020-05-09
settings: Log anisotropic filtering and gpu accuracy as readable strings

Morph1984:
Log them as readable strings instead of their respective combobox indices.

### Example:

Before:
```
Config <Info> core\settings.cpp:LogSetting:80: Renderer_GPUAccuracyLevel: 1
Config <Info> core\settings.cpp:LogSetting:80: Renderer_AnisotropicFilteringLevel: 4
```

After
```
Config <Info> core\settings.cpp:LogSetting:80: Renderer_GPUAccuracyLevel: High
Config <Info> core\settings.cpp:LogSetting:80: Renderer_AnisotropicFilteringLevel: 16x
```


FearlessTobi:
Shouldn't parsing that number be the job of the log parser. 
Right now, I only see this adding extra code without a clear benefit.

PR #3905, opened 2020-05-10
Correct a series of crashes and intructions on Async GPU and Vulkan Pipeline

FernandoS27:
This PR takes care of correcting a series of crashes and incorrections on Async GPU & Vulkan.
Resume of changes:
- Remove CPU Query Cache Invalidation on Async GPU, as Async GPU always flushes and invalidates them.
- On Async GPU, Shader & Pipeline Invalidations from the CPU will no longer be instant and instead use the sync interface.
- Corrected an issue in Vulkan's Pipeline Cache where if a shader had an invalid address, it wouldn't use a null shader instead of trying to decompile it.

### DIFF
--- a/src/video_core/rasterizer_cache.h
+++ b/src/video_core/rasterizer_cache.h
@@ -75,8 +75,8 @@ public:
 private:
     bool is_registered{};      ///< Whether the object is currently registered with the cache
     bool is_dirty{};           ///< Whether the object is dirty (out of sync with guest memory)
-    bool is_memory_marked{};   ///< Whether it's marking rasterizer memory.
-    bool is_sync_pending{};    ///< Whether it's pending deletion.
+    bool is_memory_marked{};   ///< Whether the object is marking rasterizer memory.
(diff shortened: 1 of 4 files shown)


src/video_core/rasterizer_cache.h
```diff
@@ -56,9 +56,27 @@ class RasterizerCacheObject {
...
+    }
+
+    bool IsSyncPending() const {
+        return is_sync_pending;
+    }
+
 private:
     bool is_registered{};      ///< Whether the object is currently registered with the cache
     bool is_dirty{};           ///< Whether the object is dirty (out of sync with guest memory)
+    bool is_memory_marked{};   ///< Whether it's marking rasterizer memory.
```
lioncash, line 78: If `it's` refers to `the object` in the other comments, `it's` should be changed to that to remain consistent with the other documentation.

bunnei:
@FernandoS27 is this ready to be merged? Can you address the comment above? Thank you!

PR #3909, opened 2020-05-10
Improve time zone support

bunnei:
Various improvements to time zone support. Tested against ACNH. 

* Tries to detect current time zone and use that. Windows only for now, and fairly incomplete.
* Adds a setting to manually change the time zone.

### DIFF
--- a/src/common/time_zone.cpp
+++ b/src/common/time_zone.cpp
@@ -37,13 +37,13 @@ static int ConvertOsTimeZoneOffsetToInt(const std::string& timezone) {
     }
 }
 
-int GetCurrentOffsetSeconds() {
+std::chrono::seconds GetCurrentOffsetSeconds() {
     const int offset{ConvertOsTimeZoneOffsetToInt(GetOsTimeZoneOffset())};
 
     int seconds{(offset / 100) * 60 * 60}; // Convert hour component to seconds
     seconds += (offset % 100) * 60;        // Convert minute component to seconds
 
-    return seconds;
+    return std::chrono::seconds{seconds};
 }
 
 } // namespace Common::TimeZone
--- a/src/common/time_zone.cpp
+++ b/src/common/time_zone.cpp
@@ -1,0 +1,49 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <chrono>
+#include <iomanip>
+#include <sstream>
+
+#include "common/logging/log.h"
+#include "common/time_zone.h"
+
+namespace Common::TimeZone {
+
+std::string GetDefaultTimeZone() {
+    return "GMT";
+}
+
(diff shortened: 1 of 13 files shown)


src/common/time_zone.cpp
```diff
@@ -0,0 +1,276 @@
...
+    map_zone["UTC+12"] = "NZ";
+
+    // (UTC+12:00) Fiji
+    map_zone["Fiji Standard Time"] = "NZ";
+
+    return map_zone;
+}
+
+static std::string ConvertWindowsTimeZone(const std::string& windows_timezone) {
+    static const std::map<std::string, std::string> map_zone{GenerateWindowsTimezoneTable()};
```
lioncash, line 221: If it's desirable to keep this as a map, you may wish to opt for making this:

```cpp
std::map<std::string_view, std::string_view>
```

Which will allow the input parameter to be a std::string_view as well.

src/common/time_zone.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+static std::string GetWindowsTimeZone() {
+    DYNAMIC_TIME_ZONE_INFORMATION dtzi{};
+    const auto result = GetDynamicTimeZoneInformation(&dtzi);
+    if (result == TIME_ZONE_ID_INVALID) {
+        UNREACHABLE_MSG("result is invalid");
+        return {};
+    }
+
+    const auto wlen = wcslen(dtzi.TimeZoneKeyName);
```
lioncash, line 239:
```suggestion
    const auto wlen = std::wcslen(dtzi.TimeZoneKeyName);
```

src/common/time_zone.cpp
```diff
@@ -0,0 +1,276 @@
...
+    char buf[128] = {};
+
+    ASSERT(sizeof(buf) >= wlen + 1);
+
+    wcstombs(buf, dtzi.TimeZoneKeyName, wlen);
```
lioncash, line 244:
```suggestion
    std::wcstombs(buf, dtzi.TimeZoneKeyName, wlen);
```

src/common/time_zone.cpp
```diff
@@ -0,0 +1,276 @@
...
+    if (strcmp(buf, "Coordinated Universal Time") == 0) {
```
lioncash, line 245:
```suggestion
    if (std::strcmp(buf, "Coordinated Universal Time") == 0) {
```

src/common/time_zone.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <string>
+
+namespace Common::TimeZone {
+
+/// Gets the default timezone, i.e. "GMT"
+std::string GetDefaultTimeZone();
+
+/// Gets the offset of the current timezone (from the default), in seconds
+int GetCurrentOffsetSeconds();
```
lioncash, line 15: Perhaps return this as `std::chrono::seconds` to make the interface strongly typed by default in the common library.

PR #3926, opened 2020-05-11
hid: Clear keyboard states & fix logic issue

ogniK5377:
Previously we never cleared the states of the entries and the key would stay held down, also looping over the key bytes for each key lead to setting every bit for the key state instead of the key we wanted


src/core/hle/service/hid/controllers/keyboard.cpp
```diff
@@ -38,10 +38,11 @@ void Controller_Keyboard::OnUpdate(const Core::Timing::CoreTiming& core_timing,
     cur_entry.sampling_number = last_entry.sampling_number + 1;
     cur_entry.sampling_number2 = cur_entry.sampling_number;
 
+    std::memset(cur_entry.key.data(), 0, cur_entry.key.size());
```
lioncash, line 41:
```suggestion
    cur_entry.key.fill(0);
```

PR #3929, opened 2020-05-12
file_sys/nsp: Make SetTicketKeys actually do something

FearlessTobi:
Previously, the method wasn't modifying any class state and therefore not having any effects when called.
Since this has been the case for a very long time now, I'm not sure if we couldn't just remove this method altogether.

### DIFF
--- a/src/core/file_sys/submission_package.cpp
+++ b/src/core/file_sys/submission_package.cpp
@@ -19,38 +19,6 @@
 #include "core/loader/loader.h"
 
 namespace FileSys {
-namespace {
-void SetTicketKeys(const std::vector<VirtualFile>& files) {
-    auto& keys = Core::Crypto::KeyManager::Instance();
-
-    for (const auto& ticket_file : files) {
-        if (ticket_file == nullptr) {
-            continue;
-        }
-
-        if (ticket_file->GetExtension() != "tik") {
-            continue;
-        }
-
-        if (ticket_file->GetSize() <
-            Core::Crypto::TICKET_FILE_TITLEKEY_OFFSET + sizeof(Core::Crypto::Key128)) {
-            continue;
-        }
-
-        Core::Crypto::Key128 key{};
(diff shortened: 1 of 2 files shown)


src/core/file_sys/submission_package.h
```diff
@@ -26,6 +26,8 @@ class NSP : public ReadOnlyVfsDirectory {
     explicit NSP(VirtualFile file);
     ~NSP() override;
 
+    void SetTicketKeys(const std::vector<VirtualFile>& files);
```
lioncash, line 29: This can be made private.

FearlessTobi: Done.

Morph1984:
This *should* fix some reported issues of games not detected by yuzu even after being redumped and with updated keys.

src/core/file_sys/submission_package.cpp
```diff
@@ -232,6 +200,35 @@ VirtualDir NSP::GetParentDirectory() const {
...
+        if (ticket_file->GetSize() <
+            Core::Crypto::TICKET_FILE_TITLEKEY_OFFSET + sizeof(Core::Crypto::Key128)) {
+            continue;
+        }
+
+        Core::Crypto::Key128 key{};
+        ticket_file->Read(key.data(), key.size(), Core::Crypto::TICKET_FILE_TITLEKEY_OFFSET);
+
+        // We get the name without the extension in order to create the rights ID.
+        std::string name_only(ticket_file->GetName());
```
VolcaEM, line 222: I think you can use

`const std::string name_only = ticket_file->GetName().substr(0, ticket_file->GetName().size() - 4) `

FearlessTobi: As I was just moving this code, I won't be making any changes to it, especially no micro-optimizations.

VolcaEM: Ok, I'll make a PR when this gets merged, then(?)

FearlessTobi:
Please give this a final review so we can get this merged soon. 
It is blocking some other stuff I'm working on.

bunnei (approved):

PR #3932, opened 2020-05-14
gpu_thread: Invalidate asynchronously.

bunnei:
WIP to see if this helps with Vulkan crashes -- Do not merge


RealJohnGalt:
Fine with vulkan for me, but with opengl it hangs SMO and BOTW on menu.

PR #3938, opened 2020-05-15
shader: Misc implementations used in Astral Chain

ReinUsesLisp:
Description in individual commits:
- vk_shader_decompiler: Don't assert for void returns
- shader_decompiler: Visit source nodes even when they assign to RZ
- shader/memory: Implement non-addition operations in RED 
- shader/other: Implement thread comparisons (NV_shader_thread_group)
- shader/other: Implement BAR.SYNC 0x0
- shader/other: Implement MEMBAR.CTS

No known visual difference is caused by this pull request.


FernandoS27:
Seems to be causing crashes on Astral Chain for some people. Removed temporally.

ReinUsesLisp:
Separated in different pull requests.

PR #3953, opened 2020-05-17
Add button to show and update current user profile on status bar

GodKratos:
This makes it very easy to see what user profile is active with a status bar button and switch between them by clicking the button.
Very useful when managing saves for multiple users!

This is my first time coding in C++ so I hope everything is ok.
Happy to amend in any way necessary.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -441,6 +441,20 @@ void GMainWindow::WebBrowserOpenPage(std::string_view filename, std::string_view
 
 #endif
 
+QString GetAccountUsername() {
+    Service::Account::ProfileManager manager;
+    const auto current_user = manager.GetUser(Settings::values.current_user);
+    ASSERT(current_user);
+    Service::Account::ProfileBase profile;
+    if (!manager.GetProfileBase(*current_user, profile)) {
+        return {};
+    }
+
+    const auto text = Common::StringFromFixedZeroTerminatedBuffer(
+        reinterpret_cast<const char*>(profile.username.data()), profile.username.size());
+    return QString::fromStdString(text);
+}
+
 void GMainWindow::InitializeWidgets() {
 #ifdef YUZU_ENABLE_COMPATIBILITY_REPORTING
     ui.action_Report_Compatibility->setVisible(true);
@@ -492,6 +506,48 @@ void GMainWindow::InitializeWidgets() {
         label->setContentsMargins(4, 0, 4, 0);
         statusBar()->addPermanentWidget(label);
     }
+
+    // Setup Profile button
+    profile_status_button = new QPushButton();
+    profile_status_button->setObjectName(QStringLiteral("TogglableStatusBarButton"));
+    profile_status_button->setCheckable(true);
+    profile_status_button->setChecked(true);
+    profile_status_button->setFocusPolicy(Qt::NoFocus);
+    const auto username = GetAccountUsername();
+    profile_status_button->setText(username);
+    connect(profile_status_button, &QPushButton::clicked, [=] {
+        profile_status_button->setChecked(true);
+
+        if (emulation_running) {
+            return;
+        }
+
+        // User save data
+        const auto select_profile = [this] {
+            QtProfileSelectionDialog dialog(this);
+            dialog.setWindowFlags(Qt::Dialog | Qt::CustomizeWindowHint | Qt::WindowTitleHint |
+                                  Qt::WindowSystemMenuHint | Qt::WindowCloseButtonHint);
+            dialog.setWindowModality(Qt::WindowModal);
+
+            if (dialog.exec() == QDialog::Rejected) {
+                return -1;
+            }
+
+            return dialog.GetIndex();
+        };
+
+        const auto index = select_profile();
+        if (index == -1) {
+            return;
+        }
+
+        Settings::values.current_user = index;
+        Settings::Apply();
+
+        const auto username = GetAccountUsername();
+        profile_status_button->setText(username);
+    });
+    statusBar()->insertPermanentWidget(0, profile_status_button);
 
     // Setup Dock button
     dock_status_button = new QPushButton();
@@ -1902,6 +1958,8 @@ void GMainWindow::OnConfigure() {
         ui.centralwidget->setMouseTracking(false);
     }
 
+    const auto username = GetAccountUsername();
+    profile_status_button->setText(username);
     dock_status_button->setChecked(Settings::values.use_docked_mode);
     async_status_button->setChecked(Settings::values.use_asynchronous_gpu_emulation);
 #ifdef HAS_VULKAN
--- a/src/yuzu/main.h
+++ b/src/yuzu/main.h
@@ -233,6 +233,7 @@ private:
     QLabel* emu_speed_label = nullptr;
     QLabel* game_fps_label = nullptr;
     QLabel* emu_frametime_label = nullptr;
+    QPushButton* profile_status_button = nullptr;
     QPushButton* async_status_button = nullptr;
     QPushButton* renderer_status_button = nullptr;
     QPushButton* dock_status_button = nullptr;


VolcaEM:
Has this been tested (and works)?
And does this change anything in the GUI? If so, please add a screenshot of the changes

src/yuzu/main.cpp
```diff
@@ -135,6 +135,24 @@ __declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
 }
 #endif
 
+namespace {
+
+QString GetAccountUsername() {
+    Service::Account::ProfileManager manager;
+    const auto current_user = manager.GetUser(Settings::values.current_user);
+    ASSERT(current_user);
```
VolcaEM, line 143: Asserts get ignored in Mainline and Early Access build, so asserting here is not a good idea tbh

I'd suggest you return {} instead here, if the current_user has no value or is Common::UUID{}

src/yuzu/main.cpp
```diff
@@ -493,6 +511,48 @@ void GMainWindow::InitializeWidgets() {
...
+
+        const auto index = select_profile();
+        if (index == -1) {
+            return;
+        }
+
+        Settings::values.current_user = index;
+        Settings::Apply();
+
+        const auto username = GetAccountUsername();
```
VolcaEM, line 551: Please add a check to see if the username is valid or not before setting it

src/yuzu/main.cpp
```diff
@@ -1902,6 +1962,8 @@ void GMainWindow::OnConfigure() {
         ui.centralwidget->setMouseTracking(false);
     }
 
+    const auto username = GetAccountUsername();
```
VolcaEM, line 1965: See above

src/yuzu/main.cpp
```diff
@@ -493,6 +511,48 @@ void GMainWindow::InitializeWidgets() {
         statusBar()->addPermanentWidget(label);
     }
 
+    // Setup Profile button
+    profile_status_button = new QPushButton();
+    profile_status_button->setObjectName(QStringLiteral("TogglableStatusBarButton"));
+    profile_status_button->setCheckable(true);
+    profile_status_button->setChecked(true);
+    profile_status_button->setFocusPolicy(Qt::NoFocus);
+    const auto username = GetAccountUsername();
```
VolcaEM, line 520: See above (forgot this one)

GodKratos:
> 
> 
> Has this been tested (and works)? (I haven't looked at the code for now)
> And does this change anything in the GUI? If so, please add a screenshot of the changes

Yes I've tested and it works fine and updates the profiles saves correctly.
It adds a button on the status bar that shows the current user profile name which opens up a profile selector when clicked.

image

GodKratos:
@VolcaEM does that resolve the issues you raised?

src/yuzu/main.cpp
```diff
@@ -135,6 +135,28 @@ __declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
 }
 #endif
 
+namespace {
+
+QString GetAccountUsername() {
+    const QString nouser = QString::fromStdString("No User");
+    Service::Account::ProfileManager manager;
+    const auto current_user = manager.GetUser(Settings::values.current_user);
+    if (!current_user.has_value()) {
```
VolcaEM, line 144:
```suggestion
    if (!current_user.has_value() && (current_user != Common::UUID{})) {
```

GodKratos:
@VolcaEM is this good to go now?

src/yuzu/main.cpp
```diff
@@ -135,6 +135,28 @@ __declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
 }
 #endif
 
+namespace {
+
+QString GetAccountUsername() {
```
VolcaEM, line 140: I'd use std::string here

src/yuzu/main.cpp
```diff
@@ -135,6 +135,28 @@ __declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
...
+    const QString nouser = QString::fromStdString("No User");
```
VolcaEM, line 141:
```suggestion
    const std::string nouser = "No User";
```

src/yuzu/main.cpp
```diff
@@ -135,6 +135,28 @@ __declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
...
+    }
+    Service::Account::ProfileBase profile;
+    if (!manager.GetProfileBase(*current_user, profile)) {
+        return nouser;
+    }
+
+    const auto text = Common::StringFromFixedZeroTerminatedBuffer(
+        reinterpret_cast<const char*>(profile.username.data()), profile.username.size());
+
+    return text.empty() ? nouser : QString::fromStdString(text);
```
VolcaEM, line 155:
```suggestion
    return text.empty() ? nouser : text;
```

src/yuzu/main.cpp
```diff
@@ -493,6 +515,48 @@ void GMainWindow::InitializeWidgets() {
...
     }
 
+    // Setup Profile button
+    profile_status_button = new QPushButton();
+    profile_status_button->setObjectName(QStringLiteral("TogglableStatusBarButton"));
+    profile_status_button->setCheckable(true);
+    profile_status_button->setChecked(true);
+    profile_status_button->setFocusPolicy(Qt::NoFocus);
+    const auto username = GetAccountUsername();
+    profile_status_button->setText(username);
```
VolcaEM, line 525:
```suggestion
    profile_status_button->setText(QString::fromStdString(username));
```

src/yuzu/main.cpp
```diff
@@ -1902,6 +1966,8 @@ void GMainWindow::OnConfigure() {
         ui.centralwidget->setMouseTracking(false);
     }
 
+    const auto username = GetAccountUsername();
+    profile_status_button->setText(username);
```
VolcaEM, line 1970:
```suggestion
    profile_status_button->setText(QString::fromStdString(username));
```

GodKratos:
@VolcaEM do you really think that change is worthwhile? Everywhere that function is called requires a QString so it gets converted after every call instead of being converted once within the function? The only benefit I can see is of nouser being stored as a std:string for the brief moment it exists but that wouldn't achieve a lot...

VolcaEM:
Now that I really think about it, no, it's fine like this; I thought yuzu didn't use QString if not when necessary but I was wrong, it's used a lot. Sorry for not noticing this earlier.
The only issue I have now with this PR is that GetAccountUsername should be moved somewhere else, like under ProfileSelectorSelectProfile

GodKratos:
Yes I agree @VolcaEM, I thought that that function should be more globally accessible but wasn't familiar enough with the codebase to know exactly where it should go. Looking at the profile settings class I thought it actually needed a rework to make the profile configurations more global but wasn't confident enough in performing all of that as part of my first attempt at C++.
I think as far as this feature add is concerned how I've done it is good for now and the enhancement you suggest should be considered as part of a larger rework of the profile system.

GodKratos:
@VolcaEM, so now what happens?

VolcaEM:
Wait until someone else comments/until this gets merged, I can't merge this

GodKratos:
@bunnei @ReinUsesLisp how do these things get committed?

ogniK5377:
Is there any reasoning behind this PR? Generally, users have one profile and don't swap profiles for each game. Personally I don't think it should be in the status bar where users modify the most used options

GodKratos:
> 
> 
> Is there any reasoning behind this PR? Generally, users have one profile and don't swap profiles for each game. Personally I don't think it should be in the status bar where users modify the most used options

How can you generalise like that to know how everyone uses the app? I have two kids and myself who all play on it and we switch between profiles for different saves. I'm sure there will be plenty of other users who also do something similar.

Firstly, you can't even tell which profile is active when you go into settings let alone when you open the app so unless you are sure who was playing last you have to go into settings and select a profile to use every time you start the app to get the right saves. Having the profile visible straight away and being able to switch instantly is extremely convenient in that situation.

FearlessTobi:
Doesn't the "Prompt for user on Game Boot" option then already accomplish what you need?

GodKratos:
> 
> 
> Doesn't the "Prompt for user on Game Boot" option then already accomplish what you need?

Huh, I'd never actually noticed that option before, how long has that been there?
That works but TBH, that's kind of annoying when the same person plays multiple times. It's much nicer being able to see which profile is active at a glance and just changing it when necessary.

PR #3966, opened 2020-05-20
yuzu/frontend: Remove internal resolution option

Morph1984:
Removes the internal resolution combobox to minimize user confusion as this setting is currently non-functional.

image

### DIFF
--- a/src/core/settings.h
+++ b/src/core/settings.h
@@ -437,7 +437,7 @@ struct Values {
     bool renderer_debug;
     int vulkan_device;
 
-    float resolution_factor;
+    u16 resolution_factor{1};
     int aspect_ratio;
     int max_anisotropy;
     bool use_frame_limit;
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -689,10 +689,7 @@ void GMainWindow::InitializeHotkeys() {
                 Settings::values.use_frame_limit = !Settings::values.use_frame_limit;
                 UpdateStatusBar();
             });
-    // TODO: Remove this comment/static whenever the next major release of
-    // MSVC occurs and we make it a requirement (see:
-    // https://developercommunity.visualstudio.com/content/problem/93922/constexprs-are-trying-to-be-captured-in-lambda-fun.html)
-    static constexpr u16 SPEED_LIMIT_STEP = 5;
+    constexpr u16 SPEED_LIMIT_STEP = 5;
     connect(hotkey_registry.GetHotkey(main_window, QStringLiteral("Increase Speed Limit"), this),
             &QShortcut::activated, this, [&] {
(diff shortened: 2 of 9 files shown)


FearlessTobi:
Why can't we just remove the code from the ui file instead of having to hide it in code?
It's not like adding it back would be a huge task.

superheri:
Instead of removing it, why not implement it instead? This combobox was added for a reason...

SilverBeamx:
Please, refrain from such snarky and pointless comments. The feature was removed because it was not ready, and it will be re-implemented if and when it'll work properly. Also, yuzu devs have different priorities. Being rude and ungrateful won't change or speed this up.

bunnei:
+1 to Tobi's comment... let's just remove it and add it back when we are ready

bunnei:
@Morph1984 ping

src/yuzu/main.cpp
```diff
@@ -689,10 +689,7 @@ void GMainWindow::InitializeHotkeys() {
                 Settings::values.use_frame_limit = !Settings::values.use_frame_limit;
                 UpdateStatusBar();
             });
-    // TODO: Remove this comment/static whenever the next major release of
-    // MSVC occurs and we make it a requirement (see:
-    // https://developercommunity.visualstudio.com/content/problem/93922/constexprs-are-trying-to-be-captured-in-lambda-fun.html)
-    static constexpr u16 SPEED_LIMIT_STEP = 5;
```
FearlessTobi, line 695: Did you actually verify this was fixed?

src/yuzu/main.cpp
```diff
@@ -1989,7 +1986,7 @@ void GMainWindow::OnCaptureScreenshot() {
         const QString path = png_dialog.selectedFiles().first();
         if (!path.isEmpty()) {
             UISettings::values.screenshot_path = QFileInfo(path).path();
-            render_window->CaptureScreenshot(UISettings::values.screenshot_resolution_factor, path);
+            render_window->CaptureScreenshot(Settings::values.resolution_factor, path);
```
FearlessTobi, line 1989: I don't understand why this was changed.

src/core/settings.h
```diff
@@ -437,7 +437,7 @@ struct Values {
     bool renderer_debug;
     int vulkan_device;
 
-    float resolution_factor;
+    u16 resolution_factor{1};
```
FearlessTobi, line 440: Are we still using this anywhere?

Morph1984: Yes, in vi, am, video_core, and its still logged.

src/yuzu/main.cpp
Morph1984: Yes it's fixed in VS 2019

PR #3967, opened 2020-05-20
crypto: Make KeyManager a singleton class

FearlessTobi:
Previously, we were reading the keys every time a KeyManager object was created, causing yuzu to reread the keys file multiple hundreds of times when loading the game list.
With this change, it is only loaded once.
**On my system, this decreased game list loading times by a factor of 20.**

One thing I need to check is if key rederivation is still working after this change.
This is a bit hard to do because it apparently isn't working even on yuzu master for me.


RealJohnGalt:
I tested on an EA build with this change in a new yuzu prefix. Fresh key initialization worked for me if that's what you mean by "key rederivation."

ogniK5377:
I have spoken with @DarkLordZach and have decided to merge this for the time being

PR #3976, opened 2020-05-22
qt_themes: remove unknown qss property from dark theme

Neodyblue:
Qdarkstyle's qss file uses an overflow property.
According to `https://doc.qt.io/qt-5/stylesheet-reference.html`,
the property `overflow` doesn't exist, which leads to a warning message
in the console.

### DIFF
--- a/dist/qt_themes/qdarkstyle/style.qss
+++ b/dist/qt_themes/qdarkstyle/style.qss
@@ -673,10 +673,6 @@ QTabWidget::pane {
     border-bottom-left-radius: 2px;
 }
 
-QTabWidget::tab-bar {
-    overflow: visible;
-}
-
 QTabBar {
     qproperty-drawBase: 0;
     border-radius: 3px;


FearlessTobi:
@Simek Was there a specific reason you added this property?

Simek:
@FearlessTobi I'm not sure right now, it might by just to ensure the "hidden" behaviour which seems to be default behaviour in QT anyway. In CSS this property hides the content which exceeds the boundaries on the parent element or container. 

So if QT do not support "overflow" the line should be safe to remove without affecting the current UI.

BTW. When I was working on UI tweaks I have used this guide as a reference: https://doc.qt.io/qt-5/stylesheet-examples.html#customizing-qtabwidget-and-qtabbar

Neodyblue:
> @FearlessTobi I'm not sure right now, it might by just to ensure the "hidden" behaviour which seems to be default behaviour in QT anyway. In CSS this property hides the content which exceeds the boundaries on the parent element or container.

Yes, however to achieve that you need `overflow` to be set to `hidden` or `scroll`, here it was set to `visible` which would have rendered outside of the container.

PR #3986, opened 2020-05-23
shader_cache: Implement a generic runtime shader cache

ReinUsesLisp:
The shader cache is barely invalidated but we were using an approach that had the cost of a full cache like the buffer or texture caches. To avoid paying the extra cost, use a page based hash table for faster invalidation lookups.

Implement a generic shader cache for fast lookups and invalidations.
Invalidations lookups are cheap but expensive when a shader is invalidated.

Use two mutexes instead of one to avoid locking invalidations for
lookups and vice versa. When a shader has to be removed, lookups are
locked as expected.

Drop the usage of shared pointers. Pass star pointers around and store in unique pointers.


src/video_core/shader_cache.h
```diff
@@ -0,0 +1,171 @@
...
+#include "common/assert.h"
+#include "common/common_types.h"
+#include "video_core/rasterizer_interface.h"
+
+namespace VideoCommon {
+
+template <class T>
+class ShaderCache {
+    static constexpr u64 PAGE_SHIFT = 14;
+    static constexpr u64 PAGE_SIZE = u64(1) << PAGE_SHIFT;
```
VolcaEM, line 23: I'd use static_cast but this is fine too

PR #3991, opened 2020-05-25
texture_cache: Implement depth stencil texture swizzles

ReinUsesLisp:
Stop ignoring image swizzles on depth and stencil images.

This doesn't fix a known issue on Xenoblade Chronicles 2 where an OpenGL
texture changes swizzles twice before being used. A proper fix would be
having a small texture view cache for this like we do on Vulkan.


ReinUsesLisp:
> This doesn't fix a known issue on Xenoblade Chronicles 2 where an OpenGL
texture changes swizzles twice before being used. A proper fix would be
having a small texture view cache for this like we do on Vulkan.

This is fixed now.
The last commit fixes texture swizzle issues on Xenoblade Chronicles 2.

ReinUsesLisp:
Untagging because there's an unrelated issue.

ReinUsesLisp:
Addressed in this same pull request. Tagging again.

PR #3996, opened 2020-05-26
fixed_pipeline_state,gl_rasterizer: Swap negative viewport checks for front faces

ReinUsesLisp:
These fixes are based on gdkchan's explanation of the problem, found in [this](https://github.com/Ryujinx/Ryujinx/pull/1277) pull request from Ryujinx.

The implementation is different. On Vulkan I removed a leftover from OpenGL and everything works fine, on OpenGL I ported an the old Vulkan code back. As described in 606a62d, this is not a complete implementation because we only check for the first viewport.
The reason I didn't port gdkchan's solution is because I don't want to use viewport swizzles on games not using viewport swizzles and injecting shader code to flip `gl_Position` can introduce issues on transform feedbacks.

This fixes front face issues on Xenoblade Chronicles 2 and any other game doing bad things on NVN queues.

Credits to gdkchan from [Ryujinx](ryujinx.org/) for finding what causes this issue.

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -1031,6 +1031,26 @@ void RasterizerOpenGL::SyncViewport() {
     const auto& regs = gpu.regs;
 
     const bool dirty_viewport = flags[Dirty::Viewports];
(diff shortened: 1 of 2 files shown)


FernandoS27:
Weird I tried this 2 months ago but it had a few regressions, so I abandoned it. Would be nice to test if it has any regressions.

PR #3999, opened 2020-05-27
texture_cache: Optimize GetSurfacesInRegion

ReinUsesLisp:
Avoid unnecessary work on GetSurfacesInRegion using `std::unordered_map::find` instead of `operator[]` and avoid heap allocations using `boost::container::small_vector` with a size of one.

While we are at it, use standard algorithms when possible.


ReinUsesLisp:
Profiling Xenoblade Chronicles 2:

master
imagen

After d2b2557
imagen

After b8b6f94
imagen

PR #4002, opened 2020-05-28
patch_manager: Add support for case-sensitivity on Linux

lat9nq:
Some patch developers distribute their patches with odd-cased subdirectories. On Windows systems, this is fine: file names do not have case, so 'exefs' and 'exeFS' refer to the same directory. On most Linux systems however, case matters when naming files and directories. Thus, the current patch manager implementation in Linux does not recognize mods that use alternate-cases when naming specifically the `exefs`, `romfs`, `romfs_ext`, and `cheats` directories.

This changes patch_manager functions to use a case-insensitive variant of `GetSubdirectory`. With this change, patches now work on *nix systems even when patch directories are named with odd cases.

I am obviously open to changes, especially moving `FindSubdirectoryCaseless` to a different file. It is in patch_manager because I was afraid changing a more core VFS function would have other consequences, and as far as I know the rest of the file system interactions work fine on Linux without need for case handling.

### DIFF
--- a/src/core/file_sys/patch_manager.cpp
+++ b/src/core/file_sys/patch_manager.cpp
@@ -56,8 +56,7 @@ std::shared_ptr<VfsDirectory> FindSubdirectoryCaseless(const std::shared_ptr<Vfs
 #else
     const auto subdirs = dir->GetSubdirectories();
     for (const auto& subdir : subdirs) {
-        std::string dir_name = subdir->GetName();
-        dir_name = Common::ToLower(dir_name);
+        std::string dir_name = Common::ToLower(subdir->GetName());
         if (dir_name == name) {
             return subdir;
         }
--- a/src/core/file_sys/patch_manager.cpp
+++ b/src/core/file_sys/patch_manager.cpp
@@ -6,6 +6,7 @@
 #include <array>
 #include <cstddef>
 #include <cstring>
+#include <boost/algorithm/string/case_conv.hpp>
 
 #include "common/file_util.h"
 #include "common/hex_util.h"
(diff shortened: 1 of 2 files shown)


src/core/file_sys/patch_manager.h
```diff
@@ -29,6 +29,11 @@ enum class TitleVersionFormat : u8 {
 std::string FormatTitleVersion(u32 version,
                                TitleVersionFormat format = TitleVersionFormat::ThreeElements);
 
+// Returns a directory with name matching name case-insensitive. Returns nullptr if directory
+// doesn't have a directory with name.
+std::shared_ptr<VfsDirectory> FindSubdirectoryCaseless(const std::shared_ptr<VfsDirectory> dir,
+                                                       const std::string& name);
```
lioncash, line 35:
```suggestion
                                                       std::string_view name);
```

src/core/file_sys/patch_manager.cpp
```diff
@@ -48,6 +49,24 @@ std::string FormatTitleVersion(u32 version, TitleVersionFormat format) {
...
 
+std::shared_ptr<VfsDirectory> FindSubdirectoryCaseless(const std::shared_ptr<VfsDirectory> dir,
+                                                       const std::string& name) {
+#ifdef _WIN32
+    return dir->GetSubdirectory(name);
+#else
+    const auto subdirs = dir->GetSubdirectories();
+    for (const auto& subdir : subdirs) {
+        std::string dir_name = subdir->GetName();
+        boost::algorithm::to_lower(dir_name);
```
lioncash, line 60: We have our own ToLower function in `common/string_util.h` that can be used instead of boost.

bunnei:
LGTM, thanks!

Issue #3342, opened 2020-01-23
Texture Format: G8R24 unimplemented Luigi's Mansion 3 v1.2.0


CJBok:
Following log is spammed in Luigi's Mansion 3 V1.2.0

`Debug <Critical> video_core\texture_cache\format_lookup_table.cpp:GetPixelFormat:182: Assertion Failed! texture format=14 srgb=false components={4 2 2 2}`

Usefull for GPU Project?

FernandoS27:
Sadly there's no Vulkan/OpenGL equivalent. It needs to be converted by hand.

Morph1984:
@CJBok This is G24R8, not G8R24

Morph1984:
#3020 related

MysticExile:
Fixed by #4005

Issue #3020, opened 2019-10-23
Missing GPU texture format 'G24R8_UINT'


GDhex12:
* Needed for Witcher 3

<meta name="titleids"  content="0100bfe00e9ca000">

MysticExile:
Fixed by #4005

PR #4005, opened 2020-05-28
format_lookup_table: Implement G24S8 format as S8Z24

ReinUsesLisp:
Implement G24S8 as S8Z24, this is deduced from the component type. If sampling individual components has to be implemented has not been tested.


Morph1984:
Closes #3342 and #3020

PR #4006, opened 2020-05-28
glsl: Squash constant buffers into a single SSBO when we hit the limit

ReinUsesLisp:
Avoids compilation errors at the cost of shader build times and runtime
performance when a game hits the limit of uniform buffers we can use.


bunnei:
@ReinUsesLisp can you rebase this please?

PR #4008, opened 2020-05-28
texture_cache: Always load textures on Recycle

ReinUsesLisp:
This is a temporary fix. We should handle textures that have the
candidate as a super texture and they are an inner mipmap with copies.


ReinUsesLisp:
#4012 fixes this in a better way.

PR #4010, opened 2020-05-29
kernel: ResourceLimit::Reserve remove useless while loop

ogniK5377:
Timeout is a u64, it will always be >= 0

### DIFF
--- a/src/core/hle/kernel/resource_limit.cpp
+++ b/src/core/hle/kernel/resource_limit.cpp
@@ -24,13 +24,9 @@ bool ResourceLimit::Reserve(ResourceType resource, s64 amount, u64 timeout) {
     const std::size_t index{ResourceTypeToIndex(resource)};
 
     s64 new_value = current[index] + amount;
-    while (new_value > limit[index] && available[index] + amount <= limit[index]) {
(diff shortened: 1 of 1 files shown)


src/core/hle/kernel/resource_limit.cpp
```diff
@@ -24,13 +24,9 @@ bool ResourceLimit::Reserve(ResourceType resource, s64 amount, u64 timeout) {
     const std::size_t index{ResourceTypeToIndex(resource)};
 
     s64 new_value = current[index] + amount;
-    while (new_value > limit[index] && available[index] + amount <= limit[index]) {
+    if (new_value > limit[index] && available[index] + amount <= limit[index]) {
```
lioncash, line 27: I feel like this is swapping out one bug for another, considering this will now never run until new_value matches the specified exit condition that was being tested for before.

@bunnei, was this just a typo with using `>=` instead of `>` in the break condition?

ogniK5377: Yeah I messaged bunnei about this as well, apparently this was here for a while

ogniK5377: I feel as this should be an s64 instead

PR #4011, opened 2020-05-29
common_types: Add size_t as a common type

ogniK5377:
Writing size_t instead of std::size_t allows for better code formatting. Slowly the code base can be switched to size_t instead of std::size_t but shouldn't be done until prometheus is merged.


BreadFish64:
Hmm, `usize` like in Rust is more ergonomic to type

lioncash:
Posting from discord:

tbh I don't even know why the alias is necessary, given all implementations we care about provide `size_t` alongside `std::size_t` in the `<cstddef>` header (and will be the case for any worthwhile C++ implementation, given it'd break a ton of software code to not actually do that).

So we can just remove the `std::` qualifiers if it's really necessary without any other changes.

PR #4016, opened 2020-05-29
shader/other: Fix hardcoded value in S2R INVOCATION_INFO

ReinUsesLisp:
Geometry shaders built from Nvidia's compiler check for bits[16:23] to
be less than or equal to 0 with VSETP to default to a "safe" value of
0x8000'0000 (safe from hardware's perspective). To avoid hitting this
path in the shader, return 0x0001'0000 from S2R INVOCATION_INFO.

What S2R INVOCATION_INFO returns from a geometry shader has yet to be
tested.

- Fixes the first layer of a 3D texture on UE4 games. Rainbow will still be there because there are 31 3D texture slices that contain trash.


ReinUsesLisp:
Changed the value to `0x00ff'0000`. It seems to be the maximum value of vertices a primitive can emit, so passing a `0xff` here will be safe since we run on top of the host hardware that already checks this for us.

An optimization pass will remove this conditionals.

PR #4025, opened 2020-05-31
gl_device: Enable compute shaders for Intel proprietary drivers

Morph1984:
Previously we were disabling compute shaders on Intel's proprietary driver due to broken compute. This has been fixed in the latest Intel drivers. Re-enable compute for Intel proprietary drivers and remove the check for broken compute.


goldenx86:
No more T posing in OGL.

Morph1984:
To be clear, the T posing in 3.1.0 and above is **not fixed** by this

PR #4031, opened 2020-06-01
gl_shader_decompiler: Fix geometry shader outputs on Intel drivers

Morph1984:
On Intel's proprietary drivers, gl_Layer and gl_ViewportIndex are not allowed members of gl_PerVertex block, causing the shader to fail to compile. Fix this by declaring these variables outside of gl_PerVertex.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -619,6 +619,21 @@ private:
             }
         }
 
+        if (stage != ShaderType::Geometry &&
+            (stage != ShaderType::Vertex || device.HasVertexViewportLayer())) {
+            if (ir.UsesLayer()) {
+                code.AddLine("int gl_Layer;");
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -647,6 +634,21 @@ class GLSLDecompiler final {
...
+            if (ir.UsesViewportIndex()) {
+                code.AddLine("int gl_ViewportIndex;");
+            }
+        } else if ((ir.UsesLayer() || ir.UsesViewportIndex()) && stage == ShaderType::Vertex &&
+                   !device.HasVertexViewportLayer()) {
+            LOG_ERROR(
+                Render_OpenGL,
+                "GL_ARB_shader_viewport_layer_array is not available and its required by a shader");
+        }
+        code.AddNewLine();
```
VolcaEM, line 651: Why did you add this? Readability?

PR #4032, opened 2020-06-01
hid: Stub GetXpadIDs

VolcaEM:
Allows Minecraft: Nintendo Switch Edition (a.k.a. old Minecraft) to boot and go ingame


rcl0ng:
this update broke mouse, keyboard and controller support over steam big picture mode.

rcl0ng:
It is correct that it was a config error. I had 'Run as Admin' turned on which breaks controller support over the Steam Link app.

PR #4044, opened 2020-06-04
Downgrade "handle not signaled" error to trace

ogniK5377:
Clogs logs quite a bit


bunnei:
Is this expected behavior? If no -- it should at the very least be debug or warning.

PR #4068, opened 2020-06-10
externals: Replace conan opus with git submodule

ReinUsesLisp:
Apparently conan's prebuilt package ships without a runtime check for
AVX, crashing yuzu on devices without it.

Reverting this to externals seems to fix it. This wasn't tested on a CPU
without AVX, but it was tested with the following change on Windows:

bcdedit /set xsavedisable 1

master crashes inside opus code with an illegal instruction, this commit
doesn't.


degasus:
Could we please keep the FindOpus helper and try to use it? So with a shared installation of opus, we would still use it.

PR #4076, opened 2020-06-11
Direct Memory Access

vdwjeremy:
While troubleshooting dynarmic high cpu usage, I noticed it goes through a software page table for all memory access.
The reason for it, I'm guessing, is to take the most out the host OS allowed allocation space while still mapping the full 64 bit addressing space.

However this comes at a cost : any memory access requires
- 1 store/load for the page indirection
- 1 store/load for the target address
- 1 compare and jump for the fallback
In addition to it the page size in yuzu is 4K, which makes a large page table, that can accelerate CPU cache fill.

The switch addressing space being 512GB, I was wondering if it would be possible to map directly the guest virtual memory to the host virtual memory. That would reduce memory accesses to a single load/store with immediate offset, it would also move the page table logic to the hardware and solve the unaligned reads/writes.
The feature is not available yet in dynarmic A64, but it could be adapted from the A32 version.

The only diff in the PR for now is a test that Windows can allocate 512GB of contiguous virtual adress space.
I'm opening this PR to start the discussion and get your feedback, in case this was already considered and discarded, or if other part of the project could be incompatible.

### DIFF
--- a/src/common/virtual_buffer.cpp
+++ b/src/common/virtual_buffer.cpp
@@ -22,9 +22,20 @@
 
 namespace Common {
 
+const VAddr address_space_size{1ULL << 39};
+void* vmem{nullptr};
+void* next_base{nullptr};
+
 void* AllocateMemoryPages(std::size_t size) {
+
 #ifdef _WIN32
-    void* base{VirtualAlloc(nullptr, size, MEM_COMMIT, PAGE_READWRITE)};
+    if (vmem == nullptr) {
+        vmem = VirtualAlloc(nullptr, address_space_size, MEM_RESERVE, PAGE_READWRITE);
+        ASSERT(vmem);
+        next_base = vmem;
+    }
+    void* base{VirtualAlloc(next_base, size, MEM_COMMIT, PAGE_READWRITE)};
+    next_base = (char*)base + size;
 #else
     void* base{mmap(nullptr, size, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0)};
 
@@ -34,6 +45,8 @@ void* AllocateMemoryPages(std::size_t size) {
 #endif
 
     ASSERT(base);
+    ASSERT(base >= vmem);
+    ASSERT(base < (char*)vmem + address_space_size);
 
     return base;
 }
@@ -43,7 +56,7 @@ void FreeMemoryPages(void* base, std::size_t size) {
         return;
     }
 #ifdef _WIN32
-    ASSERT(VirtualFree(base, 0, MEM_RELEASE));
+    ASSERT(VirtualFree(base, size, MEM_DECOMMIT));
 #else
     ASSERT(munmap(base, size) == 0);
 #endif


MerryMage:
Hi! Thanks for attempting to solve this issue.

1. The major road-block to this approach for which you will have to find a solution for is as follows:
    The address space allocation granularity on Windows is 64 KiB.
2. You are incorrect when you say a fastmem approach will only require one load/store instruction. You will also have to verify whether the load/store is within the valid address space, which means this requires a minimum of 3 instructions.
    Otherwise you're providing a trivial read/write gadget which a malicious guest applicaiton can exploit.

I would be interested in hearing your ideas on the Windows address space granuarlity issue.

src/common/virtual_buffer.cpp
```diff
@@ -22,9 +22,20 @@
 
 namespace Common {
 
+const VAddr address_space_size{1ULL << 39};
+void* vmem{nullptr};
```
lioncash, line 26: Please don't introduce mutable file-scope variables into the common library. We almost have all global variables removed from the codebase and adding more of them hinders this goal.

MerryMage:
For documentation purposes I will try to clarify the constraints here.

1. One needs to allocate an appopriately sized block of shared memory with `CreateFileMappingA`. Shared memory is required for memory mirroring.
2. For each address space, one needs to allocate the 512GB region with `VirtualAlloc2`. I would advise using `MEM_RESERVE_PLACEHOLDER` for concurrency reasons.
    VirtualAlloc has a reservation granularity of 64KiB and a commit granularity of 4KiB.
3. For each region you wish to map, you must partition the previously reserved address space with `VirtualFree(MEM_RESERVE_PLACEHOLDER)`. Reading the MSDN documentation will clarify why this is necessary. You would also map in the previously allocated shared memory with `MapViewOfFile3`.
    MapViewOfFile has a granularity of 64KiB.
    I note that `VirtualProtect` has a granularity of 4KiB, which can be useful emulate a partial allocation. This gets tricky to handle, however.
4. You can unmap mapped memory with `UnmapViewOfFile2`.

You may be tempted to use `VirtualQuery`, but I would advise manually tracking state because calling into the kernel often isn't good for performance.

src/common/virtual_buffer.cpp
vdwjeremy: alright, I'll keep that in mind when going forward

vdwjeremy:
Thanks for your feedback

Good point for the virtual address higher range, a malicious/bad program could overflow it. We could use a bitfield to limit the address space.
Concerning the pages alignment I'm not sure to see where the problem will be
- The memory sections defined in [address_space_info.cpp#L31](https://github.com/yuzu-emu/yuzu/blob/88141bb2d44f26d1015be2cd832284bf322bc3b2/src/core/hle/kernel/memory/address_space_info.cpp#L31) are also aligned with 64K
- Mutexes are aligned on 4 bytes [mutex.cpp#L69](https://github.com/yuzu-emu/yuzu/blob/5ddcc1b2f582b10ac52ba9d329bfb4e49360c738/src/core/hle/kernel/mutex.cpp#L69), so the behaviour is the same
Does it have something to do with the way RasterizerCachedMemory is mapped ? [#L491](https://github.com/yuzu-emu/yuzu/blob/b8f5c71f2d7f819821acf036175cce65ab1ae12c/src/core/memory.cpp#L491)

amayra:
how about linux ?

MerryMage:
> Concerning the pages alignment I'm not sure to see where the problem will be

Switch pages are 4KiB in size and need to be placed on 4KiB boundaries. If you can only place a page at 64KiB boundaries, this poses issues.

vdwjeremy:
Implementation started in #MerryMage/dynarmic/pull/528

bunnei:
@vdwjeremy Do you plan to revisit this change?

v1993:
It probably should be noted that https://github.com/MerryMage/dynarmic/tree/fm incorporating above PR exists.

vdwjeremy:
hi @bunnei,
sorry I don't think I will be able to invest much time in this feature in the coming months,
if you are willing to give it a go, indeed https://github.com/MerryMage/dynarmic/tree/fm implements fastmem which can be used to create a flat addressing space.
I would advise benchmarking the concept before making architectural changes as perfomance gains seems to vary a lot depending on the usage.

degasus:
Hi, I'm sorry for missing this PR. We have picked your dynarmic patch and it is in their master now. Thanks a lot.
However this one here sounds like an early proof of concept (just like my first linux one). So I think it is fine to close those one here and point to #6422 for the full implementation.

Thanks a lot for your work on fastmem. Now half a year later, you have provided the first patches to get a huge speed up.

PR #4082, opened 2020-06-13
maxwell_to_gl: Implement MirrorOnceClampOGL wrap mode using GL_MIRROR_CLAMP_EXT

Morph1984:
Like MirrorOnceBorder, this requires the `GL_EXT_texture_mirror_clamp` extension. This extension is unfortunately not available on Intel's drivers (both Windows proprietary and Linux Mesa). Use `GL_MIRROR_CLAMP_TO_EDGE` as a fallback if the extension is unavailable.

A Vulkan equivalent is not available so it will be left unimplemented.

Used in Kirby Star Allies. Thanks to gidoly from the yuzu discord for the log.

### DIFF
--- a/src/video_core/renderer_opengl/maxwell_to_gl.h
+++ b/src/video_core/renderer_opengl/maxwell_to_gl.h
@@ -191,6 +191,12 @@ inline GLenum WrapMode(Tegra::Texture::WrapMode wrap_mode) {
         } else {
             return GL_MIRROR_CLAMP_TO_EDGE;
         }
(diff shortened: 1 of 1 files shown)


lioncash:
The general rule of thumb is to implement these when a known game using them is found so that the feature usage is documented.

ogniK5377:
Do not merge until this has been hardware tested, @ReinUsesLisp

bunnei:
We cannot have things indefinitely in EA -- @Morph1984 @ReinUsesLisp reminder to close out on this one

PR #4086, opened 2020-06-15
xbyak_abi: Cleanup

MerryMage:
Minor changes not affecting behaviour. (With the exception of the first commit, which is a bugfix.)


bunnei:
EA-tagging for a little while, just for sanity checking

PR #4093, opened 2020-06-16
externals: Revert to libressl, as build is broken with find_package(OpenSSL).

bunnei:


bunnei:
@degasus I am merging this so that mainline releases do not hit this issue tomorrow morning. This reverts your change https://github.com/yuzu-emu/yuzu/commit/b52097804334307103cfaba72cd6e27db804f650#diff-c03bfeb2bf107d5021458c91aeabba95. I'm not sure the way to properly fix this (so we do not use the system OpenSSL on the the Azure build machines), but please take a look when you get a chance. Thanks!

PR #4097, opened 2020-06-17
Fix framebuffer size on fractional scaling display

kevinxucs:
Fix OpenGL renderer framebuffer size on fractional scaling display by using devicePixelRatioF()

Before:
Screenshot_20200616_204427
After:
Screenshot_20200616_205301


bunnei:
Thanks!

PR #3955, opened 2020-05-17
Remake Kernel Scheduling, CPU Management & Boot Management (Prometheus)

FernandoS27:
This PR rewrites a series of key components from yuzu such as the Kernel Scheduler, CPU Management and the Boot Management.

The Kernel Scheduler has been based on TuxSH's firmware 7.0.0's kernel idb, SciresM's firmware 9.0.0's kernel idb and Mesopshere (both classic Mesasphere by TuxSH and modern Mesosphere by SciresM). The algorithms now correctly use Critical Sections/Scheduler Locks instead of the system of messy callbacks inherited from Citra. Guest Threads now use fibers with a custom library created in #3396. The new kernel is the same indifferent of SingleCore or Multicore Emulation, no hacks are applied to differ them, instead merry go round between threads on SingleCore occurs on CPU management. Thus guaranteeing the kernel is almost the same except for differences between Cycle Timing and Host Timing. In order to ease implementation of Multicore, a Host Timer is used instead of a Cycle Timer. The new Host Timer can have a reaction inaccuracy of up to 1 milisecond with average reaction delay of 50 microseconds.

Other effects of this PR:
- Dynarmic has been updated with corrections to Exclusive Memory Handling. In retrospective, yuzu has been updated to handle Atomic Writes using compiler CAS intrinsics.
- All timing events on the emulator now use nanoseconds instead of CPU Cycles.
- CPU Thread and Booting Thread are no longer the same thread, this may have effects not foreseen before. (this is the same for both multicore and singlecore).
- Pausing now differs from a step emulation method to a kernel assisted method. In order to pause the emulator, special suspend kernel threads with max priority are set to run, yielding execution to emulation control. Before this was done in the step emulation thread by not going into the next step.
- SingleCore uses Cycle Timing and everything except the GPU (unless async gpu is off) is run on a single thread. For deterministic emulation such as homebrew testing/emulator debugging, ity is suggested to run the emulator on SingleCore without Async GPU. This guarantees the most accuracy and the less randomization.

What needs to be checked:
- [ ] Audition of every affected SVC against the IDB and Scheduler Primitives.
- [x] Guarantee SingleCore has the same functionality as in mainline and that no regressions are present.

Further Work:
[...]

PR #4098, opened 2020-06-17
configuration: implement per-game configurations

lat9nq:
This PR implements an optional configuration for each game in a user's library. The majority of the UI changes takes place in the current game properties dialog. In it, the user gains access to the same General, System, Graphics, Advanced Graphics, and Audio tabs that are currently present in the Emulation -> Configure... dialog, barring minor tweaks. The Add-ons scrollbox is moved into an Add-ons tab and features the present interface for enabling game patches. Each existing checkbox is set to a tristate-checkbox, the partial state meaning that the setting will defer to the global setting. Combo boxes each include the entry "Use global configuration" at the beginning of it, meaning to defer to the global setting. Other settings such as the volume slider and the background color chooser each get a new combo box that explicitly sets their global state. The game configuration dialog saves a config file specific to that game for each game and not in the global configuration file. Upon booting a game, yuzu loads the game using the correct configuration depending on how or if the user configured it, applying each setting if the setting is marked as non-global.

This strategy involves creating a new template class that encompasses the type of many of the settings in Settings::Values. The class keeps a two records of each setting and a boolean that marks whether the object should use the global or game specific value. Upon access, it either writes or returns the appropriate value. 

In no particular order, work that still needs to be done includes:
- [x] Saving game-specific configurations to individual files
- [x] Opening either settings dialog and reading the correct settings
- [x] Booting a game using the correct values
  - Reads the game config on boot with a new Config object
- [x] Opening either configuration dialog during a game and configuring the correct values without affecting the game's execution
  - Solved by disabling the specific settings in the main configuration dialog if a game specific one overwrites it, and guarding saves from that dialog when disabled
- [x] Hierarchically configuring the settings, rather than a bunch of settings in one fell swoop
- ~~QoL additions~~ (moving to a separate PR)
  - ~~An indicator in the games list to show the particular game has a modified configuration~~
  - ~~A reset button to clear any local setting~~
  - ~~An option to boot the game without the per-game configuration on the right-click menu~~
- etc., this list will grow as needed.

Suggestions are welcome. 

[Imgur album](https://imgur.com/a/tzpQESS) so it's easier to make certain suggestions.

Example log: [yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/4855935/yuzu_log.txt)

Update: Per-game Input configurations will not be included in this PR.


src/core/settings.h
```diff
@@ -382,7 +382,7 @@ enum class GPUAccuracy : u32 {
     Extreme = 2,
 };
 
-struct Values {
+typedef struct Values_t {
```
lioncash, line 385: A typedef here isn't necessary. This is a C-ism that doesn't apply to C++

src/yuzu/configuration/configure_per_game_addons.h
```diff
@@ -19,35 +18,36 @@ class QTreeView;
...
-    ~ConfigurePerGameGeneral() override;
+    explicit ConfigurePerGameAddons(QWidget* parent = nullptr);
+    ~ConfigurePerGameAddons() override;
 
     /// Save all button configurations to settings file
     void ApplyConfiguration();
 
     void LoadFromFile(FileSys::VirtualFile file);
 
+    void SetTitleId(u64 t_id);
```
lioncash, line 36:
```suggestion
    void SetTitleId(u64 id);
```

Ditto in the cpp file.

src/core/settings.h
ogniK5377, line 385:
```suggestion
struct Values {
```

src/core/settings.h
```diff
@@ -488,7 +488,11 @@ struct Values {
 
     // Add-Ons
     std::map<u64, std::vector<std::string>> disabled_addons;
-} extern values;
+} Values;
```
ogniK5377, line 491:
```suggestion
};
```

bunnei:
Hi @lat9nq, I haven't looked at the code too closely yet, but great to see someone working on this feature! However, I've thought about this one myself a bit, and rather than having global settings or game specific settings (with a pointer at one or the other), I was wondering if it made sense to make this more hierarchical? I.e. global settings would be applied, and then, if there was a game specific setting(s), that would override only that specific specific setting(s). E.g., I could imagine a user might want to only configure their controller once for all games, but then override specific settings like docked mode, or the renderer, on a per game basis.

lat9nq:
Thanks for the input, @bunnei! Since I first described my approach, it's has changed slightly so now the Input config is not game-specific whatsoever, and I split the Settings::Values struct into two mutually exclusive structs that together contain the same data. I duplicate the one that has the settings I want to change, leaving the other one to be the global settings and isn't overwritten.

My way of doing it is a bit simpler, but if I do it hierarchically, it makes the config much more future-proof, thus a novel setting doesn't need to be manually set for each game. It would also make it easier to do input configs *I think*, but I'd still want to keep that to its own config file in a future PR.

I'll upload what I've done today (after fixing a regression) and then stop to think about how to make it hierarchical. Personally I like that idea a lot more. I'm mostly unsure about how to do that in the UI, though. Currently it uses the exact same QT widgets that the normal `Emulation -> Configure...` dialog does (e.g. ConfigureGeneral, ConfigureAdvancedGraphics, etc.). I didn't want to create a bunch of new widgets and then force the devs to add a setting twice for each new feature.

src/core/settings.h
```diff
@@ -382,20 +382,93 @@ enum class GPUAccuracy : u32 {
...
+    // Misceallaneous
+    Setting<std::string> log_filter;
+    Setting<bool> use_dev_keys;
+
+    // Renderer
+    Setting<RendererBackend> renderer_backend;
+    Setting<bool> renderer_debug;
+    Setting<int> vulkan_device;
+
+    Setting<u16> resolution_factor;
```
Morph1984, line 443: default initialize this to 1

src/yuzu/configuration/config.cpp
```diff
@@ -630,12 +644,15 @@ void Config::ReadRendererValues() {
...
+    Settings::values.vulkan_device =
+        ReadSetting(QStringLiteral("vulkan_device"), 0).toInt();
+    Settings::values.aspect_ratio =
+        ReadSetting(QStringLiteral("aspect_ratio"), 0).toInt();
+    Settings::values.max_anisotropy =
+        ReadSetting(QStringLiteral("max_anisotropy"), 0).toInt();
     Settings::values.use_frame_limit =
         ReadSetting(QStringLiteral("use_frame_limit"), true).toBool();
-    Settings::values.frame_limit = ReadSetting(QStringLiteral("frame_limit"), 100).toUInt();
+    Settings::values.frame_limit = ReadSetting(QStringLiteral("frame_limit"), 100).toInt();
```
Morph1984, line 655: this was changed to UInt since frame limit is a u16

src/yuzu/configuration/config.cpp
```diff
@@ -853,7 +873,8 @@ void Config::SavePlayerValues() {
 }
 
 void Config::SaveDebugValues() {
-    WriteSetting(QStringLiteral("debug_pad_enabled"), Settings::values.debug_pad_enabled, false);
+    WriteSetting(QStringLiteral("debug_pad_enabled"), Settings::values.debug_pad_enabled,
+                 false);
```
Morph1984, line 877: did clangformat do this?

src/yuzu/configuration/config.cpp
```diff
@@ -938,21 +963,25 @@ void Config::SaveControlValues() {
     WriteSetting(QStringLiteral("motion_device"),
                  QString::fromStdString(Settings::values.motion_device),
                  QStringLiteral("engine:motion_emu,update_period:100,sensitivity:0.01"));
-    WriteSetting(QStringLiteral("keyboard_enabled"), Settings::values.keyboard_enabled, false);
+    WriteSetting(QStringLiteral("keyboard_enabled"), Settings::values.keyboard_enabled,
+                 false);
```
Morph1984, line 967: ditto

src/yuzu/configuration/config.cpp
```diff
@@ -938,21 +963,25 @@ void Config::SaveControlValues() {
...
     WriteSetting(QStringLiteral("udp_input_address"),
                  QString::fromStdString(Settings::values.udp_input_address),
                  QString::fromUtf8(InputCommon::CemuhookUDP::DEFAULT_ADDR));
     WriteSetting(QStringLiteral("udp_input_port"), Settings::values.udp_input_port,
                  InputCommon::CemuhookUDP::DEFAULT_PORT);
     WriteSetting(QStringLiteral("udp_pad_index"), Settings::values.udp_pad_index, 0);
+    WriteSetting(QStringLiteral("use_docked_mode"), Settings::values.use_docked_mode,
+                 false);
```
Morph1984, line 975: ditto

lat9nq:
Left [this Imgur album](https://imgur.com/a/tzpQESS) in the main post if people want to look at it and make suggestions.

lat9nq:
Do not tag this for early-access, since it has merge conflicts with #3955

src/yuzu/configuration/configure_graphics.cpp
```diff
@@ -26,20 +27,17 @@ ConfigureGraphics::ConfigureGraphics(QWidget* parent)
...
     connect(ui->device, static_cast<void (QComboBox::*)(int)>(&QComboBox::activated), this,
             [this](int device) { UpdateDeviceSelection(device); });
 
-    connect(ui->bg_button, &QPushButton::clicked, this, [this] {
-        const QColor new_bg_color = QColorDialog::getColor(bg_color);
-        if (!new_bg_color.isValid()) {
-            return;
-        }
-        UpdateBackgroundColorButton(new_bg_color);
-    });
```
lat9nq, line 42: Should not have been deleted

src/yuzu/configuration/configure_graphics.cpp
```diff
@@ -132,5 +196,44 @@ void ConfigureGraphics::RetrieveVulkanDevices() {
...
+    connect(ui->bg_combobox, static_cast<void (QComboBox::*)(int)>(&QComboBox::activated), this,
+            [this](int index) { ui->bg_button->setEnabled(index == 1); });
+
+    connect(ui->bg_button, &QPushButton::clicked, this, [this] {
+        const QColor new_bg_color = QColorDialog::getColor(bg_color);
+        if (!new_bg_color.isValid()) {
+            return;
+        }
+        UpdateBackgroundColorButton(new_bg_color);
+    });
```
lat9nq, line 233: Should be in the constructor

src/yuzu/main.cpp
```diff
@@ -543,6 +543,7 @@ void GMainWindow::InitializeWidgets() {
     async_status_button->setText(tr("ASYNC"));
     async_status_button->setCheckable(true);
     async_status_button->setChecked(Settings::values.use_asynchronous_gpu_emulation);
+    statusBar()->insertPermanentWidget(0, async_status_button);
```
lat9nq, line 546: Thought this was moved as a result of the rebase, need to undo

lat9nq:
@bunnei one of the changes in this PR affects when yuzu logs the game settings. When I tested it, the discord bot does not successfully parse the logs from this PR. I feel like the responsible thing would be to make the necessary changes to the bot and PR it to support these logs. Where is the bot's log parsing source code?

bunnei:
@lat9nq I think we are ready to test this more broadly in EA -- however, this change should not update dynarmic. Can you rebase and drop the changes to dynarmic?

Also, no worries about the bot -- we can update that as appropriate.

Thanks!

lat9nq:
Thanks, @bunnei! I guess when this check completes, you can give it the EA tag!

A note about the logs, this PR tells yuzu to wait until the game boots before printing the settings, so `yuzu is starting...` should appear immediately before the relevant settings appear.

Example [yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/4854872/yuzu_log.txt)

src/core/settings.h
```diff
@@ -382,20 +382,97 @@ enum class GPUAccuracy : u32 {
...
+
+    // Core
+    Setting<bool> use_multi_core;
+
+    // Renderer
+    Setting<RendererBackend> renderer_backend;
+    Setting<bool> renderer_debug;
+    Setting<int> vulkan_device;
+
+    Setting<u16> resolution_factor = Setting(static_cast<u16>(1));
```
Morph1984, line 447: No need to cast, use 1U instead
```suggestion
    Setting<u16> resolution_factor = Setting(1U);
```

lat9nq: Build failed since it is expecting a short unsigned int, not an unsigned int. Is there a way to write a short unsigned int literal?

FearlessTobi:
We had some reports of settings resetting with this merged. Is this PR providing a clear migration path for settings?

lat9nq:
The only setting that should be reset is docked mode as it was moved to control settings, all the others should be untouched. Do you mean that all settings were being reset when the PR was tested?

src/yuzu/main.cpp
```diff
@@ -1035,6 +1035,23 @@ void GMainWindow::BootGame(const QString& filename) {
     LOG_INFO(Frontend, "yuzu starting...");
     StoreRecentFile(filename); // Put the filename on top of the list
 
+    u64 title_id;
```
Morph1984, line 1038:
```suggestion
    u64 title_id{0}
```

src/yuzu/main.cpp
```diff
@@ -1035,6 +1035,23 @@ void GMainWindow::BootGame(const QString& filename) {
...
+    const auto v_file = Core::GetGameFileFromPath(vfs, filename.toUtf8().constData());
+    const auto loader = Loader::GetLoader(v_file);
+    if (loader == nullptr || loader->ReadProgramId(title_id) != Loader::ResultStatus::Success) {
+        QMessageBox::information(this, tr("Properties"),
+                                 tr("The game properties could not be loaded."));
+        return;
+    }
+
+    // Load per game settings
+    Config per_game_config(fmt::format("{:016X}", title_id) + ".ini", false);
```
Morph1984, line 1049: Homebrew has titleid 0, skip loading per game settings if this is the case.

Morph1984: Currently, launching any homebrew nro will cause this to pop up 
image

Morph1984: Alternatively, you could attempt to apply a custom configuration if and only if the loader is not nullptr or the loader is able to get a title id

src/yuzu/main.cpp
```diff
@@ -1035,19 +1035,15 @@ void GMainWindow::BootGame(const QString& filename) {
     LOG_INFO(Frontend, "yuzu starting...");
     StoreRecentFile(filename); // Put the filename on top of the list
 
-    u64 title_id;
+    u64 title_id{0Ul};
```
Morph1984, line 1038: An `unsigned long` is not universal, yuzu's code uses either `u64 title_id{}` or `u64 title_id{0}`

src/core/settings.h
Morph1984: Using u16(1) would be your best choice then in this case. There's no clean way to represent an unsigned short using an int literal

lat9nq: Reverting because that change causes ctest to fail.

src/yuzu/configuration/config.cpp
```diff
@@ -1234,6 +1265,34 @@ QVariant Config::ReadSetting(const QString& name, const QVariant& default_value)
     return result;
 }
 
+template <typename Type>
+void Config::ReadSettingGlobal(Settings::Setting<Type>& setting, const QString& name) {
+    bool use_global = qt_config->value(name + QStringLiteral("/use_global"), true).toBool();
```
VolcaEM, line 1270:
```suggestion
    const bool use_global = qt_config->value(name + QStringLiteral("/use_global"), true).toBool();
```

src/yuzu/configuration/config.cpp
```diff
@@ -1234,6 +1265,34 @@ QVariant Config::ReadSetting(const QString& name, const QVariant& default_value)
...
+    setting.SetGlobal(use_global);
+    if (global || !use_global) {
+        setting.SetValue(ReadSetting(name).value<Type>());
+    }
+}
+
+template <typename Type>
+void Config::ReadSettingGlobal(Settings::Setting<Type>& setting, const QString& name,
+                               const QVariant& default_value) {
+    bool use_global = qt_config->value(name + QStringLiteral("/use_global"), true).toBool();
```
VolcaEM, line 1280:
```suggestion
    const bool use_global = qt_config->value(name + QStringLiteral("/use_global"), true).toBool();
```

src/yuzu/configuration/config.cpp
```diff
@@ -1234,6 +1265,34 @@ QVariant Config::ReadSetting(const QString& name, const QVariant& default_value)
...
+    setting.SetGlobal(use_global);
+    if (global || !use_global) {
+        setting.SetValue(ReadSetting(name, default_value).value<Type>());
+    }
+}
+
+template <typename Type>
+void Config::ReadSettingGlobal(Type& setting, const QString& name,
+                               const QVariant& default_value) const {
+    bool use_global = qt_config->value(name + QStringLiteral("/use_global"), true).toBool();
```
VolcaEM, line 1290:
```suggestion
    const bool use_global = qt_config->value(name + QStringLiteral("/use_global"), true).toBool();
```

src/core/settings.h
```diff
@@ -382,20 +382,97 @@ enum class GPUAccuracy : u32 {
     Extreme = 2,
 };
 
+extern bool configuring_global;
+
+template <typename Type>
+class Setting {
```
VolcaEM, line 388: (Not sure)
```suggestion
class Setting final {
```

src/yuzu/configuration/configuration_shared.h
```diff
@@ -0,0 +1,37 @@
...
+#include <QComboBox>
+#include <QString>
+#include "core/settings.h"
+
+namespace ConfigurationShared {
+
+constexpr int USE_GLOBAL_INDEX = 0;
+constexpr int USE_GLOBAL_SEPARATOR_INDEX = 1;
+constexpr int USE_GLOBAL_OFFSET = 2;
+const QString use_global_text = QString::fromUtf8("Use global configuration");
```
VolcaEM, line 17: (Not sure)
```suggestion
static const QString use_global_text = QString::fromUtf8("Use global configuration");
```

src/yuzu/configuration/configure_graphics_advanced.cpp
```diff
@@ -5,40 +5,96 @@
...
 void ConfigureGraphicsAdvanced::ApplyConfiguration() {
-    auto gpu_accuracy = static_cast<Settings::GPUAccuracy>(ui->gpu_accuracy->currentIndex());
-    Settings::values.gpu_accuracy = gpu_accuracy;
-    Settings::values.use_vsync = ui->use_vsync->isChecked();
-    Settings::values.use_assembly_shaders = ui->use_assembly_shaders->isChecked();
-    Settings::values.use_fast_gpu_time = ui->use_fast_gpu_time->isChecked();
-    Settings::values.force_30fps_mode = ui->force_30fps_mode->isChecked();
-    Settings::values.max_anisotropy = ui->anisotropic_filtering_combobox->currentIndex();
+    // Subtract 2 if configuring per-game (separator and "use global configuration" take 2 slots)
+    auto gpu_accuracy = static_cast<Settings::GPUAccuracy>(
```
VolcaEM, line 54:
```suggestion
    const auto gpu_accuracy = static_cast<Settings::GPUAccuracy>(
```

src/yuzu/configuration/configure_per_game.cpp
```diff
@@ -0,0 +1,139 @@
+// Copyright 2016 Citra Emulator Project
```
VolcaEM, line 1: Are you sure this is right?

src/core/settings.h
```diff
@@ -382,20 +382,97 @@ enum class GPUAccuracy : u32 {
...
+        use_global = true;
+    }
+    Setting(Type val) {
+        use_global = true;
+        global = val;
+    }
+    ~Setting() = default;
+    void SetGlobal(bool to_global) {
+        use_global = to_global;
+    };
```
lioncash, line 400: Semicolons aren't necessary on the closing brace of a function, ditto for all occurrences below.

src/core/settings.h
```diff
@@ -382,20 +382,97 @@ enum class GPUAccuracy : u32 {
...
+    operator Type() const {
+        return GetValue();
+    };
+    Type operator=(const Type& b) {
+        SetValue(b);
+        return b;
+    };
+
+private:
+    bool use_global{};
```
lioncash, line 426:
```suggestion
    bool use_global = true;
```

Then, the assignments in the constructor can be removed.

src/core/settings.h
```diff
@@ -382,20 +382,97 @@ enum class GPUAccuracy : u32 {
     Extreme = 2,
 };
 
+extern bool configuring_global;
+
+template <typename Type>
+class Setting final {
+public:
+    Setting() {
```
lioncash, line 390:
```suggestion
    Setting() = default;
```

After previous suggestions are applied.

src/core/settings.h
```diff
@@ -382,20 +382,97 @@ enum class GPUAccuracy : u32 {
...
+        use_global = true;
+    }
+    Setting(Type val) {
```
lioncash, line 393:
```suggestion
    explicit Setting(Type val) : global{val} {}
```

src/core/settings.h
```diff
@@ -382,20 +382,97 @@ enum class GPUAccuracy : u32 {
...
+        if (use_global) {
+            global = value;
+        } else {
+            local = value;
+        }
+    };
+    operator Type() const {
+        return GetValue();
+    };
+    Type operator=(const Type& b) {
```
lioncash, line 420: This seems like an awfully bizarre assignment operator, given it's not returning a reference to `this`.

Maybe make this its own named function (or just use GetValue/SetValue() to begin with?

src/yuzu/configuration/configuration_shared.h
```diff
@@ -0,0 +1,37 @@
...
+#include <QComboBox>
+#include <QString>
+#include "core/settings.h"
+
+namespace ConfigurationShared {
+
+constexpr int USE_GLOBAL_INDEX = 0;
+constexpr int USE_GLOBAL_SEPARATOR_INDEX = 1;
+constexpr int USE_GLOBAL_OFFSET = 2;
+static const QString use_global_text = QString::fromUtf8("Use global configuration");
```
lioncash, line 17: This can be made non-static within `InsertGlobalItem` as `tr("Use global configuration")`, so that the text can actually be translated at runtime in the future.

src/yuzu/configuration/config.h
```diff
@@ -15,7 +16,7 @@ class QSettings;
 
 class Config {
 public:
-    Config();
+    Config(const std::string& config_loc = "qt-config.ini", bool is_global = true);
```
lioncash, line 19:
```suggestion
    explicit Config(const std::string& config_loc = "qt-config.ini", bool is_global = true);
```

src/yuzu/configuration/configure_per_game.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "ui_configure_per_game.h"
+#include "yuzu/configuration/config.h"
+#include "yuzu/configuration/configure_input.h"
+#include "yuzu/configuration/configure_per_game.h"
+#include "yuzu/uisettings.h"
+#include "yuzu/util/util.h"
+
+ConfigurePerGame::ConfigurePerGame(QWidget* parent, u64 title_id)
+    : QDialog(parent), ui(std::make_unique<Ui::ConfigurePerGame>()), title_id(title_id) {
+    game_config = std::make_unique<Config>(fmt::format("{:016X}", title_id) + ".ini", false);
```
lioncash, line 32:
```suggestion
    game_config = std::make_unique<Config>(fmt::format("{:016X}.ini", title_id), false);
```

src/yuzu/configuration/configure_per_game.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+void ConfigurePerGame::LoadConfiguration() {
+    if (file == nullptr) {
+        return;
+    }
+
+    ui->addonsTab->LoadFromFile(file);
+
+    ui->display_title_id->setText(QString::fromStdString(fmt::format("{:016X}", title_id)));
```
lioncash, line 88:
```suggestion
    ui->display_title_id->setText(QStringLiteral("%1").arg(title_id, 16, 16, 0).toUpper()));
```

src/yuzu/main.cpp
```diff
@@ -1035,6 +1035,17 @@ void GMainWindow::BootGame(const QString& filename) {
     LOG_INFO(Frontend, "yuzu starting...");
     StoreRecentFile(filename); // Put the filename on top of the list
 
+    u64 title_id{0};
+
+    const auto v_file = Core::GetGameFileFromPath(vfs, filename.toUtf8().constData());
+    const auto loader = Loader::GetLoader(v_file);
+    if (!(loader == nullptr || loader->ReadProgramId(title_id) != Loader::ResultStatus::Success)) {
+        // Load per game settings
+        Config per_game_config(fmt::format("{:016X}", title_id) + ".ini", false);
```
lioncash, line 1044:
```suggestion
        Config per_game_config(fmt::format("{:016X}.ini", title_id), false);
```

src/yuzu/configuration/configure_per_game.cpp
lat9nq: Doing that produces a warning:
```
yuzu/src/yuzu/configuration/configure_per_game.cpp:88:79: warning: ISO C++ says that these are ambiguous, even though the worst conversion for the first is better than the worst conversion for the second:
   88 |     ui->display_title_id->setText(QStringLiteral("%1").arg(title_id, 16, 16, 0).toUpper());
      |                                                                               ^
```

lat9nq: Nevermind, missed a parenthesis.

lat9nq: That change breaks title id formatting (e.g. got `72174692027785216` instead of `01006A800016E000`).

lioncash: The final zero was supposed to be `QLatin1Char{'0'}`, my bad.

bunnei:
LGTM -- Really awesome work here!

PR #4099, opened 2020-06-17
Fix compilation on macOS

MerryMage:
Please note: This does not enable yuzu to be playable on macOS.

### DIFF
--- a/src/video_core/renderer_vulkan/renderer_vulkan.cpp
+++ b/src/video_core/renderer_vulkan/renderer_vulkan.cpp
@@ -13,6 +13,7 @@
 #include <fmt/format.h>
 
 #include "common/dynamic_library.h"
+#include "common/file_util.h"
 #include "common/logging/log.h"
 #include "common/telemetry.h"
 #include "core/core.h"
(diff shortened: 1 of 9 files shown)


src/yuzu/bootmanager.cpp
```diff
@@ -571,6 +579,11 @@ bool GRenderWindow::InitializeOpenGL() {
         std::make_unique<OpenGLSharedContext>(context->GetShareContext(), child->windowHandle()));
 
     return true;
+#else
+    QMessageBox::critical(this, tr("OpenGL not available!"),
```
lioncash, line 583: This should be a warning, not a critical error given this by itself isn't a stability issue

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -870,7 +870,7 @@ void RasterizerVulkan::BeginTransformFeedback() {
     UNIMPLEMENTED_IF(binding.buffer_offset != 0);
 
     const GPUVAddr gpu_addr = binding.Address();
-    const std::size_t size = binding.buffer_size;
+    const VkDeviceSize size = static_cast<VkDeviceSize>(binding.buffer_size);
```
lioncash, line 873:
```suggestion
    const auto size = static_cast<VkDeviceSize>(binding.buffer_size);
```

src/video_core/renderer_vulkan/renderer_vulkan.cpp
```diff
@@ -76,7 +77,8 @@ Common::DynamicLibrary OpenVulkanLibrary() {
     char* libvulkan_env = getenv("LIBVULKAN_PATH");
     if (!libvulkan_env || !library.Open(libvulkan_env)) {
         // Use the libvulkan.dylib from the application bundle.
-        std::string filename = File::GetBundleDirectory() + "/Contents/Frameworks/libvulkan.dylib";
+        std::string filename =
```
VolcaEM, line 80: I think this can be const

MerryMage:
Addressed above comments.

PR #4105, opened 2020-06-18
gl_rasterizer: Use NV_vertex_buffer_unified_memory for vertex buffer robustness

ReinUsesLisp:
Switch games are allowed to bind less data than what they use in a vertex buffer, the expected behavior here is that these values are read as zero. At the moment of writing this only D3D12, OpenGL through [NV_vertex_buffer_unified_memory](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_vertex_buffer_unified_memory.txt) and NVN support vertex buffer with a size limit.

In theory this could be emulated on Vulkan creating a new [VkBuffer](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBufferCreateInfo.html) for each [handle, offset, length] tuple and binding the expected data to it. This is likely going to be slow and memory expensive when used on the vertex buffer and we have to do it on all draws because we can't know
without analysing indices when a game is going to read vertex data out of bounds.

This is not a problem on OpenGL's `BufferAddressRangeNV` because it takes a length parameter, unlike Vulkan's [CmdBindVertexBuffers](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdBindVertexBuffers.html) that only takes buffers and offsets (the length is implicit in VkBuffer). It isn't a problem on D3D12 either, because [D3D12_VERTEX_BUFFER_VIEW](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_vertex_buffer_view) on [IASetVertexBuffers](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-iasetvertexbuffers) takes `SizeInBytes` as a parameter (although I am not familiar with robustness on D3D12).

Currently this only implements buffer ranges for vertex buffers, although indices can also be affected. A KHR_robustness profile is not created, but Nvidia's driver reads out of bound vertex data as zero anyway, this might have to be changed in the future.

On Turing GPUs Nvidia's driver crashes when the buffer is marked as resident or on DeleteBuffers, hence this functionality is disabled on these devices for now.

Resident buffers are required for bindless storage buffers that allows us to properly implement per-stage global memory on GLASM, and as a small plus `BufferAddressRangeNV` is around 5 times faster than `BindVertexBuffer` with a non-dummy handle.

- This fixes SMO random triangles when capturing an enemy, getting hit, or looking at the environment on certain maps. Currently only Nvidia devices that are not Turing based have this fixed.

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -188,16 +188,32 @@ bool IsASTCSupported() {
     return true;
 }
 
+/// @brief Returns true when a GL_RENDERER is a Turing GPU
+/// @param renderer GL_RENDERER string
+bool IsTuring(std::string_view renderer) {
+    static constexpr std::array<std::string_view, 12> TURING_GPUS = {
+        "GTX 1650",        "GTX 1660",        "RTX 2060",        "RTX 2070",
+        "RTX 2080",        "TITAN RTX",       "Quadro RTX 3000", "Quadro RTX 4000",
+        "Quadro RTX 5000", "Quadro RTX 6000", "Quadro RTX 8000", "Tesla T4",
+    };
+    return std::any_of(TURING_GPUS.begin(), TURING_GPUS.end(),
+                       [renderer](std::string_view candidate) {
+                           return renderer.find(candidate) != std::string_view::npos;
+                       });
+}
+
 } // Anonymous namespace
 
 Device::Device()
     : max_uniform_buffers{BuildMaxUniformBuffers()}, base_bindings{BuildBaseBindings()} {
     const std::string_view vendor = reinterpret_cast<const char*>(glGetString(GL_VENDOR));
+    const std::string_view renderer = reinterpret_cast<const char*>(glGetString(GL_RENDERER));
     const std::string_view version = reinterpret_cast<const char*>(glGetString(GL_VERSION));
     const std::vector extensions = GetExtensions();
 
     const bool is_nvidia = vendor == "NVIDIA Corporation";
     const bool is_amd = vendor == "ATI Technologies Inc.";
+    const bool is_turing = is_nvidia && IsTuring(renderer);
 
     bool disable_fast_buffer_sub_data = false;
     if (is_nvidia && version == "4.6.0 NVIDIA 443.24") {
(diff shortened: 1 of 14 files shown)


src/video_core/renderer_opengl/gl_device.cpp
```diff
@@ -180,16 +180,33 @@ bool IsASTCSupported() {
...
 }
 
+/// @brief Returns true when a GL_RENDERER is a Turing GPU
+/// @param renderer GL_RENDERER string
+bool IsTuring(std::string_view renderer) {
+    static constexpr std::array<std::string_view, 7> TURING_GPUS = {
+        "GTX 1650", "GTX 1660",        "RTX 2060",        "RTX 2070",
+        "RTX 2080", "Quadro RTX 5000", "Quadro RTX 6000",
+    };
+    for (const std::string_view candidate : TURING_GPUS) {
```
lioncash, line 190: std::any_of

src/video_core/renderer_opengl/gl_device.cpp
```diff
@@ -180,16 +180,31 @@ bool IsASTCSupported() {
...
 Device::Device()
     : max_uniform_buffers{BuildMaxUniformBuffers()}, base_bindings{BuildBaseBindings()} {
     const std::string_view vendor = reinterpret_cast<const char*>(glGetString(GL_VENDOR));
+    const std::string_view renderer = reinterpret_cast<const char*>(glGetString(GL_RENDERER));
     const std::string_view version = reinterpret_cast<const char*>(glGetString(GL_VERSION));
     const std::vector extensions = GetExtensions();
 
     const bool is_nvidia = vendor == "NVIDIA Corporation";
     const bool is_amd = vendor == "ATI Technologies Inc.";
+    const bool is_turing = is_nvidia && IsTuring(renderer);
```
VolcaEM, line 207: Why not just IsTuring?

src/video_core/renderer_opengl/gl_device.cpp
```diff
@@ -188,16 +188,31 @@ bool IsASTCSupported() {
     return true;
 }
 
+/// @brief Returns true when a GL_RENDERER is a Turing GPU
+/// @param renderer GL_RENDERER string
+bool IsTuring(std::string_view renderer) {
+    static constexpr std::array<std::string_view, 7> TURING_GPUS = {
+        "GTX 1650", "GTX 1660",        "RTX 2060",        "RTX 2070",
+        "RTX 2080", "Quadro RTX 5000", "Quadro RTX 6000",
+    };
```
Morph1984, line 197: Add more Turing GPUs
```cpp
    static constexpr std::array<std::string_view, 12> TURING_GPUS = {
        "GTX 1650",        "GTX 1660",        "RTX 2060",        "RTX 2070",
        "RTX 2080",        "TITAN RTX",       "Quadro RTX 3000", "Quadro RTX 4000",
        "Quadro RTX 5000", "Quadro RTX 6000", "Quadro RTX 8000", "Tesla T4",
    };
```

oscarbg: Sorry for joining discussion but what about Volta GPU Titan V, I say because it's very similar to Turing at least in compute side.. have you tested it on a Volta GPU?

Morph1984: I don't think anyone has a Titan V to test with, unfortunately.

oscarbg: @Morph1984 I asked because I have it! how can I help in testing?

VolcaEM: Try joining the Discord server (you can find the invite in the README) and asking the developers there, it would be much faster than asking here

PR #4112, opened 2020-06-19
macro: Add support for "middle methods" on the code cache

ogniK5377:
Macro code is just uploaded sequentially from a starting address, however that does not mean the entry point for the macro is at that address. This PR adds preliminary support for executing macros in the middle of our cached code.

### DIFF
--- a/src/video_core/macro/macro.cpp
+++ b/src/video_core/macro/macro.cpp
@@ -35,10 +35,19 @@ void MacroEngine::Execute(Engines::Maxwell3D& maxwell3d, u32 method,
         }
     } else {
         // Macro not compiled, check if it's uploaded and if so, compile it
+        std::pair<bool, u32> mid_method;
         auto macro_code = uploaded_macro_code.find(method);
         if (macro_code == uploaded_macro_code.end()) {
-            UNREACHABLE_MSG("Macro 0x{0:x} was not uploaded", method);
-            return;
+            for (const auto& [method_base, code] : uploaded_macro_code) {
+                if (method >= method_base && (method - method_base) < code.size()) {
+                    mid_method = {true, method_base};
+                    break;
+                }
+            }
+            if (!mid_method.first) {
+                UNREACHABLE_MSG("Macro 0x{0:x} was not uploaded", method);
+                return;
+            }
         }
         auto& cache_info = macro_cache[method];
         cache_info.hash = boost::hash_value(macro_code->second);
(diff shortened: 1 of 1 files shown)


MerryMage:
Curious: Is a macro guaranteed never to have any back-references such that cutting off code never causes issues? :o

src/video_core/macro/macro.cpp
```diff
@@ -21,12 +21,33 @@ void MacroEngine::Execute(u32 method, const std::vector<u32>& parameters) {
         compiled_macro->second->Execute(parameters, method);
     } else {
         // Macro not compiled, check if it's uploaded and if so, compile it
+        std::pair<bool, u32> mid_method;
         auto macro_code = uploaded_macro_code.find(method);
```
MerryMage, line 25: If `std::unordered_map<u32, std::vector<u32>>` → `std::map<u32, std::vector<u32>>`, you could instead implement this as:

```
const auto macro_code = uploaded_macro_code.lower_bound(method);
if (macro_code == uploaded_macro_code.end() || method >= macro_code->first + macro_code->second.size()) {
    UNREACHABLE_MSG("Macro 0x{0:x} was not uploaded", method);
    return;
}
if (macro_code->first == method) {
    macro_cache[method] = Compile(macro_code->second);
} else {
    const auto& macro_cached = macro_code->second;
    const auto rebased_method = method - macro_code->first;
    auto& code = uploaded_macro_code[method];
    code.resize(macro_cached.size() - rebased_method);
    std::memcpy(code.data(), macro_cached.data() + rebased_method,
                code.size() * sizeof(u32));
    macro_cache[method] = Compile(code);
}
```

I still feel like the truncation is unnecessarily complicated, but *shrug*.

src/video_core/macro/macro.cpp
```diff
@@ -21,12 +21,33 @@ void MacroEngine::Execute(u32 method, const std::vector<u32>& parameters) {
         compiled_macro->second->Execute(parameters, method);
     } else {
         // Macro not compiled, check if it's uploaded and if so, compile it
+        std::pair<bool, u32> mid_method;
```
MerryMage, line 24: This is a `std::optional`.

ReinUsesLisp:
Untagged due to conflicts.

src/video_core/macro/macro.cpp
ogniK5377: Sadly this wouldn't work, homebrew tends to upload all their macros at "0" and we'll only have 1 entry. This would lead to upper_bound/lower_bound always point end for any non zero value

PR #4113, opened 2020-06-19
Fix compilation when not building with boxcat

ogniK5377:
Fixes compilation when trying to build without boxcat enabled.

Zip is also used by vfs and should not be exclusive to building with boxcat.

### DIFF
--- a/src/yuzu/configuration/configure_service.cpp
+++ b/src/yuzu/configuration/configure_service.cpp
@@ -68,6 +68,7 @@ void ConfigureService::SetConfiguration() {
 }
 
 std::pair<QString, QString> ConfigureService::BCATDownloadEvents() {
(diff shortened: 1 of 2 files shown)


src/yuzu/configuration/configure_service.cpp
```diff
@@ -106,6 +107,10 @@ std::pair<QString, QString> ConfigureService::BCATDownloadEvents() {
                    .arg(FormatEventStatusString(value));
     }
     return {QStringLiteral("Current Boxcat Events"), std::move(out)};
+#else
+    return {QStringLiteral("Current Boxcat Events"),
```
lioncash, line 111: Pretty sure both this and the string above should be translatable, given they're both shown to the user.

PR #4120, opened 2020-06-20
gl_arb_decompiler: Avoid several string copies

lioncash:
Variables that are marked as const cannot have the move constructor invoked when returning from a function (the move constructor requires a non-const variable so it can "steal" the resources from it.

@ReinUsesLisp Is the transformation made at 1159 (moving the string into the conditional) safe, or does it need to purposely be outside of the if statement?.


ReinUsesLisp:
It's safe. That function is basically doing a `std::visit`.

PR #4127, opened 2020-06-20
texture_cache: Fix incorrect address used in a DeduceSurface() call

lioncash:
Previously the source was being deduced twice in a row.


src/video_core/texture_cache/texture_cache.h
```diff
@@ -1053,7 +1053,7 @@ class TextureCache {
     void DeduceBestBlit(SurfaceParams& src_params, SurfaceParams& dst_params,
                         const GPUVAddr src_gpu_addr, const GPUVAddr dst_gpu_addr) {
         auto deduced_src = DeduceSurface(src_gpu_addr, src_params);
-        auto deduced_dst = DeduceSurface(src_gpu_addr, src_params);
+        auto deduced_dst = DeduceSurface(dst_gpu_addr, src_params);
```
Morph1984, line 1056: Shouldn't it also be dst_params instead of src_params as we are trying to deduce the destination surface?

PR #4129, opened 2020-06-20
gl_shader_decompiler: Workaround textureLod when GL_EXT_texture_shadow_lod is not available

### LINKED ISSUE #3969: OpenGl: using textureLod not supported for sampler2DArrayShadow
When loading up BOTW (version 1.6.0 + DLC 1,2) on an AMD R9 290 graphics card multiple shaders will fail to compile with the following error:

```
[  87.355571] Render.OpenGL <Error> video_core\renderer_opengl\gl_shader_util.cpp:LoadShader:48: Error compiling compute shader:
Compute shader failed to compile with the following errors:
ERROR: 0:3698: error(#202) No matching overloaded function found: textureLod
ERROR: error(#273) 1 compilation errors.  No code generated
 
[  87.355590] Render.OpenGL <Error> video_core/renderer_opengl/gl_shader_util.h:LoadProgram:79: Error linking shader:
Compute shader(s) were not successfully compiled before glLinkProgram() was called.  Link failed. 
```
 

This is because the textureLod(sampler2DArrayShadow, vec4, float) function requires the [EXT_texture_shadow_lod](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_shadow_lod.txt) extension which is not supported on that graphics card.

According to a similar issue in another Project (https://github.com/KhronosGroup/SPIRV-Cross/issues/207) you could use the textureGrad function as a workaround for sampler2DArrayShadow sampling.

yuzu Mainline 260
Full log with shader source: https://pastebin.com/WQ6U3va0

Morph1984:
Fixes #3969 

Enable `GL_EXT_texture_shadow_lod` if exposed by the gpu driver.

Utilizes the workaround described [here](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L5872) and [here](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_glsl.cpp#L5949) if `GL_EXT_texture_shadow_lod` isn't available, allowing the compute shaders in The Legend of Zelda: Breath of the Wild to compile on Intel/AMD's proprietary drivers.

In my short testing I have not noticed any visual changes.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -37,6 +37,7 @@ using Tegra::Shader::IpaMode;
 using Tegra::Shader::IpaSampleMode;
 using Tegra::Shader::PixelImap;
 using Tegra::Shader::Register;
+using Tegra::Shader::TextureType;
 using VideoCommon::Shader::BuildTransformFeedback;
 using VideoCommon::Shader::Registry;
 
@@ -526,6 +527,9 @@ private:
         if (device.HasImageLoadFormatted()) {
             code.AddLine("#extension GL_EXT_shader_image_load_formatted : require");
         }
+        if (device.HasTextureShadowLod()) {
+            code.AddLine("#extension GL_EXT_texture_shadow_lod : require");
+        }
         if (device.HasWarpIntrinsics()) {
             code.AddLine("#extension GL_NV_gpu_shader5 : require");
             code.AddLine("#extension GL_NV_shader_thread_group : require");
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -909,13 +913,13 @@ class GLSLDecompiler final {
                     return "samplerBuffer";
                 }
                 switch (sampler.type) {
-                case Tegra::Shader::TextureType::Texture1D:
```
VolcaEM, line 912: Why? (above, too)

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2041,8 +2059,17 @@ class GLSLDecompiler final {
...
 
-        std::string expr = GenerateTexture(
-            operation, "Lod", {TextureArgument{Type::Float, meta->lod}, TextureOffset{}});
+        std::string expr{};
+
+        if (!meta->sampler.is_shadow || device.HasTextureShadowLod()) {
+            expr = GenerateTexture(operation, "Lod",
+                                   {TextureArgument{Type::Float, meta->lod}, TextureOffset{}});
+        } else if (meta->sampler.is_shadow && !device.HasTextureShadowLod()) {
+            LOG_ERROR(Render_OpenGL,
```
VolcaEM, line 2068: Maybe log as warning instead of error

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
Morph1984:
```cpp
using Tegra::Shader::TextureType;
```

VolcaEM: Oops I missed that

PR #4136, opened 2020-06-21
Add a "Open Mods Page" button to the GUI

VolcaEM:
ui_change

It simply opens [the Mods page](https://github.com/yuzu-emu/yuzu/wiki/Switch-Mods) when clicked

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1800,7 +1800,7 @@ void GMainWindow::OnMenuReportCompatibility() {
 }
 
 void GMainWindow::OnSwitchModsPage() {
-    const std::string mods_page_url = "https://github.com/yuzu-emu/yuzu/wiki/Switch-Mods";
+    const auto mods_page_url = QStringLiteral("https://github.com/yuzu-emu/yuzu/wiki/Switch-Mods");
     const QString mods_page_url_qs = QString::fromStdString(mods_page_url);
     const QUrl mods_page(mods_page_url_qs);
     const bool open = QDesktopServices::openUrl(mods_page);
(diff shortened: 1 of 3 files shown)


src/yuzu/main.cpp
```diff
@@ -1797,6 +1799,17 @@ void GMainWindow::OnMenuReportCompatibility() {
     }
 }
 
+void GMainWindow::OnSwitchModsPage() {
+    const std::string mods_page_url = "https://github.com/yuzu-emu/yuzu/wiki/Switch-Mods";
```
lioncash, line 1803:
```suggestion
    const auto mods_page_url = QStringLiteral("https://github.com/yuzu-emu/yuzu/wiki/Switch-Mods");
```

provides the same behavior without the need to convert between string types.

src/yuzu/main.cpp
```diff
@@ -1797,6 +1799,17 @@ void GMainWindow::OnMenuReportCompatibility() {
     }
 }
 
+void GMainWindow::OnSwitchModsPage() {
+    const auto mods_page_url = QStringLiteral("https://github.com/yuzu-emu/yuzu/wiki/Switch-Mods");
+    const QString mods_page_url_qs = QString::fromStdString(mods_page_url);
```
lioncash, line 1804: This line can be removed now, since the URL above it is now always in the correct format.

src/yuzu/main.cpp
```diff
@@ -1797,6 +1799,16 @@ void GMainWindow::OnMenuReportCompatibility() {
     }
 }
 
+void GMainWindow::OnSwitchModsPage() {
```
FearlessTobi, line 1802: nit: Shouldn't this be named `OnOpenModsPage`?

VolcaEM: Yes, I noticed this when I made the PR but forgot about it 
I will change this later

PR #4139, opened 2020-06-21
vk_rasterizer: Add requirements for VK_EXT_extended_dynamic_state

ReinUsesLisp:
This adds the requirements needed for using [VK_EXT_extended_dynamic_state](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_EXT_extended_dynamic_state) that will allow us to reduce the number of cached pipelines and reduce the size of the pipeline key.

This extension also allows us to implement #4105 for all vendors supporting this extension.

Nvidia hasn't published a [beta driver](https://developer.nvidia.com/vulkan-driver) at the moment of writing this so the extension itself is enabled when present, but it's unused.

### DIFF
--- a/src/video_core/renderer_vulkan/fixed_pipeline_state.cpp
+++ b/src/video_core/renderer_vulkan/fixed_pipeline_state.cpp
@@ -39,28 +39,21 @@ constexpr std::array POLYGON_OFFSET_ENABLE_LUT = {
 
 } // Anonymous namespace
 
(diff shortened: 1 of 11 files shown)


goldenx86:
GPU vendors seem to be in pain recently, there hasn't been any new driver in ages.

ReinUsesLisp:
This introduces validation errors I couldn't see before adding #4140. Investigating.

Miksel12:
Nvidia just released a new beta with VK_EXT_extended_dynamic_state.

PR #4140, opened 2020-06-22
renderer_vulkan: Update validation layer name and test before enabling

ReinUsesLisp:
Update validation layer string to `VK_LAYER_KHRONOS_validation`.

While we are at it, properly check for available validation layers before enabling them.


Morph1984:
This PR allows Intel GPUs to initialize the Vulkan renderer again

PR #4146, opened 2020-06-22
gl_rasterizer: Return early in Draw if the offset is nullptr

VolcaEM:
I haven't tested this properly so there might be regressions


FernandoS27:
use `ASSERT_OR_EXECUTE` for that, so it logs an error.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -616,6 +616,9 @@ void RasterizerOpenGL::Draw(bool is_indexed, bool is_instanced) {
         const GLint base_vertex = static_cast<GLint>(gpu.regs.vb_element_base);
         const GLsizei num_vertices = static_cast<GLsizei>(gpu.regs.index_array.count);
         const GLvoid* offset = reinterpret_cast<const GLvoid*>(index_buffer_offset);
+        if (offset == nullptr) {
+            return;
+        }
```
ogniK5377, line 621:
```suggestion
        ASSERT_OR_EXECUTE(offset != nullptr, { return; });
```

ogniK5377:
Is this still being worked on or can It be closed?

PR #4150, opened 2020-06-24
vulkan: Use VK_EXT_extended_dynamic_state when available

ReinUsesLisp:
Supersedes #4139.

Same as the linked PR, this time implementing the usage of [VK_EXT_extended_dynamic_state](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_EXT_extended_dynamic_state).

The graphics pipeline key has been rearranged to support hashing and `memcmp`ing with a variable size, this way we ignore state that can be changed dynamically.

As Miksel12 has pointed [here](https://github.com/yuzu-emu/yuzu/pull/4139#issuecomment-648377016), beta drivers for Nvidia are available, and there's also a merge request on mesa for ANV. This implementation has only been tested against Nvidia's driver without validation layers (as these don't exist/are not public yet).

### DIFF
--- a/src/video_core/renderer_vulkan/fixed_pipeline_state.cpp
+++ b/src/video_core/renderer_vulkan/fixed_pipeline_state.cpp
@@ -39,28 +39,21 @@ constexpr std::array POLYGON_OFFSET_ENABLE_LUT = {
 
 } // Anonymous namespace
 
-void FixedPipelineState::DepthStencil::Fill(const Maxwell& regs) noexcept {
-    raw = 0;
-    front.action_stencil_fail.Assign(PackStencilOp(regs.stencil_front_op_fail));
-    front.action_depth_fail.Assign(PackStencilOp(regs.stencil_front_op_zfail));
-    front.action_depth_pass.Assign(PackStencilOp(regs.stencil_front_op_zpass));
-    front.test_func.Assign(PackComparisonOp(regs.stencil_front_func_func));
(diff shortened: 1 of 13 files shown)


src/video_core/renderer_vulkan/fixed_pipeline_state.cpp
```diff
@@ -147,20 +131,57 @@ void FixedPipelineState::BlendingAttachment::Fill(const Maxwell& regs, std::size
...
-    rasterizer.Fill(regs);
-    depth_stencil.Fill(regs);
-    color_blending.Fill(regs);
-    viewport_swizzles.Fill(regs);
+void FixedPipelineState::DynamicState::Fill(const Maxwell& regs) {
+    const u32 topology_index = static_cast<u32>(regs.draw.topology.Value());
+    u32 packed_front_face = PackFrontFace(regs.front_face);
+    if (regs.screen_y_control.triangle_rast_flip != 0) {
+        // Flip front face
+        packed_front_face = 1 - packed_front_face;
```
VolcaEM, line 139: packed_front_face is an unsigned int (u32), so, if you subtract a number that is greater than 1, there will be an underflow, if I'm right: are you sure this is correct?

ReinUsesLisp: `packed_front_face` can only be 1 or 0, so there's no underflow.

PR #4156, opened 2020-06-25
gl_texture_cache: Avoid copying between images with different formats

ReinUsesLisp:
Skip copies between images with different formats. Copies between illegal formats (e.g. RG8UI and R16F) corrupt Nvidia's driver on Turing devices and some of these are forbidden by ARB_copy_image.

As a temporary fix to avoid corrupting the driver, avoid copying these images.

@FernandoS27 Any thoughts on a proper fix for this? Currently we even copy between R16F and Z16 on Breath of the Wild.


FernandoS27:
R16F to Z16F should be permitted though...

ReinUsesLisp:
Superseded by #4176.

PR #4164, opened 2020-06-25
hotkeys: Add a "Mute Audio" hotkey

Kewlan:
Mutes the audio by returning 0 to VolumeAdjustSamples() when toggled. The already existing volume setting won't be affected.

I decided to not make it save to the config file as muting audio may be a temporary thing for the current play session, but I'll change it if requested.

### DIFF
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -127,6 +127,13 @@ void LogSettings() {
     LogSetting("Services_BCATBoxcatLocal", Settings::values.bcat_boxcat_local);
 }
 
+float Volume() {
+    if (values.audio_muted) {
+        return 0.0f;
+    }
+    return values.volume;
+}
+
 bool IsGPULevelExtreme() {
     return values.gpu_accuracy == GPUAccuracy::Extreme;
 }
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -736,6 +736,9 @@ void GMainWindow::InitializeHotkeys() {
                                     Settings::values.use_docked_mode);
                 dock_status_button->setChecked(Settings::values.use_docked_mode);
             });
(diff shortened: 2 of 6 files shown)


src/core/settings.cpp
```diff
@@ -127,6 +127,12 @@ void LogSettings() {
     LogSetting("Services_BCATBoxcatLocal", Settings::values.bcat_boxcat_local);
 }
 
+float Volume() {
+    if (values.audio_muted)
```
lioncash, line 131: This should be braced

src/yuzu/main.cpp
```diff
@@ -736,6 +736,9 @@ void GMainWindow::InitializeHotkeys() {
                                     Settings::values.use_docked_mode);
                 dock_status_button->setChecked(Settings::values.use_docked_mode);
             });
+    connect(hotkey_registry.GetHotkey(main_window, QStringLiteral("Mute Audio"), this),
+            &QShortcut::activated, this,
+            [&] { Settings::values.audio_muted = !Settings::values.audio_muted; });
```
lioncash, line 741:
```suggestion
            [] { Settings::values.audio_muted = !Settings::values.audio_muted; });
```

src/core/settings.cpp
```diff
@@ -127,6 +127,13 @@ void LogSettings() {
     LogSetting("Services_BCATBoxcatLocal", Settings::values.bcat_boxcat_local);
 }
 
+float Volume() {
+    if (values.audio_muted) {
+        return 0;
```
lioncash, line 132: Missed this in the initial review (my bad), but this should be a float literal (`0.0f`) to prevent conversion warnings.

Kewlan: No worries. Thanks for reviewing so fast and also teaching how to solve it, like with `std::pair` last time. 🙂

PR #4169, opened 2020-06-26
am: Update function tables and add missing classes

VolcaEM:
This was based on Switchbrew page: https://switchbrew.org/wiki/Applet_Manager_services


lioncash:
Classes should only be added if they're actually used in the source code, otherwise this can lead to unused class warnings.

Also please don't add comments all over the function tables

PR #4170, opened 2020-06-26
aoc: Update function table

VolcaEM:
This was based on Switchbrew page: https://switchbrew.org/wiki/NS_Services#aoc:u


lioncash:
Please remove the comments and simply add the entry. Looks good otherwise.

PR #4171, opened 2020-06-26
apm: Add comments to function table(s)

VolcaEM:
This was based on Switchbrew page: https://switchbrew.org/wiki/PPC_services


lioncash:
These comments aren't necessary, given one can look at SwitchBrew for this. This also provides more line noise than necessary, particularly if a load of these comments are added to a table.

Knowing a version isn't particularly relevant, since any functions that do get hit are documented in the commit implementing those commands.

PR #4173, opened 2020-06-26
bpc: Update function tables

VolcaEM:
This was based on Switchbrew page: https://switchbrew.org/wiki/PCV_services


lioncash:
Looks great once comments are removed

PR #4174, opened 2020-06-26
btdrv: Update function table

VolcaEM:
This was based on Switchbrew page: https://switchbrew.org/wiki/Bluetooth_Driver_services


lioncash:
Looks great once the comments are removed.

PR #4182, opened 2020-06-27
hotkeys: Fix issues caused when changing the fullscreen hotkey

Kewlan:
Currently when changing the fullscreen hotkey the old hotkey still works for entering fullscreen, while the new hotkey can toggle between. The context menu will also show the old hotkey until yuzu is restarted. And if you switch back to the first hotkey you'll have to press it twice to enter the fullscreen.

This fixes that.

At first I tried with 
`ui.action_Fullscreen->setShortcut(hotkey_registry.GetKeySequence(main_window, fullscreen));`
to keep it in the same style as the others above, but that caused the double press issue.

I also added 
`ui.action_Fullscreen->setShortcutContext(hotkey_registry.GetShortcutContext(main_window, fullscreen));`
for the same reason (I don't know what it does) but this fix works without it, so I don't know if it should be removed.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -666,6 +666,11 @@ void GMainWindow::InitializeHotkeys() {
     ui.action_Capture_Screenshot->setShortcutContext(
         hotkey_registry.GetShortcutContext(main_window, capture_screenshot));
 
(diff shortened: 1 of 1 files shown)


bunnei:
LGTM, thanks!

PR #4194, opened 2020-06-28
shader_cache: Fix use-after-free and orphan invalidation cache entries

ReinUsesLisp:
This fixes some cases where entries could have been removed multiple
times reading freed memory. To address this issue this commit removes
duplicates from entries marked for removal and sorts out the removal
process to fix another use-after-free situation.

Another issue fixed in this commit is orphan invalidation cache entries.
Previously only the entries that were invalidated in the current
operations had its entries removed. This led to more use-after-free
situations when these entries were actually invalidated but referenced
an object that didn't exist.

- Fixes vertex explosions on Xenoblade Chronicles 2
- It may fix shader issues on Smash Ultimate

### DIFF
--- a/src/video_core/shader_cache.h
+++ b/src/video_core/shader_cache.h
@@ -20,6 +20,7 @@ namespace VideoCommon {
 template <class T>
 class ShaderCache {
     static constexpr u64 PAGE_BITS = 14;
+    static constexpr u64 PAGE_SIZE = u64(1) << PAGE_BITS;
 
     struct Entry {
         VAddr addr_start;
@@ -87,8 +88,8 @@ protected:
         const VAddr addr_end = addr + size;
         Entry* const entry = NewEntry(addr, addr_end, data.get());
 
-        const u64 page_end = addr_end >> PAGE_BITS;
-        for (u64 page = addr >> PAGE_BITS; page <= page_end; ++page) {
+        const u64 page_end = (addr_end + PAGE_SIZE - 1) >> PAGE_BITS;
+        for (u64 page = addr >> PAGE_BITS; page < page_end; ++page) {
             invalidation_cache[page].push_back(entry);
         }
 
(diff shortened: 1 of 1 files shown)


src/video_core/shader_cache.h
```diff
@@ -20,6 +20,7 @@ namespace VideoCommon {
 template <class T>
 class ShaderCache {
     static constexpr u64 PAGE_BITS = 14;
+    static constexpr u64 PAGE_SIZE = u64(1) << PAGE_BITS;
```
ogniK5377, line 23:
```suggestion
    static constexpr u64 PAGE_SIZE = 1UUL << PAGE_BITS;
```

Morph1984: UUL or ULL

VolcaEM: unsigned long long so I guess ULL

FernandoS27: Well it turns out Unsigned long long does not mean the same in linux and windows, so his approach is safer.

ogniK5377:
Untagged due to softlocks being caused in pokemon

ReinUsesLisp:
Pushed a fix for it, there was a bug in `InvalidatePageEntries`. `index` was being increased when it shouldn't.
I played 15 battles on Pokémon Sword without experiencing a softlock.

Here's the diff:
```diff
diff --git a/src/video_core/shader_cache.h b/src/video_core/shader_cache.h
index 4a8fe7c2a..b7608fc7b 100644
--- a/src/video_core/shader_cache.h
+++ b/src/video_core/shader_cache.h
@@ -155,7 +155,8 @@ private:
     /// @param addr_end        Non-inclusive end address of the invalidation
     /// @pre invalidation_mutex is locked
     void InvalidatePageEntries(std::vector<Entry*>& entries, VAddr addr, VAddr addr_end) {
-        for (std::size_t index = 0; index < entries.size(); ++index) {
+        std::size_t index = 0;
+        while (index < entries.size()) {
             Entry* const entry = entries[index];
             if (!entry->Overlaps(addr, addr_end)) {
                 ++index;
```

PR #4195, opened 2020-06-28
gl_rasterizer: Avoid copying attrib

Morph1984:
Make attrib a const reference and make VertexType() take in a const reference of attrib as its parameter.
This saves ~40 MB in memory usage in Super Mario Odyssey


ogniK5377:
Copying from discord, there's no reason to pass a u32 as a const ref, it should use more memory. Generally, for small types we can just flat out copy them because of how small they are

Morph1984:
Closing as per @ogniK5377 's comment, and more tests have revealed that memory savings are negligible after more games have been tested, and after committed memory is taken into account.

PR #4204, opened 2020-06-29
renderer_vulkan: Create and properly use Vulkan 1.0 instances when 1.1 is not available

ReinUsesLisp:
Enable the required capabilities to use Vulkan 1.0 without validation errors and disable those that are not compatible with it.


bunnei:
@ReinUsesLisp needs a rebase

bunnei (approved):

PR #4206, opened 2020-06-29
cmake: fix fmt linking

RealJohnGalt:
On gcc/ld, and clang/lld, fmt::v6 symbols are excluded, so linking
fails. This fixes the issue.

Note: This was included in the FindBoost changes I shared with
BlinkHawk, however only they were merged. I'm not sure if it was missed,
or if there was an issue with this part of the change.


bunnei:
LGTM, thanks!

bunnei:
Reverting, since this breaks our build pipeline

PR #4208, opened 2020-06-29
common: unbreak build on BSDs

jbeich:
Regressed by #3954. From [error log](https://github.com/yuzu-emu/yuzu/files/4848484/yuzu-s20200629.log):
```c++
src/common/memory_detect.cpp:15:10: fatal error: 'sys/sysinfo.h' file not found
#include <sys/sysinfo.h>
         ^~~~~~~~~~~~~~~
```

[sysinfo(2)](http://man7.org/linux/man-pages/man2/sysinfo.2.html) doesn't exist on any of BSDs. [libsysinfo](https://github.com/bsdimp/libsysinfo) relies on [kvm_getswapinfo(3)](https://man.freebsd.org/kvm_getswapinfo/3) which doesn't exist on NetBSD and OpenBSD. Also, installing libsysinfo [is dangerous](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=242236).

### DIFF
--- a/src/common/memory_detect.cpp
+++ b/src/common/memory_detect.cpp
@@ -9,10 +9,12 @@
 // clang-format on
 #else
 #include <sys/types.h>
-#ifdef __APPLE__
+#if defined(__APPLE__) || defined(__FreeBSD__)
 #include <sys/sysctl.h>
-#else
+#elif defined(__linux__)
 #include <sys/sysinfo.h>
+#else
+#include <unistd.h>
 #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


src/common/memory_detect.cpp
```diff
@@ -42,11 +44,22 @@ static MemoryInfo Detect() {
     sysctlbyname("vm.swapusage", &vmusage, &sizeof_vmusage, NULL, 0);
     mem_info.TotalPhysicalMemory = ramsize;
     mem_info.TotalSwapMemory = vmusage.xsu_total;
-#else
+#elif defined(__FreeBSD__)
+    u_long physmem, swap_total;
+    std::size_t sizeof_u_long = sizeof(u_long);
+    // sysctlbyname(const char *, void *, size_t *, const void *, size_t);
+    sysctlbyname("hw.physmem", &physmem, &sizeof_u_long, NULL, 0);
```
lioncash, line 51:
```suggestion
    sysctlbyname("hw.physmem", &physmem, &sizeof_u_long, nullptr, 0);
```

src/common/memory_detect.cpp
```diff
@@ -42,11 +44,22 @@ static MemoryInfo Detect() {
...
+    sysctlbyname("vm.swap_total", &swap_total, &sizeof_u_long, NULL, 0);
```
lioncash, line 52:
```suggestion
    sysctlbyname("vm.swap_total", &swap_total, &sizeof_u_long, nullptr, 0);
```

src/common/memory_detect.cpp
jbeich: Changing only FreeBSD wouldn't be consistent, so I've applied to macOS as well.

jbeich: On FreeBSD the change is purely cosmestic, see https://github.com/freebsd/freebsd/commit/c8ed04c26b6758354853a6bed4629f71d0d01a7d

PR #4213, opened 2020-06-30
nvdrv/nvmap: Add cases for "Compr" and "Base"

VolcaEM:
From Switchbrew (https://switchbrew.org/wiki/NV_services):

"Returns info about a nvmap object. Identical to Linux driver, but extended with further params.

struct {
   __in  u32 handle;
   __in  u32 param;  // 1=SIZE, 2=ALIGNMENT, 3=BASE (returns error), 4=HEAP (always 0x40000000), 5=KIND, 6=COMPR (unused)
   __out u32 result;
 }; "

Base always returns an error (not sure if it's "InvalidValue" though), while Compr is unused, so nothing should change functionality-wise, I think

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvmap.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvmap.cpp
@@ -200,8 +200,13 @@ u32 nvmap::IocParam(const std::vector<u8>& input, std::vector<u8>& output) {
     case ParamTypes::Kind:
         params.result = object->kind;
         break;
+    case ParamTypes::Compr:
+        params.result = 0;
+        break;
+    case ParamTypes::Base:
     default:
-        UNIMPLEMENTED();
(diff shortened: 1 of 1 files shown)


ogniK5377:
Has this been tested? Do games request this?

VolcaEM:
I tested this on 4-5 games but none have ever used Compr or Base so far, from what I can tell (I 
had added a temporary LOG_CRITICAL in each case)

FearlessTobi:
What is the point of this PR then?

VolcaEM:
First and foremost, accuracy to the original hardware, which is very important to me
Secondly, maybe some game uses this and we don't know that yet; SwitchBrew might be wrong (about the fact that Compr is unused) and has been wrong in the past
This also isn't a change that reduces performance in any way and only adds 6 lines of code so I don't see why not make the PR

VolcaEM:
Doesn't seem to affect anything anyway

PR #214, opened 2018-02-22
Return RESULT_SUCCESS for unimplemented functions

mailwl:
... Like Citra does.
this PR lets yuzu not to crash on unimplemented functions

PR #4237, opened 2020-07-04
service: Auto stub fallback

ogniK5377:
For simple services, we can implement an automatic stub fallback to help with compatibility until a proper implementation is done.

This should not be merged as it's a hack, however, it's to improve compatibility until we have an actual implementation of HLE functions.


ogniK5377:
Untagged until an option is added, some games tend to stop booting with this fix

FearlessTobi:
Maybe you should elaborate, how this PR is any different from the one that was closed in 2018.
https://github.com/yuzu-emu/yuzu/pull/214

PR #4247, opened 2020-07-05
prepo: Stub SaveReport and SaveReportWithUser with PlayReportType=New

Morph1984:
Stub these 2 SaveReport commands introduced in the 10.0 nnsdk nso

### DIFF
--- a/src/core/hle/service/prepo/prepo.cpp
+++ b/src/core/hle/service/prepo/prepo.cpp
@@ -55,7 +55,8 @@ private:
         const auto process_id = rp.PopRaw<u64>();
 
         std::vector<std::vector<u8>> data{ctx.ReadBuffer(0)};
-        if constexpr (Type == Core::Reporter::PlayReportType::Old2) {
+        if constexpr (Type == Core::Reporter::PlayReportType::Old2 ||
+                      Type == Core::Reporter::PlayReportType::New) {
             data.emplace_back(ctx.ReadBuffer(1));
         }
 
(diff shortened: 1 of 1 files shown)


lioncash:
It should probably be uncovered what is actually new with these functions rather than passing it through, otherwise this is likely silently doing the wrong thing which is particularly problematic here, given this is mixing an incomplete/unknown implementation (`New`) with implementations that are complete (`Old` and `Old2`) for `SaveReport` and `SaveReportWithUser` templates.

src/core/hle/service/prepo/prepo.cpp
```diff
@@ -76,7 +76,8 @@ class PlayReport final : public ServiceFramework<PlayReport> {
         const auto process_id = rp.PopRaw<u64>();
         std::vector<std::vector<u8>> data{ctx.ReadBuffer(0)};
 
-        if constexpr (Type == Core::Reporter::PlayReportType::Old2) {
+        if constexpr (Type == Core::Reporter::PlayReportType::Old2 ||
+                      Type == Core::Reporter::PlayReportType::New) {
```
lioncash, line 80: Why isn't this similarly added to the if constexpr branch in `SaveReport` as well?

VolcaEM:
> It should probably be uncovered what is actually new with these functions rather than passing it through, otherwise this is likely silently doing the wrong thing which is particularly problematic here, given this is mixing an incomplete/unknown implementation (`New`) with implementations that are complete (`Old` and `Old2`) for `SaveReport` and `SaveReportWithUser` templates.
"New" prepo simply adds a check for another function in the "client" side, but it's very possible that they added new fields or changed previous fields

lioncash:
Yes. Which is why I said it should be investigated in the first place.

PR #4248, opened 2020-07-05
AM/ISelfController: Stub CreateManagedDisplaySeparableLayer

Morph1984:
Stub this by sending 1 layer id instead of 2 as yuzu does not support multiple layers per display.
No adverse side effects have been observed.
- Used by Animal Crossing: New Horizons Update 1.3.0


FernandoS27:
Just create the second layer and save it in a variable.

PR #4249, opened 2020-07-05
registered_cache: Remove previous update/dlc if it exists on install

Morph1984:
- This checks for and removes old updates or dlc based on title id. If a content meta nca exists within the registered cache, it will attempt to remove all the ncas associated with the content meta before installing a new update/dlc
- Now updates and DLC can be properly overwritten

Note: If there are more than 1 update versions installed, this will not remove the extras properly.

### DIFF
--- a/src/core/file_sys/registered_cache.cpp
+++ b/src/core/file_sys/registered_cache.cpp
@@ -547,7 +547,7 @@ InstallResult RegisteredCache::InstallEntry(const XCI& xci, bool overwrite_if_ex
     return InstallEntry(*xci.GetSecurePartitionNSP(), overwrite_if_exists, copy);
 }
 
-bool RegisteredCache::RemoveExistingEntry(const u64 title_id) {
+bool RegisteredCache::RemoveExistingEntry(u64 title_id) {
     const auto delete_nca = [this](const NcaID& id) {
         const auto path = GetRelativePathFromNcaID(id, false, true, false);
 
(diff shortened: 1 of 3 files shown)


src/core/file_sys/registered_cache.cpp
```diff
@@ -560,31 +616,44 @@ InstallResult RegisteredCache::InstallEntry(const NSP& nsp, bool overwrite_if_ex
...
 
-    // Install Metadata File
     const auto meta_id_raw = (*meta_iter)->GetName().substr(0, 32);
     const auto meta_id = Common::HexStringToArray<16>(meta_id_raw);
 
+    const auto section0 = (*meta_iter)->GetSubdirectories()[0];
+    const auto cnmt_file = section0->GetFiles()[0];
+    const CNMT cnmt(cnmt_file);
+
+    // Get the title id stored within the CNMT
```
lioncash, line 626: This is self explanatory from the code.

src/core/file_sys/registered_cache.cpp
```diff
@@ -560,31 +616,44 @@ InstallResult RegisteredCache::InstallEntry(const NSP& nsp, bool overwrite_if_ex
...
+    const auto title_id = cnmt.GetTitleID();
+    // Removes an entry if it exists
```
lioncash, line 628: Ditto

src/core/file_sys/registered_cache.h
```diff
@@ -154,6 +155,9 @@ class RegisteredCache : public ContentProvider {
         std::optional<TitleType> title_type = {}, std::optional<ContentRecordType> record_type = {},
         std::optional<u64> title_id = {}) const override;
 
+    // Removes an existing entry based on title id
+    bool RemoveExistingEntry(const u64 title_id);
```
lioncash, line 159:
```suggestion
    bool RemoveExistingEntry(u64 title_id);
```

src/core/file_sys/registered_cache.cpp
```diff
@@ -547,6 +547,62 @@ InstallResult RegisteredCache::InstallEntry(const XCI& xci, bool overwrite_if_ex
...
+            return false;
+        }
+
+        Core::Crypto::SHA256Hash hash{};
+        mbedtls_sha256_ret(id.data(), id.size(), hash.data(), 0);
+        const auto dirname = fmt::format("000000{:02X}", hash[0]);
+
+        const auto dir2 = GetOrCreateDirectoryRelative(dir, dirname);
+
+        const auto res = dir2->DeleteFile(fmt::format("{}.nca", Common::HexToString(id, false)));
```
VolcaEM, line 564: Does this check if the file exists?

src/core/file_sys/registered_cache.cpp
```diff
@@ -547,6 +547,62 @@ InstallResult RegisteredCache::InstallEntry(const XCI& xci, bool overwrite_if_ex
...
+
+        return res;
+    };
+
+    // Get the Content Provider
+    const auto& installed = Core::System::GetInstance().GetContentProvider();
```
VolcaEM, line 570:
```suggestion
    const auto& content_provider = Core::System::GetInstance().GetContentProvider();
```

src/core/file_sys/registered_cache.cpp
```diff
@@ -547,6 +547,62 @@ InstallResult RegisteredCache::InstallEntry(const XCI& xci, bool overwrite_if_ex
...
+    // If an update exists, remove
+    if (installed.HasEntry(title_id, ContentRecordType::Meta)) {
+        LOG_INFO(Loader,
+                 "Previous Update (v{}) for title_id={:016X} detected! Attempting to remove...",
+                 installed.GetEntryVersion(title_id).value_or(0), title_id);
+        // Get all the ncas associated with the current update CNMT and delete them
+        const auto& meta_old_id =
```
lioncash, line 577:
```suggestion
        const auto meta_old_id =
```

`GetNcaIDFromMetadata` returns by value. Ditto for all other occurrences below.

src/core/file_sys/registered_cache.cpp
lioncash, line 570: If possible, I'd avoid making use of the global system instance here, and instead opt for either storing a reference to the active system instance/content provider in the constructor, or passing a reference to the system instance/content provider into the function as a parameter.

src/core/file_sys/registered_cache.cpp
Morph1984: Yes, the implementation of DeleteFile already does all the necessary checks.

src/core/file_sys/registered_cache.cpp
```diff
@@ -560,31 +609,42 @@ InstallResult RegisteredCache::InstallEntry(const NSP& nsp, bool overwrite_if_ex
...
         const auto res2 = RawInstallNCA(*nca, copy, overwrite_if_exists, record.nca_id);
-        if (res2 != InstallResult::Success)
+        if (res2 != InstallResult::Success) {
             return res2;
+        }
     }
 
     Refresh();
+    if (result) {
+        return InstallResult::ErrorAlreadyExists;
```
Morph1984, line 646: This is intentional due to #3385

PR #4254, opened 2020-07-06
hotkeys: (HACK) don't set a shortcut for the screenshot hotkey

VolcaEM:
This *should* make it work again as intended, at least with the "default" settings (I have no idea why, but it worked for me)
I wasn't able to find the root of the problem unfortunately, but the bug was most probably caused by hotkey_registry
The screenshot hotkey didn't work as intended while in-game, and this PR attempts to fix that problem (in a "hacky" way, though)

This has only been tested by me, so it should be considered "untested"


lioncash:
Sorry, but a hack like this is the kind of thing that we don't particularly like to merge. Especially if it involves commenting code out and leaving it in the codebase, given this seemingly papers over an underlying issue.

PR #4258, opened 2020-07-06
configuration: Fix hotkeys that open file dialog in windowed mode

ameerj:
Also fixed a crash when emu_thread is nullptr and attempting to screenshot


bunnei:
LGTM, thank you!

PR #4261, opened 2020-07-07
input_common: GC Controller save and compare against analog origin state

ameerj:
This is an attempt to try and make my previous implementation a bit more universal and robust. After the merge to mainline I received quite a few reports that for some adapters/controllers the GC analog devices were not working as intended.

This saves the origin state of the device as yuzu is booting up, assuming it is in the idle state, then uses the origin state as a comparison to detect if an input is placed. May need to tweak some of the numbers as I get more data from users/testers.


src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -376,4 +385,36 @@ const std::array<GCState, 4>& Adapter::GetPadState() const {
...
+    if (padaxis == PadAxes::StickY)
+        return origin_status[port].stick_y;
+    if (padaxis == PadAxes::SubstickX)
+        return origin_status[port].substick_x;
+    if (padaxis == PadAxes::SubstickY)
+        return origin_status[port].substick_x;
+    if (padaxis == PadAxes::TriggerLeft)
+        return origin_status[port].trigger_left;
+    if (padaxis == PadAxes::TriggerRight)
+        return origin_status[port].trigger_right;
```
VolcaEM, line 401: {} for consistency

src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -376,4 +385,36 @@ const std::array<GCState, 4>& Adapter::GetPadState() const {
...
```
VolcaEM, line 417: same as ^

src/input_common/gcadapter/gc_poller.cpp
```diff
@@ -144,14 +146,19 @@ void GCButtonFactory::EndConfiguration() {
...
     float GetAxis(int axis) const {
         std::lock_guard lock{mutex};
         // division is not by a perfect 128 to account for some variance in center location
         // e.g. my device idled at 131 in X, 120 in Y, and full range of motion was in range
         // [20-230]
-        return (gcadapter->GetPadState()[port].axes.at(axis) - 128.0f) / 95.0f;
+        if (axis % 2 == 0)
+            return (gcadapter->GetPadState()[port].axes.at(axis) - origin_value_x) / 95.0f;
+        else
+            return (gcadapter->GetPadState()[port].axes.at(axis) - origin_value_y) / 95.0f;
```
VolcaEM, line 161: same as ^

src/input_common/gcadapter/gc_adapter.cpp
ameerj: Force of habit. Thanks for catching that

src/input_common/gcadapter/gc_adapter.h
```diff
@@ -102,6 +96,9 @@ class Adapter {
     std::array<GCState, 4>& GetPadState();
     const std::array<GCState, 4>& GetPadState() const;
 
+    int GetOriginValue(int port, int axis);
+    const int GetOriginValue(int port, int axis) const;
```
lioncash, line 100:
```suggestion
    int GetOriginValue(int port, int axis) const;
```

src/input_common/gcadapter/gc_adapter.h
```diff
@@ -102,6 +96,9 @@ class Adapter {
     std::array<GCState, 4>& GetPadState();
     const std::array<GCState, 4>& GetPadState() const;
 
+    int GetOriginValue(int port, int axis);
```
lioncash, line 99: This version of the function doesn't need to exist, since it doesn't modify any class instance state and returns by value. The const variant is the only implementation needed.

src/input_common/gcadapter/gc_adapter.h
```diff
@@ -155,6 +152,8 @@ class Adapter {
 
     std::array<Common::SPSCQueue<GCPadStatus>, 4> pad_queue;
     std::array<GCState, 4> state;
+    std::array<bool, 4> get_origin;
```
lioncash, line 155:
```suggestion
    std::array<bool, 4> get_origin{};
```

src/input_common/gcadapter/gc_poller.cpp
```diff
@@ -144,14 +146,20 @@ void GCButtonFactory::EndConfiguration() {
...
+          origin_value_x(adapter->GetOriginValue(port_, axis_x_)),
+          origin_value_y(adapter->GetOriginValue(port_, axis_y_)) {}
 
     float GetAxis(int axis) const {
         std::lock_guard lock{mutex};
         // division is not by a perfect 128 to account for some variance in center location
         // e.g. my device idled at 131 in X, 120 in Y, and full range of motion was in range
         // [20-230]
-        return (gcadapter->GetPadState()[port].axes.at(axis) - 128.0f) / 95.0f;
+        if (axis % 2 == 0) {
```
lioncash, line 158:
```cpp
const auto origin_value = axis % 2 == 0 ? origin_value_x : origin_value_y;

return (gcadapter->GetPadState()[port].axes.at(axis) - origin_value) / 95.0f;
```

avoids duplicating the whole expression.

src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -376,4 +385,51 @@ const std::array<GCState, 4>& Adapter::GetPadState() const {
...
+    if (padaxis == PadAxes::TriggerLeft) {
+        return origin_status[port].trigger_left;
+    }
+    if (padaxis == PadAxes::TriggerRight) {
+        return origin_status[port].trigger_right;
+    }
+    return 0;
+}
+
+const int Adapter::GetOriginValue(int port, int axis) const {
```
lioncash, line 412:
```cpp
int Adapter::GetOriginValue(int port, int axis) const {
    const auto& status = origin_status[port];

    switch (static_cast<PadAxes>(axis)) {
    case PadAxes::StickX:
        return status.stick_x;
    case PadAxes::StickY:
        return status.stick_y;
    case PadAxes::SubstickX:
        return status.substick_x;
    case PadAxes::SubstickY:
        return status.substick_y;
    case PadAxes::TriggerLeft:
        return status.trigger_left;
    case PadAxes::TriggerRight:
        return status.trigger_right;
    default:
        return 0;
    }
}
```

ameerj:
With this latest commit, users of GC adapter are expected to have it plugged in before startup. There is no longer a thread constantly checking the usb devices looking for the adapter. 

This _could_ yield some performance back to the rest of the program, especially for those who don't use the adapter.

A future refactoring/cleanup of the code is planned

bunnei:
@ameerj needs a rebase

PR #4263, opened 2020-07-07
screenshots: Option to save screenshots immediately in a specified directory + Linux workaround

lat9nq:
Currently, to save a screenshot, there is no option but to choose a path and filename each time the user wanted to take a screenshot. In Linux, the Capture Screenshot QFileDialog is blank **if** it is visible, thus it is completely unusable.

This PR allows the user to save a screenshot immediately, shortcutting the process in Windows, and provides a workaround for the dialog issue in Linux until an actual solution is found. To control this behavior, in Emulation -> Configure... -> General -> UI there are two new options. The first lets the user choose whether yuzu prompts where to store the screenshot (marked as Windows only). The second is a text box that allows the user to specify a default path to save screenshots. Beside that is a button that opens a QFileDialog letting the user choose the default screenshots path. The default path is `[yuzu directory]/screenshots`.

This PR was intended to make use of `UISettings::values.screenshot_path`, but ended up removing it altogether in favor of adding an entry to `FileUtil`. IMO it is debatable whether it belongs there, and I am open to changing how the default directory is determined.

This will hopefully be a welcome change to Windows users, but this was originally written as a workaround for the Linux issue. It seems all QFileDialogs have issues when a game is running in Linux, but the Capture Screenshot dialog is the most visible when a game is running, given its utility. Thus we have the motivation for this PR.

It is worth noting that building yuzu with MinGW and running it in Wine under Linux does not produce the same result as running it natively in Linux. The QFileDialog and saving screenshots with the prompt works well in Wine.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2153,17 +2153,26 @@ void GMainWindow::OnToggleFilterBar() {
 
 void GMainWindow::OnCaptureScreenshot() {
     OnPauseGame();
-    QFileDialog png_dialog(this, tr("Capture Screenshot"), UISettings::values.screenshot_path,
-                           tr("PNG Image (*.png)"));
-    png_dialog.setAcceptMode(QFileDialog::AcceptSave);
-    png_dialog.setDefaultSuffix(QStringLiteral("png"));
-    if (png_dialog.exec()) {
-        const QString path = png_dialog.selectedFiles().first();
-        if (!path.isEmpty()) {
-            UISettings::values.screenshot_path = QFileInfo(path).path();
-            render_window->CaptureScreenshot(UISettings::values.screenshot_resolution_factor, path);
+
+    const u64 title_id = Core::System::GetInstance().CurrentProcess()->GetTitleID();
+    const auto screenshot_path = FileUtil::GetUserPath(FileUtil::UserPath::ScreenshotsDir);
+    const auto date = QDateTime::currentDateTime()
+                          .toString(QStringLiteral("yyyy-MM-dd_hh-mm-ss-zzz"))
+                          .toStdString();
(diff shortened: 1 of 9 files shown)


src/yuzu/main.cpp
```diff
@@ -2052,17 +2052,26 @@ void GMainWindow::OnToggleFilterBar() {
...
-    png_dialog.setAcceptMode(QFileDialog::AcceptSave);
-    png_dialog.setDefaultSuffix(QStringLiteral("png"));
-    if (png_dialog.exec()) {
-        const QString path = png_dialog.selectedFiles().first();
-        if (!path.isEmpty()) {
-            UISettings::values.screenshot_path = QFileInfo(path).path();
-            render_window->CaptureScreenshot(UISettings::values.screenshot_resolution_factor, path);
+
+    const u64 title_id = Core::System::GetInstance().CurrentProcess()->GetTitleID();
+    const auto screenshot_path = FileUtil::GetUserPath(FileUtil::UserPath::ScreenshotsDir);
```
lioncash, line 2057:
```cpp
const auto screenshot_path = QString::fromStdString(FileUtil::GetUserPath(FileUtil::UserPath::ScreenshotsDir));
const auto date = QDateTime::currentDateTime()
                      .toString(QStringLiteral("yyyy-MM-dd_hh-mm-ss-zzz"));
QString filename = QStringLiteral("%1%2_%3.png")
                      .arg(screenshot_path)
                      .arg(title_id, 16, 16, QLatin1Char{'0'})
                      .arg(date);
```

Allows the same behavior, but doesn't churn string copies as much.

bunnei:
@lat9nq can you rebase this when you get a chance? Thank you!

PR #4137, opened 2020-06-21
GC Adapter Implementation

ameerj:
This is exciting. This is a libusb implementation for the First-party Nintendo GameCube adapter. There is inspiration from the Dolphin team implementation, I tried not to stray too far to ensure functionality. This is also my first contribution to an established C++ project, so I apologize if my code is sloppy.

The current implementation works as another input device. Mapping works just as usual. And all 4 ports work as expected. I played for about a half hour with my brother using two controllers and there was no issue.

Currently rumble, analog triggers, and the hot plug feature are not implemented.

PR #4266, opened 2020-07-07
gcadapter: unbreak build on FreeBSD

jbeich:
Regressed by #4137

- FreeBSD, OpenBSD, macOS use libc++ by default
- DragonFly, FreeBSD ship non-reference libusb
- 121af3646dad added `find_package(LibUSB)` but forgot to copy FindLibUSB.cmake from Dolphin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,8 +331,10 @@ endif()
 
 # Ensure libusb is properly configured (based on dolphin libusb include)
 find_package(LibUSB)
-add_subdirectory(externals/libusb)
-set(LIBUSB_LIBRARIES usb)
+if (NOT LIBUSB_FOUND)
+    add_subdirectory(externals/libusb)
+    set(LIBUSB_LIBRARIES usb)
+endif()
 
 
 # Prefer the -pthread flag on Linux.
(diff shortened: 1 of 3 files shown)


bunnei:
LGTM, thank you!

lat9nq:
Breaks building on Linux systems that have recent versions of LibUSB installed.

jbeich:
@lat9nq, can you open an issue documenting the error message? I don't have Linux. Maybe Yuzu can adapt by checking `LIBUSB_API_VERSION` in a few places.

lat9nq:
Sure, but first there's a possible fix by removing the NOT here https://cdn.discordapp.com/attachments/476077512984756254/730793171344424970/unknown.png. It fixes it for Linux systems. Can you test that on your system?

lat9nq:
Nevermind, I'll open an issue. That change breaks on the linux-fresh build container.

jbeich:
`NOT` cannot be removed because it'd make Yuzu use bundled libusb regardless of whether system libusb is available. As documented in 2af7e40 commit message bundled libusb doesn't build on FreeBSD.

PR #4269, opened 2020-07-07
configuration: Move "Force 30 FPS mode" to debug options

VolcaEM:
Ui screenshot


slashiee:
How does this even fall under being a debug option?

VolcaEM:
> 
> 
> How does this even fall under being a debug option?

My original plan was to remove it entirely, as, as far as I am aware, since it doesn't work with multicore and, now that multicore has been merged, most people have it turned on, if their CPU supports it.
Though, some people have asked me to *not* remove it, and that's why I placed it in the debugging options instead.
You shouldn't turn it on unless you want to use single-core and it's almost never the case now (if not for debugging), and that being a rare case, I find that it should stay in the "debugging options".

PR #4273, opened 2020-07-08
video_core: Add asynchronous shader decompilation and compilation

ogniK5377:
Massive thanks to Exzap from Cemu for their help!

This PR adds asynchronous shader building to yuzu. There are multiple issues with asynchronous shader building on a system like the Nintendo Switch. We cannot build every shader like this as there are many shaders that are only ran once. To fix this we attempt to deduce which shaders can and can't be built in parallel based on the GPU regs we know at the time.

All in all, most shader stutter should be reduced.


src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -436,14 +470,49 @@ ProgramSharedPtr ShaderCacheOpenGL::GeneratePrecompiledProgram(
...
     const GPUVAddr address{GetShaderAddress(system, program)};
 
+    if (device.UseAsynchronousShaders() && async_shaders.HasCompletedWork()) {
+        auto completed_work = async_shaders.GetCompletedWork();
+        for (auto& work : completed_work) {
+            Shader* shader = TryGet(work.cpu_address);
+            if (shader == nullptr) {
+                continue;
+            }
+            using namespace VideoCommon::Shader;
```
VolcaEM, line 492: I think this should be moved to the top of the file

ogniK5377: no

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,159 @@
...
+                        cpu_addr};
+    std::unique_lock lock(queue_mutex);
+    pending_queue.push_back(std::move(params));
+}
+
+void AsyncShaders::ShaderCompilerThread(Core::Frontend::GraphicsContext* context) {
+    using namespace std::chrono_literals;
+    while (!is_thread_exiting.load()) {
+        // Partial lock to allow all threads to read at the same time
+        if (HasWorkQueued()) {
```
lioncash, line 116: This can be inverted to unindent code.

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,159 @@
...
+                        compiler_settings,
+                        registry,
+                        cpu_addr};
+    std::unique_lock lock(queue_mutex);
+    pending_queue.push_back(std::move(params));
+}
+
+void AsyncShaders::ShaderCompilerThread(Core::Frontend::GraphicsContext* context) {
+    using namespace std::chrono_literals;
+    while (!is_thread_exiting.load()) {
```
lioncash, line 114: If the atomicity is only desirable, and no reordering needs to be cared about, this can be changed into:

```suggestion
    while (!is_thread_exiting.load(std::memory_order_relaxed)) {
```

src/video_core/shader/async_shaders.h
```diff
@@ -0,0 +1,103 @@
...
+
+    void QueueOpenGLShader(const OpenGL::Device& device, Tegra::Engines::ShaderType shader_type,
+                           u64 uid, std::vector<u64> code, std::vector<u64> code_b, u32 main_offset,
+                           VideoCommon::Shader::CompilerSettings compiler_settings,
+                           const VideoCommon::Shader::Registry& registry, VAddr cpu_addr);
+
+private:
+    void ShaderCompilerThread(Core::Frontend::GraphicsContext* context);
+
+    struct WorkerParams {
```
lioncash, line 80: Fairly sure this can be forward declared and be defined in the cpp file like:

```cpp
// Header

class Thing {
private:
    struct WorkerParams;
};

// cpp file:

struct Thing::WorkerParams {
    AsyncShaders::Backend backend;
    OpenGL::Device device;
    Tegra::Engines::ShaderType shader_type;
    u64 uid;
    std::vector<u64> code;
    std::vector<u64> code_b;
    u32 main_offset;
    VideoCommon::Shader::CompilerSettings compiler_settings;
    VideoCommon::Shader::Registry registry;
    VAddr cpu_address;
};
```

src/yuzu/configuration/configure_graphics_advanced.ui
```diff
@@ -72,6 +72,16 @@
           </property>
          </widget>
         </item>
+        <item>
+         <widget class="QCheckBox" name="use_asynchronous_shaders">
+          <property name="toolTip">
+           <string>Enabling this reduces shader stutter. Enables asynchronous shader compilation. This feature is experimental.</string>
```
lioncash, line 78:
```suggestion
           <string>Enables asynchronous shader compilation, which may reduce shader stutter. This feature is experimental.</string>
```

It likely a better way of combining this while reducing the redundancy in using variants of "Enable"

src/video_core/shader/async_shaders.h
ogniK5377: Doesn't look like it since we pass it into deque

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -162,6 +163,18 @@ RasterizerOpenGL::RasterizerOpenGL(Core::System& system, Core::Frontend::EmuWind
...
+
+    if (device.UseAsynchronousShaders()) {
+        // Get the amount of threads we can use(this can return zero)
+        const auto cpu_thread_count = std::thread::hardware_concurrency();
+        // Deduce how many "extra" threads we have to use.
+        const auto max_threads_unused = std::max(4u, cpu_thread_count) - 4;
+        // Specify an upper bound for how many threads we can use
+        const auto max_allowed_shader_threads = std::max(2u, max_threads_unused);
+        // Always allow at least 1 thread regardless of our settings
+        const auto worker_count = std::min(1u, max_allowed_shader_threads);
```
EliEron, line 175: I might be misunderstanding this code but should't this be `std::max` not `std::min`? As it's written now it cannot be set higher than 1u, but can be set lower which goes against the accompanying comment. It also makes the values assigned above it rather pointless.

ogniK5377: Nice catch, Thanks!

src/yuzu/configuration/config.cpp
```diff
@@ -1119,6 +1121,10 @@ void Config::SaveRendererValues() {
     WriteSettingGlobal(QStringLiteral("use_vsync"), Settings::values.use_vsync, true);
     WriteSettingGlobal(QStringLiteral("use_assembly_shaders"),
                        Settings::values.use_assembly_shaders, false);
+    WriteSettingGlobal(QStringLiteral("use_asynchronous_shaders"),
+                       Settings::values.use_asynchronous_shaders, false);
+    WriteSettingGlobal(QStringLiteral("use_asynchronous_shaders"),
+                       Settings::values.use_asynchronous_shaders, false);
```
Morph1984, line 1127: Duplicate

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -162,6 +163,23 @@ RasterizerOpenGL::RasterizerOpenGL(Core::System& system, Core::Frontend::EmuWind
...
+        constexpr auto RESERVED_THREADS = 6u;
+        // Get the amount of threads we can use(this can return zero)
+        const auto cpu_thread_count =
+            std::max(RESERVED_THREADS, std::thread::hardware_concurrency());
+        // Deduce how many "extra" threads we have to use.
+        const auto max_threads_unused = cpu_thread_count - RESERVED_THREADS;
+        // Always allow at least 1 thread regardless of our settings
+        const auto max_worker_count = std::max(1u, max_threads_unused);
+        // Don't use more than MAX_THREADS
+        const auto worker_count = std::min(max_worker_count, MAX_THREADS);
```
Morph1984, line 180:
```cpp
        // Max worker threads we should allow
        constexpr auto MAX_THREADS = 4U;
        // Deduce how many threads we can use
        const auto threads_used = std::thread::hardware_concurrency / 4;
        // Always allow at least 1 thread regardless of our settings
        const auto max_worker_count = std::max(1U, threads_used);
        // Don't use more than MAX_THREADS
        const auto worker_count = std::min(max_worker_count, MAX_THREADS);
        async_shaders.AllocateWorkers(worker_count);
```

Copying from Discord so this doesn't get missed

Thread usage will look like this:
```
CPU Threads  Shader Threads
       <= 7               1
       8-11               2
      12-15               3
      >= 16               4
```

bunnei:
Untagging, since there is a soft EA conflict. @ReinUsesLisp @ogniK5377 if this is good -- we can merge this weekend. If not, let's try to address the remaining issues sooner than later. Thanks!

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,175 @@
...
+// Refer to the license.txt file included.
+
+#include <chrono>
+#include "video_core/engines/maxwell_3d.h"
+#include "video_core/renderer_base.h"
+#include "video_core/renderer_opengl/gl_shader_cache.h"
+#include "video_core/shader/async_shaders.h"
+
+namespace VideoCommon::Shader {
+AsyncShaders::AsyncShaders(Core::Frontend::EmuWindow& emu_window) : emu_window(emu_window) {}
```
ReinUsesLisp, line 12: New line here

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,175 @@
...
+AsyncShaders::~AsyncShaders() {
```
ReinUsesLisp, line 13: Ditto

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,175 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <chrono>
```
ReinUsesLisp, line 5: Missing includes `<mutex>`, `<thread>`, `<condition_variable>`, `<vector>`

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // Clear our worker threads
+    worker_threads.clear();
+}
+
+bool AsyncShaders::HasWorkQueued() {
+    return !pending_queue.empty();
+}
+
+bool AsyncShaders::HasCompletedWork() {
+    std::shared_lock lock(completed_mutex);
```
ReinUsesLisp, line 67:
```suggestion
    std::shared_lock lock{completed_mutex};
```

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,175 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+std::vector<AsyncShaders::Result> AsyncShaders::GetCompletedWork() {
+    std::vector<AsyncShaders::Result> results;
+    {
+        std::unique_lock lock(completed_mutex);
```
ReinUsesLisp, line 93:
```suggestion
        std::unique_lock lock{completed_mutex};
```

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,175 @@
...
+                        cpu_addr};
+    std::unique_lock lock(queue_mutex);
+    pending_queue.push_back(std::move(params));
+    cv.notify_one();
+}
+
+void AsyncShaders::ShaderCompilerThread(Core::Frontend::GraphicsContext* context) {
+    using namespace std::chrono_literals;
+    while (!is_thread_exiting.load(std::memory_order_relaxed)) {
+        std::unique_lock<std::mutex> lock(queue_mutex);
```
ReinUsesLisp, line 127:
```suggestion
        std::unique_lock lock{queue_mutex};
```

src/video_core/shader/async_shaders.cpp
```diff
@@ -0,0 +1,175 @@
...
+        cv.wait(lock, [&] { return HasWorkQueued() || is_thread_exiting; });
```
ReinUsesLisp, line 128:
```suggestion
        cv.wait(lock, [this] { return HasWorkQueued() || is_thread_exiting; });
```

src/video_core/shader_notify.cpp
```diff
@@ -0,0 +1,42 @@
...
+    const auto diff = now - last_update;
+    if (diff > UPDATE_TICK) {
+        std::shared_lock lock(mutex);
+        last_updated_count = accurate_count;
+    }
+    return last_updated_count;
+}
+
+std::size_t ShaderNotify::GetShadersBuildingAccurate() {
+    std::shared_lock lock(mutex);
```
ReinUsesLisp, line 28:
```suggestion
    std::shared_lock lock{mutex};
```

src/video_core/shader_notify.cpp
```diff
@@ -0,0 +1,42 @@
...
+    return accurate_count;
+}
+
+void ShaderNotify::MarkShaderComplete() {
+    std::unique_lock lock(mutex);
```
ReinUsesLisp, line 33:
```suggestion
    std::unique_lock lock{mutex};
```

src/video_core/shader_notify.cpp
```diff
@@ -0,0 +1,42 @@
...
+    accurate_count--;
+}
+
+void ShaderNotify::MarkSharderBuilding() {
+    std::unique_lock lock(mutex);
```
ReinUsesLisp, line 38:
```suggestion
    std::unique_lock lock{mutex};
```

PR #4275, opened 2020-07-08
AM: fix GetDesiredLanguage:

CrazyMax:
try to get a control metadata from application update when is failed to get from the basic version.

Tested on Kirby Star Allies


bunnei:
LGTM, thank you!

PR #4276, opened 2020-07-08
input_common: Add more libusb error checks for GC Adapter

ameerj:
Some yuz-ers have reported crashes relating to libusb. This PR adds more error checking on libusb calls, logging errors as encountered and halting executions earlier when errors are encountered.

### DIFF
--- a/src/input_common/gcadapter/gc_adapter.cpp
+++ b/src/input_common/gcadapter/gc_adapter.cpp
@@ -24,9 +24,13 @@ Adapter::Adapter() {
     LOG_INFO(Input, "GC Adapter Initialization started");
 
     current_status = NO_ADAPTER_DETECTED;
-    libusb_init(&libusb_ctx);
 
-    StartScanThread();
+    const int init_res = libusb_init(&libusb_ctx);
+    if (init_res == LIBUSB_SUCCESS) {
+        StartScanThread();
+    } else {
+        LOG_ERROR(Input, "libusb could not be initialized. failed with error = {}", init_res);
+    }
 }
 
 GCPadStatus Adapter::GetPadStatus(int port, const std::array<u8, 37>& adapter_payload) {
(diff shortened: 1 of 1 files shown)


src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -215,6 +219,10 @@ void Adapter::Setup() {
 
     // populate the list of devices, get the count
     const std::size_t device_count = libusb_get_device_list(libusb_ctx, &devices);
```
bunnei, line 221: This needs to be signed

bunnei: e.g. ssize_t

src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -279,7 +288,13 @@ bool Adapter::CheckDeviceAccess(libusb_device* device) {
 
 void Adapter::GetGCEndpoint(libusb_device* device) {
     libusb_config_descriptor* config = nullptr;
-    libusb_get_config_descriptor(device, 0, &config);
+    const int config_descriptor_error = libusb_get_config_descriptor(device, 0, &config);
+    if (config_descriptor_error) {
```
bunnei, line 292: Should this be compared against LIBUSB_SUCCESS?

src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -215,6 +219,10 @@ void Adapter::Setup() {
 
     // populate the list of devices, get the count
     const std::size_t device_count = libusb_get_device_list(libusb_ctx, &devices);
+    if (device_count < 0) {
+        LOG_ERROR(Input, "libusb_get_device_list failed with error: {}", device_count);
+        return;
+    }
```
bunnei, line 225: Should we also check that devices != nullptr?

src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -215,6 +219,10 @@ void Adapter::Setup() {
```
bunnei, line 219: Please initialize devices above to `{}` so the memory is zero'd

src/input_common/gcadapter/gc_adapter.cpp
ameerj: `LIBUSB_SUCCESS` is 0, just a shorthand but I addressed this for better clarity.

PR #4282, opened 2020-07-09
filesystem: Set various NAND partition sizes to their defaults

Morph1984:
Removes the ability to configure the NAND partition sizes and sets them to hardware defaults:

NAND Total (size of rawnand.bin) = 29820 MiB
NAND User = 26624 MiB
NAND System = 2560 MiB

Furthermore, set the SDMC size to 1 TiB, putting it below the MBR partition table limit of 2 TiB, but allowing for a large enough size such that the SDMC will never run out of space.

Lastly, set the free space of the NAND user partition to be 1 MiB lower than its total space (explained in the last commit).
Due to the ability to install more than 26 GiB of updates and dlc to NAND, we should never return 0 here as some games like Bioshock 2 check this prior to save creation.

image


src/core/file_sys/sdmc_factory.cpp
```diff
@@ -10,6 +10,8 @@
 
 namespace FileSys {
 
+constexpr u64 SDMC_TOTAL_SIZE = 0x10000000000ULL; // 1 TiB
+
```
VolcaEM, line 14: I don't think this is right, and we should probably leave it as an option

BSoDGamingYT: Why?

slashiee: It seems fine to me. Besides, nobody is gonna cap out 1 TB of emulated space for an emulator.

src/core/file_sys/bis_factory.cpp
```diff
@@ -12,6 +12,10 @@
 
 namespace FileSys {
 
+constexpr u64 NAND_USER_SIZE = 0x680000000ULL;  // 26624 MiB
```
lioncash, line 15:
```suggestion
constexpr u64 NAND_USER_SIZE = 0x680000000;  // 26624 MiB
```

doesn't need to be included, given it's an integer constant being assigned to a u64. This will always be typed correctly, given there's no ambiguity. I realize the previous code did that sort of prefixing, but even then it wasn't necessary either.

src/core/file_sys/bis_factory.cpp
```diff
@@ -12,6 +12,10 @@
...
+constexpr u64 NAND_SYSTEM_SIZE = 0xA0000000;    // 2560 MiB
+constexpr u64 NAND_TOTAL_SIZE = 0x747C00000ULL; // 29820 MiB
```
lioncash, line 17: Ditto

src/core/file_sys/sdmc_factory.cpp
```diff
@@ -10,6 +10,8 @@
 
 namespace FileSys {
 
+constexpr u64 SDMC_TOTAL_SIZE = 0x10000000000ULL; // 1 TiB
```
lioncash, line 13: Ditto

PR #4291, opened 2020-07-11
input_common: First implementation of controller rumble

### LINKED ISSUE #3894: Rumble Support
This issue is to complete rumble support.

# Summary
we break out all the data and process it, we just don't vibrate the controller itself. Ultimately we just want to call SDL to update the rumble state and such.

https://github.com/yuzu-emu/yuzu/blob/fe2173429a678a497207deb5849438ed7bccbc35/src/core/hle/service/hid/controllers/npad.cpp#L513

# Background
@FearlessTobi already did a v0 of this change:

https://github.com/yuzu-emu/yuzu/compare/master...FearlessTobi:new-rumble

`input_common` code uses the factory pattern so that the implementation of the devices is completely independent of what the emulation core sees. 

The emulation core gets a `ButtonDevice` for each of the buttons and a `AnalogDevice` for each of the joysticks.

This works well for everything that is an `InputDevice` aka something that the emu core will read state from, but that doesn't really work for rumble, which is more like an `OutputDevice`. 

We tried adding `SendFeedback` to the `InputDevice` class, but that doesn't really make sense since an Input device represents each individual button or analog on a controller, and not the whole controller itself.

Let's add a new abstraction for `OutputDevice` which has a `SetState` virtual method and `RumbleDevice` should be an Output device that takes whatever the core needs to pass to it.

Maybe the stuff for `SDLHaptic` can be moved directly into `SDLJoystick` instead of in its own class.

german77:
This is an approach to implement a basic controller rumble. It enables vibrations on Pro Controllers that can use sdl_haptics. 
The first button of player settings is used as a reference to know which controller corresponds to which player.

From testing and debugging I found that controller_ids should be splitted into 4 bytes. The second rightmost byte always contains the player number.

Most of the code added should be changed with a better method. I'm looking to fix it until everyone it's happy with the code.

Closes #3894

### DIFF
--- a/src/core/hle/service/hid/controllers/npad.cpp
+++ b/src/core/hle/service/hid/controllers/npad.cpp
@@ -609,20 +609,31 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode) 
     }
 }
 
-void Controller_NPad::VibrateController(const std::vector<u32>& controller_ids,
+void Controller_NPad::VibrateController(const std::vector<u32>& controllers,
                                         const std::vector<Vibration>& vibrations) {
-    LOG_DEBUG(Service_HID, "(STUBBED) called");
+    LOG_TRACE(Service_HID, "called");
 
     if (!Settings::values.vibration_enabled || !can_controllers_vibrate) {
         return;
     }
-    for (std::size_t i = 0; i < controller_ids.size(); i++) {
-        std::size_t controller_pos = NPadIdToIndex(static_cast<u32>(i));
-        if (connected_controllers[controller_pos].is_connected) {
-            // TODO(ogniK): Vibrate the physical controller
+    bool success = true;
+    for (std::size_t i = 0; i < controllers.size(); ++i) {
+        if (!connected_controllers[i].is_connected) {
+            continue;
+        }
+        using namespace Settings::NativeButton;
+        const auto& button_state = buttons[i];
+        if (button_state[A - BUTTON_HID_BEGIN]) {
+            if (button_state[A - BUTTON_HID_BEGIN]->SetRumblePlay(
+                    vibrations[0].amp_high, vibrations[0].amp_low, vibrations[0].freq_high,
+                    vibrations[0].freq_low)) {
+                success = false;
+            }
         }
     }
-    last_processed_vibration = vibrations.back();
+    if (success) {
+        last_processed_vibration = vibrations.back();
+    }
 }
 
 Controller_NPad::Vibration Controller_NPad::GetLastVibration() const {
--- a/src/core/hle/service/hid/controllers/npad.h
+++ b/src/core/hle/service/hid/controllers/npad.h
@@ -121,7 +121,7 @@ public:
 
     void SetNpadMode(u32 npad_id, NPadAssignments assignment_mode);
 
-    void VibrateController(const std::vector<u32>& controller_ids,
+    void VibrateController(const std::vector<u32>& controllers,
                            const std::vector<Vibration>& vibrations);
 
     Vibration GetLastVibration() const;
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <array>
 #include <atomic>
+#include <chrono>
 #include <cmath>
 #include <functional>
 #include <mutex>
@@ -78,6 +79,33 @@ public:
         return state.axes.at(axis) / (32767.0f * range);
     }
 
+    bool RumblePlay(f32 amp_low, f32 amp_high, int time) {
+        const u16 raw_amp_low = static_cast<u16>(amp_low * 0xFFFF);
+        const u16 raw_amp_high = static_cast<u16>(amp_high * 0xFFFF);
+        // Lower drastically the number of state changes
+        if (raw_amp_low >> 11 == last_state_rumble_low >> 11 &&
+            raw_amp_high >> 11 == last_state_rumble_high >> 11) {
+            if (raw_amp_low + raw_amp_high != 0 ||
+                last_state_rumble_low + last_state_rumble_high == 0) {
+                return false;
+            }
+        }
+        // Don't change state if last vibration was < 20ms
+        const auto now = std::chrono::system_clock::now();
+        if (std::chrono::duration_cast<std::chrono::milliseconds>(now - last_vibration) <
+            std::chrono::milliseconds(20)) {
+            return raw_amp_low + raw_amp_high == 0;
+        }
+
+        last_vibration = now;
+        last_state_rumble_low = raw_amp_low;
+        last_state_rumble_high = raw_amp_high;
+        if (sdl_joystick) {
+            SDL_JoystickRumble(sdl_joystick.get(), raw_amp_low, raw_amp_high, time);
+        }
+        return false;
+    }
+
     std::tuple<float, float> GetAnalog(int axis_x, int axis_y, float range) const {
         float x = GetAxis(axis_x, range);
         float y = GetAxis(axis_y, range);
@@ -139,6 +167,9 @@ private:
     } state;
     std::string guid;
     int port;
+    u16 last_state_rumble_high;
+    u16 last_state_rumble_low;
+    std::chrono::time_point<std::chrono::system_clock> last_vibration;
     std::unique_ptr<SDL_Joystick, decltype(&SDL_JoystickClose)> sdl_joystick;
     std::unique_ptr<SDL_GameController, decltype(&SDL_GameControllerClose)> sdl_controller;
     mutable std::mutex mutex;
@@ -207,7 +238,7 @@ void SDLState::InitJoystick(int joystick_index) {
         sdl_gamecontroller = SDL_GameControllerOpen(joystick_index);
     }
     if (!sdl_joystick) {
-        LOG_ERROR(Input, "failed to open joystick {}", joystick_index);
+        LOG_ERROR(Input, "Failed to open joystick {}", joystick_index);
         return;
     }
     const std::string guid = GetGUID(sdl_joystick);
@@ -301,6 +332,12 @@ public:
 
     bool GetStatus() const override {
         return joystick->GetButton(button);
+    }
+
+    bool SetRumblePlay(f32 amp_high, f32 amp_low, f32 freq_high, f32 freq_low) const override {
+        const f32 new_amp_low = pow(amp_low, 0.5f) * (3.0f - 2.0f * pow(amp_low, 0.15f));
+        const f32 new_amp_high = pow(amp_high, 0.5f) * (3.0f - 2.0f * pow(amp_high, 0.15f));
+        return joystick->RumblePlay(new_amp_low, new_amp_high, 250);
     }
 
 private:
--- a/src/core/frontend/input.h
+++ b/src/core/frontend/input.h
@@ -33,6 +33,9 @@ public:
     virtual bool GetAnalogDirectionStatus(AnalogDirection direction) const {
         return {};
     }
+    virtual bool SetRumblePlay(f32 amp_high, f32 amp_low, f32 freq_high, f32 freq_low) const {
+        return {};
+    }
 };
 
 /// An abstract class template for a factory that can create input devices.
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -802,18 +802,18 @@ void Hid::EndPermitVibrationSession(Kernel::HLERequestContext& ctx) {
 
 void Hid::SendVibrationValue(Kernel::HLERequestContext& ctx) {
     IPC::RequestParser rp{ctx};
-    const auto controller_id{rp.Pop<u32>()};
+    const auto controller{rp.Pop<u32>()};
     const auto vibration_values{rp.PopRaw<Controller_NPad::Vibration>()};
     const auto applet_resource_user_id{rp.Pop<u64>()};
 
-    LOG_DEBUG(Service_HID, "called, controller_id={}, applet_resource_user_id={}", controller_id,
+    LOG_DEBUG(Service_HID, "called, controller={}, applet_resource_user_id={}", controller,
               applet_resource_user_id);
 
     IPC::ResponseBuilder rb{ctx, 2};
     rb.Push(RESULT_SUCCESS);
 
     applet_resource->GetController<Controller_NPad>(HidController::NPad)
-        .VibrateController({controller_id}, {vibration_values});
+        .VibrateController({controller}, {vibration_values});
 }
 
 void Hid::SendVibrationValues(Kernel::HLERequestContext& ctx) {
@@ -831,8 +831,6 @@ void Hid::SendVibrationValues(Kernel::HLERequestContext& ctx) {
 
     std::memcpy(controller_list.data(), controllers.data(), controllers.size());
     std::memcpy(vibration_list.data(), vibrations.data(), vibrations.size());
-    std::transform(controller_list.begin(), controller_list.end(), controller_list.begin(),
-                   [](u32 controller_id) { return controller_id - 3; });
 
     applet_resource->GetController<Controller_NPad>(HidController::NPad)
         .VibrateController(controller_list, vibration_list);


src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
 void Controller_NPad::VibrateController(const std::vector<u32>& controller_ids,
                                         const std::vector<Vibration>& vibrations) {
-    LOG_DEBUG(Service_HID, "(STUBBED) called");
 
     if (!can_controllers_vibrate) {
         return;
     }
     for (std::size_t i = 0; i < controller_ids.size(); i++) {
-        std::size_t controller_pos = NPadIdToIndex(static_cast<u32>(i));
+        const auto byte0 = controller_ids[i] % 256; // posible motor direction?
```
lioncash, line 523: Possible

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+        const auto byte1 = (controller_ids[i] >> 8) % 256; // controller number
+        const auto byte2 = (controller_ids[i] >> 16) % 256; // posible start/stop?
```
lioncash, line 525: Ditto

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+        const auto byte3 = (controller_ids[i] >> 24) % 256; // unknown
+        const auto controller_pos = NPadIdToIndex(static_cast<u32>(byte1));
+
         if (connected_controllers[controller_pos].is_connected) {
-            // TODO(ogniK): Vibrate the physical controller
+            f32 amplitude = vibrations[0].amp_high > vibrations[0].amp_low ? vibrations[0].amp_high
+                                                                     : vibrations[0].amp_low;
+            if (amplitude > 0) {
+                int joystick_index = connected_controllers[controller_pos].rumble_port;
```
lioncash, line 533:
```suggestion
                const int joystick_index = connected_controllers[controller_pos].rumble_port;
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+
+                //Amplitudes are to small, The motors can't spin with that little force
```
lioncash, line 535: Comments should have a space after the `//`. Also `to` should be `too`.

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+                amplitude *= 6;
```
lioncash, line 536:
```suggestion
                amplitude *= 6.0f;
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+                if (amplitude > 1)
```
lioncash, line 537: This and the following if statement can be collapsed into:

```cpp
amplitude = std::clamp(amplitude, 0.2f, 1.0f);
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+                    amplitude = 1;
+                if (amplitude < 0.2)
+                    amplitude = (f32)0.2;
+
+                //StartRumble function of sdl_impl.cpp should be called instead of this!!
+                SDL_Joystick* sdl_joystick = SDL_JoystickOpen(joystick_index);
+                if (!sdl_joystick) {
+                    LOG_ERROR(Service_HID, "failed to open joystick {}", joystick_index);
```
lioncash, line 545:
```suggestion
                    LOG_ERROR(Service_HID, "Failed to open joystick {}", joystick_index);
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+                    return;
+                }
+
+                SDL_Haptic* sdl_haptic = SDL_HapticOpenFromJoystick(sdl_joystick);
+                if (!sdl_haptic) {
+                    LOG_INFO(Service_HID, "controller does not support haptics {}", joystick_index);
```
lioncash, line 551:
```suggestion
                    LOG_INFO(Service_HID, "Controller does not support haptics {}", joystick_index);
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +515,51 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+                } else {
+                    if (SDL_HapticRumbleInit(sdl_haptic) < 0) {
+                        LOG_INFO(Service_HID, "Unable to initialize rumble {}", joystick_index);
+                    }
+                }
+
+                if (SDL_HapticRumblePlay(sdl_haptic, amplitude, 15) != 0) {
+                    LOG_INFO(Service_HID, "Warning: Unable to play rumble! {}", SDL_GetError());
+                }
+                //End of StartRumble
```
lioncash, line 561: This comment is unnecessary.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -240,6 +261,7 @@ void SDLState::HandleGameControllerEvent(const SDL_Event& event) {
         if (auto joystick = GetSDLJoystickBySDLID(event.jbutton.which)) {
             joystick->SetButton(event.jbutton.button, false);
         }
+
```
lioncash, line 264: This seems like an unnecessary change.

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -235,6 +237,17 @@ void Controller_NPad::OnLoadInputDevices() {
         std::transform(players[i].analogs.begin() + Settings::NativeAnalog::STICK_HID_BEGIN,
                        players[i].analogs.begin() + Settings::NativeAnalog::STICK_HID_END,
                        sticks[i].begin(), Input::CreateDevice<Input::AnalogDevice>);
+        if (players[i].type == Settings::ControllerType::ProController) {
+            std::string button = players[i].buttons[0];
+            std::string find("port:");
```
lioncash, line 242:
```suggestion
            constexpr std::string_view find = "port:";
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -235,6 +237,17 @@ void Controller_NPad::OnLoadInputDevices() {
...
+            size_t index;
+            if ((index = button.find(find)) != std::string::npos) {
+                int port = -1;
+                button = button.substr(index + 5, 1);
+                sscanf(button.c_str(), "%d", &port);
```
lioncash, line 247: This can be [std::from_chars](https://en.cppreference.com/w/cpp/utility/from_chars). Which doesn't require you to make a substring, preventing an unnecessary copy.

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -17,6 +18,7 @@
 #include "core/hle/kernel/writable_event.h"
 #include "core/hle/service/hid/controllers/npad.h"
 #include "core/settings.h"
+#include <regex>
```
lioncash, line 21: This seems to be unused.

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <array>
 #include <cstring>
+#include <SDL.h>
```
lioncash, line 8: I can't help but feel that the usage of SDL should be abstracted out into input-common, so the interface doesn't depend directly on a specific library, so that we can support other input libraries in the future if necessary (e.g. if we ever support Android way further down the line).

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +514,47 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
 void Controller_NPad::VibrateController(const std::vector<u32>& controller_ids,
                                         const std::vector<Vibration>& vibrations) {
-    LOG_DEBUG(Service_HID, "(STUBBED) called");
 
     if (!can_controllers_vibrate) {
         return;
     }
     for (std::size_t i = 0; i < controller_ids.size(); i++) {
-        std::size_t controller_pos = NPadIdToIndex(static_cast<u32>(i));
+        const u8 byte0 = controller_ids[i] % 256;         // possible motor direction?
```
lioncash, line 522:
```suggestion
        const u8 byte0 = static_cast<u8>(controller_ids[i] % 256);         // possible motor direction?
```

ditto for the other byte expressions, below this, otherwise this will cause type truncation warnings.

legend800:
How do I set this up to test? Connect Nin. pro controller and configure buttons and try in game like SMO?

german77:
> How do I set this up to test? Connect Nin. pro controller and configure buttons and try in game like SMO?

Yes, If the device supports haptics it should work without a problem. I have tested it with a xbox controller.

ogniK5377:
Can confirm this works beautifully in SMO. It might be worse adding a feature to adjust the rumble strength though!

legend800:
FWIW, Petergov also solved a similar perf issue recently on wiimote vibration in Cemu.

german77:
> There have been reports of this PR killing FPS on Xbox One controllers when rumbling.
> I had the exact same issue when making an own implementation for rumble.
> Going to block this for the time being so we can look into whether this is a code or a driver/SDL issue.

Rumble is called up 16 times per frame. I will see if something can be done to fix it.

legend800:
Test Results from latest build:

- It's not working in 3 games:  Streets of Rage 4, Ultra Street Fighter 2, Fast RMX (double checked settings)
- And as mentioned, the rumble is very strong, especially on an xone pad - like almost double expected. Someway to tone it down is needed. Until then, I gotta disable, it's just too annoying, despite being a huge fan of rumble since I got my first N64 rumble pak for Star Fox 64. :)

Tested Games with working rumble:
•	Bayonetta 2
•	Crash Team Racing
•	Donkey Kong Country TF
•	Mario Kart 8
•	Mario Tennis
•	Mother Russia
•	Smash Bros
•	Super Mario Maker 2
•	Super Mario Odyssey

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +502,29 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+            f32 amplitude = std::max(vibrations[0].amp_low, vibrations[0].amp_high);
+            if (amplitude > 0 && vibration_start) {
+                // Amplitudes are too small, The motors can't spin with that little force
+                amplitude *= 6.0f;
+                amplitude = std::clamp(amplitude, 0.2f, 1.0f);
+
+                using namespace Settings::NativeButton;
+                const auto& button_state = buttons[controller_pos];
+                button_state[A - BUTTON_HID_BEGIN]->SetRumblePlay(amplitude, 1);
+            }
```
ogniK5377, line 527: So I did some testing here for a few things, firstly the rumble should keep rumbling until a new rumble value is passed. That means an amplitude of 0 is considered value here. Passing `SDL_HAPTIC_INFINITY` as the time should be sufficient and dropping the amplitude > 0 check.

As for the scaling of the amplitude, a linear scale by multiplying by 6 tends to be a bit overkill in most games. I suggest replacing amplitude with something as follows
`const auto new_amplitude = pow(amplitude, 0.45f) * (3.0f - 2.0f * pow(amplitude, 0.225f));`

```suggestion
            const f32 amplitude = std::max(vibrations[0].amp_low, vibrations[0].amp_high);
            // Amplitudes are too small, The motors can't spin with that little force
            const auto new_amplitude =
                pow(amplitude, 0.45f) * (3.0f - 2.0f * pow(amplitude, 0.225f));

            using namespace Settings::NativeButton;
            const auto& button_state = buttons[controller_pos];
            button_state[A - BUTTON_HID_BEGIN]->SetRumblePlay(new_amplitude, SDL_HAPTIC_INFINITY);
```

german77:
Rumble on Ultra Street Fighter 2 works you just need to enable in the control options.
Fast RMX is calling a stubbed function and is not sending any rumble signals, I'm trying to implement it but it may be in another PR.

Rumble strength should be better now.

I have tested rumble on different machines and controllers. I did not encounter any stutter. It looks like it's a driver issue.

legend800:
Yeah, strength is better, still strong, but a future slider will address that. 

Can confirm USF2 has rumble - looks like it only does so on certain moves / last attack, which is why I missed it. Good luck with Fast RMX, SOR4 and other games that do funky things.

Does this add true HD rumble (for Switch controllers) or just generic rumble? I can't get my Pro controller to rumble at all to check, even with BetterJoy, but others don't have this issue. So some config thing on my side I guess.

Thanks for the follow-ups and great job on this!

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -502,15 +502,26 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
-    LOG_DEBUG(Service_HID, "(STUBBED) called");
 
     if (!can_controllers_vibrate) {
         return;
     }
     for (std::size_t i = 0; i < controller_ids.size(); i++) {
-        std::size_t controller_pos = NPadIdToIndex(static_cast<u32>(i));
-        if (connected_controllers[controller_pos].is_connected) {
-            // TODO(ogniK): Vibrate the physical controller
+        const u8 byte0 = static_cast<u8>(controller_ids[i] % 256); // possible motor direction?
```
FearlessTobi, line 510: Can't we use a Bitfield for extracting those values?
Also either way, those variables could have better naming in order to be able to omit those comments.

german77: Could you check if what I did was right? I never used Bitfield before

ExsertWizard:
> I have tested rumble on different machines and controllers. I did not encounter any stutter. It looks like it's a driver issue.

No one has mentioned this so far but, I recently did some testing and the stutter only occurs on Xbox One Controllers connected over Bluetooth, it works perfectly when connected over USB.

It seems that some strange issue occurs only in SMO, performance is constantly capped to 10fps, even with the in-game rumble option off.
On the other hand, Super Smash Bros. Ultimate stutters only when rumble activity is present, with big dips in performance when the controller's heavy vibration motors are actuated, not so much with the light ones.

german77:
Never tested with Bluetooth. SMO always rumble even when you feel nothing, SSBU only when It's necessary. I will take a look and see if I finally get that stutter and fix it. Worst case scenario rumble will be disabled with Bluetooth controllers.

raven02:
@german77 I also got same lag issue in ACHN when jump using Xbox One Bluetooth controller . Looking forward to get it fixed or at least has option to enable/disable .

bunnei:
Untagging, since this seems to cause a major perf issue in SMO (maybe other games) with Xbox One controllers. @german77 can you take a look at this issue? Thank you!

german77:
> Untagging, since this seems to cause a major perf issue in SMO (maybe other games) with Xbox One controllers. @german77 can you take a look at this issue? Thank you!

I'm tying to find a solution but no luck yet. It seems that this problems was present before 2017.  https://answers.microsoft.com/en-us/windows/forum/games_windows_10/xbox-one-s-controller-causing-major-performance/e69a5864-794f-4355-83d8-a018b7ade00f?page=7

bunnei:
Thanks for the quick turnaround @german77 ! Retagging for further testing :)

raven02:
Retested latest commit on ACHN , SMO and MK8D , it works perfectly with Xbox One Bluetooth controller.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -71,6 +71,23 @@ class SDLJoystick {
...
+        if (raw_amp_low >> 13 == last_state_rumble_low >> 13 &&
+            raw_amp_high >> 13 == last_state_rumble_high >> 13) {
+            if (!(raw_amp_low + raw_amp_high == 0 &&
+                  last_state_rumble_low + last_state_rumble_high != 0)) {
+                return true;
+            }
+        }
+        last_state_rumble_low = raw_amp_low;
+        last_state_rumble_high = raw_amp_high;
+        printf("%d %d \n", raw_amp_low, raw_amp_high);
```
ogniK5377, line 87: Unneeded printf

legend800:
Looks like this breaks Mario Tennis loading. Crashes with rumble on, fine with it off in today's EA. Notice the last events logged were rumble actions.

Rumble On:
[yuzu_log(1).txt](https://github.com/yuzu-emu/yuzu/files/5013403/yuzu_log.1.txt)

Rumble Off:
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5013404/yuzu_log.txt)
2020-08-02 18_29_38-#patreon-general

bunnei:
@german77 this needs a rebase when you get a chance. Thank you!

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -500,17 +500,23 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
     }
 }
 
-void Controller_NPad::VibrateController(const std::vector<u32>& controller_ids,
+void Controller_NPad::VibrateController(const std::vector<ControllerIds>& controller_ids,
                                         const std::vector<Vibration>& vibrations) {
-    LOG_DEBUG(Service_HID, "(STUBBED) called");
-
```
Morph1984, line 506: Please do not remove this or change to LOG_TRACE

Morph1984:
Also, squash your commits by running
`git rebase -i upstream/master`
Change all `pick` to `fixup` except the first commit

Then run: `git submodule update --recursive`

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -500,17 +500,24 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
...
+    LOG_TRACE(Service_HID, "called");
```
Morph1984, line 505: Leave the newline intact
```suggestion
    LOG_TRACE(Service_HID, "called");
    
```

legend800:
Mario Tennis works great now! Overall rumble intensity feels about right now too. Great progress!

2020-08-05 21_40_23-Window

src/core/hle/service/hid/controllers/npad.h
```diff
@@ -50,6 +50,17 @@ class Controller_NPad final : public ControllerBase {
     };
     static_assert(sizeof(NPadType) == 4, "NPadType is an invalid size");
 
+    struct ControllerIds {
```
lioncash, line 53:
```suggestion
    struct ControllerID {
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -500,18 +500,18 @@ void Controller_NPad::SetNpadMode(u32 npad_id, NPadAssignments assignment_mode)
     }
 }
 
-void Controller_NPad::VibrateController(const std::vector<ControllerIds>& controller_ids,
+void Controller_NPad::VibrateController(const std::vector<ControllerID>& controller_id,
```
lioncash, line 503: the variable itself can remain as `controller_ids`, given it's a container holding several controller IDs.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -71,6 +72,38 @@ class SDLJoystick {
...
+        // Lower drastically the number of state changes
+        if (raw_amp_low >> 11 == last_state_rumble_low >> 11 &&
+            raw_amp_high >> 11 == last_state_rumble_high >> 11) {
+            if (!(raw_amp_low + raw_amp_high == 0 &&
+                  last_state_rumble_low + last_state_rumble_high != 0)) {
+                return false;
+            }
+        }
+        // Don't change state if last vibration was < 20ms
+        std::chrono::time_point<std::chrono::system_clock> now;
```
lioncash, line 87:
```cpp
const auto now = std::chrono::system_clock::now();
```

Is a way more concise way of composing this line.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -71,6 +72,38 @@ class SDLJoystick {
...
+        now = std::chrono::system_clock::now();
+        if (std::chrono::duration_cast<std::chrono::milliseconds>(now - last_vibration) <
+            std::chrono::milliseconds(20)) {
+            if (raw_amp_low + raw_amp_high == 0) {
```
lioncash, line 91:
```cpp
return raw_amp_low + raw_amp_high == 0;
```

Is more concise here.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -71,6 +72,38 @@ class SDLJoystick {
         return state.axes.at(axis) / 32767.0f;
     }
 
+    bool RumblePlay(f32 amp_low, f32 amp_high, int time) {
+        const u16 raw_amp_low = static_cast<u16>(amp_low * 0xFFFF);
+        const u16 raw_amp_high = static_cast<u16>(amp_high * 0xFFFF);
+        // Lower drastically the number of state changes
+        if (raw_amp_low >> 11 == last_state_rumble_low >> 11 &&
+            raw_amp_high >> 11 == last_state_rumble_high >> 11) {
+            if (!(raw_amp_low + raw_amp_high == 0 &&
```
lioncash, line 81: Inverting the not allows this to be written without it needing to be present in the condition itself:

```suggestion
            if (raw_amp_low + raw_amp_high != 0 ||
                last_state_rumble_low + last_state_rumble_high == 0) {
```

Shoegzer:
@german77 Your PR works well!  The issues compiling as I had reported [here](https://github.com/yuzu-emu/yuzu/issues/4500) have been resolved and I was able to merge the PR and test, haven't encountered any issues performance or otherwise and rumble strength feels about right compared with the same games on my switch (though a slider would be good to have eventually).  Great work, and thank you for this nice feature.

Test system is linux mint 20, kernel 5.4, self-compiled with gcc 9.3.0.  Testing with Sony DS4 pad.

v1993:
On topic of slowdown and stutter: make sure that you're using SDL2 at least 2.0.12. Its [changelog](https://www.libsdl.org/tmp/SDL/WhatsNew.txt) mentions:

```
* Fixed blocking game controller rumble calls when using the HIDAPI driver
```

Last SDL2 available in Ubuntu repos is 2.0.10, one potentially having this problem. Conan have updated version.

v1993:
@legend800 This implements only basic rumble, but #4411 should allow proper HD Rumble functional on Switch controllers once done and merged, complete with fully accurate strength. HD Rumble is not that HD anyways, actually (there are only two acoustic actuators in entire Pro Controller).

Shoegzer:
@v1993 Thanks for pointing that out.  The rumble in this PR seemed sufficiently accurate to me in initial testing but perhaps I didn't test it enough to spot differences between HD vs. non-HD rumble.  Also, from what you say I assume the code in #4411 only provides HD rumble with native controllers and cannot approximate/simulate the feature with others such as the DS4?

german77:
The current input engine do not allow HD rumble on other controllers even if they support it. This PR only uses amplitude and ignores the frequency of the rumble. For HD rumble to work on DS4 and other future controllers that support it. A new input engine will be needed.

Shoegzer:
Thanks @german77, and that's certainly fine.  Everything is iterative and evolutionary.  I appreciate your efforts either way.

v1993:
@Shoegzer It is not possible to implement HD rumble using SDL2 and it is not even their fault. AFAIK it is not possible at all (at least) in Linux without directly accessing HID device like #4411 does (normal driver maps strength values to amplitude+frequency, for each side on procon). This means that one currently have to write low-lever driver in emulator to support rumble beyond left strength+right strength.

Proper solution would be to introduce a way to report rumble as two values per rumbler instead of one, but it would require additions both to SDL2 (somewhat trivial ones) and lower level stuff (less trivial ones). Long story short, it would need work far beyond bounds of yuzu.

P.S.: that PR looks abandoned, sadly, and I would expect it to have conflicts with existing driver at least on Linux.

german77:
@v1993 #4411 is not abandoned at all. It has many changes that are not committed right now. It fully supports motion, button mapping, using two controllers as one and other things. I'm waiting for motion and this PR to get into mainline. Devs do not like when I start chaining PR to get a single one working.

Shoegzer:
@v1993 Thanks, that makes sense.  I would think SDL2 doesn't support such a feature because there hasn't been a need for it, at least until more recently with projects like this.  Perhaps someone on this team or others such as ryujinx who understands HID matters can contribute code upstream at some future point for more widespread benefit.

What makes you say the PR is abandoned?  It doesn't have recent activity but that's normal as I assuming development is being done in spare time when RL issues aren't in the way, and that's not predictable.  I'm sure it will come in time and with patience.  That said, I do certainly hope linux issues such as the one you raised are addressed and I'm sure they will.

v1993:
@Shoegzer The thing is that SDL2 can't communicate to underlying driver to send HD Rumble command due to driver not exposing such interface.

Shoegzer:
@v1993 Thanks and yeah, I was thinking the development would have to progress on both low- and higher-level, was just focused on SDL2 when I wrote that.

toccata10:
Thanks for this rumble feature. I'm working with Archlinux. 
I have rumble on my ds4v2 when connected on Usb, but not on bluetooth. I asked on Discord, and some guy said that "rumble over BT has never worked for me".

So I tested both usb and bluetooth with fftest: the rumble works fine. I also tried with rpcs3, with both "evdev" and the "Dualshock4" config: they all rumble with bluetooth. Same for cemu (with wine and xinput emulation).

On a side note, I also tried ds4drv-cemuhook. And: motion works with bluetooth (but no rumble) and doesn't with usb (but rumble).

v1993:
@toccata10 This sounds interesting. From what I know, Wine uses SDL2 to implement joysticks and so does Yuzu. The difference might be that wine probably uses full Haptics API while Yuzu utilizes newer function that allows to send vibration easier.

FearlessTobi (changes requested):

Shoegzer:
Getting no rumble with this PR as merged to mainline with [this](https://github.com/yuzu-emu/yuzu-mainline/commit/4d0ae1a17a031488eadfa9133dac157fb71900c0) commit.  To test, I connected a DS4 pad, selected "Pro Controller", reconfigured buttons and toggled the "Vibration" checkbox on and off but no rumble occurs.  I then entered games that feature rumble on the actual Switch but no rumble occurs there either.  Rumble worked with the unmerged PR as I had noted above.

@german77 any thoughts here?

Test system is linux mint 20, kernel 5.4 and a DS4 pad, compiled using gcc 9.3.0

v1993:
@Shoegzer I doubt that it is a regression if rumble works in some games but not others. There were notes that some of methods games use to send vibrations aren't implemented by this PR.

Morph1984:
@Shoegzer You literally scared me with that lol, but yes, there are some games where rumble does not work currently as @v1993 mentioned

german77:
Please post any games where rumble is absent here #4738

Shoegzer:
Yeah, I was a bit concerned myself!  Sorry for the false alarm.

@v1993 I did notice that Pinball FX3's input options feature a toggle only for "HD Rumble", so it does make sense that the current rumble code doesn't expose any rumble in light of @german77's comment above:

> The current input engine do not allow HD rumble on other controllers even if they support it. This PR only uses amplitude and ignores the frequency of the rumble. For HD rumble to work on DS4 and other future controllers that support it. A new input engine will be needed.

PR https://github.com/yuzu-emu/yuzu/pull/4411 wouldn't help with use cases like this either, as that's for officail Joycon support rather than other pads such as the DS4 this test was done with.  It would be nice if support for approximating the HD rumble experience for non-official input devices could be added at some future point though.

v1993:
@Shoegzer That's not really it. This PR reads HD Rumble as it's really the only sort of rumble Switch produces (but not all ways (HD) rumble can be triggered, which is your problem) and *converts it to simple rumble*. Your quote is about HD Rumble being played as simple rumble on supported controllers as opposed to HD Rumble as one might expect.

Shoegzer:
I appreciate the explanation @v1993. I wasn't familiar with how the Switch handles rumble so I assumed it had multiple modes but apparently not, it's just that the method of triggering HD rumble in Pinball FX3 is different from other titles that have been known to work with this PR.  Also it seems the DS4 _could_ actually handle HD rumble (with frequency as well as amplitude), so the feature could be replicated rather than approximated given a rewrite of the input enigne that might recognize it.

PR #4292, opened 2020-07-11
hle: service: mii: Rewrite service to properly support creation of random and default miis.

bunnei:
This change rewrites our Mii services implementation. We've pretty much removed the old implementation, which included code to handle the Mii database (which was incomplete/incorrect, and based on assumptions from ~2 years ago). With the rewrite, we focus just on the basics that games need, which are:
* Proper default Mii data
* The ability to correctly generate random Mii data

This fixes Mario Kart 8 Deluxe, as well as the Mii Fighter mode in Super Smash Bros. Ultimate (screenshots below). These changes require a dump of the Mii system archive, which is used for the actual Mii models.

Credit to Ryujinx team for the Mii lookup tables used for generating random data, as well as the table of default Mii's that matches hardware (credit and copyright info is also cited in the source).

image
image
image
image


src/core/hle/service/mii/manager.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+template <typename T, std::size_t SourceArraySize, std::size_t DestArraySize>
+std::array<T, DestArraySize> ResizeArray(const std::array<T, SourceArraySize>& in) {
+    std::array<T, DestArraySize> out{};
+    std::memcpy(out.data(), in.data(), sizeof(T) * std::min(SourceArraySize, DestArraySize));
+    return out;
+}
+
+MiiInfo ConvertStoreDataToInfo(const MiiStoreData& data) {
+    MiiStoreBitFields bf{};
```
lioncash, line 48:
```suggestion
    MiiStoreBitFields bf;
```

Particularly given it's being memcpy'd into completely immediately on the next line.

src/core/hle/service/mii/manager.cpp
```diff
@@ -0,0 +1,482 @@
...
+MiiInfo MiiManager::BuildDefault(std::size_t index) {
+    return ConvertStoreDataToInfo(BuildDefaultStoreData(
+        GetArrayValue<DefaultMii>(RawData::DefaultMii.data(), index), user_id));
+}
+
+ResultVal<std::vector<MiiInfoElement>> MiiManager::GetDefault(SourceFlag source_flag) {
+    std::vector<MiiInfoElement> result;
+
+    if ((source_flag & SourceFlag::Default) == SourceFlag::None) {
+        return MakeResult<std::vector<MiiInfoElement>>(result);
```
lioncash, line 463:
```suggestion
        return MakeResult(result);
```

src/core/hle/service/mii/manager.cpp
```diff
@@ -0,0 +1,482 @@
...
+    }
+
+    for (std::size_t index = 0; index < DefaultMiiCount; index++) {
+        result.emplace_back(BuildDefault(index), Source::Default);
+    }
+
+    return MakeResult<std::vector<MiiInfoElement>>(result);
```
lioncash, line 470:
```suggestion
    return MakeResult(std::move(result));
```

src/core/hle/service/mii/raw_data.h
```diff
@@ -0,0 +1,2265 @@
...
+
+#pragma once
+
+#include <array>
+
+#include "common/common_types.h"
+
+namespace Service::Mii::RawData {
+
+static constexpr std::array<u8, 1728> DefaultMii{
```
lioncash, line 29: These should be defined in a cpp file and be made extern const in the header, so that these don't copied entirely into every translation unit that may reference these identifiers.

Thog:
Nice work on getting Mii support added.
When we were working on reversing Mii's for Ryujinx, we guessed a significant amount of the enums (names & properties) and structures, which seem to have been referenced here directly from Ryujinx code as well.

It would be appreciated if the appropriate license statements could be added to the top of those files as well.

Otherwise the code looks fine 👍

bunnei:
@Thog thanks for taking a look. Yes, sorry, several of the enum definitions were from Ryujinx. However, these were largely unused, and considering we have not independently verified them, I have just gone ahead and removed these.

PR #4294, opened 2020-07-11
configuration: Add settings to enable/disable specific CPU optimizations

MerryMage:
Expose fine-grained CPU optimization flags for debugging purposes only.

image
image
image


MerryMage:
Added tooltips and changed layouts as per internal discussion.

FearlessTobi:
Should we maybe add a Debug category and move the CPU and the Debug page there?
I'm just a bit worried that having this CPU category so prominently in the settings could lead some users to mess around with it.

raven02:
Should "Multicore CPU emulation" under CPU tab ?

MerryMage:
@FearlessTobi Added a Debug tab. Hope this is a sufficient deterent for you.

PR #4297, opened 2020-07-11
main/profile_select: Don't prompt for profile selection when only one is available

FearlessTobi:
This exists as an option on the real Switch as well and saves some hassle for the users.


raven02:
Looking forward to test it with LM3 which always prompt for user profile selection a title screen.

EliEron:
It would be nice if this was also applied to the "Open Save data folder" action as well. That action uses the dialog defined in the OnGameListOpenFolder method, and thus is not affected by this PR. Ideally it should be refactored to use the same dialog code as the ProfileSelectorSelectProfile method.

PR #4298, opened 2020-07-11
yuzu/configure_filesystem: Remove "Select Cache Directory" option

FearlessTobi:
This tab of the settings is already extremely bloated and the setting itself is quite useless.
With a gamelist of almost 30 games, the cache directory is smaller than 1MB for me and therefore I don't see why it needs to be configurable.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -615,12 +615,6 @@ void Config::ReadDataStorageValues() {
                                 QString::fromStdString(FS::GetUserPath(FS::UserPath::DumpDir)))
                         .toString()
                         .toStdString());
-    FS::GetUserPath(FS::UserPath::CacheDir,
-                    qt_config
-                        ->value(QStringLiteral("cache_directory"),
-                                QString::fromStdString(FS::GetUserPath(FS::UserPath::CacheDir)))
-                        .toString()
-                        .toStdString());
     Settings::values.gamecard_inserted =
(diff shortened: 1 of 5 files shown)


legend800:
Please try to not remove this. Giving users the ability to move things off their system drive (C) is important for backups and stuff. Google drive doesn't let you sync the Users dir. I also prefer to have everything in the app root anyway for convenience.

2020-07-11 10_30_13-yuzu Configuration

2020-07-11 10_28_15-D__Video Games_Switch_sys

Only these left that I can't back up:
2020-07-11 10_28_04-C__Users_Wes_AppData_Roaming_yuzu

FearlessTobi:
I don't really understand your argument here.
First of all, the cache doesn't need to be backed up, as it can literally be regenerated in a few seconds when loading the game list without it.
The option to move certain directories was specially made for those who contain large files and not for dirs where the files are so small that their size isn't even noticed by a normal user.
And lastly, you can always create a user folder in your yuzu root, giving you the ability to store your configuration and data in the yuzu install directory.

legend800:
Oh, I thought that stored shader cache but I guess there's another folder specifically for that strangely called 'shader'. The other emu has a cache folder with cpu/gpu/shader beneath so I was thinking of that. 

In that case, yeah not a big deal. It was really just a comment for creating a more 'portable' experience and not storing stuff on the win installation drive if possible.

Also, I created a user folder in the root and nothing happened. What gets stored here if the dir exist? Or are you saying I can create any folder and point the other (not cache) file paths there?

bunnei (approved):

PR #4300, opened 2020-07-11
Port citra-emu/citra#5441: "Common: remove a mod from AlignUp"

FearlessTobi:
See citra-emu/citra#5441 for more details.

**Original description**:
In cases where the size is not a known constant when inlining, `AlignUp<std::size_t>` currently generates two 64-bit div instructions.
This generates one div and a cmov which is significantly cheaper.
https://godbolt.org/z/S848QG


bunnei:
@FearlessTobi FYI reverting since things have been broken, please resubmit once you have a proper fix

PR #4303, opened 2020-07-11
common/alignment: Fix compilation errors

FearlessTobi:
Please wait for CI to pass before merging this.

### DIFF
--- a/src/common/alignment.h
+++ b/src/common/alignment.h
@@ -11,7 +11,9 @@ namespace Common {
 template <typename T>
 constexpr T AlignUp(T value, std::size_t size) {
     static_assert(std::is_unsigned_v<T>, "T must be an unsigned value.");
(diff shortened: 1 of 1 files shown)


src/common/alignment.h
```diff
@@ -11,7 +11,7 @@ namespace Common {
 template <typename T>
 constexpr T AlignUp(T value, std::size_t size) {
     static_assert(std::is_unsigned_v<T>, "T must be an unsigned value.");
-    auto mod{value % size};
+    T mod{value % size};
```
lioncash, line 14: I believe it would be more correct here to do:

```cpp
auto mod{static_cast<T>(value % size)};

bunnei:
FYI https://github.com/yuzu-emu/yuzu/pull/4305 -- please resubmit the original change when you have a working fix. Thanks!

PR #4280, opened 2020-07-09
cmake: pass libusb include directory

jbeich:
Regressed by #4266 
Fixes #4279

Reference libusb (unlike FreeBSD implementation) has `libusb-1.0/` directory prefix, so include directory from `pkg-config --cflags libusb-1.0` needs to be passed properly.

PR #4304, opened 2020-07-11
Fix build by adding libusb include dir

kevinxucs:
Fix build by adding libusb include dir


lioncash:
Hi! Thanks for the pull request, however #4280 already addresses this.

PR #4306, opened 2020-07-12
core/network: Add network abstraction

ReinUsesLisp:
This commit adds a network abstraction designed to implement bsd:s and nifm services but at the same time work as a generic abstraction to implement any networking code we have to use from core.

This is implemented on top of BSD sockets on Unix systems and winsock on Windows. The code is designed around winsocks having compatibility definitions to support both BSD and Windows sockets.

It is preferred to use this instead of `boost::asio` because the later is too high level for our needs. When implementing bsd:s procedures like `socket` we can't know ahead of time if this is going to be a client socket or an acceptor, but `boost::asio` wants to know this ahead of time.

The Linux code hasn't been tested at runtime because compilation is broken there and when attempting to remove the commits that broke it, I had submodule issues.

### DIFF
--- a/src/core/network/network.cpp
+++ b/src/core/network/network.cpp
@@ -1,0 +1,652 @@
+// Copyright 2020 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <algorithm>
+#include <cstring>
+#include <limits>
+#include <utility>
+#include <vector>
+
+#ifdef _WIN32
+#define _WINSOCK_DEPRECATED_NO_WARNINGS // gethostname
+#include <winsock2.h>
+#elif __unix__
+#include <errno.h>
+#include <fcntl.h>
+#include <netdb.h>
+#include <poll.h>
+#include <unistd.h>
+#else
+#error "Unimplemented platform"
+#endif
+
+#include "common/assert.h"
+#include "common/common_types.h"
+#include "common/logging/log.h"
+#include "core/network/network.h"
+#include "core/network/sockets.h"
+
+namespace Network {
+
+namespace {
+
+#ifdef _WIN32
+
+using socklen_t = int;
+
+void Initialize() {
+    WSADATA wsa_data;
+    (void)WSAStartup(MAKEWORD(2, 2), &wsa_data);
+}
+
+void Finalize() {
+    WSACleanup();
+}
+
+constexpr IPv4Address TranslateIPv4(in_addr addr) {
+    auto& bytes = addr.S_un.S_un_b;
+    return IPv4Address{bytes.s_b1, bytes.s_b2, bytes.s_b3, bytes.s_b4};
+}
+
+sockaddr TranslateFromSockAddrIn(SockAddrIn input) {
+    sockaddr_in result;
+
+#ifdef __unix__
+    result.sin_len = sizeof(result);
+#endif
+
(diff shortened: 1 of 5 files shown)


src/core/network/network.cpp
```diff
@@ -0,0 +1,640 @@
...
+        WSADATA wsa_data;
+        (void)WSAStartup(MAKEWORD(2, 2), &wsa_data);
+    }
+
+    ~Singleton() {
+        WSACleanup();
+    }
+};
+
+Singleton singleton;
```
lioncash, line 50: This should live and be released once the system instance ends. Using this for the entire duration of the application lifecycle is unnecessary

src/core/network/network.h
```diff
@@ -0,0 +1,87 @@
...
+void Initialize();
+
+/// Finalize network resources
+void Finalize();
+
+/// @brief Returns host's IPv4 address
+/// @return Pair of an array of human ordered IPv4 address (e.g. 192.168.0.1) and an error code
+std::pair<IPv4Address, Errno> GetHostIPv4Address();
+
+} // namespace Network
```
lioncash, line 87: Missing newline

src/core/network/network.cpp
```diff
@@ -0,0 +1,647 @@
...
+    value.l_linger = static_cast<u_short>(linger_value);
+    return value;
+}
+
+int LastError() {
+    return WSAGetLastError();
+}
+
+bool EnableNonBlock(SOCKET fd, bool enable) {
+    u_long value = enable ? 1 : 0;
```
VolcaEM, line 96:
```suggestion
    const u_long value = enable ? 1 : 0;
```

src/core/network/network.cpp
```diff
@@ -0,0 +1,647 @@
...
+    value.l_linger = linger_value;
+    return value;
+}
+
+int LastError() {
+    return errno;
+}
+
+bool EnableNonBlock(int fd, bool enable) {
+    int flags = fcntl(fd, F_GETFD);
```
VolcaEM, line 168:
```suggestion
    const int flags = fcntl(fd, F_GETFD);
```

src/core/network/network.cpp
```diff
@@ -0,0 +1,647 @@
...
+
+NetworkInstance::~NetworkInstance() {
+    Finalize();
+}
+
+std::pair<IPv4Address, Errno> GetHostIPv4Address() {
+    std::array<char, 256> name{};
+    if (gethostname(name.data(), static_cast<int>(name.size()) - 1) == SOCKET_ERROR) {
+        UNIMPLEMENTED_MSG("Unhandled gethostname error");
+        return {IPv4Address{}, Errno::SUCCESS};
```
VolcaEM, line 311: Why return success? Same below

src/core/network/network.cpp
ReinUsesLisp: It can't be const because the callee expects it to be a mutable.

src/core/network/network.cpp
ReinUsesLisp: It can't be const because the callee expects it to be a mutable.

src/core/network/network.cpp
ReinUsesLisp: Because I haven't encountered an environment where this fails. UNIMPLEMENTED_MSG will log it. We can add a meaningful error code once that happens.

PR #4309, opened 2020-07-12
fs: Fix RomFS building when zero byte files are present

Morph1984:
When zero byte files are present, the key (offset) for that file is identical to the file right after. A `std::map` isn't able to fit key-value pairs with identical keys (offsets). Therefore, the solution is to use `std::multimap` which permits multiple entries with the same key.

This most prominently fixes Pokemon Sword and Shield weather with any RomFS mod applied.

image

image

image

image


theboy181:
SWEEET!

PR #4313, opened 2020-07-12
CMakeLists: Unicorn requires Python >= 3.5

MerryMage:


MerryMage:
Ignore me, I was using a version of unicorn that was far ahead of the 3-year-old version yuzu uses.

PR #4314, opened 2020-07-12
gcadapter: Tidy up compiler warnings

lioncash:
Tidies up some compiler warnings and also removes some deprecated usages of lambda captures in C++20.


RealJohnGalt:
This fixes the random crash on yuzu launch that I was getting with the gc adapter code merged. :+1:

PR #4323, opened 2020-07-13
kernel/scheduler: Use std::mutex instead of spin lock

ReinUsesLisp:
Profiling shows that this is a highly contested mutex, causing diminishing
results compared to a OS lock. std::mutex implementations can spin for a
while before falling back to an OS lock.

This avoids wasting precious CPU cycles in a no-op.

### DIFF
--- a/src/core/hle/kernel/scheduler.h
+++ b/src/core/hle/kernel/scheduler.h
@@ -188,7 +188,7 @@ private:
 
     /// Scheduler lock mechanisms.
     bool is_locked{};
-    Common::SpinLock inner_lock{};
+    std::mutex inner_lock;
     std::atomic<s64> scope_lock{};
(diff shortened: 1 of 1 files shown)


FernandoS27:
My tests shows this caused instability. Either way, the original Switch OS uses a SpinLock for Critical sections which this lock emulates, it should remain as so.

ReinUsesLisp:
What do you mean by instability? If it causes crashes that means there's a race condition hidden by this. And what games use hundreds of switches? I can't test it without knowing that.

We can't have a busy wait wasting 8% of yuzu's time.

FernandoS27:
Not that kind of stability issues but of threads starving and causing weird slow-catch behaviors. Maybe it was another thing.

bunnei (approved):

PR #4330, opened 2020-07-13
input_configuration: Add range logic for analog sticks

ameerj:
This feature allows the user to modify the apparent radius/range of their analog sticks. In my testing, setting the range to below 50% made the sticks extremely sensitive (moving at rest), so the range slider allows for setting the value from [50, 150]

This PR now only provides the Analog stick logic for handling range coming from the UI params as to not conflict with other UI changes.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -461,7 +461,7 @@ public:
         const int axis_x = params.Get("axis_x", 0);
         const int axis_y = params.Get("axis_y", 1);
         const float deadzone = std::clamp(params.Get("deadzone", 0.0f), 0.0f, .99f);
-        const float range = std::clamp(params.Get("range", 0.0f), 0.0f, .99f) + 0.50f;
+        const float range = std::clamp(params.Get("range", 0.0f), 0.50f, 1.50f);
         auto joystick = state.GetSDLJoystickByGUID(guid, port);
 
         // This is necessary so accessing GetAxis with axis_x and axis_y won't crash
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -461,7 +461,7 @@ public:
         const int axis_x = params.Get("axis_x", 0);
         const int axis_y = params.Get("axis_y", 1);
         const float deadzone = std::clamp(params.Get("deadzone", 0.0f), 0.0f, .99f);
-        const float range = std::clamp(params.Get("range", 0.0f), 0.50f, 1.50f);
+        const float range = std::clamp(params.Get("range", 1.0f), 0.50f, 1.50f);
         auto joystick = state.GetSDLJoystickByGUID(guid, port);
 
         // This is necessary so accessing GetAxis with axis_x and axis_y won't crash
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -66,14 +66,14 @@ public:
         state.axes.insert_or_assign(axis, value);
     }
 
-    float GetAxis(int axis) const {
+    float GetAxis(int axis, float range) const {
         std::lock_guard lock{mutex};
-        return state.axes.at(axis) / 32767.0f;
+        return state.axes.at(axis) / (32767.0f * range);
     }
 
-    std::tuple<float, float> GetAnalog(int axis_x, int axis_y) const {
-        float x = GetAxis(axis_x);
-        float y = GetAxis(axis_y);
+    std::tuple<float, float> GetAnalog(int axis_x, int axis_y, float range) const {
+        float x = GetAxis(axis_x, range);
+        float y = GetAxis(axis_y, range);
         y = -y; // 3DS uses an y-axis inverse from SDL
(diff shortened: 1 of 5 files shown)


src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -376,6 +378,15 @@ ConfigureInputPlayer::ConfigureInputPlayer(QWidget* parent, std::size_t player_i
                 analogs_param[analog_id].Set("modifier_scale", slider_value / 100.0f);
             }
         });
+        connect(analog_map_range_slider[analog_id], &QSlider::valueChanged, [=] {
```
lioncash, line 381:
```suggestion
        connect(analog_map_range_slider[analog_id], &QSlider::valueChanged, [this, =] {
```

lioncash:
I feel `"Set Analog Stick"` can be omitted from the dialog, given it's obvious alone that the sliders are setting the deadzone and range without it. This should tidy up the dialog box a little bit.

ameerj:
> 
> 
> I feel `"Set Analog Stick"` can be omitted from the dialog, given it's obvious alone that the sliders are setting the deadzone and range without it. This should tidy up the dialog box a little bit.

@lioncash  `"Set Analog Stick"` is actually what allows yuzu internally from distinguishing if the user wants to use an `Analog` device or a binary `Axis Button` if the user decides to manually input the "button" associated with each direction. It's a bit clunky I agree, but would need a non-trivial rewrite.

FearlessTobi:
What would be a common usecase of this feature?

ameerj:
> 
> 
> What would be a common usecase of this feature?

@FearlessTobi 
Not all controllers are created equally. Some controllers don't fit the SDL standard range, and are not recognized as a full push even when pushing the stick all the way. This can be fixed by having the `range` value set to < 100%

Another use case is for controllers that very sharply indicate a full stick press, setting the range to above 100% gives a more gradual shift in stick intensity.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -66,14 +66,14 @@ class SDLJoystick {
...
+        return state.axes.at(axis) / (32767.0f * range);
     }
 
-    std::tuple<float, float> GetAnalog(int axis_x, int axis_y) const {
-        float x = GetAxis(axis_x);
-        float y = GetAxis(axis_y);
+    std::tuple<float, float> GetAnalog(int axis_x, int axis_y, float range) const {
+        float x = GetAxis(axis_x, range);
+        float y = GetAxis(axis_y, range);
         y = -y; // 3DS uses an y-axis inverse from SDL
```
VolcaEM, line 77: Is this still here from Citra?

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -66,14 +66,14 @@ class SDLJoystick {
...
         std::lock_guard lock{mutex};
-        return state.axes.at(axis) / 32767.0f;
+        return state.axes.at(axis) / (32767.0f * range);
     }
 
-    std::tuple<float, float> GetAnalog(int axis_x, int axis_y) const {
-        float x = GetAxis(axis_x);
-        float y = GetAxis(axis_y);
+    std::tuple<float, float> GetAnalog(int axis_x, int axis_y, float range) const {
+        float x = GetAxis(axis_x, range);
```
VolcaEM, line 75: const?

ameerj: Not quite. In the if block below, x and y are divided by a factor of r.

src/yuzu/configuration/configure_input_player.ui
```diff
@@ -811,7 +840,37 @@
           </item>
          </layout>
         </item>
-        <item row="6" column="1">
+
```
Morph1984, line 843: Unnecessary newline, ditto for the one below

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -458,13 +461,13 @@ class SDLAnalogFactory final : public Input::Factory<Input::AnalogDevice> {
         const int axis_x = params.Get("axis_x", 0);
         const int axis_y = params.Get("axis_y", 1);
         const float deadzone = std::clamp(params.Get("deadzone", 0.0f), 0.0f, .99f);
-
+        const float range = std::clamp(params.Get("range", 0.0f), 0.0f, .99f) + 0.50f;
```
Morph1984, line 464: Is there a reason why the range isn't from 0,0f to 1.0f?

ameerj: For some reason, the Qt sliders here range from 0-0.99. Couldn't figure out how to set it to 1.0. So I followed the existing standard.

Morph1984:
I would suggest squashing https://github.com/yuzu-emu/yuzu/pull/4330/commits/a750cb0f27f92b9a759ed6bc35d36fbad7adfd3c into https://github.com/yuzu-emu/yuzu/pull/4330/commits/2e97820fc88721f8ae1a39b1552ad8c08dd2d246
To prevent this from happening in the future be sure to rebase with upstream/master by running `git rebase upstream/master` and running `git submodule update` immediately after

PR #4334, opened 2020-07-14
configure_per_game: Clearer per-game settings

lat9nq:
In the original implementation of per-game configurations, the partially checked state of tristate check boxes were used to signal that a given setting would inherit its global configuration. In addition, many combo boxes stated "Use global configuration" to indicate the same behavior. Unfortunately it is not immediately clear to new users how tristate check boxes are supposed to work, and, assuming it is anyways, it is not convenient to go back-and-forth between the global config and the per-game config when the user forgets the global setting and wants to override it.

This PR aims to solve both of those problems. Check boxes still have three states, but they follow a different pattern:

1. they initially inherit the global state (can be on or off),
2. the first click negates the checkbox, and highlights it blue to indicate it will override the global config,
3. the second click negates it again, but it remains highlighted to still indicate an override,
4. a third click sets it back to the global state, without negating the checkbox since it will already be in that state.

In addition, combo boxes now say "Use global configuration ([global setting])" where [global setting] is replaced with the text corresponding to the selected global value. Their labels are also highlighted blue, with some exceptions, to indicate they will override the global setting.

This is purely a visual change to make per-game configs easier to understand and work with. Ideally nothing functional has changed.

Here's a screenshot of the roughest UI change, suggestions are welcome.
[](https://i.imgur.com/pEFSkst.png)
And [here's a video](https://cdn.discordapp.com/attachments/476077512984756254/732344258396422184/2020-07-13_17-09-52.mp4) of how checkboxes work.

Rei came up with this idea for this UI update, and BSoD also helped me sift through ideas to make the UI cleaner, so big thanks to them for this.

### DIFF
--- a/src/yuzu/configuration/configuration_shared.cpp
+++ b/src/yuzu/configuration/configuration_shared.cpp
@@ -10,10 +10,6 @@
 #include "yuzu/configuration/configuration_shared.h"
 #include "yuzu/configuration/configure_per_game.h"
 
-namespace ConfigurationShared {
-Trackers trackers = {};
-}
-
 void ConfigurationShared::ApplyPerGameSetting(Settings::Setting<bool>* setting,
                                               const QCheckBox* checkbox) {
     if (checkbox->checkState() == Qt::PartiallyChecked) {
--- a/src/yuzu/configuration/configuration_shared.cpp
+++ b/src/yuzu/configuration/configuration_shared.cpp
@@ -89,11 +89,11 @@ void ConfigurationShared::SetPerGameSetting(
 void ConfigurationShared::SetHighlight(QWidget* widget, const std::string& name, bool highlighted) {
     if (highlighted) {
         widget->setStyleSheet(
-            QStringLiteral("QWidget#%1 { border:2px solid;border-color:rgba(0,203,255,0.5) }")
+            QStringLiteral("QWidget#%1 { background-color:rgba(0,203,255,0.5) }")
                 .arg(QString::fromStdString(name)));
     } else {
         widget->setStyleSheet(
-            QStringLiteral("QWidget#%1 { border:2px solid;border-color:rgba(0,0,0,0) }")
+            QStringLiteral("QWidget#%1 { background-color:rgba(0,0,0,0) }")
                 .arg(QString::fromStdString(name)));
     }
     widget->show();
--- a/src/yuzu/configuration/configuration_shared.cpp
+++ b/src/yuzu/configuration/configuration_shared.cpp
@@ -95,42 +95,42 @@ void ConfigurationShared::SetHighlight(QWidget* widget, const std::string& name,
 
 void ConfigurationShared::SetColoredTristate(QCheckBox* checkbox, const std::string& name,
                                              const Settings::Setting<bool>& setting,
-                                             ConfigurationShared::CheckState& tracker) {
+                                             CheckState& tracker) {
     if (setting.UsingGlobal()) {
         tracker = CheckState::Global;
     } else {
         tracker = (setting.GetValue() == setting.GetValue(true)) ? CheckState::On : CheckState::Off;
     }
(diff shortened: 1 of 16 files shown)


src/yuzu/configuration/configuration_shared.cpp
```diff
@@ -4,10 +4,16 @@
...
 #include <QCheckBox>
 #include <QComboBox>
+#include <QObject>
+#include <QString>
 #include "core/settings.h"
 #include "yuzu/configuration/configuration_shared.h"
 #include "yuzu/configuration/configure_per_game.h"
 
+namespace ConfigurationShared {
+Trackers trackers = {};
```
lioncash, line 14: Please don't use global variables. We're trying to remove all usages of them in the codebase, and this runs counter to that.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -161,37 +148,42 @@ void ConfigureSystem::ApplyConfiguration() {
...
+                Settings::values.custom_rtc.SetValue(std::nullopt);
+            }
             break;
-        case Qt::PartiallyChecked:
+        case ConfigurationShared::CheckState::Global:
             Settings::values.custom_rtc.SetGlobal(false);
             Settings::values.custom_rtc.SetValue(std::nullopt);
             Settings::values.custom_rtc.SetGlobal(true);
             break;
+        case ConfigurationShared::CheckState::Count:;
```
lioncash, line 186: This should be using a break statement to make behavior more explicit. This just looks weird to most readers.

src/yuzu/configuration/configure_system.cpp
```diff
@@ -161,37 +148,42 @@ void ConfigureSystem::ApplyConfiguration() {
...
+                Settings::values.rng_seed.SetValue(std::nullopt);
+            }
             break;
-        case Qt::PartiallyChecked:
+        case ConfigurationShared::CheckState::Global:
             Settings::values.rng_seed.SetGlobal(false);
             Settings::values.rng_seed.SetValue(std::nullopt);
             Settings::values.rng_seed.SetGlobal(true);
             break;
+        case ConfigurationShared::CheckState::Count:;
```
lioncash, line 167: Ditto

src/yuzu/configuration/configuration_shared.h
```diff
@@ -15,8 +15,17 @@ constexpr int USE_GLOBAL_INDEX = 0;
 constexpr int USE_GLOBAL_SEPARATOR_INDEX = 1;
 constexpr int USE_GLOBAL_OFFSET = 2;
 
+enum CheckState {
```
lioncash, line 18:
```suggestion
enum class CheckState {
```

This makes it strongly typed and prevents generic names from entering the global namespace. It also allows the enum to be forward declared.

src/yuzu/configuration/configure_audio.h
```diff
@@ -6,6 +6,7 @@
 
 #include <memory>
 #include <QWidget>
+#include "yuzu/configuration/configuration_shared.h"
```
lioncash, line 9: This can be a forward declaration with the previous suggestion. Ditto for occurrences in other headers where trackers exist.

src/yuzu/configuration/configure_audio.h
```diff
@@ -37,4 +38,8 @@ class ConfigureAudio : public QWidget {
     void SetupPerGameUI();
 
     std::unique_ptr<Ui::ConfigureAudio> ui;
+
+    struct Trackers {
```
lioncash, line 42: We generally keep struct/class definitions separate from their uses, since it makes refactoring nicer if the struct needs to be moved in a future change. Ditto for other occurrences.

Kewlan:
I should have mentioned this in the first pull request, but what do you think about putting the config files in their own folder?

PR #4337, opened 2020-07-14
main: Set async gpu properly after loading per-game setting

lat9nq:
Another error that got pass me and only noticed when I was doing the per-game settings UI rework. This prevents asynchronous GPU emulation from being disabled while multi core is enabled as a result of a poorly put together per-game config.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1053,6 +1053,10 @@ void GMainWindow::BootGame(const QString& filename) {
     if (!(loader == nullptr || loader->ReadProgramId(title_id) != Loader::ResultStatus::Success)) {
         // Load per game settings
         Config per_game_config(fmt::format("{:016X}.ini", title_id), false);
+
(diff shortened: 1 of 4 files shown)


src/yuzu/main.cpp
```diff
@@ -1053,6 +1053,10 @@ void GMainWindow::BootGame(const QString& filename) {
     if (!(loader == nullptr || loader->ReadProgramId(title_id) != Loader::ResultStatus::Success)) {
         // Load per game settings
         Config per_game_config(fmt::format("{:016X}.ini", title_id), false);
+
+        Settings::values.use_asynchronous_gpu_emulation.SetValue(
```
lioncash, line 1057: Part of me feels like this is leaking behavior that should be handled elsewhere. I really don't think "BootGame" is an adequate place to handle changing configuration settings. Especially if we potentially have more settings of the same nature in the future.

PR #4341, opened 2020-07-14
Make the ExeFS dumper take applied updates into account

EliEron:
Currently the "Dump ExeFS" option always dumps the base ExeFS files regardless of whether you have updates enabled. This happens because the code responsible for dumping is placed above the code that checks for and applies updates. Simply swapping these around fixes the issue.

I am somewhat inexperienced when it comes to git and pull requests, so if I make any mistakes I can only apologize.

### DIFF
--- a/src/core/file_sys/patch_manager.cpp
+++ b/src/core/file_sys/patch_manager.cpp
@@ -80,16 +80,6 @@ VirtualDir PatchManager::PatchExeFS(VirtualDir exefs) const {
     if (exefs == nullptr)
         return exefs;
 
-    if (Settings::values.dump_exefs) {
-        LOG_INFO(Loader, "Dumping ExeFS for title_id={:016X}", title_id);
-        const auto dump_dir =
-            Core::System::GetInstance().GetFileSystemController().GetModificationDumpRoot(title_id);
-        if (dump_dir != nullptr) {
-            const auto exefs_dir = GetOrCreateDirectoryRelative(dump_dir, "/exefs");
-            VfsRawCopyD(exefs, exefs_dir);
-        }
-    }
-
     const auto& installed = Core::System::GetInstance().GetContentProvider();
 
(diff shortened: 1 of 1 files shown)


src/core/file_sys/patch_manager.cpp
```diff
@@ -80,16 +80,6 @@ VirtualDir PatchManager::PatchExeFS(VirtualDir exefs) const {
...
-        LOG_INFO(Loader, "Dumping ExeFS for title_id={:016X}", title_id);
-        const auto dump_dir =
-            Core::System::GetInstance().GetFileSystemController().GetModificationDumpRoot(title_id);
-        if (dump_dir != nullptr) {
-            const auto exefs_dir = GetOrCreateDirectoryRelative(dump_dir, "/exefs");
-            VfsRawCopyD(exefs, exefs_dir);
-        }
-    }
-
     const auto& installed = Core::System::GetInstance().GetContentProvider();
```
ogniK5377, line 83: It's a little out of scope but we might as well clean up the function while we're at it
```suggestion
    auto &system = Core::System::GetInstance();
    auto &fs_controller = system.GetFileSystemController();
    const auto& installed = system.GetContentProvider();
```
Just so we can eliminate calling `Core::System::GetInstance()` everywhere here

EliEron: Good idea, I've added a new commit that removes the redundant `Core::System::GetInstance()` from the function.

goldenx86:
I was working on adding https://github.com/yuzu-emu/yuzu/pull/7899 to the next progress report and noticed your discussion on it, can I mention you for this PR too?

EliEron:
Sure, that would be great 👍.

Also I have to say I'm impressive by how much research you do for these Progress reports, even digging though PR comments, it can't be easy work😄.

PR #4345, opened 2020-07-15
filesystem: Create subdirectories prior to creating a file

Morph1984:
If subdirectories exist in the given path parameter and don't exist in the real filesystem create them prior to creating the files within.
This fixes the softlocks upon save creation in The Legend of Zelda: Breath of the Wild


Gustav0ar:
> If subdirectories exist create them prior to creating the files within.
> This fixes the softlocks upon save creation in The Legend of Zelda: Breath of the Wild

If subdirectories *don't* exist .....

Morph1984:
> If subdirectories _don't_ exist .....

Should elaborate: If subdirectories exist *in the given path parameter* but don't exist in the real filesystem

PR #4352, opened 2020-07-16
configuration: Toggle controller vibrations

Morph1984:
Allow toggling controller vibrations on/off, this is exactly the same as the setting found in System Settings on hardware.

This setting is enabled by default

Does not conflict with #4291 

image

### DIFF
--- a/src/core/hle/service/hid/controllers/npad.cpp
+++ b/src/core/hle/service/hid/controllers/npad.cpp
@@ -504,7 +504,7 @@ void Controller_NPad::VibrateController(const std::vector<u32>& controller_ids,
                                         const std::vector<Vibration>& vibrations) {
     LOG_DEBUG(Service_HID, "(STUBBED) called");
 
-    if (!can_controllers_vibrate) {
+    if (!Settings::values.vibration_enabled || !can_controllers_vibrate) {
         return;
     }
     for (std::size_t i = 0; i < controller_ids.size(); i++) {
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -845,8 +845,7 @@ void Hid::CreateActiveVibrationDeviceList(Kernel::HLERequestContext& ctx) {
 void Hid::PermitVibration(Kernel::HLERequestContext& ctx) {
     IPC::RequestParser rp{ctx};
     const auto can_vibrate{rp.Pop<bool>()};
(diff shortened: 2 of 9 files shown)


german77:
I was wondering on how to add this option. Thanks for the example.

Morph1984:
> I was wondering on how to add this option. Thanks for the example.

No problem, great work on implementing rumble!

FearlessTobi:
Should we make this a per-controller setting? I see some cases where that might be useful

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -634,11 +634,11 @@ Controller_NPad::LedPattern Controller_NPad::GetLedPattern(u32 npad_id) {
...
 
 void Controller_NPad::SetVibrationEnabled(bool can_vibrate) {
-    can_controllers_vibrate = can_vibrate;
+    Settings::values.vibration_enabled = can_vibrate;
 }
 
 bool Controller_NPad::IsVibrationEnabled() const {
-    return can_controllers_vibrate;
+    return Settings::values.vibration_enabled;
 }
```
ogniK5377, line 642: These should not be changed because they are game-specific settings. It should remain under can_controllers_vibrate

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -504,7 +504,7 @@ void Controller_NPad::VibrateController(const std::vector<u32>& controller_ids,
                                         const std::vector<Vibration>& vibrations) {
     LOG_DEBUG(Service_HID, "(STUBBED) called");
 
-    if (!can_controllers_vibrate) {
+    if (!Settings::values.vibration_enabled) {
```
ogniK5377, line 507:
```suggestion
    if (!can_controllers_vibrate || !Settings::values.vibration_enabled) {
```

raven02:
Looking forward to have this in EA as #4352 has issue on those with Xbox One Bluetooth Controller .

FearlessTobi:
If this isn't a per-controller option and rumble causes issue on quite a few controllers, should this really be enabled by default?

german77:
> If this isn't a per-controller option and rumble causes issues on quite a few controllers, should this really be enabled by default?

Issues with rumble where fixed in latest EA. I haven't had more complains and some have confirmed that this is not an issue anymore

PR #4356, opened 2020-07-16
cpu_manager: Minor tidying up/header inclusions

lioncash:
Previously the header for cpu_manager was relying upon indirect includes. This PR makes them explicit, so include orderings being changed in other files don't break the compilation of things in the header.

While we're at it, we can also tidy up some unused variables and getters.


Gustav0ar:
Just confirming that this fixes the Linux build for me, thanks

PR #4369, opened 2020-07-17
macro_hle: Remove unnecessary std::make_pair calls

lioncash:
The purpose of make_pair is generally to deduce the types within the pair without explicitly specifying the types, so these usages were generally unnecessary, particularly when the type is enforced by the array declaration.


ogniK5377:
Just address clang issues

lioncash:
Done

PR #4372, opened 2020-07-17
game_list: Add "Remove" context menu

### LINKED ISSUE #2633: Deleting games installed to the NAND through the yuzu UI
The current process of deleting a game title, DLC/Update, that has been previously installed to the nand through the yuzu UI is a bit confusing. You need to guess for the file you want to delete based on the file size.

When a file is installed to the nand, the file is meant to be renamed to the NCAID of that particular title, however only part of the file is hashed resulting in the incorrect NCAID. 

Even if the file is already titled by it's [NCAID](https://imgur.com/AfPyKw1), the [new name is incorrect](https://imgur.com/gwWgf32) while the hash is still the same as the previous name.

Seeing as how the current method to find the right file to delete is impossible without generating your own hash of the file to check against some sort of database, please incorporate a function within the UI to delete the file without having to navigate folders and alphanumeric files that would confuse a casual user of the program.

Morph1984:
Adds the following actions to the context menu upon right clicking on a game entry.
- Remove Installed Update
- Remove All Installed DLC
- Remove Shader Cache
- Remove Custom Configuration
- Remove All Installed Contents

image

### Update/DLC:
https://imgur.com/a/PctG5k7

### Shader Cache:
https://imgur.com/a/G1VcuSE

### Custom Configuration:
https://imgur.com/a/NNZvHmt

### All Installed Contents:
https://imgur.com/a/1bopB5a

Closes #2633 

Update: Now supports removing SDMC installed titles as well as NAND installed titles.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1428,16 +1428,15 @@ void GMainWindow::OnGameListRemoveInstalledEntry(u64 program_id, InstalledEntryT
         RemoveAddOnContent(program_id, entry_type);
         break;
     }
-    game_list->PopulateAsync(UISettings::values.game_dirs);
     FileUtil::DeleteDirRecursively(FileUtil::GetUserPath(FileUtil::UserPath::CacheDir) + DIR_SEP +
                                    "game_list");
+    game_list->PopulateAsync(UISettings::values.game_dirs);
 }
 
 void GMainWindow::RemoveBaseContent(u64 program_id, const QString& entry_type) {
-    const auto res = Core::System::GetInstance()
-                         .GetFileSystemController()
-                         .GetUserNANDContents()
-                         ->RemoveExistingEntry(program_id);
+    const auto& fs_controller = Core::System::GetInstance().GetFileSystemController();
+    const auto res = fs_controller.GetUserNANDContents()->RemoveExistingEntry(program_id) ||
+                     fs_controller.GetSDMCContents()->RemoveExistingEntry(program_id);
 
     if (res) {
         QMessageBox::information(this, tr("Successfully Removed"),
@@ -1450,10 +1449,10 @@ void GMainWindow::RemoveBaseContent(u64 program_id, const QString& entry_type) {
 }
 
 void GMainWindow::RemoveUpdateContent(u64 program_id, const QString& entry_type) {
-    const auto res = Core::System::GetInstance()
-                         .GetFileSystemController()
-                         .GetUserNANDContents()
-                         ->RemoveExistingEntry(program_id | 0x800);
+    const auto update_id = program_id | 0x800;
+    const auto& fs_controller = Core::System::GetInstance().GetFileSystemController();
+    const auto res = fs_controller.GetUserNANDContents()->RemoveExistingEntry(update_id) ||
+                     fs_controller.GetSDMCContents()->RemoveExistingEntry(update_id);
 
     if (res) {
         QMessageBox::information(this, tr("Successfully Removed"),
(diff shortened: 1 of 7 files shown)


src/yuzu/main.cpp
```diff
@@ -1386,6 +1387,163 @@ static bool RomFSRawCopy(QProgressDialog& dialog, const FileSys::VirtualDir& src
     return true;
 }
 
+void GMainWindow::OnGameListRemoveInstalledEntry(u64 program_id, InstalledEntryType type) {
+    QString entry_type;
```
lioncash, line 1391:
```cpp
const QString entry_type = [this, type] {
    switch (type) {
    case InstalledEntryType::Game:
        return tr("Contents");
    case InstalledEntryType::Update:
        return tr("Update");
    case InstalledEntryType::AddOnContent:
        return tr("DLC");
    default:
        return QString{};
    }
}();
```

Lets you join the declaration and assignment together. Ditto for a similar switch in `OnGameListRemoveFile`

src/yuzu/main.cpp
```diff
@@ -1386,6 +1387,163 @@ static bool RomFSRawCopy(QProgressDialog& dialog, const FileSys::VirtualDir& src
...
+            this, tr("Remove Entry"), tr("Remove Installed Game %1?").arg(entry_type),
+            QMessageBox::Yes | QMessageBox::No, QMessageBox::No) != QMessageBox::Yes) {
+        return;
+    }
+
+    bool res;
+
+    switch (type) {
+    case InstalledEntryType::Game:
+        res = Core::System::GetInstance()
```
lioncash, line 1415: I'd likely suggest moving the individual cases to their own functions and then calling them in the switch here, given how large each case is. Ditto for the switch cases in `OnGameListRemoveFile`.

src/yuzu/main.cpp
```diff
@@ -1386,6 +1387,176 @@ static bool RomFSRawCopy(QProgressDialog& dialog, const FileSys::VirtualDir& src
...
+    QMessageBox::information(this, tr("Successfully Removed"),
+                             tr("Successfully removed %1 installed DLC.").arg(count));
+}
+
+void GMainWindow::OnGameListRemoveFile(u64 program_id, GameListRemoveTarget target) {
+    const QString question = [this, target] {
+        switch (target) {
+        case GameListRemoveTarget::ShaderCache:
+            return tr("Delete Transferable Shader Cache?");
+
```
lioncash, line 1492: out-of-place newline.

Kewlan:
Something to consider is if some or all of the files should be moved to the trash bin instead. Maybe just for the shader cache and config file since updates and DLC are duplicated when installed anyway.

Morph1984:
> Something to consider is if some or all of the files should be moved to the trash bin instead. Maybe just for the shader cache and config file since updates and DLC are duplicated when installed anyway.

Unfortunately Qt does not provide this kind of functionality, so it would add bloat to the code to implement it and potentially support it outside of Windows.

Kewlan:
> Unfortunately Qt does not provide this kind of functionality, so it would add bloat to the code to implement it and potentially support it outside of Windows.

Qt 5.15 added `QFile::moveToTrash()` so I guess this will have to wait until yuzu's Qt version is updated.

Its-Rei:
what's the value of moving to trash here? the user intentionally selected to delete whatever he chose and then also confirmed said deletion with a confirmation window.

Kewlan:
It's for if they end up regretting later of course. Why not add the functionality once possible?

src/yuzu/main.cpp
```diff
@@ -1314,14 +1317,12 @@ void GMainWindow::OnGameListOpenFolder(GameListOpenTarget target, const std::str
 }
 
 void GMainWindow::OnTransferableShaderCacheOpenFile(u64 program_id) {
-    ASSERT(program_id != 0);
```
VolcaEM, line 1317: Why did you remove this? Homebrew(s) aren't supposed to have shader caches

Morph1984: 1. The assert would cause yuzu to crash when asserts are enabled (not ignored)
2. Homebrew does not create a shader cache, so it simply would not find one and return an error message, handling it gracefully
3. I have hidden most of the context menu options in the following commit
image

slashiee:
> 
> 
> It's for if they end up regretting later of course. Why not add the functionality once possible?

If they end up regretting, then they can just reinstall the Update/DLC that they didn't want removed.

Kewlan:
> If they end up regretting, then they can just reinstall the Update/DLC that they didn't want removed.

Yes, that's why I suggested it should only move the shader cache and the config file to the trash in my previous message.

ToxaGaming:
> > If they end up regretting, then they can just reinstall the Update/DLC that they didn't want removed.
> 
> Yes, that's why I suggested it should only move the shader cache and the config file to the trash in my previous message.

The confirmation window is there to specifically make sure that you won't regret deleting or accidentally delete the shader cache/config file. Besides I'm pretty sure most users forget the contents of the trash bin and empty it whenever they want to delete something from their PC

Kewlan:
There's such an odd amount of resistance against what I suggested that I think I gotta clarify that I meant that `QFile::moveToTrash()` can be added later, not this entire PR. Is that the misunderstanding or does everyone just prefer it to delete it directly? It's true this won't matter for most people, but I want to add this extra measure for those few where it does.

FearlessTobi:
@Kewlan I think it's just that forcing a new QT version just for that is pretty needless.
As was stated before, the original files can just be dumped again and therefore I don't see any reason why this shouldn't be deleted directly when using this UI option.

PR #4377, opened 2020-07-18
qt-themes: Add Midnight Blue qdarkstyle theme (2.8.1)

Morph1984:
This adds the "Midnight Blue" and "Midnight Blue Colorful" themes to yuzu.

This is the latest version (2.8.1) of the [qdarkstyle theme](https://github.com/ColinDuquesnoy/QDarkStyleSheet). As this drastically changes the look of the UI, this was added as a separate theme instead of overwriting the previous dark theme.

image

Status bar icons (Left = Enabled, Right = Disabled)
image


ogniK5377:
For midnight blue, it might be better to make these icons white for better readability:
image

The colorful option seems fine though

Morph1984:
@ogniK5377 Should be resolved now, the icons were falling back to the default due to an incorrect icon resource path

PR #4380, opened 2020-07-19
swkbd: Return result for Calc request for inlined swkbd

ogniK5377:
Fixes random swkbd popups in monster hunter. Future PR will implement inlined swkbd when more research is done

### DIFF
--- a/src/core/hle/service/am/applets/software_keyboard.cpp
+++ b/src/core/hle/service/am/applets/software_keyboard.cpp
@@ -14,7 +14,7 @@
 namespace Service::AM::Applets {
 
 namespace {
-enum class Request : u32_le {
+enum class Request : u32 {
     Finalize = 0x4,
     SetUserWordInfo = 0x6,
     SetCustomizeDic = 0x7,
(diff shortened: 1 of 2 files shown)


src/core/hle/service/am/applets/software_keyboard.cpp
```diff
@@ -13,11 +13,23 @@
 
 namespace Service::AM::Applets {
 
+namespace {
+enum class Request : u32_le {
```
lioncash, line 17:
```suggestion
enum class Request : u32 {
```

This will break on big endian otherwise, given `u32_le` becomes a struct in that scenario.

src/core/hle/service/am/applets/software_keyboard.cpp
```diff
@@ -13,11 +13,23 @@
...
+    UnknownD = 0xd,
+    UnknownE = 0xe,
+};
+constexpr std::size_t SWKBD_INLINE_INIT_SIZE = 0x8;
 constexpr std::size_t SWKBD_OUTPUT_BUFFER_SIZE = 0x7D8;
 constexpr std::size_t SWKBD_OUTPUT_INTERACTIVE_BUFFER_SIZE = 0x7D4;
 constexpr std::size_t DEFAULT_MAX_LENGTH = 500;
 constexpr bool INTERACTIVE_STATUS_OK = false;
-
+} // namespace
```
lioncash, line 32:
```suggestion
} // Anonymous namespace
```

PR #4381, opened 2020-07-19
main: Fix Open Save/Mod Locations for installed titles

Morph1984:
Previously NAND/SDMC installed titles would open device saves when they are supposed to be user saves. This is due to the control nca not being read and thus returns 0 for both GetDefaultNormalSaveSize() and GetDeviceSaveDataSize(). Fix this by utilizing the patch manager to read the control nca.


Morph1984:
Conflicts with #4372

Morph1984:
Rebased

PR #4408, opened 2020-07-22
config: Organize config files into directories

lat9nq:
Keeps the config folder a little neater by moving config files to `config/custom/[title id].ini`. Moves existing config files from the old location to the new one if able.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -5,6 +5,7 @@
 #include <array>
 #include <QKeySequence>
 #include <QSettings>
+#include "common/common_paths.h"
 #include "common/file_util.h"
 #include "configure_input_simple.h"
 #include "core/hle/service/acc/profile_manager.h"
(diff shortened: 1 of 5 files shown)


src/yuzu/configuration/config.cpp
```diff
@@ -13,10 +14,28 @@
...
-    qt_config_loc = FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir) + config_file;
+    qt_config_loc = FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir) + "qt-config.ini";
     FileUtil::CreateFullPath(qt_config_loc);
+    qt_config =
+        std::make_unique<QSettings>(QString::fromStdString(qt_config_loc), QSettings::IniFormat);
+    Reload();
+}
+
+Config::Config(u64 title_id, bool is_global) {
+    // TODO: Ditto
```
lioncash, line 27: This should restate the same thing that needs to be done, to prevent a scenario where the first one gets resolved but the second one accidentally gets left in, and people don't know what "Ditto" is referring to.

src/yuzu/configuration/config.cpp
```diff
@@ -13,10 +14,28 @@
...
+    std::string old_filename = fmt::format(
+        "{}{:016X}.ini", FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir), title_id);
+    qt_config_loc = fmt::format("{}custom" DIR_SEP "{:016X}" DIR_SEP "qt-config.ini",
+                                FileUtil::GetUserPath(FileUtil::UserPath::ConfigDir), title_id);
+    FileUtil::CreateFullPath(qt_config_loc);
+    if (FileUtil::Exists(old_filename)) {
```
lioncash, line 33: This should likely have a TODO: to eventually remove the migration code after a certain amount of time has passed (like a year or something).

lat9nq:
Superseded by #4866

PR #4411, opened 2020-07-24
Joycon Adapter Implementation

german77:
This PR should allows us to connect directly to official joycon controllers without external software or been limited in functionality with other libraries. It's heavily based on GC adapter implementation, betterjoy and [Nintendo_Switch_Reverse_Engineering](https://github.com/dekuNukem/Nintendo_Switch_Reverse_Engineering). 

Hidapi library has problems in yuzu. Since the problem's with this library are critical it can't be used at the moment. If someone knows a different library or is able to found the issue please contact with me.

Hidapi fails to return a list of connected devices. Sometimes it only returns the first element other times returns null. Once the connection is established it works as expected.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,6 +34,9 @@
 [submodule "xbyak"]
     path = externals/xbyak
     url = https://github.com/herumi/xbyak.git
+[submodule "externals/hidapi"]
+	path = externals/hidapi
+	url = https://github.com/german77/hidapi.git
 [submodule "opus"]
 	path = externals/opus/opus
 	url = https://github.com/xiph/opus.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,9 +34,9 @@
 [submodule "xbyak"]
     path = externals/xbyak
     url = https://github.com/herumi/xbyak.git
-[submodule "externals/hidapi"]
-	path = externals/hidapi
-	url = https://github.com/german77/hidapi.git
 [submodule "opus"]
 	path = externals/opus/opus
 	url = https://github.com/xiph/opus.git
+[submodule "externals/hidapi/hidapi"]
+	path = externals/hidapi/hidapi
+	url = https://github.com/libusb/hidapi.git
--- a/src/input_common/joycon/jc_adapter.cpp
+++ b/src/input_common/joycon/jc_adapter.cpp
@@ -1,0 +1,974 @@
+// Copyright 2020 Yuzu Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include <span>
+#include <thread>
+#include "common/logging/log.h"
+#include "input_common/joycon/jc_adapter.h"
+
+namespace JCAdapter {
+
+/// Used to loop through and assign button in poller
+constexpr std::array<PadButton, 22> PadButtonArray{
+    PadButton::BUTTON_DOWN,    PadButton::BUTTON_UP,    PadButton::BUTTON_RIGHT,
+    PadButton::BUTTON_LEFT,    PadButton::BUTTON_L_SR,  PadButton::BUTTON_L_SL,
+    PadButton::TRIGGER_L,      PadButton::TRIGGER_ZL,   PadButton::BUTTON_Y,
+    PadButton::BUTTON_X,       PadButton::BUTTON_B,     PadButton::BUTTON_A,
+    PadButton::BUTTON_R_SL,    PadButton::BUTTON_R_SR,  PadButton::TRIGGER_R,
+    PadButton::TRIGGER_ZR,     PadButton::BUTTON_MINUS, PadButton::BUTTON_PLUS,
+    PadButton::STICK_R,        PadButton::STICK_L,      PadButton::BUTTON_HOME,
+    PadButton::BUTTON_CAPTURE,
+};
+
+Joycons::Joycons() {
+    LOG_INFO(Input, "JC Adapter Initialization started");
+    global_counter = 0;
+    adapter_thread_running = false;
+    const int init_res = hid_init();
+    if (init_res == 0) {
+        Setup();
+    } else {
+        LOG_ERROR(Input, "HIDapi could not be initialized. failed with error = {}", init_res);
+    }
+}
+
+void Joycons::SendWrite(hid_device* handle, std::vector<u8> buffer, int size) {
+    hid_write(handle, buffer.data(), size);
+}
+
+u8 Joycons::GetCounter() {
+    const u8 counter = global_counter;
+    global_counter = (global_counter + 1) & 0x0F;
+    return counter;
+}
+
+std::vector<u8> Joycons::SubCommand(hid_device* handle, SubComamnd sc, std::vector<u8> buffer,
+                                    int size) {
+    std::vector<u8> local_buffer(size + 11);
+
+    local_buffer[0] = static_cast<u8>(Output::RUMBLE_AND_SUBCMD);
+    local_buffer[1] = GetCounter();
+    for (int i = 0; i < 8; ++i) {
+        local_buffer[i + 2] = default_buffer[i];
+    }
+    local_buffer[10] = static_cast<u8>(sc);
+    for (int i = 0; i < size; ++i) {
+        local_buffer[11 + i] = buffer[i];
+    }
+
+    SendWrite(handle, local_buffer, size + 11);
+    return GetResponse(handle, sc);
+}
+
+std::vector<u8> Joycons::GetResponse(hid_device* handle, SubComamnd sc) {
+    int tries = 0;
+    std::vector<u8> buffer(max_resp_size);
+    do {
+        int result = hid_read_timeout(handle, buffer.data(), max_resp_size, 100);
+        if (result < 1) {
+            LOG_ERROR(Input, "No response from joystick");
+        }
+        tries++;
+    } while (tries < 10 && buffer[0] != 0x21 && buffer[14] != static_cast<u8>(sc));
+    return buffer;
+}
+
+std::vector<u8> Joycons::ReadSPI(hid_device* handle, CalAddr addr, u8 size) {
+    std::vector<u8> buffer = {0x00, 0x00, 0x00, 0x00, size};
+    std::vector<u8> local_buffer(size + 20);
+
+    buffer[0] = static_cast<u8>(static_cast<u16>(addr) & 0x00FF);
+    buffer[1] = static_cast<u8>((static_cast<u16>(addr) & 0xFF00) >> 8);
+    for (int i = 0; i < 100; ++i) {
+        local_buffer = SubCommand(handle, SubComamnd::SPI_FLASH_READ, buffer, 5);
+        if (local_buffer[15] == buffer[0] && local_buffer[16] == buffer[1]) {
+            break;
+        }
+    }
+    return std::vector<u8>(local_buffer.begin() + 20, local_buffer.end());
+}
+
+void Joycons::SetJoyStickCal(std::vector<u8> buffer, JoyStick& axis1, JoyStick& axis2, bool left) {
+    if (left) {
+        axis1.max = (u16)(((buffer[1] & 0x0F) << 8) | buffer[0]);
+        axis2.max = (u16)((buffer[2] << 4) | (buffer[1] >> 4));
+        axis1.center = (u16)(((buffer[4] & 0x0F) << 8) | buffer[3]);
+        axis2.center = (u16)((buffer[5] << 4) | (buffer[4] >> 4));
+        axis1.min = (u16)(((buffer[7] & 0x0F) << 8) | buffer[6]);
+        axis2.min = (u16)((buffer[8] << 4) | (buffer[7] >> 4));
+    } else {
+        axis1.center = (u16)(((buffer[1] & 0x0F) << 8) | buffer[0]);
+        axis2.center = (u16)((buffer[2] << 4) | (buffer[1] >> 4));
+        axis1.min = (u16)(((buffer[4] & 0x0F) << 8) | buffer[3]);
+        axis2.min = (u16)((buffer[5] << 4) | (buffer[4] >> 4));
+        axis1.max = (u16)(((buffer[7] & 0x0F) << 8) | buffer[6]);
+        axis2.max = (u16)((buffer[8] << 4) | (buffer[7] >> 4));
+    }
+
+    /* Nintendo fix for drifting stick
+    buffer = ReadSPI(0x60, (isLeft ? 0x86 : 0x98), 16);
+    joycon[0].joystick.deadzone = (u16)((buffer[4] << 8) & 0xF00 | buffer[3]);*/
+}
+
+void Joycons::SetImuCal(Joycon& jc, std::vector<u8> buffer) {
+    for (size_t i = 0; i < jc.imu.size(); ++i) {
+        jc.imu[i].acc.offset = (u16)(buffer[0 + (i * 2)] | (buffer[1 + (i * 2)] << 8));
+        jc.imu[i].acc.scale = (u16)(buffer[0 + 6 + (i * 2)] | (buffer[1 + 6 + (i * 2)] << 8));
+        jc.imu[i].gyr.offset = (u16)(buffer[0 + 12 + (i * 2)] | (buffer[1 + 12 + (i * 2)] << 8));
+        jc.imu[i].gyr.scale = (u16)(buffer[0 + 18 + (i * 2)] | (buffer[1 + 18 + (i * 2)] << 8));
+    }
+}
+
+void Joycons::GetUserCalibrationData(Joycon& jc) {
+    std::vector<u8> buffer;
+    hid_set_nonblocking(jc.handle, 0);
+
+    switch (jc.type) {
+    case JoyControllerTypes::Left:
+        buffer = ReadSPI(jc.handle, CalAddr::USER_LEFT_DATA, 9);
+        SetJoyStickCal(buffer, jc.axis[0], jc.axis[1], true);
+        break;
+    case JoyControllerTypes::Right:
+        buffer = ReadSPI(jc.handle, CalAddr::USER_RIGHT_DATA, 9);
+        SetJoyStickCal(buffer, jc.axis[2], jc.axis[3], false);
+        break;
+    case JoyControllerTypes::Pro:
+        buffer = ReadSPI(jc.handle, CalAddr::USER_LEFT_DATA, 9);
+        SetJoyStickCal(buffer, jc.axis[0], jc.axis[1], true);
+        buffer = ReadSPI(jc.handle, CalAddr::USER_RIGHT_DATA, 9);
+        SetJoyStickCal(buffer, jc.axis[2], jc.axis[3], false);
+        break;
+    }
+
+    buffer = ReadSPI(jc.handle, CalAddr::USER_IMU_DATA, 24);
+    SetImuCal(jc, buffer);
+    hid_set_nonblocking(jc.handle, 1);
+}
+
+void Joycons::GetFactoryCalibrationData(Joycon& jc) {
+    std::vector<u8> buffer;
+    hid_set_nonblocking(jc.handle, 0);
+
+    switch (jc.type) {
+    case JoyControllerTypes::Left:
+        buffer = ReadSPI(jc.handle, CalAddr::FACT_LEFT_DATA, 9);
+        SetJoyStickCal(buffer, jc.axis[0], jc.axis[1], true);
+        break;
+    case JoyControllerTypes::Right:
+        buffer = ReadSPI(jc.handle, CalAddr::FACT_RIGHT_DATA, 9);
+        SetJoyStickCal(buffer, jc.axis[2], jc.axis[3], false);
+        break;
+    case JoyControllerTypes::Pro:
+        buffer = ReadSPI(jc.handle, CalAddr::FACT_LEFT_DATA, 9);
+        SetJoyStickCal(buffer, jc.axis[0], jc.axis[1], true);
+        buffer = ReadSPI(jc.handle, CalAddr::FACT_RIGHT_DATA, 9);
+        SetJoyStickCal(buffer, jc.axis[2], jc.axis[3], false);
+        break;
+    }
+
+    buffer = ReadSPI(jc.handle, CalAddr::FACT_IMU_DATA, 24);
+    SetImuCal(jc, buffer);
+    hid_set_nonblocking(jc.handle, 1);
+}
+
+s16 Joycons::GetRawIMUValues(size_t sensor, size_t axis, std::vector<u8> buffer) {
+    const size_t offset = (sensor * 6) + (axis * 2);
+    return static_cast<s16>(buffer[13 + offset] | (buffer[14 + offset] << 8));
+}
+
+f32 Joycons::TransformAccValue(s16 raw, ImuData cal, AccSensitivity sen) {
+    // const f32 value = (raw - cal.offset) * cal.scale / 65535.0f / 1000.0f;
+    const f32 value = raw * (1.0f / (cal.scale - cal.offset)) * 4;
+    switch (sen) {
+    case AccSensitivity::G2:
+        return value / 4.0f;
+    case AccSensitivity::G4:
+        return value / 2.0f;
+    case AccSensitivity::G8:
+        return value;
+    case AccSensitivity::G16:
+        return value * 2.0f;
+    }
+    return value;
+}
+
+f32 Joycons::TransformGyrValue(s16 raw, ImuData cal, GyrSensitivity sen) {
+    // const f32 value = (raw - cal.offset) * cal.scale / 65535.0f / 360.0f / 3.8f;
+    const f32 value = (raw - cal.offset) * (936.0f / (cal.scale - cal.offset)) / 360.0f;
+    switch (sen) {
+    case GyrSensitivity::DPS250:
+        return value / 8.0f;
+    case GyrSensitivity::DPS500:
+        return value / 4.0f;
+    case GyrSensitivity::DPS1000:
+        return value / 2.0f;
+    case GyrSensitivity::DPS2000:
+        return value;
+    }
+    return value;
+}
+
+void Joycons::GetIMUValues(Joycon& jc, std::vector<u8> buffer) {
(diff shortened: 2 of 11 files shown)


bylaws:
You might want to check out the soon to be upstream linux joycon driver https://github.com/nicman23/dkms-hid-nintendo/blob/master/src/hid-nintendo.c for a more detailed reference. You should also ensure that yuzu's implementation doesn't conflict with this driver.

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+#include <thread>
+#include <unordered_map>
+#include <hidapi.h>
+#include "common/common_types.h"
+#include "common/threadsafe_queue.h"
+
+namespace JCAdapter {
+
+enum class PadButton {
+    PAD_BUTTON_LEFT = 0x0001,
```
lioncash, line 18: `ButtonLeft`, etc. `PAD_` isn't necessary anymore, given the enum is an enum class now.

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+    PadAxes axis{PadAxes::Undefined};
+    u8 axis_value{255};
+};
+
+struct JCState {
+    std::unordered_map<int, bool> buttons;
+    std::unordered_map<int, u16> axes;
+};
+
+enum class JoyControllerTypes { None, Left, Right, Pro};
```
lioncash, line 62:
```suggestion
enum class JoyControllerTypes { None, Left, Right, Pro, };
```

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+
+    std::array<JCState, 4> state;
+    std::array<bool, 4> get_origin;
+    std::array<JCPadStatus, 4> origin_status;
+    std::array<Common::SPSCQueue<JCPadStatus>, 4> pad_queue;
+    std::array<JoyControllerTypes, 4> adapter_controllers_status{};
+
+    u8 global_counter;
+    const u32 report_len = 49;
+    const u8 default_buffer[8] = {0x0, 0x1, 0x40, 0x40, 0x0, 0x1, 0x40, 0x40};
```
lioncash, line 136:
```suggestion
    static constexpr u32 report_len = 49;
    static constexpr std::array<u8, 8> default_buffer{0x0, 0x1, 0x40, 0x40, 0x0, 0x1, 0x40, 0x40};
```

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+    struct HDRumble {
+        f32 freq_high;
+        f32 freq_low;
+        f32 amp_high;
+        f32 amp_low;
+    } hdRumble;
```
lioncash, line 142:
```suggestion
    };
    HDRumble hd_rumble;
```

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+    ~Joycons();
+    /// Used for polling
+    void BeginConfiguration();
+    void EndConfiguration();
+
+    bool DeviceConnected(std::size_t port);
+
+    int GetOriginValue(int port, int axis) const;
+
+    void Set_rumble(f32 amp_high, f32 amp_low, f32 freq_high, f32 freq_low);
```
lioncash, line 79:
```suggestion
    void SetRumble(f32 amp_high, f32 amp_low, f32 freq_high, f32 freq_low);
```

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+    /// Initialize the JC Adapter capture and read sequence
+    Joycons();
+
+    /// Close the adapter read thread and release the adapter
+    ~Joycons();
+    /// Used for polling
+    void BeginConfiguration();
+    void EndConfiguration();
+
+    bool DeviceConnected(std::size_t port);
```
lioncash, line 75:
```suggestion
    bool DeviceConnected(std::size_t port) const;
```

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
+// Copyright 2014 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#pragma once
+#include <algorithm>
```
lioncash, line 6: Newline after the pragma.

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+#include <functional>
+#include <mutex>
+#include <thread>
+#include <unordered_map>
+#include <hidapi.h>
```
lioncash, line 11: This should be replaced with a forward declaration if possible and be moved to the cpp file.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+        device->vendor_id, device->product_id);
+        return false;
+    }
+    hid_set_nonblocking(joycon_handle, 1);
+    return true;
+}
+
+void Joycons::GetJCEndpoint(hid_device_info* device) {
+    adapter_thread_running = true;
+    adapter_input_thread = std::thread([=] { Read(); }); // Read input
```
lioncash, line 307:
```suggestion
    adapter_input_thread = std::thread(&Joycons::Read, this);
```

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+
+    std::array<Common::SPSCQueue<JCPadStatus>, 4>& GetPadQueue();
+    const std::array<Common::SPSCQueue<JCPadStatus>, 4>& GetPadQueue() const;
+
+    std::array<JCState, 4>& GetPadState();
+    const std::array<JCState, 4>& GetPadState() const;
+
+private:
+    void SendRumble();
+    void Send_Write(u8* buffer, int size);
```
lioncash, line 89:
```suggestion
    void SendWrite(std::span<u8> buffer);
```

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+
+    u8 get_counter();
```
lioncash, line 91:
```suggestion
    u8 GetCounter();
```

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,145 @@
...
+
+    void PadToState(const JCPadStatus& pad, JCState& state);
+
+    void Read();
+
+    /// Resets status of device connected to port
+    void ResetDeviceType(std::size_t port);
+
+    /// Returns true if we successfully gain access to JC Adapter
+    bool CheckDeviceAccess(hid_device_info* device);
```
lioncash, line 107:
```suggestion
    bool CheckDeviceAccess(hid_device_info* device) const;
```

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+void Joycons::Send_Write(u8* buffer, int size) {
+    /*for (int i = 0; i < report_len; ++i) {
+        printf("%02x ", buffer[i]);
+    }
+    printf("\n");*/
+    hid_write(joycon_handle, buffer, size);
+}
+
+u8* Joycons::SubCommand(u8 sc, u8* buffer, int size) {
+    u8* buffer_ = new u8[report_len];
```
lioncash, line 99: The way this is currently used, a memory leak will occur.

You may want to use a std::vector here, given:

```cpp
std::vector<u8> local_buffer(report_len);
```

will automatically zero out its buffer, removing the need for the loop below this.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+void Joycons::Send_Write(u8* buffer, int size) {
+    /*for (int i = 0; i < report_len; ++i) {
+        printf("%02x ", buffer[i]);
+    }
+    printf("\n");*/
+    hid_write(joycon_handle, buffer, size);
+}
+
+u8* Joycons::SubCommand(u8 sc, u8* buffer, int size) {
```
lioncash, line 98:
```suggestion
u8* Joycons::SubCommand(u8 sc, std::span<u8> buffer) {
```

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+    buffer_[10] = sc;
+    for (int i = 0; i < size; i++) {
+        buffer_[11 + i] = buffer[i];
+    }
+
+    Send_Write(buffer_, size + 11);
+    return GetResponse(sc);
+}
+
+u8* Joycons::GetResponse(u8 sc) {
```
lioncash, line 118:
```suggestion
std::vector<u8> Joycons::GetResponse(u8 sc) const {
```

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+    int tries = 0;
+    u8* buffer = new u8[report_len];
```
lioncash, line 120:
```suggestion
    std::vector<u8> buffer(report_len);
```

The for loop below this can then be removed.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+    for (int i = 0; i < report_len; ++i) {
+        buffer[i] = 0;
+    }
+    do {
+        int result = hid_read_timeout(joycon_handle, buffer, report_len, 100);
```
lioncash, line 125:
```suggestion
        const int result = hid_read_timeout(joycon_handle, buffer, report_len, 100);
```

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+        if (result < 1) {
+            LOG_ERROR(Input, "No response from joystick");
+        }
+        tries++;
+    } while (tries < 10 && buffer[0] != 0x21 && buffer[14] != sc);
+    return buffer;
+}
+
+u8 Joycons::get_counter() {
+    u8 counter = global_counter;
```
lioncash, line 135:
```suggestion
    const u8 counter = global_counter;
```

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+    for (size_t i = 0; i < pad.axis_values.size(); ++i) {
+        state.axes.insert_or_assign(static_cast<u8>(i), pad.axis_values[i]);
+    }
+}
+
+void Joycons::Read() {
+    LOG_DEBUG(Input, "JC Adapter Read() thread started");
+
+    u8* buffer = new u8[report_len];
```
lioncash, line 215:
```suggestion
    std::vector<u8> buffer(report_len);
```

The for loop below this can then be removed.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+    SubCommand(0x30, new u8[1]{leds}, 1);
+    SubCommand(0x40, new u8[1]{static_cast<u8>(imu_enabled ? 0x1 : 0x0)}, 1);
+    SubCommand(0x48, new u8[1]{0x01}, 1);
+    SubCommand(0x41, new u8[4]{0x03, 0x00, 0x00, 0x01}, 4); // higher gyro performance
+    SubCommand(0x3, new u8[1]{0x30}, 1);
+    hid_set_nonblocking(joycon_handle, 1);
+
+    while (adapter_thread_running) {
+        //printf("button update\n");
+        int status = hid_read_timeout(joycon_handle, buffer, report_len, 15);
```
lioncash, line 236:
```suggestion
        const int status = hid_read_timeout(joycon_handle, buffer, report_len, 15);
```

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+    }
+    std::array<JCPadStatus, 4> pads;
+
+    bool imu_enabled = true;
+    u8 leds = 1;
+    global_counter = 0;
+
+    //init joycon
+    hid_set_nonblocking(joycon_handle, 0);
+    SubCommand(0x30, new u8[1]{leds}, 1);
```
lioncash, line 227: All of these new calls will result in a memory leak with the current API. See other review comment about using std::span here instead.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+    }
+}
+
+void Joycons::Setup() {
+    // Initialize all controllers as unplugged
+    adapter_controllers_status.fill(JoyControllerTypes::None);
+    // Initialize all ports to store axis origin values
+    get_origin.fill(true);
+
+    struct hid_device_info *devs, *cur_dev;
```
lioncash, line 267:
```suggestion
    hid_device_info* devs = hid_enumerate(0x057e, 0x2006);
    hid_device_info* cur_dev = devs;
```

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+void Joycons::Set_rumble(f32 amp_high, f32 amp_low, f32 freq_high, f32 freq_low) {
+    hdRumble.freq_high = std::clamp(freq_high, 81.75177f, 1252.572266f);
+    hdRumble.freq_low = std::clamp(freq_low, 40.875885f, 626.286133f);
+    hdRumble.amp_high = std::clamp(amp_high, 0.0f, 1.0f);
+    hdRumble.amp_low = std::clamp(amp_low, 0.0f, 1.0f);
+    SendRumble();
+}
+void Joycons::SendRumble() {
+    u8* buffer = new u8[49];
```
lioncash, line 42:
```suggestion
    std::array<u8, 49> buffer{};
```

Will allow you to also remove the loop below this.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+        } else if (hdRumble.amp_high < 0.11249) {
+            encoded_hamp = static_cast<u8>(roundf((log(hdRumble.amp_high) * 11.556f) + 55.3f));
+        } else if (hdRumble.amp_high < 0.22498) {
+            encoded_hamp = static_cast<u8>(roundf((log2f(hdRumble.amp_high) * 32) + 131));
+        } else {
+            encoded_hamp = static_cast<u8>(roundf((log2f(hdRumble.amp_high) * 64) + 200));
+        }
+
+        // encoded_lamp should use amp_low data, for now sticking to documentation
+        u8 encoded_lamp = encoded_hamp >> 1;
```
lioncash, line 69:
```suggestion
        const u8 encoded_lamp = encoded_hamp >> 1;
```

This and most of the assignments below also need an explicit cast to `u8`, since operations on types smaller than `int` promote to `int`, so this will cause truncation warnings to occur. One solution that may be less noisy is to make the involved variables `u32`s and cast them at the end when they're being stored into the buffer.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+        u8 parity = (encoded_lamp % 2) * 0x80;
```
lioncash, line 70:
```suggestion
        const u8 parity = (encoded_lamp % 2) * 0x80;
```

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+/// Used to loop through and assign button in poller
+constexpr std::array<PadButton, 12> PadButtonArray{
+    PadButton::PAD_BUTTON_LEFT, PadButton::PAD_BUTTON_RIGHT, PadButton::PAD_BUTTON_DOWN,
+    PadButton::PAD_BUTTON_UP,   PadButton::PAD_TRIGGER_Z,    PadButton::PAD_TRIGGER_R,
+    PadButton::PAD_TRIGGER_L,   PadButton::PAD_BUTTON_A,     PadButton::PAD_BUTTON_B,
+    PadButton::PAD_BUTTON_X,    PadButton::PAD_BUTTON_Y,     PadButton::PAD_BUTTON_START,
+};
+
+Joycons::Joycons() {
+    if (joycon_handle != nullptr) {
```
lioncash, line 21: I think this branch is unnecessary, given the variable will always be nullptr when the constructor runs.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,372 @@
...
+        return;
+    }
+    LOG_INFO(Input, "JC Adapter Initialization started");
+
+    const int init_res = hid_init();
+    if (init_res ==0) {
```
lioncash, line 27:
```suggestion
    if (init_res == 0) {
```

ogniK5377:
Rebase when possible! Thanks!

.gitmodules
```diff
@@ -37,6 +37,9 @@
 [submodule "externals/libusb"]
 	path = externals/libusb
 	url = https://github.com/ameerj/libusb
+[submodule "externals/hidapi"]
+	path = externals/hidapi
+	url = https://github.com/german77/hidapi.git
```
FearlessTobi, line 42: We shouldn't have submodules pointing to private repositories.
I see it has been done before, but this doesn't mean we should continue doing so.

german77: hidapi lacks of a cmake file. I took one from dolphin. how do I use the original repository and still have the cmake file?

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,376 @@
...
+#include "common/logging/log.h"
+#include "input_common/joycon/Jc_adapter.h"
+
+namespace JCAdapter {
+
+/// Used to loop through and assign button in poller
+constexpr std::array<PadButton, 20> PadButtonArray{
+    PadButton::BUTTON_LEFT,  PadButton::BUTTON_RIGHT, PadButton::BUTTON_DOWN,
+    PadButton::BUTTON_UP,    PadButton::TRIGGER_ZR,   PadButton::TRIGGER_ZL,
+    PadButton::TRIGGER_SR,   PadButton::TRIGGER_SL,   PadButton::TRIGGER_R,
```
ameerj, line 17: I don't think the shoulder buttons on the joycons are considered triggers.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,376 @@
...
+    }
+    return counter;
+}
+
+std::vector<u8> Joycons::SubCommand(SubComamnd sc, std::vector<u8> buffer, int size) {
+    std::vector<u8> local_buffer(size + 11);
+
+    local_buffer[0] = static_cast<u8>(Output::RUMBLE_AND_SUBCMD);
+    local_buffer[1] = GetCounter();
+    for (int i = 0; i < 8; i++) {
```
ameerj, line 58: `++i` is preferred in loops

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+        buffer[2] = static_cast<u8>(encoded_hf & 0xff);
+        buffer[3] = static_cast<u8>((encoded_hf >> 8) & 0xff);
+        buffer[4] = static_cast<u8>(encoded_lf & 0xff);
+
+        buffer[3] |= static_cast<u8>(encoded_hamp);
+        buffer[4] |= static_cast<u8>(parity);
+        buffer[5] = static_cast<u8>(encoded_amp);
+
+        encoded_amp = encoded_hamp >> 1;
```
ameerj, line 186: Can you try to put this into a loop. looks a bit repetitive.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,376 @@
...
+                adapter_thread_running = false;
+                adapter_controllers_status[0] = JoyControllerTypes::None;
+                break;
+            }
+            for (int i = 0; i < 10; i++) {
+                printf("%02x ", buffer[i]);
+            }
+            printf("\n");
+            pads[0] = GetLeftPadButtons(0, buffer);
+            if (DeviceConnected(0)) {
```
ameerj, line 255: `pad_queue` need only be updated/pushed into while polling/configuring

src/input_common/joycon/jc_poller.cpp
```diff
@@ -0,0 +1,165 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
```
ameerj, line 4: Since this looks almost exactly like the gc_poller. It might be worthwhile in the future to abstract this out to reduce the code duplication.

src/input_common/joycon/jc_adapter.h
```diff
@@ -0,0 +1,216 @@
...
+#include <thread>
+#include <unordered_map>
+#include <hidapi.h>
+#include "common/common_types.h"
+#include "common/threadsafe_queue.h"
+
+namespace JCAdapter {
+
+enum class PadButton {
+    BUTTON_LEFT = 0x08,
```
ameerj, line 20: Not sure if this is still under development. But sort this in increasing order for the value.

And if these are OR'd together, then you definitely don't want duplicate values.

.gitmodules
lioncash: Normally we place the original repo into a folder within a folder in externals, and then place the CMake file outside of it like so.

e.g. In other words:

1. Make a folder in externals called hidapi. Place the CMakeLists file here.
2. Make another folder in the hidapi folder also called hidapi. This is where the source from the original repo will go.

src/input_common/joycon/jc_adapter.cpp
```diff
@@ -0,0 +1,775 @@
...
+        joycon[port].hd_rumble.freq_high = 160.0f;
+        joycon[port].hd_rumble.freq_low = 80.0f;
+    }
+    int port = 0;
+
+    hid_device_info* devs = hid_enumerate(0x057e, 0x2006);
+    hid_device_info* cur_dev = devs;
+
+    while (cur_dev) {
+        printf("Device Found\n  type: %04hx %04hx\n  path: %s\n  serial_number: %ls",
```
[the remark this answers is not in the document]
german77: This part it's where I have problems with the library. All logs and duplicated code in this function will be removed once it works correctly. For now this will help with debugging.

cooperate:
What is the current status of this PR?  Is there anything I can do to help test?

german77:
This PR is in standby mode until motion and rumble are on mainline. If you want to test it contact me on discord "naii the baf".

v1993:
Now that both motion and rumble are in mainline, will this PR be revived?

I've tried testing it some time ago, but it failed to pull required submodule and I wasn't able to advance past CMake step.

german77:
It has been updated to make cmake work. I will fix the merge conflict today. There are machines where even a compiled exe does not work for unknow reasons. 

Better joy for example can work at the same time with the native joycon. The communication it's basically UDP it does not care who is the sender, It should work with hid_nintendo without a problem.

v1993:
BetterJoy is a windows utility and hid_nintendo is linux driver, so it isn't applicable here. In addition, Cemuhook protocol have numerous issues that make me avoid it until it's replaced with better one.

I've also tested on last commit present here and it failed the way I described (missing submodule). Running `git submodule update --init` haven't changed anything.

v1993:
Note: currently this fails to build for me when merged into master (merge conflict fixed by hands) with following error:

```
FAILED: externals/hidapi/CMakeFiles/hidapi.dir/hidapi/libusb/hid.c.o
gcc-10 -DARCHITECTURE_x86_64=1 -DBOOST_ASIO_DISABLE_CONCEPTS -I/home/v/.conan/data/boost/1.73.0/_/_/package/104297d4d8bdbca7e9351b0f32c32f11628164c7/include -I/home/v/.conan/data/catch2/2.13.0/_/_/package/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9/include -I/home/v/.conan/data/fmt/7.0.3/_/_/package/b173bbda18164d49a449ffadc1c9e817f49e819d/include -I/home/v/.conan/data/nlohmann_json/3.8.0/_/_/package/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9/include -I/home/v/.conan/data/zlib/1.2.11/_/_/package/19729b9559f3ae196cad45cb2b97468ccb75dcd1/include -I/home/v/.conan/data/bzip2/1.0.8/_/_/package/91a8b22c2c5a149bc617cfc06cdd21bf23b12567/include -I../externals/hidapi/hidapi/hidapi    -O3 -DNDEBUG  -flto -fno-fat-lto-objects   -D_GLIBCXX_USE_CXX11_ABI=1 -MD -MT externals/hidapi/CMakeFiles/hidapi.dir/hidapi/libusb/hid.c.o -MF externals/hidapi/CMakeFiles/hidapi.dir/hidapi/libusb/hid.c.o.d -o externals/hidapi/CMakeFiles/hidapi.dir/hidapi/libusb/hid.c.o   -c ../externals/hidapi/hidapi/libusb/hid.c
../externals/hidapi/hidapi/libusb/hid.c:47:10: fatal error: libusb.h: No such file or directory
   47 | #include <libusb.h>
      |          ^~~~~~~~~~
compilation terminated.
```

Correct path to lost header is `externals/libusb/libusb/libusb/libusb.h` FYI.

v1993:
@german77 You say that HIDAPI library have critical issues. It perhaps would be helpful to know what are those issues in order to not run into them with other libraries.

cooperate:
> @german77 You say that HIDAPI library have critical issues. It perhaps would be helpful to know what are those issues in order to not run into them with other libraries.

If I remember correctly while testing this, iterating over devices does not work correctly - hence why german77 had to duplicate his code over and over again.  Something is wrong with the libraries implementation in the project.

RinMaru:
this still open for testing?

german77:
I have been actively trying to make it work. The main show stopper here is that hidapi conflicts with hidapi from sdl2. I found that if I rename all functions to have unique names it will work wonderfully

german77:
This PR is very outdated, not worth keeping. A new one will be opened when it works

Issue #4424, opened 2020-07-25
callers of NVFlinger::Lock() do not hold onto the unique_lock


comex:
`NVFlinger::Lock` returns a `unique_lock`, a lock guard which will automatically unlock the mutex when it is destroyed.

Each its three callers (two in vi.cpp and one in nvflinger.cpp) simply calls `Lock();`, discarding the return value.  This causes the mutex to be locked and then immediately unlocked.  I am pretty sure this is not intended.

Instead, they should probably do something like `auto guard = Lock();`, in order to hold onto the mutex for the rest of the scope.

lioncash:
Resolved in #4426. Thanks for the report!

PR #4426, opened 2020-07-25
nvflinger: Use return value of Lock()

lioncash:
comex reported in #4424 that we were incorrectly discarding the return value of Lock() which is correct.

While we're at it, we can make it more unlikely for issues like this to happen in the future by marking parts of the interface with the [[[nodiscard]]](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) attribute.

### DIFF
--- a/src/core/hle/service/nvflinger/nvflinger.h
+++ b/src/core/hle/service/nvflinger/nvflinger.h
@@ -86,7 +86,7 @@ public:
 
     s64 GetNextTicks() const;
 
-    std::unique_lock<std::mutex> Lock() {
+    std::unique_lock<std::mutex> Lock() const {
         return std::unique_lock{*guard};
     }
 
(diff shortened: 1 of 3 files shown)


src/core/hle/service/nvflinger/nvflinger.h
```diff
@@ -54,54 +54,52 @@ class NVFlinger final {
...
 
-    std::unique_lock<std::mutex> Lock() {
-        return std::unique_lock{*guard};
-    }
+    [[nodiscard]] std::unique_lock<std::mutex> Lock() const { return std::unique_lock{*guard}; }
 
-private:
-    /// Finds the display identified by the specified ID.
-    VI::Display* FindDisplay(u64 display_id);
+    private :
```
lioncash, line 91: If there's a particular configuration we can use in our clang format file to prevent clang-format wanting extremely wonky formatting like this in the presence of C++ attributes, I'd love to know about it.

PR #4430, opened 2020-07-26
hle: nvdrv: Rewrite of GPU memory management.

bunnei:
Rewrites our GPU Virtual Memory Manager. This greatly simplifies things, should reduce committed memory by the process, fixes many bugs, and allows us to implement a few new features. Among these, we implement the "remap" flag of the MapBufferEx ioctl. This fixes many of the graphical bugs with Mario Kart 8 Deluxe, as well as some of the softlock issues with Dragon Quest XI.

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvmap.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvmap.cpp
@@ -18,7 +18,12 @@ enum {
 };
 }
 
-nvmap::nvmap(Core::System& system) : nvdevice(system) {}
+nvmap::nvmap(Core::System& system) : nvdevice(system) {
+    // Handle 0 appears to be used when remapping, so we create a placeholder empty nvmap object to
+    // represent this.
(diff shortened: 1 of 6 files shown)


src/core/hle/service/nvdrv/devices/nvmap.cpp
```diff
@@ -50,6 +55,21 @@ u32 nvmap::ioctl(Ioctl command, const std::vector<u8>& input, const std::vector<
...
+    // Create a new nvmap object and obtain a handle to it.
+    auto object = std::make_shared<Object>();
+    object->id = next_id++;
+    object->size = size;
+    object->status = Object::Status::Created;
+    object->refcount = 1;
+
+    const u32 handle = next_handle++;
+
+    handles[handle] = std::move(object);
```
lioncash, line 68:
```suggestion
    handles.insert_or_assign(handle, std::move(object));
```

PR #4432, opened 2020-07-26
video_core/gpu: Correct the size of the puller registers

bylaws:
The puller register array is made up of u32s however the `NUM_REGS` value is the size in bytes, so switch it to avoid making the struct unnecessary large. Also fix a small typo in a comment.


ReinUsesLisp:
I forgot to write it: lgtm, thanks for looking into this

PR #4445, opened 2020-07-28
renderer_opengl: Use 1/4 of all threads for async shader compilation

Morph1984:
This bumps up the maximum threads for asynchronous shader compilation to 4 threads, and ensures that no more than 25% of all threads are used.

The number of shader threads spawned are as follows:
```
CPU Threads  Shader Threads
       <= 7               1
       8-11               2
      12-15               3
      >= 16               4
```


goldenx86:
Just look at CPU use and estimate the number of free threads, use all of them and make those power bills go brrrrr

ameerj:
Just a thought. Could we give the user a choice for how many threads they want to use for async compilation? From 1 thread to 
~1/3 of the total cpu threads, with it defaulting to 1/4?

PR #4448, opened 2020-07-29
game_list_worker: Do not clear entries when > 1 gamedir is present

Morph1984:
Previously the map of entries was being cleared while looping through each game directory, this resulted into all game directories except the last game dir to lose content metadata information. Fix this by clearing the entries only once.


Morph1984:
@CrazyMax This is why patch manager isn't able to find the control data and unable to display the control title version, it is now fixed.

PR #4451, opened 2020-07-29
logging/settings: Increase maximum log size to 100 MB and add extended logging option

slashiee:
This PR increases the maximum size the log can go up to from 50 MB to 100 MB.

The extended logging option is automatically disabled on boot but can be enabled afterwards, allowing the log file to go up to 1 GB during that session.

This PR also fixes a few errors that are present in the general debug menu, such as tool tips not appearing.

### DIFF
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -23,6 +23,7 @@
 #include "common/logging/text_formatter.h"
 #include "common/string_util.h"
 #include "common/threadsafe_queue.h"
+#include "core/settings.h"
 
 namespace Log {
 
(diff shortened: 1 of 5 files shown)


Morph1984:
> Instead of arbitrarily increasing the limit to 100MB, shouldn't we be looking into what is causing the log spam and find ways to work around it?
> Also why is there a limit when enabling "extended logging"? Isn't the purpose of this feature to remove the issue of the log getting cut off because of its size?

It is a safeguard to prevent drive space from being filled up by the usual spam suspects like unknown npad id, unmapped reads, etc

Extended logging already provides 20x more space for logs than what is currently available, but can be increased if you'd like

slashiee:
> Instead of arbitrarily increasing the limit to 100MB, shouldn't we be looking into what is causing the log spam and find ways to work around it?

Emulator progression is probably the best way to go towards the log spam, as there isn't really any good way to work around it. This is especially true if they are errors, which a lot of them are.

Clamping duplicate entries together and noting how many times it appears like RPCS3 could work, but if there's something in-between each duplicate (which happens a lot of the time) then it won't work. Compressing the log is also a good idea, but I currently do not have enough coding knowledge to attempt that.

If you want me to I can revert the default size back to 50 MB. I only increased it because it was suggested by a few people.

> Also why is there a limit when enabling "extended logging"? Isn't the purpose of this feature to remove the issue of the log getting cut off because of its size?

The reason why there's still a limit with extended logging enabled is because otherwise stuff like unmapped memory errors will cause the log to fill up the hard drive rather quickly. Capping it at 1GB will ensure the log won't become 10+ GB filled with these errors while still logging everything important that's needed past whatever the default size provides.

Morph1984 (approved):
I have no objections to this PR, we can't have these just lying around

bunnei (approved):

Issue #4132, opened 2020-06-20
High input lag/latency


Jatoxo:
Input latency in yuzu is very high for me. Both with keyboard or with a controller, there is a very noticeable delay. This doesn't just apply to keypresses, analog sticks and the touch feature is affected as well. The video I am attaching is in real time, the delay is clearly visible. I'll also include a generic log. The issue seems persistent regardless of version or game playing. Testing over on Ryujinx input latency is much lower

Video:
![ezgif-4-40ca74412c5b](https://user-images.githubusercontent.com/39669932/85211608-0372dd80-b34b-11ea-97f9-325244d5846b.gif)

Log:
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/4808560/yuzu_log.txt)

lat9nq:
I've noticed the same issue. In both yuzu and yuzu-cmd, there's is a noticeable delay between my input to the gamepad and the response in yuzu. The issue was **not** observed in other applications, including of course the HTML5 Gamepad Tester as well as Citra.

[Video](https://streamable.com/u8qz9q), [Gamepad Tester](https://html5gamepad.com/), [yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/4809036/yuzu_log.txt)

My specs are:
Manjaro 20.0.3, Linux 5.7.2-6-tkg-pds
AMD Ryzen 7 2700X
GTX 980 Ti on 440.82
16GB 3200CL16 @ 3400 MHz
Controller is, in this case, a HORI NSW-108, but identical latency also observed on a HORI NSW-001, a PowerA 1507843-01, and a Sony DS3 CECHZC2U over USB.

Given that gamepad support in Linux is baked into the kernel, this is likely not a driver issue.

FearlessTobi:
Our polling rate is probably wrong, causing this delay to happen. 

(https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/service/hid/hid.cpp#L42)

resadent:
I guess this issue has to do with input lag too: https://github.com/yuzu-emu/yuzu/issues/4053. Hope yuzu gains exclusive fullscreen too in the future :).

Pajn:
@resadent This has nothing to do with fullscreen. On linux there isn't even a concept of "exclusive fullscreen"

Morph1984:
Some things I found:
[...]

PR #4452, opened 2020-07-30
HID: Make all input shared memory the same

german77:
As explained by #4132, yuzu had minor problems with input lag. From my testing I had 1 to 4 frames of lag. After changing the sleep period to 5ms now I have 0 to 2 frames of lag. 
5ms is the period that the acelerometer and gyroscope are polled from a joycon.
After more testing I found that this does not solve the problem. Just make it less likely to be delayed by many frames.

Solution:
The input memory consist of an array of 17 values. An unsync input poller is equivalent of reading an input value before or after the actual updated value. That means you read a random value of that input array. In the worst case scenario you lag 17 frames.

A game with 1200fps has near 0ms of input lag because it reads the whole input array in one screen refresh.

By making the full input array contain only the latest value. It will ensure that in the next frame the correct data will be read. This will fix problems with input lag while I found a way to sync the function with the emulator.

Needs to be tested as it has potential to break some games.

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -115,7 +115,10 @@ void IAppletResource::UpdateControllers(u64 userdata, s64 ns_late) {
         if (should_reload) {
             controller->OnLoadInputDevices();
         }
-        controller->OnUpdate(core_timing, shared_mem->GetPointer(), SHARED_MEMORY_SIZE);
+        // TODO(german77): Sync this function with the emulator to remove 'loop'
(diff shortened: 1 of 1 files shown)


german77:
I think this problem has to be fixed from the root. Closing it until I found a solution that keeps the emulation accuracy.

FearlessTobi:
Arbitrarily putting a loop here seems more like a hack than a proper solution.

ogniK5377:
In favor of using #4643 over this one specifically because #4643 actually changes the rate a which input is polled whereas this one is just updating 17 times every poll which in turn can break input in many games

PR #4453, opened 2020-07-30
textures/decoders: Fix block linear to pitch copies

ReinUsesLisp:
There were two issues with block linear copies. First the swizzling was wrong and this commit reimplements it.

The other issue was that these copies are generally used to download render targets from the GPU and yuzu was not downloading them from host GPU memory unless the extreme GPU accuracy setting was selected. This commit enables cached memory reads for all accuracy levels.

- Fixes level thumbnails in Super Mario Maker 2.


ReinUsesLisp:
Before:
imagen
After:
imagen

asLody:
This has been tested to cause a huge performance regression.

BSoDGamingYT:
> This has been tested to cause a huge performance regression.

Noting the games affected would be helpful

PR #4457, opened 2020-07-31
am: Unstub SetScreenShotPermission

ogniK5377:


theboy181:
Will this allow the use of taking screenshots like on the HW, or is this for some compatibility requirement for a specific game?

PR #4461, opened 2020-08-02
Fix thread naming on Linux, which limits names to 15 bytes.

comex:
- In `SetCurrentThreadName`, when on Linux, truncate to 15 bytes, as (at least on glibc) `pthread_set_name_np` will otherwise return `ERANGE` and do nothing.
- Also, add logging in case `pthread_set_name_np` returns an error anyway.  This is Linux-specific, as the Apple and BSD versions of `pthread_set_name_np` return `void`.
- To make things slightly more succinct, change `SetCurrentThreadName` to take `const std::string&` rather than `const char *`.
- Change the name for CPU threads in multi-core mode from "yuzu:CoreCPUThread_N" (19 bytes) to "yuzu:CPUCore_N" (14 bytes) so it fits into the Linux limit.  Some other thread names are also cut off, but I didn't bother addressing them as you can guess them from the truncated versions.  For a CPU thread, truncation means you can't see which core it is!

### DIFF
--- a/src/common/thread.cpp
+++ b/src/common/thread.cpp
@@ -2,6 +2,8 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include "common/common_funcs.h"
+#include "common/logging/log.h"
 #include "common/thread.h"
 #ifdef __APPLE__
 #include <mach/mach.h>
@@ -18,6 +20,8 @@
 #ifndef _WIN32
 #include <unistd.h>
 #endif
+
+#include <string>
 
 #ifdef __FreeBSD__
 #define cpu_set_t cpuset_t
@@ -110,6 +114,14 @@ void SetCurrentThreadName(const char* name) {
     pthread_set_name_np(pthread_self(), name);
 #elif defined(__NetBSD__)
     pthread_setname_np(pthread_self(), "%s", (void*)name);
+#elif defined(__linux__)
+    // Linux limits thread names to 15 characters and will outright reject any
+    // attempt to set a longer name with ERANGE.
+    std::string truncated(name, std::min(strlen(name), static_cast<size_t>(15)));
+    if (int e = pthread_setname_np(pthread_self(), truncated.c_str())) {
+        errno = e;
+        LOG_ERROR(Common, "Failed to set thread name to '{}': {}", truncated, GetLastErrorMsg());
+    }
 #else
     pthread_setname_np(pthread_self(), name);
 #endif
--- a/src/core/cpu_manager.cpp
+++ b/src/core/cpu_manager.cpp
@@ -331,7 +331,7 @@ void CpuManager::RunThread(std::size_t core) {
     system.RegisterCoreThread(core);
     std::string name;
     if (is_multicore) {
-        name = "yuzu:CoreCPUThread_" + std::to_string(core);
+        name = "yuzu:CPUCore_" + std::to_string(core);
     } else {
         name = "yuzu:CPUThread";
     }


src/common/thread.cpp
```diff
@@ -76,7 +78,7 @@ void SetCurrentThreadPriority(ThreadPriority new_priority) {
 // Sets the debugger-visible name of the current thread.
 // Uses trick documented in:
 // https://docs.microsoft.com/en-us/visualstudio/debugger/how-to-set-a-thread-name-in-native-code
-void SetCurrentThreadName(const char* name) {
+void SetCurrentThreadName(const std::string& name) {
```
ogniK5377, line 81:
```suggestion
void SetCurrentThreadName(std::string_view name) {
```

src/common/thread.cpp
```diff
@@ -103,13 +105,21 @@ void SetCurrentThreadName(const char* name) {
 
 // MinGW with the POSIX threading model does not support pthread_setname_np
 #if !defined(_WIN32) || defined(_MSC_VER)
-void SetCurrentThreadName(const char* name) {
+void SetCurrentThreadName(const std::string& name) {
```
ogniK5377, line 108:
```suggestion
void SetCurrentThreadName(std::string_view name) {
```

src/common/thread.h
```diff
@@ -95,6 +95,6 @@ enum class ThreadPriority : u32 {
 
 void SetCurrentThreadPriority(ThreadPriority new_priority);
 
-void SetCurrentThreadName(const char* name);
+void SetCurrentThreadName(const std::string& name);
```
ogniK5377, line 98:
```suggestion
void SetCurrentThreadName(std::string_view name);
```

src/common/thread.cpp
comex: Ah, right, I forgot that `std::string_view` was in C++17 rather than only being in C++20 like `std::span`.

comex:
Updated.

src/common/thread.cpp
```diff
@@ -103,21 +107,29 @@ void SetCurrentThreadName(const char* name) {
 
 // MinGW with the POSIX threading model does not support pthread_setname_np
 #if !defined(_WIN32) || defined(_MSC_VER)
-void SetCurrentThreadName(const char* name) {
+void SetCurrentThreadName(std::string_view name) {
```
lioncash, line 110: Fairly sure using a std::string_view over std::string here is problematic, given std::string_view doesn't guarantee null-termination, which is generally what these APIs expect on other platforms.

comex: Oh, jeez, I’m embarrassed.  How does this even compile on Windows?  There shouldn’t be a `c_str` method on `string_view` for precisely that reason.

bunnei:
This currently breaks our build system. Here is the error:

```
##[error]src\common\thread.cpp(96,29): Error C2039: 'c_str': is not a member of 'std::basic_string_view<char,std::char_traits<char>>'D:\a\1\s\src\common\string_util.cpp(157,31): warning C4996: 'std::codecvt_utf8_utf16<short,1114111,(std::codecvt_mode)0>': warning STL4017: std::wbuffer_convert, std::wstring_convert, and the <codecvt> header (containing std::codecvt_mode, std::codecvt_utf8, std::codecvt_utf16, and std::codecvt_utf8_utf16) are deprecated in C++17. (The std::codecvt class template is NOT deprecated.) The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from <Windows.h> instead. You can define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.
    34>D:\a\1\s\src\common\thread.cpp(96,29): error C2039: 'c_str': is not a member of 'std::basic_string_view<char,std::char_traits<char>>'D:\a\1\s\src\common\string_util.cpp(157,31): warning C4996: 'std::codecvt_utf8_utf16<short,1114111,(std::codecvt_mode)0>': warning STL4017: std::wbuffer_convert, std::wstring_convert, and the <codecvt> header (containing std::codecvt_mode, std::codecvt_utf8, std::codecvt_utf16, and std::codecvt_utf8_utf16) are deprecated in C++17. (The std::codecvt class template is NOT deprecated.) The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from <Windows.h> instead. You can define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning. [d:\a\1\s\build\src\common\common.vcxproj]
    41>ClCompile:
         platform_util.c
```

comex:
Yep, the patch is wrong, as @lioncash identified before.  What I'm confused about is why the Windows check builds succeeded in the first place.  Anyway, I'll fix it.

comex:
Updated to go back to the original function signature.  Sorry for screwing up the patch.

comex:
Ping.

FearlessTobi (approved):

PR #4465, opened 2020-08-03
ipc: Allow all trivially copyable objects to be passed directly into WriteBuffer

ogniK5377:
With the support of C++20, we can use concepts to deduce if a type is an STL container or not.

### DIFF
--- a/src/common/concepts.h
+++ b/src/common/concepts.h
@@ -4,15 +4,17 @@
 
 #pragma once
 
-#include <concepts>
-
 // Check if type is like an STL container
 template <typename T>
 concept IsSTLContainer = requires(T t) {
     typename T::value_type;
+    typename T::iterator;
+    typename T::const_iterator;
     // TODO(ogniK): Replace below is std::same_as<void> when MSVC supports it.
     t.begin();
     t.end();
+    t.cbegin();
+    t.cend();
     t.data();
     t.size();
 };
--- a/src/common/concepts.h
+++ b/src/common/concepts.h
@@ -1,0 +1,18 @@
+// Copyright 2020 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <concepts>
+
+// Check if type is like an STL container
+template <typename T>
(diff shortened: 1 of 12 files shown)


src/common/concepts.h
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <concepts>
```
lioncash, line 7: This currently doesn't need to be included, given it uses nothing from it yet.

src/common/concepts.h
```diff
@@ -0,0 +1,18 @@
...
+
+// Check if type is like an STL container
+template <typename T>
+concept IsSTLContainer = requires(T t) {
+    typename T::value_type;
```
lioncash, line 12:
```cpp
typename T::iterator;
typename T::const_iterator;
```

should also be added.

src/common/concepts.h
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
```
lioncash, line 6: This file should contain the Common namespace.

src/core/hle/kernel/hle_ipc.h
```diff
@@ -199,17 +200,18 @@ class HLERequestContext {
...
-    std::size_t WriteBuffer(const ContiguousContainer& container,
-                            std::size_t buffer_index = 0) const {
-        using ContiguousType = typename ContiguousContainer::value_type;
-
-        static_assert(std::is_trivially_copyable_v<ContiguousType>,
-                      "Container to WriteBuffer must contain trivially copyable objects");
-
-        return WriteBuffer(std::data(container), std::size(container) * sizeof(ContiguousType),
-                           buffer_index);
+    template <typename T, typename = std::enable_if_t<!std::is_pointer_v<T>>>
```
lioncash, line 203: `@tparam` in the documentation needs to be updated.

PR #4466, opened 2020-08-03
loader: Make IdentifyFile typesafe

ogniK5377:
Relies on #4465 for concept.h Common::IsBaseOf

Thanks to std::is_base_of_v we can ensure that everything passed is actually derived from "AppLoader" and not rely on preprocessor macros now.

### DIFF
--- a/src/core/loader/loader.cpp
+++ b/src/core/loader/loader.cpp
@@ -23,6 +23,8 @@
 
 namespace Loader {
 
+namespace {
+
 template <Common::IsBaseOf<AppLoader> T>
 std::optional<FileType> IdentifyFileLoader(FileSys::VirtualFile file) {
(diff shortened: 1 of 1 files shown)


src/core/loader/loader.cpp
```diff
@@ -21,27 +23,37 @@
...
-    CHECK_TYPE(NSO)
-    CHECK_TYPE(NRO)
-    CHECK_TYPE(NCA)
-    CHECK_TYPE(XCI)
-    CHECK_TYPE(NAX)
-    CHECK_TYPE(NSP)
-    CHECK_TYPE(KIP)
-
-#undef CHECK_TYPE
+template <Common::IsBaseOf<AppLoader> T>
```
lioncash, line 26: This template function should be placed within an anonymous namespace.

PR #4472, opened 2020-08-03
perf_stats: Mark GetMeanFrametime() as const 

lioncash:
The general pattern is to mark mutexes as mutable when it comes to matters of constness, given the mutex acts as a transient member of a data structure: it's not a direct invariant of the data structure.

While we're at it, we can make use of designated initializers to directly initialize structs instead of zeroing them out beforehand.

### DIFF
--- a/src/core/perf_stats.h
+++ b/src/core/perf_stats.h
@@ -30,7 +30,6 @@ struct PerfStatsResults {
 class PerfStats {
 public:
     explicit PerfStats(u64 title_id);
-
     ~PerfStats();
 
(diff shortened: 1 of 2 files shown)


src/core/perf_stats.h
```diff
@@ -44,7 +43,7 @@ class PerfStats {
     /**
      * Returns the Arthimetic Mean of all frametime values stored in the performance history.
```
Morph1984, line 44: While we're at it we can also correct this to `Arithmetic`

PR #4484, opened 2020-08-03
aes_util: Allow SetIV() to be non-allocating

lioncash:
In a few places, the data to be set as the IV is already within an array. We shouldn't require this data to be heap-allocated if it doesn't need to be.

### DIFF
--- a/src/core/crypto/aes_util.h
+++ b/src/core/crypto/aes_util.h
@@ -6,7 +6,6 @@
 
 #include <memory>
 #include <type_traits>
-#include <vector>
 #include "common/common_types.h"
 #include "core/file_sys/vfs.h"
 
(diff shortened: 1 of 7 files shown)


src/core/crypto/aes_util.h
```diff
@@ -32,10 +31,12 @@ class AESCipher {
 
 public:
     AESCipher(Key key, Mode mode);
-
     ~AESCipher();
 
-    void SetIV(std::vector<u8> iv);
+    template <typename ContiguousContainer>
```
ogniK5377, line 36:
```suggestion
    template <Common::IsSTLContainer ContiguousContainer>
```

lioncash: This is overly restrictive. Any type can be used here as long as they provide a data and size member function.

lioncash: ContiguousContainer is a more adequate container here, since the data *needs* to be a contiguous block of memory. Once the `<concept>` header is available on all platforms, then that concept can be used.

ogniK5377: it might be worth adding a concept for it specifically since the compiler emits nicer errors if they don't match

PR #3666, opened 2020-04-15
Implement a new virtual memory manager

bunnei:
This rewrites our process virtual memory manager (what used to be `hle/kernel/vm_manager.cpp` and `hle/kernel/vm_manager.h`). The goals of this rewrite are:
* Reorganize the design, to be closer to the real HOS, and use similar primitives for memory management.
* Manage the process memory layout more accurately, e.g. to emulate games such as Super Smash Bros. Ultimate without the memory coalescing workaround.
* Functionality with an eventual multicore impl. of yuzu's kernel
* Back the VMM by a single block of memory representing the Switch's physical RAM, which we then can reserve using OS virtual mapping functions to reduce unnecessary memory usage. This furthermore avoids the endless allocations that the old VMM could do, which resulted in runaway memory usage.
* Enable a correct implementation of SetHeapSize, which our old memory manager did not support. This is used by Donkey Kong Tropical Freeze to resize the heap, which did not work correctly with the old VMM.
* Simplify several other kernel primitives that rely on the VMM (e.g. SharedMemory, TransferMemory, etc. impls. are quite a bit simpler now).

Several major components of this are taken (with modifications) from Mesopshere/Atmosphere-NX. I've annotated these files with a copyright comment. Credit also to @gdkchan as the design here is somewhat similar to Ryujinx, and a few ideas were directly borrowed (the main things that come to mind here is the physical memory layout, which is much simpler than Mesopshere).

This has been in EA for several days now, and for the most part is working pretty well. This does seem to be causing a regression on Linux (https://github.com/yuzu-emu/yuzu/issues/3639) which needs to be investigated before merging.

I also ended up removing a few things that we have not been able to adequately test yet, primarily because I can't find games that use them (a few SVCs). If these are reported, I will try my best to re-implement these before this is merged.

PR #4493, opened 2020-08-05
common/virtual_buffer: drop unused includes

jbeich:
Regressed by #3666. Similar to #4208. From [DragonFly build log](https://sting.dragonflybsd.org/dports/logs/emulators___yuzu-qt5.log):
```
src/common/virtual_buffer.cpp
src/common/virtual_buffer.cpp:16:10: fatal error: sys/sysinfo.h: No such file or directory
 #include <sys/sysinfo.h>
          ^~~~~~~~~~~~~~~
```

### DIFF
--- a/src/common/virtual_buffer.cpp
+++ b/src/common/virtual_buffer.cpp
@@ -5,16 +5,7 @@
 #ifdef _WIN32
 #include <windows.h>
 #else
-#include <stdio.h>
 #include <sys/mman.h>
-#include <sys/types.h>
-#if defined __APPLE__ || defined __FreeBSD__ || defined __OpenBSD__
-#include <sys/sysctl.h>
-#elif defined __HAIKU__
-#include <OS.h>
-#else
-#include <sys/sysinfo.h>
-#endif
 #endif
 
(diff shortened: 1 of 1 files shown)


src/common/virtual_buffer.cpp
```diff
@@ -5,16 +5,7 @@
 #ifdef _WIN32
 #include <windows.h>
 #else
-#include <stdio.h>
 #include <sys/mman.h>
-#include <sys/types.h>
-#if defined __APPLE__ || defined __FreeBSD__ || defined __OpenBSD__
```
ameerj, line 11: You might want to keep 
```c++
#if defined __APPLE__
```
Only remove the directive that impacts you directly.

jbeich: Builds fine on FreeBSD, so `#include <sys/sysctl.h>` is confurmed to be unused.

jbeich: Also, Yuzu (unlike Dolphin) never supported Haiku.

bunnei:
LGTM, thank you!

PR #4510, opened 2020-08-11
uuid: Directly compare UUID instead of checking per element

ogniK5377:
We can now update this for C++20


lioncash:
Commit message title needs to be updated to match the PR.

Our compilers on the buildbot need to be updated (if possible) in order for compilation to succeed.

Morph1984:
Superseded by #6684

PR #4512, opened 2020-08-11
video_core: Use contains when possible

ogniK5377:


lioncash:
clang/libc++ currently don't implement this yet (see the compiler [support](https://en.cppreference.com/w/cpp/compiler_support) page)

ogniK5377:
Will keep it open until support is added

v1993:
Looks like it's supported now.

ameerj:
These files no longer exist as of the shader decompiler rewrite. Do you want to keep this open @ogniK5377 ?

ameerj:
Closing since the files being changed no longer exist in the project.

PR #4513, opened 2020-08-11
arb_decompiler: Use bit_cast for FClamp

ogniK5377:


lioncash:
This is not supported by either clang or GCC at the moment (see the [compiler support](https://en.cppreference.com/w/cpp/compiler_support) page)

v1993:
GCC libstdc++ 11 implements this.

PR #4515, opened 2020-08-11
main: Add an option to modify the currrent game's configuration

lat9nq:
Creates a new entry in the Emulation menu labelled "Configure Current Game..." that is only enabled if a game is currently being run in yuzu. When selected, it opens the game properties dialog for the current game.

Thanks to BSoD for reminding me to do this.


bunnei:
LGTM - Thanks!

PR #4576, opened 2020-08-24
video_core: Make use of defaulted three-way comparison where applicable

lioncash:
Allows the compiler to generate the comparison operators in a much more succinct manner.


lioncash:
Seems like the Windows buildbot doesn't support this yet :|

bunnei:
@lioncash Should this be closed then?

ameerj:
Many of these files no longer exist given the various rewrites and changes over the last year. Is this something you want to keep opened @lioncash ?

ameerj:
Closing due to the above.

PR #4584, opened 2020-08-26
externals: Track upstream libusb directly

lioncash:
Allows us to directly track and manage how we integrate libusb into yuzu without needing to use a fork as an intermediary to bridge the gap.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,9 +34,6 @@
 [submodule "xbyak"]
     path = externals/xbyak
     url = https://github.com/herumi/xbyak.git
-[submodule "externals/libusb"]
-	path = externals/libusb
-	url = https://github.com/ameerj/libusb
 [submodule "opus"]
 	path = externals/opus/opus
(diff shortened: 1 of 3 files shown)


ameerj:
Downloaded the artifact and mapping the GC controls worked just as well as previously. Thank you for this!

bunnei:
EA tagging for a day or so, just to be safe

bunnei:
Untagging because we are hitting `error: refusing to lose untracked file at 'externals/libusb/config.h.in'` on build pipeline -- @lioncash any thoughts on if this is a real error?

lioncash:
This is likely an error that's just an artifact of trying to merge this commit into another existing branch (since externals/libusb is already a submodule in the existing branch that this commit is being applied into after the fact (assuming I'm understanding the early-access commit applying process)

If it's directly merged into master it'll likely disappear (otherwise the same issue would have been hit by the buildbots for this PR).

PR #4588, opened 2020-08-26
common/thread: Fix data race in is_set

ReinUsesLisp:
As report by tsan, Event::Set can write is_set while WaitFor and friends are reading from it. To address this issue, make is_set an atomic.

### DIFF
--- a/src/common/thread.h
+++ b/src/common/thread.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <atomic>
 #include <chrono>
 #include <condition_variable>
 #include <cstddef>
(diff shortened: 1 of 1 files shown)


src/common/thread.h
```diff
@@ -54,9 +55,9 @@ class Event {
     }
 
 private:
-    bool is_set = false;
     std::condition_variable condvar;
     std::mutex mutex;
+    std::atomic_bool is_set{false};
```
lioncash, line 60: This can be a `std::atomic_flag`

ReinUsesLisp: I just tried using `std::atomic_flag`. `libc++` lacks C++20's `test` required to trivially implement this abstraction with it.

PR #4599, opened 2020-08-28
vk_device: Blacklist AMD proprietary from VK_EXT_extended_dynamic_state

ReinUsesLisp:
Vertex binding's <stride> is bugged on AMD's proprietary drivers when
using VK_EXT_extended_dynamic_state. Blacklist it for now while we
investigate how to report this issue to AMD.

It might be possible the driver is interpreting `<stride>` as in OpenGL's `VertexAttribPointer` instead of `BindVertexBuffer`.

- Fixes [SMO corruption](https://user-images.githubusercontent.com/26564870/91619271-0699d380-e963-11ea-8315-ff87ff107846.png) on the latest AMD proprietary drivers.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_device.cpp
+++ b/src/video_core/renderer_vulkan/vk_device.cpp
@@ -691,7 +691,12 @@ std::vector<const char*> VKDevice::LoadExtensions() {
         }
     }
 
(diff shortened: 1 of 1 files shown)


goldenx86:
This problem still persist at least on Vega GPUs.

PR #4602, opened 2020-08-29
cmake: Always enable Vulkan

ReinUsesLisp:
This reduces the maintenance burden of having code hidden by
preprocessor macros. All Vulkan files are always included (aside from
Nsight Aftermath that is kept out-of-tree).

The original intention behind this setting was to keep Vulkan disabled while it was being developed and then it was used when Linux distros lacked an updated version of Qt. We don't have any reason to keep this cmake option.


lioncash:
LGTM once the build is fixed

lioncash:
Conflicts need to be fixed

ReinUsesLisp:
I'll revisit this later. Closing for now.

PR #4627, opened 2020-09-03
sdl_impl: Erase the SDLJoystick entry after removing a controller

### LINKED ISSUE #4268: Yuzu doesn't react to button pressing after disconnecting and connecting DS4 again during gaming
I got DS4 controller connected through micro usb. Sometimes during excessive motions it briefly disconnects from PC. After that Yuzu stops react on it until I remapped the buttons or relaunch the application entirely. It's annoying to say the least.
I'm using it without DS4 Windows wrapper, with it it's fine.

Morph1984:
Previously, disconnecting a controller still leaves a null SDLJoystick entry within the vector of SDLJoysticks mapped by GUID.
When a DirectInput device of the same GUID is reconnected, it adds that device to a new port causing non-detectable input.
Furthermore, opening the "Configure" menu would cause yuzu to crash since it first tries to resolve the name of a null SDLJoystick entry that was not removed.
Resolve this by properly erasing the SDLJoystick entry from the vector, this resolves #4268.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -273,21 +273,19 @@ void SDLState::InitJoystick(int joystick_index) {
 void SDLState::CloseJoystick(SDL_Joystick* sdl_joystick) {
     const std::string guid = GetGUID(sdl_joystick);
 
-    std::shared_ptr<SDLJoystick> found_joystick;
-    {
-        std::lock_guard lock{joystick_map_mutex};
-        // This call to guid is safe since the joystick is guaranteed to be in the map
-        const auto& joystick_guid_list = joystick_map[guid];
-        const auto joystick_it = std::find_if(joystick_guid_list.begin(), joystick_guid_list.end(),
-                                              [&sdl_joystick](const auto& joystick) {
-                                                  return joystick->GetSDLJoystick() == sdl_joystick;
-                                              });
-        found_joystick = *joystick_it;
-    }
+    std::lock_guard lock{joystick_map_mutex};
+    auto& joystick_guid_list = joystick_map[guid];
+    auto joystick_it = std::find_if(
+        joystick_guid_list.begin(), joystick_guid_list.end(),
+        [&sdl_joystick](auto& joystick) { return joystick->GetSDLJoystick() == sdl_joystick; });
 
-    // Destruct SDL_Joystick outside the lock guard because SDL can internally call the
-    // event callback which locks the mutex again.
-    found_joystick->SetSDLJoystick(nullptr, nullptr);
+    if (joystick_it != joystick_guid_list.end()) {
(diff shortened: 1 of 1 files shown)


Morph1984:
Do not merge until I find a better solution, I realized this solution is short-sighted

Morph1984:
I have pushed a new solution that should theoretically work however it is quite inconsistent? After repeatedly connecting and disconnecting the same controller it eventually fails to erase the entry and causes the same crash.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -234,22 +234,19 @@ void SDLState::InitJoystick(int joystick_index) {
...
-        const auto& joystick_guid_list = joystick_map[guid];
-        const auto joystick_it =
-            std::find_if(joystick_guid_list.begin(), joystick_guid_list.end(),
-                         [&sdl_joystick](const std::shared_ptr<SDLJoystick>& joystick) {
-                             return joystick->GetSDLJoystick() == sdl_joystick;
-                         });
-        joystick = *joystick_it;
+    auto& joystick_guid_list = joystick_map[guid];
+    auto joystick_it = std::find_if(joystick_guid_list.begin(), joystick_guid_list.end(),
+                                    [&sdl_joystick](const std::shared_ptr<SDLJoystick>& joystick) {
```
lioncash, line 239:
```suggestion
                                    [&sdl_joystick](const auto& joystick) {
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -234,22 +234,19 @@ void SDLState::InitJoystick(int joystick_index) {
...
+                                        return joystick->GetSDLJoystick() == sdl_joystick;
+                                    });
+
+    if (joystick_it != joystick_map[guid].end()) {
```
lioncash, line 243:
```suggestion
    if (joystick_it != joystick_guid_list.end()) {
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -234,22 +234,19 @@ void SDLState::InitJoystick(int joystick_index) {
...
+        (*joystick_it)->SetSDLJoystick(nullptr, nullptr);
+        joystick_it = joystick_map[guid].erase(joystick_it);
```
lioncash, line 245:
```suggestion
        joystick_it = joystick_guid_list.erase(joystick_it);
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -234,22 +234,19 @@ void SDLState::InitJoystick(int joystick_index) {
...
+        if (joystick_map[guid].empty()) {
```
lioncash, line 246:
```suggestion
        if (joystick_guid_list.empty()) {
```

bunnei (approved):

PR #4643, opened 2020-09-10
Test: Decrease pad_update_ns

FearlessTobi:
There have been reports of quite heavy input lag in the past.
Compared to Citra for example, our pad_update_ns value is very high.
So let's decrease it and see if it helps with this problem.


german77:
This may break motion in some games as it is really time sensitive. We should use another poll rate just for motion. I will have to do some real HW test to find out the actual poll rate for buttons and motion.

bunnei (approved):

bunnei:
@german77 I am going to merge this considering it has been in EA for over a week without any reported regressions. However, if you have any concerns with motion, please reach out on Discord or submit a follow up PR.

PR #4659, opened 2020-09-16
Add a stub for LoadOpenContext

ZiggyDev:
There isn't any research done on this call, but building an empty response with a param size of 2 should work for now.


FearlessTobi:
This stub is a) very likely wrong and b) you haven't even stated which games this is needed for.
Therefore I'm closing this

NielsGx:
Used by "Dr Kawashima’s Brain Training" for Nintendo Switch at least

as stated in issue #4660

PR #4674, opened 2020-09-17
renderer_vulkan: Make unconditional use of VK_KHR_timeline_semaphore

ReinUsesLisp:
This reworks how host<->device synchronization works on the Vulkan
backend. Instead of "protecting" resources with a fence and signalling
these as free when the fence is known to be signalled by the host GPU,
use timeline semaphores.

Vulkan timeline semaphores allow use to work on a subset of D3D12
fences. As far as we are concerned, timeline semaphores are a value set
by the host or the device that can be waited by either of them.

Taking advantange of this, we can have a monolithically increasing
atomic value for each submission to the graphics queue. Instead of
protecting resources with a fence, we simply store the current logical
tick (the atomic value stored in CPU memory). When we want to know if a
resource is free, it can be compared to the current GPU tick.

This greatly simplifies resource management code and the free status of
resources should have less false negatives.

To workaround bugs in validation layers, when these are attached there's
a thread waiting for timeline semaphores.


ReinUsesLisp:
This introduces issues on Breath of the Wild. Investigating.

behunin:
[See this](https://github.com/behunin/yuzu/commit/aa6e49d7357d498ce89b52e145c794697aba1709), I renamed it to be the same as [switchbrews nvhost-ctrl](https://switchbrew.org/wiki/NV_services#NVHOST_IOCTL_CTRL_SYNCPT_CLEAR_EVENT_WAIT), and removed the gpu check as switchbrew says this function is exclusive to the switch, so it seamed unnecessary.

Breath of the wild calls this often so am I thought it might help.

Also noticed that there is a lot of:
`HW.GPU <Error> ./video_core/texture_cache/texture_cache.h:TryCopyImage:1194: Illegal copy between formats={39, 87}`
`HW.GPU <Error> ./video_core/texture_cache/texture_cache.h:TryCopyImage:1194: Illegal copy between formats={87, 39}`

Which I believe is how the game does blooming on bright lights.

bunnei (approved):

PR #4676, opened 2020-09-18
am: Stub GetPreviousProgramIndex

Morph1984:
- Used in Super Mario 3D All-Stars

Also allows it to boot to the launcher with #4675 
image

### DIFF
--- a/src/core/hle/service/am/am.h
+++ b/src/core/hle/service/am/am.h
@@ -288,11 +288,13 @@ private:
     void SetApplicationCopyrightVisibility(Kernel::HLERequestContext& ctx);
     void QueryApplicationPlayStatistics(Kernel::HLERequestContext& ctx);
     void QueryApplicationPlayStatisticsByUid(Kernel::HLERequestContext& ctx);
+    void GetPreviousProgramIndex(Kernel::HLERequestContext& ctx);
     void GetGpuErrorDetectedSystemEvent(Kernel::HLERequestContext& ctx);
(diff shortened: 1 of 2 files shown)


src/core/hle/service/am/am.h
```diff
@@ -288,11 +288,13 @@ class IApplicationFunctions final : public ServiceFramework<IApplicationFunction
     void SetApplicationCopyrightVisibility(Kernel::HLERequestContext& ctx);
     void QueryApplicationPlayStatistics(Kernel::HLERequestContext& ctx);
     void QueryApplicationPlayStatisticsByUid(Kernel::HLERequestContext& ctx);
+    void GetPreviousProgramIndex(Kernel::HLERequestContext& ctx);
     void GetGpuErrorDetectedSystemEvent(Kernel::HLERequestContext& ctx);
     void GetFriendInvitationStorageChannelEvent(Kernel::HLERequestContext& ctx);
 
     bool launch_popped_application_specific = false;
     bool launch_popped_account_preselect = false;
+    s32 previous_program_index{-1};
```
lioncash, line 297: This doesn't seem to actually be assigned a value other than `-1`. Is this what the actual service does? If not, the relevant function should likely be marked as a stub rather than a complete implementation.

Morph1984: This seems to be set by ExecuteProgram, I'll mark this as a stub for now

bunnei (approved):

PR #4677, opened 2020-09-18
InputCommon: Add random motion input for buttons

german77:
This allows to assign other types of inputs as a random motion input. This effectively triggers shake actions in multiple games like SMO.

### DIFF
--- a/src/input_common/motion_input.cpp
+++ b/src/input_common/motion_input.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included
 
+#include <random>
 #include "common/math_util.h"
 #include "input_common/motion_input.h"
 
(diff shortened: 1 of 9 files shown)


src/input_common/motion_input.cpp
```diff
@@ -159,6 +159,35 @@ Common::Vec3f MotionInput::GetRotations() const {
     return rotations;
 }
 
+Input::MotionStatus MotionInput::GetMotion() const {
+    Common::Vec3f gyroscope = GetGyroscope();
```
lioncash, line 163: These variables should be const

src/input_common/motion_input.cpp
```diff
@@ -159,6 +159,35 @@ Common::Vec3f MotionInput::GetRotations() const {
...
+    Common::Vec3f accelerometer = GetAcceleration();
+    Common::Vec3f rotation = GetRotations();
+    std::array<Common::Vec3f, 3> orientation = GetOrientation();
+    return {accelerometer , gyroscope , rotation, orientation};
+}
+
+Input::MotionStatus MotionInput::GetRandomMotion(int accel_magnitude,
+                                                       int gyro_magnitude) const {
+    Common::Vec3f gyroscope = {
```
lioncash, line 171: Ditto

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -95,6 +96,10 @@ class SDLJoystick {
         return std::make_tuple(x, y);
     }
 
+    InputCommon::MotionInput GetMotion() const {
```
lioncash, line 99:
```suggestion
    const InputCommon::MotionInput& GetMotion() const {
```

MotionInput instances are expensive to copy, so its best to allow the calling code to take it by reference if it only needs to query information.

src/input_common/udp/client.cpp
```diff
@@ -250,6 +246,8 @@ void Client::OnPadData(Response::PadData data) {
         clients[client].status.touch_status = {x, y, is_active};
 
         if (configuring) {
+            Common::Vec3f gyroscope = clients[client].motion.GetGyroscope();
```
lioncash, line 249: These should be const

src/input_common/motion_input.cpp
```diff
@@ -159,6 +159,34 @@ Common::Vec3f MotionInput::GetRotations() const {
...
+    Common::Vec3f gyroscope = GetGyroscope();
+    Common::Vec3f accelerometer = GetAcceleration();
+    Common::Vec3f rotation = GetRotations();
+    std::array<Common::Vec3f, 3> orientation = GetOrientation();
+    return {accelerometer, gyroscope, rotation, orientation};
+}
+
+Input::MotionStatus MotionInput::GetRandomMotion(int accel_magnitude, int gyro_magnitude) const {
+    Common::Vec3f gyroscope = {
+        (rand() % 2000 - 1000) * 0.001f,
```
lioncash, line 172: Prefer using facilities in `<random>` over `std::rand`

v1993:
Seems like a good idea to me, but I think that "SDLMotion" sounds a bit misleading, considering that it is very well possible to use SDL for real motion input in some cases.

german77:
SDL is event triggered. Rigth now I only added buttons and axis events. When SDL implements real motion you just have to add the event in the switch case.

In other words this PR adds the backbone in SDL to implement real motion. SDLMotion is the correct name

v1993:
I see what you mean. The thing is, currently it's possible to read motion as axis events (3 axles for acceleration, 3 for gyro) using SDL on some devices. But if you intend to wait until SDL adds motion events, this name makes perfect sense.

german77:
Adding 6 new inputs to just have motion seems a little to complicated for the end user. For me keeping it as simple as posible is preferred.  Feel free to implement it if you like.

lioncash:
A lot of comments seem to be marked as resolved when they aren't actually addressed, so I've marked them as unresolved.

bunnei (approved):

PR #4678, opened 2020-09-18
acc: Partially implement LoadOpenContext

Morph1984:
This is used in multiple games such as:
- Clubhouse Games: 51 Worldwide Classics
- Grandia HD Collection
- XCOM 2 Collection
- Baldur's Gate 1/2
- Dr Kawashima's Brain Training
- Super Mario 3D All-Stars

### DIFF
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -774,6 +774,17 @@ void Module::Interface::ListQualifiedUsers(Kernel::HLERequestContext& ctx) {
     rb.Push(RESULT_SUCCESS);
 }
 
+void Module::Interface::LoadOpenContext(Kernel::HLERequestContext& ctx) {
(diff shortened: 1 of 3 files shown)


src/core/hle/service/acc/acc.cpp
```diff
@@ -774,6 +774,17 @@ void Module::Interface::ListQualifiedUsers(Kernel::HLERequestContext& ctx) {
     rb.Push(RESULT_SUCCESS);
 }
 
+void Module::Interface::LoadOpenContext(Kernel::HLERequestContext& ctx) {
+    LOG_WARNING(Service_ACC, "(Partial Implementation) called");
```
FearlessTobi, line 778: Why not just call this a stub then and be consistent with the rest of the codebase?

bunnei (approved):

PR #4686, opened 2020-09-19
nvdrv: Implement Ioctl2 command SubmitGPFIFOEx

ZiggyDev:
Needed by:
- Super Mario 3D All-Stars
And possibly some other newer games.

Although this implements the command, this does not fully fix the issue for Super Mario 3D All-Stars, as the flag "add_increment" is unimplemented. Maybe someone with a better understanding of that can add to this PR.

I tried to streamline the code by having both SubmitGPFIFO and SubmitGPFIFOEx call the same function and differentiating between them inside it, since they are essentially the same command, just different ioctl versions. If you'd rather have two separate functions you're more than welcome to edit this.


bylaws:
Iirc add_increment increments the syncpoint associated with the fence after the execution of all pushed gpfifo entries, though if async GPU is disabled it shouldn't matter.

bylaws:
Though this patch seems wrong, yuzu implements it but calls it 'KickoffPB'

bunnei:
This has been implemented already, closing

Issue #4687, opened 2020-09-19
Slowdowns in Smash Ultimate intro and Paper Mario Origami King Nvidia Opengl drivers newer than 446.14


IDSG:
I wanna know if this issue is from nvidia side (driver related bug) or some kind implementation that "worked" but some driver messed things up, reverted to 446.14 driver is not an option anymore if user having 30xx series cards, and wants  fixes for newer games, so there's something you can report back to nvidia dev team to fix the issue or make a change in yuzu to behave correctly in newer drivers? 

BTW, using vulkan everyhing works right, but vulkan doesnt have shader cache

ReinUsesLisp:
As a temporary workaround if you attach Nsight Graphics to yuzu, the issue is fixed.

goldenx86:
That means you can't profile the issue with the official tool... Great job Nvidia.

ReinUsesLisp:
Upon further investigation, it's caused by our vsync implementation.

theboy181:
I noticed that too. Any word on Blink?

IDSG:
Oh nice, it's working fine on EA 978 build, thanks, i know its a workround, hope you finde the right fix

ReinUsesLisp:
Thanks for reporting this.

PR #4688, opened 2020-09-19
renderer_opengl: Disable vsync (DO NOT MERGE)

ReinUsesLisp:
Our software vsync implementation causes performance issues on recent Nvidia drivers.
As a temporary fix, disable it.

See #4687.


ReinUsesLisp:
Superseeded by #4692.

PR #4689, opened 2020-09-20
audio_renderer: Fix biquad filter

ogniK5377:
Originally the biquad filter was disabled due to issues with crackling before.

### DIFF
--- a/src/audio_core/command_generator.cpp
+++ b/src/audio_core/command_generator.cpp
@@ -270,11 +270,9 @@ void CommandGenerator::GenerateBiquadFilterCommandForVoice(ServerVoiceInfo& voic
         }
 
         // Generate biquad filter
-        //        GenerateBiquadFilterCommand(mix_buffer_count, biquad_filter,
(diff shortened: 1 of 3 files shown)


v1993:
This is causing issues in Kirby: Star Allies. During stage three of final boss battle parts of music sound broken.

Here's how it sounds with this patch:

https://www.dropbox.com/s/yddl781mutvwmmf/kirby_bad.ogg?dl=0

Here's how it sounds normally:

https://www.dropbox.com/s/3pqfmukce8ol6qa/kirby_good.ogg?dl=0

Note that both versions include some sound effects from recorded gameplay. Curiously, issue doesn't happen while I stay in pause menu, either because of quieter music or different filtering.

v1993:
Can this, please, be excluded from EA merge until issue I've pointed out is fixed?

bunnei:
@ogniK5377 can you take a look at the above issue? Thanks!

bunnei:
@ogniK5377 ping here

PR #4690, opened 2020-09-20
style: C++20 module keyword fixup

ogniK5377:
With C++20, "module" is now considered a keyword, this PR removes mentions of the word "module".


src/video_core/renderer_vulkan/wrapper.h
```diff
@@ -14,7 +14,9 @@
 #include <vector>
 
 #define VK_NO_PROTOTYPES
+#define module vkModule
```
lioncash, line 17: There should really be a comment about why this exists

ReinUsesLisp:
These are not real errors. Latest MSVC supports modules and the code builds fine. It seems like the EDG compiler built into Visual Studio is not aware of `module` being a contextual keyword.

ameerj:
Closing since many of the files changed have diverged immensely, and the bug incorrectly flagging the use of `module` in Visual Studio has since been resolved.

PR #4713, opened 2020-09-24
Start of Integer flags implementation

behunin:
This is the start of implementing shader_ir flags.
I have only tested this on Breath of the Wild.


src/video_core/shader/shader_ir.h
```diff
@@ -346,6 +346,9 @@ class ShaderIR final {
     /// Access a bindless image sampler.
     Image& GetBindlessImage(Tegra::Shader::Register reg, Tegra::Shader::ImageType type);
 
+    /// Recursive Iteration over the OperationNode operands
+    void Iterop(NodeBlock& nb, Node var);
```
lioncash, line 350: This should likely have a more informative name, Like `RecursiveOpIterate` or something.

src/video_core/shader/shader_ir.cpp
```diff
@@ -387,9 +387,49 @@ void ShaderIR::SetInternalFlagsFromInteger(NodeBlock& bb, Node value, bool sets_
     if (!sets_cc) {
         return;
     }
-    Node zerop = Operation(OperationCode::LogicalIEqual, std::move(value), Immediate(0));
-    SetInternalFlag(bb, InternalFlag::Zero, std::move(zerop));
-    LOG_WARNING(HW_GPU, "Condition codes implementation is incomplete");
+    switch (value->index()) {
+    case 0:
```
lioncash, line 391: Magic values

src/video_core/shader/shader_ir.cpp
```diff
@@ -387,9 +387,49 @@ void ShaderIR::SetInternalFlagsFromInteger(NodeBlock& bb, Node value, bool sets_
...
+        Iterop(bb, value);
+        break;
+    case 2:
+        if (const auto gpr = std::get_if<GprNode>(value.get())) {
+            LOG_WARNING(HW_GPU, "GprNode: index={}", gpr->GetIndex());
```
lioncash, line 396: This should be a debug log

src/video_core/shader/shader_ir.cpp
```diff
@@ -387,9 +387,49 @@ void ShaderIR::SetInternalFlagsFromInteger(NodeBlock& bb, Node value, bool sets_
...
+    default:
+        Node zerop = Operation(OperationCode::LogicalIEqual, std::move(value), Immediate(0));
+        SetInternalFlag(bb, InternalFlag::Zero, std::move(zerop));
+        LOG_WARNING(HW_GPU, "Node Type: {}", value->index());
+        break;
+    }
+}
+
+void ShaderIR::Iterop(NodeBlock& nb, Node var) {
+    if (const auto op = std::get_if<OperationNode>(var.get())) {
```
lioncash, line 412: We generally use guard clauses to reduce nesting where possible:

```suggestion
    const auto* op = std::get_if<OperationNode>(var.get());
    if (op == nullptr) {
        return;
    }
    
    if (op->GetOperandsCount() == 0) {
        return;
    }
    
    // rest of the code here
```

src/video_core/shader/shader_ir.cpp
```diff
@@ -387,9 +387,49 @@ void ShaderIR::SetInternalFlagsFromInteger(NodeBlock& bb, Node value, bool sets_
...
+        if (op->GetOperandsCount() > 0) {
+            for (auto& opss : op->GetOperands()) {
```
lioncash, line 414:
```suggestion
            for (auto& operand : op->GetOperands()) {
```

src/video_core/shader/shader_ir.cpp
```diff
@@ -387,9 +387,49 @@ void ShaderIR::SetInternalFlagsFromInteger(NodeBlock& bb, Node value, bool sets_
...
+                switch (opss->index()) {
+                case 0:
+                    return Iterop(nb, opss);
+                case 2:
+                    if (const auto gpr = std::get_if<GprNode>(opss.get())) {
+                        LOG_WARNING(HW_GPU, "Child GprNode: index={}", gpr->GetIndex());
```
lioncash, line 420: Debug log

src/video_core/shader/shader_ir.cpp
```diff
@@ -387,9 +387,49 @@ void ShaderIR::SetInternalFlagsFromInteger(NodeBlock& bb, Node value, bool sets_
...
+
+void ShaderIR::Iterop(NodeBlock& nb, Node var) {
+    if (const auto op = std::get_if<OperationNode>(var.get())) {
+        if (op->GetOperandsCount() > 0) {
+            for (auto& opss : op->GetOperands()) {
+                switch (opss->index()) {
+                case 0:
+                    return Iterop(nb, opss);
+                case 2:
+                    if (const auto gpr = std::get_if<GprNode>(opss.get())) {
```
lioncash, line 419:
```suggestion
                    if (const auto* gpr = std::get_if<GprNode>(opss.get())) {
```

src/video_core/shader/node.h
```diff
@@ -464,6 +464,10 @@ class OperationNode final : public AmendNode {
         return operands.size();
     }
 
+    NodeBlock GetOperands() const {
```
lioncash, line 467:
```suggestion
    const NodeBlock& GetOperands() const {
        return operands;
    }
```

So usages of this function don't churn std::vector allocations.

src/video_core/shader/shader_ir.h
behunin: I like that suggestion, here is a few more: OperandsIteration, RecurseOperands, SearchOperands.
Whatever seems clever to the team.

src/video_core/shader/shader_ir.cpp
behunin: This is the std::variant<Types...>::index, so would a comment above the switch statement be sufficient enough to clarify?

behunin: // value->index() is the index of the std::variant, tells us which Node type we have here.

What ya think?

src/video_core/shader/shader_ir.h
behunin: /// Recursive Iteration over the OperationNode operands, searching for a GprNode.

I am favoring SearchOperands now. Cool?

src/video_core/shader/shader_ir.cpp
lioncash: This is sort of self-explanatory without really elaborating on the node type. It's likely more informative to add a comment by the number like:

```cpp
case 0: // Operation node
    ...
````

behunin:
I don't know why CI didn't build successfully. Can someone please trigger the build again?

v1993:
@behunin I think you can rebase on top of new commits in master and force-push to trigger a CI rebuild.

CI build have completed, but [tests have failed](https://dev.azure.com/yuzu-emu/yuzu/_build/results?buildId=7957&view=logs&j=b457fee1-24b5-5b05-276f-95f55538c81f&t=8274a14c-2774-5fa1-24fc-2bf7929518a4&l=3485) for some reason.

behunin:
@v1993 Thank you! I see now it's just tests that failed, I am not really worried about that and am guessing the team feels the same.

behunin:
@bunnei @ReinUsesLisp @lioncash
Good to go er nah?

bunnei (approved):

adityaruplaha (commented):
I know this has been merged already, and it's a really minor thing but...

PR #4714, opened 2020-09-24
Rename to align with switchbrew and remove gpu function

behunin:
I renamed it to be the same as [switchbrews nvhost-ctrl](https://switchbrew.org/wiki/NV_services#NVHOST_IOCTL_CTRL_SYNCPT_CLEAR_EVENT_WAIT), and removed the gpu check as switchbrew says this function is exclusive to the switch, so it seamed unnecessary.

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
@@ -36,8 +36,8 @@ u32 nvhost_ctrl::ioctl(Ioctl command, const std::vector<u8>& input, const std::v
         return IocCtrlEventRegister(input, output);
     case IoctlCommand::IocCtrlEventUnregisterCommand:
         return IocCtrlEventUnregister(input, output);
-    case IoctlCommand::IocCtrlEventSignalCommand:
-        return IocCtrlEventSignal(input, output);
+    case IoctlCommand::IocCtrlClearEventWaitCommand:
+        return IocCtrlClearEventWait(input, output);
     default:
         UNIMPLEMENTED_MSG("Unimplemented ioctl");
         return 0;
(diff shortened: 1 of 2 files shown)


ReinUsesLisp:
Your description doesn't coincide with the code changed. There's no check for this running on an emulated Switch.

src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
```diff
@@ -154,23 +154,17 @@ u32 nvhost_ctrl::IocCtrlEventUnregister(const std::vector<u8>& input, std::vecto
...
-    u32 event_id = params.user_event_id & 0x00FF;
-    LOG_WARNING(Service_NVDRV, "(STUBBED) called, user_event_id: {:X}", event_id);
+    u32 event_id = params.event_id & 0x00FF;
+    LOG_WARNING(Service_NVDRV, "cleared event wait on, event_id: {:X}", event_id);
     if (event_id >= MaxNvEvents) {
         return NvResult::BadParameter;
     }
     if (events_interface.status[event_id] == EventState::Waiting) {
-        auto& gpu = system.GPU();
-        if (gpu.CancelSyncptInterrupt(events_interface.assigned_syncpt[event_id],
```
behunin, line 169: What I am trying to say is that I removed this function call. I am sorry for the confusion.

lioncash:
Your commit message should likely be more informative, given the changes involve more than just renaming.

src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
ReinUsesLisp: Why did you remove it?

behunin: After reading the switchbrew description that I linked.

> Clears the wait signal of an event. Exclusive to the Switch.

I started to ponder why would this be exclusive to the switch and arrived at two possible reasons.

1. The switch is set to 30 fps on a lot of the games and so frame drop will be very noticeable. This function could be called to cancel waiting on long running events to reduce stutter.

2. Creating an effect of stutter to give the user a sense of being dazed or stunned. Or maybe just for an artistic intentions.

Unless someone with inside knowledge gives the real reason that this function is included we may never know why it was added...

So I am taking NVHOST_IOCTL_CTRL_SYNCPT_CLEAR_EVENT_WAIT for face value. Clear the signal to wait on the event, move on to the next. Don't check what the event may have been working on and see if you can remove it, just move on.

bunnei:
Sorry for the delay here -- change looks good.

bunnei (approved):

PR #4731, opened 2020-09-29
common: Use system zstd on Linux

lat9nq:
From what I understand, this tells CMake to use the system, not conan, version of zstd on Linux machines. Required to build with the coming MinGW Docker container updates.


v1993:
Does this also disable building conan version of zstd?

bunnei (approved):

PR #4732, opened 2020-09-29
common/wall_clock: Add virtual destructors

ReinUsesLisp:
From -fsanitize=address, this code wasn't calling the proper destructor.
Adding virtual destructors for each inherited class and the base class
fixes this bug.

While we are at it, mark the classes as final.

### DIFF
--- a/src/common/wall_clock.cpp
+++ b/src/common/wall_clock.cpp
@@ -15,7 +15,7 @@ namespace Common {
 using base_timer = std::chrono::steady_clock;
 using base_time_point = std::chrono::time_point<base_timer>;
 
-class StandardWallClock : public WallClock {
+class StandardWallClock final : public WallClock {
 public:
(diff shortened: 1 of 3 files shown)


src/common/wall_clock.cpp
```diff
@@ -15,13 +15,15 @@ namespace Common {
...
 
-class StandardWallClock : public WallClock {
+class StandardWallClock final : public WallClock {
 public:
     StandardWallClock(u64 emulated_cpu_frequency, u64 emulated_clock_frequency)
         : WallClock(emulated_cpu_frequency, emulated_clock_frequency, false) {
         start_time = base_timer::now();
     }
 
+    ~StandardWallClock() override = default;
```
lioncash, line 25: It should be sufficient to provide the virtual dtor in the base class, these extra dtors don't need to be specified if they don't do anything separately.

lioncash (approved):

PR #4735, opened 2020-09-30
Remove ext_extended_dynamic_state blacklist

goldenx86:
Latest AMD 20.9.2 driver fixed this, there's no reason to keep it blocked, as the previous stable signed driver release doesn't include the extension.


Morph1984 (approved):
Yes

lioncash (approved):

PR #4739, opened 2020-10-01
InputCommon: Fix offbrand GCAdapter

german77:
Allows offbrand GCAdapters to work, port from https://github.com/dolphin-emu/dolphin/pull/8933

### DIFF
--- a/src/input_common/gcadapter/gc_adapter.cpp
+++ b/src/input_common/gcadapter/gc_adapter.cpp
@@ -123,11 +123,10 @@ void Adapter::Read() {
                                   sizeof(adapter_payload), &payload_size, 16);
 
         if (payload_size != sizeof(adapter_payload) || adapter_payload[0] != LIBUSB_DT_HID) {
+            // This can occur for a few frames on initialization.
             LOG_ERROR(Input,
                       "Error reading payload (size: {}, type: {:02x}) Is the adapter connected?",
                       payload_size, adapter_payload[0]);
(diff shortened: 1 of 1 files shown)


src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -218,6 +217,12 @@ bool Adapter::CheckDeviceAccess(libusb_device* device) {
         }
     }
 
+    int control_transfer_error =
+        libusb_control_transfer(usb_adapter_handle, 0x21, 11, 0x0001, 0, nullptr, 0, 1000);
```
ameerj, line 221: Just out of curiosity, what are these numbers?

src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -218,6 +217,12 @@ bool Adapter::CheckDeviceAccess(libusb_device* device) {
         }
     }
 
+    int control_transfer_error =
```
lioncash, line 220:
```suggestion
    const int control_transfer_error =
```

src/input_common/gcadapter/gc_adapter.cpp
german77: Magic numbers. I haven't done any research just added the fix from dolphin.

ameerj: Can you please add a comment just to clarify the purpose of this line for anyone looking at the code later on.

PR #4741, opened 2020-10-01
HID/configuration: Use vibration strength from settings

german77:
This makes the vibration strength setting on the UI actually work.


v1993:
It appears that in all places except actual vibration force multiplications percentage value is used as opposed to tiny one. Wouldn't it make sense to stick to it everywhere, making code simpler and avoiding floats until needed?

german77:
This is a temporary fix. This PR would be closed when the sucesor is open.

PR #4742, opened 2020-10-02
HID: Only use inputs corresponding to controller type

german77:
This prevents using buttons that aren't supposed to exist in the controller.

Example:
You map all buttons xbox controller as a pro controller then change the controller type to left joycon. 
On SSBU you can still use A,X,Y,Z and right stick of the xbox controller.


bunnei (approved):

bunnei:
LGTM, thanks!

PR #4755, opened 2020-10-03
InputCommon: Implement GC vibrations

german77:
Adds a rather basic output form of vibration for GC controllers

### DIFF
--- a/src/input_common/gcadapter/gc_adapter.cpp
+++ b/src/input_common/gcadapter/gc_adapter.cpp
@@ -16,6 +16,7 @@
 
 #include "common/logging/log.h"
 #include "common/param_package.h"
+#include "core/settings.h"
 #include "input_common/gcadapter/gc_adapter.h"
 #include "input_common/settings.h"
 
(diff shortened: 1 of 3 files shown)


src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -218,6 +238,12 @@ bool Adapter::CheckDeviceAccess(libusb_device* device) {
         }
     }
 
+    int control_transfer_error =
```
lioncash, line 241:
```suggestion
    const int control_transfer_error =
```

src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -380,6 +406,23 @@ InputCommon::AnalogMapping Adapter::GetAnalogMappingForDevice(
     return mapping;
 }
 
+void Adapter::vibrate(bool p1, bool p2, bool p3, bool p4) {
+    int size = 0;
+    u8 payload[5] = {0x11, p1, p2, p3, p4};
```
lioncash, line 411: std::array

src/input_common/gcadapter/gc_adapter.h
```diff
@@ -113,6 +115,9 @@ class Adapter {
     /// For use in initialization, querying devices to find the adapter
     void Setup();
 
+    // Updates vibration state of all controllers
+    void vibrate(bool p1, bool p2, bool p3, bool p4);
```
lioncash, line 119:
```suggestion
    void Vibrate(bool p1, bool p2, bool p3, bool p4);
```

PR #4757, opened 2020-10-04
InputCommon: Add compatibility with only accelerometer and auto calibrate for drift

german77:
This would allow to have fast motion with devices that only have accelerometer and make them be able to trigger shake events. Gyro is needed for true motion tracking having only accelerometer will not detect movements on the perpendicular plane to the gravity.

Also slowly auto calibrate for drift if the device is still. Improving accuracy on the long run.

As an extra, estimate the correct orientation of the controller if an orientation reset occurs. This will remove slow drifts when a orientation reset happens.

### DIFF
--- a/src/input_common/motion_input.cpp
+++ b/src/input_common/motion_input.cpp
@@ -16,8 +16,16 @@ void MotionInput::SetAcceleration(const Common::Vec3f& acceleration) {
 
 void MotionInput::SetGyroscope(const Common::Vec3f& gyroscope) {
     gyro = gyroscope - gyro_drift;
+
+    // Auto adjust drift to minimize drift
+    if (!IsMoving(0.1f)) {
+        gyro_drift = (gyro_drift * 0.9999f) + (gyroscope * 0.0001f);
+    }
+
     if (gyro.Length2() < gyro_threshold) {
         gyro = {};
+    } else {
+        only_accelerometer = false;
     }
 }
 
@@ -49,7 +57,7 @@ bool MotionInput::IsCalibrated(f32 sensitivity) const {
     return real_error.Length() < sensitivity;
 }
 
-void MotionInput::UpdateRotation(u64 elapsed_time) {
+void MotionInput::UpdateRotation(const u64 elapsed_time) {
     const f32 sample_period = elapsed_time / 1000000.0f;
     if (sample_period > 0.1f) {
         return;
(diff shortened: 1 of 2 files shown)


src/input_common/motion_input.h
```diff
@@ -29,8 +29,8 @@ class MotionInput {
     void EnableReset(bool reset);
     void ResetRotations();
 
-    void UpdateRotation(u64 elapsed_time);
-    void UpdateOrientation(u64 elapsed_time);
+    void UpdateRotation(const u64 elapsed_time);
```
lioncash, line 32:
```suggestion
    void UpdateRotation(u64 elapsed_time);
```

Top-level const on a value type in the header doesn't do anything. It can be left on the parameter in the cpp file though.

src/input_common/motion_input.cpp
```diff
@@ -143,6 +174,67 @@ std::array<Common::Vec3f, 3> MotionInput::GetOrientation() const {
...
+    const f32 az = -normal_accel.z;
+
+    while (!IsCalibrated(0.01f) && ++iterations < 100) {
+        // Short name local variable for readability
+        f32 q1 = quat.w;
+        f32 q2 = quat.xyz[0];
+        f32 q3 = quat.xyz[1];
+        f32 q4 = quat.xyz[2];
+
+        Common::Vec3f rad_gyro = {};
```
lioncash, line 193: This should be moved closer to its actual usage spot.

src/input_common/motion_input.cpp
```diff
@@ -16,8 +16,16 @@ void MotionInput::SetAcceleration(const Common::Vec3f& acceleration) {
 
 void MotionInput::SetGyroscope(const Common::Vec3f& gyroscope) {
     gyro = gyroscope - gyro_drift;
+
+    // Auto adjust drift to minimize drift
+    if (!IsMoving(0.1f)) {
```
v1993, line 21: Maybe there could be a setting to disable this? I can imagine scenario where gyro is already properly calibrated and this causes slight drift where there would be none at all.

Would also make sense if specific input drivers (e.g. physical joycons/procon) had an option to disable this for the same reason.

german77: From my test even the slightest change in temperature makes this value drift. There is no such thing as perfectly calibrated. This compensates for this error and will not introduce more drift to an already calibrated controller because it only takes measurements when the controller is still.

bunnei (approved):

PR #4758, opened 2020-10-04
InputCommon: Fix motion from mouse

german77:
Adds the necessary changes to output correct motion data to the emulator. To use it is needed to modify settings manually by setting the motion engine to "motion_emu". Then right click on the game screen to do a motion movement.

Only one motion_emu engine can be used at the same time. I'm investigating this to allow for multiple motion_emu engines in the settings file.

This type of motion cannot trigger shake events. I'm also looking to fix this.


german77:
Superseded by #4939

PR #4764, opened 2020-10-06
Filesystem: Implement OpenDataStorageWithProgramIndex

german77:
Program_index returned 0x00. Instead I used the current process to load the file contents. Updated the names of a few functions to match switchbrew.

used by RollerCoaster Tycoon 3: Complete Edition

### DIFF
--- a/src/core/hle/service/filesystem/fsp_srv.cpp
+++ b/src/core/hle/service/filesystem/fsp_srv.cpp
@@ -717,8 +717,8 @@ FSP_SRV::FSP_SRV(Core::System& system_)
         {201, nullptr, "OpenDataStorageByProgramId"},
         {202, &FSP_SRV::OpenDataStorageByDataId, "OpenDataStorageByDataId"},
         {203, &FSP_SRV::OpenPatchDataStorageByCurrentProcess, "OpenPatchDataStorageByCurrentProcess"},
-        {204, nullptr, "OpenDataFileSystemByProgramIndex"},
-        {205, nullptr, "OpenDataStorageByProgramIndex"},
+        {204, nullptr, "OpenDataFileSystemWithProgramIndex"},
+        {205, &FSP_SRV::OpenDataStorageWithProgramIndex, "OpenDataStorageWithProgramIndex"},
         {400, nullptr, "OpenDeviceOperator"},
         {500, nullptr, "OpenSdCardDetectionEventNotifier"},
         {501, nullptr, "OpenGameCardDetectionEventNotifier"},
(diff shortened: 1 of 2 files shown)


src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -991,6 +991,30 @@ void FSP_SRV::OpenPatchDataStorageByCurrentProcess(Kernel::HLERequestContext& ct
...
+    const auto storage_index = rp.PopRaw<u8>();
+
+    LOG_DEBUG(Service_FS, "called with storage_index={:02X}", storage_index);
+
+    // TODO: use storage_index
+    // Ignore storage index and proceed with current process
+    auto romfs = fsc.OpenRomFSCurrentProcess();
+    if (romfs.Failed()) {
+        // TODO (bunnei): Find the right error code to use here
+        LOG_CRITICAL(Service_FS, "no file system interface available!");
```
lioncash, line 1005: This should be an error log

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -991,6 +991,30 @@ void FSP_SRV::OpenPatchDataStorageByCurrentProcess(Kernel::HLERequestContext& ct
...
+        IPC::ResponseBuilder rb{ctx, 2};
+        rb.Push(RESULT_UNKNOWN);
+        return;
+    }
+
+    auto storage = std::make_shared<IStorage>(std::move(romfs.Unwrap()));
+
+    IPC::ResponseBuilder rb{ctx, 2, 0, 1};
+    rb.Push(RESULT_SUCCESS);
+    rb.PushIpcInterface<IStorage>(std::move(storage));
```
lioncash, line 1015:
```suggestion
    rb.PushIpcInterface(std::move(storage));
```

src/core/hle/service/filesystem/fsp_srv.cpp
```diff
@@ -991,6 +991,30 @@ void FSP_SRV::OpenPatchDataStorageByCurrentProcess(Kernel::HLERequestContext& ct
     rb.Push(FileSys::ERROR_ENTITY_NOT_FOUND);
 }
 
+void FSP_SRV::OpenDataStorageWithProgramIndex(Kernel::HLERequestContext& ctx) {
```
lioncash, line 994: Was any RE done to ensure that this behavior is correct with what the console is doing?

german77: I din't find any source to backup the changes apart from switchbrew 
I played for a while with no problems also save files were created correctly with no errors.
At least for this game is working fine.

lioncash: If you don't verify changes via RE, I would urge you to not make changes like this. We're emulating a console, not what we think the console should be doing.

lioncash:
Marking with needs hardware test, since no verification has been done on console behavior (or documenting console behavior and indicating where we diverge from it in specifics). Especially given RCT3 is low impact.

bunnei:
The implementation here seems fine and probably can be merged if we add an UNIMPLEMENTED_IF `program_index != 0`

PR #4772, opened 2020-10-09
vk_device: Block VK_EXT_extended_dynamic_state for RDNA devices

goldenx86:
RDNA devices seem to crash when using VK_EXT_extended_dynamic_state in
the latest 20.9.2 proprietary Windows drivers. As a workaround, for now
we block device names corresponding to current RDNA released products.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_device.cpp
+++ b/src/video_core/renderer_vulkan/vk_device.cpp
@@ -79,6 +79,21 @@ VkFormatFeatureFlags GetFormatFeatures(VkFormatProperties properties, FormatType
     }
 }
 
+[[nodiscard]] bool IsRDNA(std::string_view device_name, VkDriverIdKHR driver_id) {
+    static constexpr std::array RDNA_DEVICES{
+        "5700",
+        "5600",
+        "5500",
+        "5300",
+    };
+    if (driver_id != VK_DRIVER_ID_AMD_PROPRIETARY_KHR) {
+        return false;
+    }
+    return std::any_of(RDNA_DEVICES.begin(), RDNA_DEVICES.end(), [device_name](const char* name) {
+        return device_name.find(name) != std::string_view::npos;
+    });
+}
+
 std::unordered_map<VkFormat, VkFormatProperties> GetFormatProperties(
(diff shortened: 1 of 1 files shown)


goldenx86:
This was all done by ReinUsesLisp.

Wunkolo:
I think it might be better to use the `deviceID` property of [VkPhysicalDeviceProperties](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceProperties.html) than the current string-comparison method.

Sites like [this](https://pci-ids.ucw.cz/read/PC/1002) can help identify these devices as well. Next time there's a special-case set of cards or architectures that has to be handled differently(`"blah-blah is broken on blah-blah, so we disable it"`), the `deviceID` might be a more stable metric of detecting certain cards and architectures across platforms and operating systems rather than however a driver decides to name the device in Vulkan(Linux AMD drivers will name their devices `AMD RADV RAVEN (LLVM 10.0.1)` for instance).

deviceID| GPU
-- | --
7310 | Navi 10 [Radeon Pro W5700X]
7312 | Navi 10 [Radeon Pro W5700]
731f | Navi 10 [Radeon RX 5600 OEM/5600 XT / 5700/5700 XT]
7340 | Navi 14 [Radeon RX 5500/5500M / Pro 5500M]
7341 | Navi 14 [Radeon Pro W5500]
7347 | Navi 14 [Radeon Pro W5500M]
734f | Navi 14 [Radeon Pro W5300M]
7360 | Navi 12 [Radeon Pro 5600M]

goldenx86:
We considered it, but the list becomes too extensive, and we are already limiting the scope of this to only the Windows driver.
Besides, this will most likely just be temporal, the 20.9.2 driver fixed this extension for all AMD archs besides Navi, one would expect that the next release or so will finally fix it for RDNA1.

Morph1984 (approved):

bunnei (approved):

hurricane-dorian:
Yuzu no longer crashes with VK_EXT_extended_dynamic_state  enabled in driver version 20.11.1
Tested on a RX5600
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5541900/yuzu_log.txt)

Issue #4074, opened 2020-06-11
Pokemon Sword/Shield daily events broken


MohamedAarab:
None of the daily events can be done, such as pokejobs, max raids, berry trees, etc because the time doesn't advance. Pokejobs say that there is 24 hours remaining, but it never goes down.

I tried changing my computer time/date but that didn't fix it either.

MohamedAarab:
Apparently there is an unofficial patch going around that fixes this? Perhaps the team could use this as inspiration for the official fix:
https://www.reddit.com/r/yuzu/comments/gu0vxt/why_do_daily_events_not_work_its_as_if_time/fsh5jas/

takeruadelbert:
I really hope that the team would fix this.

KostaSaizo7:
This stiil happens in Yuzu mainline 417

ackile:
Nop, this is not fixed. All of my Tree berry are empty, and i can't fight in the battle pub.

RiedleroD:
still an issue, pokejobs don't advance, even with custom RTC. Didn't try the patch bc it's work I don't want to do rn.

G-Spawn:
> still an issue, pokejobs don't advance, even with custom RTC (edit: when set while the game isn't playing). Didn't try the patch bc it's work I don't want to do rn.

You're supposed to change custom rtc while the game is running, not when it isn't.

RiedleroD:
yes, I got that now. Still very annoying though.

FedericoValeri:
> still an issue, pokejobs don't advance, even with custom RTC (edit: when set while the game isn't playing). Didn't try the patch bc it's work I don't want to do rn.

Any news?

G-Spawn:
The news is custom rtc still works. Besides that there is nothing. @FedericoValeri

Fluffisaurus:
bump, also are there any plans to fix this issue? 🙏

G-Spawn:
Yes they plan on fixing it in future. No ETA.

KostaSaizo7:
This got fixed in Ryujinx btw https://github.com/Ryujinx/Ryujinx/pull/4822

german77:
Fixed in latest mainline

PR #4792, opened 2020-10-15
service: time: Update current time with changes to RTC setting.

bunnei:
This refreshes the time service with changes to the RTC set in the UI. This can be used to advance time, e.g. for Pokemon Sword/Shield pokejobs.

### DIFF
--- a/src/core/hle/service/time/time_zone_content_manager.h
+++ b/src/core/hle/service/time/time_zone_content_manager.h
@@ -21,7 +21,9 @@ namespace Service::Time::TimeZone {
 
 class TimeZoneContentManager final {
 public:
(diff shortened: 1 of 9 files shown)


src/core/hle/service/time/time_zone_content_manager.h
```diff
@@ -21,7 +21,9 @@ namespace Service::Time::TimeZone {
 
 class TimeZoneContentManager final {
 public:
-    TimeZoneContentManager(TimeManager& time_manager, Core::System& system);
+    TimeZoneContentManager(Core::System& system);
```
lioncash, line 24:
```suggestion
    explicit TimeZoneContentManager(Core::System& system);
```

KostaSaizo7:
Unfortunately, this doesn't fix #4074

PR #4797, opened 2020-10-16
service: bcat: Check client connection before interacting with socket.

bunnei:
- Fixes a crash when BCAT service is offline.


src/core/hle/service/bcat/backend/boxcat.cpp
```diff
@@ -454,6 +454,16 @@ Boxcat::StatusResult Boxcat::GetStatus(std::optional<std::string>& global,
         {std::string("Boxcat-Client-Type"), std::string(BOXCAT_CLIENT_TYPE)},
     };
 
+    if (!client.is_valid()) {
+        LOG_CRITICAL(Service_BCAT, "Client is invalid, going offline!");
```
lioncash, line 458: This should likely be an error, given this doesn't threaten the stability of the application.

LGTM otherwise

PR #4799, opened 2020-10-17
core: Refactor loader and implement ExecuteProgram

bunnei:
This change refactors loader code to support booting an indexed program within the application container. For now, this supports booting a program NCA (by index) within an NSP or XCI file.

With this refactoring, we can implement IApplicationFunctions::ExecuteProgram within the AM service, which is used to boot another process. This is used by Super Mario 3D All-Stars to launch sub-titles (Super Mario 64, Super Mario Sunshine, etc.).

Since yuzu does not support multiple processes, we accomplish this by simply stopping the current process and launching the new process at the specified program index.

This allows Super Mario 3D All-Stars to mostly function (with the exception that the Mario games do not render, but that is unrelated to this change). Once we fix rendering, this should work.

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -1188,9 +1188,9 @@ IApplicationFunctions::IApplicationFunctions(Core::System& system_)
         {102, &IApplicationFunctions::SetApplicationCopyrightVisibility, "SetApplicationCopyrightVisibility"},
         {110, &IApplicationFunctions::QueryApplicationPlayStatistics, "QueryApplicationPlayStatistics"},
         {111, &IApplicationFunctions::QueryApplicationPlayStatisticsByUid, "QueryApplicationPlayStatisticsByUid"},
-        {120, nullptr, "ExecuteProgram"},
-        {121, nullptr, "ClearUserChannel"},
-        {122, nullptr, "UnpopToUserChannel"},
+        {120, &IApplicationFunctions::ExecuteProgram, "ExecuteProgram"},
+        {121, &IApplicationFunctions::ClearUserChannel, "ClearUserChannel"},
+        {122, &IApplicationFunctions::UnpopToUserChannel, "UnpopToUserChannel"},
         {123, &IApplicationFunctions::GetPreviousProgramIndex, "GetPreviousProgramIndex"},
         {124, nullptr, "EnableApplicationAllThreadDumpOnCrash"},
         {130, &IApplicationFunctions::GetGpuErrorDetectedSystemEvent, "GetGpuErrorDetectedSystemEvent"},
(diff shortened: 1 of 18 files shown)


src/yuzu_cmd/emu_window/emu_window_sdl2.h
```diff
@@ -23,15 +23,15 @@ class EmuWindow_SDL2 : public Core::Frontend::EmuWindow {
...
-    /// Polls window events
     void PollEvents() override;
 
     /// Whether the window is still open, and a close request hasn't yet been sent
     bool IsOpen() const;
 
-    /// Returns if window is shown (not minimized)
     bool IsShown() const override;
 
+    void ExecuteProgram(std::size_t /*program_index*/) override{};
```
lioncash, line 33:
```suggestion
    void ExecuteProgram(std::size_t /*program_index*/) override {}
```

src/yuzu_tester/emu_window/emu_window_sdl2_hide.h
```diff
@@ -17,14 +17,14 @@ class EmuWindow_SDL2_Hide : public Core::Frontend::EmuWindow {
...
 
-    /// Polls window events
     void PollEvents() override;
 
-    /// Whether the screen is being shown or not.
     bool IsShown() const override;
 
     std::unique_ptr<Core::Frontend::GraphicsContext> CreateSharedContext() const override;
 
+    void ExecuteProgram(std::size_t /*program_index*/) override{};
```
lioncash, line 26:
```suggestion
    void ExecuteProgram(std::size_t /*program_index*/) override {}
```

src/core/hle/service/am/am.cpp
```diff
@@ -1554,6 +1559,36 @@ void IApplicationFunctions::QueryApplicationPlayStatisticsByUid(Kernel::HLEReque
...
+    [[maybe_unused]] const auto unk_1 = rp.Pop<u32>();
+    [[maybe_unused]] const auto unk_2 = rp.Pop<u32>();
+    const auto program_index = rp.Pop<u64>();
+
+    IPC::ResponseBuilder rb{ctx, 2};
+    rb.Push(RESULT_SUCCESS);
+
+    // TODO(bunnei): It's kind of janky to go through the renderer to reload the program, but this
+    // requires broad refactoring to fix.
+    system.GPU().Renderer().GetRenderWindow().ExecuteProgram(program_index);
```
lioncash, line 1575: I'm not sure why this needs to be going through the renderer. Would it not be more straightforward to allow the frontend to register a callback with the system, which then gets called here?

I really don't think we should be coupling the video core ro the core library like this, even if it is a "fast and quick" way to get it working. We have several TODOs similar to this still in core that never got addressed, and I don't think adding another one like it to the mix is the way to go. If refactoring needs to be done to suit the design better, I think it should just be done and gotten over with.

lioncash (changes requested):

src/core/hle/service/am/am.cpp
bunnei: Yep -- this PR was borderline "RFC" because I was not sure how I wanted to address this problem. We'll certainly do that before merging.

src/yuzu_cmd/emu_window/emu_window_sdl2.h
bunnei: Against our clang-format rules

src/yuzu_tester/emu_window/emu_window_sdl2_hide.h
bunnei: Against our clang-format rules

lioncash (approved):

PR #4805, opened 2020-10-20
yuzu: settings: Enable multicore, asynch GPU, and assembly shaders by default.

bunnei:
- In general, this is now the preferred settings for most games.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -515,7 +515,7 @@ void Config::ReadMotionTouchValues() {
 void Config::ReadCoreValues() {
     qt_config->beginGroup(QStringLiteral("Core"));
 
-    ReadSettingGlobal(Settings::values.use_multi_core, QStringLiteral("use_multi_core"), false);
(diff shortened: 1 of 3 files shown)


v1993:
Looks like a good change. Only thing I'd like to note is that in my tests (OpenGL, Nvidia, Linux) assembly shaders allow slightly faster compilation but lower FPS compared to GLSL.

ReinUsesLisp:
While assembly shaders might generate slower code than the GLSL equivalent, in my opinion the overall user experience is better in GLASM. Users that experience better results with GLSL can always change the setting manually.
As a side note, GLASM allows us to use precise on legacy drivers (Fermi) without causing internal compiler errors.

pacproduct:
I do not know if it's because of this PR specifically, but performances on Luigi's Mansion 3 skyrocketed today. This game now runs more or less at normal speed on my environment, amazing. Was still almost unplayable yesterday as it was running at 30~60% speed yesterday! 👍

v1993:
Does this also change settings on existing installations? Unlike luckier people, I had to "enjoy" 7-10 FPS and minor graphical glitches in games until giving configuration window a visit and turning assembly shaders off.

PR #4806, opened 2020-10-20
frontend: qt: Disable web browser applet implementation.

bunnei:
- Despite being partially working, this has no functional value and causes more user issues.
- Disable for now, while we rework the original implementation.


lat9nq:
Causes a regression in SSBU
Trying to load the regular Smash mode without any initial save results in a black screen (0 fps)

This behavior is notably different from simply not compiling the web browser into yuzu (simply configuring yuzu with `cmake ..`). You would get a warning dialog about not having the browser compiled in, then the game continues normally.

Here there is no dialog, but there is also no continuation of the game.

TONYSOAD05:
Hello, for me it was working well the web applet on SSBU, Mario Odyssey, pokemon shield and MK8D. But now, if I try to use an option that uses the web applet the games crashes, could you please re-enable the web applet application? Thank you!

arcadee1977:
Very unfortunate update, hope to recover it please !

bunnei:
@lat9nq @TONYSOAD05 should be fixed in latest iteration

PR #4796, opened 2020-10-15
core: Fix clang build

lioncash:
Recent changes to the build system that made more warnings be flagged as errors caused building via clang to break.

This should fix #4795

PR #4812, opened 2020-10-21
core/CMakeLists: Enable more warnings as errors on MSVC

lioncash:
Now that all sign conversion scenarios are resolved, we can enable them as errors for MSVC as well, so any introduced violations can be caught before making a PR and waiting for the buildbot to point them out.


lioncash:
Closing, due to issues with #4796

PR #4816, opened 2020-10-21
sdl_impl: Fix controller reconnection issues

### LINKED ISSUE #4268: Yuzu doesn't react to button pressing after disconnecting and connecting DS4 again during gaming

### LINKED ISSUE #4815: Issue with reconnecting disconnected controller
This was found using Early Access 1047.
I'm using a wireless Xbox One S controller and if I disconnect the controller and reconnect it, I can no longer control the running game. I have found two ways to get around this.

1. Click configure and then click OK.

2. Click configure current game. No need to click OK here to make the controller work again.
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5414729/yuzu_log.txt)

Morph1984:
It turns out that after a controller is disconnected, there is a chance that events from the previous controller are sent/processed after it has been disconnected.
This causes the previously disconnected controller to reappear as connected due to GetSDLJoystickBySDLID() emplacing this controller back to the vector/map.

Fix this by only returning an SDLJoystick if and only if it exists in the map.

This also reverts the changes from #4627 as they are unnecessary after this change.

Fixes #4815 and properly fixes #4268

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -155,13 +155,13 @@ public:
         return sdl_joystick.get();
     }
 
-    void SetSDLJoystick(SDL_Joystick* joystick, SDL_GameController* controller) {
-        sdl_controller.reset(controller);
-        sdl_joystick.reset(joystick);
-    }
-
     SDL_GameController* GetSDLGameController() const {
         return sdl_controller.get();
+    }
+
+    void SetSDLJoystick(SDL_Joystick* joystick, SDL_GameController* controller) {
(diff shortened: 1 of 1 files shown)


v1993:
I can confirm that this finally fixes controller reconnection for me.

lioncash (approved):

PR #4817, opened 2020-10-21
main/profile_select: Don't ask for profile when there's only one.

Kewlan:
~Based on #4297.~

Reverts #4297 and moves the profile selection skip to QtProfileSelectionDialog. Now it skips in both 
`GMainWindow::ProfileSelectorSelectProfile` and `GMainWindow::OnGameListOpenFolder` when there's only one profile.


FearlessTobi:
Should we maybe try to migrate this logic to QtProfileSelectionDialog if we are going to do this in multiple places?
I'm not sure myself if that would be the best course of action, but it could be worth a thought.

FearlessTobi (approved):

bunnei (approved):

bunnei:
LGTM, thanks!

PR #4837, opened 2020-10-27
nvdec: Minor tidying up

lioncash:
Cleans up a few things that I noticed after it being merged.


ameerj (approved):
Thanks for catching these!

bunnei (approved):

PR #4842, opened 2020-10-27
web_backend: fix a regression introduced in 39c8d18

liushuyu:
* A regression was in 39c8d18 and the token verification function was broken.
* The reason being `httplib` now requires OpenSSL 1.1+ API while LibreSSL 2.x provided OpenSSL 1.0 compatible API.
* The bundled LibreSSL has been updated to 3.2.2 so it now provides OpenSSL 1.1 compatible API now.
* Also the path hint has been added so that it will find the correct path to the CA certs on *nix systems.
* An option is provided so that *nix system distributions/providers can use their own SSL implementations when compiling Yuzu/Citra to (hopefully) complies with their maintenance guidelines.
* LURLParse is also removed since `httplib` can handle `scheme:host:port` string itself now.

---
This change is [Reviewable](https://reviewable.io/reviews/yuzu-emu/yuzu/4842)


CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ option(ENABLE_VULKAN "Enables Vulkan backend" ON)
 
 option(USE_DISCORD_PRESENCE "Enables Discord Rich Presence" OFF)
 
+option(USE_SYSTEM_SSL "Use system provided SSL library" OFF)
```
lioncash, line 33: This should probably be removed and replaced with a [find_package(OpenSSL)](https://cmake.org/cmake/help/v3.19/module/FindOpenSSL.html) call and uses the found library if it was discovered on the system automatically, otherwise it falls back to externals, like we do with several other libraries.

lioncash (approved):

PR #4847, opened 2020-10-28
video_core: Enforce -Wredundant-move and -Wpessimizing-move

ReinUsesLisp:
Silence three warnings and make them errors to avoid introducing more in the future.


src/video_core/CMakeLists.txt
```diff
@@ -304,6 +304,8 @@ else()
         -Wno-error=sign-conversion
         -Werror=switch
         -Werror=unused-variable
+        -Werror=redundant-move
```
lioncash, line 307: These should be alphabetically ordered with the other error flags.

lioncash (approved):

PR #4848, opened 2020-10-28
video_core: Enforce -Werror=type-limits

ReinUsesLisp:
Silences one warning and avoids introducing more in the future.
`base_stream.GetPosition()` returns an unsigned type, so it will always be greater than or equal to zero.

### DIFF
--- a/src/video_core/CMakeLists.txt
+++ b/src/video_core/CMakeLists.txt
@@ -303,6 +303,7 @@ else()
         -Werror=conversion
         -Wno-error=sign-conversion
         -Werror=switch
+        -Werror=type-limits
         -Werror=unused-variable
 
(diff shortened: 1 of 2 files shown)


src/video_core/CMakeLists.txt
```diff
@@ -304,6 +304,7 @@ else()
         -Wno-error=sign-conversion
         -Werror=switch
         -Werror=unused-variable
+        -Werror=type-limits
```
lioncash, line 307: This should go above unused-variable, so it's alphabetically organized.

ReinUsesLisp: Addressed, although GitHub doesn't seem to catch the change.

lioncash (approved):

Issue #4580, opened 2020-08-25
Problems Operating Analog Stick with Keyboard (Diablo 3)


Samsik2:
Analog sticks in the Diablo 3 console board use a method called 14 axis rather than a typical 8 axis.

![114](https://user-images.githubusercontent.com/70080561/91158482-a6582700-e701-11ea-8aa4-a4acc712afb7.png)

So I can't select the equipment marked in yellow.

The same problem occurred in RPCS3, but the link below seems to have been resolved.

https://github.com/RPCS3/rpcs3/pull/5468
Add mouse movement and stick lerp settings

If YUZU can implement these functions, we would like you to add them.

german77:
duplicate of #2845

german77:
Fixed by #4905

Issue #2845, opened 2019-09-06
Left and Right analog keyboard circle approximation issue


Fueramanta:
Hi everyone, this used to be an issue when playing diablo 3 in rpcs3 just with keyboard, but was fixed, I was wondering if the same could be done with Yuzu. I have tried many ways to do it myself, such editing the qt-config file, and downloading Vjoy, but to no success at all.

The left analog is approximated by wasd keys in keyboard, while we can use two adjacent keys to do 45 degree of the analog, it could not do 22.5 degree, this makes 22.5 degree analog unemulatable at present.
As a test case, we can take Diablo 3>Inventory>select belt or armor
I dont know how nest could it be resolved, but I have 1 idea in mind.Mabye implement 1 key that emulates analog rotation, could be 1degree/200ms for counter closewise rotation then another key for clockwise rotation.

It would awesome if it worked the same as in rpcs3, as long as it is not too difficult to do it.

Thanks a lot.

Fueramanta:
Isn’t there anyway to fix this?, I know I could use a controller, but can’t buy one for personal reasons. I’d really appreciate it if anybody gives me an answer.

Thanks again

german77:
Fixed by #4905

PR #4905, opened 2020-11-07
Allow to dial any angle with digital joystick

german77:
Instead of returning only the cardinal angles with digital buttons allow the input to output any angle by drifting the angle slowly to the desired direction. Solving #4580, #2845. A solution with mouse will also be implemented in a future PR.

The speed of the drift can be adjusted in the config.ini. If necessary I can add the setting to the UI but as far my testing with the game I have needs no adjustment.

### DIFF
--- a/src/input_common/analog_from_button.cpp
+++ b/src/input_common/analog_from_button.cpp
@@ -46,7 +46,7 @@ public:
                 angle += TAU;
             }
         } else if ((angle >= bottom_limit && angle < to_angle) ||
-                    (angle - TAU >= bottom_limit && angle - TAU < to_angle)) {
+                   (angle - TAU >= bottom_limit && angle - TAU < to_angle)) {
             angle += modifier_angle;
             if (angle >= TAU) {
                 angle -= TAU;
--- a/src/input_common/analog_from_button.cpp
+++ b/src/input_common/analog_from_button.cpp
@@ -2,6 +2,10 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <chrono>
+#include <cmath>
+#include <thread>
+#include "common/math_util.h"
 #include "input_common/analog_from_button.h"
 
 namespace InputCommon {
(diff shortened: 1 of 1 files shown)


src/input_common/analog_from_button.cpp
```diff
@@ -59,6 +134,11 @@ class Analog final : public Input::AnalogDevice {
     Button right;
     Button modifier;
     float modifier_scale;
+    float modifier_angle;
+    float angle = 0;
+    float amplitude = 0;
+    std::thread update_thread;
+    bool update_thread_running = true;
```
Morph1984, line 141: Braced initializers
```suggestion
    float angle{};
    float amplitude{};
    std::thread update_thread;
    bool update_thread_running{true};
```

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +14,103 @@ class Analog final : public Input::AnalogDevice {
...
         }
-        if (left->GetStatus()) {
-            --x;
-        }
-        if (up->GetStatus()) {
-            ++y;
+    }
+
+    void MoveToDirection(bool enable, float to_angle) {
+        constexpr float TAU = 3.1415926f * 2.0f;
```
Morph1984, line 32: make use of the Common::PI constant in `common/math_util.h`
```suggestion
        constexpr float TAU = Common::PI * 2.0f;
```

src/input_common/analog_from_button.cpp
```diff
@@ -59,6 +134,11 @@ class Analog final : public Input::AnalogDevice {
     Button right;
     Button modifier;
     float modifier_scale;
+    float modifier_angle;
+    float angle{0};
+    float amplitude{0};
```
Morph1984, line 139: The default braced initialization for `float example_float{}` is 0.0f

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +14,101 @@ class Analog final : public Input::AnalogDevice {
...
+            }
+            if (u && d) {
+                u = false;
+                d = false;
+            }
+
+            // Move to the right
+            MoveToDirection(r & !u & !d, 0.0f);
+
+            // Move to the right up
```
lioncash, line 79:
```suggestion
            // Move to the upper right
```

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +14,101 @@ class Analog final : public Input::AnalogDevice {
...
+            MoveToDirection(r & u & !d, Common::PI * 0.25f);
+
+            // Move up
+            MoveToDirection(u & !l & !r, Common::PI * 0.5f);
+
+            // Move to the left up
```
lioncash, line 85:
```suggestion
            // Move to the upper left
```

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +14,101 @@ class Analog final : public Input::AnalogDevice {
...
+            MoveToDirection(l & u & !d, Common::PI * 0.75f);
+
+            // Move to the left
+            MoveToDirection(l & !u & !d, Common::PI);
+
+            // Move to the left down
```
lioncash, line 91:
```suggestion
            // Move to the bottom left
```

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +14,101 @@ class Analog final : public Input::AnalogDevice {
...
+            MoveToDirection(l & !u & d, Common::PI * 1.25f);
+
+            // Move down
+            MoveToDirection(d & !l & !r, Common::PI * 1.5f);
+
+            // Move to the right down
```
lioncash, line 97:
```suggestion
            // Move to the bottom right
```

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +14,101 @@ class Analog final : public Input::AnalogDevice {
...
+                r = false;
+                l = false;
+            }
+            if (u && d) {
+                u = false;
+                d = false;
+            }
+
+            // Move to the right
+            MoveToDirection(r & !u & !d, 0.0f);
```
lioncash, line 77: Note that using bitwise arithmetic on booleans can cause warnings and also be flagged by static analysis tooling. You may want to consider using the logical operators instead.

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +14,101 @@ class Analog final : public Input::AnalogDevice {
...
+            }
+            std::this_thread::sleep_for(std::chrono::milliseconds(10));
         }
+    }
 
-        const float coef = modifier->GetStatus() ? modifier_scale : 1.0f;
-        return std::make_tuple(static_cast<float>(x) * coef * (y == 0 ? 1.0f : SQRT_HALF),
-                               static_cast<float>(y) * coef * (x == 0 ? 1.0f : SQRT_HALF));
+    std::tuple<float, float> GetStatus() const override {
+        return std::make_tuple(cos(angle) * amplitude, sin(angle) * amplitude);
```
lioncash, line 111:
```suggestion
        return std::make_tuple(std::cos(angle) * amplitude, std::sin(angle) * amplitude);
```

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +14,101 @@ class Analog final : public Input::AnalogDevice {
...
+            // Move to the right down
+            MoveToDirection(r & !u & d, Common::PI * 1.75f);
+
+            // Move if a key is pressed
+            if (r | l | u | d) {
+                amplitude = coef;
+            } else {
+                amplitude = 0;
+            }
+            std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
lioncash, line 106: There should probably be a comment here indicating why the sleep is necessary.

src/input_common/analog_from_button.cpp
Morph1984: Would a `<cmath>` include be needed as well?

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +15,103 @@ class Analog final : public Input::AnalogDevice {
...
-        }
-        if (up->GetStatus()) {
-            ++y;
+    }
+
+    void MoveToDirection(bool enable, float to_angle) {
+        constexpr float TAU = Common::PI * 2.0f;
+        // Use wider angle to ease the transition.
+        constexpr float aperture = TAU * 0.15f;
+        float top_limit = to_angle + aperture;
```
lioncash, line 36: Missed in the initial review, but `top_limit` and `bottom_limit` can be const.

src/input_common/analog_from_button.cpp
```diff
@@ -11,31 +15,103 @@ class Analog final : public Input::AnalogDevice {
...
+        float bottom_limit = to_angle - aperture;
+
+        if (enable) {
```
lioncash, line 39:
```suggestion
        if (!enable) {
            return;
        }
```

Can likely be placed at the top of the function, unindenting all the contained code here.

lioncash (commented):
Just two things I missed in the previous review. Other than those it looks good

bunnei (approved):

PR #4907, opened 2020-11-08
core: Make nvservices more standardized

ogniK5377:
The most notable changes are as follows;
* Ioctl1, Ioctl2 & Ioctl3 are completely separated now. This allows us to track unimplemented Ioctls across versions easier.
* Delayed Ioctls are removed. They're not accurate and are not needed.
* All Ioctls share the same result codes now instead of redefining results all the time.
* Ioctls are now set up to use their groups & cmd numbers. This cuts down the code for nvdec & is closer to what the service does on the switch.
* GetVARegions for Ioctl3 is now implemented(this was not implemented before)
* All structs have their members initialized now

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvdevice.h
+++ b/src/core/hle/service/nvdrv/devices/nvdevice.h
@@ -24,10 +24,35 @@ public:
     explicit nvdevice(Core::System& system) : system{system} {}
     virtual ~nvdevice() = default;
 
+    /**
+     * Handles an ioctl1 request.
+     * @param command The ioctl command id.
+     * @param input A buffer containing the input data for the ioctl.
+     * @param output A buffer where the output data will be written to.
+     * @returns The result code of the ioctl.
+     */
     virtual NvResult Ioctl1(Ioctl command, const std::vector<u8>& input,
                             std::vector<u8>& output) = 0;
+
+    /**
+     * Handles an ioctl2 request.
+     * @param command The ioctl command id.
+     * @param input A buffer containing the input data for the ioctl.
+     * @param inline_input A buffer containing the input data for the ioctl which has been inlined.
+     * @param output A buffer where the output data will be written to.
+     * @returns The result code of the ioctl.
+     */
     virtual NvResult Ioctl2(Ioctl command, const std::vector<u8>& input,
                             const std::vector<u8>& inline_input, std::vector<u8>& output) = 0;
+
+    /**
+     * Handles an ioctl3 request.
+     * @param command The ioctl command id.
+     * @param input A buffer containing the input data for the ioctl.
+     * @param output A buffer where the output data will be written to.
+     * @param inline_output A buffer where the inlined output data will be written to.
+     * @returns The result code of the ioctl.
+     */
     virtual NvResult Ioctl3(Ioctl command, const std::vector<u8>& input, std::vector<u8>& output,
                             std::vector<u8>& inline_output) = 0;
 
--- a/src/core/hle/service/nvdrv/devices/nvdevice.h
+++ b/src/core/hle/service/nvdrv/devices/nvdevice.h
@@ -24,25 +24,12 @@ public:
     explicit nvdevice(Core::System& system) : system{system} {}
     virtual ~nvdevice() = default;
 
-    union Ioctl {
-        u32_le raw;
-        BitField<0, 8, u32> cmd;
-        BitField<8, 8, u32> group;
-        BitField<16, 14, u32> length;
-        BitField<30, 1, u32> is_in;
-        BitField<31, 1, u32> is_out;
-    };
-
-    /**
-     * Handles an ioctl request.
-     * @param command The ioctl command id.
-     * @param input A buffer containing the input data for the ioctl.
-     * @param output A buffer where the output data will be written to.
-     * @returns The result code of the ioctl.
-     */
-    virtual u32 ioctl(Ioctl command, const std::vector<u8>& input, const std::vector<u8>& input2,
-                      std::vector<u8>& output, std::vector<u8>& output2, IoctlCtrl& ctrl,
-                      IoctlVersion version) = 0;
+    virtual NvResult Ioctl1(Ioctl command, const std::vector<u8>& input,
+                            std::vector<u8>& output) = 0;
+    virtual NvResult Ioctl2(Ioctl command, const std::vector<u8>& input,
+                            const std::vector<u8>& inline_input, std::vector<u8>& output) = 0;
+    virtual NvResult Ioctl3(Ioctl command, const std::vector<u8>& input, std::vector<u8>& output,
+                            std::vector<u8>& inline_output) = 0;
 
 protected:
     Core::System& system;
--- a/src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
@@ -17,59 +17,70 @@
 
 namespace Service::Nvidia::Devices {
 
-namespace NvErrCodes {
-constexpr u32 Success{};
-constexpr u32 OutOfMemory{static_cast<u32>(-12)};
-constexpr u32 InvalidInput{static_cast<u32>(-22)};
-} // namespace NvErrCodes
-
 nvhost_as_gpu::nvhost_as_gpu(Core::System& system, std::shared_ptr<nvmap> nvmap_dev)
     : nvdevice(system), nvmap_dev(std::move(nvmap_dev)) {}
 nvhost_as_gpu::~nvhost_as_gpu() = default;
 
-u32 nvhost_as_gpu::ioctl(Ioctl command, const std::vector<u8>& input, const std::vector<u8>& input2,
-                         std::vector<u8>& output, std::vector<u8>& output2, IoctlCtrl& ctrl,
-                         IoctlVersion version) {
-    LOG_DEBUG(Service_NVDRV, "called, command=0x{:08X}, input_size=0x{:X}, output_size=0x{:X}",
-              command.raw, input.size(), output.size());
-
-    switch (static_cast<IoctlCommand>(command.raw)) {
-    case IoctlCommand::IocInitalizeExCommand:
-        return InitalizeEx(input, output);
-    case IoctlCommand::IocAllocateSpaceCommand:
-        return AllocateSpace(input, output);
-    case IoctlCommand::IocMapBufferExCommand:
-        return MapBufferEx(input, output);
-    case IoctlCommand::IocBindChannelCommand:
-        return BindChannel(input, output);
-    case IoctlCommand::IocGetVaRegionsCommand:
-        return GetVARegions(input, output);
-    case IoctlCommand::IocUnmapBufferCommand:
-        return UnmapBuffer(input, output);
-    case IoctlCommand::IocFreeSpaceCommand:
-        return FreeSpace(input, output);
+NvResult nvhost_as_gpu::Ioctl1(Ioctl command, const std::vector<u8>& input,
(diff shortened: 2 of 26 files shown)


src/core/hle/service/nvdrv/devices/nvdevice.h
```diff
@@ -24,25 +24,12 @@ class nvdevice {
...
-     * Handles an ioctl request.
-     * @param command The ioctl command id.
-     * @param input A buffer containing the input data for the ioctl.
-     * @param output A buffer where the output data will be written to.
-     * @returns The result code of the ioctl.
-     */
-    virtual u32 ioctl(Ioctl command, const std::vector<u8>& input, const std::vector<u8>& input2,
-                      std::vector<u8>& output, std::vector<u8>& output2, IoctlCtrl& ctrl,
-                      IoctlVersion version) = 0;
+    virtual NvResult Ioctl1(Ioctl command, const std::vector<u8>& input,
```
lioncash, line 27: These three funcs are all missing documentation comments.

src/core/hle/service/nvdrv/devices/nvhost_nvdec_common.h
```diff
@@ -18,9 +18,12 @@ class nvhost_nvdec_common : public nvdevice {
     explicit nvhost_nvdec_common(Core::System& system, std::shared_ptr<nvmap> nvmap_dev);
     ~nvhost_nvdec_common() override;
 
-    virtual u32 ioctl(Ioctl command, const std::vector<u8>& input, const std::vector<u8>& input2,
-                      std::vector<u8>& output, std::vector<u8>& output2, IoctlCtrl& ctrl,
-                      IoctlVersion version) = 0;
+    virtual NvResult Ioctl1(Ioctl command, const std::vector<u8>& input,
```
lioncash, line 21: Missing documentation comments.

src/core/hle/service/nvdrv/devices/nvhost_nvjpg.cpp
```diff
@@ -13,28 +13,43 @@ namespace Service::Nvidia::Devices {
...
-                        IoctlVersion version) {
-    LOG_DEBUG(Service_NVDRV, "called, command=0x{:08X}, input_size=0x{:X}, output_size=0x{:X}",
-              command.raw, input.size(), output.size());
-
-    switch (static_cast<IoctlCommand>(command.raw)) {
-    case IoctlCommand::IocSetNVMAPfdCommand:
-        return SetNVMAPfd(input, output);
+NvResult nvhost_nvjpg::Ioctl1(Ioctl command, const std::vector<u8>& input,
+                              std::vector<u8>& output) {
+
```
lioncash, line 18: Unnecessary newline.

src/core/hle/service/nvdrv/devices/nvhost_vic.cpp
```diff
@@ -15,36 +15,43 @@ nvhost_vic::nvhost_vic(Core::System& system, std::shared_ptr<nvmap> nvmap_dev)
...
+            return Submit(input, output);
+        case 0x2:
+            return GetSyncpoint(input, output);
+        case 0x3:
+            return GetWaitbase(input, output);
+        case 0x9:
+            return MapBuffer(input, output);
+        case 0xa:
+            return UnmapBuffer(input, output);
+        }
```
lioncash, line 32: If fallthrough here is intentional, you should use `[[fallthrough]]`, otherwise a `break;` should be here.

src/core/hle/service/nvdrv/devices/nvmap.cpp
```diff
@@ -26,35 +19,48 @@ nvmap::nvmap(Core::System& system) : nvdevice(system) {
...
+            return IocFromId(input, output);
+        case 0x4:
+            return IocAlloc(input, output);
+        case 0x5:
+            return IocFree(input, output);
+        case 0x9:
+            return IocParam(input, output);
+        case 0xe:
+            return IocGetId(input, output);
+        }
```
lioncash, line 38: A `break;` should follow this line, even if there's only one case.

src/core/hle/service/nvdrv/interface.h
```diff
@@ -33,11 +33,13 @@ class NVDRV final : public ServiceFramework<NVDRV> {
...
     void GetStatus(Kernel::HLERequestContext& ctx);
     void DumpGraphicsMemoryInfo(Kernel::HLERequestContext& ctx);
-    void IoctlBase(Kernel::HLERequestContext& ctx, IoctlVersion version);
+
+    void ServiceError(Kernel::HLERequestContext& ctx, NvResult result);
 
     std::shared_ptr<Module> nvdrv;
 
     u64 pid{};
+    bool initialized{};
```
lioncash, line 42:
```suggestion
    bool is_initialized{};
```

src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -17,59 +17,70 @@
...
 
-u32 nvhost_as_gpu::InitalizeEx(const std::vector<u8>& input, std::vector<u8>& output) {
+NvResult nvhost_as_gpu::Ioctl3(Ioctl command, const std::vector<u8>& input, std::vector<u8>& output,
+                               std::vector<u8>& inline_output) {
+    switch (command.group) {
+    case 'A':
+        switch (command.cmd) {
+        case 0x8:
+            return GetVARegions(input, output, inline_output);
+        }
```
lioncash, line 68: A `break;` should be after this line.

src/core/hle/service/nvdrv/nvdrv.cpp
```diff
@@ -61,36 +61,101 @@ Module::Module(Core::System& system) : syncpoint_manager{system.GPU()} {
...
+}
+
+NvResult Module::Ioctl3(DeviceFD fd, Ioctl command, const std::vector<u8>& input,
+                        std::vector<u8>& output, std::vector<u8>& inline_output) {
+    if (fd < 0) {
+        LOG_ERROR(Service_NVDRV, "Invalid DeviceFD={}!", fd);
+        return NvResult::InvalidState;
+    }
+
+    auto itr = open_files.find(fd);
```
lioncash, line 133:
```suggestion
    const auto itr = open_files.find(fd);
```

src/core/hle/service/nvdrv/nvdrv.cpp
```diff
@@ -61,36 +61,101 @@ Module::Module(Core::System& system) : syncpoint_manager{system.GPU()} {
...
+    return itr->second->Ioctl3(command, input, output, inline_output);
+}
+
+NvResult Module::Close(DeviceFD fd) {
+    if (fd < 0) {
+        LOG_ERROR(Service_NVDRV, "Invalid DeviceFD={}!", fd);
+        return NvResult::InvalidState;
+    }
+
+    auto itr = open_files.find(fd);
```
lioncash, line 149:
```suggestion
    const auto itr = open_files.find(fd);
```

src/core/hle/service/nvdrv/nvdrv.cpp
```diff
@@ -61,36 +61,101 @@ Module::Module(Core::System& system) : syncpoint_manager{system.GPU()} {
 
 Module::~Module() = default;
 
-u32 Module::Open(const std::string& device_name) {
-    ASSERT_MSG(devices.find(device_name) != devices.end(), "Trying to open unknown device {}",
-               device_name);
+NvResult Module::VerifyFd(DeviceFD fd) const {
```
lioncash, line 64:
```suggestion
NvResult Module::VerifyFD(DeviceFD fd) const {
```

src/core/hle/service/nvdrv/interface.cpp
```diff
@@ -23,124 +23,171 @@ void NVDRV::SignalGPUInterruptSyncpt(const u32 syncpoint_id, const u32 value) {
...
 
-    IPC::ResponseBuilder rb{ctx, 3, 1};
-    rb.Push(RESULT_SUCCESS);
+    if (!initialized) {
+        ServiceError(ctx, NvResult::NotInitialized);
+        LOG_ERROR(Service_NVDRV, "NvServices is not initalized!");
+        return;
+    }
+
+    auto nv_result = nvdrv->VerifyFd(fd);
```
lioncash, line 172:
```suggestion
    const auto nv_result = nvdrv->VerifyFd(fd);
```

bylaws (approved):
At least from an emulation perspective this is a lot better :)

src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -17,59 +17,70 @@
...
-    case IoctlCommand::IocUnmapBufferCommand:
-        return UnmapBuffer(input, output);
-    case IoctlCommand::IocFreeSpaceCommand:
-        return FreeSpace(input, output);
+NvResult nvhost_as_gpu::Ioctl1(Ioctl command, const std::vector<u8>& input,
+                               std::vector<u8>& output) {
+    switch (command.group) {
+    case 'A':
+        switch (command.cmd) {
+        case 0x1:
```
bunnei, line 29: These should still be named enum values

Thog:
Hello,

It should be noted here that the way ioctl distribution is done is literally the same as the one I made for [Ryujinx](https://github.com/Ryujinx/Ryujinx/pull/800) last year and falls under the term of the MIT license.

This was written that way to abuse ReadOnlySpan optimization on C# JIT, This is totally inefficient in C++ and could be handled in a far better way.

On the subject of ioctl decoding itself, I have chosen not to handle ioctl the same way nvservices does, to reduce human errors as much as possible.  As such, I use switch statements in my implementation because there is no efficient way to enforce type restriction on ioctl numbers in a generic way in C#.

WIth all that said, I can confidently say that it doesn't make sense that you would come up with the same design in C++, and require proper attribution of the Ryujinx project under the terms of the MIT license.

Best regards,

Thog

ogniK5377:
> Hello,
> 
> It should be noted here that the way ioctl distribution is done is literally the same as the one I made for [Ryujinx](https://github.com/Ryujinx/Ryujinx/pull/800) last year and falls under the terms of the MIT license.
> 
> This was written that way to abuse ReadOnlySpan optimization on C# JIT, This is totally inefficient in C++ and could be handled in a far better way.
> 
> On the subject of ioctl decoding itself, I have chosen not to handle ioctl the same way nvservices does, to reduce human errors as much as possible. As such, I use switch statements in my implementation because there is no efficient way to enforce type restriction on ioctl numbers in a generic way in C#.
> 
> With all that said, I can confidently say that it doesn't make sense that you would come up with the same design in C++, and require proper attribution of the Ryujinx project under the terms of the MIT license.
> 
> Best regards,
> 
> Thog

As far as I can see it doesn’t fall under anything from Ryujinx. The reason the ioctls were structured in this way was to force everything to use jump tables and keep everything standardized. On top of the standardization, it leads to more optimized code as C++ compiler is way better at handling switch cases. I even ended up writing tooling to unpack ioctls to give me their groups and cmd numbers. If I referenced Ryujinx I wouldn’t have bothered. In the spirit of open source here’s the tooling used: https://gist.github.com/ogniK5377/11ec04da6a6715399f0d12271b59e5d6

Yuzu and Ryujinx don’t even implement ioctls the same way. I have no idea why you think ioctl groups are some magic way of thinking when everything is on switchbrew. As I said in the beginning, I don’t believe this requires attribution.

ogniK5377:
In regards to your concerns about performance, refer to [here](https://godbolt.org/z/57xjTj) for asm output to see old vs new method.

bunnei:
@ogniK5377 errors look real -- Can you fix/rebase so we can re-tag this? Thanks!

bunnei (approved):

PR #4914, opened 2020-11-10
bootmanager: Log and show GL_RENDERER string when GPU is insufficient

lat9nq:
Changes the first message to not include the OpenGL version, as the error is caused by OpenGL failing to load and nothing to do with the version.

Adds a new check for OpenGL version 4.3. This will display a message with a similar error as well as the GL_RENDERER string. Adds a ERROR log message when triggered. This prevents a crash with yuzu trying to use older OpenGL versions.

Modifies the unsupported extension message to output the GL_RENDERER string in the message, as well as logging the string.

### DIFF
--- a/src/yuzu/bootmanager.cpp
+++ b/src/yuzu/bootmanager.cpp
@@ -10,6 +10,7 @@
 #include <QMessageBox>
 #include <QPainter>
 #include <QScreen>
+#include <QString>
 #include <QStringList>
 #include <QWindow>
 
@@ -603,19 +604,34 @@ bool GRenderWindow::LoadOpenGL() {
     auto context = CreateSharedContext();
     auto scope = context->Acquire();
     if (!gladLoadGL()) {
+        QMessageBox::critical(
+            this, tr("Error while initializing OpenGL!"),
+            tr("Your GPU may not support OpenGL, or you do not have the latest graphics driver."));
+        return false;
+    }
+
+    QString renderer = QString::fromUtf8(reinterpret_cast<const char*>(glGetString(GL_RENDERER)));
+
+    if (!GLAD_GL_VERSION_4_3) {
+        LOG_CRITICAL(Frontend, "GPU does not support OpenGL 4.3: {:s}", renderer.toStdString());
         QMessageBox::critical(this, tr("Error while initializing OpenGL 4.3!"),
                               tr("Your GPU may not support OpenGL 4.3, or you do not have the "
-                                 "latest graphics driver."));
+                                 "latest graphics driver.<br><br>GL Renderer:<br>%1")
+                                  .arg(renderer));
         return false;
     }
 
     QStringList unsupported_gl_extensions = GetUnsupportedGLExtensions();
     if (!unsupported_gl_extensions.empty()) {
+        LOG_CRITICAL(Frontend, "GPU does not support all needed extensions: {:s}",
+                     renderer.toStdString());
         QMessageBox::critical(
             this, tr("Error while initializing OpenGL!"),
             tr("Your GPU may not support one or more required OpenGL extensions. Please ensure you "
-               "have the latest graphics driver.<br><br>Unsupported extensions:<br>") +
-                unsupported_gl_extensions.join(QStringLiteral("<br>")));
+               "have the latest graphics driver.<br><br>GL Renderer:<br>%1<br><br>Unsupported "
+               "extensions:<br>%2")
+                .arg(renderer)
+                .arg(unsupported_gl_extensions.join(QStringLiteral("<br>"))));
         return false;
     }
     return true;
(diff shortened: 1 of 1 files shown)


src/yuzu/bootmanager.cpp
```diff
@@ -603,19 +604,34 @@ bool GRenderWindow::LoadOpenGL() {
     auto context = CreateSharedContext();
     auto scope = context->Acquire();
     if (!gladLoadGL()) {
+        QMessageBox::critical(
```
lioncash, line 607:
```suggestion
        QMessageBox::warning(
```

Since this is an error that is safely handled and doesn't threaten the stability of the emulator.

src/yuzu/bootmanager.cpp
```diff
@@ -603,19 +604,34 @@ bool GRenderWindow::LoadOpenGL() {
...
-                                 "latest graphics driver."));
+                                 "latest graphics driver.<br><br>GL Renderer:<br>%1")
+                                  .arg(renderer));
         return false;
     }
 
     QStringList unsupported_gl_extensions = GetUnsupportedGLExtensions();
     if (!unsupported_gl_extensions.empty()) {
+        LOG_CRITICAL(Frontend, "GPU does not support all needed extensions: {:s}",
+                     renderer.toStdString());
```
ameerj, line 627: I think it won't hurt to also log `unsupported_gl_extensions`

src/yuzu/bootmanager.cpp
```diff
@@ -603,19 +604,34 @@ bool GRenderWindow::LoadOpenGL() {
...
+        QMessageBox::critical(
+            this, tr("Error while initializing OpenGL!"),
+            tr("Your GPU may not support OpenGL, or you do not have the latest graphics driver."));
+        return false;
+    }
+
+    QString renderer = QString::fromUtf8(reinterpret_cast<const char*>(glGetString(GL_RENDERER)));
+
+    if (!GLAD_GL_VERSION_4_3) {
+        LOG_CRITICAL(Frontend, "GPU does not support OpenGL 4.3: {:s}", renderer.toStdString());
```
lioncash, line 616:
```suggestion
        LOG_WARNING(Frontend, "GPU does not support OpenGL 4.3: {}", renderer.toStdString());
```

src/yuzu/bootmanager.cpp
```diff
@@ -603,19 +604,34 @@ bool GRenderWindow::LoadOpenGL() {
...
                               tr("Your GPU may not support OpenGL 4.3, or you do not have the "
-                                 "latest graphics driver."));
+                                 "latest graphics driver.<br><br>GL Renderer:<br>%1")
+                                  .arg(renderer));
         return false;
     }
 
     QStringList unsupported_gl_extensions = GetUnsupportedGLExtensions();
     if (!unsupported_gl_extensions.empty()) {
+        LOG_CRITICAL(Frontend, "GPU does not support all needed extensions: {:s}",
```
lioncash, line 626:
```suggestion
        LOG_WARNING(Frontend, "GPU does not support all needed extensions: {}",
```

src/yuzu/bootmanager.cpp
lat9nq: yuzu already logs unsupported extensions in `GRenderWindow::GetUnsupportedGLExtensions()`.

src/yuzu/bootmanager.cpp
```diff
@@ -603,19 +604,34 @@ bool GRenderWindow::LoadOpenGL() {
     auto context = CreateSharedContext();
     auto scope = context->Acquire();
     if (!gladLoadGL()) {
+        QMessageBox::critical(
+            this, tr("Error while initializing OpenGL!"),
+            tr("Your GPU may not support OpenGL, or you do not have the latest graphics driver."));
+        return false;
+    }
+
+    QString renderer = QString::fromUtf8(reinterpret_cast<const char*>(glGetString(GL_RENDERER)));
```
lioncash, line 613:
```suggestion
    const QString renderer = QString::fromUtf8(reinterpret_cast<const char*>(glGetString(GL_RENDERER)));
```

This should likely be moved closer to its usage as well.

src/yuzu/bootmanager.cpp
lat9nq: > Since this is an error that is safely handled and doesn't threaten the stability of the emulator.

In `GRenderWindow::GetUnsupportedGLExtensions()`, should I change that to `LOG_WARNING` as well?

src/yuzu/bootmanager.cpp
lat9nq: Can you clarify what you mean by "This should likely be moved closer to its usage as well"? I use `renderer` both when checking for version 4.3 as well as when unsupported extensions are checked.

lioncash: Ah, didn't see that usage of it. Never mind then.

src/yuzu/bootmanager.cpp
lioncash: I would use LOG_ERROR it if it needs to stand out. `LOG_CRITICAL` should be reserved for, "Your application is probably going to crash after this occurring" scenarios—basically anything that threatens the entire stability of the application.

goldenx86:
Fantastic, this will help with support. Thanks toast.

lioncash (approved):

PR #4930, opened 2020-11-16
cdma_pusher: Minor cleanup

lioncash:
Tidies up the implementation of the CDMA pusher a little bit.

Makes some member variables unsigned, as they were entirely used within unsigned contexts, or were always being converted into unsigned contexts implicitly.

We can also remove an unnecessary intermediate `std::vector` in `Step()`. Instead, we can opt to use the popped vector directly. We can just access the raw value instead of `std::memcpy`-ing the data into the intermediate vector as a `u32`.


lioncash:
Closing in favor of #4973

PR #4932, opened 2020-11-17
audren: Make use of nodiscard, rework downmixing, release all buffers

ogniK5377:
Preliminary work for upmixing & general cleanup. Fixes basic issues in games such as Shovel Knight and slightly improves the LEGO games. Upmixing stitll needs to be implemented.

Audio levels in a few games will be fixed as we now use the downmix coefficients when possible instead of supplying our own


src/audio_core/audio_renderer.cpp
```diff
@@ -14,6 +15,59 @@
 #include "core/memory.h"
 #include "core/settings.h"
 
+namespace {
+[[nodiscard]] static constexpr s16 ClampToS16(s32 value) {
```
lioncash, line 19:
```suggestion
[[nodiscard]] static constexpr s16 ClampToS16(s32 value) {
    return static_cast<s16>(std::clamp(value, s32{std::numeric_limits<s16>::min()},
                                       s32{std::numeric_limits<s16>::max()}));
}
```

Given converting from s16 to s32 isn't lossy.

src/audio_core/sink_context.h
```diff
@@ -74,16 +74,21 @@ class SinkContext {
...
+
+    [[nodiscard]] bool HasDownMixingCoefficients() const;
+    [[nodiscard]] const std::array<float_le, 4>& GetDownmixCoefficients() const;
 
 private:
     bool in_use{false};
     s32 use_count{};
     std::array<u8, AudioCommon::MAX_CHANNEL_COUNT> buffers{};
     std::size_t sink_count{};
+    bool downmix{false};
```
lioncash, line 91:
```suggestion
    bool has_downmix_coefs{false};
```

src/audio_core/sink_context.h
```diff
@@ -74,16 +74,21 @@ class SinkContext {
...
 
-    void UpdateMainSink(SinkInfo::InParams& in);
-    bool InUse() const;
-    std::vector<u8> OutputBuffers() const;
+    void UpdateMainSink(const SinkInfo::InParams& in);
+    [[nodiscard]] bool InUse() const;
+    [[nodiscard]] std::vector<u8> OutputBuffers() const;
+
+    [[nodiscard]] bool HasDownMixingCoefficients() const;
+    [[nodiscard]] const std::array<float_le, 4>& GetDownmixCoefficients() const;
```
lioncash, line 84:
```cpp
using DownmixCoefficients = std::array<float_le, 4>;
```

can be used here to deduplicate needing to write this type several times.

src/audio_core/stream.cpp
```diff
@@ -136,4 +136,13 @@ std::vector<Buffer::Tag> Stream::GetTagsAndReleaseBuffers(std::size_t max_count)
     return tags;
 }
 
+std::vector<Buffer::Tag> Stream::GetTagsAndReleaseBuffers() {
+    std::vector<Buffer::Tag> tags;
```
lioncash, line 140:
```suggestion
    std::vector<Buffer::Tag> tags;
    tags.reserve(released_buffers.size());
```

lioncash (commented):
In the future please split the addition of nodiscard into its own commit so it's quicker to read the actual functional changes.

bunnei (approved):

PR #4951, opened 2020-11-19
hle: service: Stub OLSC Initialize and SetSaveDataBackupSettingEnabled functions.

bunnei:
- Used by Animal Cross: New Horizons v1.6.0 update, minimal stub gets this update working.

### DIFF
--- a/src/core/hle/service/olsc/olsc.cpp
+++ b/src/core/hle/service/olsc/olsc.cpp
@@ -43,8 +43,6 @@ public:
     }
 
 private:
-    bool initialized{};
-
     void Initialize(Kernel::HLERequestContext& ctx) {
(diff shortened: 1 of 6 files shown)


src/core/hle/service/olsc/olsc.cpp
```diff
@@ -0,0 +1,69 @@
...
+            {9025, nullptr, "ListDownloadableSaveDataBackupInfoAsyncForDebug"},
+            {9026, nullptr, "DownloadSaveDataBackupAsyncForDebug"},
+        };
+        // clang-format on
+
+        RegisterHandlers(functions);
+    }
+
+private:
+    bool initialized{};
```
lioncash, line 46: Our convention is for member variables to be after functions.

PR #4956, opened 2020-11-20
sdl_impl: Add XINPUT_ENABLED = 0 hint

### LINKED ISSUE #4912: Yuzu only detects the first four Wirelessly Connected Xbox One Controllers on Windows 10
I'm not sure if there is a graceful fallback to direct input or raw input but because this is a specific Windows 10 issue it might be worthwhile to use [gaming.input](https://docs.microsoft.com/en-us/uwp/api/windows.gaming.input?view=winrt-19041) for xbox one controllers, which will provide trigger feedback support and avoid the issues surronding triggers in dInput (not able to use both at once).

As it stands only the first four controllers are detected under xInput.  Any subsequent xbox one controller that connects cannot be detected by yuzu.

Morph1984:
Allows more than 4 XInput controllers to be detected by SDL

Should fix #4912, needs testing

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -697,6 +697,10 @@ SDLState::SDLState() {
     RegisterFactory<VibrationDevice>("sdl", vibration_factory);
     RegisterFactory<MotionDevice>("sdl", motion_factory);
 
+    // This hint allows for the detection of more than 4 XInput devices by
+    // detecting the rest of the devices as RawInput or DirectInput devices.
+    SDL_SetHint(SDL_HINT_XINPUT_ENABLED, "0");
+
     // If the frontend is going to manage the event loop, then we don't start one here
     start_thread = SDL_WasInit(SDL_INIT_JOYSTICK) == 0;
     if (start_thread && SDL_Init(SDL_INIT_JOYSTICK) < 0) {


src/input_common/sdl/sdl_impl.cpp
```diff
@@ -678,6 +678,8 @@ SDLState::SDLState() {
     RegisterFactory<VibrationDevice>("sdl", vibration_factory);
     RegisterFactory<MotionDevice>("sdl", motion_factory);
 
+    SDL_SetHint(SDL_HINT_XINPUT_ENABLED, "0");
```
lioncash, line 681: A comment should be placed above this to indicate what this is activating

Schplee:
Verified that 5 Xinput controllers are detected and all map individually 
image

Morph1984:
Untagging as there is a regression in mapping analog triggers

bunnei:
@Morph1984 what is the status of this one?

jabroany:
Upon compiling this, I can max detect 5 controllers (unlike 4, previously). Can anyone comfirm detection of over 5?

german77:
> Upon compiling this, I can max detect 5 controllers (unlike 4 without this commit). Can anyone confirm detection of over 5 controllers?

If you are using GC controllers they are not affected by this limit.

mrpinkcat:
Any new on that ?

Morph1984:
Closing this as this may not be the proper solution, @german7 will look into this further.

PR #4969, opened 2020-11-22
CI: move refreshenv to the configure step

liushuyu:
Then cmake can find the Vulkan SDK binaries.

It seems like Azure Pipelines resets environment variables each step. By reordering the `refreshenv` instruction, the modified `PATH` variable will be passed to `CMake` correctly.


bunnei (approved):

bunnei:
Thank you!!

PR #4974, opened 2020-11-23
Warning fixes and macOS build fixes

comex:
A whole bunch of trivial changes, explained by the commit messages.  With it, I can build yuzu on macOS with the only remaining warnings being:

- Deprecation warnings due to uses of `Core::System::GetInstance()`
- Some deprecation warning from a moc-generated C++ file (no idea what's going on with that).
- Stuff from externals/cubeb


comex:
Corrected formatting.

comex:
Changed implementation of `ResponseBuilder::Push` type checking since apparently compilers other than Clang don't support the way I did it before.

ReinUsesLisp:
Could you put these changes in separate pull requests? It makes things easier to review and merge. Some of these changes have to be discussed and will generate conflicts with other pull requests.

comex:
Okay.  These were intended to be trivial but I guess some of them ended up being a bit more than that.  But I don't want to open 29 pull requests, so I guess I'll split this into a few chunks and then do individual PRs for the less-trivial changes.

PR #4975, opened 2020-11-23
nvdrv, video_core: Don't index out of bounds when given invalid syncpoint ID

comex:
- Use .at() instead of raw indexing when dealing with untrusted indices.

- For the special case of WaitFence with syncpoint id -1, instead of crashing, log an error and ignore.  This is what I get when running Super Mario Maker 2.

### DIFF
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -95,22 +95,29 @@ void GPU::WaitFence(u32 syncpoint_id, u32 value) {
     if (!is_async) {
         return;
     }
+    if (syncpoint_id == UINT32_MAX) {
+        // TODO: Research what this does.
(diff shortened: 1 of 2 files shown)


src/video_core/gpu.cpp
```diff
@@ -95,22 +95,29 @@ void GPU::WaitFence(u32 syncpoint_id, u32 value) {
     if (!is_async) {
         return;
     }
+    if (syncpoint_id == -1u) {
```
lioncash, line 98: Please don't use arithmetic like this. Use `UINT32_MAX` or be explicit with `0xFFFFFFFF`

comex: Fixed.

bunnei:
LGTM, Thank you!

bunnei (approved):

PR #4976, opened 2020-11-23
Overhaul EmuWindow::PollEvents to fix yuzu-cmd calling SDL_PollEvents off main thread

comex:
EmuWindow::PollEvents was called from the GPU thread (or the CPU thread
in sync-GPU mode) when swapping buffers.  It had three implementations:

- In GRenderWindow, it didn't actually poll events, just set a flag and
  emit a signal to indicate that a frame was displayed.

- In EmuWindow_SDL2_Hide, it did nothing.

- In EmuWindow_SDL2, it did call SDL_PollEvents, but this is wrong
  because SDL_PollEvents is supposed to be called on the thread that set
  up video - in this case, the main thread, which was sleeping in a
  busyloop (regardless of whether sync-GPU was enabled).  On macOS this
  causes a crash.

To fix this:

- Rename EmuWindow::PollEvents to OnFrameDisplayed, and give it a
  default implementation that does nothing.

- In EmuWindow_SDL2, do not override OnFrameDisplayed, but instead have
  the main thread call SDL_WaitEvent in a loop.

### DIFF
--- a/src/video_core/renderer_opengl/renderer_opengl.cpp
+++ b/src/video_core/renderer_opengl/renderer_opengl.cpp
@@ -151,8 +151,8 @@ void RendererOpenGL::SwapBuffers(const Tegra::FramebufferConfig* framebuffer) {
 
     rasterizer->TickFrame();
 
-    render_window.PollEvents();
     context->SwapBuffers();
+    render_window.OnFrameDisplayed();
 }
 
 void RendererOpenGL::PrepareRendertarget(const Tegra::FramebufferConfig* framebuffer) {
--- a/src/core/frontend/emu_window.h
+++ b/src/core/frontend/emu_window.h
@@ -102,8 +102,8 @@ public:
         float render_surface_scale = 1.0f;
     };
 
-    /// Polls window events
-    virtual void PollEvents() = 0;
+    /// Called from GPU thread when a frame is displayed.
+    virtual void OnFrameDisplayed() {}
 
     /**
      * Returns a GraphicsContext that the frontend provides to be used for rendering.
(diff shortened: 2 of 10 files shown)


comex:
Corrected formatting.

comex:
…and re-corrected formatting, since apparently git-clang-format wasn't enough.

src/video_core/renderer_opengl/renderer_opengl.cpp
```diff
@@ -151,8 +151,8 @@ void RendererOpenGL::SwapBuffers(const Tegra::FramebufferConfig* framebuffer) {
 
     rasterizer->TickFrame();
 
-    render_window.PollEvents();
     context->SwapBuffers();
+    render_window.OnFrameDisplayed();
```
ReinUsesLisp, line 155: Is the order intended to be changed here?

ReinUsesLisp (commented):
Looks good to me, just one question marked in a separate comment.

src/video_core/renderer_opengl/renderer_opengl.cpp
comex: Ah, yeah… forgot to document that bit.  I did intentionally move it to after SwapBuffers so that OpenGL version is more consistent with the Vulkan version, as well as with the new name ("displayed" not "about to be displayed").  The exact timing doesn't really matter, as currently the only effect of OnFrameDisplayed is to start fading out the loading screen.

ReinUsesLisp:
Looks good to me, I'll test it locally and merge it afterwards.

ReinUsesLisp (approved):
Thanks for looking into this

PR #4977, opened 2020-11-24
hle: services: Fix a crash with improper NVFlinger lifetime management.

bunnei:
- This crash would happen when attempting to shutdown yuzu early on in boot.

### DIFF
--- a/src/core/hle/service/service.h
+++ b/src/core/hle/service/service.h
@@ -191,7 +191,7 @@ private:
  */
 class Services final {
 public:
-    Services(std::shared_ptr<SM::ServiceManager>& sm, Core::System& system);
+    explicit Services(std::shared_ptr<SM::ServiceManager>& sm, Core::System& system);
     ~Services();
 
 private:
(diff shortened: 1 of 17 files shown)


ReinUsesLisp (approved):

src/core/hle/service/service.h
```diff
@@ -181,10 +185,17 @@ class ServiceFramework : public ServiceFrameworkBase {
...
 
-/// Initialize ServiceManager
-void Init(std::shared_ptr<SM::ServiceManager>& sm, Core::System& system);
+/**
+ * The purpose of this class is to own any objects that need to be shared across the other service
+ * implementations. Will be torn down when the global system instance is shutdown.
+ */
+class Services final {
+public:
+    Services(std::shared_ptr<SM::ServiceManager>& sm, Core::System& system);
```
lioncash, line 194:
```suggestion
    explicit Services(std::shared_ptr<SM::ServiceManager>& sm, Core::System& system);
```

PR #4978, opened 2020-11-24
core: cpu_manager: Fix shutdown crash when closing before emulation starts.

bunnei:
- This crash would happen when shutting down yuzu while shaders were compiling.

### DIFF
--- a/src/core/cpu_manager.cpp
+++ b/src/core/cpu_manager.cpp
@@ -4,6 +4,7 @@
 
 #include "common/fiber.h"
 #include "common/microprofile.h"
+#include "common/scope_exit.h"
 #include "common/thread.h"
 #include "core/arm/exclusive_monitor.h"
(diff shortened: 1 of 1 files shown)


src/core/cpu_manager.cpp
```diff
@@ -343,6 +344,16 @@ void CpuManager::RunThread(std::size_t core) {
     data.initialized = true;
     const bool sc_sync = !is_async_gpu && !is_multicore;
     bool sc_sync_first_use = sc_sync;
+
+    // Cleanup
+    auto cleanup = detail::ScopeExit([&] {
```
ReinUsesLisp, line 349: Do we use `SCOPE_EXIT` macro? `detail` is not intended to be accessed from the outside.

bunnei: derp yes

Issue #4965, opened 2020-11-22
Regression: some games fail to boot or load


lat9nq:
Caused by #4907 @ogniK5377 

yuzu (since Mainline 444) no longer can load a save in BOTW 1.0.0 ~~nor boot MK8D 1.0.0~~ (MK8D is possibly an unrelated issue)

[Log of BOTW 1.0.0](https://github.com/yuzu-emu/yuzu/files/5579835/yuzu_log.txt)

bunnei:
Fixed by #4981

PR #4981, opened 2020-11-24
nvservices: Reintroducee IoctlCtrl

ogniK5377:
Fixes regression caused by #4907 which caused games like Breath of the Wild 1.0.0 not to boot.


lat9nq:
Fixes #4965

bunnei (approved):

PR #4996, opened 2020-11-24
Kernel: Refactor to use 4-instances of Dynarmic & various cleanups and improvements

bunnei:
The primary purpose of this PR is to change how our implementation of allocating an instance of Dynarmic per guest thread to one where we allocate an instance per physical core (4 total). This has the benefit of greatly reducing memory usage, since Dynarmic does not share block allocations across instances. This part was mostly done by @FernandoS27.

This should fix many issues related to yuzu crashing for users who do not have a large enough page file or enough RAM. It also improves performance in various places.

Additionally, this change includes a few more cleanups, and some minor changes around the scheduler code, as well as changes necessary for the 4-JITs implementation to work with NROs.

This change does surface an issue that we may need to live with for now -- There seems to be a bug where moving guest threads to other Dynarmic backends causes issues. Most notably, we see this with Diablo 3 during boot. I've fully narrowed this down to actually changing the Dynarmic instance for a guest thread, not any behavior surrounding our scheduler, so it seems like it could be a bug with exclusive monitor or some other Dynarmic state that is not being probably loaded/restored.

@MerryMage if you have any ideas on the above, it would be very helpful, thank you!

### DIFF
--- a/src/core/cpu_manager.cpp
+++ b/src/core/cpu_manager.cpp
@@ -113,22 +113,23 @@ void CpuManager::MultiCoreRunGuestThread() {
         auto& sched = kernel.CurrentScheduler();
         sched.OnThreadStart();
     }
+    auto* thread = kernel.CurrentScheduler().GetCurrentThread();
+    auto& host_context = thread->GetHostContext();
+    host_context->SetRewindPoint(GuestRewindFunction, this);
     MultiCoreRunGuestLoop();
 }
 
 void CpuManager::MultiCoreRunGuestLoop() {
     auto& kernel = system.Kernel();
-    auto* thread = kernel.CurrentScheduler().GetCurrentThread();
+
     while (true) {
         auto* physical_core = &kernel.CurrentPhysicalCore();
-        auto& arm_interface = thread->ArmInterface();
         system.EnterDynarmicProfile();
         while (!physical_core->IsInterrupted()) {
-            arm_interface.Run();
+            physical_core->Run();
             physical_core = &kernel.CurrentPhysicalCore();
         }
         system.ExitDynarmicProfile();
-        arm_interface.ClearExclusiveState();
         auto& scheduler = kernel.CurrentScheduler();
         scheduler.TryDoContextSwitch();
     }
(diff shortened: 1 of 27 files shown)


src/core/hle/kernel/physical_core.h
```diff
@@ -49,6 +58,18 @@ class PhysicalCore {
     // Shutdown this physical core.
     void Shutdown();
 
+    bool IsInitialized() const {
+        return !!arm_interface;
```
lioncash, line 62:
```suggestion
        return arm_interface != nullptr;
```

Is more intuitive to read.

src/core/hle/kernel/physical_core.h
```diff
@@ -4,9 +4,12 @@
 
 #pragma once
 
+#include <array>
 #include <cstddef>
 #include <memory>
 
+#include "core/arm/arm_interface.h"
```
lioncash, line 11: This can be a forward declaration

src/core/hle/kernel/physical_core.cpp
```diff
@@ -4,40 +4,58 @@
...
-      core_index{id}, scheduler{scheduler}, guard{std::make_unique<Common::SpinLock>()} {}
+PhysicalCore::PhysicalCore(std::size_t core_index, Core::System& system,
+                           Kernel::Scheduler& scheduler, Core::CPUInterrupts& interrupts)
+    : core_index{core_index}, system{system}, scheduler{scheduler},
+      interrupts{interrupts}, guard{std::make_unique<Common::SpinLock>()} {}
 
 PhysicalCore::~PhysicalCore() = default;
 
+void PhysicalCore::Initialize(bool is_64_bit) {
+    auto& kernel = system.Kernel();
```
lioncash, line 24: This function should preserve the `#ifdef ARCHITECTURE_x86_64` architecture check so an informative compile error can be given during attempts to build on an unsupported architecture.

src/core/cpu_manager.cpp
```diff
@@ -113,22 +113,23 @@ void CpuManager::MultiCoreRunGuestThread() {
         auto& sched = kernel.CurrentScheduler();
         sched.OnThreadStart();
     }
+    auto* thread = kernel.CurrentScheduler().GetCurrentThread();
+    auto& host_context = thread->GetHostContext();
+    host_context->SetRewindPoint(std::function<void(void*)>(GuestRewindFunction), this);
```
lioncash, line 118:
```suggestion
    host_context->SetRewindPoint(GuestRewindFunction, this);
```

src/core/hle/kernel/physical_core.cpp
bunnei: needed for `Core::CPUInterrupts`

PR #4937, opened 2020-11-18
InputCommon: Add multiple udp server support

german77:
Modifies the ui and adds the necessary code to permit multiple udp clients to connect up to 8 different udp servers

Thanks to @Its-Rei for modifying the UI

PR #5003, opened 2020-11-26
input_common: unbreak -Werror with Clang

jbeich:
Regressed by #4927. From [error log](https://github.com/yuzu-emu/yuzu/files/5599711/yuzu-qt5-s20201125.log):
```c++
error: unknown warning option '-Werror=unused-but-set-parameter'; did you mean '-Werror=unused-parameter'? [-Werror,-Wunknown-warning-option]
error: unknown warning option '-Werror=unused-but-set-variable'; did you mean '-Werror=unused-const-variable'? [-Werror,-Wunknown-warning-option]
```

### DIFF
--- a/src/input_common/CMakeLists.txt
+++ b/src/input_common/CMakeLists.txt
@@ -56,8 +56,8 @@ else()
         -Werror=reorder
         -Werror=shadow
         -Werror=sign-compare
-        -Werror=unused-but-set-parameter
-        -Werror=unused-but-set-variable
+        $<$<CXX_COMPILER_ID:GNU>:-Werror=unused-but-set-parameter>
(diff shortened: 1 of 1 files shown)


german77:
This should no longer be needed as #4937 uses this data.  I'm not sure about the changes in the cmake file

jbeich:
> This should no longer be needed as #4937 uses this data.

OK, dropped 277ad81d60dc. #4937 also had a Clang regression (reported in that PR).

> I'm not sure about the changes in the cmake file

It's already used elsewhere
https://github.com/yuzu-emu/yuzu/blob/0e15c68f54ae5217e6247603b9e31757515f89e1/src/core/CMakeLists.txt#L639-L640

lioncash (approved):

PR #4967, opened 2020-11-22
video_core/texture_cache: Rewrite the texture cache

ReinUsesLisp:
The current texture cache has several points that hurt maintainability and performance. It’s easy to break unrelated parts of the cache when doing minor changes. The cache can easily forget valuable information about the cached textures by CPU writes or simply by its normal usage.

The original intent of this pull request was to emulate bindless textures at a low level. Instead of handling textures on demand, this new cache would load all textures in the current texture pool and translate from guest texture+sampler indices during shader execution using VK_EXT_descriptor_indexing. This is not possible on OpenGL due to the lack of bindless separate sampler objects, and it can’t be hacked with GL_NV_bindless_texture or GL_ARB_bindless_texture because Nvidia and AMD’s OpenGL driver uses combined samplers for bindless.

A minor issue on emulating this on Vulkan is the lack of heap pointers to image handles on the API (in other words, physical pointers to bindless textures). Games can sample textures with the wrong target and have defined behavior (e.g. sampling 1D as 1D array, and 3D as 1D), we’d have to create one descriptor array per target and fill invalid combinations with null descriptors. Descriptors sets can only have one unbounded array and there are around 8 different texture targets, this wouldn’t fit on some vendors.

Emulating bindless textures this way had to be ditched (for now) for several reasons:
We don’t know what or when textures have been modified from the GPU (imageStore).
Decompressed ASTC textures don’t fit in VRAM on certain titles (2.5+ GiB), although OpenGL drivers can handle this gracefully, even when textures are resident. Compressing them to other formats hurts image quality significantly and it’s slow.
Aliased images have to be hacked or manually synchronized after every write. Because render targets are aligned, it’s sometimes impossible to know if it’s rendering to a 6x8 texture or 7x8, given the report size is 16x8. We can use hints for this (render area, viewport size), but they break when two render targets are bound. Keeping render targets synchronized with aliased textures after every write when bindless is used seems to be the best solution for this.

The idea can still be implemented in the future.

Aliased images are emulated through copies on demand. For example, when a 16x8 texture is rendered and then a texture in the same address is sampled with an 8x8 size, the 16x8 texture is copied to 8x8. This also analyzes mips and layers.

Rendering to compressed textures is implemented with copies. This could be emulated on Vulkan, but it’s not yet implemented due to the constraints required by the API to be able to render compressed images.

[...]

PR #5005, opened 2020-11-26
common: Add Common::DivCeil and Common::DivCeilLog2

ReinUsesLisp:
Add an equivalent to 'Common::AlignUp(n, d) / d' and a log2 alternative.

Required by #4967

### DIFF
--- a/src/common/div_ceil.h
+++ b/src/common/div_ceil.h
@@ -1,0 +1,26 @@
+// Copyright 2020 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <cstddef>
+#include <type_traits>
+
+namespace Common {
+
+/// Ceiled integer division.
+template <typename N, typename D>
+requires std::is_integral_v<N>&& std::is_unsigned_v<D>[[nodiscard]] constexpr auto DivCeil(
+    N number, D divisor) {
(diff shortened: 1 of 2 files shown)


src/common/div_ceil.h
```diff
@@ -0,0 +1,26 @@
...
+#include <cstddef>
+#include <type_traits>
+
+namespace Common {
+
+/// Ceiled integer division.
+template <typename N, typename D>
+requires std::is_integral_v<N>&& std::is_unsigned_v<D>[[nodiscard]] constexpr auto DivCeil(
+    N number, D divisor) {
+    return (number + divisor - 1) / divisor;
```
lioncash, line 16: If N is signed and D is unsigned, then this arithmetic will cause an implicit sign conversion. `number` should be casted to `D` to get rid of the implicit conversion.

ReinUsesLisp: Addressed it, should I change this now to always return a type of `D` instead of `auto`?

lioncash: Depends on how you'd want to handle it.

e.g. if N is `< sizeof(s32)` and D is `< sizeof(u32)`, the arithmetic would promote to `int` being returned from the function, which `auto` allows without an implicit truncation.

If you want to predictably make it always return an unsigned value, then you would need to cast the whole expression to D  before changing the return type.

bunnei (approved):

PR #5014, opened 2020-11-26
CMakeLists: disable -Winvalid-offsetof

comex:
This Clang warning complains when offsetof is used on a
non-standard-layout type (i.e. any class using various C++ features),
even though it works fine (and is not undefined behavior as of C++17).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -66,6 +66,7 @@ else()
         -Wextra
         -Wmissing-declarations
         -Wno-attributes
+        -Wno-invalid-offsetof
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ else()
         -Wmissing-declarations
         -Wno-attributes
         -Wno-unused-parameter
+        -Wno-invalid-offsetof
```
lioncash, line 70: This should be alphabetically organized

lioncash (approved):

PR #5015, opened 2020-11-26
tests: Fix warning about comparison between signed and unsigned

comex:


src/tests/common/bit_field.cpp
```diff
@@ -68,7 +68,7 @@ TEST_CASE("BitField", "[common]") {
                    }});
 
     // bit fields: 01101100111101'10101110'1011'101100
-    REQUIRE(be_bitfield.raw == 0b01101100'11110110'10111010'11101100);
+    REQUIRE(be_bitfield.raw == 0b01101100'11110110'10111010'11101100u);
```
lioncash, line 71: We generally use the upper-cased form for literal suffixes, since they stand out more in fonts.

lioncash (approved):

bunnei (approved):

PR #5017, opened 2020-11-26
CMakeLists: Always use `zstd::zstd`, not just on Windows

comex:
This seems to be required in order to actually use the flags for the
copy of zstd found by the top-level CMakeLists.  Without it, it seems to
blindly add -lzstd, which breaks the build for me since my copy of zstd
is not in the default library path.

Originally this used `zstd`, not `zstd::zstd`, on all platforms.  Commit
2cbce77 ("CMakeLists: use system zstd on Linux") made it
platform-dependent: despite its title, it changed the behavior on
Windows to use `zstd::zstd`, while leaving the behavior on Linux
unchanged.

I'm pretty sure `zstd::zstd` is the right choice on all platforms, but
by accident `zstd` also works on Linux because the library happens to
(usually) be installed in /usr/lib.


src/common/CMakeLists.txt
```diff
@@ -211,9 +211,4 @@ create_target_directory_groups(common)
 find_package(Boost 1.71 COMPONENTS context headers REQUIRED)
 
 target_link_libraries(common PUBLIC ${Boost_LIBRARIES} fmt::fmt microprofile)
-target_link_libraries(common PRIVATE lz4::lz4 xbyak)
-if (MSVC)
-  target_link_libraries(common PRIVATE zstd::zstd)
-else()
-  target_link_libraries(common PRIVATE zstd)
-endif()
+target_link_libraries(common PRIVATE lz4::lz4 zstd::zstd xbyak)
```
lioncash, line 214: We generally keep library order alphabetical where applicable, so zstd::zstd should come after xbyak.

comex: Fixed.

comex:
Hmm, what the heck is going on with the Windows build.  It's not printing anything about zstd, either that it found it or that it was unable to find it...

comex:
Okay, the issue is that because CMakeLists.txt explicitly sets CMAKE_FIND_PACKAGE_PREFER_CONFIG to TRUE, [CMake will](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature-and-module-mode) first search for `zstdConfig.cmake`, and only if that fails will use `Findzstd.cmake` in `externals/find-modules`.

I don't think the use of CMAKE_FIND_PACKAGE_PREFER_CONFIG makes sense.  It's not just zstd; the build setup assumes that several libraries define a target named 'foo::foo', where foo is the library name.  The corresponding `Findfoo.cmake` files in `externals/find-modules` ensure that that target is defined, but a system installation of the same library may not.

Let's see what happens if CMAKE_FIND_PACKAGE_PREFER_CONFIG is set to FALSE instead.

comex:
…Turns out `Findnlohmann_json.cmake` does not work.  At all.  Nothing ever sets `nlohmann_json_LIBRARY`.  I tried fixing it by setting it to `PC_nlohmann_json_LIBRARIES`, i.e. the libraries found via `pkg-config`, but it didn't work because for some reason `nlohmann_json`'s pkg-config file fails to define `Libs`.

comex:
…And `Findfmt.cmake` is also broken, because adds `/prefix/include/fmt` to the include path instead of `/prefix/include`.  Adding `/prefix/include/fmt` to the include path is wrong because (a) Yuzu includes fmt as `<fmt/format.h>`, not `<format.h>`, and (b) `fmt` contains a header named `locale.h`, which conflicts with the libc `<locale.h>` if it's added to the include path.

comex:
Oh, `nlohmann_json` does not even have a library.  It's header-only.  So its .pc file is not buggy, but `Findnlohmann_json.cmake` is buggy, as the lack of library path produces CMake errors.  I believe it needs to specify the imported library as `INTERFACE` rather than `UNKNOWN`.

comex:
Add another to the broken list: `FindBoost.cmake`, which is not even in `externals/find-modules` but shipped with CMake itself.  Specifically, there seems to be a problem when the Linux CI build (1) runs find_package but finds a too-old version of Boost, (2) installs a newer one, then (3) runs find_package a second time to try to detect the new one.  Perhaps there are some variables that are getting wrongly carried over from the first invocation to the second.

comex:
Nope, it's not variables being carried over.  Here's what's going on with Boost.

After it's not found on the system and Conan builds it, Conan writes a `BoostConfig.cmake` to the build directory.  This file is based on a generic Conan template and, *unlike* the `BoostConfig.cmake` installed by a standard installation of Boost, knows nothing about Boost components.

Then `src/common/CMakeLists.txt` does `find_package(Boost 1.71 COMPONENTS context headers REQUIRED)`, attempting to require the `context` and `headers` components of Boost.  What happens next?

In `master` (CMAKE_FIND_PACKAGE_PREFER_CONFIG = TRUE): The `find_package` executes `BoostConfig.cmake` directly, and *the requested components are ignored*.  You can add `bogus` to the components list and it will still build.

In this PR (CMAKE_FIND_PACKAGE_PREFER_CONFIG = FALSE): The `find_package` executes `FindBoost.cmake` packaged with CMake, which in turn [calls `find_package` recursively with `NO_MODULE`](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L460), causing CMake to load the same `BoostConfig.cmake`.  But `FindBoost.cmake` expects `BoostConfig.cmake` to set variables like `Boost_context_FOUND` for each component, and it goes on to check that that variables are set for each requested component.  Since Conan's `BoostConfig.cmake` didn't set any such variables, it fails to find any components.

For now, I worked around the issue by just removing the list of components.  Which is a hack.  To be fair, that list doesn't work as expected in the first place… except when checking a system-installed Boost, where it does work.

comex:
Closing this for now since it seems like a rabbit hole I need to prioritize other things over.

PR #5020, opened 2020-11-27
Disable analog joystick from buttons by default

german77:
Since this setting adds delay to the input, some games like SSBU don't work as intended. This will disable the setting by default and add a checkbox under advanced settings to enable it.

### DIFF
--- a/src/core/settings.h
+++ b/src/core/settings.h
@@ -178,6 +178,8 @@ struct Values {
     u16 udp_input_port;
     u8 udp_pad_index;
 
+    bool emulate_analog_keyboard;
+
     bool mouse_enabled;
     std::string mouse_device;
     MouseButtonsRaw mouse_buttons;
--- a/src/yuzu/configuration/configure_input_advanced.ui
+++ b/src/yuzu/configuration/configure_input_advanced.ui
@@ -2546,14 +2546,27 @@
                  </property>
                 </widget>
                </item>
-               <item row="4" column="2">
+               <item row="1" column="0">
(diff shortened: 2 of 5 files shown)


bunnei (approved):

src/yuzu/configuration/configure_input_advanced.ui
```diff
@@ -2546,14 +2546,27 @@
...
+               <item row="1" column="0">
+                 <widget class="QCheckBox" name="analog_joystick_button">
+                   <property name="minimumSize">
+                     <size>
+                       <width>0</width>
+                       <height>23</height>
+                     </size>
+                   </property>
+                   <property name="text">
+                     <string>Analog joystick button</string>
```
Morph1984, line 2558: Is there something more descriptive that can be used to convey that this setting is **only** meant for 1. Keyboard users and 2. Is meant for games such as diablo 3

Morph1984 (changes requested):

src/yuzu/configuration/configure_input_advanced.ui
Morph1984: I gave this a bit of thought, perhaps, "Emulate Analog with Keyboard Input"

src/core/settings.h
```diff
@@ -178,6 +178,8 @@ struct Values {
     u16 udp_input_port;
     u8 udp_pad_index;
 
+    bool analog_joystick_button;
```
Morph1984, line 181: Rename instances of `analog_joystick_button` to `emulate_analog_keyboard`

Morph1984 (approved):

PR #5121, opened 2020-12-04
core: Optimize core timing utility functions to avoid unnecessary math

bunnei:
Avoids a lot of unnecessary 128-bit math for imperceptible accuracy. Thanks @ReinUsesLisp for the profiling.


src/core/core_timing_util.h
```diff
@@ -1,24 +1,59 @@
...
-std::chrono::nanoseconds CyclesToNs(s64 cycles);
-std::chrono::microseconds CyclesToUs(s64 cycles);
-
-u64 CpuCyclesToClockCycles(u64 ticks);
+namespace detail {
+constexpr u64 CNTFREQ_ADJUSTED = Hardware::CNTFREQ / 1000;
+constexpr u64 BASE_CLOCK_RATE_ADJUSTED = Hardware::BASE_CLOCK_RATE / 1000;
+} // namespace detail
+
+constexpr s64 msToCycles(std::chrono::milliseconds ms) {
```
lioncash, line 19: While these are being touched, maybe mark these functions as `[[nodiscard]]` as well.

src/common/wall_clock.cpp
```diff
@@ -41,16 +40,11 @@ class StandardWallClock final : public WallClock {
     }
 
     u64 GetClockCycles() override {
-        std::chrono::nanoseconds time_now = GetTimeNS();
-        const u128 temporary =
-            Common::Multiply64Into128(time_now.count(), emulated_clock_frequency);
-        return Common::Divide128On32(temporary, 1000000000).first;
+        return GetTimeNS().count() * (emulated_clock_frequency / 1000) / 1000000;
```
degasus, line 43: This will still use an integer division. If you want to optimize this behavior, please precalculate `(emulated_clock_frequency / 1000) / 1000000` and just do a multiplication here.

e.g.:
```
// precalculate whenever emulated_clock_frequency changes.
uint64_t factor = (static_cast<uint128_t>(emulated_clock_frequency) << 64) / 1000_000_000;

// One MUL instruction, 3 cycles latency.
return (static_cast<uint128_t>(GetTimeNS().count()) * factor) >> 64;
```

degasus (approved):

PR #5141, opened 2020-12-05
maxwell_dma: Rename RenderEnable::Mode::FALSE and TRUE to avoid name conflict

comex:
On Apple platforms, `FALSE` and `TRUE` are defined as macros by <mach/boolean.h>, which is included by various system headers.

Note that there appear to be no actual users of the names to fix up.

### DIFF
--- a/src/video_core/engines/maxwell_dma.h
+++ b/src/video_core/engines/maxwell_dma.h
@@ -72,11 +72,13 @@ public:
 
     struct RenderEnable {
         enum class Mode : u32 {
-            FALSE = 0,
-            TRUE = 1,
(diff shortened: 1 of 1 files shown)


ReinUsesLisp:
This can be replaced with `PascalCase`. Underscore + upper case character is reserved by the standard.

comex:
Like this?  Or did you want only False and True to be Pascal cased?

ReinUsesLisp:
Changing all of them is fine, it's more consistent. CI will probably fail though.

ReinUsesLisp (approved):

PR #5142, opened 2020-12-05
network, sockets: Replace `POLL_IN`, `POLL_OUT`, etc. constants with an `enum class PollEvents`

comex:
Actually, two enum classes, since for some reason there are two separate
yet identical `PollFD` types used in the codebase.  I get that one is
ABI-compatible with the Switch while the other is an abstract type used
for the host, but why not use `WSAPOLLFD` directly for the latter?

Anyway, why make this change?  Because on Apple platforms, `POLL_IN`,
`POLL_OUT`, etc. (with an underscore) are defined as macros in
<sys/signal.h>.  (This is inherited from FreeBSD.)  So defining
a variable with the same name causes a compile error.

I could just rename the variables, but while I was at it I thought I
might as well switch to an enum for stronger typing.

Also, change the type used for values copied directly to/from the
`events` and `revents` fields of the host *native*
`pollfd`/`WSASPOLLFD`, from `u32` to `short`, as `short` is the correct
canonical type on both Unix and Windows.

### DIFF
--- a/src/core/hle/service/sockets/bsd.cpp
+++ b/src/core/hle/service/sockets/bsd.cpp
@@ -489,18 +489,18 @@ std::pair<s32, Errno> BSD::PollImpl(std::vector<u8>& write_buffer, std::vector<u
     }
 
     for (PollFD& pollfd : fds) {
-        ASSERT(pollfd.revents == 0);
+        ASSERT(False(pollfd.revents));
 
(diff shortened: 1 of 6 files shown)


comex:
Edited to fix `POLL_PRI` being used instead of `POLLPRI` (a preexisting bug which this change made into a compilation error on Linux), and use the right type in `translate`.

ReinUsesLisp (approved):

PR #5151, opened 2020-12-06
vfs_real: When moving files or directories, don't assume file opening will succeed

comex:
Found this via a warning, but it's a substantive fix.

Since this is only for a cache, it should be safe to silently drop the entry if opening fails.  I think.

### DIFF
--- a/src/core/file_sys/vfs_real.cpp
+++ b/src/core/file_sys/vfs_real.cpp
@@ -133,8 +133,11 @@ VirtualFile RealVfsFilesystem::MoveFile(std::string_view old_path_, std::string_
         }
 
         cache.erase(old_path);
-        file->Open(new_path, "r+b");
-        cache.insert_or_assign(new_path, std::move(file));
+        if (file->Open(new_path, "r+b")) {
+            cache.insert_or_assign(new_path, std::move(file));
+        } else {
+            LOG_ERROR(Service_FS, "Failed to open path {} in order to re-cache it", new_path);
+        }
     } else {
         UNREACHABLE();
         return nullptr;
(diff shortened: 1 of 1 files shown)


src/core/file_sys/vfs_real.cpp
```diff
@@ -133,8 +133,9 @@ VirtualFile RealVfsFilesystem::MoveFile(std::string_view old_path_, std::string_
         }
 
         cache.erase(old_path);
-        file->Open(new_path, "r+b");
-        cache.insert_or_assign(new_path, std::move(file));
+        if (file->Open(new_path, "r+b")) {
+            cache.insert_or_assign(new_path, std::move(file));
+        }
```
bunnei, line 138: Should we log an error if this fails?

src/core/file_sys/vfs_real.cpp
```diff
@@ -214,9 +215,10 @@ VirtualDir RealVfsFilesystem::MoveDirectory(std::string_view old_path_,
         }
 
         auto file = cached.lock();
-        file->Open(file_new_path, "r+b");
         cache.erase(file_old_path);
-        cache.insert_or_assign(std::move(file_new_path), std::move(file));
+        if (file->Open(file_new_path, "r+b")) {
+            cache.insert_or_assign(std::move(file_new_path), std::move(file));
```
bunnei, line 220: Should we log an error if this fails?

bunnei:
@comex ping on the comments above, so we can get this merged sooner than later. Thanks!

comex:
Sorry for the long delay.  I added an error log as requested.  I'm not sure whether `Service_FS` is an appropriate log class for this, since this isn't within hle/service/filesystem, but I didn't see any better options.

bunnei (approved):

PR #5154, opened 2020-12-06
hle: Type check ResponseBuilder::Push arguments, and fix use in vi.cpp

comex:
- Add a type check so that calling Push with an invalid type produces a compile error rather than a linker error.

- vi.cpp was calling Push with a variable of type `std::size_t`.  There's no explicit overload for `size_t`, but there is one for `u64`, which on most platforms is the same type as `size_t`.  On macOS, however, it isn't: both types are 64 bits, but `size_t` is `unsigned long` and `u64` is `unsigned long long`.  Regardless, it makes more sense to explicitly use `u64` here instead of `size_t`.

### DIFF
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -166,8 +166,23 @@ public:
         ValidateHeader();
     }
 
+    void PushImpl(s8 value);
+    void PushImpl(s16 value);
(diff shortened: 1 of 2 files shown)


bunnei (approved):

bunnei:
LGTM, Thank you!

PR #5173, opened 2020-12-09
common/file_util: Make use of std::filesystem

lioncash:
Migrates a bunch of our file handling code over to std::filesystem to simplify a bunch of our code and make it easier to maintain.

Notably with this, `CopyDir` will now function on Windows.

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -68,13 +68,6 @@
 #include <algorithm>
 #include <sys/stat.h>
 
-#ifndef S_ISDIR
-#define S_ISDIR(m) (((m)&S_IFMT) == S_IFDIR)
-#endif
-
-// This namespace has various generic functions related to files and paths.
-// The code still needs a ton of cleanup.
-// REMEMBER: strdup considered harmful!
 namespace Common::FS {
 namespace fs = std::filesystem;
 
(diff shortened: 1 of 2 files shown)


src/common/file_util.h
```diff
@@ -38,48 +39,34 @@ enum class UserPath {
...
-    bool isDirectory;
-    u64 size;                 // file length or number of entries from children
-    std::string physicalName; // name on disk
-    std::string virtualName;  // name in FST names table
-    std::vector<FSTEntry> children;
-};
-
-// Returns true if file filename exists
-[[nodiscard]] bool Exists(const std::string& filename);
+// Returns true if the exists
```
ReinUsesLisp, line 42:
```suggestion
// Returns true if the path exists
```

src/common/file_util.cpp
```diff
@@ -67,290 +68,102 @@
...
-        LOG_DEBUG(Common_Filesystem, "path exists {}", fullPath);
-        return true;
-    }
-
-    std::size_t position = 0;
-    while (true) {
-        // Find next sub path
-        position = fullPath.find(DIR_SEP_CHR, position);
+bool Rename(const fs::path& src, const fs::path& dst) {
+    LOG_TRACE(Common_Filesystem, "{} --> {}", src.string(), dst.string());
```
ReinUsesLisp, line 127: Is the double `-` in `-->` intentional? Ditto for `Copy`.

lioncash: This is a pre-existing log call

ReinUsesLisp (approved):

adityaruplaha (approved):
Edit: Nvm, didn't notice this was already merged.

PR #5174, opened 2020-12-09
common/file_util: Fix and deprecate CreateFullPath, add CreateDirs

ReinUsesLisp:
Fix CreateFullPath to have its intended previous behavior (whatever that was), and deprecate it in favor of the new CreateDirs function.

Unlike CreateDir, CreateDirs is marked as [[nodiscard]] to avoid new code ignoring its result value.

Thanks to Morph and ogniK for helping me debug and fix this.

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -114,18 +114,39 @@ bool CreateDir(const fs::path& path) {
     return true;
 }
 
-bool CreateFullPath(const fs::path& path) {
+bool CreateDirs(const fs::path& path) {
     LOG_TRACE(Common_Filesystem, "path {}", path.string());
+
+    if (Exists(path)) {
+        LOG_DEBUG(Common_Filesystem, "path exists {}", path.string());
+        return true;
+    }
 
     std::error_code ec;
     const bool success = fs::create_directories(path, ec);
 
     if (!success) {
-        LOG_ERROR(Common_Filesystem, "Unable to create full path: {}", ec.message());
+        LOG_ERROR(Common_Filesystem, "Unable to create directories: {}", ec.message());
         return false;
     }
 
     return true;
+}
+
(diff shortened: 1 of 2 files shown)


Morph1984 (approved):

src/common/file_util.cpp
```diff
@@ -109,20 +114,41 @@ bool CreateDir(const fs::path& path) {
...
         return false;
     }
 
     return true;
 }
 
+bool CreateFullPath(const fs::path& path) {
+    LOG_TRACE(Common_Filesystem, "path {}", path);
+
+    // Removes trailing slashes and turns any '\' into '/'
```
lioncash, line 139: Which OS is this a problem on? I'm fairly certain Windows handles both of these in their filesystem implementation. Part of the purpose of std::filesystem is to abstract away needing to care about this in the first place.

src/common/file_util.cpp
```diff
@@ -109,20 +114,41 @@ bool CreateDir(const fs::path& path) {
...
+    const auto new_path = SanitizePath(path.string(), DirectorySeparator::ForwardSlash);
+
+    if (new_path.rfind('.') == std::string::npos) {
```
lioncash, line 142: This can be inverted into:

`if (path.has_extension())`

src/common/file_util.cpp
```diff
@@ -109,20 +114,41 @@ bool CreateDir(const fs::path& path) {
...
+        // The path is a directory
+        return CreateDirs(new_path);
+    } else {
+        // The path is a file
+        // Creates directory preceding the last '/'
+        return CreateDirs(new_path.substr(0, new_path.rfind('/')));
```
lioncash, line 148:
```cpp
return CreateDirs(path.parent_path());
```

PR #5178, opened 2020-12-09
 Add option to reset window size to 1080p

### LINKED ISSUE #5175: Feature request: Ability to resize to 1080p using "reset window size"
I have a 1440p monitor, so I can't use fullscreen mode to display in 1080p, because pixels are not 1:1 (they're blurry) in full screen.

There is a feature to reset the window size, but it reset only to 720p (code : https://github.com/yuzu-emu/yuzu/blob/44f7067cab5c8999a69f8dd92574a2b6e8edfe34/src/yuzu/main.cpp#L2254 ). I would like the feature to be able to choose between 720p and 1080p when reseting the window size.
Or if possible, detect the current native resolution of the game and resize to it ?
I don't know if many games are 720p in docked mode, but maybe we could detect docked and use the docked resolution when reseting the window size instead of the undocked ?

thank you

german77:
Allows to resize between 720p and 1080p depending if undocked or docked mode is selected.

Fixes #5175

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -939,7 +939,10 @@ void GMainWindow::ConnectMenuEvents() {
             &GMainWindow::OnDisplayTitleBars);
     connect(ui.action_Show_Filter_Bar, &QAction::triggered, this, &GMainWindow::OnToggleFilterBar);
     connect(ui.action_Show_Status_Bar, &QAction::triggered, statusBar(), &QStatusBar::setVisible);
-    connect(ui.action_Reset_Window_Size, &QAction::triggered, this, &GMainWindow::ResetWindowSize);
+    connect(ui.action_Reset_Window_Size_720, &QAction::triggered, this,
+            &GMainWindow::ResetWindowSize720);
+    connect(ui.action_Reset_Window_Size_1080, &QAction::triggered, this,
+            &GMainWindow::ResetWindowSize1080);
 
     // Fullscreen
     connect(ui.action_Fullscreen, &QAction::triggered, this, &GMainWindow::ToggleFullscreen);
@@ -2258,7 +2261,7 @@ void GMainWindow::ToggleWindowMode() {
     }
 }
 
-void GMainWindow::ResetWindowSize() {
+void GMainWindow::ResetWindowSize720() {
     const auto aspect_ratio = Layout::EmulationAspectRatio(
         static_cast<Layout::AspectRatio>(Settings::values.aspect_ratio.GetValue()),
         static_cast<float>(Layout::ScreenUndocked::Height) / Layout::ScreenUndocked::Width);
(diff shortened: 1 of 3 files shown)


Kewlan:
I say this should be based on the monitor size instead of the docked state, as this will just be an inconvenience for anyone with a 1080p monitor or smaller.

german77:
> I say this should be based on the monitor size instead of the docked state, as this will just be an inconvenience for anyone with a 1080p monitor or smaller.

If the monitor is smaller I could limit the size to the current monitor resolution. That way it at least fits in the screen. The goal here is to have an easy way to set a 1:1 pixel ratio

momala454:
thank you for working on my feature request 👍

german77:
I compare against the main window size. It takes into account any UI element. There are only two resolutions 720p in undocked and 1080p in docked then it's resized to match the window size. The resolution will not change only the resize.

src/yuzu/main.cpp
```diff
@@ -2252,16 +2252,35 @@ void GMainWindow::ToggleWindowMode() {
...
 
 void GMainWindow::ResetWindowSize() {
+    const auto docked_enabled = Settings::values.use_docked_mode.GetValue();
     const auto aspect_ratio = Layout::EmulationAspectRatio(
         static_cast<Layout::AspectRatio>(Settings::values.aspect_ratio.GetValue()),
         static_cast<float>(Layout::ScreenUndocked::Height) / Layout::ScreenUndocked::Width);
     if (!ui.action_Single_Window_Mode->isChecked()) {
-        render_window->resize(Layout::ScreenUndocked::Height / aspect_ratio,
-                              Layout::ScreenUndocked::Height);
+        if (docked_enabled) {
```
lioncash, line 2260: The resize call can be deduplicated like so:

```cpp
const u32 height = docked_enabled ? Layout::ScreenDocked::Height : Layout::ScreenUndocked::Height;
render_window->resize(height / aspect_ratio, height);
```

Ditto for the other duplicated resize in the else case.

Kewlan:
Maybe it'd be better to replace **Reset Window Size** with **Set Window Size** and a sub menu of common game resolutions like **1280x720**, **1600x900**, and **1920x1080**. 
That way there'll be no potential confusion of why it resizes differently, 1080p users like me who are always in docked mode won't have to undock first to resize, and we'll have another option for games that are 900p.

momala454:
can we auto detect the current resolution ?

Kewlan:
I'm sure we can, but some games (like Super Mario Odyssey) changes the output resolution in the middle of gameplay, so I don't think changing the window size based on that is a good idea. Plus that also goes against the original purpose of this function; At first you couldn't make the window smaller than 1280x720, but when that was changed, this function was added alongside it: #3413

momala454:
ok, so it's probably better to display the current resolution on the task bar below, as a reference

Morph1984:
The original intention behind **not** resizing to 1080p for docked resolutions is for displays that can't display the entire yuzu window with the render window being 1080p, since the yuzu window adds around 80 pixels to the height

@Kewlan Your suggestion is a good one, although I would leave out 900p as the framebuffer is only set to 720p or 1080p. Resizing to 900p would mean a double conversion of 900p -> 1080p -> 900p and there will be not benefit.

Ultimately, this is my proposal:

A "Set Window Size" menu with 2 options: 720p and 1080p. We do not list the horizontal resolution as this is dependent on the aspect ratio

We should not maximize the window at any rate since that is not the intended function of this feature.

bunnei (approved):

PR #5180, opened 2020-12-09
file_util: Make ForeachDirectoryEntry() use directory iterators

lioncash:
Simplifies the implementing code so that we don't need to explicitly handle platform-specifics.

### DIFF
--- a/src/common/file_util.cpp
+++ b/src/common/file_util.cpp
@@ -218,9 +218,9 @@ bool CreateEmptyFile(const std::string& filename) {
     return true;
 }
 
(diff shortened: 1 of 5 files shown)


ReinUsesLisp (approved):

adityaruplaha (approved):

bunnei:
@lioncash needs a rebase

lioncash:
Going to need to close due to Windows' wacky UTF-16/wide-char nonsense relative to everything else that needs to be handled a little more rigorously,

PR #5187, opened 2020-12-12
fs: Revert all std::filesystem changes

Morph1984:
This reverts all the changes for std::filesystem as it causes an application crash when narrow multibyte (UTF-8) encoded characters are converted to and from UTF-16 on Windows.

Reverts #5173, #5174, #5176, #5179 and #5181


bunnei:
Supersedes #5186

PR #5190, opened 2020-12-12
controllers/npad: Validate device handles before use

### LINKED ISSUE #5188: Nekopara Vol. 3 crashes on loading screen
msvc-source-20201211-770af5fc1
Windows 10 Home SL 20H2 x64
Previous volumes (1, 2) launches successfully.
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5682879/yuzu_log.txt)

Morph1984:
Some games such as NEKOPARA Vol. 3 send invalid device handles when calling InitializeVibrationDevice. Introduce a check to validate the device handle before use.

Fixes #5188

### DIFF
--- a/src/core/hle/service/hid/controllers/npad.cpp
+++ b/src/core/hle/service/hid/controllers/npad.cpp
@@ -116,6 +116,31 @@ u32 Controller_NPad::IndexToNPad(std::size_t index) {
     }
 }
 
+bool Controller_NPad::IsNpadIdValid(u32 npad_id) {
+    switch (npad_id) {
+    case 0:
+    case 1:
+    case 2:
+    case 3:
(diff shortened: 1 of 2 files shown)


FearlessTobi:
Are you sure it's the game having a bug and not yuzu?
It just seems a bit weird to me that a game would send invalid values like this.
Either way, it would be a good idea to see if and how sanity checks are performed in the hid module on real hw.
Maybe a hwtest would also be useful.

german77:
I did some HW test.  The switch returns error 0xf6ca on invalid handle init and error 0xf4ca on send vibration values. None of the controllers actually vibrated.

bunnei (approved):

PR #5196, opened 2020-12-14
cmake/conan: Conditionally add target Boost::context

lat9nq:
Addresses an issue with the two competing versions of Conan's Boost package that are currently floating around, in this case the newer one.

Adds the Boost::context target only if it's recognized by CMake as a target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,13 +194,18 @@ macro(yuzu_find_packages)
     unset(FN_FORCE_REQUIRED)
 endmacro()
 
-if (NOT Boost_FOUND)
-    find_package(Boost 1.73.0 COMPONENTS context headers QUIET)
-    if (Boost_FOUND)
-        set(Boost_LIBRARIES Boost::boost Boost::context)
+find_package(Boost 1.73.0 COMPONENTS context headers QUIET)
+if (Boost_FOUND)
+    set(Boost_LIBRARIES Boost::boost)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -312,6 +312,10 @@ if (CONAN_REQUIRED_LIBS)
     if (NOT Boost_FOUND)
         find_package(Boost 1.73.0 REQUIRED COMPONENTS context headers)
         set(Boost_LIBRARIES Boost::boost)
+        # Conditionally add Boost::context only if the active version of the Conan Boost package provides it
+        if (TARGET Boost::context)
+            set(Boost_LIBRARIES ${Boost_LIBRARIES} Boost::context)
```
lioncash, line 317:
```suggestion
            list(APPEND Boost_LIBRARIES Boost::context)
```

CMakeLists.txt
```diff
@@ -312,6 +312,10 @@ if (CONAN_REQUIRED_LIBS)
     if (NOT Boost_FOUND)
         find_package(Boost 1.73.0 REQUIRED COMPONENTS context headers)
         set(Boost_LIBRARIES Boost::boost)
+        # Conditionally add Boost::context only if the active version of the Conan Boost package provides it
```
lioncash, line 315: May want to expand this to explain why `context` wouldn't exist in the first place.

lat9nq: I tried to expand out what I meant with that comment. Let me know if it looks good now.

bunnei (approved):

PR #5199, opened 2020-12-17
memory: Fix issues with thread safety

bunnei:
The way that we access page tables is not thread safe. Much of this code existed before our multicore CPU and asynchronous GPU implementations. Two main issues come to mind:
* We do not atomically access a page address from its metadata on reads/writes
* The current page table might change (e.g. on a reschedule) while some other thread is accessing it (e.g. the GPU thread)

This change should fix the above.

Thanks to @ReinUsesLisp for bringing this to my attention.

PR #5206, opened 2020-12-18
core: memory: Ensure thread safe access when pages are rasterizer cached

bunnei:
There is a race between memory pages being accessed and the GPU marking them as cached by the rasterizer. This change adds a lock to alleviate this issue. I also addressed this in #5199, however that change added a bit of unnecessary overhead,

Thanks to @ReinUsesLisp for bringing this to my attention.

Supersedes #5199

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -692,7 +692,7 @@ struct Memory::Impl {
     template <typename T>
     void Write(const VAddr vaddr, const T data) {
         // Avoid adding any extra logic to this fast-path block
-        if (u8* const pointer = current_page_table->pointers[vaddr >> PAGE_BITS]; pointer) {
+        if (u8* const pointer = current_page_table->pointers[vaddr >> PAGE_BITS]) {
             std::memcpy(&pointer[vaddr], &data, sizeof(T));
             return;
         }
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -646,7 +646,7 @@ struct Memory::Impl {
     template <typename T>
     T Read(const VAddr vaddr) {
         // Avoid adding any extra logic to this fast-path block
-        if (const u8* const pointer = current_page_table->pointers[vaddr >> PAGE_BITS]; pointer) {
+        if (const u8* const pointer = current_page_table->pointers[vaddr >> PAGE_BITS]) {
             T value;
             std::memcpy(&value, &pointer[vaddr], sizeof(T));
             return value;
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -4,6 +4,7 @@
 
 #include <algorithm>
 #include <cstring>
+#include <mutex>
 #include <optional>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


src/core/memory.cpp
```diff
@@ -497,7 +498,21 @@ struct Memory::Impl {
         return CopyBlock(*system.CurrentProcess(), dest_addr, src_addr, size);
     }
 
+    struct PageEntry {
+        u8* const pointer;
+        const Common::PageType attribute;
+    };
+
+    PageEntry SafePageEntry(std::size_t base) {
+        std::lock_guard<std::mutex> lock(rasterizer_cache_guard);
```
lioncash, line 507:
```suggestion
        std::lock_guard lock{rasterizer_cache_guard};
```

src/core/memory.cpp
```diff
@@ -497,7 +498,21 @@ struct Memory::Impl {
         return CopyBlock(*system.CurrentProcess(), dest_addr, src_addr, size);
     }
 
+    struct PageEntry {
+        u8* const pointer;
+        const Common::PageType attribute;
+    };
+
+    PageEntry SafePageEntry(std::size_t base) {
```
lioncash, line 506:
```suggestion
    PageEntry SafePageEntry(std::size_t base) const {
```

src/core/memory.cpp
```diff
@@ -756,6 +770,7 @@ struct Memory::Impl {
         return true;
     }
 
+    std::mutex rasterizer_cache_guard;
```
lioncash, line 773:
```suggestion
    mutable std::mutex rasterizer_cache_guard;
```

src/core/memory.cpp
```diff
@@ -670,15 +691,21 @@ struct Memory::Impl {
      */
     template <typename T>
     void Write(const VAddr vaddr, const T data) {
-        u8* const page_pointer = current_page_table->pointers[vaddr >> PAGE_BITS];
-        if (page_pointer != nullptr) {
-            // NOTE: Avoid adding any extra logic to this fast-path block
-            std::memcpy(&page_pointer[vaddr], &data, sizeof(T));
+        // Avoid adding any extra logic to this fast-path block
+        if (u8* const pointer = current_page_table->pointers[vaddr >> PAGE_BITS]; pointer) {
```
lioncash, line 695:
```suggestion
        if (u8* const pointer = current_page_table->pointers[vaddr >> PAGE_BITS]) {
```

src/core/memory.cpp
```diff
@@ -630,16 +645,22 @@ struct Memory::Impl {
      */
     template <typename T>
     T Read(const VAddr vaddr) {
-        const u8* const page_pointer = current_page_table->pointers[vaddr >> PAGE_BITS];
-        if (page_pointer != nullptr) {
-            // NOTE: Avoid adding any extra logic to this fast-path block
+        // Avoid adding any extra logic to this fast-path block
+        if (const u8* const pointer = current_page_table->pointers[vaddr >> PAGE_BITS]; pointer) {
```
lioncash, line 649:
```suggestion
        if (const u8* const pointer = current_page_table->pointers[vaddr >> PAGE_BITS]) {
```

PR #5225, opened 2020-12-24
cmake: Always enable Vulkan

ReinUsesLisp:
Removes the unnecessary burden of maintaining separate #ifdef paths and allows us sharing generic Vulkan code across APIs.

~~This is missing CI changes, bear with me until it's happy.~~ Fixed.


ReinUsesLisp:
This should be fine now. It will conflict with tagged PRs, but that's expected.

PR #5228, opened 2020-12-25
applets/controller: Make use of std::span

Morph1984:
Eliminates the need to convert an array into a vector

### DIFF
--- a/src/core/frontend/applets/controller.h
+++ b/src/core/frontend/applets/controller.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <functional>
+#include <span>
 
 #include "common/common_types.h"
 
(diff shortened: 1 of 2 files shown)


src/core/frontend/applets/controller.h
```diff
@@ -23,9 +24,9 @@ struct ControllerParameters {
     bool keep_controllers_connected{};
     bool enable_single_mode{};
     bool enable_border_color{};
-    std::vector<BorderColor> border_colors{};
+    std::span<BorderColor> border_colors{};
```
lioncash, line 27: I'm not sure using a spans as member variables is a particularly safe thing to do here, given spans are non-owning, so it's very easy to unintentionally cause undefined behavior this way.

PR #5231, opened 2020-12-26
renderer_vulkan/fixed_pipeline_state: Move enabled bindings to static state

ReinUsesLisp:
Without using VK_EXT_robustness2, we can't consider the 'enabled' (not null) vertex buffers as dynamic state, as this leads to invalid Vulkan state. Move this to static state that is always hashed and compared in the pipeline key.

The bits used for enabled vertex buffers are moved into the attribute state bitfield. This is not 'correct' as it's not an attribute state, but that struct has bits to spare, and it's used in an array of 32 elements (the exact same number of vertex buffer bindings).


goldenx86:
VK_EXT_extended_dynamic_state, not VK_EXT_robustness2.
Who are you, AyyMD?

bunnei (approved):

PR #5236, opened 2020-12-27
input_common: process udp packets only for the correct pad

gal20:
Currently, each of the four pads in a single socket process all new data packets, regardless of whether the packet is actually relevant to them. This works fine when only one pad is used in a single socket (though redundant, as each packet is processed four times, and the same motion input is reported as four different devices), but with additional pads that would result in corrupted motion input, as motion inputs from different devices are processed together.
This PR fixes this issue, by checking the pad index of the incoming packet before processing it.


bunnei:
LGTM, thank you!

bunnei (approved):

PR #5237, opened 2020-12-28
nvdec: Incorporate syncpoint manager

ameerj:
This PR includes the incorporation of the syncpoint manager to more accurately emulate nvdec. 

However, this does not implement proper async gpu support, that still remains as a TODO.


src/video_core/command_classes/sync_manager.cpp
```diff
@@ -38,7 +38,7 @@ u32 SyncptIncrManager::IncrementWhenDone(u32 class_id, u32 id) {
 }
 
 void SyncptIncrManager::SignalDone(u32 handle) {
-    const auto done_incr =
+    const auto& done_incr =
```
lioncash, line 41:
```suggestion
    const auto done_incr =
```

find_if always returns the iterator by value, so this is misleading

bunnei (approved):

PR #5239, opened 2020-12-28
.ci/templates: Enable QT translation for MSVC CI

FearlessTobi:
Previously this flag was missing, causing translation files not to be shipped with CI builds of yuzu.


FearlessTobi:
I've been able to confirm that this change works now.
Translation support is now available in latest EA.

bunnei (approved):

PR #5240, opened 2020-12-28
CMakeLists: Enable Qt translation support by default

Morph1984:
Should we have Qt translation support enabled by default on all builds? Or should they be individually enabled using a flag like in #5239


FearlessTobi:
I'm not quite sure this should be enabled by default. Building with translations increases build times a bit and could cause issues on some systems when QtLinguist dependency is missing. 
Also, for dev builds I think translations are not really needed,

Morph1984:
Good points, linux systems may not have QtLinguist installed as part of QtCore

PR #5249, opened 2020-12-30
core/memory: Read and write page table atomically

ReinUsesLisp:
Squash attributes into the pointer's integer, making them an uintptr_t
pair containing 2 bits at the bottom and then the pointer. These bits
are currently unused thanks to alignment requirements. By doing this
we no longer have to use a mutex to write and read the page table.

Configure Dynarmic to mask out these bits on pointer reads.

While we are at it, remove some unused attributes carried over from
Citra.

Read/Write and other hot functions use a two step unpacking process that
is less readable to stop MSVC from emitting an extra AND instruction in
the hot path:
```diff
 mov         rdi,rcx
 shr         rdx,0Ch
 mov         r8,qword ptr [rax+8]
 mov         rax,qword ptr [r8+rdx*8]
 mov         rdx,rax
-and         al,3
 and         rdx,0FFFFFFFFFFFFFFFCh
 je          Core::Memory::Memory::Impl::Read<unsigned char>
 mov         rax,qword ptr [vaddr]
 movzx       eax,byte ptr [rdx+rax]
```

See https://github.com/MerryMage/dynarmic/pull/566 for a more detailed explanation of how this works.

### DIFF
--- a/src/common/page_table.h
+++ b/src/common/page_table.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <atomic>
 #include <tuple>
 
 #include "common/common_types.h"
@@ -20,10 +21,6 @@ enum class PageType : u8 {
     /// Page is mapped to regular memory, but also needs to check for rasterizer cache flushing and
     /// invalidation
     RasterizerCachedMemory,
-    /// Page is mapped to a I/O region. Writing and reading to this page is handled by functions.
-    Special,
-    /// Page is allocated for use.
-    Allocated,
 };
 
 struct SpecialRegion {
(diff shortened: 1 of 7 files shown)


src/common/page_table.h
```diff
@@ -48,6 +45,43 @@ struct SpecialRegion {
  * mimics the way a real CPU page table works.
  */
 struct PageTable {
+    static constexpr int ATTRIBUTE_BITS = 2;
+
+    class PageInfo {
```
lioncash, line 50: It might be best to provide a comment explaining why the class needs to exist in the first place, especially given how uncommon it is (relatively) to directly store attributes in a pointer.

src/common/page_table.h
```diff
@@ -48,6 +45,43 @@ struct SpecialRegion {
...
+
+        [[nodiscard]] uintptr_t Raw() const noexcept {
+            return raw;
+        }
+
+        void Store(u8* pointer, PageType type) noexcept {
+            raw.store(reinterpret_cast<uintptr_t>(pointer) | static_cast<uintptr_t>(type));
+        }
+
+        static u8* ExtractPointer(uintptr_t raw) {
```
lioncash, line 73: This and ExtractType can also be noexcept.

src/common/page_table.h
```diff
@@ -48,6 +45,43 @@ struct SpecialRegion {
...
+            return ExtractType(raw);
+        }
+
+        [[nodiscard]] std::pair<u8*, PageType> PointerType() const noexcept {
+            const uintptr_t non_atomic_raw = raw;
+            return {ExtractPointer(non_atomic_raw), ExtractType(non_atomic_raw)};
+        }
+
+        [[nodiscard]] uintptr_t Raw() const noexcept {
+            return raw;
```
lioncash, line 66: If atomicity is all that matters, this can be:

```suggestion
            return raw.load(std::memory_order_relaxed);
```

src/common/page_table.h
```diff
@@ -48,6 +45,43 @@ struct SpecialRegion {
...
+        [[nodiscard]] u8* Pointer() const noexcept {
+            return ExtractPointer(raw);
+        }
+
+        [[nodiscard]] PageType Type() const noexcept {
+            return ExtractType(raw);
+        }
+
+        [[nodiscard]] std::pair<u8*, PageType> PointerType() const noexcept {
+            const uintptr_t non_atomic_raw = raw;
```
lioncash, line 61:
```suggestion
            const uintptr_t non_atomic_raw = raw.load(std::memory_order_relaxed);
```

If all that matters is atomicity

Ditto for usages in `Pointer()` and `Type()` as well

src/common/page_table.h
ReinUsesLisp: Done. Thanks for pointing it out.

bunnei (approved):

PR #4773, opened 2020-10-09
Build System: Add Flatpak CI

gal20:
Based on #3770, I created a pipeline with an updated flatpak manifest that successfully builds the application. The Builds are found [here](https://dev.azure.com/gal20/gal20.yuzu/_build).

To set up the pipeline, you'll need to upload your private ssh and gpg keys (named 'ssh.key' and 'gpg.key') as secure files, and create a variable group called 'flatpak-variables' that contains the flatpak environment variables.

Changed required before merging: 
- Uncomment the sshfs command
- Change the docker image to yuzu's (currently uses meirod's, because the linux-flatpak image doesn't exist in yuzuemu/build-environment)
- Update the runtime version to 5.15  (see yuzu-emu/build-environments#21)

Update 27.1:
The builds now fail to compile yuzu because the 5.13 runtime is too old. On the 5.15 runtime it compiles successfully.

PR #5250, opened 2020-12-30
ci/linux: Build an AppImage

lat9nq:
This builds yuzu in an [AppImage](https://appimage.org/) alongside the other archives during release. Required to allow distributing yuzu in the future with upgraded dependencies, such as Qt. As a result of using an AppImage, we remove the need for users to install dependencies before running yuzu on wild Linux distros. Since the build system recently switched to Ubuntu 18.04, this change also brings yuzu back to Ubuntu 18.04, Debian Buster, CentOS 8, and similar.

There are a few changes of note to make this possible:
- Puts the `artifacts/` directory into a variable. The variable is used to directly copy the AppImage to `artifacts/` from within the Linux-specific scripts.
- Reverts to using GNU Make in order to install yuzu to a custom directory.
- Downloads needed tools at build time to generate the AppImage.
- Script exits prematurely if there's an error, rather than continuing with the rest of the script like nothing wrong happened.

The AppImage will be released alongside the other archives, not inside and not replacing them. AppImages are already compressed, so double-compressing one would be ill-advised.

Any yuzu AppImages generated between now and April 2021 will not be valid for submission on [AppImageHub](https://appimage.github.io/) due to incompatibilities with Ubuntu 16.04 as in their [guidelines](https://docs.appimage.org/packaging-guide/distribution.html#appimagehub). Beyond that, I believe there is more work needed to make yuzu eligible. That makes this PR more like preliminary work.

Thanks to @jroweboy for informing me about [CheckRT](https://github.com/darealshinji/AppImageKit-checkrt), which is required to run programs on older Linux installs despite using newer versions of GCC to build them with. (This was several months ago, but nevertheless...)

Double-thank you to @AniLeo for talking to me about launchpad PPAs of GCC 10 and Qt for Ubuntu 18.04.

As a side note, this is also a bit of a test for the 18.04 container, since it is fairly young and hasn't checked many PRs yet.

### DIFF
--- a/.ci/scripts/linux/docker.sh
+++ b/.ci/scripts/linux/docker.sh
@@ -1,14 +1,54 @@
 #!/bin/bash -ex
 
+# Exit on error, rather than continuing with the rest of the script.
+set -e
+
 cd /yuzu
 
 ccache -s
 
 mkdir build || true && cd build
-cmake .. -G Ninja -DDISPLAY_VERSION=$1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/lib/ccache/gcc -DCMAKE_CXX_COMPILER=/usr/lib/ccache/g++ -DYUZU_ENABLE_COMPATIBILITY_REPORTING=${ENABLE_COMPATIBILITY_REPORTING:-"OFF"} -DENABLE_COMPATIBILITY_LIST_DOWNLOAD=ON -DUSE_DISCORD_PRESENCE=ON -DENABLE_QT_TRANSLATION=ON
+cmake .. -DDISPLAY_VERSION=$1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/lib/ccache/gcc -DCMAKE_CXX_COMPILER=/usr/lib/ccache/g++ -DYUZU_ENABLE_COMPATIBILITY_REPORTING=${ENABLE_COMPATIBILITY_REPORTING:-"OFF"} -DENABLE_COMPATIBILITY_LIST_DOWNLOAD=ON -DUSE_DISCORD_PRESENCE=ON -DENABLE_QT_TRANSLATION=ON -DCMAKE_INSTALL_PREFIX="/usr"
 
-ninja
+make -j$(nproc)
 
 ccache -s
 
 ctest -VV -C Release
+
+make install DESTDIR=AppDir
+rm -vf AppDir/usr/bin/yuzu-cmd AppDir/usr/bin/yuzu-tester
+
+# Download tools needed to build an AppImage
+wget -nc https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage
+wget -nc https://github.com/linuxdeploy/linuxdeploy-plugin-qt/releases/download/continuous/linuxdeploy-plugin-qt-x86_64.AppImage
+wget -nc https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage
+wget -nc https://github.com/darealshinji/AppImageKit-checkrt/releases/download/continuous/AppRun-patched-x86_64
+wget -nc https://github.com/darealshinji/AppImageKit-checkrt/releases/download/continuous/exec-x86_64.so
+# Set executable bit
+chmod 755 \
+    appimagetool-x86_64.AppImage \
+    AppRun-patched-x86_64 \
+    exec-x86_64.so \
+    linuxdeploy-x86_64.AppImage \
+    linuxdeploy-plugin-qt-x86_64.AppImage
+
+# Workaround for https://github.com/AppImage/AppImageKit/issues/828
+export APPIMAGE_EXTRACT_AND_RUN=1
+
+mkdir -p AppDir/usr/optional
+mkdir -p AppDir/usr/optional/libstdc++
+mkdir -p AppDir/usr/optional/libgcc_s
+
+# Deploy yuzu's needed dependencies
+./linuxdeploy-x86_64.AppImage --appdir AppDir --plugin qt
+
+# Workaround for building yuzu with GCC 10 but also trying to distribute it to Ubuntu 18.04 et al.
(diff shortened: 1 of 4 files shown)


adityaruplaha (approved):
Going through the code, it looks good.

I'm on mobile though, so might've missed something.

gal20:
Just a reminder, since this PR is about distributing yuzu on Linux, #4773 sets up a flatpak ci similiar to citra's.

.ci/scripts/linux/docker.sh
```diff
@@ -5,9 +5,46 @@ cd /yuzu
...
 
 mkdir build || true && cd build
-cmake .. -G Ninja -DDISPLAY_VERSION=$1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/lib/ccache/gcc -DCMAKE_CXX_COMPILER=/usr/lib/ccache/g++ -DYUZU_ENABLE_COMPATIBILITY_REPORTING=${ENABLE_COMPATIBILITY_REPORTING:-"OFF"} -DENABLE_COMPATIBILITY_LIST_DOWNLOAD=ON -DUSE_DISCORD_PRESENCE=ON -DENABLE_QT_TRANSLATION=ON
+cmake .. -DDISPLAY_VERSION=$1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/lib/ccache/gcc -DCMAKE_CXX_COMPILER=/usr/lib/ccache/g++ -DYUZU_ENABLE_COMPATIBILITY_REPORTING=${ENABLE_COMPATIBILITY_REPORTING:-"OFF"} -DENABLE_COMPATIBILITY_LIST_DOWNLOAD=ON -DUSE_DISCORD_PRESENCE=ON -DENABLE_QT_TRANSLATION=ON -DCMAKE_INSTALL_PREFIX="/usr"
 
-ninja
+make -j$(nproc)
+
+make install DESTDIR=AppDir
+rm -vf AppDir/usr/bin/yuzu-cmd AppDir/usr/bin/yuzu-tester
```
v1993, line 13: Maybe avoid building tester at all? I'm not sure if it is easy to do when using "install" target, but it probably should be.

lat9nq: We need tests to check PRs on CI? As well as it's probably a good idea to let yuzu pass tests before releasing it????

lat9nq:
> Just a reminder, since this PR is about distributing yuzu on Linux, #4773 sets up a flatpak ci similiar to citra's.

In my opinion, there are distinct advantages here in distributing an AppImage versus a Flatpak. One needs to install Flatpak, then install yuzu through a thrid-party repository, then setup the directory permissions to allow yuzu to access the user's game files. And then there's the issue of support, where all of the written knowledge of yuzu on Linux and its directory structure needs to be reworked for the specific case of running yuzu in a Flatpak environment.

An AppImage simply needs downloaded and set with the executable bit. Directory permissions would already be implicitly setup, and all of the written knowledge for yuzu still applies, since it runs as a distinct program. Nothing needs installed through a package manager, and the user does not need root permissions to run yuzu. Much less the user doesn't need to download a couple graphics drivers and another desktop environment alongside effectively what makes up an AppImage.

gal20:
> In my opinion, there are distinct advantages here in distributing an AppImage versus a Flatpak. 

I disagree. I think flatpak offers much more advantages to end-users compared to AppImage. As far as I know, AppImage doesn't offer auto-updating and desktop integration like flatpak, so using AppImage isn't all that different than downloading precompiled binaries.

> One needs to install Flatpak, then install yuzu through a thrid-party repository, then setup the directory permissions to allow yuzu to access the user's game files. 

Flatpak doesn't require any permission setup by end-users, it is pre-determined by the manifest.
Also, it's possible to simply publish the manifest to flathub rather than using a third-party repo like citra did.

> And then there's the issue of support, where all of the written knowledge of yuzu on Linux and its directory structure needs to be reworked for the specific case of running yuzu in a Flatpak environment.

A properly written app requires no modification for running in flatpak.

lat9nq:
> As far as I know, AppImage doesn't offer auto-updating and desktop integration like flatpak, so using AppImage isn't all that different than downloading precompiled binaries.

You may need to look [here](https://github.com/AppImage/AppImageUpdate). Beyond that, we also plan to get the maintenance tool officially running on Linux.

> ...it's possible to simply publish the manifest to flathub rather than using a third-party repo like citra did.

Possible, but not favorable as I believe the reason the third-party repository exists is because of frequent updates. yuzu wouldn't be any different than Citra at pushing updates, maybe worse.

> Flatpak doesn't require any permission setup by end-users, it is pre-determined by the manifest.
> A properly written app requires little to no modification for running in flatpak. Both citra and yuzu can run in a flatpak environment without any change to the code.

I can't inspect Citra at the moment because Citra was failing to install through Flatpak on my machine using the guide on their site. I ended up removing all of my Flatpak packages, so now I'm sitting here downloading a gigabyte of packages, including drivers I already have installed on the host, because my internet isn't too great.

And what of AMD/Intel users who need the most recent version of Mesa? This may not matter as much for Citra, but for yuzu this makes a big difference, and I can't find any Flatpak of mesa-git despite a few mentions of one here and there. Do they need to hop through *another* third-party repository just to get a working graphics driver? What if someone wants to use AMDVLK? Or AMDGPU-PRO? AMDVLK-PRO?

qurious-pixel:
Looks good. 
My yuzu fork builds appimages with auto-updater built in. 
Easy to setup, just that the updater needs to have the same name and path each time.
And to consider the amount of updates. Rather than each PR, I set it to update daily for yuzu, or weekly for other projects.

gal20:
> I can't inspect Citra at the moment because Citra was failing to install through Flatpak on my machine using the guide on their site. I ended up removing all of my Flatpak packages, so now I'm sitting here downloading a gigabyte of packages, including drivers I already have installed on the host, because my internet isn't too great.

Sorry you had to go through that, but this is not really a problem with flatpak. The citra repository doesn't work because of expired GPG key, and the citra team never renewed it or updated the site to reflect that. The only problem with flatpak here is not being verbose enough about the error (with `-v` you will be able to see the error more clearly, though).

bunnei (approved):

PR #5261, opened 2020-12-30
yuzu/main: Fix 'Hide mouse on inactivity' and port citra-emu/citra#5476

gal20:
The mouseMoveEvent never gets triggered for some reason, causing the mouse to never reappear, unless pressed or moved to the menu bar. Using eventFilter instead solves this issue.
I don't have a lot of experience with Qt, so any suggestions/alternative solutions are welcome.

### DIFF
--- a/src/yuzu/main.h
+++ b/src/yuzu/main.h
@@ -248,6 +248,7 @@ private slots:
     void OnCoreError(Core::System::ResultStatus, std::string);
     void OnReinitializeKeys(ReinitializeKeyBehavior behavior);
     void OnLanguageChanged(const QString& locale);
+    void OnMouseActivity();
 
 private:
     void RemoveBaseContent(u64 program_id, const QString& entry_type);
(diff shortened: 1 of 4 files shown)


src/yuzu/main.h
```diff
@@ -335,6 +335,6 @@ private slots:
     void dropEvent(QDropEvent* event) override;
     void dragEnterEvent(QDragEnterEvent* event) override;
     void dragMoveEvent(QDragMoveEvent* event) override;
-    void mouseMoveEvent(QMouseEvent* event) override;
     void mousePressEvent(QMouseEvent* event) override;
+    bool eventFilter(QObject* object, QEvent* event);
```
lioncash, line 339:
```suggestion
    bool eventFilter(QObject* object, QEvent* event) override;
```

FearlessTobi:
There is also https://github.com/citra-emu/citra/pull/5476 which still needs to be ported to yuzu.
Would it be possible to integrate it into this PR?

gal20:
> There is also [citra-emu/citra#5476](https://github.com/citra-emu/citra/pull/5476) which still needs to be ported to yuzu.
> Would it be possible to integrate it into this PR?

Done! This should now work when single window mode is disabled.

adityaruplaha (approved):

bunnei (approved):

PR #5262, opened 2020-12-30
buffer_cache/buffer_base: Add a range tracking buffer container and tests

ReinUsesLisp:
This keeps track of the modified CPU and GPU ranges on a CPU page granularity, notifying the given rasterizer about state changes in the tracking behavior of the buffer.
The buffer takes advantage of C++20's `<bit>` operations when possible to check for modified pages on a 256 KiB range as fast as possible. On MSVC this will use CPU extensions when available branching at runtime.

Due to how error prone the container design is, I've added unit tests for it. Some tests are based on bugs found using this buffer in games, so if we ever break it in the future in a way that might harm commercial games, the tests should fail.

This is part of the buffer cache rewrite, it doesn't currently change yuzu's behavior aside from the tests.

### DIFF
--- a/src/video_core/buffer_cache/buffer_base.h
+++ b/src/video_core/buffer_cache/buffer_base.h
@@ -1,0 +1,495 @@
+// Copyright 2020 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <algorithm>
+#include <bit>
+#include <limits>
+#include <utility>
+
+#include "common/alignment.h"
+#include "common/common_funcs.h"
+#include "common/common_types.h"
+#include "common/div_ceil.h"
+#include "core/memory.h"
+
+namespace VideoCommon {
+
+enum class BufferFlagBits {
+    Picked = 1 << 0,
+};
+DECLARE_ENUM_FLAG_OPERATORS(BufferFlagBits)
+
+/// Tag for creating null buffers with no storage or size
+struct NullBufferParams {};
+
+/**
+ * Range tracking buffer container.
+ *
+ * It keeps track of the modified CPU and GPU ranges on a CPU page granularity, notifying the given
+ * rasterizer about state changes in the tracking behavior of the buffer.
+ *
+ * The buffer size and address is forcefully aligned to CPU page boundaries.
+ */
+template <class RasterizerInterface>
+class BufferBase {
+    static constexpr u64 PAGES_PER_WORD = 64;
+    static constexpr u64 BYTES_PER_PAGE = Core::Memory::PAGE_SIZE;
+    static constexpr u64 BYTES_PER_WORD = PAGES_PER_WORD * BYTES_PER_PAGE;
+
+    /// Vector tracking modified pages tightly packed with small vector optimization
+    union WrittenWords {
+        /// Returns the pointer to the words state
+        [[nodiscard]] const u64* Pointer(bool is_short) const noexcept {
+            return is_short ? &stack : heap;
+        }
+
+        /// Returns the pointer to the words state
+        [[nodiscard]] u64* Pointer(bool is_short) noexcept {
+            return is_short ? &stack : heap;
+        }
+
+        u64 stack = 0; ///< Small buffers storage
+        u64* heap;     ///< Not-small buffers pointer to the storage
+    };
+
+    struct GpuCpuWords {
+        explicit GpuCpuWords() = default;
+        explicit GpuCpuWords(u64 size_bytes_) : size_bytes{size_bytes_} {
+            if (IsShort()) {
+                cpu.stack = ~u64{0};
+                gpu.stack = 0;
+            } else {
+                // Share allocation between CPU and GPU pages and set their default values
+                const size_t num_words = NumWords();
+                u64* const alloc = new u64[num_words * 2];
+                cpu.heap = alloc;
+                gpu.heap = alloc + num_words;
+                std::fill_n(cpu.heap, num_words, ~u64{0});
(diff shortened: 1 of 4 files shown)


src/video_core/buffer_cache/buffer_base.h
```diff
@@ -0,0 +1,480 @@
...
+
+    BufferBase& operator=(BufferBase&& rhs) noexcept {
+        Release();
+        rasterizer = rhs.rasterizer;
+        cpu_addr = rhs.cpu_addr;
+        size_bytes = rhs.size_bytes;
+        cpu_written_words = rhs.cpu_written_words;
+        gpu_written_words = rhs.gpu_written_words;
+        flags = rhs.flags;
+        rhs.cpu_written_words.heap = nullptr;
```
lioncash, line 95: Missing a `return *this;` after this line.

src/video_core/buffer_cache/buffer_base.h
```diff
@@ -0,0 +1,480 @@
...
+            std::fill_n(gpu_written_words.heap, num_words, 0);
+        }
+        // Clean up tailing bits
+        const u64 last_local_page = Common::DivCeil(size_bytes % BYTES_PER_WORD, BYTES_PER_PAGE);
+        const u64 shift = (PAGES_PER_WORD - last_local_page) % PAGES_PER_WORD;
+        u64& last_word = cpu_written_words.Pointer(IsShort())[NumWords() - 1];
+        last_word = last_word << shift >> shift;
+    }
+
+    explicit BufferBase(NullBufferParams) {}
```
lioncash, line 81:
```suggestion
    explicit BufferBase(NullBufferParams) = default;
```

src/video_core/buffer_cache/buffer_base.h
```diff
@@ -0,0 +1,480 @@
...
+    explicit BufferBase(RasterizerInterface& rasterizer_, VAddr cpu_addr_, u64 size_bytes_)
+        : rasterizer{&rasterizer_}, cpu_addr{Common::AlignDown(cpu_addr_, BYTES_PER_PAGE)},
+          size_bytes{Common::AlignUp(size_bytes_ + (cpu_addr_ - cpu_addr), BYTES_PER_PAGE)} {
+        if (IsShort()) {
+            cpu_written_words.stack = ~u64{0};
+            gpu_written_words.stack = 0;
+        } else {
+            // Share allocation between CPU and GPU pages and set their default values
+            const size_t num_words = NumWords();
+            u64* const alloc = new u64[num_words * 2];
```
lioncash, line 68: Maybe do this in a way that doesn't require raw new?

ReinUsesLisp: I can wrap it an internal container so the destructor and move methods for `BufferBase` can be defaulted, but I can't avoid the raw new maintaining the same behavior.

src/video_core/buffer_cache/buffer_base.h
```diff
@@ -0,0 +1,480 @@
...
+        u64 page_index = (offset / BYTES_PER_PAGE) % PAGES_PER_WORD;
+        for (u64 word_index = word_begin; word_index < word_end; ++word_index, page_index = 0) {
+            const u64 word = words[word_index] & ~(gpu ? cpu_words[word_index] : 0);
+            if (word == 0) {
+                continue;
+            }
+            const u64 page_end = std::min((word_index + 1) * PAGES_PER_WORD, page_limit);
+            const u64 local_page_end = page_end % PAGES_PER_WORD;
+            const u64 page_end_shift = (PAGES_PER_WORD - local_page_end) % PAGES_PER_WORD;
+            if ((word >> page_index << page_index) << page_end_shift != 0) {
```
lioncash, line 418:
```suggestion
            if (((word >> page_index) << page_index) << page_end_shift != 0) {
```

src/video_core/buffer_cache/buffer_base.h
```diff
@@ -0,0 +1,480 @@
...
+        u64 current_base = 0;
+        u64 current_size = 0;
+        bool on_going = false;
+        for (u64 word_index = word_index_begin; word_index < word_index_end; ++word_index) {
+            const bool is_last_word = word_index + 1 == word_index_end;
+            const u64 page_end = is_last_word ? last_word_page_end : PAGES_PER_WORD;
+            const u64 right_offset = page_begin;
+            const u64 left_offset = PAGES_PER_WORD - page_end;
+            u64 bits = ~u64{0};
+            bits = bits >> right_offset << right_offset;
```
lioncash, line 350: Consider parenthesizing the individual operation order here, ditto below

src/video_core/buffer_cache/buffer_base.h
```diff
@@ -0,0 +1,480 @@
...
+        const u64 end_word_index = Common::DivCeil(end_page_index, PAGES_PER_WORD);
+        u64 page_index = begin_page_index % PAGES_PER_WORD;
+        u64 word_index = begin_word_index;
+        while (word_index < end_word_index) {
+            const u64 next_word_first_page = (word_index + 1) * PAGES_PER_WORD;
+            const u64 left_offset =
+                std::min(next_word_first_page - end_page_index, PAGES_PER_WORD) % PAGES_PER_WORD;
+            const u64 right_offset = page_index;
+            u64 bits = ~u64{0};
+            bits = bits >> right_offset << right_offset;
```
lioncash, line 250: Parenthesizing, ditto below

src/video_core/buffer_cache/buffer_base.h
ReinUsesLisp: It seems to be illegal according to MSVC:
```
'BufferBase' is not a special member function or comparison operator which can be defaulted
```

PR #5267, opened 2020-12-31
main: Make the loader error dialog fully translatable 

lioncash:
Allows the dialog text to be fully translated, as opposed to only the title. While we're at it, we can add annotation comments to explain what the contents of the formatting specifiers are for translators.


FearlessTobi:
Could you please post a screenshot of what the popups look like with your changes now, so they can be compared to the old version?

german77:
Can the link be independent of language? I don't think we want to update 20 languages just to update a link

FearlessTobi:
When the source string is changed, the translation should be invalidated automatically. @german77

lioncash:
@FearlessTobi The popups are going to look the exact same. The difference is that the descriptions can be localized now.

bunnei (approved):

PR #5275, opened 2021-01-01
X86/NativeClock: Improve performance of clock calculations on hot path.

FernandoS27:
This Pull Request introduces a way faster method for transforming time values from native clock (RTDSC) to nanoseconds, microseconds, miliseconds, guest clock cycles and guest cpu cycles.

To reviewers: MulHigh needs to be inlined as the code itself is a hotpath, it's best it remains under an anonymous scope instead of moved to some random module in common.


src/common/x64/native_clock.cpp
```diff
@@ -2,19 +2,73 @@
...
 #else
 #include <x86intrin.h>
 #endif
 
 #include "common/uint128.h"
 #include "common/x64/native_clock.h"
 
+namespace {
+
+[[nodiscard]] u64 getFixedPoint64Factor(u64 numerator, u64 divisor) {
```
lioncash, line 25:
```suggestion
[[nodiscard]] u64 GetFixedPoint64Factor(u64 numerator, u64 divisor) {
```

src/common/x64/native_clock.cpp
```diff
@@ -2,19 +2,73 @@
...
+#ifdef __SIZEOF_INT128__
+    unsigned __int128 base = static_cast<unsigned __int128>(numerator) << 64ULL;
```
lioncash, line 27:
```suggestion
    const auto base = static_cast<unsigned __int128>(numerator) << 64;
```

src/common/x64/native_clock.cpp
```diff
@@ -2,19 +2,73 @@
...
+#else
+    return _udiv128(r[1], r[0], divisor, &remainder);
+#endif
+#else
+    // This one is bit more inaccurate.
+    return MultiplyAndDivide64(std::numeric_limits<u64>::max(), numerator, divisor);
+#endif
+}
+
+inline [[nodiscard]] u64 MulHigh(u64 a, u64 b) {
```
lioncash, line 43:
```suggestion
[[nodiscard]] u64 MultiplyHigh(u64 a, u64 b) {
```

Given we already have functions that use `Multiply` as opposed to `Mul`. `inline` also isn't necessary on an internally linked function

src/common/x64/native_clock.h
```diff
@@ -32,15 +31,31 @@ class NativeClock final : public WallClock {
 private:
     u64 GetRTSC();
 
+    union alignas(16) TimePoint {
+        TimePoint() = default;
```
lioncash, line 35: These two constructors don't need to be provided.

src/common/x64/native_clock.h
```diff
@@ -32,15 +31,31 @@ class NativeClock final : public WallClock {
...
+    TimePoint time_point;
+    // factors
+    u64 clock_rtsc_factor{};
+    u64 cpu_rtsc_factor{};
+    u64 ns_rtsc_factor{};
+    u64 us_rtsc_factor{};
+    u64 ms_rtsc_factor{};
+
     u64 rtsc_frequency;
+
```
lioncash, line 58: Unnecessary newline

src/common/x64/native_clock.h
FernandoS27: MSVC doesn't like them not being there for some reason.

lioncash: If that's the case, then the second constructor should contain a const reference, rather than a non-const reference

bunnei (approved):

PR #5278, opened 2021-01-02
dynarmic: Add Unsafe_InaccurateNaN optimization

MerryMage:
Here are less accurate versions of some things that aren't numbers.

---

In this commit we introduce a new unsafe "optimization", we add an option to remove NaN checks from JITted code. Do note that this means that the NaNs produced will not be bit-perfect with hardware.

Also **please note that this removes some checks for certain special values**, thus this produce incorrect finite results. Here are the two cases where incorect values are wrongly produced when this unsafe option is enabled:
* FRSQRTS will produce a NaN for certain large inputs (where previously it would have produced a value >10^38).
* FMA will sometimes produce a negative zero where previously a positive zero was produced.

I chose to keep NaN checks in FMULX/FMINNM/FMAXNM and related instructions that special case NaN explicitly, since it would be more likely that accuracy is required in these instructions.


goldenx86:
Best description ever.

MerryMage:
Pushed fix for CPUs without FMA: Accidentally introduced a crash while JITting on CPUs without FMA instructions due to a failed assertion.

bunnei (approved):

PR #5208, opened 2020-12-21
Service threads

bunnei:
This change moves HLE service calls from the `CoreTiming` thread to individual host threads allocated per `ServerSession`. This allows us to remove `SleepClientThread`, a messy mechanism used for blocking guest threads from host threads, because now we can just block within service code. These changes are necessary for more broad kernel refactoring that I am working on.

As a consequence, we needed to also change how synchronous GPU emulation works, since there is not a single designated thread for GPU emulation anymore. We now always have a separate GPU thread (like asynchronous GPU), and synchronize on GPU pushes and cache invalidation/flushes. This allows us to decouple async GPU from multicore, meaning both of these settings can be toggled independently.

These changes should help perf a little bit -- Booting games should be faster, as things like filesystem access get their own thread. User input and audio may also have reduced lag, etc.

Overall, this should not change things too much, and is more of a staging change for things to come.

PR #5284, opened 2021-01-03
buffer_queue: Fix data race by protecting queue_sequence access

ameerj:
Fixes a data race introduced by #5208 where `queue_sequence` was accessed and manipulated by multiple threads, but was not protected.

Thanks to @ReinUsesLisp for a reproducible test scenario confirming the data race and its fix.

### DIFF
--- a/src/core/hle/service/nvflinger/buffer_queue.cpp
+++ b/src/core/hle/service/nvflinger/buffer_queue.cpp
@@ -26,10 +26,10 @@ void BufferQueue::SetPreallocatedBuffer(u32 slot, const IGBPBuffer& igbp_buffer)
     LOG_WARNING(Service, "Adding graphics buffer {}", slot);
 
     {
-        std::unique_lock lock{queue_mutex};
+        std::unique_lock lock{free_buffers_mutex};
         free_buffers.push_back(slot);
     }
-    condition.notify_one();
+    free_buffers_condition.notify_one();
 
     buffers[slot] = {
         .slot = slot,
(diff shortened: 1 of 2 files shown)


src/core/hle/service/nvflinger/buffer_queue.cpp
```diff
@@ -64,6 +64,14 @@ std::optional<std::pair<u32, Service::Nvidia::MultiFence*>> BufferQueue::Dequeue
     auto slot = buffers.size();
 
     while (f_itr != free_buffers.end()) {
+        ASSERT(*f_itr < buffers.size());
+        if (*f_itr >= buffers.size()) {
+            LOG_ERROR(Service,
+                      "Asking to dequeue a buffer slot out of range. Slot: {}, Buffers size: {}",
+                      *f_itr, buffers.size());
+            free_buffers.erase(f_itr++); // erase bad slot element and continue
```
lioncash, line 72: Comments should go on top of the line rather than on the side (especially given in this case it applies to more than just this line).

ReinUsesLisp (changes requested):
This is a data race. It has to be fixed with locks or other multithreading primitives, instead of hidden in a log that nobody will bother to fix in a month or so.

ameerj:
@ReinUsesLisp formally addressed the issue 👍

src/core/hle/service/nvflinger/buffer_queue.cpp
```diff
@@ -100,7 +100,10 @@ void BufferQueue::QueueBuffer(u32 slot, BufferTransformFlags transform,
     buffers[slot].crop_rect = crop_rect;
     buffers[slot].swap_interval = swap_interval;
     buffers[slot].multi_fence = multi_fence;
-    queue_sequence.push_back(slot);
+    {
```
ReinUsesLisp, line 103: No need to introduce a new scope here.

ReinUsesLisp (approved):

ReinUsesLisp:
Squash the commits, it looks good to me.

Morph1984 (approved):

PR #5297, opened 2021-01-06
vulkan_memory_allocator: Improvements to the memory allocator

ReinUsesLisp:
This reworks Vulkan's memory allocator functionality and its API, using self-contained RAII types instead of an `unique_ptr`. The detailed explanation is on each individual commit.

Add memory allocation usage hints (download, upload, device local), replacing the old boolean that simply described if its expected to be host visible or not.

Move the code to the `vulkan_common` directory so we can use it from OpenGL.

Rename `Vulkan::VKMemoryManager` to `Vulkan::MemoryAllocator`, IMO a "memory allocator" is a better name for what we are currently doing.

If anyone knows a memory allocation algorithm that can be used here, please let me know. I won't add it on this commit, but it's something I'd like yuzu to have. The current allocator is not really optimized for speed, but for space usage.

This is required for the buffer cache rewrite.


src/video_core/renderer_vulkan/vk_staging_buffer_pool.h
```diff
@@ -9,63 +9,73 @@
...
-
-        std::unique_ptr<VKBuffer> buffer;
+    struct StagingBuffer {
+        vk::Buffer buffer;
+        MemoryCommit commit;
+        std::span<u8> mapped_span;
         u64 tick = 0;
+
+        StagingBufferRef Ref() const noexcept {
+            return StagingBufferRef{
```
lioncash, line 43:
```suggestion
            return {
```

Given the return type is already specified directly above

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -554,10 +554,18 @@ void TextureCacheRuntime::Finish() {
...
-        .handle = *buffer.handle,
-        .map = buffer.commit->Map(size),
+        .handle = staging_ref.buffer,
+        .span = staging_ref.mapped_span,
+    };
+}
+
+ImageBufferMap TextureCacheRuntime::MapDownloadBuffer(size_t size) {
+    const auto staging_ref = staging_buffer_pool.Request(size, MemoryUsage::Download);
+    return ImageBufferMap{
```
lioncash, line 566:
```suggestion
    return {
```

ReinUsesLisp:
Untagging while I investigate some issues, it's unknown if it's a buffer cache issue or a memory allocator bug.

ReinUsesLisp:
Retagging, it was a buffer cache issue that doesn't affect master.
Dirty flags were not working properly for the index buffer so its reported size was not trustworthy. When reusing memory allocations, garbage might have been read from buffers.

PR #5309, opened 2021-01-07
Various input profile changes

Kewlan:
**Functionality**

- The selected profile is saved and loaded between config sessions.
- Removes the save button and instead saves the profiles in a struct and only writes to disk if the configuration is confirmed.
- If multiple players have the same profile selected and you edit button mappings for one of those, it'll change all of them.
- Adds a button to rename profiles.

**Code**

- No longer saves the vibration parameters, as they're applied separately at ConfigureInputPlayer::ApplyConfiguration and when booting games.
- Saves and loads the profiles directly in ConfigureInputPlayer since it need to read the current button mappings, plus there's no more need to have extra checks in the filesystem.
- The default value of PlayerInput::input_profile is currently an empty QString, but if it should be `[none]` I'm thinking it should be a static constexpr char[]. If so, where would be the best way to put it?

**Intentions/Bug testing**
The intended functionality is as mentioned above. Any changes that are made for a player that has an input profile selected is meant to change for all other players that have the same profile selected. I think I've got it working as it should, but if someone want to help bug test they're very welcome to!

### DIFF
--- a/src/yuzu/configuration/configure_input.cpp
+++ b/src/yuzu/configuration/configure_input.cpp
@@ -129,6 +129,12 @@ void ConfigureInput::Initialize(InputCommon::InputSubsystem* input_subsystem,
                 &ConfigureInput::UpdateAllInputDevices);
         connect(player_controllers[i], &ConfigureInputPlayer::RefreshInputProfiles, this,
                 &ConfigureInput::UpdateAllInputProfiles, Qt::QueuedConnection);
+        connect(player_controllers[i], &ConfigureInputPlayer::InputProfileDeleted, this,
+                &ConfigureInput::LoadAllProfiles);
+        connect(player_controllers[i], &ConfigureInputPlayer::InputProfileRenamed, this,
+                &ConfigureInput::RenameInputProfile, Qt::QueuedConnection);
+        connect(player_controllers[i], &ConfigureInputPlayer::PlayerTabSelected, this,
+                &ConfigureInput::TabWindowChanged);
         connect(player_connected[i], &QCheckBox::stateChanged, [this, i](int state) {
             player_controllers[i]->ConnectPlayer(state == Qt::Checked);
         });
@@ -182,7 +188,11 @@ QList<QWidget*> ConfigureInput::GetSubTabs() const {
 }
 
 void ConfigureInput::ApplyConfiguration() {
+    player_controllers[current_tab]->SaveProfile(
+        player_controllers[current_tab]->GetCurrentProfile());
+
     for (auto* controller : player_controllers) {
+        controller->LoadProfile();
         controller->ApplyConfiguration();
         controller->TryDisconnectSelectedController();
     }
@@ -195,6 +205,7 @@ void ConfigureInput::ApplyConfiguration() {
     for (auto* controller : player_controllers) {
         controller->TryConnectSelectedController();
     }
+    profiles->SaveAllProfiles();
 
     advanced->ApplyConfiguration();
 
@@ -273,12 +284,28 @@ void ConfigureInput::UpdateAllInputDevices() {
     }
 }
 
-void ConfigureInput::UpdateAllInputProfiles(std::size_t player_index) {
-    for (std::size_t i = 0; i < player_controllers.size(); ++i) {
-        if (i == player_index) {
-            continue;
-        }
-
-        player_controllers[i]->UpdateInputProfiles();
+void ConfigureInput::UpdateAllInputProfiles() {
+    for (const auto& player : player_controllers) {
+        player->UpdateInputProfiles();
     }
+}
+
+void ConfigureInput::LoadAllProfiles() {
+    for (const auto& player : player_controllers) {
+        player->LoadProfile();
+    }
+}
+
+void ConfigureInput::RenameInputProfile(const QString& old_name, const QString& new_name) {
+    for (const auto& player : player_controllers) {
+        player->RenameSpecifiedProfile(old_name.toStdString(), new_name.toStdString());
+    }
+}
+
+void ConfigureInput::TabWindowChanged(std::size_t new_tab) {
+    player_controllers[current_tab]->SaveProfile(
+        player_controllers[current_tab]->GetCurrentProfile());
+    player_controllers[new_tab]->LoadProfile();
+
+    current_tab = new_tab;
 }
--- a/src/yuzu/configuration/configure_input_player.cpp
+++ b/src/yuzu/configuration/configure_input_player.cpp
@@ -542,12 +542,12 @@ ConfigureInputPlayer::ConfigureInputPlayer(QWidget* parent, std::size_t player_i
 
     connect(ui->buttonProfilesNew, &QPushButton::clicked, this,
             &ConfigureInputPlayer::CreateProfile);
+    connect(ui->buttonProfilesRename, &QPushButton::clicked, this,
+            &ConfigureInputPlayer::RenameProfile);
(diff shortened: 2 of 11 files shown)


src/yuzu/configuration/configure_input.cpp
```diff
@@ -273,12 +284,29 @@ void ConfigureInput::UpdateAllInputDevices() {
...
-        if (i == player_index) {
-            continue;
-        }
-
         player_controllers[i]->UpdateInputProfiles();
     }
 }
+
+void ConfigureInput::LoadAllProfiles() {
+    for (std::size_t i = 0; i < player_controllers.size(); ++i) {
```
lioncash, line 294: This can be a ranged for loop

src/yuzu/configuration/configure_input.cpp
```diff
@@ -273,12 +284,29 @@ void ConfigureInput::UpdateAllInputDevices() {
...
+        player_controllers[i]->LoadProfile();
+    }
+}
+
+void ConfigureInput::RenameInputProfile(QString old_name, QString new_name) {
+    for (std::size_t i = 0; i < player_controllers.size(); ++i) {
```
lioncash, line 300: This can be a ranged for loop

src/yuzu/configuration/configure_input.cpp
```diff
@@ -273,12 +284,29 @@ void ConfigureInput::UpdateAllInputDevices() {
...
     }
 }
+
+void ConfigureInput::LoadAllProfiles() {
+    for (std::size_t i = 0; i < player_controllers.size(); ++i) {
+        player_controllers[i]->LoadProfile();
+    }
+}
+
+void ConfigureInput::RenameInputProfile(QString old_name, QString new_name) {
```
lioncash, line 299:
```suggestion
void ConfigureInput::RenameInputProfile(const QString& old_name, const QString& new_name) {
```

src/yuzu/configuration/configure_input.cpp
```diff
@@ -273,12 +284,29 @@ void ConfigureInput::UpdateAllInputDevices() {
     }
 }
 
-void ConfigureInput::UpdateAllInputProfiles(std::size_t player_index) {
+void ConfigureInput::UpdateAllInputProfiles() {
     for (std::size_t i = 0; i < player_controllers.size(); ++i) {
```
lioncash, line 288: This can now be a ranged for loop

src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -1237,87 +1251,122 @@ void ConfigureInputPlayer::CreateProfile() {
...
-    ui->comboProfiles->setCurrentIndex(-1);
+    if (ui->comboProfiles->findText(QString::fromStdString(current_profile)) < 0) {
+        current_profile = unselected_profile;
+    }
+    ui->comboProfiles->setCurrentText(QString::fromStdString(current_profile));
+}
+
+void ConfigureInputPlayer::RenameSpecifiedProfile(std::string old_name, std::string new_name) {
+    if (current_profile == old_name) {
+        current_profile = new_name;
```
lioncash, line 1368:
```suggestion
        current_profile = std::move(new_name);
```

src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -1237,87 +1251,122 @@ void ConfigureInputPlayer::CreateProfile() {
...
     }
 
-    ui->comboProfiles->setCurrentIndex(-1);
+    if (ui->comboProfiles->findText(QString::fromStdString(current_profile)) < 0) {
+        current_profile = unselected_profile;
+    }
+    ui->comboProfiles->setCurrentText(QString::fromStdString(current_profile));
+}
+
+void ConfigureInputPlayer::RenameSpecifiedProfile(std::string old_name, std::string new_name) {
```
lioncash, line 1366:
```suggestion
void ConfigureInputPlayer::RenameSpecifiedProfile(std::string_view old_name, std::string new_name) {
```

src/yuzu/configuration/configure_input_player.h
```diff
@@ -75,6 +75,18 @@ class ConfigureInputPlayer : public QWidget {
     /// Updates the list of controller profiles.
     void UpdateInputProfiles();
 
+    /// Renames the specified input profile.
+    void RenameSpecifiedProfile(std::string old_name, std::string new_name);
+
+    /// Returns the current profile.
+    std::string GetCurrentProfile();
```
lioncash, line 82:
```suggestion
    std::string GetCurrentProfile() const;
```

src/yuzu/configuration/input_profiles.h
```diff
@@ -8,6 +8,8 @@
 #include <string_view>
 #include <unordered_map>
 
+#include <core/settings.h>
```
lioncash, line 11:
```suggestion
#include "core/settings.h"
```

src/yuzu/configuration/configure_input_player.cpp
lioncash: Perhaps unconditionally change the name, even if it is the same, given, it would remove the need for a parameter and always make the heap allocation for the std::string not go to waste.

With std::move within the function, the expensive part of the rename is always going to be the construction of the std::string, not the logic within the function.

german77:
Seems to work as intended

src/yuzu/configuration/configure_input_player.cpp
Kewlan: All 8 players respond to the signal, so `old_name` is needed so only the player tabs with the same input profile change their names. 

For example: Player 1, 7, and 8 have the same profile and the user renames the profile for player 1, then just these three will rename theirs. (And all players have their profile list updated.)

Kewlan: Disclaimer, I forget pointers quite often. Would making them both `const std::string&` be better?

Its-Rei:
This goes entirely against the input concept that I designed for yuzu. Profiles will be used for per-game input and the controller applet down the line. Overwriting profiles willy nilly like that and potentially screwing up the input for all the games that use this profile is totally nonsensical and will not happen. You were told as much when you suggested such a change when Mjölnir was PR'd.

Morph1984:
> The selected profile is saved and loaded between config sessions.

This is extremely problematic for users. No emulator does this and I do not want users to be confused why their profiles have their mappings changed when profiles are bound to players in the config. 

> If multiple players have the same profile selected and you edit button mappings for one of those, it'll change all of them.

This is also problematic for the issues stated above, and that input profiles are bound to a real controller so using the same profile between multiple players is not feasible.

Kewlan:
Did you two try this PR out, by the way? You can set a player to not use a profile, then it won't update for any other player or profile. Either way it shouldn't get in the way of custom game configurations or the controller applet if properly implemented.

I really don't see what's so confusing. Users will figure out how profiles are saved with no Save button, just how they figure out how to load without a Load button.

> No emulator does this [...]

Citra does, which is what I based this of because I liked it so much.

> This is also problematic for the issues stated above, and that input profiles are bound to a real controller so using the same profile between multiple players is not feasible.

This is, for example, if a user wants to control multiple characters in Smash Bros, then they only have to map one of the players to update all of them.

I would like to hear opinions from other people too, so why not tag this to Early Access and hear what users directly think?

Its-Rei:
Both authors of our input implementation told you that these changes are not welcome so no, this will not be tagged for EA.

Kewlan:
That's really disappointing that you're gonna be like this even though users may like this more.

PR #5319, opened 2021-01-09
cmake: Enforce C4062, C4265, C4388, and C5038

ReinUsesLisp:
This should match some warnings we treat as errors on gcc and clang (-Wswitch and -Wreorder), catching bugs early and reducing the number of instances where we have to edit commits to make CI happy when developing from Windows.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -45,10 +45,14 @@ if (MSVC)
 
         # Warnings
         /W3
+        /we4062 # enumerator 'identifier' in a switch of enum 'enumeration' is not handled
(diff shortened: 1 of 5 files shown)


lioncash (approved):

lioncash:
Nice to know MSVC has an equivalent for -Wreorder, I didn't know about that one =)

ReinUsesLisp:
Me neither, I found it while walking over [Compiler warnings that are off by default](https://docs.microsoft.com/en-us/cpp/preprocessor/compiler-warnings-that-are-off-by-default?view=msvc-160), from MSDN.

PR #5324, opened 2021-01-10
config: Enable docked mode by default

Morph1984:
Enables the docked mode by default.

This has 2 advantages:
1. Makes games run at their best quality/resolution.
2. A workaround for the SMO flickering issue when GLASM (OpenGL Assembly Shaders) is enabled with a NVIDIA GPU in undocked mode.


goldenx86:
Yeah I agree it is time. Players are no longer laser focused on Let's Go.

lioncash (approved):

PR #5327, opened 2021-01-10
qt: Set DISPLAY env var when not present

AniLeo:
Fixes web browser opening (Help > Open Mods Page, Help > Open Quickstart Guide)

```
$ ./yuzu-20210110-a8cc9db87-x86_64.AppImage 
Error: no DISPLAY environment variable specified
```

Same issue as https://github.com/RPCS3/rpcs3/issues/5959, same fix as https://github.com/RPCS3/rpcs3/pull/7461

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -59,6 +59,7 @@ static FileSys::VirtualFile VfsDirectoryCreateFileWrapper(const FileSys::Virtual
 #include <QPushButton>
 #include <QShortcut>
 #include <QStatusBar>
+#include <QString>
 #include <QSysInfo>
 #include <QUrl>
 #include <QtConcurrent/QtConcurrent>
@@ -2997,6 +2998,14 @@ int main(int argc, char* argv[]) {
     // the user folder in the Qt Frontend, we need to cd into that working directory
     const std::string bin_path = Common::FS::GetBundleDirectory() + DIR_SEP + "..";
     chdir(bin_path.c_str());
+#endif
+
+#ifdef __linux__
+    // Set the DISPLAY variable in order to open web browsers
+    // TODO (lat9nq): Find a better solution for AppImages to start external applications
+    if (QString::fromLocal8Bit(qgetenv("DISPLAY")).isEmpty()) {
+        qputenv("DISPLAY", ":0");
+    }
 #endif
 
     // Enables the core to make the qt created contexts current on std::threads


lat9nq:
This does not work if the host session's DISPLAY is not `:0`.

AniLeo:
Is there any scenario where `:0` is not in use while another id is? `:0` is usually the main screen - https://www.xfree86.org/current/X.7.html#sect4

We never got any issue for that since adding this, but I am not aware of another way to go about this

lat9nq:
Well, for example, on my machine I have multiple GPUs inserted for testing. To use the NVIDIA GPUs in the subsequent slots, I need to give them dummy X screens. So the primary screen is `:0.0` and the NVIDIA GPUs get `:0.1` and `:0.2`. Essentially, my X server has multiple screens defined.

I am more than willing to try and change my xorg.conf to fix this issue, but I feel like an AppImage should be adapting to any scenario, not the user changing their machine to use the AppImage. This also wouldn't work on machines with multiple X servers running.

I was digging into this a day or so ago, and I came across something called `appimage-exec-wrapper`, which restores specified environment variables when `execve` and friends are called. It, however, seems to be deprecated. I'm not sure what has happened upstream that necessitates deprecating it.

lat9nq:
Can you confirm whether this is still an issue that needs resolving? I think we can merge it for now with a `// TODO (lat9nq): Find a better solution for starting external applications` if it's still affecting people. I currently cannot reproduce the issue on my machines nor any VMs, however. And since merging it and testing locally doesn't seem to break anything, seems fine to merge (again, if this still affects people).

AniLeo:
It still happens yes, just reproduced on the latest build. Note that this doesn't happen on the native compiled application, only on the AppImage version.

src/yuzu/main.cpp
```diff
@@ -2999,6 +3000,12 @@ int main(int argc, char* argv[]) {
     chdir(bin_path.c_str());
 #endif
 
+#ifdef __linux__
+    // Set the DISPLAY variable in order to open web browsers
```
lat9nq, line 3004:
```suggestion
    // Set the DISPLAY variable in order to open web browsers
    // TODO (lat9nq): Find a better solution for AppImages to start external applications 
```
Can you add this comment that points at me? (Making a suggestion so it's easier to add with GitHub)

src/yuzu/main.cpp
```diff
@@ -2999,6 +3000,12 @@ int main(int argc, char* argv[]) {
...
+    if (QString::fromLocal8Bit(qgetenv("DISPLAY")).isEmpty())
+        qputenv("DISPLAY", ":0");
+#endif
```
lat9nq, line 3007:
```suggestion
    if (QString::fromLocal8Bit(qgetenv("DISPLAY")).isEmpty()) {
        qputenv("DISPLAY", ":0");
    }
#endif
```
Style consistency with the project.

lat9nq (changes requested):

lat9nq (approved):

bunnei (approved):

PR #5328, opened 2021-01-10
Fix:Deriving Keys dialog isn't a child of main

pineappleEA:
What the title suggests, the window "Deriving Keys" isn't set as a child of the main window.
 
Before the fix while the main window has an icon the dialog doesn't cause it doesn't belong to it:
image

After the fix the dialog has an icon cause the main window is the parent:
image


bunnei:

PR #5330, opened 2021-01-11
Fix IP validator error

german77:
Fixes error introduced in #5156. Where the last octet produced an error if the value was higher than 199. For example 192.168.1.200 

This is caused by the break line in the string. Putting the string in a single line fixes the issue.

### DIFF
--- a/src/yuzu/configuration/configure_motion_touch.cpp
+++ b/src/yuzu/configuration/configure_motion_touch.cpp
@@ -185,8 +185,9 @@ void ConfigureMotionTouch::ConnectEvents() {
 }
 
 void ConfigureMotionTouch::OnUDPAddServer() {
-    QRegExp re(tr(R"re(^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4]"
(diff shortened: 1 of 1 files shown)


lioncash:
Your commit should probably explain what the error was to begin with that's being fixed.

src/yuzu/configuration/configure_motion_touch.cpp
```diff
@@ -185,8 +185,9 @@ void ConfigureMotionTouch::ConnectEvents() {
 }
 
 void ConfigureMotionTouch::OnUDPAddServer() {
-    QRegExp re(tr(R"re(^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4]"
-                  "[0-9]|[01]?[0-9][0-9]?)$)re")); // a valid ip address
+    // Validator for IP address
+    QRegExp re(QString::fromStdString(
```
lioncash, line 189:
```suggestion
    QRegExp re(QStringLiteral(
```

lioncash:
Can you add the explanation into the commit message itself whenever you have the chance? You've updated the PR description, but not the commit.

lioncash (approved):

PR #5336, opened 2021-01-12
common/tree: Convert defines over to templates

lioncash:
Migrates the tree header to operate off of templates as opposed to a series of defines.

This allows all tree facilities to obey namespacing rules, and also allows this code to be used within modules once compiler support is in place. This also gets rid to use a macro to define functions and structs for necessary data types. With templates, these will be generated when they're actually used, eliminating the need for the separate declaration.

Now common doesn't dump a large amount of unqualified identifiers into core.


ReinUsesLisp (approved):

bunnei (approved):

bunnei:
Thanks for the cleanup here!

PR #5348, opened 2021-01-15
astc: Make the decoder more robust to invalid data

ReinUsesLisp:
Games can bind invalid textures that make us read or write out of bounds, to avoid exploding on these illegal textures; handle them gracefully with regular robustness techniques.

- Fixes ASTC address sanitizer errors on Astral Chain, and regular crashes when on regular gcc builds.

### DIFF
--- a/src/video_core/textures/astc.cpp
+++ b/src/video_core/textures/astc.cpp
@@ -42,21 +42,24 @@ constexpr u32 Popcnt(u32 n) {
 
 class InputBitStream {
 public:
-    constexpr explicit InputBitStream(const u8* ptr, std::size_t start_offset = 0)
-        : cur_byte{ptr}, next_bit{start_offset % 8} {}
+    constexpr explicit InputBitStream(std::span<const u8> data, size_t start_offset = 0)
+        : cur_byte{data.data()}, total_bits{data.size()}, next_bit{start_offset % 8} {}
 
-    constexpr std::size_t GetBitsRead() const {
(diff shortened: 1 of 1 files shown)


lioncash (approved):

src/video_core/textures/astc.cpp
```diff
@@ -42,21 +42,24 @@ constexpr u32 Popcnt(u32 n) {
...
+        : cur_byte{data.data()}, total_bits{data.size()}, next_bit{start_offset % 8} {}
 
-    constexpr std::size_t GetBitsRead() const {
+    constexpr size_t GetBitsRead() const {
         return bits_read;
     }
 
     constexpr bool ReadBit() {
-        const bool bit = (*cur_byte >> next_bit++) & 1;
+        if (bits_read >= total_bits * 8) {
```
ameerj, line 53: you can compare here with `cur_byte` rather than `bits_read`

src/video_core/textures/astc.cpp
```diff
@@ -898,7 +902,7 @@ static void DecodeColorValues(u32* out, u8* data, const u32* modes, const u32 nP
     // We now have enough to decode our integer sequence.
     IntegerEncodedVector decodedColorValues;
 
-    InputBitStream colorStream(data);
+    InputBitStream colorStream(data, 0);
```
ameerj, line 905: is the 0 necessary if the constructor defaults `start_offset = 0`?

PR #5351, opened 2021-01-15
cmake: Enforce -Wunused-function code-base wise

ReinUsesLisp:
Stops us from merging code with unused functions in the future.

If something is invoked behind conditionally evaluated code in
a way that the language can't see it (e.g. preprocessor macros), the
potentially unused function should use [[maybe_unused]].

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -69,6 +69,7 @@ else()
         -Werror=reorder
         -Werror=switch
         -Werror=uninitialized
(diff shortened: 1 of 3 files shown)


lioncash (approved):

lioncash:
Perhaps this should be enforced code-base wide? I don't think we have too many unused functions elsewhere.

ReinUsesLisp:
Testing if things build locally, then I'll push it.

PR #5358, opened 2021-01-15
common/common_funcs: Rename INSERT_UNION_PADDING_{BYTES,WORDS} to _NOINIT

ReinUsesLisp:
`INSERT_PADDING_BYTES_NOINIT` is more descriptive of the underlying behavior.


src/common/common_funcs.h
```diff
@@ -26,8 +26,8 @@
 
 /// These are similar to the INSERT_PADDING_* macros, but are needed for padding unions. This is
```
lioncash, line 27: This comment should probably be expanded to include why this would also be useful in structs as well

lioncash (approved):

PR #5361, opened 2021-01-15
vk_shader_decompiler: Show comments as OpUndef with a type

ReinUsesLisp:
Silence the new validation layer error about SPIR-V not allowing OpUndef
on a OpTypeVoid, even when the SPIR-V spec doesn't say anything against
it.

They will be inserted as an undefined int to avoid SPIRV-Cross and
validation errors, but only when a debugging tool is attached.

We should use OpString for this in the future.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_shader_decompiler.cpp
+++ b/src/video_core/renderer_vulkan/vk_shader_decompiler.cpp
@@ -1334,7 +1334,10 @@ private:
         }
 
         if (const auto comment = std::get_if<CommentNode>(&*node)) {
-            Name(OpUndef(t_void), comment->GetText());
+            if (device.HasDebuggingToolAttached()) {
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_vulkan/vk_shader_decompiler.cpp
```diff
@@ -1334,7 +1334,9 @@ class SPIRVDecompiler final : public Sirit::Module {
         }
 
         if (const auto comment = std::get_if<CommentNode>(&*node)) {
-            Name(OpUndef(t_void), comment->GetText());
+            if (device.HasDebuggingToolAttached()) {
+                Name(OpUndef(t_int), comment->GetText());
```
lioncash, line 1338: Maybe add a comment or something about using OpString in the future, so that it isn't forgotten about.

bunnei (approved):

PR #5363, opened 2021-01-16
vk_texture_cache: Support image store on sRGB images with VkImageViewUsageCreateInfo

ReinUsesLisp:
Vulkan 1.0 didn't support creating sRGB image views on an ABGR8 VkImage
with storage usage bits. VK_KHR_maintenance2 addressed this allowing to
reduce the usage bits on a VkImageView.

To allow image store on non-sRGB image views when the VkImage is created
with sRGB, always create VkImages without sRGB and add the sRGB format
on the view.

### DIFF
--- a/src/video_core/renderer_vulkan/maxwell_to_vk.h
+++ b/src/video_core/renderer_vulkan/maxwell_to_vk.h
@@ -35,7 +35,15 @@ struct FormatInfo {
     bool storage;
 };
 
-FormatInfo SurfaceFormat(const Device& device, FormatType format_type, PixelFormat pixel_format);
+/**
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_vulkan/maxwell_to_vk.h
```diff
@@ -35,7 +35,15 @@ struct FormatInfo {
     bool storage;
 };
 
-FormatInfo SurfaceFormat(const Device& device, FormatType format_type, PixelFormat pixel_format);
+/**
+ * Returns format properties supported properties in the host
```
v1993, line 39: Typo?

PR #5364, opened 2021-01-16
Fix yuzu-cmd missing _NET_WM_ICON on Linux

pineappleEA:
Just like yuzu-qt, currently yuzu-cmd doesn't have a title bar icon and the generic one appears instead. The reason for this is because the icon implantation is a Windows only one using the resource system of Visual Studio with an .rc file. For this pull request a multi-platform approach was taken that abides by the SDL documentation (https://wiki.libsdl.org/SDL_SetWindowIcon). 

Since SDL natively supports only bitmap images a yuzu.bmp image has been added to the dist folder.

If the image was loaded using the SDL_LoadBMP function, for example:
`SDL_Surface* windowIcon = SDL_LoadBMP("yuzu.bmp");`
yuzu-cmd would expect the file in a relative path on the filesystem separate from the binary which isn't very practical.

In order to embed the bitmap inside the binary it's hex code would need to be extracted into a header file but this isn't optimal either since the header has to be modified each time the icon needs to be changed, for example yuzu has two builds with different icons, Mainline and Early Access that both use different icons.

Since Qt isn't used, no resource manager is available so a CMake function was created in order to manage the assets. Instead of creating the header file yuzu_icon.h on src/yuzu_cmd/ CMake creates when it's run and overwrites it when it's run again.

Here's yuzu-cmd before:
image
As you can see it has the generic icon

And yuzu-cmd after:
image
With an icon

While this would probably work on windows too, the resource system is retained just like the yuzu-qt pr so testing is needed


src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
```diff
@@ -193,6 +194,14 @@ void EmuWindow_SDL2::WaitEvent() {
     }
 }
 
+void EmuWindow_SDL2::SetWindowIcon(){
+ 	SDL_RWops* yuzu_icon_stream = SDL_RWFromMem((void*) yuzu_icon, yuzu_icon_size);
```
ReinUsesLisp, line 198: nit: `SDL_RWFromMem` can potentially fail if the allocation fails, maybe log and abort setting the icon on failure.

src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
```diff
@@ -193,6 +194,14 @@ void EmuWindow_SDL2::WaitEvent() {
...
+    SDL_Surface* windowIcon = SDL_LoadBMP_RW(yuzu_icon_stream, 1);
```
ReinUsesLisp, line 199: nit: Ditto about this potentially failing.

ReinUsesLisp, line 199:
```suggestion
    SDL_Surface* const window_icon = SDL_LoadBMP_RW(yuzu_icon_stream, 1);
```

src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
ReinUsesLisp, line 198:
```suggestion
 	SDL_RWops* const yuzu_icon_stream = SDL_RWFromMem((void*) yuzu_icon, yuzu_icon_size);
```

ReinUsesLisp (commented):
Some small nits.

As a side note, you can use `ninja clang-format` or `make clang-format` to reformat the whole project. Just make sure not to stage unrelated changes, as clang-format changes the formatting between versions, and we use an old one for validation.

bunnei:

PR #5743, opened 2021-01-17
Fix player 1 turning on handheld and not updating handheld settings

german77:
The default connected value when loading the player 1 is true and the default connected value when saving the player 1 was false. Because of this the player 1 was always on, regardless of the config. 

Handheld configuration is a copy of the configuration of player 1 but the configuration only was updated when handheld was turned on for the first time.

This PR addresses both issues. By fixing the default value for player 1 and always updating the configuration for handheld when it's turned on.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -1005,7 +1005,8 @@ void Config::SavePlayerValue(std::size_t player_index) {
                  static_cast<u8>(Settings::ControllerType::ProController));
 
     if (!player_prefix.isEmpty()) {
-        WriteSetting(QStringLiteral("%1connected").arg(player_prefix), player.connected, false);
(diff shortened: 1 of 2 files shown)


src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -897,13 +897,13 @@ void Controller_NPad::UpdateControllerAt(NPadControllerType controller, std::siz
         connected_controllers[HANDHELD_INDEX] = {controller, true};
         InitNewlyAddedController(HANDHELD_INDEX);
         return;
```
v1993, line 899: Is there a need to keep this return now?

german77: Now that I look at it. this change was dumb and will not solve anything. The problem was in the default value of player 1

bunnei (approved):

PR #5746, opened 2021-01-17
texture_cache/util: Resolve -Wsign-compare warning

lioncash:
Resolves a -Wsign-compare warning on Clang.


ReinUsesLisp (approved):
lgtm
Can this constant be made signed? Or does that generate more comparison warnings? In theory all `levels` counters should be signed.

lioncash:
std::array takes a std::size_t as a size specifier, so changing the type would result in the cast just being shifted elsewhere.

PR #5771, opened 2021-01-20
lm: Recode LM service

ogniK5377:
Rework the service to spit out to logs instead of a seperate file as well as fix any crashes caused by lm.


src/core/hle/service/lm/lm.cpp
```diff
@@ -5,22 +5,71 @@
 #include <sstream>
 #include <string>
 
+#include <optional>
+#include <unordered_map>
+#include "boost/container_hash/hash.hpp"
```
lioncash, line 10:
```suggestion
#include <boost/container_hash/hash.hpp>
```

src/core/hle/service/lm/lm.cpp
```diff
@@ -30,54 +79,257 @@ class ILogger final : public ServiceFramework<ILogger> {
...
+            offset++;
+            if (offset >= data.size()) {
+                break;
+            }
+        } while ((data[offset] & 0x80) != 0);
+        return result;
+    }
+
+    std::string ReadString(const std::vector<u8>& data, std::size_t& offset, std::size_t length) {
+        std::string output{};
```
lioncash, line 161:
```suggestion
        std::string output(length, '\0');
```

This removes the need for a separate resize call

src/core/hle/service/lm/lm.cpp
```diff
@@ -30,54 +79,257 @@ class ILogger final : public ServiceFramework<ILogger> {
...
+        while (offset < log_data.size()) {
+            const auto key = static_cast<LogDataChunkKey>(ReadLeb128(log_data, offset));
+            const auto chunk_size = ReadLeb128(log_data, offset);
+
+            switch (key) {
+            case LogDataChunkKey::LogSessionBegin:
+            case LogDataChunkKey::LogSessionEnd:
+                break;
+            case LogDataChunkKey::TextLog:
+                text_log = std::move(ReadString(log_data, offset, chunk_size));
```
lioncash, line 206: std::moves here don't do anything. Strings returned by value from a function will already invoke the move assignment operator. Ditto for all occurrences below.

src/core/hle/service/lm/lm.cpp
```diff
@@ -30,54 +79,257 @@ class ILogger final : public ServiceFramework<ILogger> {
...
+                break;
+            case LogDataChunkKey::ProcessName:
+                process_name = std::move(ReadString(log_data, offset, chunk_size));
+                break;
+            }
+        }
+
+        std::string output_log{};
+        if (process_name && !process_name.value().empty()) {
+            output_log += fmt::format("Process: {}\n", process_name.value());
```
lioncash, line 237:
```suggestion
            output_log += fmt::format("Process: {}\n", *process_name));
```

avoids a validity check, since we know its valid at this point. Ditto for the `valid().empty()` checks as well here and below.

src/core/hle/service/lm/lm.cpp
```diff
@@ -30,54 +79,257 @@ class ILogger final : public ServiceFramework<ILogger> {
...
+            if (offset >= data.size()) {
+                break;
+            }
+        } while ((data[offset] & 0x80) != 0);
+        return result;
+    }
+
+    std::string ReadString(const std::vector<u8>& data, std::size_t& offset, std::size_t length) {
+        std::string output(length, '\0');
+        output.resize(length);
```
lioncash, line 162: This line can be removed.

src/core/hle/service/lm/lm.cpp
```diff
@@ -30,54 +79,264 @@ class ILogger final : public ServiceFramework<ILogger> {
...
+                         output_log);
+            break;
+        default:
+            LOG_CRITICAL(Service_LM, "LogManager UNKNOWN ({}):\n{}",
+                         DestinationToString(destination), output_log);
+            break;
+        }
+    }
+
+    std::string DestinationToString(LogDestination destination) {
```
lioncash, line 293:
```suggestion
    static std::string DestinationToString(LogDestination destination) {
```

Given this doesn't modify class state

src/core/hle/service/lm/lm.cpp
```diff
@@ -5,22 +5,71 @@
...
+enum class LogSeverity : u8 {
+    Trace = 0,
+    Info = 1,
+    Warning = 2,
+    Error = 3,
+    Fatal = 4,
+};
+
+// To keep flags out of hashing as well as the payload size
+struct LogPacketHeaderEntrty {
```
lioncash, line 29:
```suggestion
struct LogPacketHeaderEntry {
```

src/core/hle/service/lm/lm.cpp
```diff
@@ -5,22 +5,71 @@
...
+    u8 verbosity{};
+
+    auto operator<=>(const LogPacketHeaderEntry&) const = default;
+};
+} // namespace Service::LM
+
+namespace std {
+template <>
+struct hash<Service::LM::LogPacketHeaderEntry> {
+    std::size_t operator()(const Service::LM::LogPacketHeaderEntry& k) const {
```
ReinUsesLisp, line 42:
```suggestion
    std::size_t operator()(const Service::LM::LogPacketHeaderEntry& k) const noexcept {
```

bunnei:
Untagging, as this breaks Celeste

bunnei (approved):

PR #5776, opened 2021-01-20
lbl: Implement most of lbl

ogniK5377:
Pretty basic service, only thing left to do is handle setting applying once set:sys is implemented


Morph1984:
Supersedes #2542

bunnei (approved):

PR #5778, opened 2021-01-20
renderer_opengl: Avoid precompiled cache and force NV GL cache directory

ReinUsesLisp:
Setting __GL_SHADER_DISK_CACHE_PATH we can force the cache directory to
be in yuzu's user directory to stop commonly distributed malware from
deleting our driver shader cache. And by setting
__GL_SHADER_DISK_CACHE_SKIP_CLEANUP we can have an unbounded shader
cache size.

This has only been implemented on Windows, mostly because previous tests
didn't seem to work on Linux.

Disable the precompiled cache on Nvidia's driver. There's no need to
hide information the driver already has in its own cache.

Thanks to Exzap from Cemu for telling me what environment flags to set on Windows.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_cache.cpp
@@ -159,6 +159,10 @@ std::unordered_set<GLenum> GetSupportedFormats() {
 
 ProgramSharedPtr BuildShader(const Device& device, ShaderType shader_type, u64 unique_identifier,
                              const ShaderIR& ir, const Registry& registry, bool hint_retrievable) {
+    if (device.UseDriverCache()) {
+        // Ignore hint retrievable if we are using the driver cache
+        hint_retrievable = false;
+    }
     const std::string shader_id = MakeShaderID(unique_identifier, shader_type);
     LOG_INFO(Render_OpenGL, "{}", shader_id);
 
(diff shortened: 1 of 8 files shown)


src/video_core/renderer_opengl/gl_shader_cache.cpp
```diff
@@ -336,7 +340,7 @@ void ShaderCacheOpenGL::LoadDiskCache(u64 title_id, const std::atomic_bool& stop
     }
 
     std::vector<ShaderDiskCachePrecompiled> gl_cache;
-    if (!device.UseAssemblyShaders()) {
+    if (!device.UseAssemblyShaders() && !device.UseDriverCache()) {
         // Only load precompiled cache when we are not using assembly shaders
```
v1993, line 344:
```suggestion
        // Only load precompiled cache when we are not using assembly shaders or when using the driver cache
```

v1993:
On linux you probably must somehow set this variably before Nvidia library loads, which happens *before entering main*. Or you probably can assign environment variable in `dist/yuzu.desktop`.

P.S.: while (Linux only?) Nvidia does have its own driver cache, it often gets invalidated after an update, so recompiling all shaders before starting may still be useful.

ReinUsesLisp:
This change still builds all shaders at boot, it just builds them from GLSL instead of using glProgramBinary.

bunnei (approved):

Issue #5195, opened 2020-12-14
Exiting or using Stop causes yuzu to lock


lat9nq:
Regressed with #4542 

When a guest program is running, yuzu will lock up if the user tries to close yuzu or use Emulation -> {Stop,Restart}. In this state, yuzu needs to receive SIGKILL from an external process to exit.

Backtrace of c96930fd9d09e49355225ecd538510b9db2a7158 running es2gears, after hitting CTRL+C in GDB once yuzu is in a locked state: [gdb.txt](https://github.com/yuzu-emu/yuzu/files/5691172/gdb.txt)
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5691173/yuzu_log.txt)

Disabling Multicore/Async, using a different OpenGL renderer (such as NVIDIA or llvmpipe), switchting to Vulkan, or enabling/disabling Macro JIT does not change its behavior and produces the same backtrace.

goldenx86:
This also affects Windows under most conditions.

lat9nq:
> This also affects Windows under most conditions.

Yup, I reproduced this in Windows. Updating issue title.

gidoly:
Adressed by 7019

PR #5337, opened 2021-01-12
Fix:Yuzu freezing when stopping emulation on Linux

pineappleEA:
Closes #4161

While working on Windows, stopping emulation caused Yuzu to freeze on Linux. That includes closing the emulator window while a game was running, stopping the game running from Emulation>Stop and restarting it from Emulation>Restart. What happened was the threads weren't being notified by the cv before being destructed (ref. https://en.cppreference.com/w/cpp/thread/condition_variable/%7Econdition_variable) so they stayed waiting forever.

image

And the emulator wasn't responding until it was killed.
Before:

https://user-images.githubusercontent.com/67879877/104332812-3f44de80-54f9-11eb-994d-e36acea11f6a.mp4

Now we notify the worker threads before terminating them.
After:

https://user-images.githubusercontent.com/67879877/104333322-c7c37f00-54f9-11eb-918a-d514f91d6633.mp4

Further testing needed, especially on Windows.

PR #5780, opened 2021-01-21
video_core: Use std::jthread and std::stop_token for worker threads

ReinUsesLisp:
Marking this as a draft because `libc++` doesn't yet support `std::jthread` and `std::stop_token`.

Fixes shutdown issues and simplifies the code overall. Default destructors stop worker threads naturally.

### DIFF
--- a/src/common/threadsafe_queue.h
+++ b/src/common/threadsafe_queue.h
@@ -14,7 +14,7 @@
 #include <utility>
 
 namespace Common {
-template <typename T>
+template <typename T, bool with_stop_token = false>
 class SPSCQueue {
 public:
     SPSCQueue() {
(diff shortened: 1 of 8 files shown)


lat9nq:
Fixes #5195

v1993:
Does this supersede PR #5337?

ReinUsesLisp:
Yes, I wasn't aware that PR existed.
This notifies all the condition variables automatically when the `std::stop_token` gets signaled. We don't have to worry about notifying anything, the jthread destructors take care of it.

ReinUsesLisp:
Unmarking this as draft, we already use features libc++ doesn't support.

ReinUsesLisp:
Untagging because our current version of msvc doesn't support it.

Morph1984:
Superseded by #7019

PR #5798, opened 2021-01-23
vulkan_device: Blacklist Intel from float16 math

ReinUsesLisp:
Astral Chain crashes Intel's SPIR-V compiler when using fp16.
Disable this while the vendor works on a fix.

With this changes, Astral Chain renders better on Intel than the other vendors, we can use this to debug further what causes issues there.
It should also be less prone to memory issues due to supporting ASTC natively as the compressed texture format it is.

imagen
(The window title is showing the wrong branch name.)


ReinUsesLisp:
@ReinUsesLisp A rebase is needed.

bunnei (approved):

PR #5799, opened 2021-01-23
nvdrv: Unregister already registered events

ogniK5377:
Fixes softlocks in pokemon lets go caused by Nvdec.
Fixed softlock caused by leon in sword and shield
2021-01-21_58-35-c2154a60-b2e4-461f-9d2b-f91e37929c7d-pL9AH7hq

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
@@ -155,7 +155,14 @@ NvResult nvhost_ctrl::IocCtrlEventRegister(const std::vector<u8>& input, std::ve
         return NvResult::BadParameter;
     }
     if (events_interface.registered[event_id]) {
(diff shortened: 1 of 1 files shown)


src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
```diff
@@ -155,7 +155,14 @@ NvResult nvhost_ctrl::IocCtrlEventRegister(const std::vector<u8>& input, std::ve
         return NvResult::BadParameter;
     }
     if (events_interface.registered[event_id]) {
-        return NvResult::BadParameter;
+        const auto event_state = events_interface.status[event_id];
+        if (event_state == EventState::Registered || event_state == EventState::Waiting ||
```
lioncash, line 159:
```suggestion
        if (event_state != EventState::Free)
```

Would simplify the check a bit more.

bunnei (approved):

PR #5800, opened 2021-01-23
sdl_impl: Set the maximum vibration duration to 1 second

### LINKED ISSUE #5026: Kirby Fighters 2: vibration stuck for a long time
Introduced by and first reported in #4866.

To reproduce:

1. Boot Kirby Fighters 2 (ver. 1.0.0)
2. Proceed to the main screen and go to settings (gear icon)
3. Switch rumble OFF and then back ON

As a result, after quick rumble peak (as expected), controller will keep vibrating slightly for a long time (about 2 mins, I think).

Log: [yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5607281/yuzu_log.txt)

Morph1984:
Some games do not send a 0 amplitude vibration to stop residual vibrations. This sets the maximum duration for any vibration to 1 second.

Fixes #5026

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -81,10 +81,14 @@ public:
     }
 
     bool RumblePlay(u16 amp_low, u16 amp_high) {
+        constexpr u32 rumble_max_duration_ms = 1000;
+
         if (sdl_controller) {
-            return SDL_GameControllerRumble(sdl_controller.get(), amp_low, amp_high, 0) == 0;
+            return SDL_GameControllerRumble(sdl_controller.get(), amp_low, amp_high,
+                                            rumble_max_duration_ms) == 0;
         } else if (sdl_joystick) {
-            return SDL_JoystickRumble(sdl_joystick.get(), amp_low, amp_high, 0) == 0;
+            return SDL_JoystickRumble(sdl_joystick.get(), amp_low, amp_high,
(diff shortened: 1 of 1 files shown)


src/input_common/sdl/sdl_impl.cpp
```diff
@@ -82,9 +82,9 @@ class SDLJoystick {
 
     bool RumblePlay(u16 amp_low, u16 amp_high) {
         if (sdl_controller) {
-            return SDL_GameControllerRumble(sdl_controller.get(), amp_low, amp_high, 0) == 0;
+            return SDL_GameControllerRumble(sdl_controller.get(), amp_low, amp_high, 1000) == 0;
```
lioncash, line 85: Perhaps move the constant to a local `constexpr u32 rumble_max_duration_ms = 1000;`

Just so it's not a magic variable and also makes the time unit evident.

lioncash (approved):

v1993:
Just wondering - is this a workaround or an accurate solution that replicates behavior of Switch?

german77:
Normally you will expect the game to send a zero vibration signal. If it fails to do that it will keep vibrating forever. This is a workaround to at least prevent that.

v1993:
That's exactly what I was wondering - if it is a workaround (like it seems) or if it is also done on Switch the same way.

Codexmsters:
#5026

PR #5807, opened 2021-01-24
video_core: Silence the remaining gcc warnings and enforce them

ReinUsesLisp:
Silence all the remaining warnings in video_core generated by gcc 10.2.0. Enforce -Warray-bounds and -Wmissing-field-initializers.


lioncash (approved):

lioncash:
I'm fairly sure these two could be globally applied to our code without a lot of effort

PR #5808, opened 2021-01-24
host_shaders/cmake: Pass --quiet to glslang to keep it quiet

ReinUsesLisp:
Silence noisy builds on toolchains.


ReinUsesLisp:
I remember searching for this flag before, but I couldn't find it.

lioncash (approved):

PR #5809, opened 2021-01-24
audout: FlushAudioOutBuffers

ogniK5377:
Fixes Devil May Cry
unknown (1)

### DIFF
--- a/src/audio_core/stream.cpp
+++ b/src/audio_core/stream.cpp
@@ -51,6 +51,14 @@ void Stream::Stop() {
     UNIMPLEMENTED();
 }
 
+bool Stream::Flush() {
+    const bool had_buffers = !queued_buffers.empty();
+    while (!queued_buffers.empty()) {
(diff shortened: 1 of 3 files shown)


lioncash (approved):

src/audio_core/stream.cpp
```diff
@@ -51,6 +51,14 @@ void Stream::Stop() {
     UNIMPLEMENTED();
 }
 
+bool Stream::Flush() {
+    const bool had_buffers = !queued_buffers.empty();
+    while (!queued_buffers.empty()) {
+        queued_buffers.pop();
```
v1993, line 57: This only pops a topmost buffer and not all of them. Is this correct?

bunnei (approved):

v1993:
@bunnei What about my proposed change? Is there a reason not to do it or should I send it in as a separate PR?

PR #5815, opened 2021-01-24
Network error handling reform

comex:
`network.cpp` has several error paths which either:
- report "Unhandled host socket error=n" and return `SUCCESS`, or
- switch on a few possible errors, log them, and translate them to
  Errno; the same switch statement is copied and pasted in multiple
  places in the code

Convert these paths to use a helper function `GetAndLogLastError`, which
is roughly the equivalent of one of the switch statements, but:
- handling more cases (both ones that were already in `Errno`, and a few
  more I added), and
- using OS functions to convert the error to a string when logging, so
  it'll describe the error even if it's not one of the ones in the
  switch statement.
  - To handle this, refactor the logic in `GetLastErrorMsg` to expose
    new functions `WinErrorToString` and `UnixErrnoToString` which take
    the error number explicitly as an argument.  And improve the Windows
    version a bit.

Also, add a test which exercises two random error paths.

### DIFF
--- a/src/common/common_funcs.h
+++ b/src/common/common_funcs.h
@@ -52,8 +52,12 @@ __declspec(dllimport) void __stdcall DebugBreak(void);
 // Generic function to get last error message.
 // Call directly after the command or use the error num.
 // This function might change the error code.
-// Defined in Misc.cpp.
+// Defined in misc.cpp.
 [[nodiscard]] std::string GetLastErrorMsg();
+
+// Like GetLastErrorMsg(), but passing an explicit error code.
+// Defined in misc.cpp.
+[[nodiscard]] std::string NativeErrorToString(int e);
 
 #define DECLARE_ENUM_FLAG_OPERATORS(type)                                                          \
     [[nodiscard]] constexpr type operator|(type a, type b) noexcept {                              \
--- a/src/core/network/network.h
+++ b/src/core/network/network.h
@@ -7,6 +7,7 @@
 #include <array>
 #include <utility>
 
+#include "common/common_funcs.h"
 #include "common/common_types.h"
 
 namespace Network {
@@ -21,6 +22,11 @@ enum class Errno {
     MFILE,
     NOTCONN,
     AGAIN,
+    CONNREFUSED,
+    HOSTUNREACH,
+    NETDOWN,
+    NETUNREACH,
+    OTHER,
 };
 
 /// Address families
--- a/src/tests/core/network/network.cpp
+++ b/src/tests/core/network/network.cpp
@@ -1,0 +1,28 @@
+// Copyright 2021 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include <catch2/catch.hpp>
+
+#include "core/network/network.h"
+#include "core/network/sockets.h"
+
+TEST_CASE("Network::Errors", "[core]") {
+    Network::NetworkInstance network_instance; // initialize network
+
+    Network::Socket socks[2];
+    for (Network::Socket& sock : socks) {
+        REQUIRE(sock.Initialize(Network::Domain::INET, Network::Type::STREAM,
+                                Network::Protocol::TCP) == Network::Errno::SUCCESS);
+    }
+
+    Network::SockAddrIn addr{
+        Network::Domain::INET,
+        {127, 0, 0, 1},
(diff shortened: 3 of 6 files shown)


src/common/common_funcs.h
```diff
@@ -55,6 +55,14 @@ __declspec(dllimport) void __stdcall DebugBreak(void);
 // Defined in Misc.cpp.
 [[nodiscard]] std::string GetLastErrorMsg();
 
+// Like GetLastErrorMsg(), but passing an explicit error code.
+// Defined in Misc.cpp.
+#ifdef _WIN32
+[[nodiscard]] std::string WinErrorToString(int e);
```
ReinUsesLisp, line 61: Both functions share the same interface. Can we squash them into a single unified function? Even when `e` isn't portable.
Something among the lines of: `NativeErrorToString`?

src/core/network/network.h
```diff
@@ -7,6 +7,7 @@
 #include <array>
 #include <utility>
 
+#include "common/common_funcs.h" // DECLARE_ENUM_FLAG_OPERATORS
```
ReinUsesLisp, line 10: We don't generally specify the reason why a header is included.

src/tests/core/network/network.cpp
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 yuzu Emulator Project
```
ReinUsesLisp, line 1: Commit date doesn't match here.

src/common/common_funcs.h
comex: I considered that but thought they'd be cleaner as separate functions… but all right.

src/tests/core/network/network.cpp
comex: Fixed.

src/core/network/network.h
comex: Removed.

bunnei:
This is causing some build issues:

```
##[error]src\common\misc.cpp(19,17): Error C2664: 'DWORD FormatMessageA(DWORD,LPCVOID,DWORD,DWORD,LPSTR,DWORD,va_list *)': cannot convert argument 5 from 'LPTSTR' to 'LPSTR'
##[error]src\common\misc.cpp(26,29): Error C2664: 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>::basic_string(std::initializer_list<_Elem>,const _Alloc &)': cannot convert argument 1 from 'LPTSTR' to 'std::initializer_list<_Elem>'
```

bunnei:
@comex ping to take a look at this when you get a chance

comex:
Well, I finally got around to setting up a Yuzu build on Windows/MSVC, at which point fixing the error was trivial (`err_str` needed to be `LPSTR` rather than `LPTSTR`).  Not sure why that wasn't just caught by CI…  Does the CI Windows build not have `_UNICODE` defined, so `LPSTR` is equivalent to `LPTSTR` there when it isn't on MSVC?  Or was the CI configuration different month ago when I submitted the original version of this? \*shrug\*

comex:
Yep, sure enough, Yuzu's CMakeLists.txt explicitly does
```
    # Ensure that projects build with Unicode support.
    add_definitions(-DUNICODE -D_UNICODE)
```
on MSVC, but not on MinGW.

I tried adding the same definitions for MinGW and the build almost works fine, except that `tests.exe` fails to link due to missing `WinMain`.  Not sure if that's worth pursuing; is there a reason MinGW doesn't have them?  In any case, it shouldn't make much difference.  As far as I can tell:
- `UNICODE` primarily affects whether you get the FooW or FooA variant of Win32 functions if you ask only for Foo, but Yuzu tends to explicitly add the W or A suffix.  
- `_UNICODE` (with the underscore) affects the definition of `TCHAR`, `TSTR`, etc., but Yuzu doesn't use those (except that the original version of this PR mistakenly added a use).

bunnei:
The CI here and what we use to generate releases are indeed different. We try to keep them the same (or very close), but sometimes things get out of sync. Adding the UNICODE and _UNICODE definitions should not be necessary, that is probably carryover from Citra (and conversely some Dolphin common code that we still use). We could pursue removing these entirely, and being explicit in the few places that we actually use Win32-specific APIs.

Anyways, thanks for fixing this -- I'll tag it for our release pipeline and see how it goes.

bunnei (approved):

bunnei:
LGTM, thanks!

PR #5816, opened 2021-01-24
gpu_thread: Disable flush on read on normal GPU accuracy

ReinUsesLisp:
This is an unfortunate oversight when adding more GPU accuracies. It
makes no sense to be more accurate (and fail to do so) in normal
accuracy than in high.

I plan to merge Normal and High in the same setting in the future.


ReinUsesLisp:
I'm not tagging this for now, as I've only tested a few games.

ReinUsesLisp:
This causes issues on some titles, I'll drop this until I can merge both settings.

PR #5820, opened 2021-01-25
shader_ir: int flags part 2

behunin:
Adding Immediate Node to the search function as it is needed by Super Smash Bros.
Sadly I don't have this game so I can't test.
I set the log level to info for testing, I will change it to debug once we know this is working.


ReinUsesLisp:
@behunin I appreciate your contributions, although we can't accept changes that do not reflect the behavior seen in hardware. There are some issues with this PR, but the overall approach taken to implement the internal flags is wrong here; hence why I'm closing the pull request and reverting the original changes in #5823.

PR #5824, opened 2021-01-25
psm: IPsmSession

ogniK5377:
Used by homebrew menu

### DIFF
--- a/src/core/hle/service/ptm/psm.cpp
+++ b/src/core/hle/service/ptm/psm.cpp
@@ -117,7 +117,7 @@ private:
 
 class PSM final : public ServiceFramework<PSM> {
 public:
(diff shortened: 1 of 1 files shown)


src/core/hle/service/ptm/psm.cpp
```diff
@@ -70,6 +181,7 @@ class PSM final : public ServiceFramework<PSM> {
 
     u32 battery_charge_percentage{100}; // 100%
     ChargerType charger_type{ChargerType::RegularCharger};
+    Core::System& system;
```
lioncash, line 184: You don't need to use a separate variable here. It's already passed in via the ServiceFramework call in the constructor initializer list as a protected variable.

bunnei (approved):

Issue #5291, opened 2021-01-04
Crash when opening Web Applet


aleasto:
On linux (Fedora 33), after #5042 yuzu crashes when trying to open the Action Guide in SMO, or How To Play in SMBU.
Happens both when compiling with and without YUZU_USE_QT_WEB_ENGINE.
Reverting the merge commit stops the crash:
 - If compiling with YUZU_USE_QT_WEB_ENGINE the browser opens and works (I have no issues with inputs)
 - If compiling without, it immediatly gets back to the game (though i thought it was supposed to show a warning dialog, it doesn't)

lat9nq:
I'm able to reproduce the issue here on ace8a8e86ebff443b1bc594516e4168626943896.

[gdb-ace8a8e86.txt](https://github.com/yuzu-emu/yuzu/files/5770924/gdb-ace8a8e86.txt)

Morph1984:
Something of note is that it crashes attempting to extract the shared fonts https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/service/ns/pl_u.cpp#L69

Morph1984:
@aleasto Could you comment out this line to see if it works? https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/service/am/applets/web_browser.cpp#L241

myman2343:
hi i am having the same issue where whenever the how to play menu is open my game crashes saying 66%webapplet. i am able to load things like training mode and world of light without much issue but whenever i open smash, since i have a new account the how to play screen opens up and my game crashes. Im not really sure how to fix this since im on windows. could you help me?

aleasto:
> @aleasto Could you comment out this line to see if it works? https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/service/am/applets/web_browser.cpp#L241

Oh, that stops the crashing when building without YUZU_USE_QT_WEB_ENGINE.
Building with YUZU_USE_QT_WEB_ENGINE=ON it now crashes later (i think?), but only when **not** running in gdb :thinking:

PR #5825, opened 2021-01-25
[HACK] applets/web_browser: Temporarily disable font extraction when QtWebEngine is disabled

Morph1984:
Extracting certain fonts (OSS NintendoExtended fonts, Nintendo system firmware fonts) are causing crashes on linux based systems. These crashes are not observed on Windows builds built with MSVC. Temporarily disable font extraction while a fix is investigated.

**DO NOT MERGE THIS INTO MASTER**

If anyone has a solution to this issue, feel free to PR the fix

Linked issue: #5291


Morph1984:
Fixed by #5878

PR #5830, opened 2021-01-26
Fix Icon Theme issues on some Linux Desktop Environments/Window Managers

### LINKED ISSUE #5269: Yuzu displayes the bundled in icon theme assets only if XDG_CURRENT_DESKTOP is set to a gtk de
As the title suggests, if you try to use yuzu in kde where XDG_CURRENT_DESKTOP defaults to value KDE yuzu will look for installed system themes and it will look like this

image

If you set the env to a random value, for example:
`export XDG_CURRENT_DESKTOP=test`

Then yuzu will look like this:
image

Of course the errors `QPixmap::scaled: Pixmap is a null pixmap` are printed for every asset requested.

If you do `export XDG_CURRENT_DESKTOP=gnome` or lxde or mate (it's not case sensitive gNoMe will work too), no errors are printed and yuzu looks normal

image

This doesn't fix the _NET_WM_ICON missing since `setWindowIcon` doesn't exist in the source code.

pineappleEA:
Closes #5269 

While working on some GTK desktop environments (like GNOME), the icon theme was broken for most DEs/WMs.

yuzu on KDE:
image
Falling back to the Breeze icon theme while printing
`Icon theme ":/icons/colorful" not found.`
and many `QPixmap::scaled: Pixmap is a null pixmap` warnings.

yuzu on LXQt
image
The fallback is hicolor which didn't provide the folder icon so it looks even worse.

A temporary workaround was to export `QT_QPA_PLATFORMTHEME=gtk3` which made yuzu look like it did in GNOME.

The reason was that yuzu on a Qt Desktop was looking for the theme ":/icons/colorful" (using Light Colorful as the example) which is a path instead of "colorful" which is the name of the theme. Since the method `QIcon::themeSearchPaths();` already returns `:/icons` the path was axed from the `QIcon::setThemeName();` method. For this to work the two colorful dark themes were changed to inherit the icons from the colorful themes.

Using a custom debug build that prints the theme names and paths we can see more clearly what the Qt gui does when run normally on a Qt DE and when run with the env
image
 
When the env is set, Fallback Theme is null and the Search Paths are scattered differently.

According to https://doc.qt.io/qt-5/qicon.html#fallbackThemeName:
"On X11, if not set, the fallback icon theme depends on your desktop settings. On other platforms it is not set by default."
so perhaps a fallback theme being set "steals" the priority from the :/icons path from the Search Paths.

After all this, yuzu looks nice:
image

Another issue resolved by this PR plagued GTK DE users made QComboBox huge when it was hovered on the Midnight Blue themes.

The reason was that the checked item was Bold. Since the other themes don't utilize bold and this wasn't visible on Windows it was removed.

Bellow a screenshot from KDE
image
(For some reason the angles are missing on KDE)

And screenshots in a GTK DE
Before:
image

After:
image

Testing on Windows required.

### DIFF
--- a/dist/qt_themes/colorful_dark/icons/index.theme
+++ b/dist/qt_themes/colorful_dark/icons/index.theme
@@ -1,7 +1,7 @@
 [Icon Theme]
 Name=colorful_dark
 Comment=Colorful theme (Dark style)
-Inherits=default
+Inherits=colorful
 Directories=16x16
  
 [16x16]
--- a/dist/qt_themes/colorful_midnight_blue/icons/index.theme
+++ b/dist/qt_themes/colorful_midnight_blue/icons/index.theme
@@ -1,7 +1,7 @@
 [Icon Theme]
 Name=colorful_midnight_blue
 Comment=Colorful theme (Midnight Blue style)
-Inherits=default
+Inherits=colorful
 Directories=16x16
 
 [16x16]
--- a/dist/qt_themes/qdarkstyle_midnight_blue/style.qss
+++ b/dist/qt_themes/qdarkstyle_midnight_blue/style.qss
@@ -1257,10 +1257,6 @@ QComboBox::item:alternate {
   background: #19232D;
 }
 
-QComboBox::item:checked {
-  font-weight: bold;
-}
-
 QComboBox::item:selected {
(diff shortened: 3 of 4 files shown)


bunnei:

PR #5832, opened 2021-01-27
basic async addition for audio renderer (not for all parts yet)

16-Bit-Dog:
A mix of async thread work and keeping dedicated thread time to a non-async process to reduce latency of a task: slight reduction in time to complete audio rendering

This was a small addition; more complex changes are going to be done later...

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -205,7 +205,7 @@ ResultCode AudioRenderer::UpdateAudioRenderer(const std::vector<u8>& input_param
         ReleaseAndQueueBuffers(); 
     });
 
-    QueueAudioBufferFence1.get();
+    keepThreadReady1.get();
 
     return RESULT_SUCCESS;
 }
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -1,9 +1,10 @@
-// Copyright 2018 yuzu Emulator Project
+// Copyright 2020 yuzu Emulator Project
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
 #include <limits>
 #include <vector>
+#include <future>
 
 #include "audio_core/audio_out.h"
 #include "audio_core/audio_renderer.h"
@@ -68,6 +69,11 @@ namespace {
 } // namespace
 
 namespace AudioCore {
+
+std::vector <std::future<void>> queueMixedThreadFence;
+std::future<void> keepThreadReady1;
+std::future<void> keepThreadReady2;
+    
 AudioRenderer::AudioRenderer(Core::Timing::CoreTiming& core_timing, Core::Memory::Memory& memory_,
                              AudioCommon::AudioRendererParameter params,
                              Stream::ReleaseCallback&& release_callback,
@@ -195,7 +201,11 @@ ResultCode AudioRenderer::UpdateAudioRenderer(const std::vector<u8>& input_param
         return AudioCommon::Audren::ERR_INVALID_PARAMETERS;
     }
 
-    ReleaseAndQueueBuffers();
+    keepThreadReady1 = std::async(std::launch::async, [&] { 
+    ReleaseAndQueueBuffers(); 
+    });
+
+    QueueAudioBufferFence1.get();
 
     return RESULT_SUCCESS;
 }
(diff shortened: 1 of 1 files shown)


16-Bit-Dog:
oof, gotta fix that format of code to follow clang

src/audio_core/audio_renderer.cpp
```diff
@@ -68,6 +69,11 @@ namespace {
 } // namespace
 
 namespace AudioCore {
+
+std::vector<std::future<void>> queueMixedThreadFence;
```
lioncash, line 73: We don't use global variables

src/audio_core/audio_renderer.cpp
```diff
@@ -236,6 +248,7 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
             mix_buffers[i] =
                 command_generator.GetMixBuffer(in_params.buffer_offset + buffer_offsets[i]);
         }
+        queueMixedThreadFence[queueMixedThreadFence.size() - 1].get();
```
lioncash, line 251:
```suggestion
        queueMixedThreadFence.back().get();
```

src/audio_core/audio_renderer.cpp
```diff
@@ -313,11 +326,16 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
     elapsed_frame_count++;
     voice_context.UpdateStateByDspShared();
 }
-
 void AudioRenderer::ReleaseAndQueueBuffers() {
+
     const auto released_buffers{audio_out->GetTagsAndReleaseBuffers(stream)};
+
+    queueMixedThreadFence.resize(0); // instead of passing a s16 to the queue mixed buffer to
```
lioncash, line 333: Comments go on top of code rather than beside it

src/audio_core/audio_renderer.cpp
```diff
@@ -313,11 +326,16 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
     elapsed_frame_count++;
     voice_context.UpdateStateByDspShared();
 }
-
 void AudioRenderer::ReleaseAndQueueBuffers() {
+
```
lioncash, line 330: Unnecessary newline

src/audio_core/audio_renderer.cpp
```diff
@@ -313,11 +326,16 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
     elapsed_frame_count++;
     voice_context.UpdateStateByDspShared();
 }
-
```
lioncash, line 316: Newline should be left in here

src/audio_core/audio_renderer.cpp
```diff
@@ -195,7 +201,9 @@ ResultCode AudioRenderer::UpdateAudioRenderer(const std::vector<u8>& input_param
         return AudioCommon::Audren::ERR_INVALID_PARAMETERS;
     }
 
-    ReleaseAndQueueBuffers();
+    keepThreadReady1 = std::async(std::launch::async, [&] { ReleaseAndQueueBuffers(); });
+
+    keepThreadReady1.get();
```
lioncash, line 206: What is the purpose of this async if it's being created and then immediately being blocked on?

src/audio_core/audio_renderer.cpp
```diff
@@ -313,11 +326,16 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
...
+
     const auto released_buffers{audio_out->GetTagsAndReleaseBuffers(stream)};
+
+    queueMixedThreadFence.resize(0); // instead of passing a s16 to the queue mixed buffer to
+                                     // control, pushing values to a vector seemed about as fast
+
     for (const auto& tag : released_buffers) {
-        QueueMixedBuffer(tag);
+        keepThreadReady2 = std::async(std::launch::async, [&] { QueueMixedBuffer(tag); });
+        keepThreadReady2.get();
```
lioncash, line 338: ditto

ogniK5377:
I would suggest setting the async audio to be more closer to a mailbox type of implementation too much functionality of the switches fw

16-Bit-Dog:
thank you for all of the recommendations and pointers; I'll close the PR now to fix it up and implement more (have an idea for a interesting barrier-mailbox system I already thought up which I think may work well)

PR #5835, opened 2021-01-27
hid: Cleanup SixAxisFusionParameters

Morph1984:
Inserts padding in `Parameters` struct where applicable and removes unused device handle function parameters

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -520,6 +520,7 @@ void Hid::EnableSixAxisSensorFusion(Kernel::HLERequestContext& ctx) {
         Controller_NPad::DeviceHandle sixaxis_handle;
         u64 applet_resource_user_id;
     };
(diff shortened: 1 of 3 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -564,6 +565,7 @@ void Hid::GetSixAxisSensorFusionParameters(Kernel::HLERequestContext& ctx) {
     IPC::RequestParser rp{ctx};
     struct Parameters {
         Controller_NPad::DeviceHandle sixaxis_handle;
+        INSERT_PADDING_WORDS_NOINIT(1);
```
lioncash, line 568: if anything it would probably be better to place a static assert for size after the struct, rather than inserting explicit padding.

lioncash: This still applies, wrt the padding

lioncash (approved):

PR #5837, opened 2021-01-28
sockets: Stub GetSockOpt

german77:
Stubs GetSocktOpt by returning not connected error, needed for Scott Pilgrim vs The World. Allows to proceed in game.

### DIFF
--- a/src/core/hle/service/sockets/bsd.cpp
+++ b/src/core/hle/service/sockets/bsd.cpp
@@ -255,6 +255,21 @@ void BSD::GetSockName(Kernel::HLERequestContext& ctx) {
     rb.Push<u32>(static_cast<u32>(write_buffer.size()));
 }
 
+void BSD::GetSockOpt(Kernel::HLERequestContext& ctx) {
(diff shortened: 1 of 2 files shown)


src/core/hle/service/sockets/bsd.cpp
```diff
@@ -255,6 +255,21 @@ void BSD::GetSockName(Kernel::HLERequestContext& ctx) {
     rb.Push<u32>(static_cast<u32>(write_buffer.size()));
 }
 
+void BSD::GetSockOpt(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+    const s32 fd = rp.Pop<s32>();
+    const u32 level = rp.Pop<u32>();
+    const OptName optname = static_cast<OptName>(rp.Pop<u32>());
```
lioncash, line 262:
```suggestion
    const auto optname = static_cast<OptName>(rp.Pop<u32>());
```

since the type is evident on the rhs.

bunnei (approved):

PR #5862, opened 2021-02-01
Kernel Rework: Refactor KEvent/KReadableEvent/KWritableEvent

bunnei:
With this change, we recode KEvent/KReadableEvent/KWriteableEvent to more closely match the real Switch kernel. Summary of changes:
* Implement `KEvent` as a kernel primitive, instead of using the `EventPair` struct we previously used.
* Rename `ReadableEvent` and `WritableEvent` to `KReadableEvent` and `KWritableEvent`.
* Update `KReadableEvent` and `KWritableEvent` to more closely match the real kernel.

Unlike my previous kernel rework changes, this one is more "housekeeping" in nature and should not really change current emulation behavior.


src/common/common_funcs.h
```diff
@@ -104,6 +104,14 @@ __declspec(dllimport) void __stdcall DebugBreak(void);
         }                                                                                          \
     }
 
+/// Evaluates a boolean expression, and returns a result unless that expression is true.
+#define R_UNLESS_NOLOG(expr, res)                                                                  \
```
lioncash, line 108: Maybe remove this, given this PR only uses it in two places

src/common/scope_exit.h
```diff
@@ -31,6 +31,8 @@ ScopeExitHelper<Func> ScopeExit(Func&& func) {
 }
 } // namespace detail
 
+#define SCOPE_GUARD(body) detail::ScopeExit([&]() body)
```
lioncash, line 34: Missing a documentation comment

src/core/hle/kernel/k_event.h
```diff
@@ -0,0 +1,49 @@
...
+    std::string GetTypeName() const override {
+        return "KEvent";
+    }
+
+    static constexpr HandleType HANDLE_TYPE = HandleType::Event;
+    HandleType GetHandleType() const override {
+        return HANDLE_TYPE;
+    }
+
+    std::shared_ptr<KReadableEvent>& GetReadableEvent() {
```
lioncash, line 35: There should also be a:

```cpp
const std::shared_ptr<KReadableEvent>& GetReadableEvent() const
```

overload for both of these.

src/yuzu/debugger/wait_tree.h
```diff
@@ -19,8 +19,8 @@ class EmuThread;
 namespace Kernel {
 class HandleTable;
 class KSynchronizationObject;
+class KReadableEvent;
```
lioncash, line 22: This should go above KSynchronizationObject

PR #5864, opened 2021-02-01
arm_dynarmic_32: Print out CPSR.T on exception

MerryMage:
Eases debugging if we know the processor is in thumb mode.

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic_32.h
+++ b/src/core/arm/dynarmic/arm_dynarmic_32.h
@@ -50,6 +50,10 @@ public:
     u64 GetTPIDR_EL0() const override;
     void ChangeProcessorID(std::size_t new_core_id) override;
 
+    bool IsInThumbMode() const {
+        return (GetPSTATE() & 0x20) != 0;
+    }
+
     void SaveContext(ThreadContext32& ctx) override;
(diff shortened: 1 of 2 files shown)


src/core/arm/dynarmic/arm_dynarmic_32.h
```diff
@@ -50,6 +50,10 @@ class ARM_Dynarmic_32 final : public ARM_Interface {
     u64 GetTPIDR_EL0() const override;
     void ChangeProcessorID(std::size_t new_core_id) override;
 
+    bool IsInThumbMode() {
```
lioncash, line 53:
```suggestion
    bool IsInThumbMode() const {
```

src/core/arm/dynarmic/arm_dynarmic_32.h
```diff
@@ -50,6 +50,10 @@ class ARM_Dynarmic_32 final : public ARM_Interface {
...
+        return static_cast<bool>(GetPSTATE() & 0x20);
```
lioncash, line 54:
```suggestion
        return (GetPSTATE() & 0x20) != 0;
```

lioncash (approved):

PR #5868, opened 2021-02-02
Prevent over scheduling audio events and add motion update unschedule event

german77:
Similar to #5861 audio stream can schedule events in negative time. This can be noticeable when you pause the game for a long period of time and then unpause it. The game will stutter and lag for quite some time until the audio event get's back to positive time.

This PR also adds a missing UnscheduleEvent for the motion update event.

### DIFF
--- a/src/audio_core/stream.cpp
+++ b/src/audio_core/stream.cpp
@@ -103,7 +103,14 @@ void Stream::PlayNextBuffer(std::chrono::nanoseconds ns_late) {
 
     sink_stream.EnqueueSamples(GetNumChannels(), active_buffer->GetSamples());
 
-    core_timing.ScheduleEvent(GetBufferReleaseNS(*active_buffer) - ns_late, release_event, {});
+    const auto buffer_release_ns = GetBufferReleaseNS(*active_buffer);
+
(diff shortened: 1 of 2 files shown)


src/audio_core/stream.cpp
```diff
@@ -103,6 +103,11 @@ void Stream::PlayNextBuffer(std::chrono::nanoseconds ns_late) {
 
     sink_stream.EnqueueSamples(GetNumChannels(), active_buffer->GetSamples());
 
+    // If ns_late is higher than the update rate ignore the delay
+    if (ns_late > GetBufferReleaseNS(*active_buffer)) {
```
lioncash, line 107: `GetBufferReleaseNS(*active_buffer)` should be stored to a const local variable, and then be used here and in the ScheduleEvent call, so that no redundant work is being done.

ogniK5377:
it might be worth looking at other events that rely on ns_late. We could have some issues elsewhere which do the same thing

german77:
I did take a look. So far only audio and input did this. Din't got triggered from other places but that doesn't mean is imposible

bunnei (approved):

PR #5871, opened 2021-02-04
k_address_arbiter: Minor cleanup

lioncash:
Removes some dead code and simplifies other code. While we're in the area, we can introduce a few specific logging calls.


bunnei:
@lioncash needs a rebase when you get a chance.

lioncash:
@bunnei done

bunnei (approved):

PR #5877, opened 2021-02-04
 kernel: More accurately utilize resource_limit

ameerj:
This PR aims to more accurately utilize the `resource_limit` to reserve and release resources closer to the way the hardware does.

There is one major inaccuracy which has not been addressed yet, which is the `Process`'s resource limit usage. Currently, each new `Process` creates its own `KResourceLimit`, when it should instead use the system wide resource limit. 
This is critical particularly for `Process::GetTotalPhysicalMemoryAvailable()`, which must return a very specific result, but this requires complete accuracy with the kernel resource reservation, and currently differs in its accuracy from one title to the next. 

This leads me to believe that the size reported as required for each title may be innacurate. For now, Processes continue to create their own resource limit until it can be further investigated and RE'd.

### DIFF
--- a/src/core/hle/kernel/k_scoped_resource_reservation.h
+++ b/src/core/hle/kernel/k_scoped_resource_reservation.h
@@ -1,0 +1,67 @@
+// Copyright 2021 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+// This file references various implementation details from Atmosphere, an open-source firmware for
+// the Nintendo Switch. Copyright 2018-2020 Atmosphere-NX.
+
+#pragma once
+
+#include "common/common_types.h"
+#include "core/hle/kernel/k_resource_limit.h"
+#include "core/hle/kernel/process.h"
+
+namespace Kernel {
+
+class KScopedResourceReservation {
+public:
+    explicit KScopedResourceReservation(std::shared_ptr<KResourceLimit> l, LimitableResource r,
+                                        s64 v, s64 timeout)
+        : resource_limit(std::move(l)), value(v), resource(r) {
+        if (resource_limit && value) {
+            success = resource_limit->Reserve(resource, value, timeout);
+        } else {
+            success = true;
+        }
+    }
+
+    explicit KScopedResourceReservation(std::shared_ptr<KResourceLimit> l, LimitableResource r,
+                                        s64 v = 1)
+        : resource_limit(std::move(l)), value(v), resource(r) {
+        if (resource_limit && value) {
+            success = resource_limit->Reserve(resource, value);
+        } else {
+            success = true;
+        }
+    }
+
+    explicit KScopedResourceReservation(const Process* p, LimitableResource r, s64 v, s64 t)
+        : KScopedResourceReservation(p->GetResourceLimit(), r, v, t) {}
+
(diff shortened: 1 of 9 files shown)


src/core/hle/kernel/k_scoped_resource_reservation.h
```diff
@@ -0,0 +1,65 @@
...
+#include "core/hle/kernel/k_resource_limit.h"
+#include "core/hle/kernel/process.h"
+
+namespace Kernel {
+
+class KScopedResourceReservation {
+public:
+    KScopedResourceReservation(std::shared_ptr<KResourceLimit> l, LimitableResource r, s64 v,
+                               s64 timeout)
+        : resource_limit(l), value(v), resource(r) {
```
lioncash, line 20:
```suggestion
        : resource_limit(std::move(l)), value(v), resource(r) {
```

src/core/hle/kernel/k_scoped_resource_reservation.h
```diff
@@ -0,0 +1,65 @@
...
+        if (resource_limit && value) {
+            success = resource_limit->Reserve(resource, value, timeout);
+        } else {
+            success = true;
+        }
+    }
+
+    KScopedResourceReservation(std::shared_ptr<KResourceLimit> l, LimitableResource r, s64 v = 1)
+        : resource_limit(l), value(v), resource(r) {
```
lioncash, line 29:
```suggestion
        : resource_limit(std::move(l)), value(v), resource(r) {
```

src/core/hle/kernel/k_scoped_resource_reservation.h
```diff
@@ -0,0 +1,65 @@
...
+
+#include "common/common_types.h"
+#include "core/hle/kernel/k_resource_limit.h"
+#include "core/hle/kernel/process.h"
+
+namespace Kernel {
+
+class KScopedResourceReservation {
+public:
+    KScopedResourceReservation(std::shared_ptr<KResourceLimit> l, LimitableResource r, s64 v,
```
lioncash, line 18: Maybe mark the constructors as `explicit`

src/core/hle/kernel/k_scoped_resource_reservation.h
```diff
@@ -0,0 +1,65 @@
...
+            success = true;
+        }
+    }
+
+    KScopedResourceReservation(const Process* p, LimitableResource r, s64 v, s64 t)
+        : KScopedResourceReservation(p->GetResourceLimit(), r, v, t) {}
+    KScopedResourceReservation(const Process* p, LimitableResource r, s64 v = 1)
+        : KScopedResourceReservation(p->GetResourceLimit(), r, v) {}
+
+    ~KScopedResourceReservation() {
```
lioncash, line 42:
```suggestion
    ~KScopedResourceReservation() noexcept {
```

src/core/hle/kernel/k_scoped_resource_reservation.h
```diff
@@ -0,0 +1,65 @@
...
+            resource_limit->Release(resource, value);
+        }
+    }
+
+    /// Commit the resource reservation, destruction of this object does not release the resource
+    void Commit() {
+        resource_limit = nullptr;
+    }
+
+    bool Succeeded() const {
```
lioncash, line 54:
```suggestion
    [[nodiscard]] bool Succeeded() const {
```

bunnei (approved):

PR #5878, opened 2021-02-05
pl_u: Fix read out of bounds

### LINKED ISSUE #5291: Crash when opening Web Applet

aleasto:
Closes #5291 #5825 

Few things to note:
* Please verify that behaviour is unchanged on windows. On linux the web applet currently doesn't work anyways.
* I removed `transformed_font[1] = Common::swap32(transformed_font[1]) ^ KEY; // "re-encrypt" the size` cause I can only assume it was a wrong copy-paste from above.

* ~~Weirdly enough removing that line (or any transformed_font[x] = y line) makes the compiler warn with~~

  >   [1/5] Building CXX object src/core/CMakeFiles/core.dir/hle/service/ns/pl_u.cpp.o
  >   ../src/core/hle/service/ns/pl_u.cpp: In function ‘void Service::NS::DecryptSharedFontToTTF(const std::vector&, std::vector&)’:
  >   ../src/core/hle/service/ns/pl_u.cpp:73:16: warning: ‘void* memcpy(void*, const void*, size_t)’ specified bound 18446744073709551608 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
 
  ~~False positive?~~

### DIFF
--- a/src/core/hle/service/ns/pl_u.cpp
+++ b/src/core/hle/service/ns/pl_u.cpp
@@ -65,13 +65,18 @@ static void DecryptSharedFont(const std::vector<u32>& input, Kernel::PhysicalMem
 void DecryptSharedFontToTTF(const std::vector<u32>& input, std::vector<u8>& output) {
     ASSERT_MSG(input[0] == EXPECTED_MAGIC, "Failed to derive key, unexpected magic number");
 
+    if (input.size() < 2) {
+        LOG_ERROR(Service_NS, "Input font is empty");
+        return;
+    }
+
     const u32 KEY = input[0] ^ EXPECTED_RESULT; // Derive key using an inverse xor
     std::vector<u32> transformed_font(input.size());
     // TODO(ogniK): Figure out a better way to do this
     std::transform(input.begin(), input.end(), transformed_font.begin(),
                    [&KEY](u32 font_data) { return Common::swap32(font_data ^ KEY); });
-    transformed_font[1] = Common::swap32(transformed_font[1]) ^ KEY; // "re-encrypt" the size
-    std::memcpy(output.data(), transformed_font.data() + 2, transformed_font.size() * sizeof(u32));
+    std::memcpy(output.data(), transformed_font.data() + 2,
+                (transformed_font.size() - 2) * sizeof(u32));
 }
 
 void EncryptSharedFont(const std::vector<u32>& input, std::vector<u8>& output,


aleasto:
@Morph1984 Hi, please review this patch

ogniK5377:
Re-encrypting the key is correct behavior, the OOB issue seems to be the input size is too small, which should be investigated instead

german77:
What do you think it will happen if transformed_font.size() is less than 2? Just a hint size() is an unsigned int

aleasto:
> Re-encrypting the key is correct behavior, the OOB issue seems to be the input size is too small, which should be investigated instead

The first 2 ints were not copied back to the output array, so changing the second int doesn't matter. This is exporting to TTF which is probably why the author of this change skipped the first 2 bytes in the memcpy. Thus the line can be removed.
And no, the issue is not that input size is too small, the oob is on the input array (its transformed, rather, which shares the size)

> What do you think it will happen if transformed_font.size() is less than 2? Just a hint size() is an unsigned int
The memcpy already copies from transformed_font.data() + 2.
Great tone btw :+1:

german77:
-2 in unsigned int can be 18446744073709551608 or 4294967294 depends if the system is 64 or 32 bits. I din't mean to sound rude. Just wanted to explain the compiler warning.

aleasto:
> -2 in unsigned int can be 18446744073709551608 or 4294967294 depends if the system is 64 or 32 bits. I din't mean to sound rude. Just wanted to explain the compiler warning.

Ack.
I added size checks, could be turned into asserts at will but it's not technically fatal

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -65,13 +65,16 @@ static void DecryptSharedFont(const std::vector<u32>& input, Kernel::PhysicalMem
 void DecryptSharedFontToTTF(const std::vector<u32>& input, std::vector<u8>& output) {
     ASSERT_MSG(input[0] == EXPECTED_MAGIC, "Failed to derive key, unexpected magic number");
 
+    if (input.size() < 2)
```
lioncash, line 68: We brace if statements, even if there's only one line that constitutes the "body" of it for consistency.

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -65,13 +65,17 @@ static void DecryptSharedFont(const std::vector<u32>& input, Kernel::PhysicalMem
 void DecryptSharedFontToTTF(const std::vector<u32>& input, std::vector<u8>& output) {
     ASSERT_MSG(input[0] == EXPECTED_MAGIC, "Failed to derive key, unexpected magic number");
 
+    if (input.size() < 2) {
+        return;
```
ogniK5377, line 69: Add a LOG_ERROR print to indicate when this fails

Morph1984:
@aleasto Thanks for the patch! I was about to verify the generated files but forgot to lol, will get to it soon

src/core/hle/service/ns/pl_u.cpp
```diff
@@ -65,13 +65,18 @@ static void DecryptSharedFont(const std::vector<u32>& input, Kernel::PhysicalMem
 void DecryptSharedFontToTTF(const std::vector<u32>& input, std::vector<u8>& output) {
     ASSERT_MSG(input[0] == EXPECTED_MAGIC, "Failed to derive key, unexpected magic number");
 
+    if (input.size() < 2) {
+        LOG_ERROR(Common, "Input font is empty");
```
Morph1984, line 69:
```suggestion
        LOG_ERROR(Service_NS, "Input font is empty");
```

Morph1984 (approved):
Generated fonts are verified to be correct, thank you!

PR #5880, opened 2021-02-05
cmake: FFmpeg linking rework

lat9nq:
This adds [FFmpeg](https://ffmpeg.org/) as a submodule and builds/statically links to it on Linux. MSVC is unchanged for the most part. MinGW now uses the same libraries as MSVC, thus fixing the dynamic linking issues with [Martchus](https://github.com/Martchus/PKGBUILDs)'s FFmpeg libraries.

When configuring the submodule on Linux, it disables almost everything from FFmpeg and compiles exactly what yuzu needs. ("Exactly" as far as I know, maybe more can be disabled.) Compile times shouldn't take too much of a hit since yuzu uses only a bit of FFmpeg.

Ports FindFFmpeg.cmake from ctira-emu/citra with minor modifications (see relevant commit). This in turn fixes a couple odd warnings/errors during CMake configuration.

Adds the variable `YUZU_USE_BUNDLED_FFMPEG` to instruct CMake on where to look for FFmpeg. At the moment, it's default to ON. Before this is merged, I want to set it to OFF by default for Linux and ON for WIN32. I'm initially PR'ing this as ON everywhere to let it pass CI that way. Please do not merge until I switch that.

Necessary for future work planned by @ameerj.

### DIFF
--- a/src/video_core/command_classes/codecs/codec.cpp
+++ b/src/video_core/command_classes/codecs/codec.cpp
@@ -13,7 +13,7 @@
 #include "video_core/memory_manager.h"
 
 extern "C" {
-#include "libavutil/opt.h"
+#include <libavutil/opt.h>
 }
 
 namespace Tegra {
--- a/src/video_core/command_classes/codecs/codec.cpp
+++ b/src/video_core/command_classes/codecs/codec.cpp
@@ -13,7 +13,7 @@
 #include "video_core/memory_manager.h"
 
 extern "C" {
-#include <libavutil/opt.h>
+#include "libavutil/opt.h"
 }
 
 namespace Tegra {
--- a/.ci/scripts/windows/docker.sh
+++ b/.ci/scripts/windows/docker.sh
@@ -42,3 +42,8 @@ done
 pip3 install pefile
 python3 .ci/scripts/windows/scan_dll.py package/*.exe "package/"
 python3 .ci/scripts/windows/scan_dll.py package/imageformats/*.dll "package/"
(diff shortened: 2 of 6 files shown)


src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -13,7 +13,7 @@
 #include "video_core/memory_manager.h"
 
 extern "C" {
-#include <libavutil/opt.h>
+#include "libavutil/opt.h"
```
lioncash, line 16: Is this change necessary? We always use `<>` for external libraries.

lat9nq: Nah, it's not. I'll fix that.

lat9nq:
Requires yuzu-emu/build-environments#26: we need an assembler to compile FFmpeg on linux-fresh.

lat9nq:
If yuzu-emu/ext-windows-bin#17 gets merged before this PR, I can update this one to point to the new binaries.

lat9nq:
FFmpeg submodule version is set to 4.3.1, and WIN32 now uses the recently uploaded FFmpeg 4.3.1 binaries from yuzu-emu/ext-windows-bin. The external is now configured to be a fallback solution to the system-installed version on Linux. `YUZU_USE_BUNDLED_FFMPEG` is OFF by default on Linux and ON on WIN32. With all of that, it's ready for a final review.

ameerj (approved):

PR #5881, opened 2021-02-06
Simple async addition to audio_renderer.cpp

16-Bit-Dog:
simple addition of std::async for asynchronous processing of a function call in ReleaseAndQueueBuffers; I wrote down specifics and justification in my branch's PR (Point being that I can add a comment of my justification) - **this is not a full rework of the audio core, this is just a minor performance improvement** (though it would be compatible with dedicating a thread to the audio_core for no viability loss)

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -321,10 +321,11 @@ void AudioRenderer::ReleaseAndQueueBuffers() {
     const auto released_buffers{audio_out->GetTagsAndReleaseBuffers(stream)};
     queue_mixed_multithread.resize(released_buffers.size());
     for (const auto& tag : released_buffers) {
-        queue_mixed_multithread[thread_counter] = std::async(std::launch::async, [=, voice_context = voice_context, 
-          splitter_context = splitter_context, mix_context = mix_context] { 
-            QueueMixedBuffer(tag, released_buffers.size(), thread_counter+1);
-        });
+        queue_mixed_multithread[thread_counter] = std::async(
+            std::launch::async, [=, voice_context = voice_context, 
+                                 splitter_context = splitter_context, mix_context = mix_context] { 
+                QueueMixedBuffer(tag, released_buffers.size(), thread_counter+1);
+            });
         thread_counter++;
     }
     for (std::size_t thread = 0; thread < released_buffers.size(); thread++) {
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -317,12 +317,12 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
 }
 
 void AudioRenderer::ReleaseAndQueueBuffers() {
-    s16 thread_counter = 0;
+    std::size_t thread_counter = 0;
     const auto released_buffers{audio_out->GetTagsAndReleaseBuffers(stream)};
     queue_mixed_multithread.resize(released_buffers.size());
     for (const auto& tag : released_buffers) {
         queue_mixed_multithread[thread_counter] = std::async(
-            std::launch::async, [=, voice_context = voice_context, splitter_context = splitter_context,
+            std::launch::async, [this, tag, voice_context = voice_context, splitter_context = splitter_context,
                                  mix_context = mix_context] { QueueMixedBuffer(tag); });
         thread_counter++;
     }
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -329,7 +329,7 @@ void AudioRenderer::ReleaseAndQueueBuffers() {
         thread_counter++;
     }
     for (std::size_t thread = 0; thread < released_buffers.size(); thread++) {
-        queue_mixed_multithread[thread_to_refrence].wait();
+        queue_mixed_multithread[thread].wait();
     }
 }
 
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -322,10 +322,8 @@ void AudioRenderer::ReleaseAndQueueBuffers() {
     queue_mixed_multithread.resize(released_buffers.size());
     for (const auto& tag : released_buffers) {
         queue_mixed_multithread[thread_counter] = std::async(
-            std::launch::async, [=, voice_context = voice_context,
-                                 splitter_context = splitter_context, mix_context = mix_context] {
-                QueueMixedBuffer(tag);
-            });
+            std::launch::async, [=, voice_context = voice_context, splitter_context = splitter_context,
+                                 mix_context = mix_context] { QueueMixedBuffer(tag); });
         thread_counter++;
     }
     for (std::size_t thread = 0; thread < released_buffers.size(); thread++) {
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -202,8 +202,7 @@ ResultCode AudioRenderer::UpdateAudioRenderer(const std::vector<u8>& input_param
     return RESULT_SUCCESS;
 }
 
-void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag, s16 buffer_max,
-                                     s16 current_thread) {
+void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag, s16 buffer_max, s16 current_thread) {
     command_generator.PreCommand();
     // Clear mix buffers before our next operation
     command_generator.ClearMixBuffers();
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -321,10 +321,10 @@ void AudioRenderer::ReleaseAndQueueBuffers() {
     const auto released_buffers{audio_out->GetTagsAndReleaseBuffers(stream)};
     queue_mixed_multithread.resize(released_buffers.size());
     for (const auto& tag : released_buffers) {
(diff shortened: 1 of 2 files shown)


src/audio_core/audio_renderer.cpp
```diff
@@ -315,9 +318,18 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
...
+    queue_mixed_multithread.resize(released_buffers.size());
     for (const auto& tag : released_buffers) {
-        QueueMixedBuffer(tag);
+        queue_mixed_multithread[thread_counter] = std::async(std::launch::async, [=, voice_context = voice_context, 
+          splitter_context = splitter_context, mix_context = mix_context] { 
+            QueueMixedBuffer(tag, released_buffers.size(), thread_counter+1);
+        });
+        thread_counter++;
+    }
+    for (std::size_t thread_to_refrence = 0; thread_to_refrence < released_buffers.size(); thread_to_refrence++) {
```
lioncash, line 331:
```suggestion
    for (std::size_t thread = 0; thread < released_buffers.size(); thread++) {
```

would make this a tiny bit nicer to grok

16-Bit-Dog: I would also need to change values inside the loop*

german77:
I recommend squashing your commits to only the meaningful ones. Visual studio can help with clang

16-Bit-Dog:
> I recommend squashing your commits to only the meaningful ones. Visual studio can help with clang

not a bad idea

src/audio_core/audio_renderer.cpp
```diff
@@ -315,9 +317,17 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
 }
 
 void AudioRenderer::ReleaseAndQueueBuffers() {
+    s16 thread_counter = 0;
```
v1993, line 320:
```suggestion
    size_t thread_counter = 0;
```

src/audio_core/audio_renderer.cpp
```diff
@@ -315,9 +317,17 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
...
     const auto released_buffers{audio_out->GetTagsAndReleaseBuffers(stream)};
+    queue_mixed_multithread.resize(released_buffers.size());
     for (const auto& tag : released_buffers) {
-        QueueMixedBuffer(tag);
+        queue_mixed_multithread[thread_counter] = std::async(
+            std::launch::async, [=, voice_context = voice_context, splitter_context = splitter_context,
```
v1993, line 325: I think explicitly capturing what you need would be a better option. Especially if it captures `this`, which should always be captured explicitly.

src/audio_core/audio_renderer.cpp
```diff
@@ -315,9 +317,17 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
...
+                                 mix_context = mix_context] { QueueMixedBuffer(tag); });
+        thread_counter++;
+    }
+    for (std::size_t thread = 0; thread < released_buffers.size(); thread++) {
+        queue_mixed_multithread[thread].wait();
     }
```
v1993, line 331: How about calling `queue_mixed_multithread.clear()`? It will clean up memory used by closures while keeping vector's capacity unchanged, so no extra reallocations here.

v1993 (changes requested):

ogniK5377:
Should be implemented as a mailbox sending commands to the DSP

src/audio_core/audio_renderer.cpp
16-Bit-Dog: 👍

src/audio_core/audio_renderer.cpp
16-Bit-Dog: I'm sorry, but I don't completely under stand the suggestion here; since I am using futures, the thread will throw an exception if I clear the future before the thread is finished (referring to how you recommend .clear over highlighted content - if that was not you intention, I wrote another reply for what I think would be the other possibility below) 

if you meant that I should include .clear() after the highlighted code (I tested this right about now to be certain): .clear() would still add unneeded overhead by affecting the residing futures (at least from what I profiled, it has a meaningful enough overhead). Futures can be reused at any point in time once the current thread is done with the job assigned, so I really do not need to .clear() anyways.

If I am misunderstanding anything, I am sorry to ask - but could you reexplain if this is the case.

16-Bit-Dog:
> 
> 
> Should be implemented as a mailbox sending commands to the DSP

is this a TODO (I ask this since from what I understand the DSP is going to be handled by 1 thread)?

ogniK5377:
> > Should be implemented as a mailbox sending commands to the DSP
> 
> is this a TODO (I ask this since from what I understand the DSP is going to be handled by 1 thread)?

It's what intended behavior is as it's what the hardware does. It has a mailbox on device which handles the processing of commands, once commands are done processing the dsp signals the service that more commands are ready for submission. That's really the only multithreading that can be done which will be intended and not cause issues as it's what the hardware does

src/audio_core/audio_renderer.cpp
16-Bit-Dog: 👍

16-Bit-Dog:
> 
> 
> > > Should be implemented as a mailbox sending commands to the DSP
> > 
> > 
> > is this a TODO (I ask this since from what I understand the DSP is going to be handled by 1 thread)?
> 
> It's what intended behavior is as it's what the hardware does. It has a mailbox on device which handles the processing of commands, once commands are done processing the dsp signals the service that more commands are ready for submission. That's really the only multithreading that can be done which will be intended and not cause issues as it's what the hardware does

hm, I see... the DSP for audio is not (or not fully) implemented yet?

ogniK5377:
> > > > Should be implemented as a mailbox sending commands to the DSP
> > > 
> > > 
> > > is this a TODO (I ask this since from what I understand the DSP is going to be handled by 1 thread)?
> > 
> > 
> > It's what intended behavior is as it's what the hardware does. It has a mailbox on device which handles the processing of commands, once commands are done processing the dsp signals the service that more commands are ready for submission. That's really the only multithreading that can be done which will be intended and not cause issues as it's what the hardware does
> 
> hm, I see... the DSP for audio is not (or not fully) implemented yet?

DSP is implemented, the mailbox and multithreaded side of it isn't. Refer to 
https://github.com/yuzu-emu/yuzu/blob/61bf850f3dfd0b44aa5d56f6f3147d7a1fa47353/src/audio_core/command_generator.cpp

&

https://github.com/yuzu-emu/yuzu/blob/61bf850f3dfd0b44aa5d56f6f3147d7a1fa47353/src/audio_core/audio_renderer.cpp#L204-L220

16-Bit-Dog:
> 
> 
> > > > > Should be implemented as a mailbox sending commands to the DSP
> > > > 
> > > > 
> > > > is this a TODO (I ask this since from what I understand the DSP is going to be handled by 1 thread)?
> > > 
> > > 
> > > It's what intended behavior is as it's what the hardware does. It has a mailbox on device which handles the processing of commands, once commands are done processing the dsp signals the service that more commands are ready for submission. That's really the only multithreading that can be done which will be intended and not cause issues as it's what the hardware does
> > 
> > 
> > hm, I see... the DSP for audio is not (or not fully) implemented yet?
> 
> DSP is implemented, the mailbox and multithreaded side of it isn't. Refer to
> https://github.com/yuzu-emu/yuzu/blob/61bf850f3dfd0b44aa5d56f6f3147d7a1fa47353/src/audio_core/command_generator.cpp
> 
> &
> 
> https://github.com/yuzu-emu/yuzu/blob/61bf850f3dfd0b44aa5d56f6f3147d7a1fa47353/src/audio_core/audio_renderer.cpp#L204-L220

thank you for the answer!

16-Bit-Dog:
This is just a more generic question with the audio renderer: since everything inside UpdateAudioRenderer works based off of the previous result (nothing can be done out of order), and the most taxing part is ReleaseAndQueueBuffers, is multi threading just referring to checking for and submitting the DSP calls?

ogniK5377 (changes requested):
As this PR has many issues with it not being implemented properly, no actual performance metrics are given, and non-squashed commits I'm going to have this blocked for the time being. I suggest implementing based on https://github.com/yuzu-emu/yuzu/pull/5881#issuecomment-774446023 so it's hardware accurate and doesn't cause threading issues or audio issues as well as providing the benefit.

I also would like to recommend running clang-format locally to make sure your files are formatted

16-Bit-Dog:
> 
> 
> As this PR has many issues with it not being implemented properly, no actual performance metrics are given, and non-squashed commits I'm going to have this blocked for the time being. I suggest implementing based on [#5881 (comment)](https://github.com/yuzu-emu/yuzu/pull/5881#issuecomment-774446023) so it's hardware accurate and doesn't cause threading issues or audio issues as well as providing the benefit.
> 
> I also would like to recommend running clang-format locally to make sure your files are formatted

16-Bit-Dog:
> 
> 
> As this PR has many issues with it not being implemented properly, no actual performance metrics are given, and non-squashed commits I'm going to have this blocked for the time being. I suggest implementing based on [#5881 (comment)](https://github.com/yuzu-emu/yuzu/pull/5881#issuecomment-774446023) so it's hardware accurate and doesn't cause threading issues or audio issues as well as providing the benefit.
> 
> I also would like to recommend running clang-format locally to make sure your files are formatted

(mis-clicked which wanted to quote) - I fully understand (also I'll put together some data, since I agree my words have less value than proven data... obviously)

PR #5889, opened 2021-02-07
video_core: Delete morton

ogniK5377:
moron.h & morton.cpp are not used anywhere and are just empty files


goldenx86:
Morton antivirus.

lioncash (approved):

adityaruplaha:
moron :kek:

PR #5890, opened 2021-02-08
kernel: Unify result codes

ogniK5377:
Drop the usage of ERR_NAME convention in kernel for ResultName. Removed seperation between svc_results.h & errors.h as we mainly include both most of the time anyways.


lioncash:
The `Svc::` prefixes on the ones in svc.cpp are already done in #5872 (just a heads up so conflicts can be avoided)

bunnei (approved):

PR #5903, opened 2021-02-10
configure_input_player_widget: Silence unused variable warnings

lat9nq:
Prevents Clang 11 from throwing an error since these variables are unused.


german77:
If up_arrow isn't used delete the full array. No need to keep it anymore.

lioncash (approved):

PR #5917, opened 2021-02-13
vulkan_device: Require VK_EXT_robustness2

ReinUsesLisp:
We are already using robustness2 features without requiring it
explicitly, causing potential crashes on drivers without the extension.
Requiring this at boot allows better diagnostics for it and formalizes
our usage on the extension.


goldenx86:
Update drivers, _or else_.

lioncash (approved):

PR #5919, opened 2021-02-13
gl_stream_buffer/vk_staging_buffer_pool: Fix size check

ReinUsesLisp:
Fix a tragic off-by-one condition that causes Vulkan's stream buffer to
think it's always full, using fallback memory. The OpenGL was also
affected by this bug to a lesser extent.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_master_semaphore.h
+++ b/src/video_core/renderer_vulkan/vk_master_semaphore.h
@@ -21,12 +21,12 @@ public:
 
     /// Returns the current logical tick.
     [[nodiscard]] u64 CurrentTick() const noexcept {
-        return current_tick;
+        return current_tick.load(std::memory_order_relaxed);
     }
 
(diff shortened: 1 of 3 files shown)


ReinUsesLisp:
This popped out with a ~6% of CPU usage on Visual Studio's profiler on Super Mario Odyssey.

src/video_core/renderer_vulkan/vk_master_semaphore.h
```diff
@@ -24,6 +24,11 @@ class MasterSemaphore {
         return current_tick;
     }
 
+    /// Returns the last known GPU tick.
+    [[nodiscard]] u64 KnownGpuTick() const noexcept {
+        return gpu_tick;
```
ReinUsesLisp, line 29: This should be marked as relaxed.

lioncash (approved):

Morph1984 (approved):

PR #5977, opened 2021-02-21
acc: Stub GetNintendoAccountUserResourceCacheForApplication

Morph1984:
This command returns a Nintendo Account ID and writes 2 output buffers. The first output buffer is a NasUserBaseForApplication and the second output buffer is currently empty.

Used by:
- Pokken Tournament DX
- Super Smash Bros. Ultimate
- Super Nintendo Entertainment System - Nintendo Switch Online
- Mario Kart 8 Deluxe


lioncash:
Maybe include what title requires this being stubbed in the commit message. LGTM otherwise

PR #5980, opened 2021-02-21
[OpenGL] Signed atomic min/max, texelFetch offsets

Kelebek1:
I didn't know how to really implement the atomics. I feel like this implementation is terrible, so if anyone has any ideas on how to better do it, let me know. I didn't want to always include the helpers as part of every shader, and I also needed mutliple versions depending on the variables used. 

I decided to just log if a particular helper was needed, and for atomic in paticular, the memory types seen during decompile. Forward-declare the functions at the top, and then implement them after decompiling the shader so I knew what was needed. I tried to make it in such a way that new helper functions can be added more easily in the future.

As part of that, I decided to rename shared memory from smem to shared_mem, to match the assembly shaders and Vulkan.

### DIFF
--- a/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
+++ b/src/video_core/renderer_opengl/gl_shader_decompiler.cpp
@@ -79,6 +79,11 @@ const float fswzadd_modifiers_a[] = float[4](-1.0f,  1.0f, -1.0f,  0.0f );
 const float fswzadd_modifiers_b[] = float[4](-1.0f, -1.0f,  1.0f, -1.0f );
 )";
 
+enum class HelperFunction {
+    SignedAtomic = 0,
+    Total,
+};
+
 class ShaderWriter final {
 public:
     void AddExpression(std::string_view text) {
@@ -434,6 +439,7 @@ public:
         DeclareInternalFlags();
         DeclareCustomVariables();
         DeclarePhysicalAttributeReader();
+        DeclareHelpersForward();
 
         code.AddLine("void main() {{");
         ++code.scope;
@@ -450,6 +456,9 @@ public:
 
         --code.scope;
         code.AddLine("}}");
+
+        code.AddNewLine();
+        DeclareHelpers();
     }
 
     std::string GetResult() {
@@ -599,7 +608,7 @@ private:
                 size = limit;
             }
 
-            code.AddLine("shared uint smem[{}];", size / 4);
+            code.AddLine("shared uint {}[{}];", GetSharedMemory(), size / 4);
             code.AddNewLine();
         }
         code.AddLine("layout (local_size_x = {}, local_size_y = {}, local_size_z = {}) in;",
@@ -983,6 +992,27 @@ private:
         }
     }
 
+    void DeclareHelpersForward() {
+        code.AddLine("int Helpers_AtomicShared(uint offset, int value, bool is_min);");
+        code.AddNewLine();
+    }
+
+    void DeclareHelpers() {
+        if (IsHelperEnabled(HelperFunction::SignedAtomic)) {
+            code.AddLine(
+                R"(int Helpers_AtomicShared(uint offset, int value, bool is_min) {{
+    uint oldValue, newValue;
+    do {{
+        oldValue = {}[offset];
+        newValue = is_min ? uint(min(int(oldValue), value)) : uint(max(int(oldValue), value));
+    }} while (atomicCompSwap({}[offset], newValue, oldValue) != oldValue);
+    return int(oldValue);
+}})",
+                GetSharedMemory(), GetSharedMemory());
+            code.AddNewLine();
+        }
+    }
+
     void VisitBlock(const NodeBlock& bb) {
         for (const auto& node : bb) {
             Visit(node).CheckVoid();
@@ -1109,7 +1139,9 @@ private:
         }
 
         if (const auto smem = std::get_if<SmemNode>(&*node)) {
-            return {fmt::format("smem[{} >> 2]", Visit(smem->GetAddress()).AsUint()), Type::Uint};
+            return {
+                fmt::format("{}[{} >> 2]", GetSharedMemory(), Visit(smem->GetAddress()).AsUint()),
+                Type::Uint};
         }
 
         if (const auto internal_flag = std::get_if<InternalFlagNode>(&*node)) {
@@ -1598,7 +1630,9 @@ private:
                 Type::Uint};
         } else if (const auto smem = std::get_if<SmemNode>(&*dest)) {
             ASSERT(stage == ShaderType::Compute);
-            target = {fmt::format("smem[{} >> 2]", Visit(smem->GetAddress()).AsUint()), Type::Uint};
+            target = {
+                fmt::format("{}[{} >> 2]", GetSharedMemory(), Visit(smem->GetAddress()).AsUint()),
+                Type::Uint};
         } else if (const auto gmem = std::get_if<GmemNode>(&*dest)) {
             const std::string real = Visit(gmem->GetRealAddress()).AsUint();
             const std::string base = Visit(gmem->GetBaseAddress()).AsUint();
@@ -2115,7 +2149,14 @@ private:
         UNIMPLEMENTED_IF(meta->sampler.is_array);
         const std::size_t count = operation.GetOperandsCount();
 
-        std::string expr = "texelFetch(";
+        std::string expr = "texelFetch";
+
+        if (!meta->aoffi.empty()) {
+            expr += "Offset";
+        }
+
+        expr += '(';
+
         expr += GetSampler(meta->sampler);
         expr += ", ";
 
@@ -2137,6 +2178,20 @@ private:
             expr += ", ";
             expr += Visit(meta->lod).AsInt();
         }
+
+        if (!meta->aoffi.empty()) {
+            expr += ", ";
+            expr += constructors.at(meta->aoffi.size() - 1);
+            expr += '(';
+            for (size_t i = 0; i < meta->aoffi.size(); ++i) {
+                if (i > 0) {
+                    expr += ", ";
+                }
+                expr += Visit(meta->aoffi[i]).AsInt();
+            }
+            expr += ')';
+        }
+
         expr += ')';
         expr += GetSwizzle(meta->element);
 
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -862,9 +862,17 @@ void RasterizerOpenGL::SyncViewport() {
         if (regs.screen_y_control.y_negate != 0) {
             flip_y = !flip_y;
         }
-        const bool is_zero_to_one = regs.depth_mode == Maxwell::DepthMode::ZeroToOne;
+
         const GLenum origin = flip_y ? GL_UPPER_LEFT : GL_LOWER_LEFT;
-        const GLenum depth = is_zero_to_one ? GL_ZERO_TO_ONE : GL_NEGATIVE_ONE_TO_ONE;
+
+        // The depth reg is unreliable as depth can change after initialisation,
```
ReinUsesLisp, line 868: If this was true, using the same logic on Vulkan wouldn't work, hence why I didn't port Ryujinx' hack.
Remove this change and push it on a separate pull request so it can be discussed in detail.

I am aware this "fixes" Link's Awakening, but it's still a hack coming from a perspective of "we must transform whatever state we have to what the game is doing in NVN" instead of "we must emulate the hardware".

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2705,6 +2778,10 @@ class GLSLDecompiler final {
         }
     }
 
+    std::string GetSharedMemory() const {
```
ReinUsesLisp, line 2781: Any reason to have this on a function instead of a constant?

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -983,6 +992,33 @@ class GLSLDecompiler final {
         }
     }
 
+    void DeclareHelpersForward() {
+        code.AddLine("int Helpers_Atomic{}(uint offset, int value, bool is_min);",
```
ReinUsesLisp, line 996: There are no atomics for local memory.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -983,6 +992,33 @@ class GLSLDecompiler final {
...
+    }
+
+    void DeclareHelpers() {
+        if (helper_functions_enabled[HelperFunctions::SignedAtomic]) {
+            for (const auto& memtype : atomic_memory_types) {
+                code.AddLine(
+                    R"(int Helpers_Atomic{}(uint offset, int value, bool is_min) {{
+    uint oldValue, newValue;
+    do
+    {{
```
ReinUsesLisp, line 1010:
```suggestion
    do {{
```

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2183,8 +2242,22 @@ class GLSLDecompiler final {
     template <const std::string_view& opname, Type type>
     Expression Atomic(Operation operation) {
         if ((opname == Func::Min || opname == Func::Max) && type == Type::Int) {
-            UNIMPLEMENTED_MSG("Unimplemented Min & Max for atomic operations");
-            return {};
+            // Use a helper as a workaround due to memory being uint
+            helper_functions_enabled[HelperFunctions::SignedAtomic] = true;
+            std::string memory;
+            if (std::get_if<SmemNode>(&*operation[0])) {
```
ReinUsesLisp, line 2248: `std::holds_alternative`

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2183,8 +2242,22 @@ class GLSLDecompiler final {
...
+                memory = GetSharedMemory();
+            } else if (std::get_if<LmemNode>(&*operation[0])) {
+                memory = GetLocalMemory();
```
ReinUsesLisp, line 2251: This never happens.

src/video_core/shader/decode/texture.cpp
```diff
@@ -820,17 +819,23 @@ Node4 ShaderIR::GetTldsCode(Instruction instr, TextureType texture_type, bool is
     std::vector<Node> coords;
     for (std::size_t i = 0; i < type_coord_count; ++i) {
         const bool last = (i == (type_coord_count - 1)) && (type_coord_count > 1);
-        coords.push_back(GetRegister(last ? last_coord_register : coord_register + i));
+        coords.push_back(
+            GetRegister(last && !aoffi_enabled ? last_coord_register : coord_register + i));
```
ReinUsesLisp, line 823: Add parenthesis here to make explicit if you want to do `(cond ? a : b) + 1` or `cond ? a : (b + 1)`

ReinUsesLisp (changes requested):

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2115,7 +2155,12 @@ class GLSLDecompiler final {
         UNIMPLEMENTED_IF(meta->sampler.is_array);
         const std::size_t count = operation.GetOperandsCount();
 
-        std::string expr = "texelFetch(";
+        std::string expr = "texelFetch";
+        if (!meta->aoffi.empty()) {
+            expr += "Offset";
+        }
+        expr += "(";
```
lioncash, line 2162:
```suggestion
        expr += '(';
```

Ditto below for a few other character constants

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2758,6 +2835,10 @@ class GLSLDecompiler final {
     ShaderWriter code;
 
     std::optional<u32> max_input_vertices;
+
+    std::array<bool, HelperFunctions::Total> helper_functions_enabled;
```
lioncash, line 2839: This can be initialized in the class body like so:

```suggestion
    std::array<bool, HelperFunctions::Total> helper_functions_enabled{};
```

allowing the additions to the constructor initializer list to be removed. Similarly `atomic_memory_types`'s entry in the initializer list can be removed because it has its own constructor that always establishes its invariants.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -410,7 +415,8 @@ class GLSLDecompiler final {
                             ShaderType stage_, std::string_view identifier_,
                             std::string_view suffix_)
         : device{device_}, ir{ir_}, registry{registry_}, stage{stage_},
-          identifier{identifier_}, suffix{suffix_}, header{ir.GetHeader()} {
+          identifier{identifier_}, suffix{suffix_}, header{ir.GetHeader()},
+          helper_functions_enabled{}, atomic_memory_types{} {
```
lioncash, line 419: These two additions aren't necessary. See a previous review comment for more details.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
lioncash: The constant could also be a constexpr std::string_view, as opposed to a std::string.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -79,6 +79,11 @@ const float fswzadd_modifiers_a[] = float[4](-1.0f,  1.0f, -1.0f,  0.0f );
 const float fswzadd_modifiers_b[] = float[4](-1.0f, -1.0f,  1.0f, -1.0f );
 )";
 
+enum HelperFunctions {
```
lioncash, line 82:
```suggestion
enum class HelperFunction {
```

We generally avoid regular enums, as they pollute the surrounding scope and are also implicitly convertible. The following private helpers can also be introduced:

```cpp
void SetHelperEnabled(HelperFunction hf, bool enabled) {
    helper_functions_enabled[static_cast<u32>(hf)] = enabled;
}

bool IsHelperEnabled(HelperFunction hf) const {
    return helper_functions_enabled[static_cast<u32>(hf)];
}
```

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
Kelebek1: Ah ok, I wasn't sure about this, which is why I did it this way. Can they only be used for shared memory then and nothing else?

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
Kelebek1: I made it a function just to replicate GetLocalMemory, so there was a single source for changing the name later on.

src/video_core/renderer_opengl/gl_rasterizer.cpp
Kelebek1: Removed from the PR, I'll open a new one.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -983,6 +992,27 @@ class GLSLDecompiler final {
...
+    do {{
+        oldValue = {}[offset];
+        newValue = is_min ? uint(min(int(oldValue), value)) : uint(max(int(oldValue), value));
+    }} while (atomicCompSwap({}[offset], newValue, oldValue) != oldValue);
+    return int(oldValue);
+}})",
+                GetSharedMemory(), GetSharedMemory());
+            code.AddNewLine();
+        }
+    };
```
lioncash, line 1014: Just noticed this, but this semicolon will cause a -Wextra-semi warning, since it's not necessary.

Kelebek1: No problems! Fixed.

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
ReinUsesLisp: They can also be used with global memory (storage buffers).

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2746,6 +2808,14 @@ class GLSLDecompiler final {
         return std::min<u32>(device.GetMaxVaryings(), Maxwell::NumVaryings);
     }
 
+    void SetHelperEnabled(HelperFunction hf, bool enabled) {
+        helper_functions_enabled[static_cast<u32>(hf)] = enabled;
```
ReinUsesLisp, line 2812:
```suggestion
        helper_functions_enabled[static_cast<size_t>(hf)] = enabled;
```

src/video_core/renderer_opengl/gl_shader_decompiler.cpp
```diff
@@ -2746,6 +2808,14 @@ class GLSLDecompiler final {
...
+    }
+
+    bool IsHelperEnabled(HelperFunction hf) const {
+        return helper_functions_enabled[static_cast<u32>(hf)];
```
ReinUsesLisp, line 2816:
```suggestion
        return helper_functions_enabled[static_cast<size_t>(hf)];
```

lat9nq:
Causes a regression on NVIDIA GPUs in BOTW.

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/6209718/yuzu_log.txt)

Screenshot_2021-03-26_02-18-54

Kelebek1:
Closed for Hades.

PR #5981, opened 2021-02-22
ci: Add clang build scripts

lat9nq:
Adds scripts necessary to compile yuzu using Clang on the yuzu-emu/build-environments:linux-fresh Docker container.

These scripts are mostly a copy of those in .ci/scripts/linux, minus AppImage building since that isn't necessary. Re-uses linux-fresh since that container has Clang 12 installed.

Apparently also adds a trailing newline to the end of a script.

I hope I got the template right, but I'm not familiar with that format. Please do not merge until CI passes with Clang.

### DIFF
--- a/.ci/scripts/clang/docker.sh
+++ b/.ci/scripts/clang/docker.sh
@@ -1,0 +1,18 @@
+#!/bin/bash -ex
+
+# Exit on error, rather than continuing with the rest of the script.
+set -e
+
+cd /yuzu
+
+ccache -s
+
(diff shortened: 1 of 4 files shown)


lat9nq:
Oh, another important note: this configures Clang to use libstdc++, *not* libc++ since yuzu uses components not available in libc++.

lioncash (approved):

lioncash:
Hell yeah! Thanks for getting clang CI all set up =)

PR #5989, opened 2021-02-23
vk_command_pool: Reduce the command pool size from 4096 to 4

ReinUsesLisp:
This allows drivers to reuse memory more easily and preallocate less.
The optimal number has been measured booting Pokémon Sword.

Before this commit:
imagen

After this commit:
imagen

### DIFF
--- a/src/video_core/renderer_vulkan/vk_command_pool.cpp
+++ b/src/video_core/renderer_vulkan/vk_command_pool.cpp
@@ -10,7 +10,7 @@
 
 namespace Vulkan {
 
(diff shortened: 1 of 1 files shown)


ReinUsesLisp:
This value was kept high from to the very early days of the Vulkan backend when it was still in development. Each draw had its command buffer, so a high number made sense. This no longer applies.

bunnei (approved):

v1993:
It looks like this PR have caused #6634. Would you mind taking a look?

PR #5990, opened 2021-02-24
InputCommon: Mouse fixes

german77:
Fixes a couple of problems that have been present in the mouse implementation.

- Up to 7 different buttons are detected now. (left, right, middle, back, forward, task and another extra button)
- Mapping a mouse button no longer stays pressed after the mapping is done.
- Default hotkey has been changed to Ctrl+F9.
- Mouse will not hide anymore if hide_mouse setting is not enabled
- Mouse panning setting will be reset at reboot and only mouse sensitivity will be stored

### DIFF
--- a/src/input_common/mouse/mouse_input.cpp
+++ b/src/input_common/mouse/mouse_input.cpp
@@ -59,7 +59,7 @@ void Mouse::UpdateYuzuSettings() {
     });
 }
 
-void Mouse::PressButton(int x, int y, int button_) {
+void Mouse::PressButton(int x, int y, MouseButton button_) {
     const auto button_index = static_cast<std::size_t>(button_);
     if (button_index >= mouse_info.size()) {
         return;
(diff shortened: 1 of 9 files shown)


src/input_common/mouse/mouse_input.h
```diff
@@ -22,6 +22,8 @@ enum class MouseButton {
     Right,
```
Morph1984, line 22: Seems like a missed this in the precious PR, `Wheel` should be under `Right`

src/input_common/mouse/mouse_input.cpp
```diff
@@ -59,8 +59,27 @@ void Mouse::UpdateYuzuSettings() {
     });
 }
 
+MouseButton Mouse::QtButtonToMouseButton(int button) const {
```
Morph1984, line 62: This can be placed into an anonymous namespace and made constexpr if it isn't going to be used anywhere else.

lioncash: I don't think we should be putting frontend-specific things like this into input common, especially if we eventually support other frontends in the future (e.g. Android or some other platform).

I think it would be better to translate the inputs and pass them in from the front-end's side of things. This way the magic values below can also be replaced with Qt enum values.

v1993:
Testing this on Linux and can't bind mouse wheel movement. Am I missing something?

german77:
Mouse wheel movement is not implemented yet. I will do it once I have free time

bunnei (approved):

PR #6004, opened 2021-02-27
InputCommon: Use an unique client id for each udp socket instance

german77:
According to https://github.com/v1993/cemuhook-protocol each socket connection should have a unique client id. This should solve problems where test fails or you need to restart the server. Also removes the magic number 24872 that was used as client id for yuzu.

This is a candidate to be ported to citra since it has the same issue.

### DIFF
--- a/src/input_common/udp/client.cpp
+++ b/src/input_common/udp/client.cpp
@@ -5,6 +5,7 @@
 #include <chrono>
 #include <cstring>
 #include <functional>
+#include <random>
 #include <thread>
 #include <boost/asio.hpp>
 #include "common/logging/log.h"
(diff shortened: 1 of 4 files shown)


src/input_common/udp/client.cpp
```diff
@@ -63,6 +64,13 @@ class Socket {
     }
 
 private:
+    u32 GenerateRandomClientId() const {
+        std::random_device device;
```
v1993, line 68: Since you get only a single number from generator, consider using just

```
return std::random_device()();
```

v1993 (changes requested):
Lgtm, only one note

src/input_common/udp/client.cpp
```diff
@@ -63,6 +64,13 @@ class Socket {
...
+        std::mt19937 gen(device());
+        std::uniform_int_distribution<u32> distribution(0, static_cast<u32>(-1));
```
lioncash, line 70: Note that in the future it's much more readable to use UINT32_MAX instead of casting -1

v1993: Wouldn't it be even better to use `std::numeric_limits<u32>::max()`? It's mostly a matter of style, so I wonder which is preferred in yuzu codebase.

german77: I did tried to go for a keyword for the max value but I couldn't find it.  So I used a cast. Next time I will use the keyword.

src/input_common/udp/client.cpp
```diff
@@ -63,6 +64,10 @@ class Socket {
     }
 
 private:
+    u32 GenerateRandomClientId() const {
+        std::random_device{}();
```
v1993, line 68:
```suggestion
        return std::random_device{}();
```
Missing `return` here.

v1993 (changes requested):

src/input_common/udp/client.cpp
lioncash: @v1993 Either works, but UINT32_MAX is nice in non-generic code, as it's nicer to read.

bunnei (approved):

PR #6019, opened 2021-03-02
[Service::nifm] Fix bcat_backend's default initialisation

Kelebek1:
bcat_backend checks for "none," but is default initialised to "null," leading to incorrectly returning a Connected state without bcat enabled due to the direct string comparisons. Not a huge deal, as opening and closing the settings at any time will fix it back to "none."


bunnei (approved):

bunnei:
LGTM, nice catch

PR #6021, opened 2021-03-02
buffer_cache: Heuristically decide to skip cache on uniform buffers

ReinUsesLisp:
Some games benefit from skipping caches (Pokémon Sword), and others don't (Animal Crossing: New Horizons). Add an heuristic to decide this at runtime, an alternative solution would be to have a database of games that benefit from this but that's ugly.

The cache hit ratio has to be ~98% or better to not skip the cache. There are 16 frames of buffer.

I tested the mentioned games as well as Super Mario Odyssey and The Legend of Zelda Breath of the wild. These two games fall under the skip cache category, so they won't be affected by this change. More testing would be appreciated.

- Fixes a performance regression introduced after the initial release of the buffer cache rewrite on Vulkan on AC:NH.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -9,6 +9,7 @@
 #include <deque>
 #include <memory>
 #include <mutex>
+#include <numeric>
 #include <span>
 #include <unordered_map>
 #include <vector>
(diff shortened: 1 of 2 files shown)


bunnei (approved):

x-variable:
Any chance of a toggle on this one @bunnei ? I think it's causing additional stuttering in SSBU on vulkan

PR #6028, opened 2021-03-03
video_core: rasterizer_accelerated: Use a flat array instead of interval_map for cached pages.

bunnei:
- Uses a fixed 64MB for the cache instead of an ever growing map.
- Slightly faster by using atomics instead of a single mutex for access.
- Thanks to Rodrigo for the idea.


src/video_core/rasterizer_accelerated.cpp
```diff
@@ -2,63 +2,42 @@
...
+            ASSERT_MSG(count > 0, "Count may underflow!");
         } else {
-            ASSERT(count >= 0);
+            ASSERT_MSG(true, "Delta must be non-zero!");
         }
-    }
 
-    if (delta < 0) {
-        cached_pages.add({pages_interval, delta});
+        count += static_cast<s8>(delta);
```
lioncash, line 31: wouldn't this be a sign/unsigned mismatch, given it's adding a signed value to an unsigned type?

PR #6029, opened 2021-03-03
CMakeLists: Add /utf-8 compile option for MSVC

Morph1984:
Ensures that the source and execution character sets are in UTF-8

Also fixes the Move Up / Down arrows 
image

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,6 +27,7 @@ if (MSVC)
     # /Zo                 - Enhanced debug info for optimized builds
     # /permissive-        - Enables stricter C++ standards conformance checks
     # /EHsc               - C++-only exception handling semantics
+    # /utf-8              - Set source and execution character sets to UTF-8
     # /volatile:iso       - Use strict standards-compliant volatile semantics.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -136,6 +136,11 @@ else()
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
 endif()
 
+# Ensure MSVC detects source files as UTF-8 encoded
+if (MSVC)
+    add_compile_options(/utf-8)
```
lioncash, line 141: Shouldn't we only apply this to source code that we own? (i.e. everything in the src folder)

Morph1984: At present the [mbedtls](https://github.com/DarkLordZach/mbedtls/tree/a280e602f3a4ae001d3a83cbc3e6e04c99c22277) external possesses a file that cannot compile under different code pages. Perhaps we could try updating/replacing this external

Morph1984: An updated [mbedtls](https://github.com/yuzu-emu/mbedtls) has been provided here, I will confirm if the issue is fixed

lioncash (approved):

PR #6040, opened 2021-03-06
Enable toggle buttons for keyboard and mouse

german77:
Since keyboard and mouse has a limit in how many keys can be pressed at the same time. I added the ability to make buttons a toggle button. This way you can free a button from your keyboard by just pressing it once and it will remain pressed until you press it again.

Right click the modifier button then select "toggle button" to enable it or disable it.

### DIFF
--- a/src/input_common/keyboard.cpp
+++ b/src/input_common/keyboard.cpp
@@ -19,16 +19,18 @@ public:
 
     bool GetStatus() const override {
         if (toggle) {
-            return toggled_status.load();
+            return toggled_status.load(std::memory_order_relaxed);
         }
         return status.load();
     }
 
     void ToggleButton() {
-        if (!lock) {
-            lock = true;
-            toggled_status.store(!toggled_status.load());
+        if (lock) {
+            return;
         }
+        lock = true;
+        const bool old_toggle_status = toggled_status.load();
+        toggled_status.store(!old_toggle_status);
     }
 
     void UnlockButton() {
@@ -41,7 +43,7 @@ private:
     std::shared_ptr<KeyButtonList> key_button_list;
     std::atomic<bool> status{false};
     std::atomic<bool> toggled_status{false};
-    bool lock = {};
+    bool lock{false};
     const bool toggle;
 };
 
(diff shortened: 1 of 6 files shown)


src/input_common/keyboard.cpp
```diff
@@ -12,20 +12,37 @@ namespace InputCommon {
...
+        lock = false;
+    }
+
     friend class KeyButtonList;
 
 private:
     std::shared_ptr<KeyButtonList> key_button_list;
     std::atomic<bool> status{false};
+    std::atomic<bool> toggled_status{false};
+    bool lock = {};
```
lioncash, line 44:
```suggestion
    bool lock{false};
```

to match the convention of the variables above it

src/input_common/keyboard.cpp
```diff
@@ -12,20 +12,37 @@ namespace InputCommon {
...
+        if (toggle) {
+            return toggled_status.load();
+        }
         return status.load();
     }
 
+    void ToggleButton() {
+        if (!lock) {
+            lock = true;
+            toggled_status.store(!toggled_status.load());
```
lioncash, line 30: This is, unfortunately, not a fully atomic sequence of update operations, because this has a separate atomic load and store, and won't be transformed into a single atomc read-modify-write operation.

To make this fully atomic, you would need to either use one of the atomic [compare_exchange](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange) functions in an update loop, or convert `toggled_status` to an atomic integer type and use an simpler interface like `toggled_status.fetch_xor(1)` (which is guaranteed to be a read-modify-write operation).

german77: I don't fully understand how atomic operations work. Since `status `was an atomic variable I decided to make `toggle_status` a atomic value. The value here is non critical. It doesn't matter if it fails a couple of times.

If it's not necessary to use atomic I can switch it to a normal bool variable

lioncash: The reason you'd want to use one of the above is because a proper read-modify-write operation prevents data races from occurring (given the variable is used on multiple threads, changing it to a non-atomic value isn't really advisable, since it'd also be undefined behavior on top of the race condition)

bunnei (approved):

PR #6041, opened 2021-03-06
common: fiber: Use weak_ptr when yielding.

bunnei:
- Avoids a memory leak, as taking a strong reference of the fiber here causes a circular reference.
- Supersedes #6006 with a more narrow fix.

### DIFF
--- a/src/common/fiber.cpp
+++ b/src/common/fiber.cpp
@@ -116,16 +116,14 @@ void Fiber::Rewind() {
     boost::context::detail::jump_fcontext(impl->rewind_context, this);
 }
 
-void Fiber::YieldTo(std::weak_ptr<Fiber> weak_from, std::shared_ptr<Fiber> to) {
-    ASSERT_MSG(to != nullptr, "Next fiber is null!");
+void Fiber::YieldTo(std::weak_ptr<Fiber> weak_from, Fiber& to) {
+    to.impl->guard.lock();
+    to.impl->previous_fiber = weak_from.lock();
 
-    to->impl->guard.lock();
-    to->impl->previous_fiber = weak_from.lock();
-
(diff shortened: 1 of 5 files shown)


src/common/fiber.cpp
```diff
@@ -116,16 +116,21 @@ void Fiber::Rewind() {
...
+    to->impl->previous_fiber = weak_from.lock();
+
     auto transfer = boost::context::detail::jump_fcontext(to->impl->context, to.get());
-    ASSERT(from->impl->previous_fiber != nullptr);
-    from->impl->previous_fiber->impl->context = transfer.fctx;
-    from->impl->previous_fiber->impl->guard.unlock();
-    from->impl->previous_fiber.reset();
+
+    // "from" might no longer be valid if the thread was killed
+    if (auto from = weak_from.lock(); from != nullptr) {
```
lioncash, line 128: This can be simplified to:

```suggestion
    if (auto from = weak_from.lock()) {
```

lioncash (commented):
LGTM aside from one nitpick

src/common/fiber.cpp
```diff
@@ -116,16 +116,21 @@ void Fiber::Rewind() {
     boost::context::detail::jump_fcontext(impl->rewind_context, this);
 }
 
-void Fiber::YieldTo(std::shared_ptr<Fiber> from, std::shared_ptr<Fiber> to) {
-    ASSERT_MSG(from != nullptr, "Yielding fiber is null!");
+void Fiber::YieldTo(std::weak_ptr<Fiber> weak_from, std::shared_ptr<Fiber> to) {
```
FernandoS27, line 119: `std::shared_ptr<Fiber> to` must be a reference or it will leak.

PR #6052, opened 2021-03-10
IApplicationDisplayService: Stub GetIndirectLayerImageMap

Morph1984:
Used by games invoking the inline software keyboard such as GNOSIA

### DIFF
--- a/src/core/hle/service/vi/vi.cpp
+++ b/src/core/hle/service/vi/vi.cpp
@@ -1217,6 +1217,32 @@ private:
         }
     }
 
+    void GetIndirectLayerImageMap(Kernel::HLERequestContext& ctx) {
+        IPC::RequestParser rp{ctx};
+        const auto width = rp.Pop<s64>();
+        const auto height = rp.Pop<s64>();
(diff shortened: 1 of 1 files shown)


src/core/hle/service/vi/vi.cpp
```diff
@@ -1217,6 +1217,28 @@ class IApplicationDisplayService final : public ServiceFramework<IApplicationDis
...
+        LOG_WARNING(Service_VI,
+                    "(STUBBED) called, width={}, height={}, indirect_layer_consumer_handle={}, "
+                    "applet_resource_user_id={}",
+                    width, height, indirect_layer_consumer_handle, applet_resource_user_id);
+
+        std::vector<u8> out_buffer(0x46);
+        ctx.WriteBuffer(out_buffer);
+
+        // TODO: This outputs 2 s64s which are copied into u64s, these values are currently unknown
+        // so they are not included for now.
```
FearlessTobi, line 1236: Shouldn't we at least return 0 for those values though?

bunnei (approved):

PR #6057, opened 2021-03-11
gl_device: Intel: do not apply Windows fixes in Linux

A-w-x:
Fixes regressions introduced [here](https://github.com/yuzu-emu/yuzu/commit/7d904fef2e6ac9ce8a3df71e758a36d39b8f69e5)

Known regression: Mario Tennis home menu is a black screen
Tested on mesa 20.3.4 Iris plus gen 11

Are amd igpus affected too?


v1993:
Please accept CLA. Yuzu can not accept contributions without that.

A-w-x:
i see no regressions

@ReinUsesLisp
what games are fixed by this?
so that i can see if i can test affected games and we close this pr

ReinUsesLisp:
What do you mean? That was added because texture views are broken on Intel and AMD, but mesa is not affected. An easy way to test if texture views work is to boot Pokémon Sword without a save, the first screen should be broken.

I don't see anything wrong with excluding mesa Intel from the blacklist, as long as it's an updated driver (because texture views on blits were also broken).

v1993:
Please note that it's preferred to rebase on top of fresh master than opening another PR for future contributions.

PR #6095, opened 2021-03-21
gl_device: Block async shaders on AMD and Intel

lat9nq:
Currently, the Windows version of the Intel OpenGL driver and the AMD proprietary OpenGL driver do not properly support (or in fact degrade in performance) when asynchronous shader compilation is enabled. This blocks specifically those drivers from using this feature. This affects AMDGPU-PRO on Linux, and AMD's and Intel's OpenGL drivers on Windows.

This will log when it is enabled in settings but not compatible with the driver, much the same way assembly shaders are disabled at a guest boot time.

The only case that hasn't been tested at the time of writing is Intel on Linux.

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -210,6 +210,12 @@ Device::Device() {
     const bool is_amd = vendor == "ATI Technologies Inc.";
     const bool is_intel = vendor == "Intel";
 
+#ifdef __linux__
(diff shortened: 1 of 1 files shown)


ReinUsesLisp (approved):

Morph1984 (approved):

IGCIT:
async is working on linux for intel
mesa 20.3.4, opengl

lioncash (approved):

PR #6099, opened 2021-03-21
Kernel Rework: Derive memory regions from board layout.

bunnei:
This change reworks a bunch of our kernel memory management code to derive the memory layout based on the Switch board we are emulating. This removes the need for some hardcoded memory regions (e.g. for memory pools) and makes it easier to support future hardware revisions. There are several other cleanups/refactoring changes here as well.


src/common/common_sizes.h
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <limits>
+
+#include "common/common_types.h"
+
+enum : u64 {
```
lioncash, line 11: This should be within the Common namespace

src/core/hle/kernel/board/nintendo/nx/k_system_control.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include "common/common_sizes.h"
+#include "core/hle/kernel/board/nintendo/nx/k_system_control.h"
+#include "core/hle/kernel/board/nintendo/nx/secure_monitor.h"
+#include "core/hle/kernel/k_trace.h"
+
+namespace Kernel::Board::Nintendo::Nx {
+
+namespace impl {
+
+constexpr inline const std::size_t RequiredNonSecureSystemMemorySizeVi = 0x2238 * 4 * 1024;
```
lioncash, line 16: inline here isn't necessary at file scope within a cpp file, ditto for other occurrences.

src/core/hle/kernel/k_memory_layout.h
```diff
@@ -27,64 +71,327 @@ constexpr bool IsKernelAddress(VAddr address) {
...
+        return IsTypedAddress(region, address, GetPhysicalLinearMemoryRegionTree(),
+                              static_cast<KMemoryRegionType>(KMemoryRegionAttr_LinearMapped));
+    }
+    bool IsLinearMappedPhysicalAddress(const KMemoryRegion*& region, PAddr address,
+                                       size_t size) const {
+        return IsTypedAddress(region, address, size, GetPhysicalLinearMemoryRegionTree(),
+                              static_cast<KMemoryRegionType>(KMemoryRegionAttr_LinearMapped));
+    }
+
+    std::tuple<size_t, size_t> GetTotalAndKernelMemorySizes() const {
```
lioncash, line 211: Given this is only a tuple with two elements, this can be a std::pair, to reduce the number of template instantiations a bit.

src/core/hle/kernel/k_memory_region.h
```diff
@@ -0,0 +1,352 @@
...
+    using const_reference = TreeType::const_reference;
+    using iterator = TreeType::iterator;
+    using const_iterator = TreeType::const_iterator;
+
+private:
+    TreeType m_tree{};
+    KMemoryRegionAllocator& memory_region_allocator;
+
+public:
+    KMemoryRegionTree(KMemoryRegionAllocator& memory_region_allocator_);
```
lioncash, line 163:
```suggestion
    explicit KMemoryRegionTree(KMemoryRegionAllocator& memory_region_allocator_);
```

src/core/hle/kernel/k_memory_region.h
```diff
@@ -0,0 +1,352 @@
...
+    iterator nfind(const_reference ref) const {
+        return m_tree.nfind(ref);
+    }
+};
+
+class KMemoryRegionAllocator final : NonCopyable {
+public:
+    static constexpr size_t MaxMemoryRegions = 200;
+
+private:
```
lioncash, line 331: Maybe move this to the bottom of the class, so that all the scattered public sections can be unified into one.

src/core/hle/kernel/k_memory_region_type.h
```diff
@@ -0,0 +1,339 @@
...
+#include "common/bit_util.h"
+#include "common/common_funcs.h"
+#include "common/common_types.h"
+
+#define ARCH_ARM64
+#define BOARD_NINTENDO_NX
+
+namespace Kernel {
+
+enum KMemoryRegionType : u32 {
```
lioncash, line 16: Should this not be an enum class so that the `KMemoryRegionAttr_` prefix isn't necessary for all enum elements? Especially given operators are declared for the enum.

src/core/hle/kernel/k_memory_region_type.h
```diff
@@ -0,0 +1,339 @@
...
+            ++high;
+            low = 0;
+        }
+    }
+    return high + 1;
+}
+
+class KMemoryRegionTypeValue {
+private:
+    using ValueType = typename std::underlying_type<KMemoryRegionType>::type;
```
lioncash, line 45:
```suggestion
    using ValueType = std::underlying_type_t<KMemoryRegionType>;
```

This should likely be public as well, given the public accessor that retrieves the value of this class uses it.

src/core/hle/kernel/k_memory_region_type.h
```diff
@@ -0,0 +1,339 @@
...
+        if ((++low) == high) {
+            ++high;
+            low = 0;
+        }
+    }
+    return high + 1;
+}
+
+class KMemoryRegionTypeValue {
+private:
```
lioncash, line 44: Maybe consider moving the private interface to the bottom of the class to match our convention of prioritizing the public interface

src/core/hle/kernel/kernel.cpp
```diff
@@ -269,44 +270,309 @@ struct KernelCore::Impl {
     }
 
     void InitializeMemoryLayout() {
-        // Initialize memory layout
-        constexpr KMemoryLayout layout{KMemoryLayout::GetDefaultLayout()};
+        KMemoryLayout memory_layout;
```
lioncash, line 273: This function is getting kind of massive. Perhaps it should be broken out into constituent functions where applicable to make it nicer to digest.

PR #6109, opened 2021-03-25
HID: Initialize correctly the gesture finger_id and filter invalid inputs

german77:
Gesture only allows up to 4 points. They where initialized at 16. This PR fixes that error and adds a filter to prevent writing to invalid places in the array

### DIFF
--- a/src/core/hle/service/hid/controllers/gesture.cpp
+++ b/src/core/hle/service/hid/controllers/gesture.cpp
@@ -4,6 +4,7 @@
 
 #include <cstring>
 #include "common/common_types.h"
+#include "common/logging/log.h"
 #include "core/core_timing.h"
 #include "core/frontend/emu_window.h"
 #include "core/hle/service/hid/controllers/gesture.h"
@@ -19,9 +20,9 @@ Controller_Gesture::~Controller_Gesture() = default;
 
 void Controller_Gesture::OnInit() {
     for (std::size_t id = 0; id < MAX_FINGERS; ++id) {
-        mouse_finger_id[id] = MAX_FINGERS;
-        keyboard_finger_id[id] = MAX_FINGERS;
-        udp_finger_id[id] = MAX_FINGERS;
+        mouse_finger_id[id] = MAX_POINTS;
+        keyboard_finger_id[id] = MAX_POINTS;
+        udp_finger_id[id] = MAX_POINTS;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -142,6 +142,10 @@ std::optional<std::size_t> Controller_Gesture::GetUnusedFingerID() const {
 std::size_t Controller_Gesture::UpdateTouchInputEvent(
     const std::tuple<float, float, bool>& touch_input, std::size_t finger_id) {
     const auto& [x, y, pressed] = touch_input;
+    if (finger_id > MAX_POINTS) {
+        // Invalid finger id reset to defaults
+        return MAX_POINTS;
```
lioncash, line 147: Should there be a debug log here and in the equivalent touchscreen code to catch invalid input, or do games willingly send invalid data regularly?

german77: You are right a log error should be here. This condition should never be triggered by any game as games don't write data here. But since giving an incorrect parameter can potentially break something I decided to filter it.

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -142,6 +142,10 @@ std::optional<std::size_t> Controller_Gesture::GetUnusedFingerID() const {
 std::size_t Controller_Gesture::UpdateTouchInputEvent(
     const std::tuple<float, float, bool>& touch_input, std::size_t finger_id) {
     const auto& [x, y, pressed] = touch_input;
+    if (finger_id > MAX_POINTS) {
+        LOG_ERROR(Service_HID, "Invalid finger id {}", finger_id);
```
lioncash, line 146: `"common/logging/log.h"` needs to be included

src/core/hle/service/hid/controllers/touchscreen.cpp
```diff
@@ -118,6 +118,10 @@ std::optional<std::size_t> Controller_Touchscreen::GetUnusedFingerID() const {
 std::size_t Controller_Touchscreen::UpdateTouchInputEvent(
     const std::tuple<float, float, bool>& touch_input, std::size_t finger_id) {
     const auto& [x, y, pressed] = touch_input;
+    if (finger_id > MAX_FINGERS) {
+        LOG_ERROR(Service_HID, "Invalid finger id {}", finger_id);
```
lioncash, line 122: `"common/logging/log.h"` needs to be included.

bunnei (approved):

PR #6116, opened 2021-03-27
yuzu/main: Add user command line argument

german77:
Allows to select a specific user at boot. If the selected user is invalid the argument is ignored.
Example:
`yuzu.exe  -u 0`
`yuzu.exe -u 1 -g "path_to_game"`
`yuzu.exe -f -u 2 -g "path_to_game"`

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -320,6 +320,34 @@ GMainWindow::GMainWindow()
             continue;
         }
 
+        // Launch game with a specific user
+        if (args[i] == QStringLiteral("-u")) {
+            if (i >= args.size() - 1) {
+                continue;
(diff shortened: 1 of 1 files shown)


bunnei (approved):

asheenlevrai:
This -u X argument doesn't work on my end (running yuzu v1233).
Using it -> Yuzu launches and prompts for user selection.
Does X needs to be 0, 1, 2, ... for the successive users? (Doesn't work on my end)
or does it need to be the long string of characters corresponding to the user profile? (Doesn't work on my end either)

Thanks a lot for your help.
-a-

asheenlevrai:
Is there any argument to pass a specific controller profile?

PR #6119, opened 2021-03-28
InputCommon: Address mapping and naming issues with SDL2

german77:
Xbox 360 and One controllers where listed as `Xinput Controller`. The Pro controller doesn't need the face buttons swapped. 

This PR displays the correct controller name for both  controllers and maps the Pro controller correctly.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -761,7 +761,7 @@ std::vector<Common::ParamPackage> SDLState::GetInputDevices() {
         for (const auto& joystick : value) {
             if (auto* const controller = joystick->GetSDLGameController()) {
                 std::string name =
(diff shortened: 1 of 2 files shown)


src/input_common/sdl/sdl_impl.cpp
```diff
@@ -760,8 +760,20 @@ std::vector<Common::ParamPackage> SDLState::GetInputDevices() {
     for (const auto& [key, value] : joystick_map) {
         for (const auto& joystick : value) {
             if (auto* const controller = joystick->GetSDLGameController()) {
-                std::string name =
-                    fmt::format("{} {}", SDL_GameControllerName(controller), joystick->GetPort());
+                std::string name;
+                // Use custom name for specific controllers because the default name is not correct
```
lioncash, line 764: Maybe place this into a local helper function that returns a std::string and call it here, in case this needs to be expanded in the future

bunnei (approved):

PR #6125, opened 2021-03-30
nvdrv: Cleanup CDMA Processor on device closure

ogniK5377:
Brings us a step closer to unifying all channels to share a common interface.

### DIFF
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -494,9 +494,8 @@ void GPU::PushCommandBuffer(Tegra::ChCommandHeaderList& entries) {
 
 void GPU::ClearCommandBuffer() {
     // This condition fires when a video stream ends, clear all intermediary data
-    if (cdma_pusher) {
-        cdma_pusher.reset();
-    }
+    cdma_pusher.reset();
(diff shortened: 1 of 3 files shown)


src/video_core/gpu.cpp
```diff
@@ -496,6 +492,13 @@ void GPU::PushCommandBuffer(Tegra::ChCommandHeaderList& entries) {
     cdma_pusher->ProcessEntries(std::move(entries));
 }
 
+void GPU::ClearCommandBuffer() {
+    // This condition fires when a video stream ends, clear all intermediary data
+    if (cdma_pusher) {
```
lioncash, line 497: This can be reset unconditionally. Given `reset()` will do this check internally

src/video_core/gpu.cpp
```diff
@@ -496,6 +492,13 @@ void GPU::PushCommandBuffer(Tegra::ChCommandHeaderList& entries) {
     cdma_pusher->ProcessEntries(std::move(entries));
 }
 
+void GPU::ClearCommandBuffer() {
+    // This condition fires when a video stream ends, clear all intermediary data
```
ameerj, line 496: Can you add a log indicating that the video stream ended?

bunnei (approved):

PR #6126, opened 2021-03-30
ISelfController: Stub SetAlbumImageTakenNotificationEnabled

Morph1984:
This service call sets an internal flag whether a notification is shown when an image is captured.
Currently we do not support capturing images via the capture button, so this can be stubbed for now.


src/core/hle/service/am/am.cpp
```diff
@@ -560,6 +560,21 @@ void ISelfController::GetAccumulatedSuspendedTickChangedEvent(Kernel::HLERequest
     rb.PushCopyObjects(accumulated_suspended_tick_changed_event->GetReadableEvent());
 }
 
+void ISelfController::SetAlbumImageTakenNotificationEnabled(Kernel::HLERequestContext& ctx) {
+    IPC::RequestParser rp{ctx};
+
+    /**
```
lioncash, line 566: We generall use `//` for line comments and `/**` for documentation comments on functions.

lioncash:
LGTM once clang-format is fixed

lioncash (approved):

bunnei (approved):

PR #6130, opened 2021-04-02
common: Move assert failure handling into a cpp file.

degasus:
Advantage: Altering the handler does not need a full recompilation.
Disadvantage: noreturn is droped, so the caller is a bit slower.

We quite often run yuzu with a YOLO assertion handler. In fact, only very few
games run at all with asserts. This patch allows developers to patch the handler
without recompiling everything. The overhead of the missing "noreturn" attribute
shoul be negletable.

This will conflict on purpose with #1340.

### DIFF
--- a/src/common/assert.h
+++ b/src/common/assert.h
@@ -4,9 +4,12 @@
 
 #pragma once
 
-#include <cstdlib>
-#include "common/common_funcs.h"
 #include "common/logging/log.h"
+
+// Sometimes we want to try to continue even after hitting an assert.
+// However touching this file yields a global recompilation as this header is included almost
+// everywhere. So let's just move the handling of the failed assert to a single cpp file.
+void assert_handle_failure();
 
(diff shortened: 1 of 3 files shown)


src/common/assert.h
```diff
@@ -4,10 +4,13 @@
...
 #pragma once
 
-#include <cstdlib>
-#include "common/common_funcs.h"
 #include "common/logging/log.h"
 
+// Sometimes we want to try to continue even after hitting an assert.
+// However touching this file yields a global recompilation as this header is included almost
+// everywhere. So let's just move the handling of the failed assert to a single cpp file.
+extern void assert_handle_failure();
```
lioncash, line 12:
```suggestion
void assert_handle_failure();
```

lioncash (commented):
LGTM otherwise

This is definitely better from a productivity viewpoint

bunnei (approved):

PR #6133, opened 2021-04-02
Project Eleuthia - On-Screen Keyboard and Error Applet Overlays

Morph1984:
## Project Eleuthia - On-Screen Keyboard and Error Applet Overlays

I and @Its-Rei are back to collaborate on yet another project focused on improving the various emulated applets within yuzu.

What is Project Eleuthia you ask? It aims to introduce the very first overlay applets within yuzu - the On-Screen Keyboard (OSK) and Error Display.

### Why overlays?

Have you ever needed to enter a name within games such as Pokemon or Super Smash Bros. Ultimate and find that you can't use a controller to enter a name?
In our quest to emulate the various features of the Nintendo Switch, the software keyboard was the perfect feature to implement a controller-friendly overlay.

### What was wrong with the previous software keyboard?

In addition to it not supporting controller input, its base implementation does not properly support newer revisions of the software keyboard and the inline software keyboard - which is used by games such as Monster Hunter Generations Ultimate, Dark Souls Remastered, Daemon X Machina, et. al.

Collaborating with Rei, we managed to completely rewrite and revamp the software keyboard implementation to not only support new versions of the software keyboard, but also implement the inline software keyboard used by the aforementioned games along with the long requested controller input support. We achieved this by mimicking the Nintendo Switch's UI for the software keyboard with matching colors for the 3 themes we provide.

### Super Smash Bros. Ultimate - Line OSK

osk_ssbu_line_white

### Super Mario Maker 2 - Box OSK

osk_smm2_box_dark

### Pokemon Sword and Shield - Numeric Pad OSK

osk_swsh_num_midnight

### Text Check Error Dialog overlaid on top of the OSK

osk_ssbu_error_dark

### Here's the software keyboard in action in Pokemon Sword and Shield with controller input

https://user-images.githubusercontent.com/39850852/113447128-3ad8e500-93c7-11eb-97f2-d99f4aba597a.mp4

### Fully functional inline software keyboard in Monster Hunter Generations Ultimate - saves no longer needed

https://user-images.githubusercontent.com/39850852/113447306-93a87d80-93c7-11eb-8846-db9596ef190b.mp4

## Error Display Applet

In developing the OSK we quickly found another use for the overlays - the Error Applet. This also accepts controller input.

This applet may show when games have errors on unimplemented features, such as Pokemon Sword/Shield's Y-comm feature or Super Smash Bros. Ultimate's camera mode when taking a screenshot. This is what it looks like

### Pokemon Sword/Shield Y-comm Error

image

[...]

PR #6135, opened 2021-04-03
configure_graphics: Add Borderless Windowed fullscreen mode

Morph1984:
### This commit was split from #6133

The borderless windowed fullscreen mode solves several issues with the presentation of the overlay dialogs and on-screen keyboard in exclusive fullscreen mode, and also has other benefits such as smoother gameplay, lower latency and a significant reduction in screen tearing.

If you encounter issues with this mode, you can swap to the previous exclusive fullscreen mode under Configure > Graphics > Fullscreen Mode

image

### DIFF
--- a/src/core/settings.h
+++ b/src/core/settings.h
@@ -139,6 +139,7 @@ struct Values {
     Setting<int> vulkan_device;
 
     Setting<u16> resolution_factor{1};
+    Setting<int> fullscreen_mode;
     Setting<int> aspect_ratio;
     Setting<int> max_anisotropy;
     Setting<bool> use_frame_limit;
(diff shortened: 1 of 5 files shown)


v1993:
Testing this on Linux + KDE Plasma (dual-screen side-by-side setup) and results are rather unstable for me. Window often gets resized to wrong resolution, either too short vertically or just to seemingly random resolution. Toggling single-window mode makes matters worse, with result sometimes not rendering anything at all.

v1993:
Nitpick: Exclusive is now default on Linux, but menu labels don't seem to reflect this, making it confusing.

Morph1984:
> Nitpick: Exclusive is now default on Linux, but menu labels don't seem to reflect this, making it confusing.

I suppose we can remove the default label then

PR #6142, opened 2021-04-04
program_metadata: Avoid reference binding to misaligned address

lat9nq:
Undefined Behaviour Sanitizer identified a reference binding to a misaligned address when logging the acid_file_access permissions. Copy the value into a variable before sending it to fmt.

Issue was found while booting SSBU.

### DIFF
--- a/src/core/file_sys/program_metadata.cpp
+++ b/src/core/file_sys/program_metadata.cpp
@@ -34,11 +34,55 @@ Loader::ResultStatus ProgramMetadata::Load(VirtualFile file) {
         return Loader::ResultStatus::ErrorBadACIHeader;
     }
 
-    if (sizeof(FileAccessControl) != file->ReadObject(&acid_file_access, acid_header.fac_offset)) {
+    // Load acid_file_access per-component instead of the entire struct, since this struct does not
+    // reflect the layout of the real data.
+    std::size_t current_offset = acid_header.fac_offset;
+    if (sizeof(FileAccessControl::version) != file->ReadBytes(&acid_file_access.version,
+                                                              sizeof(FileAccessControl::version),
+                                                              current_offset)) {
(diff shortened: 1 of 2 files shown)


src/core/file_sys/program_metadata.cpp
```diff
@@ -142,13 +142,16 @@ void ProgramMetadata::Print() const {
 
     LOG_DEBUG(Service_FS, " > Address space:       {}\n", address_space);
 
+    // Avoid reference binding to misaligned address
+    const u64_le permissions = acid_file_access.permissions;
```
lioncash, line 146: This is due to the [packing](https://github.com/yuzu-emu/yuzu/blob/4cc10e7373f87aec6bf364823394baf738428e49/src/core/file_sys/program_metadata.h#L139) misaligning the data

If possible, avoiding the packing on the struct itself and only packing the data on write to disk is likely a nicer approach from a definedness standpoint. I haven't looked into how feasible that would be though.

lat9nq: Yeah, I saw that, but I didn't look into how the struct was being loaded. I was afraid that the struct's layout reflects the layout of the actual data.

lat9nq: I don't know, looking at the way that yuzu reads in the struct [here](https://github.com/yuzu-emu/yuzu/blob/master/src/core/file_sys/vfs.h#L136), [from here](https://github.com/yuzu-emu/yuzu/blob/master/src/core/file_sys/program_metadata.cpp#L37), we can either do the one copy like I proposed earlier, or unpack the struct and do a ReadBytes for each of its members.

lioncash: I'd be fine with whichever makes it more difficult for this scenario to accidentally happen again. If neither does that, I'm totally fine with this being merged as is.

bunnei (approved):

PR #6143, opened 2021-04-04
nvhost_ctrl_gpu: Avoid sending null pointer to memcpy

lat9nq:
Undefined Behaviour Sanitizer reports a null pointer is being sent to memcpy, thought it's "guaranteed to never be null". Guard it with an if statement, and log when the action has been averted.

Issue was found while booting SSBU. This change may be needed on the rest of the functions in nvhost_ctrl_gpu, but ubsan has not caught anything else there, yet. *Fixing the error, not any potential cause.*

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_ctrl_gpu.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_ctrl_gpu.cpp
@@ -248,7 +248,13 @@ NvResult nvhost_ctrl_gpu::ZBCSetTable(const std::vector<u8>& input, std::vector<
     IoctlZbcSetTable params{};
     std::memcpy(&params, input.data(), input.size());
     // TODO(ogniK): What does this even actually do?
-    std::memcpy(output.data(), &params, output.size());
+
+    // Prevent null pointer being passed as arg 1
+    if (output.empty()) {
+        LOG_WARNING(Service_NVDRV, "Avoiding passing null pointer to memcpy");
+    } else {
+        std::memcpy(output.data(), &params, output.size());
+    }
     return NvResult::Success;
 }
 


src/core/hle/service/nvdrv/devices/nvhost_ctrl_gpu.cpp
```diff
@@ -248,7 +248,13 @@ NvResult nvhost_ctrl_gpu::ZBCSetTable(const std::vector<u8>& input, std::vector<
     IoctlZbcSetTable params{};
     std::memcpy(&params, input.data(), input.size());
     // TODO(ogniK): What does this even actually do?
-    std::memcpy(output.data(), &params, output.size());
+
+    // Prevent null pointer being passed as arg 1
+    if (output.data() != nullptr) {
```
lioncash, line 253: This should really be an `empty()` check, rather than querying if the internal buffer of the vector is null

src/core/hle/service/nvdrv/devices/nvhost_ctrl_gpu.cpp
```diff
@@ -248,7 +248,13 @@ NvResult nvhost_ctrl_gpu::ZBCSetTable(const std::vector<u8>& input, std::vector<
...
+        std::memcpy(output.data(), &params, output.size());
+    } else {
+        LOG_WARNING(, "Avoiding passing null pointer to memcpy");
```
lioncash, line 256: This is missing a log category

src/core/hle/service/nvdrv/devices/nvhost_ctrl_gpu.cpp
```diff
@@ -248,7 +248,13 @@ NvResult nvhost_ctrl_gpu::ZBCSetTable(const std::vector<u8>& input, std::vector<
     IoctlZbcSetTable params{};
     std::memcpy(&params, input.data(), input.size());
     // TODO(ogniK): What does this even actually do?
-    std::memcpy(output.data(), &params, output.size());
+
+    // Prevent null pointer being passed as arg 1
+    if (!output.empty()) {
```
lioncash, line 253: This can be inverted to make the true case first, so the condition can be read without the negation

bunnei (approved):

PR #6144, opened 2021-04-04
texutre_cache: util: Avoid shifting by large exponents

lat9nq:
Undefined Behaviour Sanitizer detected shifts as large as 2048 on u32 integers. Applies a modulus on the exponents for now.

@ReinUsesLisp please let me know at least how to verify that this is the correct behaviour.


FernandoS27:
I do not agree with rabit hole fixes like this. This probably is hiding an underlying issue within the cache. It's best to go to the bottom of it and figure it out.

PR #6145, opened 2021-04-04
nvhost_nvdec_common: Avoid memcpy with null pointers

lat9nq:
Avoid sending null pointer to memcpy as reported by Undefined Behaviour Sanitizer.

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_nvdec_common.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_nvdec_common.cpp
@@ -23,17 +23,22 @@ namespace {
 template <typename T>
 std::size_t SpliceVectors(const std::vector<u8>& input, std::vector<T>& dst, std::size_t count,
                           std::size_t offset) {
-    std::memcpy(dst.data(), input.data() + offset, count * sizeof(T));
-    offset += count * sizeof(T);
-    return offset;
+    if (!dst.empty()) {
+        std::memcpy(dst.data(), input.data() + offset, count * sizeof(T));
+    }
+    return 0;
 }
 
 // Write vectors will write data to the output buffer
 template <typename T>
 std::size_t WriteVectors(std::vector<u8>& dst, const std::vector<T>& src, std::size_t offset) {
-    std::memcpy(dst.data() + offset, src.data(), src.size() * sizeof(T));
-    offset += src.size() * sizeof(T);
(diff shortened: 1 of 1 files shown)


src/core/hle/service/nvdrv/devices/nvhost_nvdec_common.cpp
```diff
@@ -23,17 +23,27 @@ namespace {
 template <typename T>
 std::size_t SpliceVectors(const std::vector<u8>& input, std::vector<T>& dst, std::size_t count,
                           std::size_t offset) {
-    std::memcpy(dst.data(), input.data() + offset, count * sizeof(T));
-    offset += count * sizeof(T);
-    return offset;
+    if (!dst.empty()) {
```
lioncash, line 26: See [here](https://github.com/yuzu-emu/yuzu/pull/6146/files#r606890328) for replacing the memcpy calls with std::copy calls, which do the correct thing and also allow you to remove the else cases.

lat9nq: Getting errors when I try to compile something like that. 
i.e. in WriteVectors(), `std::copy(src.begin(), src.begin() + src.size() * sizeof(T), dst.begin());` causes `/usr/include/c++/10.2.0/bits/stl_algobase.h:380:18: error: cannot convert ‘const Service::Nvidia::Devices::nvhost_nvdec_common::CommandBuffer’ to ‘unsigned char’ in assignment`
and a few other alike template issues.

lioncash: Ah, I missed the fact the input vector and output vector can have different types.

src/core/hle/service/nvdrv/devices/nvhost_nvdec_common.cpp
```diff
@@ -23,17 +23,27 @@ namespace {
...
+        std::memcpy(dst.data(), input.data() + offset, count * sizeof(T));
+        offset += count * sizeof(T);
+        return offset;
+    } else {
+        LOG_WARNING(Service_NVDRV, "Avoiding sending null pointer to memcpy");
```
lioncash, line 31: This and the log in the other function can be removed, Just returning 0 in both cases should be sufficient.

bunnei (approved):

PR #6146, opened 2021-04-04
vp9: Avoid memcpy with null pointers

lat9nq:
Avoid sending null pointer to memcpy as reported by Undefined Behaviour Sanitizer.

### DIFF
--- a/src/video_core/command_classes/codecs/vp9.cpp
+++ b/src/video_core/command_classes/codecs/vp9.cpp
@@ -2,8 +2,9 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include <cstring> // for std::memcpy
+#include <algorithm> // for std::copy
 #include <numeric>
+#include "common/assert.h"
 #include "video_core/command_classes/codecs/vp9.h"
 #include "video_core/gpu.h"
 #include "video_core/memory_manager.h"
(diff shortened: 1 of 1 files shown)


src/video_core/command_classes/codecs/vp9.cpp
```diff
@@ -822,10 +823,18 @@ const std::vector<u8>& VP9::ComposeFrameHeader(const NvdecCommon::NvdecRegisters
     // Write headers and frame to buffer
     frame.resize(uncompressed_header.size() + compressed_header.size() + bitstream.size());
     std::memcpy(frame.data(), uncompressed_header.data(), uncompressed_header.size());
-    std::memcpy(frame.data() + uncompressed_header.size(), compressed_header.data(),
-                compressed_header.size());
-    std::memcpy(frame.data() + uncompressed_header.size() + compressed_header.size(),
-                bitstream.data(), bitstream.size());
+    if (!compressed_header.empty()) {
```
lioncash, line 826: Maybe replace the memcpy calls with std::copy calls? These will predictably do the correct thing, even with an empty range

e.g.

```cpp
std::copy(compressed_header.begin(), compressed_header.end(),
          frame.begin() + uncompressed_header.size());
```

bunnei (approved):

PR #6147, opened 2021-04-04
buffer_base: Avoid shift with large exponent

lat9nq:
Undefined Behaviour Sanitizer reports a shift with an exponent larger than 64 on a 64-bit integer (i.e. 68 while booting SSBU.) Use modulus to restrict it instead.

@ReinUsesLisp please let me know at least how to verify that this exhibits the correct behaviour.


degasus:
I think this implementation with masks is wrong. What do you think about `if (empty_bits == PAGES_PER_WORD) break;` instead?

PR #6167, opened 2021-04-08
service: time: Fix CalculateStandardUserSystemClockDifferenceByUser

### LINKED ISSUE #6165: clock_types: Signed integer overflow in Pokémon Sword and LGPE Demo
Undefined behaviour sanitizer reports `src/./core/hle/service/time/clock_types.h:66:25: runtime error: signed integer overflow: 139682469932695 * 1000000000 cannot be represented in type 'long int'` while starting Pokemon Sword and Pokemon Let's Go Pikachu/Eevee Demo. This occurs when meeting Hop for the first time in Sword, or meeting Prof. Oak for the first time in LGPE Demo.

Backtrace (different instance than the above report): [clock_mul_overflow.txt](https://github.com/yuzu-emu/yuzu/files/6275017/clock_mul_overflow.txt)

I would have tried suggesting a fix in a PR like the other ubsan issues I've PR'd on, but this one seems to be a deeper issue that I don't understand. Minor severity?

Morph1984:
CalculateStandardUserSystemClockDifferenceByUser passes in the ClockSnapshots through 2 input buffers and not as raw arguments. Fix this by reading the 2 input buffers instead of popping raw arguments.

Fixes #6165

### DIFF
--- a/src/core/hle/service/time/time.cpp
+++ b/src/core/hle/service/time/time.cpp
@@ -321,9 +321,14 @@ void Module::Interface::CalculateStandardUserSystemClockDifferenceByUser(
     Kernel::HLERequestContext& ctx) {
     LOG_DEBUG(Service_Time, "called");
 
-    IPC::RequestParser rp{ctx};
-    const auto snapshot_a = rp.PopRaw<Clock::ClockSnapshot>();
-    const auto snapshot_b = rp.PopRaw<Clock::ClockSnapshot>();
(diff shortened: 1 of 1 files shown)


bunnei (approved):

lat9nq:
Indeed fixes the linked issue. Values being sent to `Clock::TimeSpanType::FromSeconds` are now sensible, on the order of seconds rather than (tens or millions of) years.

PR #6170, opened 2021-04-08
service: time: Setup the network clock with the local clock context

Morph1984:
Setup the network clock with the local clock context instead of a default clock context to prevent a clock source id mismatch when getting the current network time

This fixes Super Smash Bros. Ultimate's Spirit Board resetting its timers on reboot.

**Note that this does NOT fix Pokemon Sword/Shield's time progression when the game is not running**

### DIFF
--- a/src/core/hle/service/time/time_manager.cpp
+++ b/src/core/hle/service/time/time_manager.cpp
@@ -44,7 +44,11 @@ struct TimeManager::Impl final {
         const auto system_time{Clock::TimeSpanType::FromSeconds(GetExternalRtcValue())};
         SetupStandardSteadyClock(system, Common::UUID::Generate(), system_time, {}, {});
         SetupStandardLocalSystemClock(system, {}, system_time.ToSeconds());
-        SetupStandardNetworkSystemClock({}, standard_network_clock_accuracy);
+
+        Clock::SystemClockContext clock_context{};
+        standard_local_system_clock_core.GetClockContext(system, clock_context);
+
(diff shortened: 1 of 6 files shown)


bylaws:
You do actually have the network clock implemented, you just initialise it a bit wrong. Rather than introducing an inaccuracy here just setup the network clock with the steady clock context https://github.com/skyline-emu/skyline/blob/2edca83b8b3577a24d3f12c1d3a97e74b2754048/app/src/main/cpp/skyline/services/timesrv/core.cpp#L244

bylaws:
(this is also the case on hw as well, both use the same steady clock for counting)

src/core/hle/service/time/time_manager.cpp
```diff
@@ -157,11 +157,14 @@ struct TimeManager::Impl final {
...
     }
 
-    void SetupStandardNetworkSystemClock(Clock::SystemClockContext clock_context,
+    void SetupStandardNetworkSystemClock(Core::System& system,
+                                         Clock::SystemClockContext clock_context,
                                          Clock::TimeSpanType sufficient_accuracy) {
         standard_network_system_clock_core.SetUpdateCallbackInstance(
             network_system_clock_context_writer);
 
+        standard_local_system_clock_core.GetClockContext(system, clock_context);
```
bylaws, line 166: Rather than this you can just change the clock context you pass in to the call

bylaws (approved):

bunnei (approved):

PR #6172, opened 2021-04-08
externals: Search for shared opus installation.

degasus:
We had used conan for opus before, but there was a bug in the AVX detection.
However we still had the Findopus.cmake file within the repository, but not used.

This patch reenables the Findopus helper and prefer the system wide installation of opus.

This will use the opus .so/.dll on linux-fresh and mingw builds (which I think is fine). MSVC builds are not affected.


degasus:
FYI: We explicitly install opus on our windows CI: https://github.com/yuzu-emu/build-environments/blob/master/linux-mingw/Dockerfile#L29 -- So this PR will currently use it. I assume we don't want to ship this DLL, so this line might need to be dropped.

We also do so on linux-fresh, but here I think this is fine.

bunnei (approved):

PR #6180, opened 2021-04-11
renderer_vulkan: Check return value of AcquireNextImage

Joshua-Ashton:
We can get into a really bad state by ignoring this
leading to device loss and using incorrect resources.


ReinUsesLisp (approved):
lgtm, thanks for looking into this

PR #6188, opened 2021-04-12
vk_texture_cache: Make use of bit_cast where applicable

lioncash:


FernandoS27:
you should use `Common::Bitcast<T>() `until there's wide compiler support for `std::bit_cast<T>()`. Please refer to https://en.cppreference.com/w/cpp/compiler_support clang++ still does not support it

bunnei (approved):

PR #6196, opened 2021-04-14
core: settings: Add setting for debug assertions and disable by default.

bunnei:
- This is a developer-only setting and no longer needs to be enabled by default.
- Also adds "use_auto_stub" setting to SDL frontend while we are here.
- Supersedes #1340.


src/common/assert.cpp
```diff
@@ -3,9 +3,12 @@
 // Refer to the license.txt file included.
 
 #include "common/assert.h"
-
 #include "common/common_funcs.h"
 
+#include "core/settings.h"
```
lioncash, line 8: common shouldn't have a dependency on the core library

bunnei: Are you suggesting have a static global assert handler that frontends set? Seems like we also could just move settings to common, since there is other uses of settings within common (logging)

lioncash: That works as well

src/common/settings.h
```diff
@@ -11,16 +11,13 @@
...
-#include "common/common_types.h"
-#include "input_common/settings.h"
 
-namespace Core {
-class System;
-}
+#include "common/common_types.h"
+#include "common/settings_input.h"
 
 namespace Settings {
```
lioncash, line 18: Should have the `Common::Settings` (or have `Settings` changed to `Common`)

src/common/settings_input.h
```diff
@@ -6,6 +6,7 @@
 
 #include <array>
 #include <string>
+
 #include "common/common_types.h"
 
 namespace Settings {
```
lioncash, line 12: Ditto

lioncash (commented):
LGTM once the namespaces are amended

src/common/settings.h
bunnei: This was intentional for now.
* This is many hundreds of lines changed across the whole project and really seems out of scope for the original change.
* `Settings` was already unique and concise enough. Seems like a lot of unnecessary bloat to require `Common::Settings::` in hundreds of places.
* `Common` namespace is often used in places where we don't have another namespace. It's not used for all Common code (e.g. see `Log::`, among others). If we want to change this practice, someone could refactor all Common. 
* We have a few big changes that are currently changes 1000's of lines of code (Hades, Eleuthia, and a massive kernel refactor I am working on), this seems like needless pain for those involved to do right now -- at the very least, we can do a change like this during a time when there isn't as much widespread refactoring.

lioncash: Ideally, we'd just be able to do:

```cpp
Common::Settings()
```

to get at the settings instance instead of top-level:

```cpp
Settings::values
```

to keep the global itself hidden, but yeah, that can wait until after the refactors.

The only reason the Log namespace isn't in Common is because I forgot to move that over when unifying the common namespace.

lioncash (approved):

src/common/settings.h
bunnei: Sounds good, thanks!

PR #6202, opened 2021-04-16
Config for motion control of my yuzu

ToyBonGaming:


v1993:
That's not how PRs works :laughing:. You're not supposed to make everyone have your config by default. Keep your configs to yourself.

german77:
I Believe you want to keep this change only for your fork. The default values are defined by the servers not the client (yuzu).

PR #6222, opened 2021-04-21
cmake: Download Boost from ext-linux-bin if inadequate local version

lat9nq:
Since #6207, building SDL2 from externals on systems with low or no system Boost version is incompatible with Conan's version of libiconv, a requirement of Conan's Boost package. The solution proposed here is to use the same Boost package as the linux-fresh container. This tells CMake to download boost_1_75_0.tar.xz from [yuzu-emu/ext-linux-bin](https://github.com/yuzu-emu/ext-linux-bin) at CMake's configuration step, much the same way Qt and FFmpeg are downloaded for Windows.

This also makes DownloadExternals.cmake cross-platform between Windows and Linux. Although the CMake code is not entirely specific to Linux, only Linux has Boost libraries available at ext-linux-bin, whereas there is no equivalent Boost package for Windows at ext-windows-bin. This also does not work for any of the other *nix's: Conan is there as a fallback, but those will need to have at least some updated libraries visible to CMake from elsewhere (basically, this PR changes nothing for the other *nix's.)

IIRC the linux-mingw bot will check DownloadExternals.cmake for WIN32 (it needs it for FFmpeg), but I would be more comfortable if someone using the MSVC toolchain would test this PR, as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,8 @@ CMAKE_DEPENDENT_OPTION(YUZU_USE_BUNDLED_QT "Download bundled Qt binaries" ON "EN
 
 option(ENABLE_WEB_SERVICE "Enable web services (telemetry, etc.)" ON)
 
+option(YUZU_USE_BUNDLED_BOOST "Download bundled Boost" OFF)
+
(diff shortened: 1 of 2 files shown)


bunnei:
@lat9nq needs a rebase

bunnei (approved):

PR #6243, opened 2021-04-25
input_common: Reset GC sticks center by measuring multiple packets

german77:
Sometimes when connecting a new controller to the adapter it will send an invalid package that will mess up with with the joystick center. This PR eliminates this issue by reading 4 continuos packages with the same joystick center, ensuring that the center will always be set correctly.


src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -155,12 +155,19 @@ void Adapter::UpdateStateAxes(std::size_t port, const AdapterPayload& adapter_pa
     for (const PadAxes axis : axes) {
         const auto index = static_cast<std::size_t>(axis);
         const u8 axis_value = adapter_payload[offset + 3 + index];
-        if (pads[port].axis_origin[index] == 255) {
+        if (pads[port].reset_origin_counter <= 3) {
```
ameerj, line 158: you can probably move the increment into the body of this condition and avoid the second `if (pads[port].reset_origin_counter <= 3)` check.

bunnei (approved):

PR #6244, opened 2021-04-26
input_common: Implement SDL motion

german77:
Adds the motion backend for SDL2 since it's now supported. Need testing to find supported controllers.

- [x] Adds auto mapping for motion. 
- [ ] Calibrate the sensors
- [ ] Fine tune the SDL motion
- [x] Find supported controllers, currently PS4, PS5, Pro controller, Joycons.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -29,6 +29,7 @@
 #endif
 
 #include "common/logging/log.h"
+#include "common/math_util.h"
 #include "common/param_package.h"
 #include "common/settings_input.h"
 #include "common/threadsafe_queue.h"
@@ -68,11 +69,55 @@ public:
     SDLJoystick(std::string guid_, int port_, SDL_Joystick* joystick,
                 SDL_GameController* game_controller)
         : guid{std::move(guid_)}, port{port_}, sdl_joystick{joystick, &SDL_JoystickClose},
-          sdl_controller{game_controller, &SDL_GameControllerClose} {}
+          sdl_controller{game_controller, &SDL_GameControllerClose} {
+        EnableMotion();
+    }
+
+    void EnableMotion() {
+        if (sdl_controller) {
+            SDL_GameController* controller = sdl_controller.get();
+            if (SDL_GameControllerHasSensor(controller, SDL_SENSOR_ACCEL) && !has_accel) {
+                SDL_GameControllerSetSensorEnabled(controller, SDL_SENSOR_ACCEL, SDL_TRUE);
+                has_accel = true;
+            }
+            if (SDL_GameControllerHasSensor(controller, SDL_SENSOR_GYRO) && !has_gyro) {
+                SDL_GameControllerSetSensorEnabled(controller, SDL_SENSOR_GYRO, SDL_TRUE);
+                has_gyro = true;
+            }
+        }
+    }
 
     void SetButton(int button, bool value) {
         std::lock_guard lock{mutex};
         state.buttons.insert_or_assign(button, value);
+    }
+
+    void SetMotion(SDL_ControllerSensorEvent event) {
+        constexpr float gravity_constant = 9.80665f;
+        std::lock_guard lock{mutex};
+        u64 time_difference = event.timestamp - last_motion_update;
+        last_motion_update = event.timestamp;
+        switch (event.sensor) {
+        case SDL_SENSOR_ACCEL: {
+            const Common::Vec3f acceleration = {-event.data[0], event.data[2], -event.data[1]};
+            motion.SetAcceleration(acceleration / gravity_constant);
+            break;
+        }
+        case SDL_SENSOR_GYRO: {
+            const Common::Vec3f gyroscope = {event.data[0], -event.data[2], event.data[1]};
+            motion.SetGyroscope(gyroscope / (Common::PI * 2));
+            break;
+        }
+        }
+
+        // Ignore duplicated timestamps
+        if (time_difference == 0) {
+            return;
+        }
+
+        motion.SetGyroThreshold(0.0001f);
+        motion.UpdateRotation(time_difference * 1000);
+        motion.UpdateOrientation(time_difference * 1000);
     }
 
     bool GetButton(int button) const {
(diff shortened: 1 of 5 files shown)


src/input_common/sdl/sdl_impl.cpp
```diff
@@ -121,6 +133,23 @@ class SDLJoystick {
         return std::make_tuple(x, y);
     }
 
+    const bool HasGyro() const {
```
lioncash, line 136:
```suggestion
    bool HasGyro() const {
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -121,6 +133,23 @@ class SDLJoystick {
...
+        return has_gyro;
+    }
+
+    const bool HasAccel() const {
```
lioncash, line 140:
```suggestion
    bool HasAccel() const {
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -307,6 +338,59 @@ void SDLState::HandleGameControllerEvent(const SDL_Event& event) {
...
+
+    for (const auto& [key, value] : joystick_map) {
+        for (const auto& joystick : value) {
+            auto* const controller = joystick->GetSDLGameController();
+            if (!controller) {
+                continue;
+            }
+
+            MotionInput motion = joystick->GetMotion();
+            float accel[3]{};
```
lioncash, line 355: `std::array`, ditto for `gyro` below this line

This can also allow for elimination of magic `3` constants in several places below.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -307,6 +338,59 @@ void SDLState::HandleGameControllerEvent(const SDL_Event& event) {
...
+                SDL_GameControllerGetSensorData(controller, SDL_SENSOR_ACCEL, accel, 3);
+            }
+            if (joystick->HasGyro()) {
+                SDL_GameControllerGetSensorData(controller, SDL_SENSOR_GYRO, gyro, 3);
+            }
+
+            if (polling) {
+                bool is_motion_shaking = false;
+                for (std::size_t i = 0; i < 3; ++i) {
+                    if (gyro[i] > 5.0f || gyro[i] < -5.0f) {
```
lioncash, line 368: These seem like magic values, ditto for `1.75f` and `-1.75f`

src/input_common/sdl/sdl_impl.h
```diff
@@ -68,6 +70,8 @@ class SDLState : public State {
     /// Returns a custom name for specific controllers because the default name is not correct
     std::string GetControllerName(SDL_GameController* controller) const;
 
+    void Update_Motion();
```
lioncash, line 73:
```suggestion
    void UpdateMotion();
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -307,6 +338,62 @@ void SDLState::HandleGameControllerEvent(const SDL_Event& event) {
...
+                for (std::size_t i = 0; i < accel.size(); ++i) {
+                    if (gyro[i] > gyro_threshold || gyro[i] < -gyro_threshold) {
+                        is_motion_shaking = true;
+                    }
+                    if (accel[i] > accel_threshold || accel[i] < -accel_threshold) {
+                        is_motion_shaking = true;
+                    }
+                }
+                LOG_DEBUG(Input, "Motion accel ({} {} {}) gyro ({} {} {})", accel[0], accel[1],
+                          accel[2], gyro[0], gyro[1], gyro[2]);
```
Morph1984, line 379: Is this left over from debugging

v1993:
Is there a list of supported controllers anywhere? I suspect such support is also very driver-specific, being limited to those that communicate with controller directly, like via hidraw interface (i.e. evdev-based motion in Linux is probably not supported).

german77:
Pro controllers should be supported. A few people tried but no motion was detected

src/input_common/sdl/sdl_impl.cpp
german77: No this is intentional. I might move it somewhere better to reduce spam. This only called when mapping. Helps to find issues with users that report no motion

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -68,13 +68,54 @@ class SDLJoystick {
...
         state.buttons.insert_or_assign(button, value);
     }
 
+    void SetMotion(SDL_ControllerSensorEvent event) {
+        std::lock_guard lock{mutex};
+        u64 time_difference = event.timestamp - last_motion_update;
+        last_motion_update = event.timestamp;
+        switch (event.sensor) {
+        case SDL_SENSOR_ACCEL: {
+
```
Morph1984, line 96: unnecessary newline, ditto with the case directly below this one

RinMaru:
tried the New EA 1665 Motion is not detected for Pro controller

v1993:
It's actually possible to have motion working on Linux. To do so, one must build SDL2 with `-DHIDAPI` (did so separately from yuzu for testing) since hidapi is needed for motion and it's disabled by default on Linux.

I suggest to enable hidapi and build SDL2 from externals by default even in presence of system SDL2. If user insists on not using hidapi, they always can run yuzu in firejail, making SDL2 fall back to more traditional evdev input backend. Typically, you'd prefer evdev input over hidapi: it can be calibrated, it does not interfere with drivers, etc., but since motion input is rather important for yuzu, it makes a lot of sense to use hidapi even with its apparent drawbacks. Additionally, native joycons PR also uses hidapi.

It is possible to avoid using hidapi and get motion data from controllers using evdev input, but SDL does not currently support this. Once it does, it should be safe to stop using hidapi and allow system SDL2 once again. Alternatively, it would also be possible to use evdev directly like dolphin does, but it is not a simple task.

Morph1984 (approved):

Shoegzer:
Noticed your PR was merged @german77 , so I tried to test it.  However, the only game I have to test motion (Splatoon 2) freezes near startup now for some reason, and the UI doesn't pickup any motion in the ds4.  I'm guessing that is due to sdl2 not being built with -hidapi?

v1993:
@Shoegzer Please use yuzu's discord for further discussion. More information (OS? How you installed yuzu?) is needed to diagnose your issue.

PR #6245, opened 2021-04-26
cmake: Only config Boost during find_package

### LINKED ISSUE #6242: Yuzu failing to compile [Linux]
Recent commits of Yuzu fail to compile on linux systems.  I haven't had time to bisect but can if necessary, 

Testing info:

* System: Linux Mint 20.1 / kernel 5.11
* Compiler: gcc-10.2.0
* Problem encountered on commit: 2ff39f6fdc1be6e1ed83b384bb9773049061e6b9
* Last known good commit: 956b2bd332588f078229efc15d5daa5948fad89b
*  Compile log: [here](https://pastebin.com/ydFZAs6K)

lat9nq:
By default, `find_package` will perform Module search. On at least Linux Mint 20 (I'm unable to reproduce this on CentOS 8 and Arch Linux), my guess is this causes CMake to find "dirty" modules that modify the configuration state despite the local Boost version being too low/absent. This causes problems when trying to repeat `find_package` for Boost later on the downloaded binary package: CMake complains about targets having already been added by the first `find_package`.

Use `CONFIG` to put CMake into pure Config mode and avoid Module search, thus avoid entering a invalid configuration state.

Fixes #6242


v1993:
Can confirm that this fixes an issue on KUbuntu 20.10.

Shoegzer:
Confirmed working here too with Mint 20.1 as reported.  Thanks @lat9nq.

bunnei (approved):

PR #6248, opened 2021-04-26
gl_device: Intel: Disable texture view formats workaround on mesa

A-w-x:
please let me know if amd users can also be whitelisted on mesa
no device to test atm

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -241,7 +241,7 @@ Device::Device() {
     has_variable_aoffi = TestVariableAoffi();
     has_component_indexing_bug = is_amd;
     has_precise_bug = TestPreciseBug();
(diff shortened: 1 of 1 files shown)


lat9nq:
> please let me know if amd users can also be whitelisted on mesa

Please do **not** try to fix `is_amd` for different platforms.

AMD devices have different vendors depending on the driver in use. RadeonSI (Mesa) reports `AMD`, while AMDGPU-PRO (Linux) and Windows AMD OpenGL reports `ATI Technologies Inc.`. Thus, we only use AMD-specific fixes on AMD's proprietary driver, which can happen on both Windows and Linux.

A-w-x:
yes, i'm aware of that but i didn't check what was the reported name of the amd drivers in linux, that is why i asked
thanks for clarifying that

bunnei (approved):

PR #6265, opened 2021-05-01
service: filesystem: Return proper error codes for CreateFile

Morph1984:
This improves the accuracy of CreateFile by returning the correct error codes on certain conditions (parent directory does not exist, path already exists).

This fixes saving and the loading of existing saves in New Pokemon Snap

### DIFF
--- a/src/core/file_sys/errors.h
+++ b/src/core/file_sys/errors.h
@@ -9,6 +9,7 @@
 namespace FileSys {
 
 constexpr ResultCode ERROR_PATH_NOT_FOUND{ErrorModule::FS, 1};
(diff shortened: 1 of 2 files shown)


FernandoS27:
Should I mark this a do not merge in liu of the upcoming VFS rework?

Morph1984:
This shouldn't harm any future reworks. It's only a 8 LOC change

TONYSOAD05:
Oh my god, thank you very much for this fix!!! You are a savior lol, waiting to be merged on mainline :), thanks!

bunnei (approved):

PR #6267, opened 2021-05-01
hid: Improve hardware accuracy of gestures

german77:
Touch in Pokemon Snap was broken. This PR rewrites gestures in order to make them more accurate against hardware by implementing all gesture types. Removes any input errors that it previously had. 

- [x] Add multi finger gestures
- [x] Cleanup code to make it easier to read

### DIFF
--- a/src/core/hle/service/hid/controllers/gesture.cpp
+++ b/src/core/hle/service/hid/controllers/gesture.cpp
@@ -1,10 +1,9 @@
-// Copyright 2018 yuzu emulator team
+// Copyright 2021 yuzu Emulator Project
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include <cstring>
-#include "common/common_types.h"
 #include "common/logging/log.h"
+#include "common/math_util.h"
 #include "common/settings.h"
 #include "core/core_timing.h"
 #include "core/frontend/emu_window.h"
@@ -12,10 +11,19 @@
 
 namespace Service::HID {
 constexpr std::size_t SHARED_MEMORY_OFFSET = 0x3BA00;
-constexpr f32 angle_threshold = 0.08f;
-constexpr f32 pinch_threshold = 100.0f;
 
-Controller_Gesture::Controller_Gesture(Core::System& system_) : ControllerBase{system_} {}
+// HW is around 700, value is set to 400 to make it easier to trigger with mouse
+constexpr f32 swipe_threshold = 400.0f; // Threshold in pixels/s
+constexpr f32 angle_threshold = 0.015f; // Threshold in radians
+constexpr f32 pinch_threshold = 0.5f;   // Threshold in pixels
+constexpr f32 press_delay = 0.5f;       // Time in seconds
+constexpr f32 double_tap_delay = 0.35f; // Time in seconds
+
+constexpr f32 Square(s32 num) {
+    return static_cast<f32>(num * num);
+}
+
+Controller_Gesture::Controller_Gesture(Core::System& system) : ControllerBase(system) {}
 Controller_Gesture::~Controller_Gesture() = default;
 
 void Controller_Gesture::OnInit() {
@@ -24,6 +32,8 @@ void Controller_Gesture::OnInit() {
         keyboard_finger_id[id] = MAX_POINTS;
         udp_finger_id[id] = MAX_POINTS;
     }
+    shared_memory.header.entry_count = 0;
+    force_update = true;
 }
 
 void Controller_Gesture::OnRelease() {}
@@ -38,17 +48,23 @@ void Controller_Gesture::OnUpdate(const Core::Timing::CoreTiming& core_timing, u
         shared_memory.header.last_entry_index = 0;
         return;
     }
-    shared_memory.header.entry_count = 16;
 
-    const auto& last_entry = shared_memory.gesture_states[shared_memory.header.last_entry_index];
-    shared_memory.header.last_entry_index = (shared_memory.header.last_entry_index + 1) % 17;
-    auto& cur_entry = shared_memory.gesture_states[shared_memory.header.last_entry_index];
+    ReadTouchInput();
 
-    cur_entry.sampling_number = last_entry.sampling_number + 1;
-    cur_entry.sampling_number2 = cur_entry.sampling_number;
+    GestureProperties gesture = GetGestureProperties();
+    f32 time_difference = static_cast<f32>(shared_memory.header.timestamp - last_update_timestamp) /
+                          (1000 * 1000 * 1000);
 
-    // TODO(german77): Implement all gesture types
+    // Only update if necesary
+    if (!ShouldUpdateGesture(gesture, time_difference)) {
+        return;
+    }
 
+    last_update_timestamp = shared_memory.header.timestamp;
+    UpdateGestureSharedMemory(data, size, gesture, time_difference);
+}
+
+void Controller_Gesture::ReadTouchInput() {
     const Input::TouchStatus& mouse_status = touch_mouse_device->GetStatus();
     const Input::TouchStatus& udp_status = touch_udp_device->GetStatus();
     for (std::size_t id = 0; id < mouse_status.size(); ++id) {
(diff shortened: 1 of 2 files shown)


src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -64,49 +57,144 @@ void Controller_Gesture::OnUpdate(const Core::Timing::CoreTiming& core_timing, u
         }
     }
 
+    bool update_gesture = false;
```
lioncash, line 60: This function is getting quite large now, maybe consider breaking some parts of it out into their own functions.

e.g.

the updating can be like so

```cpp
bool Controller_Gesture::ShouldUpdateGesture() const {
    for (size_t id = 0; id < MAX_POINTS; id++) {
        if (gesture.points[id].x != last_gesture.points[id].x ||
            gesture.points[id].y != last_gesture.points[id].y) {
            return true;
        }
    }

    return false;
}
```

Ditto for some other self-isolated bits below.

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -64,49 +57,144 @@ void Controller_Gesture::OnUpdate(const Core::Timing::CoreTiming& core_timing, u
...
         }
 
-        type = TouchType::Touch;
-        if (gesture.mid_point.x != last_entry.x || gesture.mid_point.y != last_entry.y) {
-            type = TouchType::Pan;
+        // Promote to pan type if touch moved
+        for (size_t id = 0; id < MAX_POINTS && type == TouchType::Idle; id++) {
+            if (gesture.points[id].x != last_gesture.points[id].x ||
+                gesture.points[id].y != last_gesture.points[id].y) {
+                type = TouchType::Pan;
```
lioncash, line 115: Would you not want to break after this line, so there's no need to redundantly check any successive items in the array?

src/core/hle/service/hid/controllers/gesture.h
```diff
@@ -135,5 +135,8 @@ class Controller_Gesture final : public ControllerBase {
     std::array<size_t, MAX_FINGERS> udp_finger_id;
     std::array<Finger, MAX_POINTS> fingers;
     GestureProperties last_gesture;
+    s64_le last_update_timestamp;
```
lioncash, line 138: Wouldn't it be better here to use a std::chrono type to make the timestamp unit explicit?

german77: This timestamp doesn't need any kind of accuracy. It's just used as a relative difference to estimate a touch panning speed

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -63,50 +76,123 @@ void Controller_Gesture::OnUpdate(const Core::Timing::CoreTiming& core_timing, u
...
+    if (last_entry.type == TouchType::Touch && last_entry.point_count == 1 &&
+        time_difference > press_delay) {
+        return true;
+    }
+
+    return false;
+}
+
+void Controller_Gesture::UpdateGestureSharedMemory(u8* data, std::size_t size,
+                                                   GestureProperties gesture,
```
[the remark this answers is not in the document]
german77: Needs to be non const as it's modified

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -12,8 +12,13 @@
...
 namespace Service::HID {
 constexpr std::size_t SHARED_MEMORY_OFFSET = 0x3BA00;
-constexpr f32 angle_threshold = 0.08f;
+
+constexpr f32 angle_threshold = 0.02f;
 constexpr f32 pinch_threshold = 100.0f;
+// HW is around 700, value is set to 400 to make it easier to trigger with mouse
+constexpr float swipe_threshold = 400.0f;
+constexpr float press_delay = 0.500f;      // Time in seconds
+constexpr float double_tap_delay = 0.350f; // Time in seconds
```
FernandoS27, line 21: use `f32` instead of `float`

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -116,12 +206,99 @@ void Controller_Gesture::OnUpdate(const Core::Timing::CoreTiming& core_timing, u
...
 
     std::memcpy(data + SHARED_MEMORY_OFFSET, &shared_memory, sizeof(SharedMemory));
 }
 
+void Controller_Gesture::SetTapEvent(GestureProperties& gesture, GestureProperties& last_gesture,
+                                     TouchType& type, Attribute& attributes) {
+    type = TouchType::Tap;
+    gesture = last_gesture;
+    force_update = true;
+    float tap_time_difference =
```
FernandoS27, line 219: ditto.

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -116,12 +206,99 @@ void Controller_Gesture::OnUpdate(const Core::Timing::CoreTiming& core_timing, u
...
+void Controller_Gesture::UpdatePanEvent(GestureProperties& gesture, GestureProperties& last_gesture,
+                                        TouchType& type, float time_difference) {
+    auto& cur_entry = shared_memory.gesture_states[shared_memory.header.last_entry_index];
+    auto& last_entry =
+        shared_memory.gesture_states[(shared_memory.header.last_entry_index + 16) % 17];
+    cur_entry.delta_x = gesture.mid_point.x - last_entry.x;
+    cur_entry.delta_y = gesture.mid_point.y - last_entry.y;
+
+    cur_entry.vel_x = static_cast<float>(cur_entry.delta_x) / time_difference;
+    cur_entry.vel_y = static_cast<float>(cur_entry.delta_y) / time_difference;
```
FernandoS27, line 236: ditto

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -116,12 +206,99 @@ void Controller_Gesture::OnUpdate(const Core::Timing::CoreTiming& core_timing, u
...
+    }
+}
+
+void Controller_Gesture::EndPanEvent(GestureProperties& gesture, GestureProperties& last_gesture,
+                                     TouchType& type, float time_difference) {
+    auto& cur_entry = shared_memory.gesture_states[shared_memory.header.last_entry_index];
+    auto& last_entry =
+        shared_memory.gesture_states[(shared_memory.header.last_entry_index + 16) % 17];
+    cur_entry.vel_x =
+        static_cast<float>(last_entry.delta_x) / (last_pan_time_difference + time_difference);
```
FernandoS27, line 258: ditto.

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -116,12 +206,99 @@ void Controller_Gesture::OnUpdate(const Core::Timing::CoreTiming& core_timing, u
...
+    cur_entry.vel_y =
+        static_cast<float>(last_entry.delta_y) / (last_pan_time_difference + time_difference);
+    const float curr_vel =
+        std::sqrt((cur_entry.vel_x * cur_entry.vel_x) + (cur_entry.vel_y * cur_entry.vel_y));
```
FernandoS27, line 262: ditto.

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -12,8 +12,14 @@
...
-constexpr f32 angle_threshold = 0.08f;
-constexpr f32 pinch_threshold = 100.0f;
+
+// HW is around 700, value is set to 400 to make it easier to trigger with mouse
+constexpr f32 swipe_threshold = 400.0f;  // Threshold in pixels/s
+constexpr f32 angle_threshold = 0.015f;  // Threshold in radians
+constexpr f32 pinch_threshold = 100.0f;  // Threshold in pixels
+constexpr f32 press_delay = 0.500f;      // Time in seconds
+constexpr f32 double_tap_delay = 0.350f; // Time in seconds
+constexpr f32 PI = 3.1415926f;
```
Morph1984, line 22: this constant is defined in common/math_util.h as `Common::PI`

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -12,8 +12,14 @@
...
 namespace Service::HID {
 constexpr std::size_t SHARED_MEMORY_OFFSET = 0x3BA00;
-constexpr f32 angle_threshold = 0.08f;
-constexpr f32 pinch_threshold = 100.0f;
+
+// HW is around 700, value is set to 400 to make it easier to trigger with mouse
+constexpr f32 swipe_threshold = 400.0f;  // Threshold in pixels/s
+constexpr f32 angle_threshold = 0.015f;  // Threshold in radians
+constexpr f32 pinch_threshold = 100.0f;  // Threshold in pixels
+constexpr f32 press_delay = 0.500f;      // Time in seconds
```
Morph1984, line 20:
```suggestion
constexpr f32 press_delay = 0.5f;      // Time in seconds
```

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -182,20 +395,32 @@ Controller_Gesture::GestureProperties Controller_Gesture::GetGestureProperties()
...
+            gesture.points[id].y =
+                static_cast<int>(active_fingers[id].y * Layout::ScreenDocked::Height);
+        }
+
         gesture.mid_point.x += static_cast<int>(gesture.points[id].x / gesture.active_points);
         gesture.mid_point.y += static_cast<int>(gesture.points[id].y / gesture.active_points);
     }
 
     for (size_t id = 0; id < gesture.active_points; ++id) {
         const double distance =
```
Morph1984, line 412: eliminates the need to re-cast distance to an f32
```suggestion
        const f32 distance =
```

src/core/hle/service/hid/controllers/gesture.cpp
```diff
@@ -182,20 +395,32 @@ Controller_Gesture::GestureProperties Controller_Gesture::GetGestureProperties()
             static_cast<int>(active_fingers[id].x * Layout::ScreenUndocked::Width);
         gesture.points[id].y =
             static_cast<int>(active_fingers[id].y * Layout::ScreenUndocked::Height);
+
+        // Hack: There is no touch in docked but games still allow it
+        if (Settings::values.use_docked_mode.GetValue()) {
+            gesture.points[id].x =
+                static_cast<int>(active_fingers[id].x * Layout::ScreenDocked::Width);
+            gesture.points[id].y =
+                static_cast<int>(active_fingers[id].y * Layout::ScreenDocked::Height);
```
Morph1984, line 404:
```suggestion
                static_cast<s32>(active_fingers[id].x * Layout::ScreenDocked::Width);
            gesture.points[id].y =
                static_cast<s32>(active_fingers[id].y * Layout::ScreenDocked::Height);
```

src/core/hle/service/hid/controllers/gesture.cpp
Morph1984: Also is distance meant to be a^2 + b^2 or sqrt(a^2 + b^2)

Morph1984:
Also I'd recommend a helper function to multiply 2 numbers together, to consolidate usages of `a * a` or `std::pow(base, 2), placed in an anonymous namespace

```cpp
namespace {

constexpr f32 square(f32 num) {
    return num * num;
}

} // Anonymous namespace

**Comment by ogniK5377** [comment]
> Also I'd recommend a helper function to multiply 2 numbers together, to consolidate usages of `a * a` or `std::pow(base, 2)`, placed in an anonymous namespace. Note `constexpr` is optional since none of the usages are in a constexpr context
> 
> ```c++
> namespace {
> 
> constexpr inline f32 Square(f32 num) {
>     return num * num;
> }
> 
> } // Anonymous namespace 
> ```

Have to agree with this, most compilers do optimize for `std::pow(x, 2);` HOWEVER, MSVC does not do this. So it might be worth using what morph suggested over `std::pow(x, 2);`

File: src/core/hle/service/hid/controllers/gesture.cpp (hunk 12 above)
Comment (german77) [reply src/core/hle/service/hid/controllers/gesture.cpp]: I will have to confirm this one. Parameters did look like they are squared but now I'm in doubt

Comment (german77) [reply src/core/hle/service/hid/controllers/gesture.cpp]: It should be sqrt(a^2 + b^2)

File: src/core/hle/service/hid/controllers/gesture.cpp [hunk 14]
```diff
@@ -1,19 +1,29 @@
...
-#include "common/common_types.h"
 #include "common/logging/log.h"
+#include "common/math_util.h"
 #include "common/settings.h"
 #include "core/core_timing.h"
 #include "core/frontend/emu_window.h"
 #include "core/hle/service/hid/controllers/gesture.h"
 
+namespace {
+constexpr inline f32 Square(s32 num) {
```
lioncash, line 13:
```suggestion
constexpr f32 Square(s32 num) {
```

since this function already has internal linkage, and inline signifies external linkage.

This namespace should also be inside the Service::HID namespace.

Morph1984 (approved):

PR #6275, opened 2021-05-03
input_common: Release mouse buttons on out of focus

german77:
When the OSK opens mouse input gets pressed permanently until next event. This will release any mouse button if the game loses focus.

### DIFF
--- a/src/input_common/mouse/mouse_input.cpp
+++ b/src/input_common/mouse/mouse_input.cpp
@@ -143,6 +143,15 @@ void Mouse::ReleaseButton(MouseButton button_) {
     mouse_info[button_index].data.axis = {0, 0};
 }
 
+void Mouse::ReleaseAllButtons() {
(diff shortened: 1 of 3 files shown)


src/input_common/mouse/mouse_input.cpp
```diff
@@ -143,6 +143,15 @@ void Mouse::ReleaseButton(MouseButton button_) {
     mouse_info[button_index].data.axis = {0, 0};
 }
 
+void Mouse::ReleaseAllButtons() {
+    buttons = 0;
+    for (std::size_t button_index = 0; button_index < mouse_info.size(); button_index++) {
```
lioncash, line 148: This seems like it can be compressed into a ranged for loop like so:
```suggestion
    for (auto& button : mouse_info) {
```

bunnei (approved):

Issue #2703, opened 2019-07-09
Proper viewport flip emulation


Eloeri:
Games like Celeste and Axiom verge have the issue where the screen is rendered upside down, changes when APIs are swapped.

[](url)
<meta name="titleids"  content="01002B30028F6000,0100052004384000">

![celeste1](https://user-images.githubusercontent.com/42394653/60931265-93289480-a27e-11e9-8b64-b8ebaa914e06.jpg)

![celeste2](https://user-images.githubusercontent.com/42394653/60931264-93289480-a27e-11e9-9ed2-b44e4c7c671b.jpg)

![celeste3](https://user-images.githubusercontent.com/42394653/60931263-93289480-a27e-11e9-8b23-ea556d22611d.jpg)

![celeste4](https://user-images.githubusercontent.com/42394653/60931266-93289480-a27e-11e9-931e-f3113bc34c1b.jpg)

ReinUsesLisp:
The issue I had with Vulkan was related to Android's vertical flip not being emulated.

On OpenGL our current approach harms transform feedbacks writing to `gl_Position.y` due to in-vertex shader flipping. To fix this we have to use GL_ARB_clip_control.

alivefromupstairs:
any solution?

ReinUsesLisp:
gdkchan from Ryujinx found there's a register in viewport configuration that controls an extra flipping flag. It is exposed in Nvidia's OpenGL driver through [NV_viewport_swizzle](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_viewport_swizzle.txt). Emulating this through the extension or manually should fix these issues.

Pi-Mania:
Run in Docked-Mode Fixes Screen Flip issues.

legend800:
Same for Fight N Rage and Streets of Rage 4.

benderscruffy:
not fixed on my AMD RX 580 with #3885

edave64:
This issue still seems to occur in Undertale
![image](https://user-images.githubusercontent.com/361185/85905564-03bc1e80-b80c-11ea-8482-91dc7e002127.png)

Only occurs on Vulkan.

GPU: Radeon RX Vega M GH

1900forever:
Same issue in Streets of Rages
![20201122205253](https://user-images.githubusercontent.com/3516677/99923672-cbf6cb80-2d04-11eb-85d3-aff66d2eb470.png)

dudulins:
[...]

PR #6284, opened 2021-05-06
nvflinger: Create layers when they are queried but not found

ameerj:
Fixes Shantae softlock on boot.

### DIFF
--- a/src/core/hle/service/nvflinger/nvflinger.cpp
+++ b/src/core/hle/service/nvflinger/nvflinger.cpp
@@ -139,11 +139,15 @@ std::optional<u64> NVFlinger::CreateLayer(u64 display_id) {
     }
 
     const u64 layer_id = next_layer_id++;
+    CreateLayerAtId(*display, layer_id);
+    return layer_id;
+}
+
+void NVFlinger::CreateLayerAtId(VI::Display& display, u64 layer_id) {
     const u32 buffer_queue_id = next_buffer_queue_id++;
     buffer_queues.emplace_back(
         std::make_unique<BufferQueue>(system.Kernel(), buffer_queue_id, layer_id));
-    display->CreateLayer(layer_id, *buffer_queues.back());
-    return layer_id;
+    display.CreateLayer(layer_id, *buffer_queues.back());
 }
 
 void NVFlinger::CloseLayer(u64 layer_id) {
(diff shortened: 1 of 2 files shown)


src/core/hle/service/nvflinger/nvflinger.cpp
```diff
@@ -232,6 +236,23 @@ const VI::Layer* NVFlinger::FindLayer(u64 display_id, u64 layer_id) const {
     return display->FindLayer(layer_id);
 }
 
+VI::Layer* Service::NVFlinger::NVFlinger::FindOrCreateLayer(u64 display_id, u64 layer_id) {
```
Morph1984, line 239: Since this is already within the Service::NVFlinger namespace
```suggestion
VI::Layer* NVFlinger::FindOrCreateLayer(u64 display_id, u64 layer_id) {
```

src/core/hle/service/nvflinger/nvflinger.cpp
```diff
@@ -232,6 +236,23 @@ const VI::Layer* NVFlinger::FindLayer(u64 display_id, u64 layer_id) const {
...
+    auto* const display = FindDisplay(display_id);
+
+    if (display == nullptr) {
+        return nullptr;
+    }
+
+    auto* layer = display->FindLayer(layer_id);
+
```
Shahil-Ayato, line 247: Why is there an empty Space here?

src/core/hle/service/nvflinger/nvflinger.cpp
```diff
@@ -139,11 +139,15 @@ std::optional<u64> NVFlinger::CreateLayer(u64 display_id) {
     }
 
     const u64 layer_id = next_layer_id++;
+    CreateLayerAtId(display, layer_id);
+    return layer_id;
+}
+
+void NVFlinger::CreateLayerAtId(VI::Display* display, u64 layer_id) {
```
lioncash, line 146: Perhaps use a reference like so:

```suggestion
void NVFlinger::CreateLayerAtId(VI::Display& display, u64 layer_id) {
```

Since the only way the intended behavior of this function can work is if the display isn't null.

src/core/hle/service/nvflinger/nvflinger.cpp
ameerj: To stay stylistically consistent with the rest of the functions in this file.

RinMaru:
still softlocks on Vulkan

ameerj:
> 
> 
> still softlocks on Vulkan

@RinMaru On Vulkan the game throws an exception at `BindTransformFeedbackBuffersEXT`, which is completely out of scope for this PR, and may be addressed in the later shader decompiler rewrite.

There is a [mod](https://cdn.discordapp.com/attachments/476077620350418964/838246773129609297/0100430013120000.7z) that can work around this issue on Vulkan by skipping some of the intro cutscenes, but not worth the application since on Vulkan the game suffers from the [flipped viewport issue](https://github.com/yuzu-emu/yuzu/issues/2703).

bunnei (approved):

Issue #5254, opened 2020-12-30
Fire Emblem Shadow Dragon: Renders poorly on Vulkan


lat9nq:
Nintendo's port of Fire Emblem: Shadow Dragon and the Blade of Light renders poorly on yuzu's Vulkan backend.

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5753507/yuzu_log.txt)

yuzu's Vulkan backend with RADV ACO, Mesa 20.3.1

https://user-images.githubusercontent.com/22451773/103339168-a9b44400-4a4e-11eb-9209-2467eab07c49.mp4

As a reference, yuzu's OpenGL backend with zink/RADV ACO, Mesa 21.0.0-devel (git-062b890987)
(essentially the same driver, using zink as a translator to Vulkan instead of yuzu)

https://user-images.githubusercontent.com/22451773/103339279-f39d2a00-4a4e-11eb-9ae7-98500221a866.mp4

Apologies for poor capture setup, my OBS settings seems to have gamma off by a bit.

PR #6289, opened 2021-05-08
texture_cache: Handle out of bound texture blits

ameerj:
Some games interleave a texture blit using copy regions which are out-of-bounds of the source texture. This PR aims to handle this by first copying the valid pixels, and doing a second blit pass copying the remaining pixels from the next row of the texture.

This fixes texture corruption such as those found in `Shantae` and `Werewolf Princess Kaguya`.

The following diagrams illustrate the out of bounds read in red region, and the implemented correction of the oob pixel copies in the blue region.

oob_behavior
fixed_behavior

### Before
broken

### After

fixed

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -148,7 +148,9 @@ public:
     /// Blit an image with the given parameters
     void BlitImage(const Tegra::Engines::Fermi2D::Surface& dst,
                    const Tegra::Engines::Fermi2D::Surface& src,
-                   const Tegra::Engines::Fermi2D::Config& copy);
+                   const Tegra::Engines::Fermi2D::Config& copy,
+                   std::optional<Region2D> src_region_override = {},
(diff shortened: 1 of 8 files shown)


lat9nq:
Fixes #5254
The remaining issue in #5254 is a duplicate of #2703

FernandoS27 (approved):

src/video_core/texture_cache/texture_cache.h
```diff
@@ -661,6 +687,21 @@ void TextureCache<P>::BlitImage(const Tegra::Engines::Fermi2D::Surface& dst,
         runtime.BlitImage(dst_framebuffer, dst_view, src_view, dst_region, src_region, copy.filter,
                           copy.operation);
     }
+
+    if (width_oob) {
+        // Continue copy of the oob region of the texture on the next row
+        auto oob_src = src;
+        oob_src.height++;
```
vonchenplus, line 694: Why do you need to increase the height of src here, which will cause the cache to be invalid and cause data to be read multiple times

PR #6293, opened 2021-05-09
On Linux, build SDL2 from externals with HIDAPI support

v1993:
Default to building SDL2 with yuzu on Linux unless user specifically asks otherwise and allow HIDAPI there. This allows to have working SDL motion on Linux with #6244.

I haven't touched CI, we may want to install HIDAPI there so AppImage builds include support for motion. It seems that CI is somewhat convoluted in yuzu, so I'll really prefer if someone else will tweak it.


v1993:
Note that lack of hidapi library at build time does not cause an error, so currently CI Linux build succeeds without hidapi support being built in.

v1993:
**Please do not merge this yet!**

There is a weird issue that happens when using this PR but not with manually built SDL2. Namely, yuzu hangs on start if any controllers are connected. And crashes if you disconnect them.

v1993:
Okay, this should be ready for review now.

Note: previous issue happened only when testing together with SDL motion PR. It's gone entirely now.

lat9nq:
> Note: previous issue happened only when testing together with SDL motion PR. It's gone entirely now.

Do you foresee a situation where, given when both this PR and #6244 are merged, some user building yuzu with**out** hidapi installed will run into problems? If so, it may be beneficial to require hidapi when configuring yuzu.

lat9nq:
> @lat9nq I see what you mean here. Sadly, there doesn't seem to be a clean way to check if HIDAPI was found.

That wasn't my question. Can that situation cause problems?

Also, is it not viable to check for hidapi via pkg-config?

```CMake
Include(FindPkgConfig)
pkg_check_modules(HIDAPI QUIET hidapi-hidraw>=0.10.1)
# "HIDAPI" is an arbitrary prefix that can be changed if we need.
```

After which we can check for `HIDAPI_FOUND` and set `option(HIDAPI ...)` accordingly.

v1993:
> > @lat9nq I see what you mean here. Sadly, there doesn't seem to be a clean way to check if HIDAPI was found.
> 
> That wasn't my question. Can that situation cause problems?

It would make SDL motion not to work for them, that's about it.

> Also, is it not viable to check for hidapi via pkg-config?

SDL uses a little more convoluted check for HIDAPI: https://github.com/libsdl-org/SDL/blob/107db2d89953ee7cc03417d43da1f26bd03aad57/cmake/sdlchecks.cmake#L1155. Apparently, it doesn't seem to check for hidapi library at all?..

lat9nq:
#6244 soft-depends on this PR
This PR soft-depends on yuzu-emu/build-environments#30

lat9nq (approved):

Morph1984 (approved):

PR #6266, opened 2021-05-01
Kernel Rework: Migrate kernel objects to KAutoObject

bunnei:
This a pretty big "kernel rework", migrating our kernel objects to be use [KAutoObject](https://switchbrew.org/wiki/Kernel_objects#KAutoObject) instead of shared_ptr. This makes object ownership and dependencies explicit, and much more closely matches what the real HOS kernel does. As part of these changes:
* Several of the remaining objects that did not have the K* prefix were renamed, e.g. Process -> KProcess, for consistency.
* Several objects needed to be fleshed out a bit more for things to be setup correctly, e.g. I added KPort, changed KServerPort and KClientPort a ton, and similar with KSession, KClientSession, and KServerSession.
* I implemented UnmapSharedMemory, which should fix #6233 while I was refactoring this code.
* Several other SVC implementations, etc. were improved (better error checking, etc.) while I was reviewing these implementations.

PR #6299, opened 2021-05-10
Various improvements to IPC and session management

bunnei:
* Adds initial implementation of TIPC, the new IPC protocol introduced with 12.0.0. I've tested the behavior of this with homebrew, although I am not aware of this being used in any retail games yet.
* Updates SM service to support TIPC. This also involved fixing several issues with SM error-checking, such as validating service names being broken.
* Corrects the behavior of svcConnectToNamedPort -- instead of creating static "installed" interfaces of named ports, we now register a factory method with the kernel, and create a new interface each time this is called. This was noticed in testing homebrew with TIPC, where SM is created twice (once for classic IPC, once for TIPC), and these interfaces should have separate state.
* Implements IPC "Close" command. Previously, we never actually closed sessions. This resulted in a small leak which was never noticed, because we did not enforce maximum number of sessions. Now that we do (introduced with #6266), this resulted in crashes in several games (Nintendo Labo, Pokemon Sword/Shield, probably more).
* Similarly, we now "move" objects when requested out of the IPC, rather than copy. This ensures our sessions will be destroyed when the caller closes their handle. This also contributes to fixing this leak.
* In implementing TIPC, I also took the opportunity to greatly simplify some IPC code that has not been touched in years. Should improve performance and memory usage a little bit.
* Increases SM point buffer size, various services expect this to be larger than 0x1000.
* We slightly improve DuplicateSession implementation, to now handle the changes to move objects. In a follow-up change, I will make this properly create a new session, rather than reusing the existing one.

### DIFF
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -108,6 +108,7 @@ public:
             header.type.Assign(ctx.GetCommandType());
         }
 
+        ctx.data_size = static_cast<u32>(raw_data_size);
         header.data_size.Assign(static_cast<u32>(raw_data_size));
         if (num_handles_to_copy || num_handles_to_move) {
             header.enable_handle_descriptor.Assign(1);
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -150,8 +150,8 @@ public:
         if (context->Session()->IsDomain()) {
             context->AddDomainObject(std::move(iface));
         } else {
-            kernel.CurrentProcess()->GetResourceLimit()->Reserve(
-                Kernel::LimitableResource::Sessions, 1);
+            // kernel.CurrentProcess()->GetResourceLimit()->Reserve(
+            //    Kernel::LimitableResource::Sessions, 1);
 
             auto* session = Kernel::KSession::Create(kernel);
             session->Initialize(nullptr, iface->GetServiceName());
--- a/src/core/hle/ipc_helpers.h
+++ b/src/core/hle/ipc_helpers.h
@@ -26,7 +26,7 @@ class RequestHelperBase {
 protected:
     Kernel::HLERequestContext* context = nullptr;
     u32* cmdbuf;
-    ptrdiff_t index = 0;
+    u32 index = 0;
 
 public:
     explicit RequestHelperBase(u32* command_buffer) : cmdbuf(command_buffer) {}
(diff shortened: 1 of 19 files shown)


src/core/hle/kernel/k_client_port.cpp
```diff
@@ -58,9 +58,9 @@ bool KClientPort::IsSignaled() const {
 
 ResultCode KClientPort::CreateSession(KClientSession** out) {
     // Reserve a new session from the resource limit.
-    KScopedResourceReservation session_reservation(kernel.CurrentProcess()->GetResourceLimit(),
-                                                   LimitableResource::Sessions);
-    R_UNLESS(session_reservation.Succeeded(), ResultLimitReached);
+    // KScopedResourceReservation session_reservation(kernel.CurrentProcess()->GetResourceLimit(),
```
lioncash, line 61: Ditto

src/core/hle/kernel/kernel.h
```diff
@@ -27,6 +27,10 @@ class CoreTiming;
 struct EventType;
 } // namespace Core::Timing
 
+namespace Service::SM {
```
lioncash, line 30: Should the kernel be depending on the service code for a type?

src/core/hle/ipc_helpers.h
```diff
@@ -51,11 +53,11 @@ class RequestHelperBase {
     }
 
     u32 GetCurrentOffset() const {
-        return static_cast<u32>(index);
```
lioncash, line 54: Similarly, in AlignWithPadding() above, the cast can be removed.

src/core/hle/ipc_helpers.h
```diff
@@ -111,29 +119,40 @@ class ResponseBuilder : public RequestHelperBase {
...
+        ctx.data_payload_offset = index;
+        ctx.domain_offset = index + raw_data_size / 4;
     }
 
     template <class T>
     void PushIpcInterface(std::shared_ptr<T> iface) {
         if (context->Session()->IsDomain()) {
             context->AddDomainObject(std::move(iface));
         } else {
+            // kernel.CurrentProcess()->GetResourceLimit()->Reserve(
```
lioncash, line 153: Should there be a comment explaining why this is commented out?

src/core/hle/kernel/kernel.h
bunnei: Yes, I agree this is not right. However, there isn't really a good stop-gap solution for this right now. I'll remove this dependency with my next project (rewriting how we handle dispatching service calls)

PR #6309, opened 2021-05-13
Brazilian translation

Felipefpl:
I read the text about transifex but the truth is that i cant fix the translation there, i sent a PM on transifex explaining the reason, i hope you'll understand why i posting this translation here,


FearlessTobi:
This is still not the proper way to translate the program.
I accepted all the outstanding requests on Transifex, so if you applied you should be able to translate.
In the future, please refrain from opening any other PRs like this.
Thank you

Felipefpl:
> 
> 
> This is still not the proper way to translate the program.
> I accepted all the outstanding requests on Transifex, so if you applied you should be able to translate.
> In the future, please refrain from opening any other PRs like this.
> Thank you

The problem is: i CANT translate because of the reviewed status changed by the reviewer. You can upload the file but it wont be applied over the previous one because of this.

PR #6312, opened 2021-05-15
input_common: Rewrite sdl analog mapping and fix controller disconnection crash

german77:
Old code had a flaw where it assigned the controller of the last queued event instead of the controller that triggered the threshold. Use this opportunity to rewrite this code to make it simpler and easier to read.

This PR also fixed a common crashing issue when a controller disconnect event is triggered and It wasn't on the controller list.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -267,7 +267,9 @@ void SDLState::CloseJoystick(SDL_Joystick* sdl_joystick) {
                                               return joystick->GetSDLJoystick() == sdl_joystick;
                                           });
 
-    (*joystick_it)->SetSDLJoystick(nullptr, nullptr);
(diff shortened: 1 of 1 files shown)


src/input_common/sdl/sdl_impl.cpp
```diff
@@ -267,7 +267,9 @@ void SDLState::CloseJoystick(SDL_Joystick* sdl_joystick) {
                                               return joystick->GetSDLJoystick() == sdl_joystick;
                                           });
 
-    (*joystick_it)->SetSDLJoystick(nullptr, nullptr);
+    if (joystick_it != std::end(joystick_guid_list)) {
```
Morph1984, line 270:
```suggestion
    if (joystick_it != joystick_guid_list.end()) {
```

bunnei (approved):

PR #6313, opened 2021-05-15
gl_texture_cache: Fix GL_INVALID_OPERATION on compressed formats with dimensions < 4x4

ameerj:
Per the documentation of the [BPTC](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_compression_bptc.txt), [S3TC](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_compression_s3tc.txt), and [RGTC](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_compression_rgtc.txt) formats, `GL_INVALID_OPERATION` will be generated if either of the width and height dimensions are not a multiple of four when used in `glCompressedTextureSubImage`. 

In yuzu, this error is mainly generated on small levels of mip-mapped textures, where texture sizes can reach 2x2 and 1x1. This condition will catch these cases, reducing the noisy errors generated as a result.

### DIFF
--- a/src/video_core/renderer_opengl/gl_texture_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_texture_cache.cpp
@@ -824,6 +824,13 @@ void Image::CopyBufferToImage(const VideoCommon::BufferImageCopy& copy, size_t b
     const bool is_compressed = gl_format == GL_NONE;
     const void* const offset = reinterpret_cast<const void*>(copy.buffer_offset + buffer_offset);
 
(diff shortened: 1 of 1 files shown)


bunnei:
Untagging, seems like this might be causing a Pokemon regression

image

bunnei:
Confirmed this was the issue with manual revert:

image

ameerj:
Closing as this will require a more proper fix to resolve OGL errors without regressing any games.

PR #6317, opened 2021-05-15
perf_stats: Rework FPS counter to be more accurate

ameerj:
The FPS counter was based on metrics in the nvdisp swapbuffers call. This metric would be accurate if the gpu thread/renderer were synchronous with the nvdisp service, but that's no longer the case.

This commit moves the frame counting responsibility onto the concrete renderers after their frame draw calls. Resulting in more meaningful metrics.

The displayed FPS is now made up of the average framerate between the previous and most recent update, in order to avoid distracting FPS counter updates when framerate is oscillating between close values.

The status bar update frequency was also changed from 2 seconds to 500ms to allow for more frequent and up-tp-date information to be displayed.

### DIFF
--- a/src/core/perf_stats.cpp
+++ b/src/core/perf_stats.cpp
@@ -69,9 +69,7 @@ void PerfStats::EndSystemFrame() {
 }
 
 void PerfStats::EndGameFrame() {
-    std::lock_guard lock{object_mutex};
-
-    game_frames += 1;
+    game_frames.fetch_add(1, std::memory_order_relaxed);
 }
 
 double PerfStats::GetMeanFrametime() const {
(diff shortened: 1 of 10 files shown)


src/core/perf_stats.cpp
```diff
@@ -69,9 +69,7 @@ void PerfStats::EndSystemFrame() {
 }
 
 void PerfStats::EndGameFrame() {
-    std::lock_guard lock{object_mutex};
-
-    game_frames += 1;
+    ++game_frames;
```
lioncash, line 72: If only the atomicity of the operation is desirable (likely, since this is just a counter), this can be:

```cpp
game_frames.fetch_add(1, std::memory_order_relaxed);
```

src/core/perf_stats.cpp
```diff
@@ -108,7 +106,8 @@ PerfStatsResults PerfStats::GetAndResetStats(microseconds current_system_time_us
     reset_point_system_us = current_system_time_us;
     accumulated_frametime = Clock::duration::zero();
     system_frames = 0;
-    game_frames = 0;
+    game_frames.store(0);
```
lioncash, line 109: ditto, wrt `memory_order_relaxed`

src/core/perf_stats.cpp
```diff
@@ -94,10 +92,10 @@ PerfStatsResults PerfStats::GetAndResetStats(microseconds current_system_time_us
     const auto interval = duration_cast<DoubleSecs>(now - reset_point).count();
 
     const auto system_us_per_second = (current_system_time_us - reset_point_system_us) / interval;
-
+    const auto current_fps = static_cast<double>(game_frames.load()) / interval;
```
lioncash, line 95: ditto, wrt `memory_order_relaxed`

jduncanator:
:+1: Thanks for following up on this @ameerj

ameerj:
> 
> 
> 👍 Thanks for following up on this @ameerj

👍 My pleasure @jduncanator ! Granted I have no clue what I've followed up on in this context.

bunnei (approved):

PR #6318, opened 2021-05-16
input_common: Add dual joycon support

german77:
Adds a dual joycon support by merging two individual joycon that share the same port together.
Corrects mapping to use the proper SL/SR buttons of joycons.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -18,16 +18,6 @@
 #include <utility>
 #include <vector>
 
-// Ignore -Wimplicit-fallthrough due to https://github.com/libsdl-org/SDL/issues/4307
-#ifdef __clang__
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wimplicit-fallthrough"
-#endif
-#include <SDL.h>
-#ifdef __clang__
-#pragma clang diagnostic pop
-#endif
-
 #include "common/logging/log.h"
 #include "common/math_util.h"
 #include "common/param_package.h"
@@ -212,6 +202,40 @@ public:
     void SetSDLJoystick(SDL_Joystick* joystick, SDL_GameController* controller) {
         sdl_joystick.reset(joystick);
         sdl_controller.reset(controller);
+    }
+
+    bool IsJoyconLeft() const {
+        return std::strstr(GetControllerName().c_str(), "Joy-Con Left") != nullptr;
+    }
+
+    bool IsJoyconRight() const {
+        return std::strstr(GetControllerName().c_str(), "Joy-Con Right") != nullptr;
+    }
+
+    std::string GetControllerName() const {
+        if (sdl_controller) {
+            switch (SDL_GameControllerGetType(sdl_controller.get())) {
+            case SDL_CONTROLLER_TYPE_XBOX360:
+                return "XBox 360 Controller";
+            case SDL_CONTROLLER_TYPE_XBOXONE:
+                return "XBox One Controller";
+            default:
+                break;
+            }
+            const auto name = SDL_GameControllerName(sdl_controller.get());
+            if (name) {
+                return name;
+            }
+        }
+
+        if (sdl_joystick) {
+            const auto name = SDL_JoystickName(sdl_joystick.get());
+            if (name) {
+                return name;
+            }
+        }
+
+        return "Unknown";
     }
 
 private:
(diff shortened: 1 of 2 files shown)


src/input_common/sdl/sdl_impl.cpp
```diff
@@ -214,6 +214,22 @@ class SDLJoystick {
         sdl_controller.reset(controller);
     }
 
+    bool IsJoyconLeft() {
```
lioncash, line 217:
```suggestion
    bool IsJoyconLeft() const {
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -214,6 +214,22 @@ class SDLJoystick {
...
+        if (!sdl_controller) {
+            return false;
+        }
+        const std::string name = SDL_GameControllerName(sdl_controller.get());
+        return name.find("Joy-Con Left") != std::string::npos;
+    }
+
+    bool IsJoyconRight() {
```
lioncash, line 225:
```suggestion
    bool IsJoyconRight() const {
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -214,6 +214,22 @@ class SDLJoystick {
         sdl_controller.reset(controller);
     }
 
+    bool IsJoyconLeft() {
+        if (!sdl_controller) {
+            return false;
+        }
+        const std::string name = SDL_GameControllerName(sdl_controller.get());
```
lioncash, line 221: `SDL_GameControllerName` is able to return null

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -214,6 +214,22 @@ class SDLJoystick {
...
+        return name.find("Joy-Con Left") != std::string::npos;
```
lioncash, line 222: Maybe use `std::strstr(name, "Joy-Con Left") != nullptr;` so that this doesn't need to heap allocate a string off the return value of SDL_GameControllerName() for every call to this function.

Ditto for `IsJoyconRight`

src/input_common/sdl/sdl_impl.h
```diff
@@ -66,8 +82,26 @@ class SDLState : public State {
     /// Needs to be called before SDL_QuitSubSystem.
     void CloseJoysticks();
 
-    /// Returns a custom name for specific controllers because the default name is not correct
-    std::string GetControllerName(SDL_GameController* controller) const;
+    /// Returns the default button bindings list for generic controllers
+    const ButtonBindings GetDefaultButtonBinding() const;
```
lioncash, line 86:
```suggestion
    ButtonBindings GetDefaultButtonBinding() const;
```
Given this just returns by value. Ditto for the other functions below

src/input_common/sdl/sdl_impl.h
```diff
@@ -9,6 +9,17 @@
 #include <mutex>
 #include <thread>
 #include <unordered_map>
+
+// Ignore -Wimplicit-fallthrough due to https://github.com/libsdl-org/SDL/issues/4307
+#ifdef __clang__
```
lioncash, line 14:
```suggestion
#ifdef __GNUC__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wimplicit-fallthrouh"
#endif
```

Will make this apply to both GCC and clang, since clang supports the GCC qualifier in pragmas

Ditto for the closing ifdef

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -214,6 +204,40 @@ class SDLJoystick {
...
 
+    bool IsJoyconLeft() const {
+        return std::strstr(GetControllerName().c_str(), "Joy-Con Left") != nullptr;
+    }
+
+    bool IsJoyconRight() const {
+        return std::strstr(GetControllerName().c_str(), "Joy-Con Right") != nullptr;
+    }
+
+    const std::string GetControllerName() const {
```
lioncash, line 215:
```suggestion
    std::string GetControllerName() const {
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -856,23 +880,42 @@ SDLState::~SDLState() {
...
+            std::string name =
+                fmt::format("{} {}", joystick->GetControllerName(), joystick->GetPort());
+            devices.emplace_back(Common::ParamPackage{
+                {"class", "sdl"},
+                {"display", std::move(name)},
+                {"guid", joystick->GetGUID()},
+                {"port", std::to_string(joystick->GetPort())},
+            });
+            if (joystick->IsJoyconLeft()) {
+                joycon_pairs[joystick->GetPort()] = joystick;
```
lioncash, line 898:
```suggestion
                joycon_pairs.insert_or_assign(joystick->GetPort(), joystick);
```

src/input_common/sdl/sdl_impl.h
```diff
@@ -9,6 +9,17 @@
 #include <mutex>
 #include <thread>
 #include <unordered_map>
+
+// Ignore -Wimplicit-fallthrough due to https://github.com/libsdl-org/SDL/issues/4307
+#ifdef __GNUC__
+#pragma clang diagnostic push
```
lioncash, line 15:
```suggestion
#pragma GCC diagnostic push
```

Ditto for the other pragmas. So that this works for both GCC and clang.

This include can also probably be kept in the cpp file, by just making the class member functions and types that use it internal to the cpp file. 

e.g. Doing a quick glance over the added private functions, I couldn't see any place where the functions depend on private class state, so they can probably just be moved into an anonymous namespace. If there is class state, then the class can pass it to those functions at their respective call sites.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -1131,11 +1220,68 @@ ButtonMapping SDLState::GetButtonMappingForDevice(const Common::ParamPackage& pa
     return mapping;
 }
 
+ButtonMapping SDLState::GetDualControllerMapping(const std::shared_ptr<SDLJoystick>& joystick,
+                                                 const std::shared_ptr<SDLJoystick>& joystick2,
+                                                 ButtonBindings switch_to_sdl_button,
```
lioncash, line 1225:
```suggestion
                                                 const ButtonBindings& switch_to_sdl_button,
```

Ditto for ZButtonBindings below and also with the same parameters in GetSingleControllerMapping()

v1993:
Automatic mapping is much appreciated (especially for motion), but this is behaving pretty oddly in my tests if I try to map another (capture, for example, since it's not mapped by default) button:

1. I select "Nintendo Dual Joy-Con 0" in dropdown. Everything is mapped as expected and responds with exception of "Capture", which is "[not set]".
2. I try to map "Capture". No input is detected and it remains umapped.
3. Once yuzu hits timeout and gives up on mapping it, device dropdown switches to "Any".

It looks like merged device doesn't produce/redirect any input usable for mapping, which is counter-intuitive.

P.S.: now that SDL motion is very much alive, it would make sense to add "swap motion sources" button for dual joycon setup. Example usecase: HW: AoC (once it works) players may want to use right joycon for motion instead of left one.

german77:
I noticed the mapping problem where the emulator can't detect input correctly until it swaps to any. I'm trying to understand the logic behind this. For swapping motion I would have to think about it.

bunnei (approved):

PR #6319, opened 2021-05-16
main: Prevent installing base titles into NAND

Morph1984:
Many users have been installing their base titles into NAND instead of adding them into the games list. This prevents users from installing any base titles and warns the user about the action.

Note: The warning message is not finalized yet.

### DIFF
--- a/src/core/file_sys/registered_cache.cpp
+++ b/src/core/file_sys/registered_cache.cpp
@@ -12,6 +12,7 @@
 #include "common/logging/log.h"
 #include "core/crypto/key_manager.h"
 #include "core/file_sys/card_image.h"
+#include "core/file_sys/common_funcs.h"
 #include "core/file_sys/content_archive.h"
 #include "core/file_sys/nca_metadata.h"
 #include "core/file_sys/registered_cache.h"
@@ -592,6 +593,12 @@ InstallResult RegisteredCache::InstallEntry(const NSP& nsp, bool overwrite_if_ex
     const CNMT cnmt(cnmt_file);
 
     const auto title_id = cnmt.GetTitleID();
+    const auto version = cnmt.GetTitleVersion();
+
+    if (title_id == GetBaseTitleID(title_id) && version == 0) {
+        return InstallResult::ErrorBaseInstall;
+    }
+
     const auto result = RemoveExistingEntry(title_id);
 
     // Install Metadata File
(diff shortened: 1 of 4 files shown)


src/yuzu/main.cpp
```diff
@@ -2146,13 +2148,24 @@ void GMainWindow::OnMenuInstallToNAND() {
...
 
         --remaining;
     }
 
     install_progress->close();
 
+    if (detected_base_install) {
+        QMessageBox::warning(
+            this, tr("Install Results"),
+            tr("To avoid possible conflicts, we discourage users to install base games to NAND.\n"
```
lioncash, line 2165:
```suggestion
            tr("To avoid possible conflicts, we discourage users from installing base games to the NAND.\n"
```

src/core/file_sys/registered_cache.cpp
```diff
@@ -592,6 +593,12 @@ InstallResult RegisteredCache::InstallEntry(const NSP& nsp, bool overwrite_if_ex
     const CNMT cnmt(cnmt_file);
 
     const auto title_id = cnmt.GetTitleID();
+    const auto version = cnmt.GetTitleVersion();
+
+    if ((title_id == GetBaseTitleID(title_id)) && version == 0) {
```
lioncash, line 598:
```suggestion
    if (title_id == GetBaseTitleID(title_id) && version == 0) {
```

Jatoxo:
Maybe mention the preferred way of adding base games? If users have been installing games to nand, it might not be clear enough from the UI

bunnei (approved):

bakugo:
What's the reason for this? I'm pretty sure a real switch can install games to NAND.

Morph1984:
@bakugo Indeed it can, but the current implementation presents usability issues that are difficult to fix as of now

bakugo:
>usability issues

Like what?

Being forced to install certain types of content and keep other types of content in their original format in a separate folder is pretty awkward imo, a warning is understandable if there are issues but I've personally never faced any and would like to at least have the option to keep installing games.

Morph1984:
There are a lot of instances where users are installing their games to nand, and then reporting that they have X or Y issue which was resolved by launching from the games list

Morph1984:
This feature will return when the underlying bugs have been fixed, for now it will stay disabled until then.

PR #6323, opened 2021-05-16
Various improvements to IPC and session management (Part 2)

bunnei:
This is a continuation of #6299. With that change, we finally got around to implementing the IPC's mechanisms for moving and closing objects.  This was a follow-up to #6266, which migrated kernel objects from std::shared_ptr to KAutoObject, and enforced the real limits for kernel objects.
However, these changes highlighted a longstanding KSession leak, due to limits being enforced and sessions not being properly managed. #6299 temporarily reverted session limits to avoid a crash.
This PR aims to fix these leaks once and for all, and reverts the hacks from #6299. The primary issue was implementing CloneCurrentObject, rather than just trying to fudge reference counts.

This should fix Pokemon Sword/Shield leaks once and for all.

### DIFF
--- a/src/core/hle/kernel/hle_ipc.h
+++ b/src/core/hle/kernel/hle_ipc.h
@@ -11,7 +11,6 @@
 #include <string>
 #include <type_traits>
 #include <vector>
-#include <boost/container/small_vector.hpp>
 
 #include "common/assert.h"
 #include "common/common_types.h"
@@ -285,23 +284,23 @@ public:
     bool CanWriteBuffer(std::size_t buffer_index = 0) const;
 
     Handle GetCopyHandle(std::size_t index) const {
-        return copy_handles.at(index);
+        return incoming_copy_handles.at(index);
     }
 
     Handle GetMoveHandle(std::size_t index) const {
-        return move_handles.at(index);
+        return incoming_move_handles.at(index);
     }
 
     void AddMoveObject(KAutoObject* object) {
-        move_objects.emplace_back(object);
+        outgoing_move_objects.emplace_back(object);
     }
 
     void AddCopyObject(KAutoObject* object) {
-        copy_objects.emplace_back(object);
+        outgoing_copy_objects.emplace_back(object);
     }
 
     void AddDomainObject(SessionRequestHandlerPtr object) {
-        domain_objects.emplace_back(std::move(object));
+        outgoing_domain_objects.emplace_back(std::move(object));
     }
 
     template <typename T>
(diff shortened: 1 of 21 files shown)


src/core/hle/kernel/hle_ipc.h
```diff
@@ -84,6 +86,69 @@ class SessionRequestHandler : public std::enable_shared_from_this<SessionRequest
...
+/**
+ * Manages the underlying HLE requests for a session, and whether (or not) the session should be
+ * treated as a domain. This is managed separately from server sessions, as this state is shared
+ * when objects are cloned.
+ */
+class SessionRequestManager final {
+public:
+    SessionRequestManager() = default;
+
+    const bool IsDomain() {
```
lioncash, line 100:
```suggestion
    bool IsDomain() const {
```

src/core/hle/kernel/hle_ipc.h
```diff
@@ -84,6 +86,69 @@ class SessionRequestHandler : public std::enable_shared_from_this<SessionRequest
...
+        domain_handlers = {session_handler};
+        is_domain = true;
+    }
+
+    size_t DomainHandlerCount() const {
+        return domain_handlers.size();
+    }
+
+    bool HasSessionHandler() const {
+        return !!session_handler;
```
lioncash, line 114:
```suggestion
        return session_handler != nullptr;
```

Is more explicit here.

src/core/hle/kernel/hle_ipc.h
```diff
@@ -84,6 +86,69 @@ class SessionRequestHandler : public std::enable_shared_from_this<SessionRequest
...
+
+    void CloseDomainHandler(size_t index) {
+        if (index < DomainHandlerCount()) {
+            domain_handlers[index] = nullptr;
+        } else {
+            UNREACHABLE_MSG("Unexpected handler index {}", index);
+        }
+    }
+
+    SessionRequestHandlerPtr DomainHandler(std::size_t index) const {
```
lioncash, line 133:
```suggestion
    SessionRequestHandlerPtr DomainHandler(size_t index) const {
```

src/core/hle/kernel/k_server_port.h
```diff
@@ -32,26 +32,24 @@ class KServerPort final : public KSynchronizationObject {
...
 
-    using HLEHandler = std::shared_ptr<SessionRequestHandler>;
-
     void Initialize(KPort* parent_, std::string&& name_);
 
     /// Whether or not this server port has an HLE handler available.
-    bool HasHLEHandler() const {
-        return hle_handler != nullptr;
+    bool HasSessionRequestHandler() const {
+        return !!session_handler;
```
lioncash, line 39:
```suggestion
        return session_handler != nullptr;
```

src/core/hle/kernel/k_server_session.h
```diff
@@ -82,27 +83,37 @@ class KServerSession final : public KSynchronizationObject,
...
+        return manager;
+    }
+
+    /// Gets the session request manager, which forwards requests to the underlying service
+    const std::shared_ptr<SessionRequestManager>& GetSessionRequestManager() const {
+        return manager;
+    }
+
+    /// Sets the session request manager, which forwards requests to the underlying service
+    void SetSessionRequestManager(const std::shared_ptr<SessionRequestManager>& manager_) {
```
lioncash, line 113: Taking this by value would allow calling code to move into the function and also allow moving into the member variable.

src/core/hle/kernel/hle_ipc.h
bunnei: derp, thanks

PR #6346, opened 2021-05-21
yuzu qt: Add an Apply button to configuration dialogs

lat9nq:
Most of the code already exists to do this, but the Apply button itself was never added. This adds a button and boolean that tells yuzu to save the configuration after applying settings, even if close/Cancel is pressed on the dialog. Changes after applying will not be saved when Cancel is pressed, though.

The Apply button will only appear on the Configuration and Game Properties dialog if a game is running. This can be changed if people think it's too confusing.

### DIFF
--- a/src/yuzu/configuration/configure_dialog.h
+++ b/src/yuzu/configuration/configure_dialog.h
@@ -35,8 +35,9 @@ signals:
 
 private:
     void changeEvent(QEvent* event) override;
-
     void RetranslateUI();
+
+    void HandleApplyButtonClicked();
 
     void SetConfiguration();
     void UpdateVisibleTabs();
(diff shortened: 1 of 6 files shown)


src/yuzu/configuration/configure_dialog.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <memory>
+#include <QAbstractButton>
```
lat9nq, line 8: Unnecessary include needs removed

src/yuzu/configuration/configure_per_game.h
```diff
@@ -7,6 +7,7 @@
 #include <memory>
 #include <vector>
 
+#include <QAbstractButton>
```
lat9nq, line 10: "

ameerj (approved):

PR #6348, opened 2021-05-21
CMakeLists: Update zstd to 1.5.0

lioncash:
zstd 1.5.0 brings numerous performance improvements to the library, as well as faster decompression speed. Should result in a tiny boost to shader cache compression/decompression.

More info can be seen [here](https://github.com/facebook/zstd/releases/tag/v1.5.0)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,7 +174,7 @@ macro(yuzu_find_packages)
         "lz4               1.8         lz4/1.9.2"
         "nlohmann_json     3.8         nlohmann_json/3.8.0"
         "ZLIB              1.2         zlib/1.2.11"
-        "zstd              1.4         zstd/1.4.8"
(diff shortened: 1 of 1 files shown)


bunnei (approved):

Tachi107:
Was this just done for performance, or is yuzu still API compatible with zstd 1.4.x? If so, I believe that the "version" field should be leaved at 1.4, since Debian-based distros still ship libzstd 1.4.8. This way, Debian/Ubuntu users building yuzu will still be able to use system's libzstd if they want to, and users who don't have zstd installed will benefit from the 1.5 performance improvements

lioncash:
I don't see a benefit in using the system zstd if it's going to offer a worse experience to the user

Tachi107:
Fair enough. Distros should update the library anyway, 1.5 was released months ago

PR #6354, opened 2021-05-24
Set: Allow setting device nickname

ogniK5377:


src/yuzu_cmd/config.cpp
```diff
@@ -359,6 +359,8 @@ void Config::ReadValues() {
         Settings::values.rng_seed.SetValue(std::nullopt);
     }
 
+    Settings::values.device_name = sdl2_config->GetString("System", "device_name", "Yuzu");
```
lioncash, line 362: Shouldn't this sanitize to 128 characters, since it has no UI to limit its length?

src/yuzu_cmd/config.cpp
```diff
@@ -359,6 +359,11 @@ void Config::ReadValues() {
...
+    if (Settings::values.device_name.length() > 128) {
```
lioncash, line 363: Can't this be collapsed to:

```cpp
Settings::values.device_name.resize(std::min(Settings::values.device_name.size(), size_t{128}));

bunnei:
@ogniK5377 Can you address the comments here?

liushuyu:
This now requires a rebase and some minor fix-ups.

liamwhite (approved):

PR #6355, opened 2021-05-24
fsp-srv: Implement DisableAutoSaveDataCreation

ogniK5377:
Used by Mii Edit


lioncash:
If there's a game that relies on this, it should probably be added to the commit message description.

bunnei:
> If there's a game that relies on this, it should probably be added to the commit message description.

Addressed on check-in

PR #6356, opened 2021-05-24
hid: ApplyNpadSystemCommonPolicy

ogniK5377:
We already do this specifically for homebrew, so we can keep it stubbed out for the time being

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -1770,7 +1770,7 @@ public:
             {232, nullptr, "GetIrSensorState"},
             {233, nullptr, "GetXcdHandleForNpadWithIrSensor"},
             {301, nullptr, "ActivateNpadSystem"},
-            {303, nullptr, "ApplyNpadSystemCommonPolicy"},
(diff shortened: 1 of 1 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -1949,6 +1949,15 @@ class HidSys final : public ServiceFramework<HidSys> {
 
         RegisterHandlers(functions);
     }
+
+private:
+    void ApplyNpadSystemCommonPolicy(Kernel::HLERequestContext& ctx) {
+        // We already do this for homebrew so we can just stub it out
+        LOG_WARNING(Service_HID, "called");
```
lioncash, line 1956:
```suggestion
        LOG_WARNING(Service_HID, "(STUBBED) called");
```

Or change it to a debug log if it's not significant that it's stubbed.

ogniK5377: I think keeping it stubbed for now as this should be changed in the future with a controller rework. The only reason we have the system policy setup is because older homebrew when launched directly relies on the system to already have this set

bunnei (approved):

PR #6361, opened 2021-05-26
yuzu qt: Handle per-game configs for title id 0

lat9nq:
Currently with programs that have a 0 title id, yuzu loads the custom configuration `0000000000000000.ini` for per-game configs. This is not ideal since many homebrews share this id. Instead for these programs, we load a config that is simply the file name and `.ini` appended to it.

This also changes a few things to use a `std::string_view` instead of a `const std::string&`.

### DIFF
--- a/src/yuzu/configuration/configure_per_game.cpp
+++ b/src/yuzu/configuration/configure_per_game.cpp
@@ -4,6 +4,7 @@
 
 #include <algorithm>
 #include <memory>
+#include <string>
 #include <utility>
 
 #include <QCheckBox>
@@ -14,6 +15,7 @@
 #include <QTimer>
 #include <QTreeView>
 
+#include "common/fs/path_util.h"
 #include "core/core.h"
 #include "core/file_sys/control_metadata.h"
 #include "core/file_sys/patch_manager.h"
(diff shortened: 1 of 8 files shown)


src/yuzu/configuration/configure_per_game.cpp
```diff
@@ -26,10 +28,15 @@
 #include "yuzu/uisettings.h"
 #include "yuzu/util/util.h"
 
-ConfigurePerGame::ConfigurePerGame(QWidget* parent, u64 title_id)
+ConfigurePerGame::ConfigurePerGame(QWidget* parent, u64 title_id, std::string_view file_name)
     : QDialog(parent), ui(std::make_unique<Ui::ConfigurePerGame>()), title_id(title_id) {
-    game_config = std::make_unique<Config>(fmt::format("{:016X}", title_id),
-                                           Config::ConfigType::PerGameConfig);
+    if (title_id == 0) {
```
lioncash, line 33: Perhaps this should be modified to be like:

```cpp
const auto title_filename = title_id == 0 ? Common::FS::GetFilename(file_name)
                                          : fmt::format("{:016X}", title_id);
game_config = std::make_unique<Config>(title_filename, Config::ConfigType::PerGameConfig);
```

So only the part that's dependent on the condition is managed by it.

src/yuzu/main.cpp
```diff
@@ -1334,7 +1334,13 @@ void GMainWindow::BootGame(const QString& filename, std::size_t program_index) {
 
     if (!(loader == nullptr || loader->ReadProgramId(title_id) != Loader::ResultStatus::Success)) {
         // Load per game settings
-        Config per_game_config(fmt::format("{:016X}", title_id), Config::ConfigType::PerGameConfig);
+        if (title_id == 0) {
```
lioncash, line 1337: Ditto here

bunnei (approved):

PR #6362, opened 2021-05-26
yuzu qt: Add settings reset button to general configuration

lat9nq:
Builds on @german77's work to reset all settings back to their defaults. This include UISettings and Settings values structs, but does not affect save profiles (not managed by Settings/UISettings), input presets (we simply don't delete them), and game directories (explicitly saved).

This works from a button input in `configure_general`. When activated, it calls a callback to close the whole configure dialog, then `GMainWindow::OnConfigure()` performs the reset. It also resets the UI and calls the telemetry window prompt.

What is marked as german77's work may be a continuation of #1605, but I haven't asked.

### DIFF
--- a/src/yuzu/configuration/configure_dialog.cpp
+++ b/src/yuzu/configuration/configure_dialog.cpp
@@ -16,6 +16,10 @@
 #include "yuzu/configuration/configure_input_player.h"
 #include "yuzu/hotkeys.h"
 
+static void CloseDialog(ConfigureDialog *dialog) {
+    dialog->close();
+}
+
 ConfigureDialog::ConfigureDialog(QWidget* parent, HotkeyRegistry& registry,
                                  InputCommon::InputSubsystem* input_subsystem)
     : QDialog(parent), ui(new Ui::ConfigureDialog), registry(registry) {
@@ -26,6 +30,8 @@ ConfigureDialog::ConfigureDialog(QWidget* parent, HotkeyRegistry& registry,
     setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
 
     ui->inputTab->Initialize(input_subsystem);
+
+    ui->generalTab->SetResetCallback(&CloseDialog, this);
 
     SetConfiguration();
     PopulateSelectionList();
--- a/src/yuzu/configuration/configure_dialog.cpp
+++ b/src/yuzu/configuration/configure_dialog.cpp
@@ -16,10 +16,6 @@
 #include "yuzu/configuration/configure_input_player.h"
 #include "yuzu/hotkeys.h"
 
-static void CloseDialog(ConfigureDialog *dialog) {
-    dialog->close();
-}
-
 ConfigureDialog::ConfigureDialog(QWidget* parent, HotkeyRegistry& registry,
                                  InputCommon::InputSubsystem* input_subsystem)
     : QDialog(parent), ui(new Ui::ConfigureDialog), registry(registry) {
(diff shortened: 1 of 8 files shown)


src/yuzu/configuration/configure_general.h
```diff
@@ -34,6 +38,9 @@ class ConfigureGeneral : public QWidget {
 
     void SetupPerGameUI();
 
+    void (*ResetCallback)(ConfigureDialog*);
```
lioncash, line 41:
```suggestion
    std::function<void(ConfigureDialog*)> reset_callback;
```

Ditto with `SetResetCallback()`, which would allow lambda functions with captures to work with the interface.

Perhaps also make an alias like:

```cpp
using ResetCallbackFn = std::function<void(ConfigureDialog*)>;
```

So the verbose-ish part only needs to be specified once.

src/yuzu/main.cpp
```diff
@@ -2591,9 +2591,12 @@ void GMainWindow::OnConfigure() {
         return;
     }
 
+
```
german77, line 2613: Remove these empty spaces

src/yuzu/configuration/configure_dialog.cpp
```diff
@@ -24,6 +28,8 @@ ConfigureDialog::ConfigureDialog(QWidget* parent, HotkeyRegistry& registry,
 
     ui->inputTab->Initialize(input_subsystem);
 
+    ui->generalTab->SetResetCallback(&CloseDialog, this);
```
lioncash, line 31: With the suggested std::function changes, this can be changed into a lambda capture.

src/yuzu/configuration/configure_general.cpp
```diff
@@ -49,6 +56,26 @@ void ConfigureGeneral::SetConfiguration() {
...
+    reset_callback_param = param;
+}
+
+void ConfigureGeneral::ResetDefaults() {
+    QMessageBox::StandardButton answer = QMessageBox::question(
+        this, tr("yuzu"),
+        tr("This reset all settings and remove all per-game configurations. This will not delete "
+           "game directories, profiles, or input profiles. Proceed?"),
+        QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
+    if (answer == QMessageBox::No)
```
lioncash, line 72: This should be braced

src/yuzu/configuration/configure_general.cpp
```diff
@@ -49,6 +56,26 @@ void ConfigureGeneral::SetConfiguration() {
     }
 }
 
+// Called to set the callback when resetting settings to defaults
+void ConfigureGeneral::SetResetCallback(void (*callback)(ConfigureDialog*),
+                                        ConfigureDialog* param) {
+    ResetCallback = callback;
```
lioncash, line 62: If changed to a std::function, `callback` can be std::moved.

src/yuzu/main.cpp
```diff
@@ -2587,13 +2587,48 @@ void GMainWindow::OnConfigure() {
...
+        // Explicitly save the game directories, since reinitializing config does not do so.
+        QVector<UISettings::GameDir> old_game_dirs = UISettings::values.game_dirs;
+        QVector<u64> old_favorited_ids = UISettings::values.favorited_ids;
+
+        Settings::values.disabled_addons.clear();
+
+        config = std::make_unique<Config>();
+        UISettings::values.reset_to_defaults = false;
+
+        UISettings::values.game_dirs = old_game_dirs;
```
lioncash, line 2617:
```suggestion
        UISettings::values.game_dirs = std::move(old_game_dirs);
```

Ditto for `old_favorited_ids`.

german77:
Might be worth it to rebase against #1605 since most of my code was from there

lat9nq:
> Might be worth it to rebase against #1605 since most of my code was from there

I just set the author of that commit to @FearlessTobi and added you as a co-author.

bunnei (approved):

PR #6371, opened 2021-05-27
core/arm_interface: Call SVC after end of dynarmic block.

degasus:
So we can modify all of dynarmic states within SVC without ExceptionalExit.

Especially as the ExceptionalExit hack is dropped on upstream dynarmic.

### DIFF
--- a/src/core/arm/arm_interface.h
+++ b/src/core/arm/arm_interface.h
@@ -65,9 +65,6 @@ public:
     /// Step CPU by one instruction
     virtual void Step() = 0;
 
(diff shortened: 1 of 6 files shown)


MerryMage (approved):

degasus:
Merry: May I assume that your approveal confirms that dynarmic returns directly after the SVC call in this way? I was not able to verify this in the dynarmic code.

MerryMage:
A SVC instruction emits a `CheckHalt` IR instruction, so yes.

PR #6373, opened 2021-05-28
hle: kernel: KSlabHeap: Allow host or guest allocations.

bunnei:
- Use host allocations for kernel memory, as this is not properly emulated yet.
- Use guest allocations for TLS, as this needs to be backed by DeviceMemory.

### DIFF
--- a/src/core/hle/kernel/k_slab_heap.h
+++ b/src/core/hle/kernel/k_slab_heap.h
@@ -4,34 +4,213 @@
 
 #pragma once
 
+#include <atomic>
+
+#include "common/assert.h"
+#include "common/common_types.h"
+
 namespace Kernel {
 
 class KernelCore;
 
-/// This is a placeholder class to manage slab heaps for kernel objects. For now, we just allocate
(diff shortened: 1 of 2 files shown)


src/core/hle/kernel/k_slab_heap.h
```diff
@@ -4,34 +4,192 @@
...
+    uintptr_t start{};
+    uintptr_t end{};
+};
 
 template <typename T>
-class KSlabHeap final : NonCopyable {
+class KSlabHeap final : public KSlabHeapBase {
 public:
-    KSlabHeap() = default;
+    constexpr KSlabHeap(bool use_host_allocator_ = true)
```
lioncash, line 140:
```suggestion
    explicit constexpr KSlabHeap(bool use_host_allocator_ = true)
```

lioncash: Perhaps replace the bool with an enum like:

```cpp
enum class AllocationType {
    Host,
    Guest,
};
```

So the comment suffixing isn't necessary for clarity at usage points. It's communicated to the reader inherently by the type itself.

PR #6378, opened 2021-05-28
texture_cache: Implement a very basic GC

Kelebek1:
This PR assigns the current timestamp to an image each time it's used in the cache, and then periodically checks all images, removing any which go above their timestamp + a constant expiration time. So if an image was seen at time 0, then the next GC check after 0 + 3 minutes, it will be removed, if it wasn't refreshed in that time to extend its lifetime.

This is in no way meant to be permanent at all because it's really bad, I just wanted to try and make something placeholder while Hades/a proper and more comprehensive solution is made. This clears out images from the texture cache based on a simple time since last use, which alone works for OpenGL. For Vulkan, it also removes empty allocations based on a time metric too, as they're not always effectively recycled. RAM usage isn't affected by this, I'm not sure where those allocations are coming from. If anyone has any ideas on that let me know.

Here's a 8x sped-up video of the effect in the most extreme situation I could find. Just moving between these 2 areas eats up 12GB VRAM in 4 minutes flat. Other games I've tried don't allocate nearly this much, but I wanted to show the effect clearly: https://dl.dropbox.com/s/90o9i1zxmrk6dg0/texture_cache.mp4

This should prevent almost all VRAM crashing for now I think. But again, this isn't a good solution and something much more robust would be better. I just wanted to get *something* in as some games (like the one shown) are hard to deal with when crashing every 20-30 minutes during normal play. 

I tried to make this small and not touch too many files to make it easy to remove later, but I'm not sure if you guys want to accept this as-is though since it's not great, so feel free to reject it. :)

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -97,6 +97,7 @@ bool IsFastmemEnabled() {
         return values.cpuopt_fastmem;
     }
     return true;
+}
 
 bool UseGarbageCollect() {
     return values.use_garbage_collect.GetValue();
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -52,6 +52,8 @@ void LogSettings() {
     log_setting("Renderer_FrameLimit", values.frame_limit.GetValue());
     log_setting("Renderer_UseDiskShaderCache", values.use_disk_shader_cache.GetValue());
     log_setting("Renderer_GPUAccuracyLevel", values.gpu_accuracy.GetValue());
+    log_setting("Renderer_UseGarbageCollect", values.use_garbage_collect.GetValue());
+    log_setting("Renderer_GarbageCollectLevel", values.garbage_collect_level.GetValue());
     log_setting("Renderer_UseAsynchronousGpuEmulation",
                 values.use_asynchronous_gpu_emulation.GetValue());
     log_setting("Renderer_UseNvdecEmulation", values.use_nvdec_emulation.GetValue());
@@ -95,6 +97,24 @@ bool IsFastmemEnabled() {
         return values.cpuopt_fastmem;
     }
     return true;
+
+bool UseGarbageCollect() {
+    return values.use_garbage_collect.GetValue();
+}
+
+std::chrono::minutes GarbageCollectTimer() {
+    switch (values.garbage_collect_level.GetValue()) {
+    case Settings::GCLevel::Aggressive:
+        return std::chrono::minutes(1);
+    case Settings::GCLevel::Normal:
+        return std::chrono::minutes(4);
+    case Settings::GCLevel::Relaxed:
+        return std::chrono::minutes(10);
+    }
+    UNREACHABLE_MSG("Garbage collection set to unknown value!",
+                    static_cast<int>(values.garbage_collect_level.GetValue()));
+    values.use_garbage_collect.SetValue(false);
+    return {};
 }
 
 float Volume() {
@@ -133,6 +153,8 @@ void RestoreGlobalState(bool is_powered_on) {
     values.frame_limit.SetGlobal(true);
     values.use_disk_shader_cache.SetGlobal(true);
     values.gpu_accuracy.SetGlobal(true);
+    values.use_garbage_collect.SetGlobal(true);
+    values.garbage_collect_level.SetGlobal(true);
     values.use_asynchronous_gpu_emulation.SetGlobal(true);
     values.use_nvdec_emulation.SetGlobal(true);
     values.use_vsync.SetGlobal(true);
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -13,6 +13,7 @@
 #include <span>
 #include <type_traits>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -22,6 +23,7 @@
 #include "common/common_funcs.h"
 #include "common/common_types.h"
 #include "common/logging/log.h"
+#include "common/settings.h"
 #include "video_core/compatible_formats.h"
 #include "video_core/delayed_destruction_ring.h"
 #include "video_core/dirty_flags.h"
@@ -57,11 +59,22 @@ using VideoCore::Surface::PixelFormat;
 using VideoCore::Surface::PixelFormatFromDepthFormat;
 using VideoCore::Surface::PixelFormatFromRenderTargetFormat;
 using VideoCore::Surface::SurfaceType;
+using Clock = std::chrono::steady_clock;
 
 template <class P>
 class TextureCache {
     /// Address shift for caching images into a hash table
     static constexpr u64 PAGE_BITS = 20;
+
+    /// Time between checking for expired images
+    static constexpr std::chrono::seconds GC_TICK_TIME = std::chrono::seconds(10);
+
+    /// Number of past cache sizes to keep
+    static constexpr size_t NUM_CACHE_HISTORY = 12;
+    /// Number of past cache sizes to keep
+    static constexpr size_t CACHE_REMOVAL_MIN_MB = 16;
+    /// Number of past cache sizes to keep
+    static constexpr size_t CACHE_REMOVAL_MAX_MB = 1024;
 
     /// Enables debugging features to the texture cache
     static constexpr bool ENABLE_VALIDATION = P::ENABLE_VALIDATION;
@@ -308,6 +321,12 @@ private:
     /// Returns true if the current clear parameters clear the whole image of a given image view
     [[nodiscard]] bool IsFullClear(ImageViewId id);
 
+    /// Tick the garbage collector to free up unused images
+    void TickGC();
+
+    /// Update image expiration times from current framebuffers
+    void UpdateFramebufferReferences();
+
     Runtime& runtime;
     VideoCore::RasterizerInterface& rasterizer;
     Tegra::Engines::Maxwell3D& maxwell3d;
@@ -340,6 +359,8 @@ private:
     SlotVector<Sampler> slot_samplers;
     SlotVector<Framebuffer> slot_framebuffers;
 
+    std::unordered_set<ImageViewId> active_framebuffer_imageviews;
+
     // TODO: This data structure is not optimal and it should be reworked
     std::vector<ImageId> uncommitted_downloads;
     std::queue<std::vector<ImageId>> committed_downloads;
@@ -353,6 +374,19 @@ private:
 
     u64 modification_tick = 0;
     u64 frame_tick = 0;
+
+    const bool GC_ENABLED;
+    /// Time between checking for expired images
+    const std::chrono::minutes GC_EXPIRATION_TIME;
+
+    std::chrono::time_point<std::chrono::steady_clock> GC_TIMER;
+    std::chrono::time_point<std::chrono::steady_clock> GC_FRAMEBUFF_TIMER;
+    size_t gc_ticks;
+    std::chrono::time_point<std::chrono::steady_clock> current_time;
+    size_t current_cache_size_bytes = 0;
+    size_t current_cache_removal_size_mb = CACHE_REMOVAL_MIN_MB;
+    std::array<s64, NUM_CACHE_HISTORY> cache_size_history_mb;
+    size_t current_cache_index = 0;
 };
 
 template <class P>
@@ -361,7 +395,10 @@ TextureCache<P>::TextureCache(Runtime& runtime_, VideoCore::RasterizerInterface&
                               Tegra::Engines::KeplerCompute& kepler_compute_,
                               Tegra::MemoryManager& gpu_memory_)
     : runtime{runtime_}, rasterizer{rasterizer_}, maxwell3d{maxwell3d_},
-      kepler_compute{kepler_compute_}, gpu_memory{gpu_memory_} {
+      kepler_compute{kepler_compute_}, gpu_memory{gpu_memory_},
+      GC_ENABLED{Settings::UseGarbageCollect()},
+      GC_EXPIRATION_TIME{Settings::GarbageCollectTimer()}, GC_TIMER{Clock::now()},
+      GC_FRAMEBUFF_TIMER{GC_TIMER}, current_time{GC_TIMER} {
     // Configure null sampler
     TSCEntry sampler_descriptor{};
     sampler_descriptor.min_filter.Assign(Tegra::Texture::TextureFilter::Linear);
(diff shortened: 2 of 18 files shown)


BSoDGamingYT:
Can you explain how (In practice) this works. Is it a timed cleaning of texture data?

In my first test of Xenoblade Chronicles 2, this seems to add a graphical issue where the ground around the player character is in a persistent shadow. 

PR Build


Latest Master (How it normally renders)
Capture


For what its worth, this does seem to help this games stability considerably, as it seems to periodically reduce committed memory and dedicated VRAM usage

lioncash:
Thanks for doing this =). One minor nitpick: the means of how the GC works should probably be included in the commit message, since it adds a little more information on top of "Basic GC"

lat9nq:
Causes a regression in BOTW

This PR
Screenshot_2021-05-28_17-42-26

Early Access
Screenshot_2021-05-28_17-45-33

FernandoS27:
May I ask you, if we were to tag this, how would you like it to be tag, on early access (limited but exclusive tests) or mainline access(everyone but no pretest). This is a bit important because we want mainline to be stable but then also EA is patron only and only team members get benefits (constant contributors are offered to become team members).

Kelebek1:
I really don't mind, whatever you think is best is fine with me.

Kelebek1:
@BSoDGamingYT  @lat9nq could you try again with this new commit?

BSoDGamingYT:
> @BSoDGamingYT @lat9nq could you try again with this new commit?

Sure, will do!

BSoDGamingYT:
This new commit seems to solve the issues with XC2, however the memory cleaning seems to cause a considerable slowdown/pause in game play. This is likely due to it cleaning up 15-20gbs of committed memory and 4-5gbs of VRAM. Perhaps smaller or more periodic cleaning would be a better solution?

Kelebek1:
> This new commit seems to solve the issues with XC2, however the memory cleaning seems to cause a considerable slowdown/pause in game play. This is likely due to it cleaning up 15-20gbs of committed memory and 4-5gbs of VRAM. Perhaps smaller or more periodic cleaning would be a better solution?

Is XC2 really commiting that much memory every 4 minutes...? :/ I haven't seen that happen with any other game at all.

BSoDGamingYT:
> > This new commit seems to solve the issues with XC2, however the memory cleaning seems to cause a considerable slowdown/pause in game play. This is likely due to it cleaning up 15-20gbs of committed memory and 4-5gbs of VRAM. Perhaps smaller or more periodic cleaning would be a better solution?
> 
> Is XC2 really commiting that much memory every 4 minutes...? :/ I haven't seen that happen with any other game at all. What does the console say in terms of numbers of images removed and the total cache size?

Yes, it can commit about 20gbs worth of memory in about 2 minutes.

Kelebek1:
> > > This new commit seems to solve the issues with XC2, however the memory cleaning seems to cause a considerable slowdown/pause in game play. This is likely due to it cleaning up 15-20gbs of committed memory and 4-5gbs of VRAM. Perhaps smaller or more periodic cleaning would be a better solution?
> > 
> > 
> > Is XC2 really commiting that much memory every 4 minutes...? :/ I haven't seen that happen with any other game at all. What does the console say in terms of numbers of images removed and the total cache size?
> 
> Yes, it can commit about 20gbs worth of memory in about 2 minutes. This game is one of the worst in regards memory leaks. Both in respect to committed memory and VRAM
> 
> A quick example of how bad this game leaks
> 
> https://streamable.com/0ktf41

Alright I made a few performance improvements, and also added some scaling ramp-up to the deletions which should help the hitching a lot. It works every 10 seconds now rather than over minutes, and it now only deletes X MB per tick, and if it caps out, then the next tick will try to remove more, and so on, scaling up and down over time.

I'm not sure how much it'll help with XC2 though judging by that video. I don't want to set the expiry time too low, because then you're going to spend more time recompiling pipelines and shaders and re-loading the images back again, which causes hitching and glitchy frames while things load. I don't want to tune the entire thing based on the worst offenders, so I'm keeping it at 4 minutes for now, so you're essentially always playing with the last ~4 minutes of the game's memory before it will start to drop out. I think that's pretty fair, because most games could go 15+ minutes easily.

The only things I can think of to help this further would be to expose some GC settings in the itnerface so you can tune how aggressive it is, or just a more clever/better implementation which I can't make. I don't know how to add stuff to the frontend, and I wouldn't want to, because again I only made this as a stop-gap, not a permanent solution to anything, so I don't think it should be integrated into the UI.

Let me know how XC2 performs again though, especially the performance side of things.

src/video_core/vulkan_common/vulkan_memory_allocator.cpp
```diff
@@ -311,6 +330,20 @@ std::optional<u32> MemoryAllocator::FindType(VkMemoryPropertyFlags flags, u32 ty
     return std::nullopt;
 }
 
+void MemoryAllocator::TickFrame() {
+    const auto now{std::chrono::steady_clock::now()};
+    if (now - gc_timer >= ALLOCATION_TICK) {
```
lioncash, line 335:
```suggestion
    if (now - gc_timer < ALLOCATION_TICK) {
        return;
    }
    
    Would allow unindenting most of the code here a level.
```

src/video_core/vulkan_common/vulkan_memory_allocator.cpp
```diff
@@ -311,6 +330,20 @@ std::optional<u32> MemoryAllocator::FindType(VkMemoryPropertyFlags flags, u32 ty
...
+        for (s64 x = allocations.size() - 1; x > 0; --x) {
```
lioncash, line 336:
```suggestion
        for (s64 x = static_cast<s64>(allocations.size()) - 1; x > 0; --x) {
```

Otherwise this is performing an implicit sign conversion

src/video_core/texture_cache/texture_cache.h
```diff
@@ -373,14 +405,90 @@ TextureCache<P>::TextureCache(Runtime& runtime_, VideoCore::RasterizerInterface&
...
 template <class P>
 void TextureCache<P>::TickFrame() {
+    current_time = std::chrono::steady_clock::now();
+
     // Tick sentenced resources in this order to ensure they are destroyed in the right order
     sentenced_images.Tick();
     sentenced_framebuffers.Tick();
     sentenced_image_view.Tick();
+
+    if (current_time - gc_framebuffer_timer >= GC_UPDATE_FRAMEBUFFER_REFS) {
```
lioncash, line 421: Given this is quite large now, maybe move this code block into its own function, since its entirely related to the GC.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -540,6 +648,16 @@ ImageViewId TextureCache<P>::VisitImageView(DescriptorTable<TICEntry>& table,
 template <class P>
 FramebufferId TextureCache<P>::GetFramebufferId(const RenderTargets& key) {
     const auto [pair, is_new] = framebuffers.try_emplace(key);
+
+    std::ranges::for_each(pair->first.color_buffer_ids, [this](ImageViewId id) {
```
lioncash, line 652: Why the use of for_each, rather than a ranged for loop?

src/video_core/texture_cache/texture_cache.h
```diff
@@ -1492,4 +1619,30 @@ bool TextureCache<P>::IsFullClear(ImageViewId id) {
            scissor.max_y >= size.height;
 }
 
+template <class P>
+void TextureCache<P>::UpdateFramebufferReferences() {
+    for (u32 i = 0; i < slot_images.Size(); ++i) {
+        const SlotId id{i};
+        if (slot_images.IsIndexFree(id)) {
```
lioncash, line 1626:
```suggestion
        if (!slot_images.IsIndexFree(id)) {
            continue;
        }
```

Would allow unindenting most of the code a little more.

src/video_core/vulkan_common/vulkan_memory_allocator.h
```diff
@@ -100,6 +106,8 @@ class MemoryAllocator {
     /// Commits memory required by the image and binds it.
     MemoryCommit Commit(const vk::Image& image, MemoryUsage usage);
 
+    void TickFrame();
```
lioncash, line 109: Missing a documentation comment like the rest of the API

src/video_core/texture_cache/texture_cache.h
```diff
@@ -373,14 +405,90 @@ TextureCache<P>::TextureCache(Runtime& runtime_, VideoCore::RasterizerInterface&
     // This way the null resource becomes a compile time constant
     void(slot_image_views.insert(runtime, NullImageParams{}));
     void(slot_samplers.insert(runtime, sampler_descriptor));
+
+    std::ranges::fill(cache_size_history_mb, CACHE_REMOVAL_MAX_MB / 2);
```
lioncash, line 409:
```suggestion
    cache_size_history_mb.fill(CACHE_REMOVAL_MAX_MB / 2);
```

src/video_core/texture_cache/texture_cache.h
```diff
@@ -373,14 +405,90 @@ TextureCache<P>::TextureCache(Runtime& runtime_, VideoCore::RasterizerInterface&
...
+            }
+        }
+
+        if (num_removed > 0) {
+            LOG_INFO(HW_Memory, "Removed {} images ({}MB) from texture cache. GC removal cap: {}MB",
+                     num_removed, removed_mb, current_cache_removal_size_mb);
+        }
+
+        size_t current_average_cache_size_mb =
+            std::accumulate(cache_size_history_mb.begin(), cache_size_history_mb.end(), (size_t)0);
```
lioncash, line 456:
```suggestion
            std::accumulate(cache_size_history_mb.begin(), cache_size_history_mb.end(), size_t{0});
```

src/video_core/texture_cache/texture_cache.h
```diff
@@ -373,14 +405,90 @@ TextureCache<P>::TextureCache(Runtime& runtime_, VideoCore::RasterizerInterface&
...
+        current_average_cache_size_mb /= NUM_CACHE_HISTORY * 2;
+        f32 removal_sizef = static_cast<f32>(current_cache_removal_size_mb);
+        f32 current_averagef = static_cast<f32>(current_average_cache_size_mb);
+
+        if (gc_capped) {
+            // GC has capped out, so increase the cache removal
+            // for the next tick, up to CACHE_REMOVAL_MAX_MB
+            f32 ratio =
+                std::max<float>(0.0f, (1 + (current_averagef / removal_sizef)) / NUM_CACHE_HISTORY);
```
lioncash, line 465: The explicit type specifier on max and min below shouldn't be necessary, since both arguments are the same type.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -1492,4 +1619,30 @@ bool TextureCache<P>::IsFullClear(ImageViewId id) {
...
+                });
+
+            if (in_use) {
+                // Update the times for the aliases and views for this image,
+                // as they may not be directly referenced otherwise (i.e pause menus)
+                image.last_access_time = current_time;
+                std::ranges::for_each(image.image_view_ids, [this](ImageViewId& view) {
+                    slot_images[slot_image_views[view].image_id].last_access_time = current_time;
+                });
+                for (auto& alias : image.aliased_images) {
```
lioncash, line 1640:
```suggestion
                for (const auto& alias : image.aliased_images) {
```

Since the alias is only being queried

lat9nq:
> BSoDGamingYT lat9nq could you try again with this new commit?

Appears to fix the issue I had with RadeonSI in BOTW.

gidoly:
seems there is no slowdown on xc2 torna golden country with last commit

BSoDGamingYT:
I tested XC2 for another 5 hours and while it didnt crash for me, I can see it crashing for other users due to being out of memory. I feel like it could trigger cleaning memory far more frequently than it is currently doing.

Kelebek1:
Just for you BSoD, I cut the time in half, so things expire twice as quickly. Please try other games though, because this will cause more allocations in total, and more reloading of shaders/images.

kqwyf:
Is this PR related to GPU memory? I can not build this PR on yuzu's master branch, and I tried to clone Kelebek1:Cache branch and built it. I tried Xenoblade 2 and after several minutes it took all of my GPU memory and then broke. Things are the same when I use the official mainline build.

I'm using Intel i5-9300H, 16GB RAM and NVIDIA GeForce GTX 1660 Ti with 6GB GPU memory.

kqwyf:
I'm using NVIDIA GeForce GTX 1660 Ti with 6GB GPU RAM. I just tested Xenoblade 2, and it crashed by OOM when I quickly go across a big map. Then I change GC_IMAGE_EXPIRATION to 1 minute and the game works well. The GPU memory occupied is basically 3~4GB.
To me, it's quite a nice work. And I totally agree that the GC parameters should not be tuned by the worst occasion. Maybe it's a good idea for someone to add some option on the UI like the "Accuracy Level", which provides several pre-defined GC settings for the user, and could be set per game. This option is transparent to the GC strategy, which means that the GC implementation could be further optimized or replaced, while the option persists.

Kelebek1:
@Morph1984 I've added a setting to the UI, since that seems to be the reasonable way to deal with games needing very different settings. Could you check that adding the option is even something the Yuzu devs want, and if I've implemented it correctly? Graphics advanced may be the wrong page for it, I'm not sure. If you want it moved or anything let me know.

src/core/core.cpp
```diff
@@ -154,7 +154,7 @@ struct System::Impl {
     }
 
     ResultStatus Init(System& system, Frontend::EmuWindow& emu_window) {
-        LOG_DEBUG(Core, "initialized OK");
```
Morph1984, line 157: This should be changed in another PR, if you intend on changing it.

src/yuzu/configuration/configure_general.cpp
```diff
@@ -66,7 +65,7 @@ void ConfigureGeneral::ApplyConfiguration() {
             Settings::values.frame_limit.SetValue(ui->frame_limit->value());
         }
     } else {
-        bool global_frame_limit = use_frame_limit == ConfigurationShared::CheckState::Global;
+        const bool global_frame_limit = use_frame_limit == ConfigurationShared::CheckState::Global;
```
Morph1984, line 68: ditto

src/yuzu/configuration/configure_graphics_advanced.ui
```diff
@@ -66,6 +66,42 @@
...
         </item>
+        <item>
+         <layout class="QHBoxLayout" name="garbage_collect_layout">
+          <item>
+           <widget class="QCheckBox" name="use_garbage_collect">
+            <property name="toolTip">
+             <string>Enable periodic removal of textures from memory, reducing VRAM usage. Lower settings work more often, lowering VRAM usage but potentially causes hitching. Set to a shorter duration if you experience out of memory crashes.</string>
+            </property>
+            <property name="text">
+             <string>Garbage Collect:</string>
```
Morph1984, line 77: `Garbage Collector:`

src/common/settings.cpp
```diff
@@ -90,6 +92,25 @@ bool IsGPULevelHigh() {
            values.gpu_accuracy.GetValue() == GPUAccuracy::High;
 }
 
+bool UseGarbageCollect() {
+    return !values.use_garbage_collect.GetValue();
```
Morph1984, line 96: Hmm, why does this return the negation?

Kelebek1:
Closing this in favour of #6465.

PR #6379, opened 2021-05-28
externals: Update dynarmic.

degasus:
The new version supports fastmem on a64.

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic_32.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic_32.cpp
@@ -4,9 +4,9 @@
 
 #include <cinttypes>
 #include <memory>
-#include <dynarmic/A32/a32.h>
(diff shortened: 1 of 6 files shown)


bunnei:
Untagged until we resolve cmake issue: https://gist.github.com/bunnei/652c863c1bccaed36e85647027c2993b

bunnei (approved):

bunnei:
Tested locally, seems fine. Probably just build pipeline being weird with submodules.

v1993:
Does this PR alone make yuzu utilize fastmem? Or are further additions required to get fastmem working? Asking since you mention in few places that you have done testing and noticed 15% performance increase.

degasus:
It is only the dynarmic part, yuzu does not support it so far.
And it is only for 64 bit games, 32 but games have been supported in dynarmic for a long time already (also no yuzu support)

PR #6385, opened 2021-05-29
core/memory: Check our memory fallbacks for out-of-bound behavior.

degasus:
This makes it by far harder to crash yuzu.
Also implement the 48bit masking of AARCH64 while touching this code.


MerryMage:
What happens with memory accesses that straddle end of memory?

degasus:
Merry: I'm still implementing to split read/write access which overlaps the page boundries, but I think this should be merged+tested within another PR. This is likely to cause of behavior changes.

bunnei (approved):

PR #6389, opened 2021-05-30
input_common: Analog button, use time based position

german77:
Use a different approach for emulating an analog stick from buttons.
- Removes the need for creating a thread for each joystick instance
- Fixes some crash errors on close

This also implements callbacks events on inputs to update data as soon it's available

### DIFF
--- a/src/core/frontend/input.h
+++ b/src/core/frontend/input.h
@@ -27,6 +27,10 @@ struct AnalogProperties {
     float range;
     float threshold;
 };
+template <typename StatusType>
+struct InputCallback {
+    std::function<void(StatusType)> on_change;
+};
 
 /// An abstract class template for an input device (a button, an analog input, etc.).
 template <typename StatusType>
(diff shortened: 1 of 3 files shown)


src/core/frontend/input.h
```diff
@@ -50,6 +54,15 @@ class InputDevice {
                                [[maybe_unused]] f32 freq_high) const {
         return {};
     }
+    void SetCallback(InputCallback<StatusType> callback_) {
+        callback = callback_;
```
lioncash, line 58:
```suggestion
        callback = std::move(callback_);
```

src/core/frontend/input.h
```diff
@@ -50,6 +54,15 @@ class InputDevice {
...
+    }
+    void TriggerOnChange() {
+        if (callback.on_change) {
+            callback.on_change(std::move(GetStatus()));
```
lioncash, line 62:
```suggestion
            callback.on_change(GetStatus());
```

Since the result of the call will already be an r-value here.

src/core/frontend/input.h
```diff
@@ -50,6 +54,15 @@ class InputDevice {
...
+        }
+    }
+    InputCallback<StatusType> callback;
```
lioncash, line 65: this should be private

bunnei (approved):

PR #6393, opened 2021-05-30
externals: libusb: CMakeLists: Set platform flags for Windows/POSIX events and threads.

bunnei:
- Hopefully fixes MINGW builds.


bunnei:
Not the right fix

PR #6394, opened 2021-05-31
externals: libusb: Use autotools for MinGW

lat9nq:
After updating to 1.0.24, MinGW fails to build libusb as a result of numerous errors. So we build libusb their way and let them update the nontrivial stuff.

This only applies to MinGW: the old path is still in use for Linux toolchains as well as MSVC.

This will dynamically link libusb, since I hit build errors with the old way we used to resolve the conflict with SDL2.

git diff is going to git diff, but everything after the `else() # MINGW` is identical to the previous version, sans a indentation correction.


lat9nq:
Looks like I didn't wait long enough for yuzu-emu/build-environments#31to percolate through all the containers. Still need to wait for that container to update, then I'll rebase and try again.

ameerj (approved):

PR #6397, opened 2021-06-01
common: fs: fs_util: Add more string conversion functions

Morph1984:
These string conversion functions are required for enforcing UTF-8 encoded paths in the future,

### DIFF
--- a/src/common/fs/fs_util.h
+++ b/src/common/fs/fs_util.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <concepts>
+#include <filesystem>
 #include <span>
 #include <string>
 #include <string_view>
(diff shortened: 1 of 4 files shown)


src/common/fs/fs_util.h
```diff
@@ -22,4 +26,31 @@ concept IsChar = std::same_as<T, char>;
  */
 [[nodiscard]] std::u8string ToU8String(std::string_view utf8_string);
 
+/**
+ * Converts a buffer of bytes to a UTF8-encoded std::u8string.
```
lioncash, line 30: This should probably mention that it stops at the first null terminator in the buffer

lioncash (approved):

PR #6402, opened 2021-06-02
game_list: Stop the columns resizing on NAND install

### LINKED ISSUE #6401: Install to NAND increases the size of the Name column
Basically title.
After installing updates and DLC, the name column of the game list increases size considerably. Minor annoyance.

Kelebek1:
Rather than deleteing and re-inserting columns and working with modified indexes, just hide the one(s) we don't want to show instead, and keep everything the same. Save the new layout every time it's altered through resize, move or sort.

The column sizes were being reset a lot through startup, after the original load from config. I believe they were mainly coming from the setHeaderData calls in PopulateAsync. By hiding the column instead, they can all be removed. The header is only saved on game boot and yuzu's exit, so any modifications to the layout and then using a menu call which calls PopulateAsync would reset everything, as it calls LoadInterfaceLayout() without a save beforehand. I don't think that's needed, as populate should only be filling the rows and not altering the columns.

Closes #6401

### DIFF
--- a/src/yuzu/game_list.cpp
+++ b/src/yuzu/game_list.cpp
@@ -328,18 +328,14 @@ GameList::GameList(FileSys::VirtualFilesystem vfs, FileSys::ManualContentProvide
     tree_view->setContextMenuPolicy(Qt::CustomContextMenu);
     tree_view->setStyleSheet(QStringLiteral("QTreeView{ border: none; }"));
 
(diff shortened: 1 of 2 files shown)


goldenx86:
You earned a cookie :P

Morph1984:
🍪

bunnei (approved):

PR #6403, opened 2021-06-02
game_list: Minor for loop optimizations

Kewlan:
There's no need to check the first and last rows since they'll always be the _Favorites_ and _Add Directory_ rows.
I decided to remove the `folder_count` variables because it doesn't just hold the folder count, but also the previous mentioned rows.

Also change the name of the `clear_all` variable for consistency.


adityaruplaha:
image

Nice.

bunnei (approved):

PR #6404, opened 2021-06-02
yuzu qt: Revert some usages of string_view

lat9nq:
Fixes a heap-use-after free reported by AddressSanitizer, caused by #6361. This makes use of std::filesystem::path, but due to that we have to use their `string()` function which may not work well with some characters on Windows.

Draft because I need testing.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -16,7 +16,7 @@
 
 namespace FS = Common::FS;
 
-Config::Config(std::string_view config_name, ConfigType config_type) : type(config_type) {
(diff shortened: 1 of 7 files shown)


lat9nq:
This is not ideal, but the purpose of this PR is to fix the heap-use-after-free, and the resulting configs correspond 1-to-1 to for example funny names like `(mgba) (╯°□°）╯︵ ┻━┻.nro` (handled correctly in Linux, handled poorly but still deterministic in Windows.)

After this, I want to scrap using the file name for the file hash.

Thanks to @gidoly for testing.

bunnei (approved):

PR #6407, opened 2021-06-03
cmake: Use autotools for libusb linking generally on GNU, and cleanup

lat9nq:
Building libusb was also broken on GCC (and maybe Clang) on our CMakeLists after upgrading to 1.0.24, but it was not being checked because our 18.04 container had an old version of libusb installed on it. This builds on the MinGW work from earlier and extends it to the rest of the GNU toolchains. In addition we make use of pkg-config when present to find libusb. pkg-config is preferrable because we can specify a minimum required version (1.0.24).

After that, there's a bunch cleanup for libusb to make it similar to how we add other externals.


CMakeLists.txt
```diff
@@ -420,14 +422,22 @@ elseif (TARGET Boost::boost)
...
+    endif()
+
+    if (NOT LIBUSB_FOUND)
+        message(WARNING "libusb not found, falling back to externals")
+        set(YUZU_USE_BUNDLED_LIBUSB ON)
+    else()
+        add_library(usb INTERFACE)
+        target_include_directories(usb INTERFACE "${LIBUSB_INCLUDE_DIRS}")
+        target_link_libraries(usb INTERFACE "${LIBUSB_LIBRARIES}")
+    endif()
```
lat9nq, line 440: Re-order these

lat9nq:
Marking as draft: this PR forces the linux-fresh container to build libusb dynamically, but libusb will not be handled by AppImage tools. The best solution would be to upgrade libusb in the container, so this is required before merging.

lat9nq:
Turns out that we can override `GUID_DEVINTERFACE_USB_DEVICE` on Linux, so static linking is possible.

bunnei (approved):

PR #6408, opened 2021-06-03
[game_list] Correct light theme loading

Kelebek1:
The setLayout call in game list instanciation will call resizing signals with default values in light theme, which was then being erroneously saved. setLayout doesn't seem to call resizing for any other theme, so I'm not sure why that happens.

### DIFF
--- a/src/yuzu/game_list.cpp
+++ b/src/yuzu/game_list.cpp
@@ -341,11 +341,7 @@ GameList::GameList(FileSys::VirtualFilesystem vfs, FileSys::ManualContentProvide
     connect(tree_view, &QTreeView::customContextMenuRequested, this, &GameList::PopupContextMenu);
     connect(tree_view, &QTreeView::expanded, this, &GameList::OnItemExpanded);
     connect(tree_view, &QTreeView::collapsed, this, &GameList::OnItemExpanded);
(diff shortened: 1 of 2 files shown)


Kewlan:
What was the point of adding the three resizing signals in the previous PR? The state already gets saved when booting a game and when exiting yuzu, so they shouldn't be needed. Removing those three connections also solves this issue.

Kelebek1:
> 
> 
> What was the point of adding the three resizing signals in the previous PR? The state already gets saved when booting a game and when exiting yuzu, so they shouldn't be needed. Removing those three connections also solves this issue.

Good point. I added them originally because PopulateAsync removed and re-created the columns, but then I ended up removing that so the connects aren't needed anymore. Thanks for pointing that out.

Kewlan (approved):

lioncash:
Looks good.

In the future consider using `git commit --amend` to amend the previously made commit instead of creating a new one when the problem being solved is the same thing (just a thing to keep the git history a little cleaner).

Consider also putting the description directly inside the commit as well, so that the information gets saved into the git repository as well (provides the "why" or explanation of a change directly to the reader if `git blame` is ever used or if someone looks up context to a change, even if they happen to not have internet access at the time).

lioncash (approved):

PR #6414, opened 2021-06-05
hle: kernel: Refactor to allocate a ServiceThread per service handler.

bunnei:
- Previously, we would allocate a thread per session, which adds new threads on CloneCurrentObject.
- This results in race conditions with N sessions queuing requests to the same service interface.
- Fixes Pokken Tournament DX crashes/softlocks, which were regressed by #6347.

### DIFF
--- a/src/core/hle/kernel/hle_ipc.cpp
+++ b/src/core/hle/kernel/hle_ipc.cpp
@@ -39,7 +39,7 @@ SessionRequestHandler::~SessionRequestHandler() {
 
 SessionRequestManager::SessionRequestManager(KernelCore& kernel_) : kernel{kernel_} {}
 
-SessionRequestManager::~SessionRequestManager() {}
+SessionRequestManager::~SessionRequestManager() = default;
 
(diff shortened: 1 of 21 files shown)


src/core/hle/kernel/hle_ipc.cpp
```diff
@@ -30,9 +30,16 @@
...
+    : kernel{kernel_}, service_thread{kernel.CreateServiceThread(service_name_)} {}
 
-SessionRequestHandler::~SessionRequestHandler() = default;
+SessionRequestHandler::~SessionRequestHandler() {
+    kernel.ReleaseServiceThread(service_thread);
+}
+
+SessionRequestManager::SessionRequestManager(KernelCore& kernel_) : kernel{kernel_} {}
+
+SessionRequestManager::~SessionRequestManager() {}
```
lioncash, line 42:
```suggestion
SessionRequestManager::~SessionRequestManager() = default;
```

lioncash (commented):
LGTM otherwise

Issue #6262, opened 2021-04-30
Prevent screensaver while in-game (Xorg)


Mynacol:
I enjoy playing Switch games on my computer a lot. Thanks for this very fast and pleasant experience.

One big annoyance I encountered is the screensaver kicking in during gameplay. To work-around this, I set `xset s off -dpms` before gameplay. As this is manual and does not revert back after ending gaming, I propose to disable the screensaver while in-game. Video players like mpv or vlc do the same if you want to know how this is done.

The alternative would probably be to educate my machine to not activate the screen saver if there is joystick/controller input.

What's your stance on screensavers?

Technical Details:
* Arch Linux
* Xorg / X11
* yuzu-20210426-758433fe2.AppImage
* gaming with Dualshock 4 controller (no mouse/keyboard input during gameplay)

steven-a-wilson:
This also bothered me recently, and I'm a software developer with some (embedded) Linux and Qt experience. So I should clearly fix it myself, right?

That being said, I just spent a couple hours thinking that this should be easy and essentially coming up empty. I figured that the freedesktop.org crowd would have definitively solved this a decade ago [1], but as far as I can tell it's still a garbage fire in practice. For any given application that inhibits screensaver/sleep "successfully", there appears to be a distinct mix of unreliable methods gated by a distinct mix of questionable platform/DE/window system tests. I'd love to be proven wrong on this, but as far as I can tell there is still no serious solution to this problem for Linux generally. Everyone seems to be throwing their own custom concoction at the wall and hoping that something will stick.

[1] The xdg-screensaver script allegedly solves this problem, but upon further examination its requirements seem to be incompatible with the limitations of QWidget::WinId() [2] [3]).

[2] "WindowID must be the X Window ID of an existing window of the calling application. The window must remain in existence for the duration of the suspension." [what about Wayland systems?]

[3] "This value may change at run-time. An event with type QEvent::WinIdChange will be sent to the widget following a change in window system identifier."

bibitocarlos:
You can use lightson : [https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwif8tH1ksLwAhWhdN8KHe3dBTgQrAIoA3oECAMQBA&url=https%3A%2F%2Fbbs.archlinux.org%2Fviewtopic.php%3Fid%3D145324&usg=AOvVaw0jaI0vmjuDfq2IgGepEa7u](url)

[...]

PR #6421, opened 2021-06-06
common/idle_inhibitor: Initial implementation

steven-a-wilson:
Move idle inhibition from GMainWindow to a more self-contained IdleInhibitor class, and add support for inhibiting idle on Linux via the org.freedesktop.ScreenSaver D-Bus interface (this requires libdbus-1.so.3 at runtime, which is already in the current AppImage; there is no additional build-time dependency) or by periodically running the `xdg-screensaver` script from the xdg-utils package for polling-style interfaces.

This is intended to address #6262 in typical Linux environments, and I've personally tested against several configurations: Plasma 5 X11, Plasma 5 Wayland, GNOME 40 X11, i3+DPMS, and i3+XScreenSaver. However, there isn't a true standard for doing this on Linux (only drafts and de-facto standards), so it is likely that some environments will not be covered by this implementation.


src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL
+// WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED
+// WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE
+// AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR
+// CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+// OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
+// NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+// CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+#include "idle_inhibitor.h"
```
lioncash, line 15:
```suggestion
#include "common/idle_inhibitor.h"
```

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+struct DBusConnection;
+struct DBusError;
+struct DBusMessage;
+typedef uint32_t dbus_bool_t;
+typedef uint32_t dbus_uint32_t;
+#endif
+
+namespace {
+#if defined(__unix__)
+const char* g_app = "yuzu";
```
lioncash, line 36:
```suggestion
constexpr char g_app[] = "yuzu";
```

Ditto for g_reason

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+const char* g_reason = "Emulation is running";
+#endif
+
+#if defined(__unix__)
+const int DBUS_TYPE_INVALID = '\0';
```
lioncash, line 41: These can be constexpr

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+const int DBUS_TYPE_STRING = 's';
+const int DBUS_TYPE_UINT32 = 'u';
+const int DBUS_TIMEOUT_USE_DEFAULT = -1;
+enum DBusBusType { DBUS_BUS_SESSION, DBUS_BUS_SYSTEM, DBUS_BUS_STARTER };
```
lioncash, line 45:
```suggestion
enum DBusBusType { DBUS_BUS_SESSION, DBUS_BUS_SYSTEM, DBUS_BUS_STARTER, };
```

So that clang format doesn't single line the enum.

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+#if defined(_WIN32)
+#include <windows.h>
+#elif defined(__unix__)
+#include <chrono>
+#include <dlfcn.h>
+#include <unistd.h>
+struct DBusConnection;
+struct DBusError;
+struct DBusMessage;
+typedef uint32_t dbus_bool_t;
```
lioncash, line 30:
```cpp
using dbus_bool_t = uint32_t;
```

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+typedef uint32_t dbus_uint32_t;
```
lioncash, line 31: Ditto here.

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+            if (!initialized) {
+                LOG_ERROR(Common, "Failed to load libdbus functions");
+            }
+        } else {
+            LOG_WARNING(Common, "Failed to dlopen() libdbus-1.so.3");
+        }
+    }
+
+    ~DBus() {
+        if (handle)
```
lioncash, line 89: This should be braced.

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+namespace Common {
+IdleInhibitor::IdleInhibitor() {
+#if defined(__unix__)
+    ConnectDBus();
+#endif
+}
+
+IdleInhibitor::~IdleInhibitor() {
+    if (inhibiting_idle)
```
lioncash, line 106: Ditto (and elsewhere this is applicable further down).

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+void AllowIdle() {
+    inhibiting_idle = false;
+    SetThreadExecutionState(ES_CONTINUOUS);
+}
+
+#elif defined(__unix__)
+void IdleInhibitor::ResetScreensaver() {
+    auto poll_time = std::chrono::steady_clock::now() + std::chrono::seconds(10);
+    for (;;) {
```
lioncash, line 128:
```suggestion
    while (true) {
```

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+        if (fork() == 0)
+            execlp("xdg-screensaver", "xdg-screensaver", "reset", nullptr);
+
+        poll_time += std::chrono::seconds(10);
+    }
+}
+
+void IdleInhibitor::ConnectDBus() {
+    if (g_dbus.initialized) {
```
lioncash, line 142: Inverting this into a guard clause allows the code to be unindented a level.

```suggestion
    if (!g_dbus.initialized) {
        return;
    }
```

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+        LOG_WARNING(Common, "Falling back to xdg-screensaver polling thread");
+        poll_thread = std::make_unique<std::thread>(&Common::IdleInhibitor::ResetScreensaver, this);
+    }
+}
+
+void IdleInhibitor::AllowIdle() {
+    if (!inhibiting_idle)
+        return;
+
+    cond_mutex.lock();
```
lioncash, line 197:
```suggestion
    {
    std::lock_guard lock{cond_mutex}
    inhibiting_idle = false;
    }
```

src/common/idle_inhibitor.cpp
```diff
@@ -0,0 +1,231 @@
...
+    inhibiting_idle = false;
+    cond_mutex.unlock();
+
+    if (poll_thread && poll_thread->joinable()) {
+        poll_cond.notify_all();
+        poll_thread->join();
+        poll_thread = nullptr;
```
lioncash, line 204:
```suggestion
        poll_thread.reset();
```

Makes the deallocation here a little more explicit.

src/common/idle_inhibitor.cpp
steven-a-wilson: clang-format wants this to be indented. Should I do that, or is there another way this should be addressed?

Tatsh:
This patch works, but it does not work to stop KDE's "Power Management" -> "Energy Saving" -> "Screen Energy Saving" -> "Switch off after" setting from applying.

Can you have this also call `org.freedesktop.PowerManagement.Inhibit.Inhibit` ?

Tatsh:
Actually I do not know if that will make a difference. I will have to confirm the screen turning off was definitely caused by power settings and not respecting the inhibition (which can be queried on this interface as well as the ScreenSaver one).

foresto:
What happens if yuzu crashes after a successful ScreenSaver.Inhibit() call? Is the screen saver left inhibited?

(A different approach would be to periodically call ScreenSaver.SimulateUserActivity(), which would never leave the screen saver stuck in a disabled state.)

Tatsh:
From what I can tell the cookie will remain until unInhibit is called.

steven-a-wilson:
@Tatsh 
> This patch works, but it does not work to stop KDE's "Power Management" -> "Energy Saving" -> "Screen Energy Saving" -> "Switch off after" setting from applying.
> 
> Can you have this also call `org.freedesktop.PowerManagement.Inhibit.Inhibit` ?

That setting is inhibited by `org.freedesktop.ScreenSaver.Inhibit` on my KDE system (Arch Linux, KDE Plasma 5.22.2, KDE Frameworks 5.83.0), and I don't recall seeing any other game/media code using `org.freedesktop.PowerManagement` while researching this, so it seems like this shouldn't be necessary. In other words, I wonder if something else is going on. Is it possible that another program is turning off the screen? Also, did you have emulation running? The inhibition isn't maintained when emulation is paused or stopped (per the existing behavior on Windows).

@foresto 
> What happens if yuzu crashes after a successful ScreenSaver.Inhibit() call? Is the screen saver left inhibited?

The draft for the org.freedesktop.ScreenSaver interface says "Inhibition will stop when [...] the application disconnects from the D-Bus session bus" [1], and this appears to be the case on both of my main Linux systems (the Arch/KDE system mentioned above, and another system running Fedora 34 with GNOME 40), tested via `kill -9`. This is why I implemented the D-Bus connection in yuzu rather than running another program such as `xdg-screensaver` or `dbus-send` for systems with this interface. I wouldn't be shocked if there are other implementations that do "leak" the inhibition, though.

> (A different approach would be to periodically call ScreenSaver.SimulateUserActivity(), which would never leave the screen saver stuck in a disabled state.)

This might also be preferable because it could be done via `dbus-send`, removing the need to have the D-Bus connection handled within yuzu. However, my GNOME 40 system reports that this method is not implemented. It appears that there was a similar method in GNOME 2.x under the `org.gnome.ScreenSaver` interface, but I don't see it in GNOME 40, nor any clear indication that it survived into GNOME 3.x generally.

[1] https://people.freedesktop.org/~hadess/idle-inhibition-spec/ch03.html

foresto:
> The draft for the org.freedesktop.ScreenSaver interface says "Inhibition will stop when [...] the application disconnects from the D-Bus session bus" 

That's reassuring. Thanks.

> > (A different approach would be to periodically call ScreenSaver.SimulateUserActivity(), which would never leave the screen saver stuck in a disabled state.)
> 
> This might also be preferable because it could be done via `dbus-send`, removing the need to have the D-Bus connection handled within yuzu. However, my GNOME 40 system reports that this method is not implemented.

I wonder if GNOME 40 still provides gnome-screensaver-command. For the record, here are the stateless screen wake commands that I have found to work on various desktops and screen savers:

    gnome-screensaver-command --deactivate
    mate-screensaver-command --poke
    xfce4-screensaver-command --poke
    xscreensaver-command -deactivate
    xset dpms force on
    xset s reset

steven-a-wilson:
The CLA link doesn't seem to work. I get:

Error

There is no CLA to sign for yuzu-emu/yuzu

(could not find linked item for owner yuzu-emu and repo yuzu)

liushuyu:
I think #6702 and #7614 jointly replaced this pull request, and this pull request may be closed.

Morph1984:
Superseded by the aforementioned PRs.

PR #6422, opened 2021-06-07
Implement/Port Fastmem from Citra to Yuzu

FernandoS27:
This Pull Request finally implements a long requested feature from Citra: Fastmem. Fastmem is the name we have given to emulating the MMU by using Host allocation instead of software MMU emulation. In the previous approach, we would emulate Guest virtual memory through a page table where we would store addresses to physical memory for each virtual page the application/game had. Whenever a game accessed memory, we would look into the emulated page table check if we have a valid physical mem address and then read the memory, if the physical address wasn't valid, the game would call an emulator fallback. All this costed around 10-15 instructions.

On fastmem, we use instead about 4 instructions: 3 for checking if the address is within the emulated address space (can be disabled with Unsafe CPU -> disable address space checks) and 1 instruction for memory read/write. In case of a Page Fault, dynarmic will invalidate the code block and recompile it under the first software approach. This is typical when dealing with GPU marked memory.

**Important note for windows users**: This PR only works for newest versions of Windows 10. Please make sure to update to latest version to avoid any inconveniences. (or build 1803 or later)

I do not take credit for most of this work, since I only advised and mixed everyone's efforts together. The credit goes to everyone above.

Credits:
- @MerryMage For implementing fastmem in dynarmic.
- @degasus For implementing Linux version of fastmem and original design.
- @ReinUsesLisp For Implementing Windows version of fastmem/host memory handling.
- @BreadFish64 For showing us on his gameboy emulator that placeholder memory could be aligned to 4kb, thus, avoid the biggest layback on windows MMU emulation: Windows page granularity for most part is 64kb.
- @bunnei for setting up the kernel swiftly and nicely.

### DIFF
--- a/src/common/host_memory.cpp
+++ b/src/common/host_memory.cpp
@@ -449,21 +449,52 @@ private:
     int fd{-1}; // memfd file descriptor, -1 is the error value of memfd_create
 };
 
-#else // ^^^ Linux ^^^
+#else // ^^^ Linux ^^^ vvv Generic vvv
 
-#error Please implement the host memory for your platform
+class HostMemory::Impl {
+public:
+    explicit Impl(size_t /*backing_size */, size_t /* virtual_size */) {
+        // This is just a place holder.
+        // Please implement fastmem in a propper way on your platform.
+        throw std::bad_alloc{};
+    }
 
-#endif
+    void Map(size_t virtual_offset, size_t host_offset, size_t length) {}
+
+    void Unmap(size_t virtual_offset, size_t length) {}
+
+    void Protect(size_t virtual_offset, size_t length, bool read, bool write) {}
+
+    u8* backing_base{nullptr};
+    u8* virtual_base{nullptr};
+};
+
+#endif // ^^^ Generic ^^^
 
 HostMemory::HostMemory(size_t backing_size_, size_t virtual_size_)
-    : backing_size(backing_size_),
-      virtual_size(virtual_size_), impl{std::make_unique<HostMemory::Impl>(
-                                       AlignUp(backing_size, PageAlignment),
-                                       AlignUp(virtual_size, PageAlignment) + 3 * HugePageSize)},
-      backing_base{impl->backing_base}, virtual_base{impl->virtual_base} {
-    virtual_base += 2 * HugePageSize - 1;
-    virtual_base -= reinterpret_cast<size_t>(virtual_base) & (HugePageSize - 1);
-    virtual_base_offset = virtual_base - impl->virtual_base;
+    : backing_size(backing_size_), virtual_size(virtual_size_) {
+    try {
+        // Try to allocate a fastmem arena.
+        // The implementation will fail with std::bad_alloc on errors.
+        impl = std::make_unique<HostMemory::Impl>(AlignUp(backing_size, PageAlignment),
+                                                  AlignUp(virtual_size, PageAlignment) +
+                                                      3 * HugePageSize);
+        backing_base = impl->backing_base;
+        virtual_base = impl->virtual_base;
+
+        if (virtual_base) {
+            virtual_base += 2 * HugePageSize - 1;
+            virtual_base -= reinterpret_cast<size_t>(virtual_base) & (HugePageSize - 1);
+            virtual_base_offset = virtual_base - impl->virtual_base;
+        }
+
+    } catch (const std::bad_alloc&) {
+        LOG_CRITICAL(HW_Memory,
+                     "Fastmem unavailable, falling back to VirtualBuffer for memory allocation");
+        fallback_buffer = std::make_unique<Common::VirtualBuffer<u8>>(backing_size);
+        backing_base = fallback_buffer->data();
+        virtual_base = nullptr;
+    }
 }
 
 HostMemory::~HostMemory() = default;
@@ -478,7 +509,7 @@ void HostMemory::Map(size_t virtual_offset, size_t host_offset, size_t length) {
     ASSERT(length % PageAlignment == 0);
     ASSERT(virtual_offset + length <= virtual_size);
     ASSERT(host_offset + length <= backing_size);
-    if (length == 0) {
+    if (length == 0 || !virtual_base || !impl) {
         return;
     }
     impl->Map(virtual_offset + virtual_base_offset, host_offset, length);
@@ -488,7 +519,7 @@ void HostMemory::Unmap(size_t virtual_offset, size_t length) {
     ASSERT(virtual_offset % PageAlignment == 0);
     ASSERT(length % PageAlignment == 0);
     ASSERT(virtual_offset + length <= virtual_size);
-    if (length == 0) {
+    if (length == 0 || !virtual_base || !impl) {
         return;
     }
     impl->Unmap(virtual_offset + virtual_base_offset, length);
@@ -498,7 +529,7 @@ void HostMemory::Protect(size_t virtual_offset, size_t length, bool read, bool w
     ASSERT(virtual_offset % PageAlignment == 0);
     ASSERT(length % PageAlignment == 0);
     ASSERT(virtual_offset + length <= virtual_size);
-    if (length == 0) {
+    if (length == 0 || !virtual_base || !impl) {
         return;
     }
     impl->Protect(virtual_offset + virtual_base_offset, length, read, write);
--- a/src/common/host_memory.cpp
+++ b/src/common/host_memory.cpp
@@ -1,11 +1,5 @@
-#ifdef __linux__
-#ifndef _GNU_SOURCE
-#define _GNU_SOURCE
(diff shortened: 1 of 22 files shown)


src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -143,7 +144,7 @@ std::shared_ptr<Dynarmic::A32::Jit> ARM_Dynarmic_32::MakeJit(Common::PageTable*
 
     // Code cache size
     config.code_cache_size = 512 * 1024 * 1024;
-    config.far_code_offset = 256 * 1024 * 1024;
+    config.far_code_offset = 400 * 1024 * 1024;
```
ReinUsesLisp, line 147: Probably an unintentional change.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -181,7 +185,7 @@ std::shared_ptr<Dynarmic::A64::Jit> ARM_Dynarmic_64::MakeJit(Common::PageTable*
 
     // Code cache size
     config.code_cache_size = 512 * 1024 * 1024;
-    config.far_code_offset = 256 * 1024 * 1024;
+    config.far_code_offset = 400 * 1024 * 1024;
```
ReinUsesLisp, line 188: Probably an unintentional change.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
FernandoS27: actually intentional, it was a bug. far_code offset is not suppose to be in the middle.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
FernandoS27: Ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
degasus: Also a related change: With fastmem enabled, we emit by far less code. Especially without bound checking, the far code cache does not fill up at all (bound checking is small, exiting the block however is not)

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -160,6 +160,10 @@ std::shared_ptr<Dynarmic::A64::Jit> ARM_Dynarmic_64::MakeJit(Common::PageTable*
         config.absolute_offset_page_table = true;
         config.detect_misaligned_access_via_page_table = 16 | 32 | 64 | 128;
         config.only_detect_misalignment_via_page_table_on_page_boundary = true;
+
+        config.fastmem_pointer = page_table->fastmem_arena;
+        config.fastmem_address_space_bits = address_space_bits;
```
degasus, line 165: cpuopt_unsafe_fastmem_check

src/common/host_memory.h
```diff
@@ -0,0 +1,66 @@
...
+class HostMemory {
+public:
+    explicit HostMemory(size_t backing_size_, size_t virtual_size_);
+    ~HostMemory();
+
+    /**
+     * Copy constructors. They shall return a copy of the buffer without the mappings.
+     * TODO: Implement them with COW if needed.
+     */
+    HostMemory(HostMemory& other) = delete;
```
lioncash, line 25:
```suggestion
    HostMemory(const HostMemory& other) = delete;
```

Ditto for the copy assignment operator

src/common/host_memory.cpp
```diff
@@ -0,0 +1,438 @@
...
+#include <mutex>
+
+#include "common/alignment.h"
+#include "common/assert.h"
+#include "common/host_memory.h"
+#include "common/logging/log.h"
+
+namespace Common {
+
+static constexpr size_t PageAlignment = 0x1000;
```
lioncash, line 34: Static can be removed here.

src/common/host_memory.cpp
```diff
@@ -0,0 +1,438 @@
...
+        // We don't want to unmap, we want to reserve this memory.
+
+        void* ret = mmap(virtual_base + virtual_offset, length, PROT_NONE,
+                         MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0);
+        ASSERT_MSG(ret != MAP_FAILED, "mmap failed: {}", strerror(errno));
+    }
+
+    void Protect(size_t virtual_offset, size_t length, bool read, bool write) {
+        int flags = 0;
+        if (read)
```
lioncash, line 347: These conditions should be braced

degasus:
May I ask you to autosquash the fixup commit? How is the GPUThread change related to fastmem?

otaku976:
Nice fix , but how much we need to wait for project Hades?

FernandoS27:
> May I ask you to autosquash the fixup commit? How is the GPUThread change related to fastmem?

kinda in the protection of read part, changing that forced me to up to date the other stuffs,

v1993:
I understand that this isn't a large concern for now, but it would be nice to disable fastmem if compiled for platform that doesn't support it instead of outright causing build error, given that yuzu can still work without fastmem just fine.

FernandoS27:
> I understand that this isn't a large concern for now, but it would be nice to disable fastmem if compiled for platform that doesn't support it instead of outright causing build error, given that yuzu can still work without fastmem just fine.

Thanks for the feedback. I'll see to try to address this. Could you give some examples ? We do know there are few people trying to get yuzu into MacOS and likely Android. I'de love to add this implementatipons but I fully understand the case where there's an OS which is not supported.

src/common/host_memory.cpp
```diff
@@ -0,0 +1,438 @@
...
+                               PAGE_READWRITE, SEC_COMMIT, backing_size, nullptr, nullptr, 0);
+        if (!backing_handle) {
+            throw std::bad_alloc{};
+        }
+        // Allocate a virtual memory for the backing file map as placeholder
+        backing_base = static_cast<u8*>(VirtualAlloc2(process, nullptr, backing_size,
+                                                      MEM_RESERVE | MEM_RESERVE_PLACEHOLDER,
+                                                      PAGE_NOACCESS, nullptr, 0));
+        if (!backing_base) {
+            Release();
```
v1993, line 55: Calling Release on every error seems a bit ugly to me. How about simple RAII wrapper that stores a reference to `HostMemory::Impl` and calls `Release()` on destruction unless deactivated? Something like

```
HostMemoryGuard hmguard{*this};
...
exiting here will call `Release()` from hmguard's destructor
...
On success:
hmguard.deactivate();
return;
```
This seems to be useful in Linux implementation as well since right now it also manually calls Release before throwing.

v1993 (changes requested):

src/common/host_memory.cpp
MerryMage: `SCOPE_EXIT({ Release(); });` preferable

degasus:
> I understand that this isn't a large concern for now, but it would be nice to disable fastmem if compiled for platform that doesn't support it instead of outright causing build error, given that yuzu can still work without fastmem just fine.

This is sadly not as easy. The chance to use fastmem requires a special way to allocate the memory itself. So in the end, we can only support half of the API in a generic build. However good luck, this is likely one of the first place you want to touch after porting it to a new platform, and as long as the destination is not Windows, you can just pick the linux implementation :)

src/common/host_memory.cpp
```diff
@@ -0,0 +1,538 @@
...
+        backing_base = impl->backing_base;
+        virtual_base = impl->virtual_base;
+
+        if (virtual_base) {
+            virtual_base += 2 * HugePageSize - 1;
+            virtual_base -= reinterpret_cast<size_t>(virtual_base) & (HugePageSize - 1);
+            virtual_base_offset = virtual_base - impl->virtual_base;
+        }
+
+    } catch (const std::bad_alloc& e) {
```
Morph1984, line 491:
```suggestion
    } catch (const std::bad_alloc&) {
```

src/yuzu/configuration/configure_cpu.ui
```diff
@@ -123,6 +123,18 @@
...
         </item>
+        <item>
+         <widget class="QCheckBox" name="cpuopt_unsafe_fastmem_check">
+          <property name="toolTip">
+           <string>
+            &lt;div&gt;This option improves speed by eliminating a safety check before every memory read/write in guest. Disabling it may allow a game to read/write the emulator's memory.&lt;/div&gt;
+           </string>
+          </property>
+          <property name="text">
+           <string>Disable adress space checks</string>
```
Morph1984, line 134:
```suggestion
           <string>Disable address space checks</string>
```

lioncash (approved):

PR #6436, opened 2021-06-09
src/common/CMakeLists.txt: fix variable escaping

liushuyu:
Fix the variable escaping in `src/common/CMakeLists.txt`.

According to CMake (https://cmake.org/cmake/help/latest/command/add_custom_command.html):
>  All arguments to the commands will be escaped properly for the build tool so that the invoked command receives each argument unchanged. Note that one level of escapes is still used by the CMake language processor before add_custom_command even sees the arguments. Use of VERBATIM is recommended as it enables correct behavior. When VERBATIM is not given the behavior is platform specific because there is no protection of tool-specific special characters.

This should resolve the title bar text issue on Linux.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -21,14 +21,14 @@ find_package(Git QUIET)
 
 add_custom_command(OUTPUT scm_rev.cpp
     COMMAND ${CMAKE_COMMAND}
-      -DSRC_DIR="${CMAKE_SOURCE_DIR}"
-      -DBUILD_REPOSITORY="${BUILD_REPOSITORY}"
-      -DTITLE_BAR_FORMAT_IDLE="${TITLE_BAR_FORMAT_IDLE}"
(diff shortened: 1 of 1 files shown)


src/common/CMakeLists.txt
```diff
@@ -92,6 +92,7 @@ add_custom_command(OUTPUT scm_rev.cpp
       "${CMAKE_CURRENT_SOURCE_DIR}/scm_rev.h"
       # technically we should regenerate if the git version changed, but its not worth the effort imo
       "${CMAKE_SOURCE_DIR}/CMakeModules/GenerateSCMRev.cmake"
+      VERBATIM
```
lat9nq, line 95:
```suggestion
    VERBATIM
```
Keeps tabbing consistent with the other parameters.

lat9nq (changes requested):
Just one nit

src/common/CMakeLists.txt
liushuyu: Addressed.

lioncash (approved):

PR #6439, opened 2021-06-09
ci: common: Remove 7z packaging

lat9nq:
Removes the 7z from being package during CI, as only .tar.xz preserves information needed on Linux, and otherwise is just extremely redundant to package in addition to the .tar.xz.  This affects Linux releases and PR-verify artifacts only. MSVC releases do not use this script to my knowledge.

Draft simply so I can verify that CI is correctly not packaging the .7z.


ameerj (approved):
Thank you for this!

PR #6448, opened 2021-06-11
common: fs: Use the normal directory iterator in *Recursively functions

Morph1984:
MSVC's implementation of recursive_directory_iterator throws an exception on an error despite a std::error_code being passed into its constructor. This is most likely a bug in MSVC's implementation since directory_iterator does not throw an exception on an error.

We can replace the usage of recursive_directory_iterator for now until MSVC fixes their implementation of it.

This fixes crashes when trying to access folders without sufficient permissions (such as `System Volume Information` at the root of a drive).

### DIFF
--- a/src/common/fs/fs.cpp
+++ b/src/common/fs/fs.cpp
@@ -321,7 +321,8 @@ bool RemoveDirContentsRecursively(const fs::path& path) {
 
     std::error_code ec;
 
-    for (const auto& entry : fs::recursive_directory_iterator(path, ec)) {
+    // TODO (Morph): Replace this with recursive_directory_iterator once it's fixed in MSVC.
+    for (const auto& entry : fs::directory_iterator(path, ec)) {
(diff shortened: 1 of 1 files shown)


FernandoS27 (approved):

CookiePLMonster:
Calling this a MSVC bug is not correct, it was an implementation bug (with this PR fixing the bug correctly).

`recursive_directory_iterator` appears to build the directory tree eagerly, not lazily; therefore, the code removing directories recursively creates a bug where advancing the iterator attempts to recurse to a now removed directory. While I agree this should have set an error code, a case like this does not seem to be explicitly covered in the documentation:
https://en.cppreference.com/w/cpp/filesystem/recursive_directory_iterator/increment
and therefore it probably can do whatever.

TL;DR: The fix in PR is correct but notes about a "MSVC bug" are not.

PR #6456, opened 2021-06-13
configure_cpu_debug: Clarify settings behavior

Morph1984:
This makes it clear that the disabled settings only take effect when CPU Accuracy is set to Debug Mode.


v1993:
Suggestion: how about moving these checckboxes to the same page as mode selection and showing them only on Debug accuracy, like it's already done for Unsafe mode settings? This makes it obvious that they only affect things when on Debug accuracy (and makes UI more consistent overall).

bunnei (approved):

PR #6458, opened 2021-06-13
fsp_srv: Add a small delay to OutputAccessLogToSdCard

Morph1984:
This simulates a delay when an entry is logged.

Fixes softlocks during loading in Xenoblade Chronicles 2 when certain DLC is enabled.

Thanks to @gidoly for helping me test this.


Morph1984:
Closed as this is not an appropriate fix, see #6460

PR #6462, opened 2021-06-13
common: fs: file: Flush the file to the disk when Flush() is called

Morph1984:
std::fflush does not guarantee that file buffers are flushed to the disk.

Use _commit on Windows and fsync on all other OSes to ensure that the file is flushed to the disk.

### DIFF
--- a/src/common/fs/file.cpp
+++ b/src/common/fs/file.cpp
@@ -309,7 +309,11 @@ bool IOFile::Flush() const {
 
     errno = 0;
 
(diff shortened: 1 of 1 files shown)


Kelebek1:
Microsoft docs say `fileno` is deprecated in favour of `_fileno`, was there a specific reaosn to use `fileno`?

lioncash (approved):

bunnei (approved):

PR #6465, opened 2021-06-14
GPU: Implement a garbage collector for GPU Caches (project Reaper+)

FernandoS27:
This PR instroduces a new GC for GPU Caches, it will try keep memory at between 2.5-4Gb VRAM by removing `Least Used Resources` and Texture Cache leftovers. In general, the GC is pretty safe but it does seem to have some issues in games like SMO when travelling through kingdoms may cause graphics issues (very few to notice though). My current guess is that it's due to GPU VMM Remaps not notifying TCR. This will be fixed in a further PR.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -106,6 +106,8 @@ public:
 
     void TickFrame();
 
+    void RunGarbageCollector();
+
     void WriteMemory(VAddr cpu_addr, u64 size);
 
     void CachedWriteMemory(VAddr cpu_addr, u64 size);
@@ -350,29 +352,7 @@ BufferCache<P>::BufferCache(VideoCore::RasterizerInterface& rasterizer_,
 }
 
 template <class P>
-void BufferCache<P>::TickFrame() {
-    const bool enabled_gc = Settings::values.use_caches_gc.GetValue();
-    SCOPE_EXIT({
-        ++frame_tick;
-        delayed_destruction_ring.Tick();
-    });
-    // Calculate hits and shots and move hit bits to the right
-    const u32 hits = std::reduce(uniform_cache_hits.begin(), uniform_cache_hits.end());
-    const u32 shots = std::reduce(uniform_cache_shots.begin(), uniform_cache_shots.end());
-    std::copy_n(uniform_cache_hits.begin(), uniform_cache_hits.size() - 1,
-                uniform_cache_hits.begin() + 1);
-    std::copy_n(uniform_cache_shots.begin(), uniform_cache_shots.size() - 1,
-                uniform_cache_shots.begin() + 1);
-    uniform_cache_hits[0] = 0;
-    uniform_cache_shots[0] = 0;
-
-    const bool skip_preferred = hits * 256 < shots * 251;
-    uniform_buffer_skip_cache_size = skip_preferred ? DEFAULT_SKIP_CACHE_SIZE : 0;
-
-    const bool activate_gc = enabled_gc && total_used_memory >= EXPECTED_MEMORY;
-    if (!activate_gc) {
-        return;
-    }
+void BufferCache<P>::RunGarbageCollector() {
     const bool aggressive_gc = total_used_memory >= CRITICAL_MEMORY;
     const u64 ticks_to_destroy = aggressive_gc ? 60 : 120;
     int num_iterations = aggressive_gc ? 64 : 32;
@@ -390,6 +370,28 @@ void BufferCache<P>::TickFrame() {
             DeleteBuffer(buffer_id);
         }
     }
+}
+
+template <class P>
+void BufferCache<P>::TickFrame() {
+    // Calculate hits and shots and move hit bits to the right
+    const u32 hits = std::reduce(uniform_cache_hits.begin(), uniform_cache_hits.end());
+    const u32 shots = std::reduce(uniform_cache_shots.begin(), uniform_cache_shots.end());
+    std::copy_n(uniform_cache_hits.begin(), uniform_cache_hits.size() - 1,
+                uniform_cache_hits.begin() + 1);
+    std::copy_n(uniform_cache_shots.begin(), uniform_cache_shots.size() - 1,
+                uniform_cache_shots.begin() + 1);
+    uniform_cache_hits[0] = 0;
+    uniform_cache_shots[0] = 0;
+
+    const bool skip_preferred = hits * 256 < shots * 251;
+    uniform_buffer_skip_cache_size = skip_preferred ? DEFAULT_SKIP_CACHE_SIZE : 0;
+
+    if (Settings::values.use_caches_gc.GetValue() && total_used_memory >= EXPECTED_MEMORY) {
+        RunGarbageCollector();
+    }
+    ++frame_tick;
+    delayed_destruction_ring.Tick();
 }
 
 template <class P>
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -65,6 +65,9 @@ class BufferCache {
 
     static constexpr BufferId NULL_BUFFER_ID{0};
 
+    static constexpr u64 expected_memory = 512ULL * 1024ULL * 1024ULL;
+    static constexpr u64 critical_memory = 1024ULL * 1024ULL * 1024ULL;
+
     using Maxwell = Tegra::Engines::Maxwell3D::Regs;
 
     using Runtime = typename P::Runtime;
@@ -327,6 +330,7 @@ private:
 
     typename SlotVector<Buffer>::Iterator deletion_iterator;
     u64 frame_tick = 0;
+    u64 total_used_memory = 0;
 
     std::array<BufferId, ((1ULL << 39) >> PAGE_BITS)> page_table;
 };
@@ -346,6 +350,10 @@ BufferCache<P>::BufferCache(VideoCore::RasterizerInterface& rasterizer_,
 
 template <class P>
 void BufferCache<P>::TickFrame() {
+    SCOPE_EXIT({
+      ++frame_tick;
+      delayed_destruction_ring.Tick();
+    });
     // Calculate hits and shots and move hit bits to the right
     const u32 hits = std::reduce(uniform_cache_hits.begin(), uniform_cache_hits.end());
     const u32 shots = std::reduce(uniform_cache_shots.begin(), uniform_cache_shots.end());
@@ -359,8 +367,13 @@ void BufferCache<P>::TickFrame() {
     const bool skip_preferred = hits * 256 < shots * 251;
     uniform_buffer_skip_cache_size = skip_preferred ? DEFAULT_SKIP_CACHE_SIZE : 0;
 
-    static constexpr u64 ticks_to_destroy = 120;
-    int num_iterations = 32;
+    const bool activate_gc = total_used_memory >= expected_memory;
+    if (!activate_gc) {
+        return;
+    }
+    const bool agressive_gc = total_used_memory >= critical_memory;
+    const u64 ticks_to_destroy = agressive_gc ? 60 : 120;
+    int num_iterations = agressive_gc ? 64 : 32;
     for (; num_iterations > 0; --num_iterations) {
         if (deletion_iterator == slot_buffers.end()) {
             deletion_iterator = slot_buffers.begin();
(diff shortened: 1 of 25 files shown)


src/yuzu/configuration/configure_graphics_advanced.ui
```diff
@@ -103,6 +103,16 @@
           </property>
          </widget>
         </item>
+        <item>
+         <widget class="QCheckBox" name="use_caches_gc">
+          <property name="toolTip">
+           <string>Enables garbage collection for the GPU caches, this will try to keep VRAM within 3-4Gb and flush least used textures/buffers. This option may be unsafe on a few games</string>
```
v1993, line 109:
```suggestion
           <string>Enables garbage collection for the GPU caches, this will try to keep VRAM usage within 3-4Gb by flushing least used textures/buffers. May cause issues in a few games.</string>
```

Aside from missing dot I tried to improve description a bit.

FernandoS27: Thanks for correcting my grammar. I have adressed it, let me know if there's something else.

src/yuzu/configuration/configure_graphics_advanced.ui
```diff
@@ -103,6 +103,16 @@
           </property>
          </widget>
         </item>
+        <item>
+         <widget class="QCheckBox" name="use_caches_gc">
+          <property name="toolTip">
+           <string>Enables garbage collection for the GPU caches, this will try to keep VRAM within 3-4Gb by flushing least used textures/buffers. May cause issues in a few games.</string>
```
CrusadingNinja, line 109:
```suggestion
           <string>Enables garbage collection for the GPU caches; this will try to keep VRAM within 3-4 GB by flushing the least used textures/buffers. May cause issues in a few games.</string>
```

src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -349,7 +368,27 @@ void BufferCache<P>::TickFrame() {
...
     uniform_buffer_skip_cache_size = skip_preferred ? DEFAULT_SKIP_CACHE_SIZE : 0;
 
-    delayed_destruction_ring.Tick();
+    const bool activate_gc = enabled_gc && total_used_memory >= expected_memory;
+    if (!activate_gc) {
+        return;
+    }
+    const bool agressive_gc = total_used_memory >= critical_memory;
+    const u64 ticks_to_destroy = agressive_gc ? 60 : 120;
+    int num_iterations = agressive_gc ? 64 : 32;
```
Morph1984, line 377: agressive -> aggressive

CrusadingNinja (commented):
minor grammar nit

src/yuzu/configuration/configure_graphics_advanced.ui
FernandoS27: Thanks, I have corrected it.

src/video_core/buffer_cache/buffer_cache.h
FernandoS27: Thank zou. I kind of get confused sometimes xD.

src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -65,6 +65,9 @@ class BufferCache {
 
     static constexpr BufferId NULL_BUFFER_ID{0};
 
+    static constexpr u64 expected_memory = 512ULL * 1024ULL * 1024ULL;
+    static constexpr u64 critical_memory = 1024ULL * 1024ULL * 1024ULL;
```
ameerj, line 69: `common/common_sizes.h` defines some common constants that you can use here instead.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -75,6 +76,9 @@ class TextureCache {
     /// Sampler ID for bugged sampler ids
     static constexpr SamplerId NULL_SAMPLER_ID{0};
 
+    static constexpr u64 expected_memory = 1024ULL * 1024ULL * 1024ULL;
+    static constexpr u64 critical_memory = 2 * 1024ULL * 1024ULL * 1024ULL;
```
ameerj, line 80: Ditto

src/video_core/texture_cache/slot_vector.h
```diff
@@ -32,6 +33,60 @@ template <class T>
...
+
+        T* operator->() const noexcept {
+            return std::addressof((*slot_vector)[id]);
+        }
+
+    private:
+        Iterator(SlotVector<T>* slot_vector_, SlotId id_) noexcept
+            : slot_vector{slot_vector_}, id{id_} {}
+
+        bool IsValid(const u64* bitset) noexcept {
```
lioncash, line 82:
```suggestion
        bool IsValid(const u64* bitset) const noexcept {
```

src/video_core/texture_cache/texture_cache.h
```diff
@@ -373,11 +379,82 @@ TextureCache<P>::TextureCache(Runtime& runtime_, VideoCore::RasterizerInterface&
...
+    const bool enabled_gc = Settings::values.use_caches_gc.GetValue();
+    if (!enabled_gc) {
+        // @Note(Blinkhawk): compile error with SCOPE_EXIT on msvc.
+        sentenced_images.Tick();
+        sentenced_framebuffers.Tick();
+        sentenced_image_view.Tick();
+        ++frame_tick;
+        return;
+    }
+    const bool high_priority_mode = total_used_memory >= expected_memory;
```
lioncash, line 397: Perhaps this added block should be placed into its own `TickGC()` function that gets called here, so that the GC-specific stuff is kept separate.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -373,11 +379,82 @@ TextureCache<P>::TextureCache(Runtime& runtime_, VideoCore::RasterizerInterface&
...
+    deletion_iterator = slot_images.begin();
 }
 
 template <class P>
 void TextureCache<P>::TickFrame() {
-    // Tick sentenced resources in this order to ensure they are destroyed in the right order
+    const bool enabled_gc = Settings::values.use_caches_gc.GetValue();
+    if (!enabled_gc) {
+        // @Note(Blinkhawk): compile error with SCOPE_EXIT on msvc.
+        sentenced_images.Tick();
```
lioncash, line 391: Assuming the SCOPE_EXIT stuff cant be resolved for some reason, maybe wrap it into a lambda and call it in here and at the end of the function, so that two sections of code don't need to be remembered if any updates to those code paths are necessary.

src/yuzu_cmd/default_ini.h
```diff
@@ -219,6 +219,10 @@ use_asynchronous_gpu_emulation =
 # 0: Off, 1 (default): On
 use_vsync =
 
+# Whether to use garbage collection or not.
```
lioncash, line 222:
```suggestion
# Whether to use garbage collection or not for GPU caches.
```

Just so that the domain the GC operates in is explicitly stated.

src/yuzu/configuration/configure_graphics_advanced.ui
```diff
@@ -103,6 +103,16 @@
           </property>
          </widget>
         </item>
+        <item>
+         <widget class="QCheckBox" name="use_caches_gc">
+          <property name="toolTip">
+           <string>Enables garbage collection for the GPU caches, this will try to keep VRAM within 3-4 GB by flushing the least used textures/buffers. May cause issues in a few games.</string>
+          </property>
+          <property name="text">
+           <string>Enable GPU caches garbage collection (unsafe)</string>
```
lioncash, line 112:
```suggestion
           <string>Enable GPU cache garbage collection (unsafe)</string>
```

src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -332,10 +345,16 @@ BufferCache<P>::BufferCache(VideoCore::RasterizerInterface& rasterizer_,
       gpu_memory{gpu_memory_}, cpu_memory{cpu_memory_}, runtime{runtime_} {
     // Ensure the first slot is used for the null buffer
     void(slot_buffers.insert(runtime, NullBufferParams{}));
+    deletion_iterator = slot_buffers.end();
 }
 
 template <class P>
 void BufferCache<P>::TickFrame() {
+    const bool enabled_gc = Settings::values.use_caches_gc.GetValue();
```
lioncash, line 353: This seems like it can be moved further down near its usages.

src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -349,7 +368,27 @@ void BufferCache<P>::TickFrame() {
     const bool skip_preferred = hits * 256 < shots * 251;
     uniform_buffer_skip_cache_size = skip_preferred ? DEFAULT_SKIP_CACHE_SIZE : 0;
 
-    delayed_destruction_ring.Tick();
+    const bool activate_gc = enabled_gc && total_used_memory >= expected_memory;
```
lioncash, line 371: Maybe place this block inside its own TickGC function that's called here to have GC handling separated from everything else.

src/video_core/texture_cache/texture_cache.h
```diff
@@ -373,11 +379,82 @@ TextureCache<P>::TextureCache(Runtime& runtime_, VideoCore::RasterizerInterface&
...
+                    SwizzleImage(gpu_memory, image->gpu_addr, image->info, copies, map.mapped_span);
+                }
+            }
+            if (True(image->flags & ImageFlagBits::Tracked)) {
+                UntrackImage(*image);
+            }
+            UnregisterImage(image_id);
+            DeleteImage(image_id);
+            if (is_bad_overlap) {
+                num_iterations++;
```
v1993, line 453:
```suggestion
                ++num_iterations;
```
Sure a nitpick, but you keep using prefix form in other places and consistency won't hurt.

BSoDGamingYT:
This GC causes issues in BOTW when enabled.

If you open the inventory and close it rapidly, RAM and VRAM will jump slightly each time. 

This issue doesnt occur with the GC disabled

gidoly:
reaper works for fate extella and extella link on vukan but open gl seems not works properly

FernandoS27:
> reaper works for fate extella and extella link on vukan but open gl seems not works properly

I've recieved this report before. Since Reaper doesn't do any backend specifioc code there's 2 possible explainations. A. we are not erasing textures correctly on GL or B. the driver will take care of freeing the memory once it needs to rather than we we tell it. The second case is totally normal and we shouldn't do anything, the first case though, we should check it.

v1993:
I'm not sure about this - but what do you think about exposing "video memory amount" setting under OpenGL's advanced graphic settings? Of course, the ideal way of dealing with it would be to query this from driver like it's done on Vulkan, but it seems that it isn't really supported even via extensions.

While I do think that exposing such an option to users isn't a great idea, it seems pretty important to me to at least somehow account for varying vram sizes. Correct me if I'm wrong about this assumption, of course.

FernandoS27:
> I'm not sure about this - but what do you think about exposing "video memory amount" setting under OpenGL's advanced graphic settings? Of course, the ideal way of dealing with it would be to query this from driver like it's done on Vulkan, but it seems that it isn't really supported even via extensions.
> 
> While I do think that exposing such an option to users isn't a great idea, it seems pretty important to me to at least somehow account for varying vram sizes. Correct me if I'm wrong about this assumption, of course.

I would prefer to avoid such settings.

lat9nq:
When the reaper is enabled, each scene change (i.e. to the pause menu and back, to different rooms/maps) causes increased system memory usage in BOTW, Splatoon 2, and ACNH. Splatoon 2 sees about 0.5 GB deltas on RadeonSI, 0.4 GB on NVIDIA OpenGL. The only other game I've tested thus far is Pokemon Sword -- at least that one isn't swallowing memory here like all the others.

master does not exhibit this behavior -- RAM usage is relatively constant between scene switches on these games.

FernandoS27:
> When the reaper is enabled, each scene change (i.e. to the pause menu and back, to different rooms/maps) causes increased system memory usage in BOTW, Splatoon 2, and ACNH. Splatoon 2 sees about 0.5 GB deltas on RadeonSI, 0.4 GB on NVIDIA OpenGL. The only other game I've tested thus far is Pokemon Sword -- at least that one isn't swallowing memory here like all the others.
> 
> master does not exhibit this behavior -- RAM usage is relatively constant between scene switches on these games.

I have looked into this and the leak itself happens on the driver.

In OGL we don't eliminate memory directly, we just release resources and the driver decides when to recycle their memory or not. What I found happening on this games on OGL was that we erased a bunch of textures and then created a bunch of textures after unpausing, the OGL driver just used new memory for that. When the driver starts running out of memory, it should do the appropiate cleaning.

I still wonder though whz is it leaking RAM though.

PacoAvelar:
This function seems to break Blasphemous in game render. Menus still work

bunnei (approved):

lioncash (approved):

PR #6472, opened 2021-06-15
service: spl: Implement general SPL service

Morph1984:
SPL is used by homebrew to determine the firmware it is running on.

Many homebrew will assume that it is being run on Atmosphere, a custom firmware for the Nintendo Switch. As such, we will mimic its behavior and report that we are Atmosphere, allowing for these homebrew to run properly on yuzu.

### DIFF
--- a/src/core/hle/api_version.h
+++ b/src/core/hle/api_version.h
@@ -1,0 +1,38 @@
+// Copyright 2021 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/common_types.h"
+
+// This file contains yuzu's HLE API version constants.
+
+namespace HLE::ApiVersion {
+
+// Horizon OS version constants.
+
+constexpr u8 HOS_VERSION_MAJOR = 11;
+constexpr u8 HOS_VERSION_MINOR = 0;
+constexpr u8 HOS_VERSION_MICRO = 1;
+
+// NintendoSDK version constants.
+
+constexpr u8 SDK_REVISION_MAJOR = 1;
+constexpr u8 SDK_REVISION_MINOR = 0;
+
+constexpr char PLATFORM_STRING[] = "NX";
+constexpr char VERSION_HASH[] = "69103fcb2004dace877094c2f8c29e6113be5dbf";
+constexpr char DISPLAY_VERSION[] = "11.0.1";
+constexpr char DISPLAY_TITLE[] = "NintendoSDK Firmware for NX 11.0.1-1.0";
+
+// Atmosphere version constants.
+
+constexpr u8 ATMOSPHERE_RELEASE_VERSION_MAJOR = 0;
+constexpr u8 ATMOSPHERE_RELEASE_VERSION_MINOR = 19;
+constexpr u8 ATMOSPHERE_RELEASE_VERSION_MICRO = 4;
+
+constexpr u32 GetTargetFirmware() {
+    return u32{HOS_VERSION_MAJOR} << 24 | u32{HOS_VERSION_MINOR} << 16 |
+           u32{HOS_VERSION_MICRO} << 8 | 0U;
+}
+
+} // namespace HLE::ApiVersion
--- a/src/core/hle/service/spl/module.cpp
+++ b/src/core/hle/service/spl/module.cpp
@@ -10,6 +10,7 @@
 #include <vector>
 #include "common/logging/log.h"
 #include "common/settings.h"
+#include "core/hle/api_version.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/service/spl/csrng.h"
 #include "core/hle/service/spl/module.h"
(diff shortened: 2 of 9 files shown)


src/core/hle/service/spl/module.cpp
```diff
@@ -39,6 +79,89 @@ void Module::Interface::GetRandomBytes(Kernel::HLERequestContext& ctx) {
...
+    case ConfigItem::DeviceUniqueKeyGeneration:
+    case ConfigItem::Package2Hash:
+        return ResultSecureMonitorNotImplemented;
+    case ConfigItem::ExosphereApiVersion:
+        // Get information about the current exosphere version.
+        return MakeResult(
+            (static_cast<u64>(ApiVersion::ATMOSPHERE_RELEASE_VERSION_MAJOR & 0xFF) << 56) |
+            (static_cast<u64>(ApiVersion::ATMOSPHERE_RELEASE_VERSION_MINOR & 0xFF) << 48) |
+            (static_cast<u64>(ApiVersion::ATMOSPHERE_RELEASE_VERSION_MICRO & 0xFF) << 40) |
+            (static_cast<u64>(ApiVersion::GetTargetFirmware()) << 00));
```
lioncash, line 129:
```suggestion
            (static_cast<u64>(ApiVersion::GetTargetFirmware())));
```

src/core/hle/service/spl/spl_types.h
```diff
@@ -0,0 +1,207 @@
...
+enum class HardwareType {
+    Icosa = 0,
+    Copper = 1,
+    Hoag = 2,
+    Iowa = 3,
+    Calcio = 4,
+    Aula = 5,
+};
+
+enum SocType {
```
lioncash, line 84: enum class

src/core/hle/service/spl/spl_types.h
```diff
@@ -0,0 +1,207 @@
...
+    SocType_Erista = 0,
+    SocType_Mariko = 1,
+};
+
+enum HardwareState {
```
lioncash, line 89: enum class

src/core/hle/service/spl/spl_types.h
```diff
@@ -0,0 +1,207 @@
...
+    HardwareState_Development = 0,
+    HardwareState_Production = 1,
+};
+
+enum MemoryArrangement {
```
lioncash, line 94: enum class

src/core/hle/service/spl/spl_types.h
```diff
@@ -0,0 +1,207 @@
...
+            u8 rtc_intr;
+            u8 nv_erc;
+            u8 boot_reason;
+        };
+        u32 value;
+    };
+};
+static_assert(sizeof(BootReasonValue) == sizeof(u32), "BootReasonValue definition!");
+
+enum BootReason {
```
lioncash, line 121: enum class

src/core/hle/service/spl/spl_types.h
```diff
@@ -0,0 +1,207 @@
...
+
+    // Note: MemoryArrangement_Dynamic is not official.
+    // Atmosphere uses it to maintain compatibility with firmwares prior to 6.0.0,
+    // which removed the explicit retrieval of memory arrangement from PM.
+    MemoryArrangement_Dynamic = 5,
+    MemoryArrangement_Count,
+};
+
+struct BootReasonValue {
+    union {
```
lioncash, line 109: Consider having functions that return these individual bytes, or use the BitField class.

Modifying or reading out of a non-active union member is undefined behavior.

src/core/hle/service/spl/spl_types.h
```diff
@@ -0,0 +1,207 @@
...
+    BootReason_OnKey = 2,
+    BootReason_RtcAlarm1 = 3,
+    BootReason_RtcAlarm2 = 4,
+};
+
+#pragma pack(push, 1)
+
+struct AesKey {
+    union {
+        std::array<u8, AES_128_KEY_SIZE> data;
```
lioncash, line 133: Similarly for these, these don't have a common initial sequence with one another, so using one to inspect the other would also be undefined behavior here.

You may want to consider accessors for reading/writing out 64-bit values instead of a union.

src/core/hle/api_version.h
```diff
@@ -0,0 +1,38 @@
+// Copyright 2021 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "common/common_types.h"
+
+// This file contains yuzu's HLE API version constants.
+
+namespace ApiVersion {
```
lioncash, line 9:
```suggestion
namespace HLE::ApiVersion {
```

or

```suggestion
namespace Core::ApiVersion {
```

src/core/hle/service/spl/module.cpp
```diff
@@ -39,6 +79,89 @@ void Module::Interface::GetRandomBytes(Kernel::HLERequestContext& ctx) {
...
+    case ConfigItem::IsChargerHiZModeEnabled:
+    case ConfigItem::QuestState:
+    case ConfigItem::RegulatorType:
+    case ConfigItem::DeviceUniqueKeyGeneration:
+    case ConfigItem::Package2Hash:
+        return ResultSecureMonitorNotImplemented;
+    case ConfigItem::ExosphereApiVersion:
+        // Get information about the current exosphere version.
+        return MakeResult(
+            (static_cast<u64>(ApiVersion::ATMOSPHERE_RELEASE_VERSION_MAJOR & 0xFF) << 56) |
```
lioncash, line 126:
```suggestion
            (u64{ApiVersion::ATMOSPHERE_RELEASE_VERSION_MAJOR} << 56) |
```

Ditto for the others below.

src/core/hle/service/spl/spl_types.h
```diff
@@ -0,0 +1,230 @@
...
+    Icosa = 0,
+    Copper = 1,
+    Hoag = 2,
+    Iowa = 3,
+    Calcio = 4,
+    Aula = 5,
+};
+
+enum class SocType {
+    SocType_Erista = 0,
```
lioncash, line 88: Given these are now enum classes, the prefix is no longer necessary. Ditto below

lioncash (approved):

bunnei (approved):

PR #6475, opened 2021-06-16
nvflinger: Add experimental toggle to disable buffer swap interval limits

ameerj:
This adds a toggle which lifts the swap interval waiting on the vsync thread, allowing nvflinger buffers to be pushed for processing by the GPU as soon as they become available.

Display/driver vsync must be disabled for this to setting take effect. The speed at which the title will be running will be increased. There is the potential to cause instabilities in some titles, and the toggle may not work on all titles. 

As such, it remains as a debug setting. The main use case is to allow for potentially easier benchmarking by surpassing the framerate limits of most titles. It can be toggled during runtime via the settings or a hotkey, which is defaulted to `Ctrl+U`

### DIFF
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -149,6 +149,7 @@ struct Values {
     Setting<bool> use_nvdec_emulation;
     Setting<bool> accelerate_astc;
     Setting<bool> use_vsync;
+    Setting<bool> disable_fps_limit;
     Setting<bool> use_assembly_shaders;
     Setting<bool> use_asynchronous_shaders;
     Setting<bool> use_fast_gpu_time;
--- a/src/core/hle/service/nvflinger/nvflinger.cpp
+++ b/src/core/hle/service/nvflinger/nvflinger.cpp
@@ -307,6 +307,9 @@ void NVFlinger::Compose() {
 }
 
 s64 NVFlinger::GetNextTicks() const {
+    if (Settings::values.disable_fps_limit.GetValue()) {
+        return 0;
+    }
     constexpr s64 max_hertz = 120LL;
     return (1000000000 * (1LL << swap_interval)) / max_hertz;
 }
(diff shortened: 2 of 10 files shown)


src/yuzu/configuration/configure_debug.ui
```diff
@@ -138,6 +138,24 @@
...
+        <property name="toolTip">
+         <string>
+          &lt;html&gt;&lt;head/&gt;&lt;body&gt;
+           &lt;p&gt;Presents guest frames as they become available, disabling the fps limit in most titles.&lt;/p&gt;
+           &lt;p&gt;NOTE: Will cause instabilities.&lt;/p&gt;
+          &lt;/body&gt;&lt;/html&gt;
+         </string>
+        </property>
+        <property name="text">
+         <string>Unlimit Framerate (Experimental)</string>
```
bunnei, line 155: nit: Prefer Unlocked framerate or something like that; Not only is it not unlimited, but abbreviations can be weird for non ESL

src/yuzu/configuration/configure_debug.ui
```diff
@@ -138,6 +138,24 @@
         </property>
        </widget>
       </item>
+      <item>
+       <widget class="QCheckBox" name="unlimit_fps">
```
bunnei, line 142: IMO debug is starting to become a bucket of random ass advanced settings. I'd prefer this in Graphics/Advanced, because it is an advanced graphics setting.  There are lots of reasons users will use this outside of debugging (e.g. benchmark their PC, or humblebrag). "Debugging" literally means settings that are only useful for developers while debugging.

src/yuzu/configuration/configure_debug.ui
```diff
@@ -138,6 +138,24 @@
...
+        <property name="enabled">
+         <bool>true</bool>
+        </property>
+        <property name="toolTip">
+         <string>
+          &lt;html&gt;&lt;head/&gt;&lt;body&gt;
+           &lt;p&gt;Presents guest frames as they become available, disabling the fps limit in most titles.&lt;/p&gt;
```
bunnei, line 149: nit: FPS

src/yuzu/configuration/configure_debug.ui
bunnei: Better yet: "Disable framerate limit (Experimental)"

bunnei (approved):

PR #6476, opened 2021-06-17
textures: astc: Make CPU ASTC decompression parallel

Morph1984:
This uses ThreadWorker in common to parallelize CPU ASTC decompression, providing a massive speed increase in decompressing large ASTC textures in games such as Astral Chain.

### DIFF
--- a/src/video_core/textures/astc.cpp
+++ b/src/video_core/textures/astc.cpp
@@ -23,7 +23,9 @@
 
 #include <boost/container/static_vector.hpp>
 
+#include "common/alignment.h"
 #include "common/common_types.h"
+#include "common/thread_worker.h"
 #include "video_core/textures/astc.h"
 
 class InputBitStream {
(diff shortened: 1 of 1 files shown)


src/video_core/textures/astc.cpp
```diff
@@ -1548,30 +1552,62 @@ static void DecompressBlock(std::span<const u8, 16> inBuf, const u32 blockWidth,
...
 
-                // Blocks can be at most 12x12
-                std::array<u32, 12 * 12> uncompData;
-                DecompressBlock(blockPtr, block_width, block_height, uncompData);
+    u32 num_blocks =
+        depth * Common::DivideUp(height, block_height) * Common::DivideUp(width, block_width);
 
-                u32 decompWidth = std::min(block_width, width - x);
-                u32 decompHeight = std::min(block_height, height - y);
+    std::vector<ASTCBlockInfo> astc_blocks(num_blocks);
```
degasus, line 1568: This allocates heap for each texture, this is quite a large overhead. Is there a chance you can move the calculation of blocks `x,y,z` within decompress_impl and just use an outer indexing with a single integer up to `depth * (height/block_height) * (width/block_width)`?

src/video_core/textures/astc.cpp
```diff
@@ -1548,30 +1552,62 @@ static void DecompressBlock(std::span<const u8, 16> inBuf, const u32 blockWidth,
...
 
-                const std::span<u8> outRow = output.subspan(depth_offset + (y * width + x) * 4);
-                for (u32 jj = 0; jj < decompHeight; jj++) {
-                    std::memcpy(outRow.data() + jj * width * 4,
-                                uncompData.data() + jj * block_width, decompWidth * 4);
-                }
+    for (u32 z = 0; z < depth; ++z) {
+        for (u32 y = 0; y < height; y += block_height) {
+            for (u32 x = 0; x < width; x += block_width) {
+                astc_blocks.emplace_back(ASTCBlockInfo{
```
degasus, line 1573: Please be careful with splitting the x blocks in chunks. You need at least once cache line (64 bytes) as stride as margin between two threads to avoid cache fights. With 4 bytes per pixel, this is still 16 pixel width and so at least two blocks.

What do you think about running only lines in parallel instead of blocks?

Morph1984:
I did some more comparison tests, documenting them here:

Tests were done on a Ryzen 5 1600, 3.7 GHz 6C/12T:

Previous impl: 47 seconds
Previous impl (with OpenMP #pragma omp parallel for): 12 seconds
New impl (non-parallel): 80 seconds
New impl (parallel): 16 seconds

Not sure why the non-parallel version of the new impl is much slower, this should be investigated.

v1993:
If OpenMP version beats everything, why not use it? Aside from it not being very C++-like solution.

FernandoS27:
> If OpenMP version beats everything, why not use it? Aside from it not being very C++-like solution.

mostly because that forces to add OpenMP as a requirement to the software and may hinder our portability.

v1993:
> mostly because that forces to add OpenMP as a requirement to the software and may hinder our portability.

Fair enough. Perhaps then test simple thread pool and see what results are, given that it's what OpenMP does anyways.

lat9nq:
This PR and libstdc++ 11 requires linking to Intel's TBB when compiling with GCC 11 and Clang 12. 
The build containers seemed to pass because they simply don't support this thread library whatsoever.

Morph1984:
@lioncash Should we be making use of OpenMP instead? Its pretty much the only choice we have given libstdc++ requires linking to Intel's TBB

ameerj:
Frankly i feel as if this PR is adding too much complexity for what is essentially a debug fallback. 
The compute shader decoder has become much more stable as of #6496 to not warrant the need to add more dependencies to the project for the cpu decoder.

Morph1984:
@v1993 @lat9nq I've done away with the parallelism TS requirement and used `ThreadWorker` in the common library

Morph1984:
Here are the updated benchmark numbers:

On a Ryzen 5 1600, 3.7 GHz 6C/12T:

Previous impl: 47 seconds
Previous impl (with OpenMP #pragma omp parallel for): 12 seconds
ThreadedWorker impl (1 thread): 47-48 seconds
ThreadedWorker impl (11 threads): 16-17 seconds

On an i7-11700K, 5.0 GHz 8C/16T:
Previous impl: 28 seconds
Previous impl (with OpenMP #pragma omp parallel for): 5.5 seconds
ThreadedWorker impl (15 threads): 7 seconds

ameerj:
This MT approach is on my 4.9 ghz 8c/16t 9900k is consistently slower than my GT 730 compute shader decoding by a very large margin. 

MT:

https://user-images.githubusercontent.com/52414509/124858239-ad83ce80-df7b-11eb-88dc-3ef4a13d2eca.mp4


730:

https://user-images.githubusercontent.com/52414509/124858231-a9f04780-df7b-11eb-9a7e-36162b9bfa14.mp4

PR #6478, opened 2021-06-18
vulkan_debug_callback: Skip logging known false-positive validation errors

ameerj:
Some validation errors are known to be false positive. We can skip them from being logged to avoid flooding the log stream with errors that are not applicable to yuzu

### DIFF
--- a/src/video_core/vulkan_common/vulkan_debug_callback.cpp
+++ b/src/video_core/vulkan_common/vulkan_debug_callback.cpp
@@ -12,6 +12,14 @@ VkBool32 Callback(VkDebugUtilsMessageSeverityFlagBitsEXT severity,
                   VkDebugUtilsMessageTypeFlagsEXT type,
                   const VkDebugUtilsMessengerCallbackDataEXT* data,
                   [[maybe_unused]] void* user_data) {
+    // Skip logging known false-positive validation errors
+    switch (static_cast<u32>(data->messageIdNumber)) {
+    case 0x682a878au: // VUID-vkCmdBindVertexBuffers2EXT-pBuffers-parameter
+    case 0x99fb7dfdu: // UNASSIGNED-RequiredParameter (vkCmdBindVertexBuffers2EXT pBuffers[0])
+        return VK_FALSE;
+    default:
+        break;
+    }
     const std::string_view message{data->pMessage};
     if (severity & VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT) {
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_vulkan/renderer_vulkan.cpp
```diff
@@ -124,6 +125,9 @@ RendererVulkan::RendererVulkan(Core::TelemetrySession& telemetry_session_,
 
 RendererVulkan::~RendererVulkan() {
     void(device.GetLogical().WaitIdle());
+    if (Settings::values.renderer_debug) {
+        void(Common::FS::RemoveFile(std::filesystem::current_path() / "vk_layer_settings.txt"));
```
lioncash, line 129: ping @Morph1984 perhaps we should drop the `[[nodiscard]]` attribute on RemoveFile. It's relatively normal to have scenarios where one doesn't really care whether or not the remove operation succeeds, but rather that it attempts a removal.

src/video_core/vulkan_common/vulkan_instance.cpp
```diff
@@ -127,6 +129,11 @@ vk::Instance CreateInstance(const Common::DynamicLibrary& library, vk::InstanceD
         LOG_ERROR(Render_Vulkan, "Failed to load Vulkan function pointers");
         throw vk::Exception(VK_ERROR_INITIALIZATION_FAILED);
     }
+    if (enable_layers) {
+        const auto path{std::filesystem::current_path() / "vk_layer_settings.txt"};
+        const auto layer_settings = Common::FS::IOFile(path, Common::FS::FileAccessMode::Write);
+        void(layer_settings.Write("khronos_validation.message_id_filter=0x99fb7dfd,0x682a878a"));
```
lioncash, line 135: Should anything be logged out if the filesystem fails to open the file or if the write fails?

src/video_core/renderer_vulkan/renderer_vulkan.cpp
ameerj: This is no longer applicable to this PR as I've changed the approach per Rodrigo's suggestions. but feel free to continue the discussion.

Morph1984: @lioncash Yea there are a few places where `[[nodiscard]]` attribute can be removed, such as in RemoveFile like you mentioned and maybe IOFile::Flush?

ReinUsesLisp (approved):

PR #6485, opened 2021-06-18
input_common: TAS with script playback & recording

MonsterDruide1:
This PR introduces a new input device, `TAS`, which will automatically play back a set script of inputs with frame-perfection.

For the most part, the descriptions of commit messages should explain everything clear enough. As stated in the first commit, recording still lacks frame-perfection and perfect reducability, so there will probably be some follow-up PRs once I figure out why inaccuracies happen.

A script file has the same format as [TAS-nx](https://github.com/hamhub7/TAS-nx/tree/master/scripts) uses, so final files will look like this:
```
1 KEY_B 0;0 0;0
6 KEY_ZL 0;0 0;0
41 KEY_ZL;KEY_Y 0;0 0;0
43 KEY_X;KEY_A 32767;0 0;0
44 KEY_A 32767;0 0;0
45 KEY_A 32767;0 0;0
46 KEY_A 32767;0 0;0
47 KEY_A 32767;0 0;0
```

Finished segments of Super Mario Odyssey with video examples and script files can be found [here](https://docs.google.com/spreadsheets/d/19NrdMccG3-8VogOhAI7I9qDRoZYhPfHciPvyUqj7eCg), where the [Mech-TAS](https://youtu.be/_MLAm2Kun88?t=9) probably is the most impressive one so far.

To play back TAS scripts on Yuzu, select the folder with scripts in the configuration menu below `System` -> `Filesystem`. The file itself has normal text format and has to be called `script0-1.txt` for controller 1, `script0-2.txt` for controller 2 and so forth (with max. 8 players).

After placing the file at the correct location, it can be read into Yuzu with the (default) hotkey `CTRL+F6` (`refresh`). In the bottom left corner, it will display the amount of frames the script file has.  Playback can be started or stopped using `CTRL+F5`.

However, for playback to actually work, the correct input device has to be selected: In the `Controls` menu, select `TAS` from hte device list for the controller that the script should be played on.

Recording a new script file is really simple: Just make sure that the proper device (*not* TAS) is connected on P1, and press `CTRL+F7` to start recording. When done, just press the same keystroke again (`CTRL+F7`). The new script will be saved at the location previously selected, as the filename `record.txt`.

For debugging purposes, the common controller debugger can be used (`View` -> `Debugging` -> `Controller P1`).

### DIFF
--- a/src/common/fs/fs_paths.h
+++ b/src/common/fs/fs_paths.h
@@ -21,6 +21,7 @@
 #define SCREENSHOTS_DIR "screenshots"
 #define SDMC_DIR "sdmc"
 #define SHADER_DIR "shader"
+#define TAS_DIR "scripts"
 
 // yuzu-specific files
 
--- a/src/common/fs/fs_paths.h
+++ b/src/common/fs/fs_paths.h
@@ -21,7 +21,7 @@
 #define SCREENSHOTS_DIR "screenshots"
 #define SDMC_DIR "sdmc"
 #define SHADER_DIR "shader"
-#define TAS_DIR "scripts"
+#define TAS_DIR "tas"
 
 // yuzu-specific files
 
--- a/src/common/fs/path_util.cpp
+++ b/src/common/fs/path_util.cpp
@@ -116,7 +116,7 @@ private:
         GenerateYuzuPath(YuzuPath::ScreenshotsDir, yuzu_path / SCREENSHOTS_DIR);
         GenerateYuzuPath(YuzuPath::SDMCDir, yuzu_path / SDMC_DIR);
         GenerateYuzuPath(YuzuPath::ShaderDir, yuzu_path / SHADER_DIR);
-        GenerateYuzuPath(YuzuPath::TASFile, yuzu_path / TAS_DIR);
+        GenerateYuzuPath(YuzuPath::TASDir, yuzu_path / TAS_DIR);
     }
 
     ~PathManagerImpl() = default;
--- a/src/common/fs/path_util.cpp
+++ b/src/common/fs/path_util.cpp
@@ -116,6 +116,8 @@ private:
         GenerateYuzuPath(YuzuPath::ScreenshotsDir, yuzu_path / SCREENSHOTS_DIR);
         GenerateYuzuPath(YuzuPath::SDMCDir, yuzu_path / SDMC_DIR);
         GenerateYuzuPath(YuzuPath::ShaderDir, yuzu_path / SHADER_DIR);
+
+        GenerateYuzuPath(YuzuPath::TASFile, fs::path{""});
     }
 
     ~PathManagerImpl() = default;
--- a/src/common/fs/path_util.cpp
+++ b/src/common/fs/path_util.cpp
@@ -116,8 +116,7 @@ private:
         GenerateYuzuPath(YuzuPath::ScreenshotsDir, yuzu_path / SCREENSHOTS_DIR);
         GenerateYuzuPath(YuzuPath::SDMCDir, yuzu_path / SDMC_DIR);
         GenerateYuzuPath(YuzuPath::ShaderDir, yuzu_path / SHADER_DIR);
-
-        GenerateYuzuPath(YuzuPath::TASFile, fs::path{""});
+        GenerateYuzuPath(YuzuPath::TASFile, yuzu_path / TAS_DIR);
     }
 
     ~PathManagerImpl() = default;
--- a/src/common/fs/path_util.h
+++ b/src/common/fs/path_util.h
@@ -23,8 +23,7 @@ enum class YuzuPath {
     ScreenshotsDir, // Where yuzu screenshots are stored.
     SDMCDir,        // Where the emulated SDMC is stored.
     ShaderDir,      // Where shaders are stored.
-
-    TASFile, // Where the current script file is stored.
+    TASDir,         // Where the current script file is stored.
 };
 
 /**
--- a/src/common/fs/path_util.h
+++ b/src/common/fs/path_util.h
@@ -23,6 +23,8 @@ enum class YuzuPath {
     ScreenshotsDir, // Where yuzu screenshots are stored.
     SDMCDir,        // Where the emulated SDMC is stored.
     ShaderDir,      // Where shaders are stored.
+
+    TASFile, // Where the current script file is stored.
 };
 
 /**
--- a/src/common/fs/path_util.h
+++ b/src/common/fs/path_util.h
@@ -23,7 +23,7 @@ enum class YuzuPath {
     ScreenshotsDir, // Where yuzu screenshots are stored.
     SDMCDir,        // Where the emulated SDMC is stored.
     ShaderDir,      // Where shaders are stored.
-    TASDir,         // Where the current script file is stored.
+    TASDir,         // Where TAS scripts are stored.
 };
 
 /**
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -14,6 +14,7 @@
 #include <utility>
 #include <vector>
 
+#include <input_common/main.h>
 #include "common/common_types.h"
 #include "common/settings_input.h"
 
@@ -499,6 +500,7 @@ struct Values {
 
     // Controls
     InputSetting<std::array<PlayerInput, 10>> players;
+    std::shared_ptr<InputCommon::InputSubsystem> inputSubsystem = NULL;
 
     Setting<bool> use_docked_mode{true, "use_docked_mode"};
 
@@ -512,9 +514,14 @@ struct Values {
                                             "motion_device"};
     BasicSetting<std::string> udp_input_servers{"127.0.0.1:26760", "udp_input_servers"};
 
+    BasicSetting<bool> tas_enable{false, "tas_enable"};
+    BasicSetting<bool> tas_reset{ false, "tas_reset" };
+    BasicSetting<bool> tas_record{ false, "tas_record" };
+
     BasicSetting<bool> mouse_panning{false, "mouse_panning"};
     BasicRangedSetting<u8> mouse_panning_sensitivity{10, 1, 100, "mouse_panning_sensitivity"};
     BasicSetting<bool> mouse_enabled{false, "mouse_enabled"};
+
     std::string mouse_device;
     MouseButtonsRaw mouse_buttons;
 
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -550,6 +550,9 @@ struct Values {
     BasicSetting<bool> gamecard_current_game{false, "gamecard_current_game"};
     BasicSetting<std::string> gamecard_path{std::string(), "gamecard_path"};
 
+    // TAS
+    bool pauseTasOnLoad;
+
     // Debugging
     bool record_frame_times;
     BasicSetting<bool> use_gdbstub{false, "use_gdbstub"};
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -14,7 +14,6 @@
 #include <utility>
 #include <vector>
 
-#include <input_common/main.h>
 #include "common/common_types.h"
 #include "common/settings_input.h"
 
@@ -515,9 +514,9 @@ struct Values {
 
     BasicSetting<bool> pause_tas_on_load { false, "pause_tas_on_load" };
     BasicSetting<bool> tas_enable{ false, "tas_enable" };
-    BasicSetting<bool> tas_reset{ false, "tas_reset" };
-    BasicSetting<bool> tas_record{ false, "tas_record" };
-    BasicSetting<bool> is_cpu_boxted{ false, "    BasicSetting<bool> is_cpu_boxted{ false, "cpuBoosted" };
+    BasicSetting<bool> tas_loop{ false, "tas_loop" };
+    BasicSetting<bool> tas_swap_controllers{ false, "tas_swap_controllers" };
+    BasicSetting<bool> is_cpu_boosted{ false,  "is_cpu_boosted" };
 " };
 
     BasicSetting<bool> mouse_panning{false, "mouse_panning"};
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -512,17 +512,14 @@ struct Values {
                                             "motion_device"};
     BasicSetting<std::string> udp_input_servers{"127.0.0.1:26760", "udp_input_servers"};
 
-    BasicSetting<bool> pause_tas_on_load { false, "pause_tas_on_load" };
-    BasicSetting<bool> tas_enable{ false, "tas_enable" };
-    BasicSetting<bool> tas_loop{ false, "tas_loop" };
-    BasicSetting<bool> tas_swap_controllers{ false, "tas_swap_controllers" };
-    BasicSetting<bool> is_cpu_boosted{ false,  "is_cpu_boosted" };
-" };
+    BasicSetting<bool> pause_tas_on_load{true, "pause_tas_on_load"};
+    BasicSetting<bool> tas_enable{false, "tas_enable"};
+    BasicSetting<bool> tas_loop{false, "tas_loop"};
+    BasicSetting<bool> tas_swap_controllers{true, "tas_swap_controllers"};
 
     BasicSetting<bool> mouse_panning{false, "mouse_panning"};
     BasicRangedSetting<u8> mouse_panning_sensitivity{10, 1, 100, "mouse_panning_sensitivity"};
     BasicSetting<bool> mouse_enabled{false, "mouse_enabled"};
-
     std::string mouse_device;
     MouseButtonsRaw mouse_buttons;
 
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -500,7 +500,6 @@ struct Values {
 
     // Controls
     InputSetting<std::array<PlayerInput, 10>> players;
-    std::shared_ptr<InputCommon::InputSubsystem> inputSubsystem = NULL;
 
     Setting<bool> use_docked_mode{true, "use_docked_mode"};
 
@@ -514,9 +513,12 @@ struct Values {
                                             "motion_device"};
     BasicSetting<std::string> udp_input_servers{"127.0.0.1:26760", "udp_input_servers"};
 
-    BasicSetting<bool> tas_enable{false, "tas_enable"};
+    BasicSetting<bool> pause_tas_on_load { false, "pause_tas_on_load" };
+    BasicSetting<bool> tas_enable{ false, "tas_enable" };
     BasicSetting<bool> tas_reset{ false, "tas_reset" };
     BasicSetting<bool> tas_record{ false, "tas_record" };
+    BasicSetting<bool> is_cpu_boxted{ false, "    BasicSetting<bool> is_cpu_boxted{ false, "cpuBoosted" };
+" };
 
     BasicSetting<bool> mouse_panning{false, "mouse_panning"};
     BasicRangedSetting<u8> mouse_panning_sensitivity{10, 1, 100, "mouse_panning_sensitivity"};
@@ -549,9 +551,6 @@ struct Values {
     BasicSetting<bool> gamecard_inserted{false, "gamecard_inserted"};
     BasicSetting<bool> gamecard_current_game{false, "gamecard_current_game"};
     BasicSetting<std::string> gamecard_path{std::string(), "gamecard_path"};
-
-    // TAS
-    bool pauseTasOnLoad;
 
     // Debugging
     bool record_frame_times;
--- a/src/core/hle/service/vi/vi.cpp
+++ b/src/core/hle/service/vi/vi.cpp
@@ -32,8 +32,6 @@
 #include "core/hle/service/vi/vi_s.h"
 #include "core/hle/service/vi/vi_u.h"
 
-#include "input_common/tas/tas_input.h"
-
 namespace Service::VI {
 
 constexpr ResultCode ERR_OPERATION_FAILED{ErrorModule::VI, 1};
@@ -597,7 +595,6 @@ private:
 
             IGBPQueueBufferResponseParcel response{1280, 720};
             ctx.WriteBuffer(response.Serialize());
-            Settings::values.input_subsystem->GetTas()->UpdateThread();
             break;
         }
         case TransactionId::Query: {
--- a/src/core/hle/service/vi/vi.cpp
+++ b/src/core/hle/service/vi/vi.cpp
@@ -32,6 +32,8 @@
 #include "core/hle/service/vi/vi_s.h"
 #include "core/hle/service/vi/vi_u.h"
 
+#include "input_common/tas/tas_input.h"
+
 namespace Service::VI {
 
 constexpr ResultCode ERR_OPERATION_FAILED{ErrorModule::VI, 1};
@@ -595,6 +597,7 @@ private:
 
             IGBPQueueBufferResponseParcel response{1280, 720};
             ctx.WriteBuffer(response.Serialize());
+            Settings::values.input_subsystem->GetTas()->UpdateThread();
             break;
         }
         case TransactionId::Query: {
--- a/src/input_common/main.cpp
+++ b/src/input_common/main.cpp
@@ -5,6 +5,7 @@
 #include <memory>
 #include <thread>
 #include "common/param_package.h"
+#include "common/settings.h"
 #include "input_common/analog_from_button.h"
 #include "input_common/gcadapter/gc_adapter.h"
 #include "input_common/gcadapter/gc_poller.h"
@@ -114,8 +115,11 @@ struct InputSubsystem::Impl {
         std::vector<Common::ParamPackage> devices = {
             Common::ParamPackage{{"display", "Any"}, {"class", "any"}},
             Common::ParamPackage{{"display", "Keyboard/Mouse"}, {"class", "keyboard"}},
-            Common::ParamPackage{{"display", "TAS"}, {"class", "tas"}},
         };
+        if (Settings::values.tas_enable) {
+            devices.push_back(
+                Common::ParamPackage{{"display", "TAS Controller"}, {"class", "tas"}});
+        }
 #ifdef HAVE_SDL2
         auto sdl_devices = sdl->GetInputDevices();
         devices.insert(devices.end(), sdl_devices.begin(), sdl_devices.end());
--- a/src/input_common/main.cpp
+++ b/src/input_common/main.cpp
@@ -117,7 +117,7 @@ struct InputSubsystem::Impl {
             Common::ParamPackage{{"display", "Keyboard/Mouse"}, {"class", "keyboard"}},
         };
         if (Settings::values.tas_enable) {
-            devices.push_back(
+            devices.emplace_back(
                 Common::ParamPackage{{"display", "TAS Controller"}, {"class", "tas"}});
         }
 #ifdef HAVE_SDL2
--- a/src/input_common/main.cpp
+++ b/src/input_common/main.cpp
@@ -13,6 +13,8 @@
 #include "input_common/motion_from_button.h"
 #include "input_common/mouse/mouse_input.h"
 #include "input_common/mouse/mouse_poller.h"
+#include "input_common/tas/tas_input.h"
+#include "input_common/tas/tas_poller.h"
 #include "input_common/touch_from_button.h"
 #include "input_common/udp/client.h"
 #include "input_common/udp/udp.h"
@@ -60,6 +62,12 @@ struct InputSubsystem::Impl {
         Input::RegisterFactory<Input::MotionDevice>("mouse", mousemotion);
         mousetouch = std::make_shared<MouseTouchFactory>(mouse);
         Input::RegisterFactory<Input::TouchDevice>("mouse", mousetouch);
+
+        tas = std::make_shared<TasInput::Tas>();
+        tasbuttons = std::make_shared<TasButtonFactory>(tas);
+        Input::RegisterFactory<Input::ButtonDevice>("tas", tasbuttons);
+        tasanalog = std::make_shared<TasAnalogFactory>(tas);
+        Input::RegisterFactory<Input::AnalogDevice>("tas", tasanalog);
     }
 
     void Shutdown() {
@@ -94,12 +102,19 @@ struct InputSubsystem::Impl {
         mouseanalog.reset();
         mousemotion.reset();
         mousetouch.reset();
+
+        Input::UnregisterFactory<Input::ButtonDevice>("tas");
+        Input::UnregisterFactory<Input::AnalogDevice>("tas");
+
+        tasbuttons.reset();
+        tasanalog.reset();
     }
 
     [[nodiscard]] std::vector<Common::ParamPackage> GetInputDevices() const {
         std::vector<Common::ParamPackage> devices = {
             Common::ParamPackage{{"display", "Any"}, {"class", "any"}},
             Common::ParamPackage{{"display", "Keyboard/Mouse"}, {"class", "keyboard"}},
+            Common::ParamPackage{{"display", "TAS"}, {"class", "tas"}},
         };
 #ifdef HAVE_SDL2
         auto sdl_devices = sdl->GetInputDevices();
@@ -120,6 +135,9 @@ struct InputSubsystem::Impl {
         if (params.Get("class", "") == "gcpad") {
             return gcadapter->GetAnalogMappingForDevice(params);
         }
+        if (params.Get("class", "") == "tas") {
+            return tas->GetAnalogMappingForDevice(params);
+        }
 #ifdef HAVE_SDL2
         if (params.Get("class", "") == "sdl") {
             return sdl->GetAnalogMappingForDevice(params);
@@ -135,6 +153,9 @@ struct InputSubsystem::Impl {
         }
         if (params.Get("class", "") == "gcpad") {
             return gcadapter->GetButtonMappingForDevice(params);
+        }
+        if (params.Get("class", "") == "tas") {
+            return tas->GetButtonMappingForDevice(params);
         }
 #ifdef HAVE_SDL2
         if (params.Get("class", "") == "sdl") {
@@ -174,9 +195,12 @@ struct InputSubsystem::Impl {
     std::shared_ptr<MouseAnalogFactory> mouseanalog;
     std::shared_ptr<MouseMotionFactory> mousemotion;
     std::shared_ptr<MouseTouchFactory> mousetouch;
+    std::shared_ptr<TasButtonFactory> tasbuttons;
+    std::shared_ptr<TasAnalogFactory> tasanalog;
     std::shared_ptr<CemuhookUDP::Client> udp;
     std::shared_ptr<GCAdapter::Adapter> gcadapter;
     std::shared_ptr<MouseInput::Mouse> mouse;
+    std::shared_ptr<TasInput::Tas> tas;
 };
 
 InputSubsystem::InputSubsystem() : impl{std::make_unique<Impl>()} {}
@@ -205,6 +229,14 @@ MouseInput::Mouse* InputSubsystem::GetMouse() {
 
 const MouseInput::Mouse* InputSubsystem::GetMouse() const {
     return impl->mouse.get();
+}
+
+TasInput::Tas* InputSubsystem::GetTas() {
+    return impl->tas.get();
+}
+
+const TasInput::Tas* InputSubsystem::GetTas() const {
+    return impl->tas.get();
 }
 
 std::vector<Common::ParamPackage> InputSubsystem::GetInputDevices() const {
@@ -285,6 +317,22 @@ MouseTouchFactory* InputSubsystem::GetMouseTouch() {
 
 const MouseTouchFactory* InputSubsystem::GetMouseTouch() const {
     return impl->mousetouch.get();
+}
+
+TasButtonFactory* InputSubsystem::GetTasButtons() {
+    return impl->tasbuttons.get();
+}
+
+const TasButtonFactory* InputSubsystem::GetTasButtons() const {
+    return impl->tasbuttons.get();
+}
+
+TasAnalogFactory* InputSubsystem::GetTasAnalogs() {
+    return impl->tasanalog.get();
+}
+
+const TasAnalogFactory* InputSubsystem::GetTasAnalogs() const {
+    return impl->tasanalog.get();
 }
 
 void InputSubsystem::ReloadInputDevices() {
--- a/src/input_common/main.h
+++ b/src/input_common/main.h
@@ -29,6 +29,10 @@ namespace MouseInput {
 class Mouse;
 }
 
+namespace TasInput {
+class Tas;
+}
+
 namespace InputCommon {
 namespace Polling {
 
@@ -64,6 +68,8 @@ class MouseButtonFactory;
 class MouseAnalogFactory;
 class MouseMotionFactory;
 class MouseTouchFactory;
+class TasButtonFactory;
+class TasAnalogFactory;
 class Keyboard;
 
 /**
@@ -103,6 +109,11 @@ public:
     /// Retrieves the underlying mouse device.
     [[nodiscard]] const MouseInput::Mouse* GetMouse() const;
 
+    /// Retrieves the underlying tas device.
+    [[nodiscard]] TasInput::Tas* GetTas();
+
+    /// Retrieves the underlying tas device.
+    [[nodiscard]] const TasInput::Tas* GetTas() const;
     /**
      * Returns all available input devices that this Factory can create a new device with.
      * Each returned ParamPackage should have a `display` field used for display, a class field for
@@ -167,6 +178,18 @@ public:
 
     /// Retrieves the underlying udp touch handler.
     [[nodiscard]] const MouseTouchFactory* GetMouseTouch() const;
+
+    /// Retrieves the underlying tas button handler.
+    [[nodiscard]] TasButtonFactory* GetTasButtons();
+
+    /// Retrieves the underlying tas button handler.
+    [[nodiscard]] const TasButtonFactory* GetTasButtons() const;
+
+    /// Retrieves the underlying tas touch handler.
+    [[nodiscard]] TasAnalogFactory* GetTasAnalogs();
+
+    /// Retrieves the underlying tas touch handler.
+    [[nodiscard]] const TasAnalogFactory* GetTasAnalogs() const;
 
     /// Reloads the input devices
     void ReloadInputDevices();
--- a/src/input_common/main.h
+++ b/src/input_common/main.h
@@ -155,28 +155,28 @@ public:
     /// Retrieves the underlying udp touch handler.
     [[nodiscard]] const UDPTouchFactory* GetUDPTouch() const;
 
-    /// Retrieves the underlying GameCube button handler.
+    /// Retrieves the underlying mouse button handler.
     [[nodiscard]] MouseButtonFactory* GetMouseButtons();
 
-    /// Retrieves the underlying GameCube button handler.
+    /// Retrieves the underlying mouse button handler.
     [[nodiscard]] const MouseButtonFactory* GetMouseButtons() const;
 
-    /// Retrieves the underlying udp touch handler.
+    /// Retrieves the underlying mouse analog handler.
     [[nodiscard]] MouseAnalogFactory* GetMouseAnalogs();
 
-    /// Retrieves the underlying udp touch handler.
+    /// Retrieves the underlying mouse analog handler.
     [[nodiscard]] const MouseAnalogFactory* GetMouseAnalogs() const;
 
-    /// Retrieves the underlying udp motion handler.
+    /// Retrieves the underlying mouse motion handler.
     [[nodiscard]] MouseMotionFactory* GetMouseMotions();
 
-    /// Retrieves the underlying udp motion handler.
+    /// Retrieves the underlying mouse motion handler.
     [[nodiscard]] const MouseMotionFactory* GetMouseMotions() const;
 
-    /// Retrieves the underlying udp touch handler.
+    /// Retrieves the underlying mouse touch handler.
     [[nodiscard]] MouseTouchFactory* GetMouseTouch();
 
-    /// Retrieves the underlying udp touch handler.
+    /// Retrieves the underlying mouse touch handler.
     [[nodiscard]] const MouseTouchFactory* GetMouseTouch() const;
 
     /// Retrieves the underlying tas button handler.
@@ -185,10 +185,10 @@ public:
     /// Retrieves the underlying tas button handler.
     [[nodiscard]] const TasButtonFactory* GetTasButtons() const;
 
-    /// Retrieves the underlying tas touch handler.
+    /// Retrieves the underlying tas analogs handler.
     [[nodiscard]] TasAnalogFactory* GetTasAnalogs();
 
-    /// Retrieves the underlying tas touch handler.
+    /// Retrieves the underlying tas analogs handler.
     [[nodiscard]] const TasAnalogFactory* GetTasAnalogs() const;
 
     /// Reloads the input devices
--- a/src/input_common/tas/tas_input.cpp
+++ b/src/input_common/tas/tas_input.cpp
@@ -61,8 +61,8 @@ void Tas::LoadTasFile(size_t player_index) {
         commands[player_index].clear();
     }
     std::string file =
-        Common::FS::ReadStringFromFile(Common::FS::GetYuzuPathString(Common::FS::YuzuPath::TASDir) +
-                                           "script0-" + std::to_string(player_index + 1) + ".txt",
+        Common::FS::ReadStringFromFile(Common::FS::GetYuzuPath(Common::FS::YuzuPath::TASDir) /
+                                           fmt::format("script0-{}.txt", player_index + 1),
                                        Common::FS::FileType::BinaryFile);
     std::stringstream command_line(file);
     std::string line;
@@ -102,7 +102,7 @@ void Tas::LoadTasFile(size_t player_index) {
     LOG_INFO(Input, "TAS file loaded! {} frames", frame_no);
 }
 
-void Tas::WriteTasFile(std::string file_name) {
+void Tas::WriteTasFile(std::u8string file_name) {
     std::string output_text;
     for (size_t frame = 0; frame < record_commands.size(); frame++) {
         if (!output_text.empty()) {
@@ -112,8 +112,8 @@ void Tas::WriteTasFile(std::string file_name) {
         output_text += std::to_string(frame) + " " + WriteCommandButtons(line.buttons) + " " +
                        WriteCommandAxis(line.l_axis) + " " + WriteCommandAxis(line.r_axis);
     }
-    const size_t bytes_written = Common::FS::WriteStringToFile(
-        Common::FS::GetYuzuPathString(Common::FS::YuzuPath::TASDir) + file_name,
+    const auto bytes_written = Common::FS::WriteStringToFile(
+        Common::FS::GetYuzuPath(Common::FS::YuzuPath::TASDir) / file_name,
         Common::FS::FileType::TextFile, output_text);
     if (bytes_written == output_text.size()) {
         LOG_INFO(Input, "TAS file written to file!");
@@ -217,6 +217,9 @@ void Tas::UpdateThread() {
             is_running = Settings::values.tas_loop;
             current_command = 0;
             tas_data.fill({});
+            if (!is_running) {
+                SwapToStoredController();
+            }
         }
     } else {
         tas_data.fill({});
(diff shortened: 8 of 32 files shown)


src/core/hle/service/vi/vi.cpp
```diff
@@ -32,6 +32,8 @@
 #include "core/hle/service/vi/vi_s.h"
 #include "core/hle/service/vi/vi_u.h"
 
+#include <input_common/tas/tas_input.h>
```
lioncash, line 35: This should be using `"` instead of `<` like regular includes.

src/common/settings.h
```diff
@@ -182,6 +184,11 @@ struct Values {
     std::string motion_device;
     std::string udp_input_servers;
 
+    bool tas_enable;
+    bool tas_reset;
+    bool tas_record;
+    bool cpuBoosted;
```
lioncash, line 190:
```suggestion
    bool is_cpu_boosted;
```

src/common/settings.h
```diff
@@ -172,6 +173,7 @@ struct Values {
 
     // Controls
     InputSetting<std::array<PlayerInput, 10>> players;
+    std::shared_ptr<InputCommon::InputSubsystem> inputSubsystem = NULL;
```
lioncash, line 176:
```suggestion
    std::shared_ptr<InputCommon::InputSubsystem> input_subsystem;
```

This also shouldn't be inside of the common library.

src/common/settings.h
```diff
@@ -12,6 +12,7 @@
 #include <string>
 #include <vector>
 
+#include <input_common/main.h>
```
lioncash, line 15: the common library shouldn't be depending on anything from input_common. `common` should be able to be built entirely by itself.

I'd strongly consider placing anything from input common elsewhere.

src/common/settings.h
```diff
@@ -212,6 +219,9 @@ struct Values {
     bool gamecard_current_game;
     std::string gamecard_path;
 
+    // TAS
+    bool pauseTasOnLoad;
```
lioncash, line 223:
```suggestion
    bool pause_tas_on_load;
```

src/common/fs/path_util.cpp
```diff
@@ -116,6 +116,8 @@ class PathManagerImpl {
         GenerateYuzuPath(YuzuPath::ScreenshotsDir, yuzu_path / SCREENSHOTS_DIR);
         GenerateYuzuPath(YuzuPath::SDMCDir, yuzu_path / SDMC_DIR);
         GenerateYuzuPath(YuzuPath::ShaderDir, yuzu_path / SHADER_DIR);
+
+        GenerateYuzuPath(YuzuPath::TASFile, fs::path{""});
```
lioncash, line 120: This should probably have a dedicated directory instead of the empty string.

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+
+#include "common/common_types.h"
+#include "core/frontend/input.h"
+#include "input_common/main.h"
+
+#define PLAYER_NUMBER 8
+
+namespace TasInput {
+
+using TasAnalog = std::tuple<float, float>;
```
lioncash, line 19:
```suggestion
using TasAnalog = std::pair<float, float>;
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+
+#include <array>
+#include <mutex>
+#include <thread>
+
+#include "common/common_types.h"
+#include "core/frontend/input.h"
+#include "input_common/main.h"
+
+#define PLAYER_NUMBER 8
```
lioncash, line 15: This should be a constexpr variable inside of the TasInput namespace, not a define.

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,340 @@
...
+}
+
+void Tas::RefreshTasFile() {
+    refresh_tas_fle = true;
+}
+void Tas::LoadTasFiles() {
+    scriptLength = 0;
+    for (int i = 0; i < PLAYER_NUMBER; i++) {
+        LoadTasFile(i);
+        if (newCommands[i].size() > scriptLength)
```
lioncash, line 37: All conditionals should be braced, even if there's only one statement inside of them.

Ditto elsewhere in the file.

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+    BUTTON_UP = 0x002000,
+    BUTTON_RIGHT = 0x004000,
+    BUTTON_DOWN = 0x008000,
+    BUTTON_SL = 0x010000,
+    BUTTON_SR = 0x020000,
+    BUTTON_HOME = 0x040000,
+    BUTTON_CAPTURE = 0x080000,
+};
+
+static const std::array<std::pair<std::string, TasButton>, 20> text_to_tas_button = {
```
lioncash, line 44: This shoiuld be in the cpp file. You'll also want to store `std::string_view` instances in it instead of `std::string`, so that this can be constexpr.

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+    u32 buttons{};
+    std::array<float, 4> axis{};
+};
+
+class Tas {
+public:
+    Tas();
+    ~Tas();
+
+    static std::string buttonsToString(u32 button) {
```
lioncash, line 85: We use `CamelCase` for function names. Ditto elsewhere this is applicable.

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+        TasAnalog l_axis{};
+        TasAnalog r_axis{};
+    };
+    void LoadTasFile(int playerIndex);
+    void WriteTasFile();
+    TasAnalog ReadCommandAxis(const std::string line) const;
+    u32 ReadCommandButtons(const std::string line) const;
+    std::string WriteCommandButtons(u32 data) const;
+    std::string WriteCommandAxis(TasAnalog data) const;
+    std::pair<float, float> flipY(std::pair<float, float> old) const;
```
lioncash, line 152:
```suggestion
    std::pair<float, float> FlipY(std::pair<float, float> old) const;
```

This can likely also be a static member function inside of the cpp file.

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+
+    size_t scriptLength{0};
```
lioncash, line 154:
```suggestion
    size_t script_length{0};
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+    std::array<TasData, PLAYER_NUMBER> tas_data;
+    bool update_thread_running{true};
+    bool refresh_tas_fle{false};
+    std::array<std::vector<TASCommand>, PLAYER_NUMBER> newCommands{};
```
lioncash, line 158:
```suggestion
    std::array<std::vector<TASCommand>, PLAYER_NUMBER> new_commands{};
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+    std::vector<TASCommand> recordCommands{};
```
lioncash, line 159:
```suggestion
    std::vector<TASCommand> record_commands{};
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+    std::size_t current_command{0};
+    TASCommand lastInput{}; // only used for recording
```
lioncash, line 161:
```suggestion
    TASCommand last_input{}; // only used for recording
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+
+private:
+    struct TASCommand {
+        u32 buttons{};
+        TasAnalog l_axis{};
+        TasAnalog r_axis{};
+    };
+    void LoadTasFile(int playerIndex);
+    void WriteTasFile();
+    TasAnalog ReadCommandAxis(const std::string line) const;
```
lioncash, line 148:
```suggestion
    TasAnalog ReadCommandAxis(const std::string& line) const;
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+    u32 ReadCommandButtons(const std::string line) const;
```
lioncash, line 149:
```suggestion
    u32 ReadCommandButtons(const std::string& line) const;
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+        if ((button & static_cast<u32>(TasInput::TasButton::BUTTON_HOME)) != 0)
+            returns += ", HOME";
+        if ((button & static_cast<u32>(TasInput::TasButton::BUTTON_CAPTURE)) != 0)
+            returns += ", CAPTURE";
+        return returns.length() != 0 ? returns.substr(2) : "";
+    }
+
+    void RefreshTasFile();
+    void LoadTasFiles();
+    void RecordInput(u32 buttons, std::array<std::pair<float, float>, 2> axes);
```
lioncash, line 132:
```suggestion
    void RecordInput(u32 buttons, const std::array<std::pair<float, float>, 2>& axes);
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+    InputCommon::AnalogMapping GetAnalogMappingForDevice(const Common::ParamPackage& params) const;
+    [[nodiscard]] const TasData& GetTasState(std::size_t pad) const;
+
+private:
+    struct TASCommand {
+        u32 buttons{};
+        TasAnalog l_axis{};
+        TasAnalog r_axis{};
+    };
+    void LoadTasFile(int playerIndex);
```
lioncash, line 146:
```suggestion
    void LoadTasFile(int player_index);
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,163 @@
...
+            returns += ", DOWN";
+        if ((button & static_cast<u32>(TasInput::TasButton::BUTTON_SL)) != 0)
+            returns += ", SL";
+        if ((button & static_cast<u32>(TasInput::TasButton::BUTTON_SR)) != 0)
+            returns += ", SR";
+        if ((button & static_cast<u32>(TasInput::TasButton::BUTTON_HOME)) != 0)
+            returns += ", HOME";
+        if ((button & static_cast<u32>(TasInput::TasButton::BUTTON_CAPTURE)) != 0)
+            returns += ", CAPTURE";
+        return returns.length() != 0 ? returns.substr(2) : "";
```
lioncash, line 127:
```suggestion
        return returns.empty() ? "" : returns.substr(2);
```

src/yuzu/main.h
```diff
@@ -315,6 +315,7 @@ private slots:
     QLabel* emu_speed_label = nullptr;
     QLabel* game_fps_label = nullptr;
     QLabel* emu_frametime_label = nullptr;
+    QLabel* TASlabel;
```
lioncash, line 318:
```suggestion
    QLabel* tas_label;
```

src/yuzu/main.cpp
```diff
@@ -204,6 +206,7 @@ GMainWindow::GMainWindow()
     : input_subsystem{std::make_shared<InputCommon::InputSubsystem>()},
       config{std::make_unique<Config>()}, vfs{std::make_shared<FileSys::RealVfsFilesystem>()},
       provider{std::make_unique<FileSys::ManualContentProvider>()} {
+    Settings::values.inputSubsystem = input_subsystem;
```
lioncash, line 209: This will cause the input subsystem to live indefinitely until the config instance is destroyed on program exit. I'd strongly consider not storing a shared_ptr here.

src/yuzu/main.cpp
```diff
@@ -2871,6 +2893,8 @@ void GMainWindow::UpdateStatusBar() {
         return;
     }
 
+    TASlabel->setText(tr(input_subsystem->GetTas()->GetStatusDescription().c_str()));
```
lioncash, line 2896: Qt Linguist will be unable to translate these strings since the strings returned by the function aren't marked for translation.

I'd consider making the UI query the TAS input state and specifying the tranlation strings in the frontend to use based off of the TAS state.

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+std::string debugButtons(u32 buttons) {
+    return "{ " + TasInput::Tas::buttonsToString(buttons) + " }";
+}
+
+std::string debugJoystick(float x, float y) {
+    return "[ " + std::to_string(x) + "," + std::to_string(y) + " ]";
+}
+
+std::string debugInput(TasData data) {
```
lioncash, line 136:
```suggestion
static std::string DebugInput(const TasData& data) {
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,340 @@
...
+    return "{ " + debugButtons(data.buttons) + " , " + debugJoystick(data.axis[0], data.axis[1]) +
+           " , " + debugJoystick(data.axis[2], data.axis[3]) + " }";
+}
+
+std::string debugInputs(std::array<TasData, PLAYER_NUMBER> arr) {
```
lioncash, line 141:
```suggestion
static std::string DebugInputs(const std::array<TasData, PLAYER_NUMBER>& arr) {
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,340 @@
...
+    }
+    if (Settings::values.tas_enable) {
+        return "Playing TAS: " + std::to_string(current_command) + "/" +
+               std::to_string(scriptLength);
+    }
+    return "TAS not running: " + std::to_string(current_command) + "/" +
+           std::to_string(scriptLength);
+}
+
+std::string debugButtons(u32 buttons) {
```
lioncash, line 128:
```suggestion
static std::string DebugButtons(u32 buttons) {
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,340 @@
...
+    return "{ " + TasInput::Tas::buttonsToString(buttons) + " }";
+}
+
+std::string debugJoystick(float x, float y) {
```
lioncash, line 132:
```suggestion
static std::string DebugJoystick(float x, float y) {
```

src/common/fs/path_util.cpp
MonsterDruide1: I don't think that this folder should go into the `AppData`, as it has to be accessed very quickly. (repeatedly changing the script file to adjust timings/angles) Therefor I wanted to force the users to pick a custom directory before even starting to do anything related to TASing.

lioncash: Sure, but that isn't good UX. We should have a reasonable basic default already selected (an AppData scripts folder), which users can then customize to be something else.

src/input_common/tas/tas_input.h
MonsterDruide1: Does that mean completely removing the function prototype from the header file and just making it a static method in the `cpp` like the `Debug[...]`-methods?

src/input_common/tas/tas_input.cpp
MonsterDruide1: Does this mean wrapping that expression in additional brackets, like this?
```cpp
if ((commands[i].size() > script_length))
```

lioncash: It means adding braces like 

```cpp
if (cond) {
    // stuff here
}
```

instead of

```cpp
if (cond)
    // thing here
```

src/input_common/tas/tas_input.h
lioncash: Yeah

src/yuzu/main.cpp
MonsterDruide1: This is directly related to [this](https://github.com/yuzu-emu/yuzu/pull/6485#discussion_r654657077) review, so I'll just mark this one as resolved. If I / someone else finds a way around storing an instance of the `InputSubsystem` in the settings, this will be fixed as well.

src/yuzu/main.cpp
MonsterDruide1: Where do the strings have to end up if they should be translatable?

lioncash: The strings need to be in the frontend, which is why I suggested that the TAS class be designed in a way to allow other observer classes to listen on state changes.

MonsterDruide1: I mean is there any specific file I should route it to? Or is the `main.cpp` file of `yuzu`'s project enough?

src/input_common/tas/tas_input.h
```diff
@@ -18,6 +18,12 @@ constexpr int PLAYER_NUMBER = 8;
 
 using TasAnalog = std::pair<float, float>;
 
+enum class TasState {
+    RUNNING,
```
lioncash, line 22: Our conventions for enum classes is to title case the entries,

so:

```cpp
Running,
Recording,
Stopped,
```

src/yuzu/main.cpp
```diff
@@ -2887,13 +2887,27 @@ void GMainWindow::UpdateWindowTitle(const std::string& title_name,
     }
 }
 
+static std::string GetTasStateDescription(TasInput::TasState state) {
```
lioncash, line 2890: You may wish to make this a member function and apply tr() to each of the strings directly. This way call sites don't need to remember to translate the strings.

src/input_common/tas/tas_input.h
```diff
@@ -108,7 +114,7 @@ class Tas {
     void LoadTasFiles();
     void RecordInput(u32 buttons, const std::array<std::pair<float, float>, 2>& axes);
     void UpdateThread();
-    std::string GetStatusDescription();
+    std::tuple<TasState, size_t, size_t> GetStatus();
```
lioncash, line 117: This can be const qualified. You may also want to document what the returned values indicate, so anyone not familiar with the interface doesn't need to go to the implementation to figure that out.

src/yuzu/main.cpp
lioncash: Yeah, this is fine for now.

src/core/hle/service/apm/interface.cpp
```diff
@@ -120,6 +121,7 @@ void APM_Sys::SetCpuBoostMode(Kernel::HLERequestContext& ctx) {
 
     LOG_DEBUG(Service_APM, "called, mode={:08X}", mode);
 
+    Settings::values.is_cpu_boosted = (static_cast<u32>(mode) == 1);
```
lioncash, line 124:
```suggestion
    Settings::values.is_cpu_boosted = mode == CpuBoostMode::Full;
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+        commands[player_index].push_back(command);
+        frameNo++;
+    }
+    LOG_INFO(Input, "TAS file loaded! {} frames", frameNo);
+}
+
+void Tas::WriteTasFile() {
+    LOG_DEBUG(Input, "WriteTasFile()");
+    std::string output_text = "";
+    for (int frame = 0; frame < (signed)record_commands.size(); frame++) {
```
lioncash, line 116:
```suggestion
    for (size_t frame = 0; frame < record_commands.size(); frame++) {
```

Seems to work fine here.

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+        if (!output_text.empty()) {
+            output_text += "\n";
+        }
+        TASCommand line = record_commands.at(frame);
```
lioncash, line 120:
```suggestion
        const TASCommand& line = record_commands[frame];
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+        output_text += std::to_string(frame) + " " + WriteCommandButtons(line.buttons) + " " +
+                       WriteCommandAxis(line.l_axis) + " " + WriteCommandAxis(line.r_axis);
+    }
+    size_t bytesWritten = Common::FS::WriteStringToFile(
```
lioncash, line 124:
```suggestion
    const size_t bytes_written = Common::FS::WriteStringToFile(
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+        };
+        commands[player_index].push_back(command);
+        frameNo++;
+    }
+    LOG_INFO(Input, "TAS file loaded! {} frames", frameNo);
+}
+
+void Tas::WriteTasFile() {
+    LOG_DEBUG(Input, "WriteTasFile()");
+    std::string output_text = "";
```
lioncash, line 115:
```suggestion
    std::string output_text;
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+    if (!commands[player_index].empty()) {
+        commands[player_index].clear();
+    }
+    std::string file = Common::FS::ReadStringFromFile(
+        Common::FS::GetYuzuPathString(Common::FS::YuzuPath::TASFile) + "script0-" +
+            std::to_string(player_index + 1) + ".txt",
+        Common::FS::FileType::BinaryFile);
+    std::stringstream command_line(file);
+    std::string line;
+    int frameNo = 0;
```
lioncash, line 76:
```suggestion
    int frame_no = 0;
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+    TASCommand empty = {.buttons = 0, .l_axis = {0.f, 0.f}, .r_axis = {0.f, 0.f}};
```
lioncash, line 77:
```suggestion
    TASCommand empty = {};
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+        state = TasState::RUNNING;
+    } else {
+        state = TasState::STOPPED;
+    }
+
+    return {state, current_command, script_length};
+}
+
+static std::string DebugButtons(u32 buttons) {
+    return "{ " + TasInput::Tas::ButtonsToString(buttons) + " }";
```
lioncash, line 159:
```suggestion
    return fmt::format("{{{}}}", Tas::ButtonsToString(buttons));
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+}
+
+static std::string DebugJoystick(float x, float y) {
+    return "[ " + std::to_string(x) + "," + std::to_string(y) + " ]";
```
lioncash, line 163:
```suggestion
    return fmt::format("[{}, {}]", x, y);
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+}
+
+static std::string DebugInput(const TasData& data) {
+    return "{ " + DebugButtons(data.buttons) + " , " + DebugJoystick(data.axis[0], data.axis[1]) +
```
lioncash, line 167:
```suggestion
    return fmt::format("{{{}, {}, {}}}", DebugButtons(data.buttons), DebugJoystick(data.axis[0], data.axis[1]),
                       DebugJoystick(data.axis[2], data.axis[3]));
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+    auto [x, y] = data;
+    std::string line;
+    line += std::to_string(static_cast<int>(x * 32767));
+    line += ";";
+    line += std::to_string(static_cast<int>(y * 32767));
+    return line;
+}
+
+std::string Tas::WriteCommandButtons(u32 data) const {
+    if (data == 0)
```
lioncash, line 287: This needs to be braced.

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+            }
+            Settings::values.tas_reset = false;
+            LoadTasFiles();
+            LOG_DEBUG(Input, "tas_reset done");
+        }
+        if (Settings::values.tas_enable) {
+            if ((signed)current_command < script_length) {
+                LOG_INFO(Input, "Playing TAS {}/{}", current_command, script_length);
+                size_t frame = current_command++;
+                for (size_t i = 0; i < PLAYER_NUMBER; i++) {
```
lioncash, line 214: This should probably be querying `commands.size()` otherwise this is a signed comparison mismatch

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,372 @@
...
+Tas::~Tas() {
+    update_thread_running = false;
+}
+
+void Tas::RefreshTasFile() {
+    refresh_tas_fle = true;
+}
+void Tas::LoadTasFiles() {
+    script_length = 0;
+    for (size_t i = 0; i < PLAYER_NUMBER; i++) {
```
lioncash, line 58:
```suggestion
    for (size_t i = 0; i < commands.size(); i++) {
```

lioncash:
Just to make it a top-level comment so it isn't missed by other reviewers:

Note that the input subsystem being housed inside the Settings struct is a hard blocker. The instance itself isn't a setting, and it seems like the Settings struct is being used as a way to house the input subsystem as a global variable so that it can be accessed from anywhere in the codebase.

This should be separated out in a way that doesn't require this kind of coupling.

src/common/fs/fs_paths.h
```diff
@@ -21,6 +21,7 @@
 #define SCREENSHOTS_DIR "screenshots"
 #define SDMC_DIR "sdmc"
 #define SHADER_DIR "shader"
+#define TAS_DIR "scripts"
```
Morph1984, line 24: This should likely be "tas" instead of "scripts"

src/yuzu/configuration/configure_filesystem.ui
```diff
@@ -219,6 +219,55 @@
...
+                    <width>60</width>
+                    <height>20</height>
+                  </size>
+                </property>
+              </spacer>
+            </item>
+            <item row="1" column="0" colspan="4">
+              <widget class="QCheckBox" name="tas_pause_on_load">
+                <property name="text">
+                  <string>Pause TAS execution during loads (SMO - 1.3)</string>
```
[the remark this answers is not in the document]
MonsterDruide1: A full new section, to be selected from the left side? Or just a new tab in one of the existing sections?

src/yuzu/main.cpp
MonsterDruide1: Isn't this already a member function? Sorry if I don't understand all of the terminology for C++, but as far as I understand that already is a member function, as it's not defined in the header.

Also, is `tr()` causing the translation?

src/input_common/tas/tas_input.cpp
MonsterDruide1: If an empty command is that easy to create, I can straight up remove that variable fully and just put the `{}` everywhere I'd use `empty`, correct?

src/input_common/tas/tas_input.cpp
MonsterDruide1: `commands.size()` is constant with `PLAYER_NUMBER`, so which is better to use here?

src/input_common/tas/tas_input.cpp
MonsterDruide1: (same as above)

src/yuzu/configuration/configure_filesystem.ui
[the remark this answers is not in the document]
Morph1984: Either that or a new window that can be opened from the menu bar

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,145 @@
...
+#include <mutex>
+#include <thread>
+
+#include "common/common_types.h"
+#include "core/frontend/input.h"
+#include "input_common/main.h"
+
+namespace TasInput {
+
+constexpr int PLAYER_NUMBER = 8;
```
Morph1984, line 17:
```suggestion
constexpr size_t PLAYER_NUMBER = 8;
```
This avoids the signed mis-match in previous review comments

src/yuzu/main.cpp
Morph1984: A member function means that it is a function declared in the GMainWindow class.

The `static` qualifier means this function's scope is within the main.cpp file itself, but not a member of the GMainWindow class.

Morph1984:
Also please avoid using C-style casts, use static_cast<> or use appropriate types to reduce the need for casting.

xdidx:
First of all, I thank you for this PR that interests me.
Then I want to let you know that I started looking at the Yuzu code yesterday, following the discovery of this PR. So I sincerely hope that my remarks will be relevant and not too off the mark.
I tried your version and a few things bothered me:
- When saving the "record.txt" file for the first time, "WriteStringToFile" returns 0 because "!IsFile(path)" 
1) I think it's because the file is not created if it doesn't exist, is this intentional? 

- If I create a recording and want to test it, I have to manually change the filename from "record.txt" to "script0-1.txt" in addition to changing the input, I find it really inconvenient:
2) Is it possible to automatically change the input to TAS at the start of reading and then put the input back on the one selected before reading?
3) Is it possible to request a file name to save the recording?
4) Is it possible to load a file from the explorer instead of just using "script0"?

Morph1984:
@xdidx 1. Is fixed in #6493

german77:
Changing the input automatically can be done. 
Originally recording would overwrite the first player file and reloading the tas was enough see your own recording. This behavior was changed with the usage of TAS editors since yuzu only reads files, it doesn't really edits them and recording again would destroy the previous file without previous warning.

This PR only adds the fundamental of the TAS feature. I'm sure in the future it can be polished to make it more user friendly for TAS editors and users

AntoSkate:
i made a smo coins autofarm with this, it works godd but it would be cool if you could put it in loop

src/yuzu/main.cpp
MonsterDruide1: So... what would be the correct solution here? Removing `static`? Adding it to the header file and prepending `GMainWindow::`?

src/yuzu/main.cpp
```diff
@@ -758,6 +762,12 @@ void GMainWindow::InitializeWidgets() {
         statusBar()->addPermanentWidget(label);
     }
 
+    tas_label = new QLabel();
+    tas_label->setObjectName(QStringLiteral("TASlabel"));
+    tas_label->setText(tr(""));
```
lioncash, line 767: This line doesn't need to be here. It's empty by default and there's no need for translators to translate an empty string

src/yuzu/main.cpp
lioncash: The latter. Making it a const member function of GMainWindow gives it a translation context so that `tr()` can translate the string.

You then apply `tr()` to each of the strings once its a member function.

src/yuzu/main.cpp
```diff
@@ -2865,12 +2895,36 @@ void GMainWindow::UpdateWindowTitle(const std::string& title_name,
...
 
+    if (Settings::values.tas_enable) {
+        auto [tas_status, current_tas_frame, total_tas_frames] =
+            input_subsystem->GetTas()->GetStatus();
+        tas_label->setText(tr("%1 TAS %2/%3")
+                               .arg(tr(GetTasStateDescription(tas_status).c_str()))
+                               .arg(current_tas_frame)
+                               .arg(total_tas_frames));
+    } else {
+        tas_label->setText(tr(""));
```
lioncash, line 2925:
```suggestion
        tas_label->clear();
```

src/yuzu/configuration/configure_input_player_widget.cpp
```diff
@@ -218,15 +218,31 @@ void PlayerControlPreview::UpdateInput() {
...
+        controller_callback.input(std::move(input));
     }
 
     if (mapping_active) {
         blink_counter = (blink_counter + 1) % 50;
     }
 }
 
+void PlayerControlPreview::SetCallBack(ControllerCallback callback_) {
+    controller_callback = callback_;
```
lioncash, line 243:
```suggestion
    controller_callback = std::move(callback_);
```

german77:
> i made a smo coins autofarm with this, it works godd but it would be cool if you could put it in loop

I added an option to loop the script

src/yuzu/main.cpp
```diff
@@ -2865,12 +2896,32 @@ void GMainWindow::UpdateWindowTitle(const std::string& title_name,
     }
 }
 
+const QString GMainWindow::GetTasStateDescription() {
```
lioncash, line 2899:
```suggestion
QString GMainWindow::GetTasStateDescription() const {
```

src/core/frontend/input.cpp
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#include "core/frontend/input.h"
+
+namespace Input {
+
+std::function<void(void)> tas_update_callback;
```
lioncash, line 9: Don't introduce global state to the frontend or core. We've worked hard to eliminate this sort of thing. This is also a hard blocker.

MonsterDruide1: Does this mean no global variables **at all**, or just not within `core/frontend`? So would a global variable within `Service::VI` be allowed?

lioncash: None at all

src/yuzu/main.h
```diff
@@ -315,6 +317,7 @@ private slots:
     QLabel* emu_speed_label = nullptr;
     QLabel* game_fps_label = nullptr;
     QLabel* emu_frametime_label = nullptr;
+    QLabel* tas_label;
```
Morph1984, line 320:
```suggestion
    QLabel* tas_label = nullptr;
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,381 @@
...
+    std::string output_text;
+    for (size_t frame = 0; frame < record_commands.size(); frame++) {
+        if (!output_text.empty()) {
+            output_text += "\n";
+        }
+        const TASCommand& line = record_commands[frame];
+        output_text += std::to_string(frame) + " " + WriteCommandButtons(line.buttons) + " " +
+                       WriteCommandAxis(line.l_axis) + " " + WriteCommandAxis(line.r_axis);
+    }
+    const size_t bytes_written = Common::FS::WriteStringToFile(
```
Morph1984, line 115:
```suggestion
    const auto bytes_written = Common::FS::WriteStringToFile(
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,381 @@
...
+        Common::FS::GetYuzuPathString(Common::FS::YuzuPath::TASDir) + file_name,
```
Morph1984, line 116:
```suggestion
        Common::FS::GetYuzuPath(Common::FS::YuzuPath::TASDir) / file_name,
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,381 @@
...
+            .l_axis = ReadCommandAxis(seglist.at(2)),
+            .r_axis = ReadCommandAxis(seglist.at(3)),
+        };
+        commands[player_index].push_back(command);
+        frame_no++;
+    }
+    LOG_INFO(Input, "TAS file loaded! {} frames", frame_no);
+}
+
+void Tas::WriteTasFile(std::string file_name) {
```
Morph1984, line 105: Please use a std::u8string
```suggestion
void Tas::WriteTasFile(std::u8string file_name) {
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,381 @@
...
+    }
+}
+
+void Tas::LoadTasFile(size_t player_index) {
+    if (!commands[player_index].empty()) {
+        commands[player_index].clear();
+    }
+    std::string file =
+        Common::FS::ReadStringFromFile(Common::FS::GetYuzuPathString(Common::FS::YuzuPath::TASDir) +
+                                           "script0-" + std::to_string(player_index + 1) + ".txt",
```
Morph1984, line 65:
```suggestion
        Common::FS::ReadStringFromFile(Common::FS::GetYuzuPath(Common::FS::YuzuPath::TASDir) /
                                           fmt::format("script0-{}.txt", player_index + 1),
```

src/common/fs/path_util.h
```diff
@@ -23,6 +23,7 @@ enum class YuzuPath {
     ScreenshotsDir, // Where yuzu screenshots are stored.
     SDMCDir,        // Where the emulated SDMC is stored.
     ShaderDir,      // Where shaders are stored.
+    TASDir,         // Where the current script file is stored.
```
Morph1984, line 26:
```suggestion
    TASDir,         // Where TAS scripts are stored.
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,381 @@
...
+}
+
+bool Tas::Record() {
+    is_recording = !is_recording;
+    return is_recording;
+}
+
+void Tas::SaveRecording(bool overwrite_file) {
+    if (!is_recording && !record_commands.empty()) {
+        WriteTasFile("record.txt");
```
Morph1984, line 310:
```suggestion
        WriteTasFile(u8"record.txt");
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,381 @@
...
+        if (overwrite_file) {
+            WriteTasFile("script0-1.txt");
```
Morph1984, line 312:
```suggestion
            WriteTasFile(u8"script0-1.txt");
```

src/yuzu/main.cpp
```diff
@@ -2871,12 +2908,32 @@ void GMainWindow::UpdateWindowTitle(const std::string& title_name,
...
+    auto [tas_status, current_tas_frame, total_tas_frames] = input_subsystem->GetTas()->GetStatus();
+    switch (tas_status) {
+    case TasInput::TasState::Running:
+        return tr("TAS state: Running %1/%2").arg(current_tas_frame).arg(total_tas_frames);
+    case TasInput::TasState::Recording:
+        return tr("TAS state: Recording %1").arg(total_tas_frames);
+    case TasInput::TasState::Stopped:
+        return tr("TAS state: Idle %1/%2").arg(current_tas_frame).arg(total_tas_frames);
+    default:
+        return tr("INVALID TAS STATE");
```
Morph1984, line 2921:
```suggestion
        return tr("TAS State: Invalid");
```

src/yuzu/configuration/config.cpp
```diff
@@ -515,6 +518,12 @@ void Config::ReadControlValues() {
     Settings::values.mouse_panning_sensitivity =
         ReadSetting(QStringLiteral("mouse_panning_sensitivity"), 1).toFloat();
 
+    Settings::values.tas_enable = ReadSetting(QStringLiteral("enable_tas"), false).toBool();
+    Settings::values.tas_loop = ReadSetting(QStringLiteral("loop_tas"), false).toBool();
+    Settings::values.tas_swap_controllers =
+        ReadSetting(QStringLiteral("swap_tas_controllers"), true).toBool();
+    Settings::values.pause_tas_on_load =
+        ReadSetting(QStringLiteral("tas_pause_on_load"), true).toBool();
```
Morph1984, line 526: Try to keep the setting name in `values` consistent with the setting itself

src/yuzu/configuration/configure_input_player_widget.cpp
```diff
@@ -220,13 +220,34 @@ void PlayerControlPreview::UpdateInput() {
...
         update();
+        ControllerInput input{};
+        input.changed = true;
+        input.axis_values[Settings::NativeAnalog::LStick] = {
+            axis_values[Settings::NativeAnalog::LStick].value.x(),
+            axis_values[Settings::NativeAnalog::LStick].value.y()};
+        input.axis_values[Settings::NativeAnalog::RStick] = {
+            axis_values[Settings::NativeAnalog::RStick].value.x(),
+            axis_values[Settings::NativeAnalog::RStick].value.y()};
+        input.button_values = button_values;
```
Morph1984, line 231: Use designated initializers

src/input_common/main.h
```diff
@@ -168,6 +179,18 @@ class InputSubsystem {
...
+    [[nodiscard]] TasButtonFactory* GetTasButtons();
+
+    /// Retrieves the underlying tas button handler.
+    [[nodiscard]] const TasButtonFactory* GetTasButtons() const;
+
+    /// Retrieves the underlying tas touch handler.
+    [[nodiscard]] TasAnalogFactory* GetTasAnalogs();
+
+    /// Retrieves the underlying tas touch handler.
+    [[nodiscard]] const TasAnalogFactory* GetTasAnalogs() const;
```
Morph1984, line 192: Seems like these comments were copy pasted incorrectly

src/yuzu/configuration/configure_input_player_widget.cpp
```diff
@@ -220,13 +220,34 @@ void PlayerControlPreview::UpdateInput() {
...
+            .changed = true,
+        };
+
+        if (controller_callback.input != nullptr) {
+            controller_callback.input(std::move(input));
+        }
+    }
+
+    if (controller_callback.update != nullptr) {
+        controller_callback.update(std::move(true));
```
lioncash, line 239:
```suggestion
        controller_callback.update(true);
```

The move does nothing here.

src/yuzu/configuration/configure_input_player_widget.cpp
```diff
@@ -220,13 +220,34 @@ void PlayerControlPreview::UpdateInput() {
 
     if (input_changed) {
         update();
+        ControllerInput input{
+            .axis_values =
+                {std::pair<float, float>{axis_values[Settings::NativeAnalog::LStick].value.x(),
```
lioncash, line 225: std::make_pair

Ditto for the one below

src/yuzu/configuration/configure_input_player_widget.cpp
```diff
@@ -220,13 +220,34 @@ void PlayerControlPreview::UpdateInput() {
 
     if (input_changed) {
         update();
+        ControllerInput input{
```
lioncash, line 223: Maybe move this into the if statement body below, so this is only put on the stack when it needs to be.

src/input_common/main.cpp
```diff
@@ -94,13 +103,23 @@ struct InputSubsystem::Impl {
...
+        tasanalog.reset();
     }
 
     [[nodiscard]] std::vector<Common::ParamPackage> GetInputDevices() const {
         std::vector<Common::ParamPackage> devices = {
             Common::ParamPackage{{"display", "Any"}, {"class", "any"}},
             Common::ParamPackage{{"display", "Keyboard/Mouse"}, {"class", "keyboard"}},
         };
+        if (Settings::values.tas_enable) {
+            devices.push_back(
```
Morph1984, line 120:
```suggestion
            devices.emplace_back(
```

src/yuzu/debugger/controller.cpp
```diff
@@ -67,3 +73,13 @@ void ControllerDialog::hideEvent(QHideEvent* ev) {
     widget->SetConnectedStatus(false);
     QWidget::hideEvent(ev);
 }
+
+void ControllerDialog::InputController(ControllerInput input) {
+    u32 buttons = 0;
+    int index = 0;
+    for (bool btn : input.button_values) {
+        buttons += (btn ? 1 : 0) << index;
```
Morph1984, line 81:
```suggestion
        buttons |= (btn ? 1 : 0) << index;
```

src/yuzu/main.cpp
```diff
@@ -1042,6 +1049,21 @@ void GMainWindow::InitializeHotkeys() {
...
+            &QShortcut::activated, this, [&] { input_subsystem->GetTas()->Reset(); });
+    connect(hotkey_registry.GetHotkey(main_window, QStringLiteral("TAS Record"), this),
+            &QShortcut::activated, this, [&] {
+                bool is_recording = input_subsystem->GetTas()->Record();
+                if (!is_recording) {
+                    QMessageBox::StandardButton reply;
+                    reply = QMessageBox::question(this, tr("TAS Recording"),
+                                                  tr("Overwrite file of player 1?"),
+                                                  QMessageBox::Yes | QMessageBox::No);
+                    input_subsystem->GetTas()->SaveRecording(reply == QMessageBox::Yes);
```
Morph1984, line 1064:
```suggestion
                    const auto res = QMessageBox::question(this, tr("TAS Recording"),
                                                  tr("Overwrite file of player 1?"),
                                                  QMessageBox::Yes | QMessageBox::No);
                    input_subsystem->GetTas()->SaveRecording(res == QMessageBox::Yes);
```

src/input_common/tas/tas_input.h
```diff
@@ -0,0 +1,233 @@
...
+    BUTTON_MINUS = 0x000800,
+    BUTTON_LEFT = 0x001000,
+    BUTTON_UP = 0x002000,
+    BUTTON_RIGHT = 0x004000,
+    BUTTON_DOWN = 0x008000,
+    BUTTON_SL = 0x010000,
+    BUTTON_SR = 0x020000,
+    BUTTON_HOME = 0x040000,
+    BUTTON_CAPTURE = 0x080000,
+};
```
Morph1984, line 79: You may want to convert these to bit shifting, for example:
```cpp
ButtonA = 1U << 0,
ButtonB = 1U << 1,
ButtonX = 1U << 2,
```
etc

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,431 @@
...
+            is_running = Settings::values.tas_loop.GetValue();
+            current_command = 0;
+            tas_data.fill({});
+            if (!is_running) {
+                SwapToStoredController();
+            }
+        }
+    } else {
+        tas_data.fill({});
+    }
```
Morph1984, line 227: This can be inverted to unindent the larger block inside the if

```cpp
if (!running) {
    tas_data.fill({});
    return;
}
```

src/input_common/tas/tas_input.cpp
```diff
@@ -0,0 +1,431 @@
...
+    std::stringstream linestream(line);
+    std::string segment;
+    std::vector<std::string> seglist;
+
+    while (std::getline(linestream, segment, ';')) {
+        seglist.push_back(segment);
+    }
+
+    const float x = std::stof(seglist.at(0)) / 32767.f;
+    const float y = std::stof(seglist.at(1)) / 32767.f;
```
Morph1984, line 241:
```suggestion
    const float x = std::stof(seglist.at(0)) / 32767.0f;
    const float y = std::stof(seglist.at(1)) / 32767.0f;
```

bunnei:
Untagging because this conflicts with other PRs, will retag once we get those merged.

src/core/frontend/input.cpp
MonsterDruide1: Fixed for now with the [Fallback to simple update](https://github.com/yuzu-emu/yuzu/pull/6485/commits/78a0c98b33f8362af0ee97536584e78f86e05720) commit. It will be more likely to desync with this method of updating, so this PR won't be the end of the journey.

MiYakuGaming:
Is it possible to get a recording to be continued from where it last left off?

MonsterDruide1:
> Is it possible to get a recording to be continued from where it last left off?

No, not as simple. You could move the first file to a different file/name (so that it doesn't get overridden), then record a new file starting at the desired location (using playback to reach it?), and finally merge both files using external tools as [TAS-Editor](https://GitHub.com/Jadefalke2/TAS-Editor).

german77:
It's definitively posible but not practical. The TAS is blind to anything and it depends heavily on timing to work correctly. Pausing just breaks all the timing. Ignoring the fact that if you move while it's paused it will be really difficult to start from the initial point.

src/yuzu/configuration/configure_tas.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    switch (target) {
+    case DirectoryTarget::TAS:
+        caption = tr("Select TAS Load Directory...");
+        break;
+    }
+
+    QString str = QFileDialog::getExistingDirectory(this, caption, edit->text());
+
+    if (str.isNull() || str.isEmpty()) {
```
lioncash, line 58:
```suggestion
    if (str.isEmpty()) {
```

Is sufficient for this case

src/yuzu/configuration/configure_tas.ui
```diff
@@ -0,0 +1,194 @@
...
+          <item>
+            <widget class="QGroupBox" name="groupBox">
+              <property name="title">
+                <string>Tool Assisted Speedrun tool</string>
+              </property>
+              <layout class="QGridLayout" name="gridLayout">
+                <item row="0" column="0" colspan="1">
+                  <widget class="QLabel" name="label_1">
+                    <property name="text">
+                      <string>This tool reads controller input from a script files under the same format as TAS-nx files. For more delailed explanation please consult the FAQ on the yuzu website.</string>
```
lioncash, line 28:
```suggestion
                      <string>Reads controller input from scripts in the same format as TAS-nx scripts. For a more detailed explanation please consult the FAQ on the yuzu website.</string>
```

src/yuzu/configuration/configure_tas.ui
```diff
@@ -0,0 +1,194 @@
...
+    <property name="windowTitle">
+      <string>Dialog</string>
+    </property>
+    <layout class="QVBoxLayout" name="verticalLayout_1">
+      <item>
+        <layout class="QHBoxLayout" name="horizontalLayout">
+          <item>
+            <widget class="QGroupBox" name="groupBox">
+              <property name="title">
+                <string>Tool Assisted Speedrun tool</string>
```
lioncash, line 22:
```suggestion
                <string>TAS</string>
```

src/yuzu/configuration/configure_tas.ui
```diff
@@ -0,0 +1,194 @@
...
+                  <widget class="QLabel" name="label_3">
+                    <property name="text">
+                      <string>Ctrl + F5: Start or stop TAS bot</string>
+                    </property>
+                  </widget>
+                </item>
+                <item row="3" column="0" colspan="1">
+                  <widget class="QLabel" name="label_4">
+                    <property name="text">
+                      <string>Ctrl + F6: Reload file or restart bot from the beggining</string>
```
lioncash, line 55:
```suggestion
                      <string>Ctrl + F6: Reload file or restart from the beginning</string>
```

src/yuzu/configuration/configure_tas.ui
```diff
@@ -0,0 +1,194 @@
...
+                    </property>
+                    <property name="wordWrap">
+                      <bool>true</bool>
+                    </property>
+                  </widget>
+                </item>
+                <item row="2" column="0" colspan="1">
+                  <widget class="QLabel" name="label_3">
+                    <property name="text">
+                      <string>Ctrl + F5: Start or stop TAS bot</string>
```
lioncash, line 48:
```suggestion
                      <string>Ctrl + F5: Start or stop TAS</string>
```

OZtistic:
Probably a bad idea to use static text in the TAS Configuration dialog (configure_tas.ui) describing the Hotkey combos, when these keys can be changed in **_Emulation --> Configure --> General --> Hotkeys_** and it won't be updated to reflect the changes on the fly
https://github.com/yuzu-emu/yuzu/blob/7234a7dec5bbb36af36bf2bbc9d8b5bcdecfaefc/src/yuzu/configuration/configure_tas.ui#L48
https://github.com/yuzu-emu/yuzu/blob/7234a7dec5bbb36af36bf2bbc9d8b5bcdecfaefc/src/yuzu/configuration/configure_tas.ui#L55
https://github.com/yuzu-emu/yuzu/blob/7234a7dec5bbb36af36bf2bbc9d8b5bcdecfaefc/src/yuzu/configuration/configure_tas.ui#L62

I would propose simply referring users to the Hotkeys page for the key combos, or having the text update with the current bindings set in General --> Hotkeys for the TAS keys

OZtistic:
https://github.com/yuzu-emu/yuzu/blob/7234a7dec5bbb36af36bf2bbc9d8b5bcdecfaefc/src/yuzu/configuration/configure_tas.ui#L120
This can be changed to simply _Directory_

Ditto for this:
https://github.com/yuzu-emu/yuzu/blob/7234a7dec5bbb36af36bf2bbc9d8b5bcdecfaefc/src/yuzu/configuration/configure_tas.ui#L76
to simply _Settings_

MonsterDruide1:
@OZtistic Implemented all of those. Thanks for your suggestions!

Emayeah:
> It's definitively posible but not practical. The TAS is blind to anything and it depends heavily on timing to work correctly. Pausing just breaks all the timing. Ignoring the fact that if you move while it's paused it will be really difficult to start from the initial point.

someone should port the citra tas thing to yuzu, this relies too much on timings, and any game changing rng (like for example Mario kart ? boxes or enemies) break the stuff. I am still curious how the citra tas implementation works tho...

PR #6487, opened 2021-06-18
common/fs: Fix file write/append access when file does not exist.

ameerj:
This fixes errors when a file is being opened for write/append access, but the file does not exist initially.

These functions were not respecting the FileAccessMode which specify that the file will be created if it does not exist.

### DIFF
--- a/src/common/fs/file.cpp
+++ b/src/common/fs/file.cpp
@@ -172,23 +172,13 @@ std::string ReadStringFromFile(const std::filesystem::path& path, FileType type)
 
 size_t WriteStringToFile(const std::filesystem::path& path, FileType type,
                          std::string_view string) {
-    if (!IsFile(path)) {
-        return 0;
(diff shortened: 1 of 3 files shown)


src/common/fs/file.cpp
```diff
@@ -172,23 +172,13 @@ std::string ReadStringFromFile(const std::filesystem::path& path, FileType type)
...
-    if (!IsFile(path)) {
-        return 0;
-    }
-
     IOFile io_file{path, FileAccessMode::Write, type};
-
     return io_file.WriteString(string);
 }
 
 size_t AppendStringToFile(const std::filesystem::path& path, FileType type,
```
lioncash, line 179: The documentation comments should be updated to indicate what the new behavior is for both of these functions

bunnei:
Closing because this was superseded by #6493

PR #6488, opened 2021-06-19
configure_input_player_widget: Mark variables as volatile

lat9nq:
A GCC bug or ill-behaved optimization causes this to act strangely on Arch Linux's MinGW GCC compiler (both GCC 10 and 11). Mark it as volatile so the compiler avoids more optimizations here.

### DIFF
--- a/src/yuzu/configuration/configure_input_player_widget.cpp
+++ b/src/yuzu/configuration/configure_input_player_widget.cpp
@@ -2505,8 +2505,8 @@ void PlayerControlPreview::DrawArrowButton(QPainter& p, const QPointF center,
     QPoint offset;
 
     for (std::size_t point = 0; point < up_arrow_button.size() / 2; ++point) {
-        const float up_arrow_x = up_arrow_button[point * 2 + 0];
-        const float up_arrow_y = up_arrow_button[point * 2 + 1];
+        const volatile float up_arrow_x = up_arrow_button[point * 2 + 0];
+        const volatile float up_arrow_y = up_arrow_button[point * 2 + 1];
 
         switch (direction) {
         case Direction::Up:


lioncash:
This should probably be commented explaining why, and maybe find the report for GCC (if any exist, to link to this).

lioncash:
Does UBSan report anything when executing this code, out of curiosity?

lat9nq:
> Does UBSan report anything when executing this code without volatile qualifiers, out of curiosity?

Nope, UBSan didn't catch anything. Granted, I could only test this with a native executable -- there's no UBSan with the MinGW compiler Arch provides.



> This should probably be commented explaining why, and maybe find the report for GCC (if any exist, to link to this).

I can add a comment, but I don't really know what causes this or how to look for a report of this nature.

lioncash:
I mean, we should investigate and understand *why* this is happening instead of marking it volatile and forgetting about it. Especially when it impacts other platforms.

Does reorganizing the data into a std::array of std::pairs fix it?

lat9nq:
Changing it to `std::pair`s didn't work, but passing the raw floats in instead of the `QPointF center` did fix it.

To give you an idea of the issue:


Which makes this change even more puzzling, since the left dpad button is the second one drawn in the cluster.

lat9nq:
Also, the left dpad button is consistently offset in both Windows and Wine, and setting the center to 0 makes the dpad draw offscreen, so the problem offset isn't 0.

lat9nq:
`-O2` doesn't have this behavior. Should mention that earlier I tried adding `LOG_DEBUG(Frontend, "({:f}, {:f})", up_arrow_x, up_arrow_y);` after setting `up_arrow_y`, but that also hid the behavior.

v1993:
I know it's somewhat easier said than done, but perhaps look at assembly of troublesome section? This will at least show what code GCC outputs and probably give some clues.



> Should mention that earlier I tried adding `LOG_DEBUG(Frontend, "({:f}, {:f})", up_arrow_x, up_arrow_y);` after setting `up_arrow_y`, but that also hid the behavior.

Explicit access probably made the troublesome optimization impossible, so it wasn't done. What happens if you print only one variable?

PR #6493, opened 2021-06-19
common: fs: Miscellaneous changes

Morph1984:
Remove [[nodiscard]] attribute on Remove* functions (RemoveFile, RemoveDir, RemoveDirRecursively, RemoveDirContentsRecursively)
There are a lot of scenarios where we don't particularly care whether or not the removal operation and just simply attempt a removal.

Remove [[nodiscard]] attribute from Flush
Similarly, Flush() is typically called to attempt to flush a file into the disk. In the one case where this is used, we do not care whether the flush has succeeded or not, making [[nodiscard]] unnecessary.

While we're at it, we can also Remove IsFile checks in FileOpen/(Write/Append)StringToFile
These checks were preventing files with the Write or Append file access modes from being created, as per the documented behavior in FileAccessMode. This removes the check as the file open functions within IOFile::Open implicitly check for these.

Due to this, we will have to fix the Mode to FileAccessMode conversion to enforce requiring the file to exist prior to opening.

Supersedes #6487


ameerj:
> Supersedes #6487

Not sure why this is the case considering your changes relating to my PR are not meaningfully different than the changes that I've made. You could have more easily left a review on my PR rather than make the same changes yourself.

bunnei (approved):

Issue #4791, opened 2020-10-14
Kirby Star Allies - broken sound effects


v1993:
Some of audio effects are broken in this game: they play with extreme crackling - so bad it's essentially impossible to hear effect itself or sometimes don't play at all. Easiest one to reproduce (there are more): going through mid-level door, health gauge of boss/miniboss filling up.

It appears to be a regression. Some time ago (in June-July IIRC) various other sound effects weren't playing, but those that are broken now worked properly.

FearlessTobi:
Missing a log file.

Moonlacer:
Should be able to be closed now @v1993, unless you have had the same issue even now

v1993:
Still wasn't able to test, but most likely fixed. Closing.

PR #6498, opened 2021-06-20
[audio_core] Decouple audio update and processing, and process at variable rate

Kelebek1:
Currently, processing of audio samples is called from AudioRenderer's Update method, using a fixed 4 buffers to process the given samples. Games call Update at variable rates, depending on framerate and/or sample count, which causes inconsistency in audio processing. From what I've seen, 60 FPS games update every ~0.004s, but 30 FPS/160 sample games update somewhere between 0.02 and 0.04, 5-10x slower. Not enough samples get fed to the backend, leading to a lot of audio skipping.

This PR seeks to address this by de-coupling the audio consumption and the audio update. Update remains the same without calling for buffer queuing, and the consume now schedules itself to run based on the sample rate and count.

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -12,6 +12,7 @@
 #include "audio_core/voice_context.h"
 #include "common/logging/log.h"
 #include "common/settings.h"
+#include "core/core_timing.h"
 #include "core/memory.h"
 
 namespace {
@@ -68,7 +69,9 @@ namespace {
 } // namespace
 
 namespace AudioCore {
-AudioRenderer::AudioRenderer(Core::Timing::CoreTiming& core_timing, Core::Memory::Memory& memory_,
+constexpr s32 NUM_BUFFERS = 2;
+
+AudioRenderer::AudioRenderer(Core::Timing::CoreTiming& core_timing_, Core::Memory::Memory& memory_,
                              AudioCommon::AudioRendererParameter params,
                              Stream::ReleaseCallback&& release_callback,
                              std::size_t instance_number)
@@ -77,7 +80,8 @@ AudioRenderer::AudioRenderer(Core::Timing::CoreTiming& core_timing, Core::Memory
       sink_context(params.sink_count), splitter_context(),
       voices(params.voice_count), memory{memory_},
       command_generator(worker_params, voice_context, mix_context, splitter_context, effect_context,
-                        memory) {
+                        memory),
+      core_timing{core_timing_} {
     behavior_info.SetUserRevision(params.revision);
     splitter_context.Initialize(behavior_info, params.splitter_count,
                                 params.num_splitter_send_channels);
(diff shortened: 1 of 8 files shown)


FernandoS27:
This seems to not only fix UE4 games audio but also all known audio softlocks like in games such as DKTF & XC2

src/audio_core/audio_renderer.cpp
```diff
@@ -315,10 +327,24 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
...
+
+    {
+        std::scoped_lock l{lock};
+        const auto released_buffers{audio_out->GetTagsAndReleaseBuffers(stream)};
+        for (const auto& tag : released_buffers) {
+            QueueMixedBuffer(tag);
+        }
     }
+
+    const f32 sampleRate = static_cast<f32>(GetSampleRate());
```
lioncash, line 342:
```suggestion
    const f32 sample_rate = static_cast<f32>(GetSampleRate());
```

Ditto for other variables below

src/audio_core/audio_renderer.h
```diff
@@ -33,6 +34,7 @@ class Memory;
 
 namespace AudioCore {
 using DSPStateHolder = std::array<VoiceState*, AudioCommon::MAX_CHANNEL_COUNT>;
+constexpr s32 NUM_BUFFERS = 2;
```
lioncash, line 37: This can be moved into the cpp file, since it's an internal detail.

src/audio_core/audio_renderer.cpp
```diff
@@ -86,16 +88,27 @@ AudioRenderer::AudioRenderer(Core::Timing::CoreTiming& core_timing, Core::Memory
...
-    audio_out->StartStream(stream);
-
-    QueueMixedBuffer(0);
-    QueueMixedBuffer(1);
-    QueueMixedBuffer(2);
-    QueueMixedBuffer(3);
+    process_event = Core::Timing::CreateEvent(
+        fmt::format("AudioRenderer-Instance{}-Consume", instance_number),
+        [this](std::uintptr_t, std::chrono::nanoseconds) { ReleaseAndQueueBuffers(); });
+    for (size_t i = 0; i < NUM_BUFFERS; ++i) {
```
lioncash, line 94: This will result in a sign comparison mismatch warning, since `i` is unsigned, but `NUM_BUFFERS` is signed.

src/audio_core/audio_renderer.h
```diff
@@ -68,6 +72,9 @@ class AudioRenderer {
     Core::Memory::Memory& memory;
     CommandGenerator command_generator;
     std::size_t elapsed_frame_count{};
+    Core::Timing::CoreTiming& core_timing;
+    std::shared_ptr<Core::Timing::EventType> process_event;
+    std::mutex lock;
```
lioncash, line 77: this should probably be given a slightly better name, given it's a mutex, and there's a convention in the codebase for `lock`, `l`, `lck`, etc to be given to the `scoped_lock`, `lock_guard` or whatever other locking type we have that actually does the locking.

lioncash:
Can you also add the PR description into the first commit in the changes as well, so that the information is also stored in the git repository once the PR is merged?

v1993:
Don't have time to test now, but likely fixes #4791.

src/audio_core/audio_renderer.cpp
```diff
@@ -114,89 +129,88 @@ Stream::State AudioRenderer::GetStreamState() const {
 
 ResultCode AudioRenderer::UpdateAudioRenderer(const std::vector<u8>& input_params,
                                               std::vector<u8>& output_params) {
+    {
```
ReinUsesLisp, line 132: This extra scope is not needed.

ameerj:
Seems to cause regressions in DKCTF and NSMBU, where audio can get cut off and not return.

Kelebek1:
Regressions should now be fixed.

bunnei (approved):

PR #6499, opened 2021-06-20
Update dynarmic and add new unsafe CPU option.

FernandoS27:
This PR updates dynarmic and adds an option to loose the rounding mode on ASIMD instructions (32 bits).

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -122,6 +122,7 @@ void RestoreGlobalState(bool is_powered_on) {
     values.cpu_accuracy.SetGlobal(true);
     values.cpuopt_unsafe_unfuse_fma.SetGlobal(true);
     values.cpuopt_unsafe_reduce_fp_error.SetGlobal(true);
+    values.cpuopt_unsafe_ignore_standard_fpcr.SetGlobal(true);
     values.cpuopt_unsafe_inaccurate_nan.SetGlobal(true);
     values.cpuopt_unsafe_fastmem_check.SetGlobal(true);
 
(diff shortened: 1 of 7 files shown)


v1993:
I find this PR rather confusing right now. Does 32-bit refer to host architecture (ok, it does not, but end users probably won't know that)? Guest architecture? Size of values instruction operates on?

If it refers to instruction arguments, then `(32 bits only)` is only a pretty technical detail that can be omitted. If it refers to guest architecture - doesn't Yuzu lack support for 32-bit guests as for now?

P.S.: where I can get stuff you use to come up with branch names? It seems pretty good...

MerryMage:
> Yuzu lack support for 32-bit guests as for now?

Yuzu supports 32-bit guests.

v1993:
Ah, thanks for correcting me then. So, does this option affect only 32-bit guests? Also, are 32-bit guests common in Yuzu?

MerryMage:
This option only affects 32-bit guests, because only 32-bit ASIMD instructions have the concept of "default FPCR", which may by different from the current FPCR. Switching rounding modes for this can be expensive.

v1993:
Thanks a lot for explanation! I'm suggesting to go with this description then:
```
Inexact rounding for ASIMD instructions in 32-bit applications
This option improves speed of 32-bit ASIMD floating-point instructions by skipping rounding mode changes.
```

bunnei (approved):

v1993:
@bunnei Uh, what do you think about my rewording suggestion? Should I send it as follow-up PR or something?

PR #6501, opened 2021-06-20
nvhost-as-gpu: correct Remap method when no object is present.

FernandoS27:
Special thanks to ByLaws from Skyline-emu for pointing this out.

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
@@ -146,6 +146,14 @@ NvResult nvhost_as_gpu::Remap(const std::vector<u8>& input, std::vector<u8>& out
         LOG_DEBUG(Service_NVDRV, "remap entry, offset=0x{:X} handle=0x{:X} pages=0x{:X}",
                   entry.offset, entry.nvmap_handle, entry.pages);
 
+        if (entry.nvmap_handle == 0) {
+            // If nvmap handle is null, we should unmap instead.
+            const auto offset{static_cast<GPUVAddr>(entry.offset) << 0x10};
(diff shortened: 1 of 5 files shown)


src/core/hle/service/nvdrv/devices/nvhost_as_gpu.cpp
```diff
@@ -146,6 +146,15 @@ NvResult nvhost_as_gpu::Remap(const std::vector<u8>& input, std::vector<u8>& out
         LOG_DEBUG(Service_NVDRV, "remap entry, offset=0x{:X} handle=0x{:X} pages=0x{:X}",
                   entry.offset, entry.nvmap_handle, entry.pages);
 
+        if (entry.nvmap_handle == 0) {
+            // If nvmap handle is null, we should unmap instead.
+            // Credits to ByLaws from Skyline-emu for pointing this out.
```
v1993, line 151: How about moving credits to commit message instead? I don't think this commit will help future readers much.

FernandoS27:
It seems there are other underlying issues in the gpu memory manager. I need to check this.

src/video_core/memory_manager.h
```diff
@@ -155,8 +163,22 @@ class MemoryManager final {
...
+        }
+
+        bool operator==(const MemoryRegion&) const = default;
+        bool operator!=(const MemoryRegion&) const = default;
+
+        GPUVAddr address{};
+        size_t size{};
+    };
+
+    std::set<MemoryRegion> map_ranges;
```
v1993, line 181: Have you considered using `unordered_set` (it may or may not have better performance)? It should be trivial to make hash function for `MemoryRegion` since it's just two integers and profile to see if there's any difference.

FernandoS27: unordered_set is unordered, I need it to be ordered :).

PR #6502, opened 2021-06-21
main: Add GPU Vendor name to running title bar

ameerj:
Adds the GPU driver vendor to the title bar of yuzu while running a title, making it easier to confirm which driver is currently being used as the renderer.

Currently the mappings are for the most common drivers and drivers which benefit from being clearly identified (i.e. the OPENSWR software renderer, which misleadingly reports the vendor name as `Intel Corporation`)

Huge thanks to @lat9nq for confirming and populating Linux driver mappings.

Example screenshot:
image

### DIFF
--- a/src/video_core/renderer_base.h
+++ b/src/video_core/renderer_base.h
@@ -42,6 +42,8 @@ public:
 
     [[nodiscard]] virtual RasterizerInterface* ReadRasterizer() = 0;
 
+    [[nodiscard]] virtual std::string GetDeviceVendor() const = 0;
+
     // Getter/setter functions:
     // ------------------------
 
--- a/src/video_core/vulkan_common/vulkan_device.h
+++ b/src/video_core/vulkan_common/vulkan_device.h
@@ -45,6 +45,9 @@ public:
     /// Reports a shader to Nsight Aftermath.
     void SaveShader(const std::vector<u32>& spirv) const;
 
+    /// Returns the name of the VkDriverId reported from Vulkan.
+    std::string GetDriverName() const;
+
     /// Returns the dispatch loader with direct function pointers of the device.
     const vk::DeviceDispatch& GetDispatchLoader() const {
         return dld;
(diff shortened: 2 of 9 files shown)


src/video_core/renderer_base.h
```diff
@@ -42,6 +42,8 @@ class RendererBase : NonCopyable {
 
     [[nodiscard]] virtual RasterizerInterface* ReadRasterizer() = 0;
 
+    [[nodiscard]] virtual std::string_view GetDeviceVendor() = 0;
```
lioncash, line 45:
```suggestion
    [[nodiscard]] virtual std::string_view GetDeviceVendor() const = 0;
```

Perhaps this should return a std::string as opposed to a std::string, given it's in a virtual interface, to ensure no renderer potentially returns a dangling string, regardless of any refactoring the takes place in the future

src/video_core/vulkan_common/vulkan_device.h
```diff
@@ -45,6 +45,9 @@ class Device {
     /// Reports a shader to Nsight Aftermath.
     void SaveShader(const std::vector<u32>& spirv) const;
 
+    /// Returns the name of the VkDriverId reported from Vulkan.
+    std::string_view GetDriverIdName() const;
```
lioncash, line 49:
```suggestion
    std::string_view GetDriverIDName() const;
```

Given we have other functions in the interface with it uppercased

src/yuzu/main.cpp
```diff
@@ -1418,7 +1419,7 @@ void GMainWindow::BootGame(const QString& filename, std::size_t program_index, S
             std::filesystem::path{filename.toStdU16String()}.filename());
     }
     LOG_INFO(Frontend, "Booting game: {:016X} | {} | {}", title_id, title_name, title_version);
-    UpdateWindowTitle(title_name, title_version);
+    UpdateWindowTitle(title_name, title_version, system.GPU().Renderer().GetDeviceVendor());
```
ReinUsesLisp, line 1422: Call GetDeviceVendor once and store it in a local variable

Morph1984 (approved):

PR #6514, opened 2021-06-22
Simple resizing of Per-Game configuration window and removal of useless Help question mark button in the title bar

OZtistic:
Corrects window sizing of Per-Game configuration window and removes the unused Help hint button from the title bar
image

### DIFF
--- a/src/yuzu/configuration/configure_per_game.ui
+++ b/src/yuzu/configuration/configure_per_game.ui
@@ -6,9 +6,14 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>800</width>
+    <width>900</width>
     <height>600</height>
    </rect>
+  </property>
+  <property name="minimumSize">
+   <size>
+    <width>900</width>
+   </size>
   </property>
   <property name="windowTitle">
    <string>Dialog</string>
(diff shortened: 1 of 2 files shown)


german77:
You might want to add this property after the geometry property to prevent this issue in all cases 

```
  <property name="minimumSize">
   <size>
    <width>840</width>
   </size>
  </property>
```

src/yuzu/configuration/configure_per_game.ui
```diff
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>800</width>
+    <width>840</width>
```
lat9nq, line 9:
```suggestion
    <width>900</width>
```

Can you set this to 900 instead? Needs to be slightly larger to work with the default Qt theme in Linux, or at least in general it'd work better with larger fonts.

I already made this change in hades, but if you could do that here the larger window size would get out faster.

OZtistic: should be fixed now

german77 (approved):

Morph1984 (approved):
nice

PR #6515, opened 2021-06-23
input_common: Fix data race on GC implementation

german77:
Use jthread instead of std::thread. Credits to @ReinUsesLisp I only fixed hotplug issues


src/input_common/gcadapter/gc_adapter.cpp
```diff
@@ -39,17 +97,16 @@ Adapter::~Adapter() {
     Reset();
```
ReinUsesLisp, line 97: If we properly order member variable decarations, we can avoid defining a destructor.

src/input_common/gcadapter/gc_adapter.h
```diff
@@ -164,5 +156,6 @@ class Adapter {
     bool configuring{false};
     bool rumble_enabled{true};
     bool vibration_changed{true};
+    bool adapter_input_thread_running{false};
```
ReinUsesLisp, line 159: This doesn't seem to be used. It's only set to true once and read within the same thread.

german77: I use it to stop the main loop cleanly. Before you where destroying the thread from the thread. Locking the system and causing yuzu to crash

ReinUsesLisp: If that's the case, this should be a jthread.request_stop() instead.

lat9nq:
After building locally with GCC 10 and libusb dynamically linked to the executable, this PR causes a regression during startup when a GC adapter is attached to the machine: 
[gdb.txt](https://github.com/yuzu-emu/yuzu/files/6712380/gdb.txt)

PR #6517, opened 2021-06-23
externals: Update fmt to 8.0.0

lioncash:
Keeps us up to date with the latest major release.

Also allows compilers that support it to perform compile-time format string checking, allowing format string errors to be caught sooner.

8.0.0 introduces quite a bunch of changes, so the entire changelog can be viewed [here](https://github.com/fmtlib/fmt/releases/tag/8.0.0).


lioncash:
Currently blocked on https://github.com/MerryMage/dynarmic/pull/629, will update this once merged.

bunnei (approved):

PR #6519, opened 2021-06-23
common: Replace common_sizes into user-literals

Wunkolo:
Removes `common_sizes.h` in favor of having `_KiB`, `_MiB`, `_GiB`, etc
user-literals within `common_funcs.h`.

To keep the global namespace clean, users will have to use:
```cpp
using namespace Common::Literals;
```
to access these literals.

Replaced some memory-constants here and there to the new user-literals.

Ex:

```diff
-    constexpr std::size_t MAX_BYTES_WRITTEN = 100 * 1024 * 1024;
-    constexpr std::size_t MAX_BYTES_WRITTEN_EXTENDED = 1024 * 1024 * 1024;
+   constexpr std::size_t MAX_BYTES_WRITTEN = 100_MiB;
+   constexpr std::size_t MAX_BYTES_WRITTEN_EXTENDED = 1024_MiB;
```


bunnei (approved):

src/common/common_funcs.h
```diff
@@ -126,4 +126,26 @@ constexpr std::size_t Size(const C& c) {
     }
 }
 
+namespace Literals {
```
lioncash, line 129: These should still remain inside common_sizes, so we don't need to drag in unrelated things just to get custom literals.

src/common/common_funcs.h
```diff
@@ -126,4 +126,26 @@ constexpr std::size_t Size(const C& c) {
...
+constexpr u64 operator""_KiB(unsigned long long int x) {
+    return 1024ull * x;
```
lioncash, line 131: We use uppercase suffixing to clearly differentiate between 1 and l in fonts and prevent misreads.

```suggestion
    return 1024ULL * x;
```

src/core/hle/kernel/k_address_space_info.cpp
```diff
@@ -5,34 +5,37 @@
...
+#include "common/common_funcs.h"
 #include "core/hle/kernel/k_address_space_info.h"
 
 namespace Kernel {
 
 namespace {
 
+using namespace Common::Literals;
+
+constexpr u64 Size_Invalid = ~u64(0);
```
lioncash, line 17:
```suggestion
constexpr u64 Size_Invalid = UINT64_MAX;
```

src/common/common_funcs.h
```diff
@@ -126,4 +126,26 @@ constexpr std::size_t Size(const C& c) {
     }
 }
 
+namespace Literals {
+constexpr u64 operator""_KiB(unsigned long long int x) {
```
Morph1984, line 130: Shouldn't `unsigned long long int` just be condensed into `u64`

lioncash: I believe this is because, of the handful of types that the C++ standard allows on user-defined literals, the only regular integral type allowed is `unsigned long long`.

Depending on the implementation, implementors may choose to implement `uint64_t` as an alias to `unsigned long` *or* `unsigned long long`, given they're the same size on LP64 platforms; so using `uint64_t` or `u64` here can cause a compilation failure if the implementation happens to implement `uint64_t` as an alias to `unsigned long` (which I believe either libstdc++ or libc++ actually do this).

src/common/common_funcs.h
Wunkolo: Maybe `literals.h` rather than `common_sizes.h`? These new literals effectively remove the need to have "Common Sizes" and maybe other future user-literals can be in this new header.

lioncash: Sounds good!

src/common/literals.h
```diff
@@ -0,0 +1,32 @@
+// Copyright 2021 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+
+namespace Common {
```
lioncash, line 9: Can't this coalesce with the `Literals` namespace into `Common::Literals`

lioncash (commented):
literals.h should also be added to the CMakeLists file.

Aside from that and a nit, this looks good.

src/common/literals.h
Wunkolo: Yeah, makes it much more concise to just do `namespace Common::Literals { ...`

lioncash (approved):

src/common/logging/backend.cpp
```diff
@@ -169,10 +171,11 @@ FileBackend::FileBackend(const std::filesystem::path& filename) {
 FileBackend::~FileBackend() = default;
 
 void FileBackend::Write(const Entry& entry) {
+    using namespace Common::Literals;
     // prevent logs from going over the maximum size (in case its spamming and the user doesn't
     // know)
-    constexpr std::size_t MAX_BYTES_WRITTEN = 100 * 1024 * 1024;
-    constexpr std::size_t MAX_BYTES_WRITTEN_EXTENDED = 1024 * 1024 * 1024;
+    constexpr std::size_t MAX_BYTES_WRITTEN = 100_MiB;
+    constexpr std::size_t MAX_BYTES_WRITTEN_EXTENDED = 1024_MiB;
```
Morph1984, line 178: or 1 GiB 🙂

src/common/literals.h
```diff
@@ -0,0 +1,31 @@
...
+
+constexpr u64 operator""_TiB(unsigned long long int x) {
+    return 1024_GiB * x;
+}
+
+constexpr u64 operator""_PiB(unsigned long long int x) {
+    return 1024_TiB * x;
+}
+
+} // namespace Common::Literals
```
Morph1984, line 31: Missing newline at eof

PR #6521, opened 2021-06-24
Add missing includes

ogniK5377:
Add missing includes to fix any compile errors

### DIFF
--- a/src/audio_core/sink_context.h
+++ b/src/audio_core/sink_context.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <vector>
 #include "audio_core/common.h"
 #include "common/common_funcs.h"
 #include "common/common_types.h"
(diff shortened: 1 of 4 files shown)


src/audio_core/sink_context.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <vector>
```
lioncash, line 7: Technically `<array>` should also be added here as well, since it's used below as well.

src/core/hle/service/mii/manager.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <vector>
```
lioncash, line 7: Ditto

lioncash:
LGTM otherwise

lioncash (approved):

PR #6523, opened 2021-06-24
[audio_core] Skip in-use voices with 0 channels

Kelebek1:
This avoids null pointer dereferences and prevents crashing in Xenoblade Chronicles 2. 

Needs more investigation as to why this is occuring as I don't think it should be, possibly something wrong with incoming voice parameter data.


lioncash:
This seems like a hack, which makes me skeptical about merging this. Particularly given this papers over the issue and removes any incentive of uncovering the underlying bug.

Kelebek1:
Closing this, as according to BSoD many more asserts throw errors besides channel count. I'm pretty convinced the parameters are just very corrupt all-round, and it's looking like a memory issue.

PR #6525, opened 2021-06-25
nvdec: Fix Submit Ioctl data source, vic frame dimension computations

ameerj:
This fixes NVDEC related issues in Mario Golf: Super Rush. The Submit Ioctl was meant to read directly from the Object's CPU memory address, rather than the mapped GPU address.

Fixing this unearthed issues in the VIC's computation of the output frame dimensions, which needed to be adjusted to avoid out of bound memory issues.

There are also various refactoring in nvhost_nvdec_common to maintain consistency.

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_nvdec_common.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_nvdec_common.cpp
@@ -93,15 +93,9 @@ NvResult nvhost_nvdec_common::Submit(const std::vector<u8>& input, std::vector<u
     for (const auto& cmd_buffer : command_buffers) {
         const auto object = nvmap_dev->GetObject(cmd_buffer.memory_id);
         ASSERT_OR_EXECUTE(object, return NvResult::InvalidState;);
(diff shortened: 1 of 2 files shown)


Itamai:
This PR causes a regression in Mario Golf: Super Rush
Hardware:
Ryzen 7 3700x
AMD RX580 | 8 GB
16GB RAM

On PR Version:
Freezes at the beginning with a blackscreen no sound on both render APIs
On Mainline:
There is still a blackscreen but audio is playing and you are able to press start to skip the intro and go to the main menu
[yuzu_logbackup.txt](https://github.com/yuzu-emu/yuzu/files/6718366/yuzu_logbackup.txt)

ameerj:
@Itamai Mario Golf has a much deeper issue with memory management in regards to nvdec emulation. It seems this causes issues with the gpu synchronization as it is waiting on invalid syncpoints and subsequently hangs.
I removed this PR from EA. for the time being please disable nvdec emulation for this game in the per game configuration.

FernandoS27 (approved):

PR #6531, opened 2021-06-26
Out of bound blit

vonchenplus:
When I was testing DQ3, I found some strange lines(see before image). Through my debugging, the problem caused by the out-of-bounds blit is not handled properly(#6289). This PR aims to handle this by adding an offset to source texture address to eliminate copy problems caused by out-of-bounds Blit.

**Before**
image

**After:**

image

### DIFF
--- a/src/video_core/engines/fermi_2d.cpp
+++ b/src/video_core/engines/fermi_2d.cpp
@@ -68,7 +68,8 @@ void Fermi2D::Blit() {
     Surface src = regs.src;
     const auto bytes_per_pixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(src.format));
     const auto need_align_to_pitch =
-        src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch && static_cast<s32>(src.width) == config.src_x1 &&
+        src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch &&
+        static_cast<s32>(src.width) == config.src_x1 &&
         config.src_x1 > static_cast<s32>(src.pitch / bytes_per_pixel) && config.src_x0 > 0;
     if (need_align_to_pitch) {
         auto address = src.Address() + config.src_x0 * bytes_per_pixel;
--- a/src/video_core/engines/fermi_2d.cpp
+++ b/src/video_core/engines/fermi_2d.cpp
@@ -7,6 +7,10 @@
 #include "video_core/engines/fermi_2d.h"
 #include "video_core/memory_manager.h"
 #include "video_core/rasterizer_interface.h"
+#include "video_core/surface.h"
+
+using VideoCore::Surface::BytesPerBlock;
+using VideoCore::Surface::PixelFormatFromRenderTargetFormat;
 
 namespace Tegra::Engines {
 
@@ -49,7 +53,7 @@ void Fermi2D::Blit() {
     UNIMPLEMENTED_IF_MSG(regs.clip_enable != 0, "Clipped blit enabled");
 
     const auto& args = regs.pixels_from_memory;
-    const Config config{
+    Config config{
         .operation = regs.operation,
         .filter = args.sample_mode.filter,
         .dst_x0 = args.dst_x0,
@@ -61,7 +65,20 @@ void Fermi2D::Blit() {
         .src_x1 = static_cast<s32>((args.du_dx * args.dst_width + args.src_x0) >> 32),
         .src_y1 = static_cast<s32>((args.dv_dy * args.dst_height + args.src_y0) >> 32),
     };
-    if (!rasterizer->AccelerateSurfaceCopy(regs.src, regs.dst, config)) {
+    Surface src = regs.src;
+    const auto bytes_per_pixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(src.format));
+    const auto need_align_to_pitch =
+        src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch && src.width == config.src_x1 &&
+        config.src_x1 > static_cast<s32>(src.pitch / bytes_per_pixel) && config.src_x0 > 0;
+    if (need_align_to_pitch) {
+        auto address = src.Address() + config.src_x0 * bytes_per_pixel;
+        src.addr_upper = static_cast<u32>(address >> 32);
+        src.addr_lower = static_cast<u32>(address);
+        src.width -= config.src_x0;
+        config.src_x1 -= config.src_x0;
+        config.src_x0 = 0;
+    }
+    if (!rasterizer->AccelerateSurfaceCopy(src, regs.dst, config)) {
         UNIMPLEMENTED();
     }
 }
--- a/src/video_core/engines/fermi_2d.cpp
+++ b/src/video_core/engines/fermi_2d.cpp
@@ -68,7 +68,7 @@ void Fermi2D::Blit() {
     Surface src = regs.src;
     const auto bytes_per_pixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(src.format));
     const auto need_align_to_pitch =
-        src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch && src.width == config.src_x1 &&
+        src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch && static_cast<s32>(src.width) == config.src_x1 &&
         config.src_x1 > static_cast<s32>(src.pitch / bytes_per_pixel) && config.src_x0 > 0;
     if (need_align_to_pitch) {
         auto address = src.Address() + config.src_x0 * bytes_per_pixel;
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -158,9 +158,7 @@ public:
     /// Blit an image with the given parameters
     void BlitImage(const Tegra::Engines::Fermi2D::Surface& dst,
                    const Tegra::Engines::Fermi2D::Surface& src,
-                   const Tegra::Engines::Fermi2D::Config& copy,
-                   std::optional<Region2D> src_region_override = {},
-                   std::optional<Region2D> dst_region_override = {});
+                   const Tegra::Engines::Fermi2D::Config& copy);
 
     /// Invalidate the contents of the color buffer index
     /// These contents become unspecified, the cache can assume aggressive optimizations.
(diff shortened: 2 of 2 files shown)


src/video_core/texture_cache/texture_cache.h
```diff
@@ -60,6 +60,10 @@ using VideoCore::Surface::PixelFormatFromRenderTargetFormat;
 using VideoCore::Surface::SurfaceType;
 using namespace Common::Literals;
 
+namespace {
```
lioncash, line 63: Why is this in an anonymous namespace instead of being declared with the other using declarations above?

src/video_core/texture_cache/texture_cache.h
```diff
@@ -158,9 +162,8 @@ class TextureCache {
     /// Blit an image with the given parameters
     void BlitImage(const Tegra::Engines::Fermi2D::Surface& dst,
                    const Tegra::Engines::Fermi2D::Surface& src,
-                   const Tegra::Engines::Fermi2D::Config& copy,
-                   std::optional<Region2D> src_region_override = {},
-                   std::optional<Region2D> dst_region_override = {});
+                   const int offset,
```
lioncash, line 165:
```suggestion
                   int offset,
```

Given top-level const doesn't do anythin in a declaration for pass-by-value.

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,20 @@ void Fermi2D::Blit() {
         .src_x1 = static_cast<s32>((args.du_dx * args.dst_width + args.src_x0) >> 32),
         .src_y1 = static_cast<s32>((args.dv_dy * args.dst_height + args.src_y0) >> 32),
     };
-    if (!rasterizer->AccelerateSurfaceCopy(regs.src, regs.dst, config)) {
+
+    int offset = 0;
+    auto bytesPerPixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(regs.src.format));
```
lioncash, line 70:
```suggestion
    const auto bytes_per_pixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(regs.src.format));
```

src/video_core/texture_cache/texture_cache.h
vonchenplus: Sorry, I submitted the debugging code

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,20 @@ void Fermi2D::Blit() {
         .src_x1 = static_cast<s32>((args.du_dx * args.dst_width + args.src_x0) >> 32),
         .src_y1 = static_cast<s32>((args.dv_dy * args.dst_height + args.src_y0) >> 32),
     };
-    if (!rasterizer->AccelerateSurfaceCopy(regs.src, regs.dst, config)) {
+
+    int offset = 0;
```
Kelebek1, line 69: We try to use s32/f32 instead of int/float etc, use s32 for all of these instead of int to match the rest of these files.

ameerj:
Thanks for looking further into this issue. 
Just as an fyi, Vulkan is a lot stricter when it comes to the out of bound blits, and this implementation seems to have regressed Shantae when using Vulkan:
image

vonchenplus:
Hello @ameerj ,How can I enter your menu？I load the game directly into follow:

image

v1993:
I may be wrong, but those are probably two *different* games (from same series). Shantae and  the Seven Sirens is a switch game, Shantae is a GBC/GBA game rerelease running in emulator.

vonchenplus:
Okay, got it

vonchenplus:
Hello @ameerj , I tested Shantae locally and did not find the problem you mentioned.

By the way, my graphics card is GTX1070Ti.

image

ameerj:
I tested on the the windows artifact from this PR which can be found here: https://dev.azure.com/yuzu-emu/yuzu/_build/results?buildId=13305&view=results
It's compiled using MINGW. It may be indicative of undefined behavior if it works on MSVC but not MINGW.

vonchenplus:
Hello @ameerj , I tested the yuzu-pr-verify-windows-mingw version from the link you provided and it still works normally. I think may be related to the video card driver.

image

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,19 @@ void Fermi2D::Blit() {
...
     };
-    if (!rasterizer->AccelerateSurfaceCopy(regs.src, regs.dst, config)) {
+
+    s32 src_address_offset = 0;
+    const auto bytes_per_pixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(regs.src.format));
+    if (regs.src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch &&
+        regs.src.width == config.src_x1 &&
+        config.src_x1 > static_cast<s32>(regs.src.pitch / bytes_per_pixel) && config.src_x0 > 0) {
+        src_address_offset = config.src_x0 * bytes_per_pixel;
+        regs.src.width -= config.src_x0;
```
ReinUsesLisp, line 75: Register values shouldn't be changed on dispatch. Use a local variable instead.

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,19 @@ void Fermi2D::Blit() {
         .src_x1 = static_cast<s32>((args.du_dx * args.dst_width + args.src_x0) >> 32),
         .src_y1 = static_cast<s32>((args.dv_dy * args.dst_height + args.src_y0) >> 32),
     };
-    if (!rasterizer->AccelerateSurfaceCopy(regs.src, regs.dst, config)) {
+
+    Surface src = regs.src;
```
ameerj, line 69: You can simplify this PR's diff by simply adding the `src_address_offset` to this src `Surface` struct's address, rather than passing the value along and adding it in 2 locations later on.

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,19 @@ void Fermi2D::Blit() {
...
+    s32 src_address_offset = 0;
+    const auto bytes_per_pixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(src.format));
+    if (src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch && src.width == config.src_x1 &&
```
ameerj, line 72: can you break up this condition into a bool to give it a more descriptive name for what it's checking for?

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,19 @@ void Fermi2D::Blit() {
...
+        config.src_x1 > static_cast<s32>(src.pitch / bytes_per_pixel) && config.src_x0 > 0) {
+        src_address_offset = config.src_x0 * bytes_per_pixel;
+        src.width -= config.src_x0;
+        config.src_x1 -= config.src_x0;
+        config.src_x0 = 0;
+    }
+
```
ameerj, line 79: unneeded newline

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,19 @@ void Fermi2D::Blit() {
         .src_x1 = static_cast<s32>((args.du_dx * args.dst_width + args.src_x0) >> 32),
         .src_y1 = static_cast<s32>((args.dv_dy * args.dst_height + args.src_y0) >> 32),
     };
-    if (!rasterizer->AccelerateSurfaceCopy(regs.src, regs.dst, config)) {
+
```
ameerj, line 68: unneeded newline

goldenx86:
Hi, i'm one of the authors of the monthly progress reports. I'm writing the next one and I wanted to know if it is ok to mention you and include your PR.

vonchenplus:
> Hi, i'm one of the authors of the monthly progress reports. I'm writing the next one and I wanted to know if it is ok to mention you and include your PR.

It's all right

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,20 @@ void Fermi2D::Blit() {
...
+    Surface src = regs.src;
+    const auto bytes_per_pixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(src.format));
+    const auto is_copy_out_of_bound =
+        src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch && src.width == config.src_x1 &&
+        config.src_x1 > static_cast<s32>(src.pitch / bytes_per_pixel) && config.src_x0 > 0;
+    if (is_copy_out_of_bound) {
+        auto address = src.Address() + config.src_x0 * bytes_per_pixel;
+        src.addr_upper = static_cast<u32>(address >> 32);
+        src.addr_lower = static_cast<u32>(address);
+        src.width -= config.src_x0;
```
bylaws, line 77: What exactly is this inside the if statement doing if it's out of bounds? It could be worth adding a comment

vonchenplus: The solution I have described in PR，by adding an offset to source texture address to eliminate copy problems caused by out-of-bounds Blit

bylaws: What does the offset mean though?

vonchenplus: This question is cumbersome to explain, I hope the following answers can help further understanding. Please look at the picture below. I have drawn three boxes, namely red, green and blue. Red represents the original area, green represents the area of the pitch, and blue represents the result of the algorithm modification. When we find that the data is not aligned (pitch > width * bytes_per_pixel), because the subsequent data copy is based on the pitch, we need to add an offset to correct this problem.

image

bylaws: Ah I see! Thanks for explaining I understand now

src/video_core/engines/fermi_2d.cpp
```diff
@@ -61,7 +65,20 @@ void Fermi2D::Blit() {
         .src_x1 = static_cast<s32>((args.du_dx * args.dst_width + args.src_x0) >> 32),
         .src_y1 = static_cast<s32>((args.dv_dy * args.dst_height + args.src_y0) >> 32),
     };
-    if (!rasterizer->AccelerateSurfaceCopy(regs.src, regs.dst, config)) {
+    Surface src = regs.src;
+    const auto bytes_per_pixel = BytesPerBlock(PixelFormatFromRenderTargetFormat(src.format));
+    const auto need_align_to_pitch =
+        src.linear == Tegra::Engines::Fermi2D::MemoryLayout::Pitch && src.width == config.src_x1 &&
```
bunnei, line 71: There is a signed/unsigned mismatch error here, untagging this until you get a chance to rebase. Thanks!

bunnei (approved):

bunnei:
LGTM, Thank you!

PR #6532, opened 2021-06-26
libusb: Apple is a POSIX system

MerryMage:


Morph1984 (approved):
Indeed, Apple is a POSIX system

PR #6535, opened 2021-06-28
main: Display the instruction set of the running title in the window name

ameerj:
Displays whether the currently running title uses 64-bit instructions or only 32-bit instructions.

Useful to know which games would benefit from certain optimizations, since some of the CPU/dynarmic unsafe options apply only to 32-bit titles.

Example screenshot:
image


Moonlacer:
Yesss, thanks for making this, it will be much easier now to tell whether a game will work or not with the recent unlimit framerate option for 32 bit games. We now no longer have to try each game to be sure whether or not they will work with the unlimit framerate option 👍

PR #6537, opened 2021-06-28
general: Enforce multiple warnings in MSVC

Morph1984:
Enforce C4242 in core, video_core, and input_common
Enforce C4018, C4267, C4305, C4389 throughout the entire project

### DIFF
--- a/src/video_core/command_classes/vic.cpp
+++ b/src/video_core/command_classes/vic.cpp
@@ -9,16 +9,9 @@ extern "C" {
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wconversion"
 #endif
-#ifdef _MSC_VER
-#pragma warning(disable : 4244) // conversion from 'type' to 'type', possible loss of data
-#pragma warning(push)
-#endif
 #include <libswscale/swscale.h>
 #if defined(__GNUC__) || defined(__clang__)
 #pragma GCC diagnostic pop
-#endif
-#ifdef _MSC_VER
-#pragma warning(pop)
 #endif
 }
 
--- a/src/video_core/textures/astc.cpp
+++ b/src/video_core/textures/astc.cpp
@@ -1365,8 +1365,8 @@ static void DecompressBlock(std::span<const u8, 16> inBuf, const u32 blockWidth,
     // each partition.
 
     // Determine partitions, partition index, and color endpoint modes
-    s32 planeIdx = -1;
-    u32 partitionIndex;
+    u32 planeIdx{UINT32_MAX};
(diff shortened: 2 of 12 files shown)


src/video_core/command_classes/vic.cpp
```diff
@@ -10,6 +10,7 @@ extern "C" {
 #pragma GCC diagnostic ignored "-Wconversion"
 #endif
 #ifdef _MSC_VER
+#pragma warning(disable : 4242) // conversion from 'type' to 'type', possible loss of data
 #pragma warning(disable : 4244) // conversion from 'type' to 'type', possible loss of data
 #pragma warning(push)
```
lioncash, line 15: This push needs to be above the warning disabling pragmas, otherwise this isn't going to pop them off the warning context and keep them disabled.

src/video_core/textures/astc.cpp
```diff
@@ -1365,8 +1365,8 @@ static void DecompressBlock(std::span<const u8, 16> inBuf, const u32 blockWidth,
     // each partition.
 
     // Determine partitions, partition index, and color endpoint modes
-    s32 planeIdx = -1;
-    u32 partitionIndex;
+    u32 planeIdx{};
```
lioncash, line 1368: Shouldn't this be assigned to UINT32_MAX?

src/input_common/udp/client.cpp
```diff
@@ -7,7 +7,16 @@
 #include <functional>
 #include <random>
 #include <thread>
+
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4242) // conversion from 'type' to 'type', possible loss of data
```
MerryMage, line 13: @lioncash Would
```
        /experimental:external
        /external:W0
        /external:anglebrackets
```
as compiler flags be more appropriate?

lioncash: If it works properly on the buildbot as well, then definitely.

Morph1984: https://docs.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-160

Apparently above VS 2019 16.10 we don't need `experimental:external`, question is if the buildbots are on that

Morph1984: Since dynarmic uses it, I guess it should be fine?

Morph1984: I've added the above and removed the #pragma warning directives, thanks @MerryMage

lioncash (approved):

bunnei (approved):

PR #6539, opened 2021-06-28
general: Move most settings' defaults and labels into their definition

lat9nq:
So the other day, I tried adding a setting to yuzu. I did most of it correctly, and yuzu would save it for the most part like normal, but one specific value of the setting would cause yuzu to revert it back to the default. I spent an hour debugging it, only to find out that it was because I set conflicting defaults when I call `WriteSettingGlobal` and `ReadSettingGlobal` in `config.cpp`.

This PR aims to put an end to these kinds of silly situations. This wraps most of the settings in `common/settings.h` and `yuzu/uisettings.h` into a resource managing class called `BasicSetting`, and the old `Setting` class becomes a subclass of `BasicSetting`. In addition, these classes now both require setting a default and label during definition.

Wherever applicable in each frontend, new helper functions were added to easily take advantage of the new default and label properties, and some of the old ones from the Qt frontend were updated to use this.

Also adds a bit of documentation. Also fixes conflicting defaults for `Settings::values.touch_device` in the Qt frontend. Also fixes a regression from #4098 where `float` settings use the native QVariant encoding, which is not human-readable.

Should note that I didn't force any of the changes from BasicSetting on the InputSetting class. @Morph1984 your input would be valued here on what the best course of action would be.

The default configuration in the Qt frontend changes slightly as a result of this:
```diff
8c8
< volume=@Variant(\0\0\0\x87?\x80\0\0)
---
> volume=1
78c78
< mouse_panning_sensitivity=@Variant(\0\0\0\x87?\x80\0\0)
---
> mouse_panning_sensitivity=1
763c763
< touch_device\default=false
---
> touch_device\default=true
874c874
< bg_blue=@Variant(\0\0\0\x87\0\0\0\0)
---
> bg_blue=0
876c876
< bg_green=@Variant(\0\0\0\x87\0\0\0\0)
---
> bg_green=0
878c878
< bg_red=@Variant(\0\0\0\x87\0\0\0\0)
---
> bg_red=0
961a962
> Screenshots\enable_screenshot_save_as\default=true
1087a1089
> yuzu_token\default=true
1088a1091
> yuzu_username\default=true
```


src/yuzu_cmd/config.h
```diff
@@ -22,4 +24,8 @@ class Config {
     ~Config();
 
     void Reload();
+
+private:
+    template <typename Type>
+    void ReadSetting(const std::string& group, Settings::BasicSetting<Type> setting);
```
lioncash, line 30:
```suggestion
    void ReadSetting(const std::string& group, const Settings::BasicSetting<Type>& setting);
```

lat9nq: I agree it should be a reference, but this function is modifying the setting that is passed in. (I'll also add documentation for it, seems I forgot about this one.)

src/yuzu/configuration/config.cpp
```diff
@@ -272,6 +272,87 @@ void Config::Initialize(const std::string& config_name) {
...
+    qt_config->setValue(name + QStringLiteral("/default"),
+                        setting.GetValue() == setting.GetDefault());
+    qt_config->setValue(name, setting.GetValue());
+}
+
+// Explicit float definition: use a double as Qt doesn't write legible floats to config files
+template <>
+void Config::WriteGlobalSetting(const Settings::Setting<float>& setting) {
+    const QString name = QString::fromStdString(setting.GetLabel());
+    const double value = setting.GetValue(global);
```
ameerj, line 333: ditto about use of double vs float. (perhaps the use of `auto` is better?)

src/yuzu/configuration/configure_profile_manager.cpp
```diff
@@ -245,15 +245,18 @@ void ConfigureProfileManager::DeleteUser() {
         this, tr("Confirm Delete"),
         tr("You are about to delete user with name \"%1\". Are you sure?").arg(username));
 
-    if (confirm == QMessageBox::No)
+    if (confirm == QMessageBox::No) {
         return;
+    }
```
ameerj, line 251: unneeded newline

src/yuzu/configuration/configure_profile_manager.cpp
```diff
@@ -245,15 +245,18 @@ void ConfigureProfileManager::DeleteUser() {
...
-    if (Settings::values.current_user == tree_view->currentIndex().row())
+    if (Settings::values.current_user.GetValue() == tree_view->currentIndex().row()) {
         Settings::values.current_user = 0;
+    }
     UpdateCurrentUser();
 
-    if (!profile_manager->RemoveUser(*uuid))
+    if (!profile_manager->RemoveUser(*uuid)) {
         return;
+    }
```
ameerj, line 260: unneeded newline

ameerj (commented):
Thank you for taking the time to make these changes. The usage of this new settings API looks to be really clean and a lot less clunky and error prone.

Just a few nits.

src/yuzu/configuration/config.cpp
```diff
@@ -272,6 +272,87 @@ void Config::Initialize(const std::string& config_name) {
...
+    const QString name = QString::fromStdString(setting.GetLabel());
+    qt_config->setValue(name + QStringLiteral("/default"),
+                        setting.GetValue() == setting.GetDefault());
+    qt_config->setValue(name, QString::fromStdString(setting.GetValue()));
+}
+// Explicit float definition: use a double as Qt doesn't write legible floats to config files
+template <>
+void Config::WriteBasicSetting(const Settings::BasicSetting<float>& setting) {
+    const QString name = QString::fromStdString(setting.GetLabel());
+    const double value = setting.GetValue();
```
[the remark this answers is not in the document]
lat9nq: Is the above comment not clear enough? `Explicit float definition: use a double as Qt doesn't write legible floats to config files`

I am intentionally using a double here. This is the difference between `bg_blue=@Variant(\0\0\0\x87\0\0\0\0)` (float) and `bg_blue=0` (double).

src/yuzu/configuration/configure_profile_manager.cpp
lat9nq: I didn't add the newline being discussed here? And I assume the newline is here for phrasing, rather than having just a wall of code.

src/common/settings.h
```diff
@@ -34,68 +36,236 @@ enum class CPUAccuracy : u32 {
...
+        return global;
+    }
+
+    /**
+     * Sets the setting to the given value.
+     *
+     * @param value The desired value
+     */
+    void SetValue(const Type& value) {
+        Type temp{value};
```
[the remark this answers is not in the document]
lat9nq: Among other primitives, `Type` is being used with `std::string` and `std::optional<u32>`.

src/input_common/mouse/mouse_poller.cpp
```diff
@@ -84,7 +84,8 @@ class MouseAnalog final : public Input::AnalogDevice {
         std::lock_guard lock{mutex};
         const auto axis_value =
             static_cast<float>(mouse_input->GetMouseState(button).axis.at(axis));
-        return axis_value * Settings::values.mouse_panning_sensitivity / (100.0f * range);
+        return axis_value * Settings::values.mouse_panning_sensitivity.GetValue() /
+               (100.0f * range);
```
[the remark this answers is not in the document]
lat9nq: Shouldn't `f32` be `float` to match the setting's type? Or should `mouse_panning_sensitivity` be a `f32`?

src/yuzu/configuration/config.cpp
```diff
@@ -272,6 +272,87 @@ void Config::Initialize(const std::string& config_name) {
...
+    qt_config->setValue(name + QStringLiteral("/default"),
+                        setting.GetValue() == setting.GetDefault());
```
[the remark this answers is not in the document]
lat9nq: `value` and `GetDefault()` will use floating point numbers with different precisions. Perhaps unwise to compare them directly? (Should I even be comparing `float`s like this?)

ameerj (approved):

PR #6540, opened 2021-06-29
Slightly refactor NVDEC and codecs for readability and safety

Kelebek1:
Moves NVDEC registers into NVDEC, rather than in Codec, and change the way the methods are called.

Add offsets to the structs and introduce bitfields. H264 especially currently relies on raw bitshifting and AND'ing the flags, this PR makes the code clearer to read what's being used, and less error-prone.

Statically assert struct member offsets for some safety.

Remove a lot of unnecessary initialisation in the structs.

Rename a few members based on official docs, as per ameerj's request.

### DIFF
--- a/src/video_core/command_classes/codecs/codec.cpp
+++ b/src/video_core/command_classes/codecs/codec.cpp
@@ -23,8 +23,8 @@ void AVFrameDeleter(AVFrame* ptr) {
     av_free(ptr);
 }
 
-Codec::Codec(GPU& gpu_)
-    : gpu(gpu_), h264_decoder(std::make_unique<Decoder::H264>(gpu)),
+Codec::Codec(GPU& gpu_, const NvdecCommon::NvdecRegisters& regs)
+    : gpu(gpu_), state{regs}, h264_decoder(std::make_unique<Decoder::H264>(gpu)),
       vp9_decoder(std::make_unique<Decoder::VP9>(gpu)) {}
 
 Codec::~Codec() {
@@ -43,46 +43,48 @@ Codec::~Codec() {
     avcodec_close(av_codec_ctx);
 }
 
-void Codec::SetTargetCodec(NvdecCommon::VideoCodec codec) {
-    if (current_codec != codec) {
-        LOG_INFO(Service_NVDRV, "NVDEC video codec initialized to {}", static_cast<u32>(codec));
-        current_codec = codec;
+void Codec::Initialize() {
+    AVCodecID codec{AV_CODEC_ID_NONE};
+    switch (current_codec) {
+    case NvdecCommon::VideoCodec::H264:
+        codec = AV_CODEC_ID_H264;
+        break;
+    case NvdecCommon::VideoCodec::Vp9:
+        codec = AV_CODEC_ID_VP9;
+        break;
+    default:
+        return;
     }
+    av_codec = avcodec_find_decoder(codec);
+    av_codec_ctx = avcodec_alloc_context3(av_codec);
+    av_opt_set(av_codec_ctx->priv_data, "tune", "zerolatency", 0);
+
+    // TODO(ameerj): libavcodec gpu hw acceleration
+
+    const auto av_error = avcodec_open2(av_codec_ctx, av_codec, nullptr);
+    if (av_error < 0) {
+        LOG_ERROR(Service_NVDRV, "avcodec_open2() Failed.");
+        avcodec_close(av_codec_ctx);
+        return;
+    }
+    initialized = true;
+    return;
 }
 
-void Codec::StateWrite(u32 offset, u64 arguments) {
-    u8* const state_offset = reinterpret_cast<u8*>(&state) + offset * sizeof(u64);
-    std::memcpy(state_offset, &arguments, sizeof(u64));
+void Codec::SetTargetCodec(NvdecCommon::VideoCodec codec) {
+    if (current_codec != codec) {
+        current_codec = codec;
+        LOG_INFO(Service_NVDRV, "NVDEC video codec initialized to {}", GetCurrentCodecName());
+    }
 }
 
 void Codec::Decode() {
-    bool is_first_frame = false;
+    const bool is_first_frame = !initialized;
     if (!initialized) {
-        if (current_codec == NvdecCommon::VideoCodec::H264) {
-            av_codec = avcodec_find_decoder(AV_CODEC_ID_H264);
-        } else if (current_codec == NvdecCommon::VideoCodec::Vp9) {
-            av_codec = avcodec_find_decoder(AV_CODEC_ID_VP9);
-        } else {
-            LOG_ERROR(Service_NVDRV, "Unknown video codec {}", current_codec);
-            return;
-        }
-
-        av_codec_ctx = avcodec_alloc_context3(av_codec);
-        av_opt_set(av_codec_ctx->priv_data, "tune", "zerolatency", 0);
-
-        // TODO(ameerj): libavcodec gpu hw acceleration
-
-        const auto av_error = avcodec_open2(av_codec_ctx, av_codec, nullptr);
-        if (av_error < 0) {
-            LOG_ERROR(Service_NVDRV, "avcodec_open2() Failed.");
-            avcodec_close(av_codec_ctx);
-            return;
-        }
-        initialized = true;
-        is_first_frame = true;
+        Initialize();
     }
-    bool vp9_hidden_frame = false;
 
+    bool vp9_hidden_frame = false;
     AVPacket packet{};
     av_init_packet(&packet);
     std::vector<u8> frame_data;
@@ -95,7 +97,7 @@ void Codec::Decode() {
     }
 
     packet.data = frame_data.data();
-    packet.size = static_cast<int>(frame_data.size());
+    packet.size = static_cast<s32>(frame_data.size());
 
     avcodec_send_packet(av_codec_ctx, &packet);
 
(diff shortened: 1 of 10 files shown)


src/video_core/command_classes/codecs/h264.cpp
```diff
@@ -45,134 +45,130 @@ H264::~H264() = default;
...
-            writer.WriteUe(0);
-        }
-
-        const s32 pic_height = context.h264_parameter_set.pic_height_in_map_units /
-                               (context.h264_parameter_set.frame_mbs_only_flag ? 1 : 2);
+        return frame;
+    }
 
-        writer.WriteUe(16);
+    /// Encode header
```
ReinUsesLisp, line 58:
```suggestion
    // Encode header
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -23,8 +23,8 @@ void AVFrameDeleter(AVFrame* ptr) {
     av_free(ptr);
 }
 
-Codec::Codec(GPU& gpu_)
-    : gpu(gpu_), h264_decoder(std::make_unique<Decoder::H264>(gpu)),
+Codec::Codec(GPU& gpu_, NvdecCommon::NvdecRegisters& regs)
```
Morph1984, line 26: is `regs` meant to be a mutable reference?

src/video_core/command_classes/codecs/h264.h
```diff
@@ -48,7 +50,7 @@ class H264BitWriter {
 
     /// Based on section 7.3.2.1.1.1 and Table 7-4 in the H.264 specification
     /// Writes the scaling matrices of the sream
-    void WriteScalingList(const std::vector<u8>& list, s32 start, s32 count);
+    void WriteScalingList(const std::span<u8>& list, s32 start, s32 count);
```
Morph1984, line 53: A span is a view of the underlying bytes
```suggestion
    void WriteScalingList(std::span<const u8> list, s32 start, s32 count);
```

Morph1984 (approved):

src/video_core/command_classes/nvdec_common.h
```diff
@@ -4,45 +4,92 @@
...
+// These registers should be 32-bit and half the size,
+// but uses 64-bit here due to not emulating multiple GPU channels.
+// NVDEV should use a 32-bit address space, but is mapped to 64-bit,
+// doubling the sizes here is compensating for that.
+struct NvdecRegisters {
+    static constexpr std::size_t NUM_REGS = 0x178;
+
+    union {
+        struct {
+            /* 0x0000 */ INSERT_PADDING_WORDS_NOINIT(256);
```
ameerj, line 30: I would prefer having the offset comments be on the right side of each member, and prefer `///< 0x42` over `/* 0x42 */`

src/video_core/command_classes/nvdec_common.h
```diff
@@ -4,45 +4,92 @@
...
     None = 0x0,
     H264 = 0x3,
     Vp8 = 0x5,
     H265 = 0x7,
     Vp9 = 0x9,
 };
 
+// These registers should be 32-bit and half the size,
+// but uses 64-bit here due to not emulating multiple GPU channels.
+// NVDEV should use a 32-bit address space, but is mapped to 64-bit,
```
ameerj, line 23:
```suggestion
// NVDEC should use a 32-bit address space, but is mapped to 64-bit,
```
And you can delete the first two lines, as you're repeating the same information twice.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -43,42 +43,41 @@ Codec::~Codec() {
...
-        // TODO(ameerj): libavcodec gpu hw acceleration
-
-        const auto av_error = avcodec_open2(av_codec_ctx, av_codec, nullptr);
-        if (av_error < 0) {
-            LOG_ERROR(Service_NVDRV, "avcodec_open2() Failed.");
-            avcodec_close(av_codec_ctx);
-            return;
-        }
-        initialized = true;
+        Initialize();
```
ameerj, line 80: Moving the logic out to a separate function will cause issues because this will no longer return early when an unsupported codec is encountered.

Kelebek1: I don't think this is possible, since it's already behind a switch of only valid codecs when called from https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/command_classes/nvdec.cpp#L37. It feels like we're checking the codec too much as-is, maybe need to move more logic into the respective derived class so we don't need to do that?

For now I'll add the check as you suggest.

src/video_core/command_classes/codecs/vp9_types.h
```diff
@@ -15,10 +15,10 @@ class GPU;
 
 namespace Decoder {
 struct Vp9FrameDimensions {
-    s16 width{};
-    s16 height{};
-    s16 luma_pitch{};
-    s16 chroma_pitch{};
+    /* 0x00 */ s16 width;
```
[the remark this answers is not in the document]
Kelebek1: I just added it to any structs which have a specific and required size that's statically asserted.

src/video_core/command_classes/codecs/h264.h
```diff
@@ -78,40 +80,99 @@ class H264 {
...
-    std::vector<u8> frame;
-    GPU& gpu;
+    static_assert(sizeof(H264DecoderContext) == 0x2A0, "H264DecoderContext is an invalid size");
+
+#define ASSERT_POSITION(field_name, position)                                                      \
+    static_assert(offsetof(H264ParameterSet, field_name) == position * sizeof(u32),                \
+                  "Field " #field_name " has invalid position")
+
+    ASSERT_POSITION(log2_max_pic_order_cnt_lsb_minus4, 0x0);
+    ASSERT_POSITION(delta_pic_order_always_zero_flag, 0x1);
```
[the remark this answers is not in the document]
Kelebek1: I think it can be helpful to have multiple representations. For the GPU engine registers for example, envytools, nouveau and nVidia use byte offsets, but deko3d uses u32 offsets, which was Rodrigo's preferred reference, and they're all asserted as being `offset / u32`. That was my reasoning anyways, I'll change these over though. What about the NVDEC registers themselves?

src/video_core/command_classes/codecs/codec.cpp
ameerj: Ah if the invoking function does the check already then ignore this suggestion.

src/video_core/command_classes/codecs/h264.h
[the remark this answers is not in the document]
ameerj: Like you said, docs refer to these data in offsets of bytes, so it helps to keep that consistent.
The state registers are also addressed by offsets in bytes, what is called the "method" in this implementation. So I think it makes sense to keep the asserts based on the data we expect to be working with, which will be offsets in units of bytes.

src/video_core/command_classes/nvdec_common.h
```diff
@@ -4,45 +4,90 @@
...
     Vp8 = 0x5,
     H265 = 0x7,
     Vp9 = 0x9,
 };
 
-} // namespace Tegra::NvdecCommon
+// NVDEV should use a 32-bit address space, but is mapped to 64-bit,
+// doubling the sizes here is compensating for that.
+struct NvdecRegisters {
+    static constexpr std::size_t NUM_REGS = 0x178;
```
ameerj, line 24:
```suggestion
    constexpr std::size_t NUM_REGS = 0x178;
```
Don't use static in a header, otherwise it will duplicate this in every cpp file that includes the header

src/video_core/command_classes/nvdec_common.h
```diff
@@ -4,45 +4,90 @@
...
+ASSERT_REG_POSITION(vp9_entropy_probs_offset, 0x170);
+ASSERT_REG_POSITION(vp9_backward_updates_offset, 0x171);
+ASSERT_REG_POSITION(vp9_last_frame_segmap_offset, 0x172);
+ASSERT_REG_POSITION(vp9_curr_frame_segmap_offset, 0x173);
+ASSERT_REG_POSITION(vp9_last_frame_mvs_offset, 0x175);
+ASSERT_REG_POSITION(vp9_curr_frame_mvs_offset, 0x176);
+
+#undef ASSERT_REG_POSITION
+
+} // namespace Tegra::NvdecCommon
```
ameerj, line 93: missing newline

src/video_core/command_classes/codecs/h264.h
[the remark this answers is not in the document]
Kelebek1: The state regs are addressed in words, `method` is a word index, not a byte offset.

src/video_core/command_classes/nvdec_common.h
Kelebek1: I get an error about not being able to use constexpr if I remove it. I just copied the other engines which are all static as well, and the same error seems to pop up for those if I try to remove static.

ameerj: Ah sorry I missed that this was a member variable of the struct.

src/video_core/command_classes/codecs/h264.h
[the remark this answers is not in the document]
ameerj: You're right. I was thinking back to my old implementation, which needed a byte offset.

ameerj (approved):
Thanks again for taking the time to make these changes!

PR #6542, opened 2021-06-29
Fix when the program index is greater than 0 invalid

vonchenplus:
This PR aims to fix error will occur when the program index is greater than 0, and I think we also need UI support to load multiple programs.

### DIFF
--- a/src/core/file_sys/card_image.cpp
+++ b/src/core/file_sys/card_image.cpp
@@ -62,13 +62,12 @@ XCI::XCI(VirtualFile file_, std::size_t program_index)
     }
 
     secure_partition = std::make_shared<NSP>(
-        main_hfs.GetFile(partition_names[static_cast<std::size_t>(XCIPartition::Secure)]),
-        program_index);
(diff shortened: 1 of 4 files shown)


german77:
Can you provide the name of a game or homebrew that hit this problem? I would like to do some testing on my side

vonchenplus:
You can test the DQ(1, 2, 3) collection.

 Currently it can only be tested by modifying the following code:

image

Morph1984:
@vonchenplus This is likely not the proper fix for this game. The program index is specific to multi-program titles like Super Mario 3D All-Stars, where the program index is simply an offset from a program ID.

For example:
StardustLauncher has Program ID 010049900F546000
SM64 has Program ID 010049900F546001
Sunshine has Program ID 010049900F546002
Galaxy has Program ID 010049900F546003

vonchenplus:
@Morph1984 Yes, you are right. I checked the previous submission records, it just support for loading indexed programs. I think I need to submit another solution to support collection.

PR #6551, opened 2021-07-03
Improve management of kernel objects

bunnei:
Currently, the emulation session can create tons of kernel objects that are not tracked. This is particularly a problem with HLE services, which are created and torn down between emulation sessions (unlike what would happen on a real Switch), yet often create many kernel objects themselves for things like sessions, events, etc. which are not tracked. This can easily end up with ~1000+ dangling kernel objects after emulation is shutdown.

This change tries to address this by tracking the worst offenders here. These are:

* Adding global tracking of kernel objects, and logging which are dangling after emulation is shutdown.
* Ensuring handle tables are actually shutdown, freeing all of the object references held by the handle table.
* Ensuring events created by HLE services are properly shutdown, in which we added some new helper code to manage this.
* Reserving a dummy KProcess for each HLE service, to allow us to better manage resource limits.
* Globally tracking KServerPorts and KServerSessions and ensuring they are shutdown. This is a simplification/workaround, as services should manage these themselves, but we'll need to entirely rethink how we emulate HLE services before this happens.
* Better management of guest threads and processes.

Some brief testing results:
* Puyo Puyo Tetris, after going in game:
  * Before: 946 dangling kernel objects after close
  * After: 16 dangling kernel objects after close
* Super Mario Odyssey, after going in game:
  * Before: 1046 dangling kernel objects after close
  * After: 33 dangling kernel objects after close
* Breath of the Wild, after going in game:
  * Before: 987 dangling kernel objects after close
  * After: 43 dangling kernel objects after close
* Pokemon Sword, after going in game:
  * Before: 1007 dangling kernel objects after close
  * After: 17 dangling kernel objects after close

Not perfect, but a lot better.

### DIFF
--- a/src/core/hle/service/kernel_helpers.cpp
+++ b/src/core/hle/service/kernel_helpers.cpp
@@ -1,0 +1,64 @@
+// Copyright 2021 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "core/core.h"
+#include "core/hle/kernel/k_event.h"
+#include "core/hle/kernel/k_process.h"
+#include "core/hle/kernel/k_readable_event.h"
+#include "core/hle/kernel/k_resource_limit.h"
+#include "core/hle/kernel/k_scoped_resource_reservation.h"
+#include "core/hle/kernel/k_writable_event.h"
+#include "core/hle/service/kernel_helpers.h"
+
+namespace Service::KernelHelpers {
+
+ServiceContext::ServiceContext(Core::System& system_, std::string name_)
+    : kernel(system_.Kernel()) {
+    process = Kernel::KProcess::Create(kernel);
+    ASSERT(Kernel::KProcess::Initialize(process, system_, std::move(name_),
+                                        Kernel::KProcess::ProcessType::Userland)
+               .IsSuccess());
+}
+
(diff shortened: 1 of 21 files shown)


Itamai:
Hello bunnei,

this PR causes a regression in Diablo 3: Eternal Collection.

Behavior on PR:
Yuzu gets to the launching screen then it freezes and crashes.

Behavior on current Mainline(676):
Game boots normally without any issues.

Kind Regards

[yuzu_logbackup.txt](https://github.com/yuzu-emu/yuzu/files/6759576/yuzu_logbackup.txt)

src/core/hle/service/kernel_helpers.cpp
```diff
@@ -0,0 +1,64 @@
+// Copyright 2021 yuzu emulator team
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
```
v1993, line 5:
```suggestion
```

Caught by GCC 11 - `#pragma once` in cpp file

v1993 (changes requested):

v1993:
While I'm not fully certain, this PR seems to be partially responsible for #6630. Segfaults when stopping emulation also happen without it, but seemingly less often (i.e. not always). Linked issue have few backtraces (and ASAN probably can come in handy as well).

Issue #6174, opened 2021-04-09
Xeboblade definitive edition face animation glitch regression


gidoly:
this issue started on yuzu mainline 538 (which buffercache rewrite merged so possibly buffer cache regression

https://youtu.be/bs0dJ9NOmpQ

you can see face animation glitches
<img width="92" alt="20210409_160858" src="https://user-images.githubusercontent.com/66776795/114142549-0b3d5600-994e-11eb-926b-0755d3006d44.png">

lips render not properly
<img width="458" alt="20210409_160922" src="https://user-images.githubusercontent.com/66776795/114142562-0ed0dd00-994e-11eb-9f86-d1a5d1f31c47.png">

character reyns forehead and eye brows constantly bugging if you look close
and most notably ending part of shulks face

ryzen 5800x
16gb ram
2080s

gidoly:
fixed on high accuracy fix

Issue #6091, opened 2021-03-20
video_core: Many particle effects are static


lat9nq:
Caused by #5741 

So far observed in BOTW (remote bomb, lightning), ZLA (bombs), and Splatoon 2 (entering levels in Octo Canyon, spawning). A bit difficult to explain in English, but when an explosion emits a large number of particles outward, those particles don't move outward and may flicker.

https://user-images.githubusercontent.com/22451773/111859615-15b19480-893a-11eb-99c7-61e9cba7b4cf.mp4

https://user-images.githubusercontent.com/22451773/111859620-19ddb200-893a-11eb-9bc7-0a65eab27c20.mp4

https://user-images.githubusercontent.com/22451773/111860597-8bb8fa00-8940-11eb-9b80-0ae225d55b4e.mp4

I had to exit back to Octo Canyon and re-enter the level to observe the issue in Splatoon 2.

Log of Splatoon 2: [yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/6174940/yuzu_log.txt)

EDIT: More issues have been brought to attention that can be rolled into this issue.

BOTW: In fact, most particle effects are broken.

https://user-images.githubusercontent.com/22451773/111863217-12c29e00-8952-11eb-9204-c3366b84b566.mp4

https://user-images.githubusercontent.com/22451773/111863218-135b3480-8952-11eb-8b87-1a1ec6aea11e.mp4

ACNH: Chimney smoke is buggy and occasionally stops moving. This is difficult to record: my screen capture options for some reason causes the smoke to start working correctly. Due to this variability, it is potentially a separate issue. You can see it at the beginning of this video.

https://user-images.githubusercontent.com/22451773/111863855-cda06b00-8955-11eb-9710-25829997fd64.mp4

Also on ACNH, it was reported that snowflakes floating in the wind are buggy, but I was unable to reproduce this.

Edit 2021-03-25: ACNH rain is broken

https://user-images.githubusercontent.com/22451773/112544685-3b61f200-8daf-11eb-8723-98dbc50c1101.mp4

Edit 2021-04-08: Pokemon Sword precipitation

https://user-images.githubusercontent.com/22451773/113969756-1183e300-9825-11eb-81f5-e6d5d5f285cf.mp4

AntoSkate:
In Splatoon 2 the launchpads, scrolls and sardinium particles are also broken



[...]

PR #5741, opened 2021-01-17
video_core: Reimplement the buffer cache

ReinUsesLisp:
Reimplement the buffer cache using cached bindings and page level
granularity for modification tracking. This also drops the usage of
shared pointers and virtual functions from the cache.

- Bindings are cached, allowing to skip work when the game changes few
  bits between draws.
- OpenGL Assembly shaders no longer copy when a region has been modified
  from the GPU to emulate constant buffers, instead GL_EXT_memory_object
  is used to alias sub-buffers within the same allocation.
- OpenGL Assembly shaders stream constant buffer data using
  glProgramBufferParametersIuivNV, from NV_parameter_buffer_object. In
  theory this should save one hash table resolve inside the driver
  compared to glBufferSubData.
- A new OpenGL stream buffer is implemented based on fences for drivers
  that are not Nvidia's proprietary, due to their low performance on
  partial glBufferSubData calls synchronized with 3D rendering (that
  some games use a lot).
- Most optimizations are shared between APIs now, allowing Vulkan to
  cache more bindings than before, skipping unnecessary work.

This commit adds the necessary infrastructure to use Vulkan object from
OpenGL. Overall, it improves performance and fixes some bugs present on
the old cache. There are still some edge cases hit by some games that
harm performance on some vendors, this are planned to be fixed in later
commits.

Regression tracking:
- [x] Fire Emblem: Three Houses crashing before showing anything on Vulkan.
- [x] Xenoblade Vulkan won't load fonts.
- [x] Fire Emblem Warriors black screen at boot.
- [x] Animal Crossing: New Horizons black screen.
- [x] Fire Emblem: Three Houses crashes on OpenGL assembly shaders.
- [x] Age of Calamity doesn't render on GLSL.
- [x] Bravely Default 2 Demo black screen.
- [x] Bravely Default 2 crashes on Vulkan.
- [x] Final Fantasy X crashes at boot on Vulkan.
- [x] Dragonball Fighters Z doesn't render on GLSL.
- [x] Hollow Knight has pixelated fog on GLSL.
- [x] Hollow Knight crashes on Vulkan.
- [x] Link's Awakening boot screen doesn't render on GLSL.
- [x] Oninaki vertex explosions.
- [x] Shin Megami Tensei 3 has broken graphics on GLSL.
- [x] ~~Mario and Rabbids has graphical issues.~~ Happens on mainline Vulkan
- [x] Several games softlock on Vulkan.
- [x] Astral Chain can crash/softlock on some cutscenes.
[...]

PR #6557, opened 2021-07-04
Buffer Cache: Fix High downloads / Fence manager: Improve fence checking.

FernandoS27:
This PR aims to fix downloads in the buffer cache and improve further the fence manager to not fully wait after a command buffer.

This PR fixes: Botw's particles (regressed on BCR), ZLA's BowWoW, Pokemon Sword's particles, Vertex explosions on UE4 games: Yoshi's Crafted World and Bravely Default 2 (needs  Hades).

### DIFF
--- a/src/video_core/buffer_cache/buffer_base.h
+++ b/src/video_core/buffer_cache/buffer_base.h
@@ -235,6 +235,11 @@ public:
         ForEachModifiedRange<Type::GPU>(query_cpu_range, size, clear, func);
     }
 
+    template <typename Func>
+    void ForEachDownloadRangeAndClear(VAddr query_cpu_range, u64 size, Func&& func) {
+        ForEachModifiedRange<Type::GPU>(query_cpu_range, size, true, func);
+    }
+
     /// Call 'func' for each GPU modified range and unmark those pages as GPU modified
     template <typename Func>
     void ForEachDownloadRange(Func&& func) {
--- a/src/video_core/buffer_cache/buffer_base.h
+++ b/src/video_core/buffer_cache/buffer_base.h
@@ -468,7 +468,7 @@ private:
 
             const u64 current_word = state_words[word_index] & bits;
             if (clear) {
-              state_words[word_index] &= ~bits;
+                state_words[word_index] &= ~bits;
             }
 
             if constexpr (type == Type::CPU) {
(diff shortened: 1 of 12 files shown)


v1993:
Does this mean that this PR (partially?) fixes #5741?

src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -153,13 +157,17 @@ class BufferCache {
...
     void CommitAsyncFlushes();
+    void CommitAsyncFlushesHigh();
 
     /// Pop asynchronous downloads
     void PopAsyncFlushes();
 
     /// Return true when a CPU region is modified from the GPU
     [[nodiscard]] bool IsRegionGpuModified(VAddr addr, size_t size);
 
+    /// Return true when a CPU region is modified from the GPU
```
v1993, line 168: This comment is a direct copy of above, so it's probably wrong

lioncash: The member function itself can likely be const qualified as well.

src/video_core/buffer_cache/buffer_base.h
```diff
@@ -415,7 +415,7 @@ class BufferBase {
      * @param func            Function to call for each turned off region
      */
     template <Type type, typename Func>
-    void ForEachModifiedRange(VAddr query_cpu_range, s64 size, Func&& func) {
+    void ForEachModifiedRange(VAddr query_cpu_range, s64 size, bool clear, Func&& func) {
```
lioncash, line 418: perhaps also add a named function like

`ForEachModifiedRangeAndClear()`

That wraps the calling of this with `true`,

So we don't have ambiguous callsites like

```cpp
buffer.ForEachDownloadRange(c, WORD, true, [&](u64 offset, u64 size) { ++num; });
```

Which don't really indicate what `true` is actually doing.

FernandoS27:
I have removed it from EA due to calls for regressions.

src/video_core/buffer_cache/buffer_cache.h
v1993: Should `IsRegionGpuModified` be const qualified as well, then?

bunnei:
Removing from EA as with some warnings changes, this is now causing some build errors. Feel free to re-tag once you can rebase and fix them.

gidoly:
fixes #6174

AntoSkate:
also fixes #6091

ameerj (approved):

PR #6559, opened 2021-07-05
Resolve some compiler warnings

german77:
Address some simple warnings on the compiler

### DIFF
--- a/src/yuzu/debugger/profiler.cpp
+++ b/src/yuzu/debugger/profiler.cpp
@@ -143,24 +143,25 @@ void MicroProfileWidget::hideEvent(QHideEvent* ev) {
 }
 
 void MicroProfileWidget::mouseMoveEvent(QMouseEvent* ev) {
-    MicroProfileMousePosition(ev->x() / x_scale, ev->y() / y_scale, 0);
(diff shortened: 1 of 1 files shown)


src/core/hle/kernel/svc.cpp
```diff
@@ -411,7 +411,7 @@ static ResultCode WaitSynchronization(Core::System& system, s32* index, VAddr ha
               handles_address, num_handles, nano_seconds);
 
     // Ensure number of handles is valid.
-    R_UNLESS(0 <= num_handles && num_handles <= ArgumentHandleCountMax, ResultOutOfRange);
+    R_UNLESS(num_handles <= ArgumentHandleCountMax, ResultOutOfRange);
```
Morph1984, line 414: This should not be removed, as num_handles is meant to be an s32

Morph1984:
Additionally, the commit should state which compiler warning(s) it addresses

src/core/hle/kernel/svc.cpp
german77: num_handles is declared as a u64. It will be always a positive number. Unless this will be changed in the future.

Morph1984 (approved):

PR #6562, opened 2021-07-05
common: fs: More misc. changes

Morph1984:
The changes are described in their individual commits.


german77:
Is IOFile::Commit() used?

Morph1984:
> Is IOFile::Commit() used?

No, but I've added it just in case it's needed.

bunnei (approved):

PR #6569, opened 2021-07-07
audio_core: Preserve front channel volume after 6 to 2 downmix

Kelebek1:
Many games report 6 channel output while only providing data for 2. We only output 2-channel audio regardless, and in the downmixing, front left/right are only providing 36% of their volume. This is done assuming all of the other channels also contain valid data, but in a few games they don't. This PR alters the downmixing to preserve front left/right, so volume is not lost.

This improves Link's Awakening, New Super Mario Bros U, Disgaea 6, Super Kirby Clash.

Buffer raw pointers have been replaced with spans as well. This was mainly to aid debugging, as MSVC won't show elements of pointers, but will show the whole span.

### DIFF
--- a/src/audio_core/audio_renderer.cpp
+++ b/src/audio_core/audio_renderer.cpp
@@ -29,10 +29,9 @@ namespace {
                                        (static_cast<float>(r_channel) * r_mix_amount)));
 }
 
-[[nodiscard]] static constexpr std::tuple<s16, s16> Mix6To2(s16 fl_channel, s16 fr_channel,
-                                                            s16 fc_channel,
-                                                            [[maybe_unused]] s16 lf_channel,
-                                                            s16 bl_channel, s16 br_channel) {
+[[maybe_unused, nodiscard]] static constexpr std::tuple<s16, s16> Mix6To2(
+    s16 fl_channel, s16 fr_channel, s16 fc_channel, [[maybe_unused]] s16 lf_channel, s16 bl_channel,
+    s16 br_channel) {
     // Front channels are mixed 36.94%, Center channels are mixed to be 26.12% & the back channels
     // are mixed to be 36.94%
 
@@ -57,11 +56,11 @@ namespace {
     const std::array<float_le, 4>& coeff) {
     const auto left =
         static_cast<float>(fl_channel) * coeff[0] + static_cast<float>(fc_channel) * coeff[1] +
-        static_cast<float>(lf_channel) * coeff[2] + static_cast<float>(bl_channel) * coeff[0];
+        static_cast<float>(lf_channel) * coeff[2] + static_cast<float>(bl_channel) * coeff[3];
 
     const auto right =
         static_cast<float>(fr_channel) * coeff[0] + static_cast<float>(fc_channel) * coeff[1] +
-        static_cast<float>(lf_channel) * coeff[2] + static_cast<float>(br_channel) * coeff[0];
+        static_cast<float>(lf_channel) * coeff[2] + static_cast<float>(br_channel) * coeff[3];
 
     return {ClampToS16(static_cast<s32>(left)), ClampToS16(static_cast<s32>(right))};
 }
@@ -241,7 +240,7 @@ void AudioRenderer::QueueMixedBuffer(Buffer::Tag tag) {
         const auto channel_count = buffer_offsets.size();
         const auto& final_mix = mix_context.GetFinalMixInfo();
         const auto& in_params = final_mix.GetInParams();
-        std::vector<s32*> mix_buffers(channel_count);
+        std::vector<std::span<s32>> mix_buffers(channel_count);
         for (std::size_t i = 0; i < channel_count; i++) {
             mix_buffers[i] =
                 command_generator.GetMixBuffer(in_params.buffer_offset + buffer_offsets[i]);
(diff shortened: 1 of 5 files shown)


src/audio_core/audio_renderer.cpp
```diff
@@ -53,15 +53,18 @@ namespace {
...
+    // Leaving it here but commented out for now, as we might want to add it back in
     const auto left =
         static_cast<float>(fl_channel) * coeff[0] + static_cast<float>(fc_channel) * coeff[1] +
-        static_cast<float>(lf_channel) * coeff[2] + static_cast<float>(bl_channel) * coeff[0];
+        /* static_cast<float>(lf_channel) * coeff[2] +*/ static_cast<float>(bl_channel) * coeff[3];
 
     const auto right =
         static_cast<float>(fr_channel) * coeff[0] + static_cast<float>(fc_channel) * coeff[1] +
-        static_cast<float>(lf_channel) * coeff[2] + static_cast<float>(br_channel) * coeff[0];
+        /*static_cast<float>(lf_channel) * coeff[2] +*/ static_cast<float>(br_channel) * coeff[3];
```
ogniK5377, line 67: This should be undone, games provide constants for downmixing and choose to mix the LFE as they please

src/audio_core/sink_context.cpp
```diff
@@ -15,10 +15,17 @@ std::size_t SinkContext::GetCount() const {
...
-    if (has_downmix_coefs) {
+    if (in.device.down_matrix_enabled) {
         downmix_coefficients = in.device.down_matrix_coef;
+    } else {
+        downmix_coefficients = {
+            1.0f,   // front
+            0.707f, // center
+            0.0f,   // lfe
+            0.707f, // back
+        };
```
ogniK5377, line 26:
```suggestion
        downmix_coefficients = {
            0.3694f,   // front
            0.2612f, // center
            0.0f,   // lfe
            0.3694f, // back
        };
```
Might yield better results as this is what we had before. Is there any benefit to the above constants as opposed to what we had before?

ogniK5377 (changes requested):

src/audio_core/audio_renderer.cpp
Kelebek1: Everything I could find kinda said just ignore it when downmixing to 2 channels. I didn't know what default coeff to use for it though. Do you have any ideas on what coefficient should be used for it?

src/audio_core/sink_context.cpp
Kelebek1: Front can't be 0.36 or we're in the same position as before, the front will only get 36% of the output and volume will be stuck low. I got these numbers from googling around, and from section 7.8.2 in this doc: http://www.atsc.org/wp-content/uploads/2015/03/A52-201212-17.pdf

Also -3dB (pow10(-3 / 20)) just seems to be used everywhere, it's in this example on Wiki as well: https://en.wikipedia.org/wiki/Pan_law

src/audio_core/audio_renderer.cpp
```diff
@@ -53,15 +52,15 @@ namespace {
 }
 
 [[nodiscard]] static constexpr std::tuple<s16, s16> Mix6To2WithCoefficients(
-    s16 fl_channel, s16 fr_channel, s16 fc_channel, s16 lf_channel, s16 bl_channel, s16 br_channel,
-    const std::array<float_le, 4>& coeff) {
+    s16 fl_channel, s16 fr_channel, s16 fc_channel, [[maybe_unused]] s16 lf_channel, s16 bl_channel,
```
Morph1984, line 55: [[maybe_unused]] is no longer needed here since `lf_channel` is used below
```suggestion
    s16 fl_channel, s16 fr_channel, s16 fc_channel, s16 lf_channel, s16 bl_channel,
```

src/audio_core/audio_renderer.cpp
```diff
@@ -29,10 +29,9 @@ namespace {
                                        (static_cast<float>(r_channel) * r_mix_amount)));
 }
 
-[[nodiscard]] static constexpr std::tuple<s16, s16> Mix6To2(s16 fl_channel, s16 fr_channel,
-                                                            s16 fc_channel,
-                                                            [[maybe_unused]] s16 lf_channel,
-                                                            s16 bl_channel, s16 br_channel) {
+[[maybe_unused]] [[nodiscard]] static constexpr std::tuple<s16, s16> Mix6To2(
```
Morph1984, line 32: As Breadfish suggested on discord
```suggestion
[[maybe_unused, nodiscard]] static constexpr std::tuple<s16, s16> Mix6To2(
```

Morph1984:
You may want to format your commit title more succinctly, The explanation for your change goes into the commit description itself.

For your current change you may want to state the following:
Commit TItle:
audio_core: Preserve front channel volume after 6 to 2 downmix
Commit Description:
Many games report 6 channel output while only providing data for 2. We only output 2-channel audio regardless, and in the downmixing, front left/right only provide 36% of their volume. This is done assuming all of the other channels also contain valid data, but in many games they don't. This PR alters the downmixing to preserve front left/right, so volume is not lost.

This improves volume in Link's Awakening, New Super Mario Bros U, Disgaea 6, Super Kirby Clash.

src/audio_core/sink_context.cpp
ogniK5377: Fair point, can be kept then. Just wanted to make sure 👍

ogniK5377 (approved):

Morph1984 (approved):

PR #6571, opened 2021-07-08
audio_core: Replace NaN mix volume samples with silence

Kelebek1:
As the comment in the code says really, NaNs are causing crazy values in the end, essentially capping every sample in the mix to a maximum or minimum value, flipping the sign as well. I don't see much other way around this. When everything is s32 we can't propogate NaNs all the way to the sink, but even if we did we'd still have to replace them with silence there instead anyways. 

This stops XC2 blowing out the audio in certain places.

### DIFF
--- a/src/audio_core/command_generator.cpp
+++ b/src/audio_core/command_generator.cpp
@@ -42,6 +42,15 @@ void ApplyMix(std::span<s32> output, std::span<const s32> input, s32 gain, s32 s
 
 s32 ApplyMixRamp(std::span<s32> output, std::span<const s32> input, float gain, float delta,
                  s32 sample_count) {
+    // XC2 passes in NaN mix volumes, causing further issues as we handle everything as s32 rather
+    // than float, so the NaN propogation is lost. As the samples get further modified for
+    // volume etc, they can get out of NaN range, so a later heuristic for catching this is
+    // more difficult. Handle it here by setting these samples to silence.
+    if (std::isnan(gain)) {
+        gain = 0.0f;
+        delta = 0.0f;
+    }
+
     s32 x = 0;
     for (s32 i = 0; i < sample_count; i++) {
         x = static_cast<s32>(static_cast<float>(input[i]) * gain);


src/audio_core/command_generator.cpp
```diff
@@ -41,6 +41,15 @@ void ApplyMix(s32* output, const s32* input, s32 gain, s32 sample_count) {
 }
 
 s32 ApplyMixRamp(s32* output, const s32* input, float gain, float delta, s32 sample_count) {
+    // XC2 passes in NaN mix volumes, causing further issues as we handle everything as s32 rather
+    // than float, so the NaN propogation is lost. As the samples get further modified for
+    // volume etc, they can get out of NaN range, so a later heuristic for catching NaNs is
+    // more difficult. Handle that here by setting these samples to silence.
+    if (gain != gain) {
```
lioncash, line 48:
```suggestion
    if (std::isnan(gain)) {
```

FernandoS27, line 48: foreadability use `std::isnan` wich emits literally this code but also safeguards it from fastmath and unsafe fp optimizations.

v1993:
1. ~~Have you considered using `std::isnan`?~~ Already commented on while writing this.
2. Does game actually produce NaNs on console too? Just want to be sure this isn't hiding a deeper bug.

FernandoS27:
> 1. ~Have you considered using `std::isnan`?~ Already commented on (twice!) while writing this.
> 2. Does game actually produce NaNs on console too? Just want to be sure this isn't hiding a deeper bug.

They could in theory send NaNs the question if they do or not is a sepparate one but it does make sense to ask it.

what I wonder is what does the Audio renderer do with NaN volumes. I suggest writting a simple test on libnx for that. (hmn does libnx support audren or only audio out?)

ogniK5377:
I'm pretty sure games shouldn't be sending NaNs, perhaps this should be checked against hardware with a test. It's most likely covering up a much bigger issue

Kelebek1:
For what it's worth, Ryuijinx has a more complete audio implementation, and they have this exact same behaviour of getting NaNs. They just get saved by their saturate function, where casting a NaN to a short gives them back 0.

lioncash:
Commit message should include the game the change is fixing

lioncash:
The game fix line is spliced into the main commit title. Consider adding a newline so that the fix line is part of the commit message body, rather than the title.

Kelebek1:
> 
> 
> The game fix line is spliced into the main commit title. Consider adding a newline so that the fix line is part of the commit message body, rather than the title.

Sorry, I didn't realise there were differences, I thought a message is just a message. It should be ok now.

bunnei (approved):

PR #6573, opened 2021-07-08
core,common,yuzu qt: Add CPU accuracy option 'Auto'

lat9nq:
The current CPU accuracy settings in yuzu are fairly polarized and require more than common knowledge to know what the optimal settings for yuzu would be. This adds a curated option called 'Auto' that applies a few at the moment known-good unsafe optimizations to Dynarmic.

Specifically, sets `Unfuse FMA`, `Faster ASIMD instructions` (32 bits only), `Inaccurate NaN handling`, and `Disable address space checks`.

Also adds migration code, since the old CPU Accuracy setting won't translate well here since we're adding one at the beginning of the list. On first boot with the new setting, just use the default setting.

Now moves the CPU debugging tab and the Debug tab as sub tabs of General -> Debug.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -93,7 +93,7 @@ bool IsGPULevelHigh() {
 }
 
 bool IsFastmemEnabled() {
-    if (values.cpu_accuracy.GetValue() == CPUAccuracy::DebugMode) {
+    if (values.cpu_debug_mode) {
         return static_cast<bool>(values.cpuopt_fastmem);
     }
     return true;
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -34,7 +34,6 @@ enum class CPUAccuracy : u32 {
     Auto = 0,
     Accurate = 1,
     Unsafe = 2,
-    DebugMode = 3,
 };
 
 /** The BasicSetting class is a simple resource manager. It defines a label and default value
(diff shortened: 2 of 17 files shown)


v1993:
I'm a little uncertain about this. Sure, having some almost-safe optimizations on by default can benefit users, but it also makes default choice not work for all games. Perhaps phrase it as "We recommend setting accuracy to "Auto" for most games." or something like that?

On that - I'd personally expect "Auto" to adjust setting automatically like name implies, probably based on some sort of game list, so maybe "Default" would be a better name for what it actually is.

It's probably a good idea to write about it in monthly report (even as about upcoming change) since once day people will boot SSBU and alike to find them misbehaving out of nowhere.

FernandoS27:
> I'm a little uncertain about this. Sure, having some almost-safe optimizations on by default can benefit users, but it also makes default choice not work for all games. Perhaps phrase it as "We recommend setting accuracy to "Auto" for most games." or something like that?
> 
> On that - I'd personally expect "Auto" to adjust setting automatically like name implies, probably based on some sort of game list, so maybe "Default" would be a better name for what it actually is.
> 
> It's probably a good idea to write about it in monthly report (even as about upcoming change) since once day people will boot SSBU and alike to find them misbehaving out of nowhere.

The unsafe options selected don't affect SSBU. Of those, Unfused FMA will only be unsafe if your CPU doesn't have native FMA. Either way you had to activate it if you had a very old CPU.

We decided to activate this is they benefit better most users and reducing the complexity of having to tell them to use unsafe when they don't have good enough hardware or the game uses something very heavy.

We mostly noticed other emulators not only of other consoles but also on the switch. Have their default settings on the lowest accuracy necessary. We decided on a balance instead.

v1993:
Makes sense, then. I feel like bisecting around this change is going to be painful tho.

PR #6575, opened 2021-07-08
Settings: Eliminate ASYNC & MULTICORE Toggles and add GPU Accuracy to  status bar

FernandoS27:
This Pull Request seeks to eliminate the MULTICORE and ASYNC toggles. This toggles were added because on their release they weren't too stable and caused issues in some games. Now-a-days almost none ever uses this toggles anymore, so we are removing them as part of our options simplification.

We are also adding a new toggle: GPU Accuracy. Which allows to switch between Normal and High GPU accuracy. This is meant for users who can't trade the accuracy for the performance. Normal, is pretty safe but breaks effects in many games.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2907,12 +2907,12 @@ void GMainWindow::UpdateStatusBar() {
 void GMainWindow::UpdateGPUAccuracyButton() {
     switch (Settings::values.gpu_accuracy.GetValue()) {
     case Settings::GPUAccuracy::Normal: {
-        gpu_accuracy_button->setText(tr("GPU NORMAL "));
+        gpu_accuracy_button->setText(tr("GPU NORMAL"));
         gpu_accuracy_button->setChecked(false);
         break;
     }
     case Settings::GPUAccuracy::High: {
-        gpu_accuracy_button->setText(tr("GPU HIGH   "));
+        gpu_accuracy_button->setText(tr("GPU HIGH"));
         gpu_accuracy_button->setChecked(true);
         break;
     }
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -789,41 +789,28 @@ void GMainWindow::InitializeWidgets() {
     dock_status_button->setChecked(Settings::values.use_docked_mode.GetValue());
     statusBar()->insertPermanentWidget(0, dock_status_button);
 
-    // Setup ASync button
-    async_status_button = new QPushButton();
-    async_status_button->setObjectName(QStringLiteral("TogglableStatusBarButton"));
-    async_status_button->setFocusPolicy(Qt::NoFocus);
-    connect(async_status_button, &QPushButton::clicked, [&] {
-        if (emulation_running) {
-            return;
+    gpu_accuracy_button = new QPushButton();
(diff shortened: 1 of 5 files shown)


goldenx86:
Almost looks like it's preliminary work for something else.

src/yuzu/main.cpp
```diff
@@ -2921,12 +2904,35 @@ void GMainWindow::UpdateStatusBar() {
...
+        gpu_accuracy_button->setChecked(false);
+        break;
+    }
+    case Settings::GPUAccuracy::High: {
+        gpu_accuracy_button->setText(tr("GPU HIGH   "));
+        gpu_accuracy_button->setChecked(true);
+        break;
+    }
+    case Settings::GPUAccuracy::Extreme: {
+        gpu_accuracy_button->setText(tr("GPU EXTREME"));
```
Morph1984, line 2920: The spacing here is rather weird, considering that the font may not be monospace
```suggestion
        gpu_accuracy_button->setText(tr("GPU NORMAL"));
        gpu_accuracy_button->setChecked(false);
        break;
    }
    case Settings::GPUAccuracy::High: {
        gpu_accuracy_button->setText(tr("GPU HIGH"));
        gpu_accuracy_button->setChecked(true);
        break;
    }
    case Settings::GPUAccuracy::Extreme: {
        gpu_accuracy_button->setText(tr("GPU EXTREME"));
```

v1993:
Is this new toggle mostly aimed at games that need different GPU accuracy in different places? Because I doubt they have any other uses now that we have per-game configuration.

I'm surprised this is getting such negative reception. Just to be clear: you still can switch multicore and async on and off to your heart's content! Just do this in global or per-game settings, okay?

> Almost looks like it's preliminary work for something else.

I feel like "something else" is called "Project Hades" :eyes:.

FernandoS27:
> Is this new toggle mostly aimed at games that need different GPU accuracy in different places? Because I doubt they have any other uses now that we have per-game configuration.
> 
> I'm surprised this is getting such negative reception. Just to be clear: you still can switch multicore and async on and off to your heart's content! Just do this in global or per-game settings, okay?
> 
> > Almost looks like it's preliminary work for something else.
> 
> I feel like "something else" is called "Project Hades" 👀.

The toggle, is just to make that option more visible as it's the bisggest discussion we normally have. It's also so people see what they are sacrificing for perf and when we get a report we know which GPu accuracy option they had right away.

It's not a real negative reception, the guys downvoting, we know them, aren't yuzu users. They are a bunch of Ryujinx fanatics that go around downvoting PRs, promoting Ryujinx in our discord, and causing trouble around. Just ignore them. Eloiri though is just trolling.

gidoly:
Seems changing accuracy on per game config while play game not correspond with toggle buttons

v1993:
Apparently, GPU accuracy toggle makes controllers vibrate for a moment when used. I think I understand how this issue came to be (changing setting toggles settings reload which makes input system check for vibration support which makes controllers vibrate briefly), but it probably should be fixed anyways (probably in a separate PR tho).

yzct12345:
> Now-a-days almost none ever uses this toggles anymore

Until #6689 gets merged, it would inconvenience me a bit to not have the ASYNC toggle when switching between Animal Crossing and Pokemon Let's Go

bunnei (approved):

PR #6576, opened 2021-07-08
settings: Disable FPS unlimit setting between title launches

ameerj:
Some titles crash if the FPS limit is disabled when launching. This change ensures that titles launch with the limit in-place to avoid issues.

In order to simplify the change, the UI toggle was removed as it would have always been overridden at launch to be disabled.
The setting can still be toggled during gameplay with the hotkey, and indicated by the fps label in the status bar.

### DIFF
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -327,7 +327,7 @@ struct Values {
     Setting<bool> use_nvdec_emulation{true, "use_nvdec_emulation"};
     Setting<bool> accelerate_astc{true, "accelerate_astc"};
     Setting<bool> use_vsync{true, "use_vsync"};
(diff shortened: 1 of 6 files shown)


v1993:
While it makes sense, this will limit discoverability of this feature considerably. On the other hand, it's not a feature your average user needs, so it's probably OK.

german77:
We don't recommend this setting at all. It's used mainly for performance testing

v1993:
@german77 I'm aware of that and that's why I'm sorta OK with this. But those who actually want to do performance testing still have to find out about it somehow.

Morph1984 (approved):

PR #6598, opened 2021-07-11
MacroHLE: Add MultidrawIndirect HLE Macro.

FernandoS27:
Currently MultidrawIndirect Macro is bugged in the MacroJit. This PR **does not** aim to have a high level construct of MultidrawIndirect on the host but to replicate the original macro in C++ and be more useful for debugging purposes.

This shouldn't be merged until macro jit is fixed. As this macro HLE will hide the bug. @ogniK5377 I'de love if you could have a look.

This fixes most rendering issues in Monster Hunter Rise. There are still some medium rendering issues.

### DIFF
--- a/src/video_core/macro/macro_hle.cpp
+++ b/src/video_core/macro/macro_hle.cpp
@@ -3,6 +3,8 @@
 
 #include <array>
 #include <vector>
+#include "common/scope_exit.h"
+#include "video_core/dirty_flags.h"
 #include "video_core/engines/maxwell_3d.h"
 #include "video_core/macro/macro.h"
 #include "video_core/macro/macro_hle.h"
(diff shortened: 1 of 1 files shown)


src/video_core/macro/macro_hle.cpp
```diff
@@ -77,12 +80,70 @@ void HLE_0217920100488FF7(Engines::Maxwell3D& maxwell3d, const std::vector<u32>&
     maxwell3d.CallMethodFromMME(0x8e5, 0x0);
     maxwell3d.mme_draw.current_mode = Engines::Maxwell3D::MMEDrawMode::Undefined;
 }
+
+// Multidraw Indirect
```
ReinUsesLisp, line 84:
```suggestion
// Multi draw elements indirect
```

src/video_core/macro/macro_hle.cpp
```diff
@@ -56,6 +58,7 @@ void HLE_0217920100488FF7(Engines::Maxwell3D& maxwell3d, const std::vector<u32>&
     maxwell3d.regs.index_array.first = parameters[3];
     maxwell3d.regs.reg_array[0x446] = element_base; // vertex id base?
     maxwell3d.regs.index_array.count = parameters[1];
+    maxwell3d.dirty.flags[VideoCommon::Dirty::IndexBuffer] = true;
```
v1993, line 61: What's the reason for this seemingly unrelated addition?

FernandoS27: This tells the buffer cache that it must recheck the index buffer configuration.

v1993: Is this change related to addition of HLE'd MultidrawIndirect macro or a separate fix?

FernandoS27: it's unique to the macro.

liamwhite (approved):

PR #6614, opened 2021-07-12
input_common/sdl: Implement HD rumble

german77:
This PR adds HD rumble support for the joycons and pro controllers. It needs a SDL version that supports HD rumble (https://github.com/german77/SDL/tree/hd_rumble)

This is a WIP just to demonstrate proof of concept. The encoding has some errors but works way better than the current on/off rumble support SDL has for the joycons/pro controllers. Once the error in the encoding is fixed and the official SDL repo supports HD rumble this can be added to yuzu.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -77,6 +77,14 @@ public:
         }
     }
 
+    bool HasHDRumble() const {
+        if (sdl_controller) {
+            return (SDL_GameControllerGetType(sdl_controller.get()) ==
+                    SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO);
+        }
+        return false;
+    }
+
     void SetButton(int button, bool value) {
         std::lock_guard lock{mutex};
         state.buttons.insert_or_assign(button, value);
@@ -160,8 +168,14 @@ public:
         return static_cast<float>(state.axes.at(axis)) / (32767.0f * range);
     }
 
-    bool RumblePlay(u16 amp_low, u16 amp_high) {
+    bool RumblePlay(u16 amp_low, u16 amp_high, u16 freq_low, u16 freq_high) {
         constexpr u32 rumble_max_duration_ms = 1000;
+
+        if (HasHDRumble()) {
+            const u64 effect =
+                (u64{amp_low} << 48) | (u64{amp_low} << 32) | (u64{amp_low} << 16) | freq_high;
+            return SDL_GameControllerSendEffect(sdl_controller.get(), &effect, sizeof(effect)) == 0;
+        }
 
         if (sdl_controller) {
             return SDL_GameControllerRumble(sdl_controller.get(), amp_low, amp_high,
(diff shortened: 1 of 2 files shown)


src/input_common/sdl/sdl_impl.cpp
```diff
@@ -77,6 +77,14 @@ class SDLJoystick {
         }
     }
 
+    bool HasHDRumble() {
+        if (sdl_controller) {
+            return (SDL_GameControllerGetType(sdl_controller.get()) ==
+                    SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO);
+        }
+        return false;
```
v1993, line 85: Does this actually support joycons? Because it looks like check is performed only for procon.

german77: SDL has the ugly way of returning that value. It actually refers to controllers that use the switch hidapi driver. Not the actual controller type

v1993: Understood. They really should have at least named it without `_PRO` part (I guess it always returns `UNKNOWN` on non-specific drivers? now that's terrible).

v1993: Another note - this check should probably ultimately end up in SDL code (I do realize it's PoC, but still). As far as I know there are some other controllers that support HD rumble out there, so leaving ability to support them would be nice.

german77: I would need to talk to them. I'm pretty sure they might have something on mind

Zangetsu38:
Squash last commit, no need create commit for cancel preview change and fix format ^^

v1993:
I'd prefer to wait until main branch of SDL gets HD rumble support back before merging, just to be sure. Can be made EA in the meantime tho.

german77:
It was actually merged then reverted aparently pro controller rumble is weak but joycons are fine. Just did a quick commit to test if it worked on yuzu

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -77,6 +77,14 @@ class SDLJoystick {
         }
     }
 
+    bool HasHDRumble() {
```
lioncash, line 80:
```suggestion
    bool HasHDRumble() const {
```

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -125,9 +133,16 @@ class SDLJoystick {
         return static_cast<float>(state.axes.at(axis)) / (32767.0f * range);
     }
 
-    bool RumblePlay(u16 amp_low, u16 amp_high) {
+    bool RumblePlay(u16 amp_low, u16 amp_high, u16 freq_low, u16 freq_high) {
         constexpr u32 rumble_max_duration_ms = 1000;
 
+        if (HasHDRumble()) {
+            const u64 effect = (static_cast<u64>(amp_low) << 48) |
```
lioncash, line 140:
```suggestion
            const u64 effect = (u64{amp_low} << 48) |
```

Ditto for the ones below, since this isn't a narrowing conversion, but a widening one.

german77:
HD rumble doesn't work well with Pro Controllers

eadmaster:
quick question: once this gets merged HD rumble will work without any specific OS driver/module on the supported Nintendo controllers?

v1993:
Yes, correct, that will be the case.

german77:
This PR will be reopened when ready. Closing it since it will be outdated

PR #6629, opened 2021-07-13
DMAEngine: Accelerate BufferClear [accelerateDMA Part 2]

FernandoS27:
This part adds BufferClears to accelerateDMA for both Vulkan and OGL.

This fixes the black textures on some characters caused by faulty normals (the compute shader messed up as the buffer wasn't clear on time).

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -329,6 +329,8 @@ private:
 
     [[nodiscard]] bool HasFastUniformBufferBound(size_t stage, u32 binding_index) const noexcept;
 
+    void ClearDownload(IntervalType subtract_interval);
+
     VideoCore::RasterizerInterface& rasterizer;
     Tegra::Engines::Maxwell3D& maxwell3d;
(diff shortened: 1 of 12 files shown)


src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -781,6 +822,27 @@ bool BufferCache<P>::IsRegionGpuModified(VAddr addr, size_t size) {
     return false;
 }
 
+template <class P>
+bool BufferCache<P>::IsRegionRegistered(VAddr addr, size_t size) {
```
lioncash, line 826:
```suggestion
bool BufferCache<P>::IsRegionRegistered(VAddr addr, size_t size) const {
```

lioncash (commented):
LGTM other than one nit

src/video_core/buffer_cache/buffer_cache.h
FernandoS27: can't be sadly, due to the access on the slot vector having side effects.

bunnei (approved):

PR #6639, opened 2021-07-14
general: Reduce compile time / linker usage on MSVC

Morph1984:
This PR aims to reduce compilation time by reducing the usage of the linker "cl.exe" when linking files with identical filenames (controller.cpp, interface.cpp, module.cpp, etc)

Given that the linker is single threaded, this significantly impacts the compilation time on MSVC.

These changes provide a significant reduction in compilation time (1 minute savings) when tested on my machine, which has a Ryzen 5 1600 6 core CPU at 3.7 GHz.

Before:
yuzu full build: 9 minutes and 32 seconds
yuzu build with precompiled externals: 6 minutes and 59 seconds

After:
yuzu full build: 8 minutes and 24 seconds
yuzu build with precompiled externals: 5 minutes and 52 seconds

It is uncertain whether this helps other compilers like GCC or Clang, testing will be much appreciated


lioncash (approved):

v1993:
Are they using hash table or something with file names as keys internally?

v1993:
Tested on GCC with Gold linker and LTO - no difference at all. Linking took exactly 5m 42s both before and after.

PR #6647, opened 2021-07-15
cmake: Be more specific with Qt

lat9nq:
As-is causes issues with building yuzu using MinGW GCC on Linux-based machines. Only set the variable when needed.

Also, we currently search for Qt on Linux regardless of the YUZU_USE_BUNDLED_QT setting. Disable the initial search if that variable is set to ON.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -254,7 +254,9 @@ if(ENABLE_QT)
 
     # Check for system Qt on Linux, fallback to bundled Qt
     if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
-        find_package(Qt5 ${QT_VERSION} COMPONENTS Widgets)
+        if (NOT YUZU_USE_BUNDLED_QT)
+            find_package(Qt5 ${QT_VERSION} COMPONENTS Widgets)
+        endif()
         if (NOT Qt5_FOUND OR YUZU_USE_BUNDLED_QT)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -354,8 +356,10 @@ if(ENABLE_QT)
         endif()
 
         set(QT_PREFIX_HINT HINTS "${QT_PREFIX}")
+
+        set(YUZU_QT_NO_CMAKE_SYSTEM_PATH "NO_CMAKE_SYSTEM_PATH")
     endif()
-    find_package(Qt5 ${QT_VERSION} REQUIRED COMPONENTS Widgets ${QT_PREFIX_HINT} NO_CMAKE_SYSTEM_PATH)
+    find_package(Qt5 ${QT_VERSION} REQUIRED COMPONENTS Widgets ${QT_PREFIX_HINT} ${YUZU_QT_NO_CMAKE_SYSTEM_PATH})
```
lioncash, line 362: In the event the conditional above is entered, this would be attempting to read out of an undefined variable wouldn't it? If so, this will cause a warning when configuring cmake with the `--warn-uninitialized` option.

lat9nq: Thanks, I didn't know that was an option with CMake. On my machine this warning comes up 128 more times when enabled, so this seems like something I can look into another time.

bunnei (approved):

PR #6204, opened 2021-04-16
externals: Add SDL 2.0.12

lat9nq:
Since Bintray is (soon to be) no more and has been removed from yuzu, there needs to be a way to acquire updated SDL2. Since Ubuntu 20.04's version is older than our minimum required version (2.0.12), add it as an external.

Fixes #6203

PR #6652, opened 2021-07-16
yuzu-cmd: Linux Vulkan fixes

lat9nq:
Likely due to a bug introduced in #6204 and #6293, `yuzu-cmd`'s Vulkan backend reports an unimplemented window manager and exits. This was due to including SDL.h which includes a dummy SDL_config.h by default.

Solution here is to manually include the generated SDL_config.h before SDL.h does. This changes `YUZU_ALLOW_SYSTEM_SDL2` to `YUZU_USE_EXTERNAL_SDL2` and now assumes that if the variable is ON, then we are using the external, else we are using the system SDL installation.

In addition, this adds more informative else cases when the reported window manager *does* have an implementation, but support was not compiled in.

As a followup to f973274b99d6683d257d77b5d2fd1994853ec3a7, we can remove the workaround for Clang as it has been fixed upstream by libsdl-org/SDL@25fc40b.


CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ project(yuzu)
 option(ENABLE_SDL2 "Enable the SDL2 frontend" ON)
 CMAKE_DEPENDENT_OPTION(YUZU_USE_BUNDLED_SDL2 "Download bundled SDL2 binaries" ON "ENABLE_SDL2;MSVC" OFF)
 # On Linux system SDL2 is likely to be lacking HIDAPI support which have drawbacks but is needed for SDL motion
-option(YUZU_ALLOW_SYSTEM_SDL2 "Try using system SDL2 before fallling back to one from externals" OFF)
+CMAKE_DEPENDENT_OPTION(YUZU_USE_EXTERNAL_SDL2 "Compiled external SDL2" ON "ENABLE_SDL2;NOT MSVC" OFF)
```
v1993, line 16: Did you mean "Compile"? Also, IMO, "Use SDL2 from externals" would be better.

iWeaker4you:
Thanks, it fixes many games crash with vulkan

bunnei (approved):

PR #6654, opened 2021-07-16
input_common: Make button threshold customizable

### LINKED ISSUE #6644: [Feature Request] Let us decide the deadzone of controller triggers (right now the input is registered at 80% of pressure)
Hello! Basically Yuzu (both latest stable version and Early Access version) register trigger inputs almost at full pressure (around 80% more or less) and it's quite annoying.
I've made a video to easily show the behavior (it was too big to attach as a file):
https://www.youtube.com/watch?v=Lj9K-dMxtFs

Here a video from another user showing it live in a game with an Xbox One controller:
https://streamable.com/t0c95s

I've tried with the Dualshock 4 too and it's the same, it's something related to how Yuzu handle inputs. 

The fact that ZL and ZR are digital inputs makes the matter even more annoying, since the soon you read the input the better, the fact that you have to press the triggers almost all the way down gives an unresponsive feedback.

I think that this really need a specific function to adjust the deadzone of the triggers to decide if inputs are going to be registered as soon you press it or at specific grade of pressure, especially since all other aspect of controller support are highly customizable, it would be really useful to have something for triggers too, thanks.

german77:
Kills two birds with one stone:
- Analog triggers from SDL2 had a range from -1.0 to 1.0 we where setting the trigger to 0.5 which is around 75% of the button pressed. This PR fixes the SDL threshold value to 50%.
- Adds an UI option on right click to set any threshold value for each button

Closes #6644

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -660,7 +660,8 @@ public:
 
         if (params.Has("axis")) {
             const int axis = params.Get("axis", 0);
-            const float threshold = params.Get("threshold", 0.5f);
+            // Convert range from (0.0, 1.0) to (-1.0, 1.0)
+            const float threshold = (params.Get("threshold", 0.5f) - 0.5f) * 2.0f;
             const std::string direction_name = params.Get("direction", "");
             bool trigger_if_greater;
             if (direction_name == "+") {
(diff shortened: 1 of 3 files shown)


Struggleton:
The ability to right click on the button and set the threshold is not obvious. Maybe to keep in line with the rest of the UI, add a slider under the triggers/bumpers in each of their groupboxes. (Mockup below)

sliderMockup

Optimally the sliders would adapt to the selected theme

german77:
I like the idea but it will be a little complicated.  Let me se what I can do.

bunnei (approved):

PR #6658, opened 2021-07-16
bootmanager: Create a dummy render widget

Morph1984:
This ensures that Qt positions the render window at the correct position on initializing the respective render backends.

Fixes the black bars / offset render window when starting yuzu on a secondary monitor,

### DIFF
--- a/src/yuzu/bootmanager.cpp
+++ b/src/yuzu/bootmanager.cpp
@@ -567,6 +567,12 @@ std::unique_ptr<Core::Frontend::GraphicsContext> GRenderWindow::CreateSharedCont
 bool GRenderWindow::InitRenderTarget() {
     ReleaseRenderTarget();
 
+    {
+        // Create a dummy render widget so that Qt
+        // places the render window at the correct position.
+        const RenderWidget dummy_widget{this};
+    }
+
     first_frame = false;
 
     switch (Settings::values.renderer_backend.GetValue()) {


german77:
I can confirm this fixes the issue

src/yuzu/bootmanager.cpp
```diff
@@ -564,6 +564,15 @@ std::unique_ptr<Core::Frontend::GraphicsContext> GRenderWindow::CreateSharedCont
 bool GRenderWindow::InitRenderTarget() {
     ReleaseRenderTarget();
 
+    {
+        // Create a dummy render widget so that Qt
+        // places the render window at the correct position.
+        auto dummy_widget = new RenderWidget(this);
+        dummy_widget->windowHandle()->create();
+        dummy_widget->deleteLater();
+        dummy_widget = nullptr;
```
ameerj, line 573: The fix for this issue can be simplified to the following one-liner.
```suggestion
        const RenderWidget dummy_widget(this);
```

ameerj (commented):
This smells more like a work around. Might be worth investigating a proper fix without needing a new `RenderWidget` created

Morph1984:
While it may be a workaround, Qt works in mysterious ways with loads of oddities here and there, so this is the only simplest solution that I could come up with.

bunnei (approved):

PR #6659, opened 2021-07-16
input_common: Fix mouse panning behaivour

german77:
#6579 altered the default values for this setting. This PR corrects them and on top of that fixes an issue with multiple screens where the joystick was always pulled down

### DIFF
--- a/src/yuzu/bootmanager.cpp
+++ b/src/yuzu/bootmanager.cpp
@@ -411,8 +411,9 @@ void GRenderWindow::mousePressEvent(QMouseEvent* event) {
     if (event->source() == Qt::MouseEventSynthesizedBySystem) {
         return;
     }
-
-    auto pos = event->pos();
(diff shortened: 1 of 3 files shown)


src/yuzu/bootmanager.cpp
```diff
@@ -411,8 +411,9 @@ void GRenderWindow::mousePressEvent(QMouseEvent* event) {
     if (event->source() == Qt::MouseEventSynthesizedBySystem) {
         return;
     }
-
-    auto pos = event->pos();
+    // Qt sometimes returns the parent coordinates. To avoid this we read the global mouse
+    // coordinates and map them to the current render area
+    auto pos = mapFromGlobal(QCursor::pos());
```
ameerj, line 416: can pos be made `const`?

ameerj (approved):

PR #6667, opened 2021-07-17
ci,linux: Support Patreon releases

lat9nq:
Soft-requires yuzu-emu/liftinstall#28

The Early Access AppImage needs to be accessible through liftinstall, so a couple modifications need to made:

The DIR_NAME needs to not include the revision info.
The EA AppImage name cannot contain revision info.
The EA AppImage has to be packaged with the rest of the yuzu package, which means both binaries and the source are bundled with it now in an archive.

In addition, fix the source archive so yuzu can actually be built from it.

This also makes the yuzu mainline releases compatible with liftinstall and its shortcuts.

### DIFF
--- a/.ci/scripts/linux/upload.sh
+++ b/.ci/scripts/linux/upload.sh
@@ -20,7 +20,9 @@ fi
 mkdir "$DIR_NAME"
 
 cp build/bin/yuzu-cmd "$DIR_NAME"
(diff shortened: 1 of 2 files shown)


iMonZ:
Is it possible that this approach will work with flatpak?
Since a lot of users preferring the flatpak version instead of the uncomfortable AppImage version

Morph1984 (approved):

PR #6671, opened 2021-07-18
applets/web: Addressing QT Navigation issues in Linux

jls47:
The QT web applet was having issues with navigation in Super Mario Odyssey's Action Guide.  Neither face button presses nor analog stick movement were being picked up.  By highlighting the first available clickable element and dealing with the button value mysteriously changing before the press can be fully acknowledged, the problem seems to have been resolved in SMO at least.  Could use testing in Windows and other games, though.

### DIFF
--- a/src/yuzu/applets/qt_web_browser.cpp
+++ b/src/yuzu/applets/qt_web_browser.cpp
@@ -107,6 +107,7 @@ void QtNXWebEngineView::LoadLocalWebPage(const std::string& main_url,
     is_local = true;
 
     LoadExtractedFonts();
+    FocusFirstLinkElement();
     SetUserAgent(UserAgent::WebApplet);
     SetFinished(false);
     SetExitReason(Service::AM::Applets::WebExitReason::EndButtonPressed);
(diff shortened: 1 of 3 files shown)


src/yuzu/applets/qt_web_browser.cpp
```diff
@@ -364,6 +367,20 @@ void QtNXWebEngineView::LoadExtractedFonts() {
...
+    QWebEngineScript focus_link_element;
+
+    constexpr char FOCUS_LINK_ELEMENT_SCRIPT[] =
+        R"(document.getElementsByTagName("a")[0].focus();)";
+
+    focus_link_element.setName(QStringLiteral("focus_link_element.js"));
+    focus_link_element.setSourceCode(QString::fromStdString(FOCUS_LINK_ELEMENT_SCRIPT));
+    focus_link_element.setWorldId(QWebEngineScript::MainWorld);
+    focus_link_element.setInjectionPoint(QWebEngineScript::Deferred);
+    focus_link_element.setRunsOnSubFrames(true);
```
v1993, line 380: Is this line needed? If it is, probably add check into script that at least one element was found.

jls47: I can't quite tell which line you're referring to-I mostly followed what had been done elsewhere in the file.  That said, I'll add a handler to the script to make sure it doesn't just become a js error factory.

v1993: I've been referring to the last line, that's how github reviews show it. It should be a fairly trivial check anyways.

Morph1984 (approved):

PR #6679, opened 2021-07-18
Fix Pokemon Let's Go on Vulkan

yzct12345:
fixes the crash after catching the first Pokemon. it was doing strange stuff with textures coming back outside. ignoring the invalid operation fixes the vulkan crash and doesn't cause any visual glitches

### DIFF
--- a/src/video_core/renderer_vulkan/vk_texture_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_texture_cache.cpp
@@ -608,7 +608,10 @@ void TextureCacheRuntime::BlitImage(Framebuffer* dst_framebuffer, ImageView& dst
     const VkImageAspectFlags aspect_mask = ImageAspectMask(src.format);
     const bool is_dst_msaa = dst.Samples() != VK_SAMPLE_COUNT_1_BIT;
     const bool is_src_msaa = src.Samples() != VK_SAMPLE_COUNT_1_BIT;
-    ASSERT(aspect_mask == ImageAspectMask(dst.format));
+    if (aspect_mask != ImageAspectMask(dst.format) {
+        UNIMPLEMENTED_MSG("Incompatible blit from format {} to {}", src.format, dst.format));
+        return;
+    }
     if (aspect_mask == VK_IMAGE_ASPECT_COLOR_BIT && !is_src_msaa && !is_dst_msaa) {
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -608,7 +608,10 @@ void TextureCacheRuntime::BlitImage(Framebuffer* dst_framebuffer, ImageView& dst
     const VkImageAspectFlags aspect_mask = ImageAspectMask(src.format);
     const bool is_dst_msaa = dst.Samples() != VK_SAMPLE_COUNT_1_BIT;
     const bool is_src_msaa = src.Samples() != VK_SAMPLE_COUNT_1_BIT;
-    ASSERT(aspect_mask == ImageAspectMask(dst.format));
+    if (aspect_mask != ImageAspectMask(dst.format) {
```
vitor-k, line 611:
```suggestion
    if (aspect_mask != ImageAspectMask(dst.format)) {
```

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -608,7 +608,10 @@ void TextureCacheRuntime::BlitImage(Framebuffer* dst_framebuffer, ImageView& dst
...
+        UNIMPLEMENTED_MSG("Incompatible blit from format {} to {}", src.format, dst.format));
```
vitor-k, line 612:
```suggestion
        UNIMPLEMENTED_MSG("Incompatible blit from format {} to {}", src.format, dst.format);
```

vitor-k (commented):
You don't need to close and open a new PR on every change, just push to the branch and the PR will be updated.

yzct12345:
> You don't need to close and open a new PR on every change, just push to the branch and the PR will be updated.

Yeah sorry i accidentally deleted the branch in the other PR somehow and couldn't reopen it

FernandoS27:
wow cool, nice catch. Mind if I add it to EA for quick testing before merging?

It's not a proper solution, but it's better to print an ugly assert than to crash.

FernandoS27:
Nvm, it's fine I'll just merge.

Thank you very much for your contribution.

PR #6683, opened 2021-07-20
input/sdl_impl: fix rumble support on DualSense.

Zangetsu38:
# About:
- fix rumble on DualSense
- Dualsense return 78 value for using rumble, return 0 only if both frequency is 0, and after check function diferent value is normal

value return can be different 0, is not error is normal, error is only -1.

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -130,10 +130,10 @@ public:
 
         if (sdl_controller) {
             return SDL_GameControllerRumble(sdl_controller.get(), amp_low, amp_high,
-                                            rumble_max_duration_ms) == 0;
+                                            rumble_max_duration_ms) != -1;
         } else if (sdl_joystick) {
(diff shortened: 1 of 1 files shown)


FernandoS27 (approved):

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -126,14 +126,12 @@ class SDLJoystick {
     }
 
     bool RumblePlay(u16 amp_low, u16 amp_high) {
-        constexpr u32 rumble_max_duration_ms = 1000;
```
Morph1984, line 129: I'm curious, why was this removed?

Zangetsu38: I restore later, apparently some game is stupid coded and no return 0 after rumble finish 🤦‍♂️🤷‍♂️

Zangetsu38:
i have restored max duration, some stupid game is apparently bad coded and no send 0 in frequency for stop vibration after it finished

Morph1984 (approved):

PR #6685, opened 2021-07-20
gl_texture_cache: Workaround slow PBO downloads on radeonsi

ReinUsesLisp:
There's an optimization bug on non-git mesa versions where not
specifying GL_CLIENT_STORAGE_BIT causes very slow reads on the CPU
side.

Add this bit for all vendors.

- Fixes performance issues on radeonsi on Skyward Sword (from ~8 FPS to 60 FPS)


FernandoS27 (approved):

ReinUsesLisp:
This also improved other games, like The Legend of Zelda: Breath of the Wild. Basically any game where we download textures.

PR #6687, opened 2021-07-20
vk_swapchain: Manually wait semaphores on Intel Mesa

yzct12345:
Not a single game works on my Linux laptop without this pull request. Everything was stuck on a black screen at 0 CPU. I don't know what's wrong with the Intel driver, but this makes things playable for me.


ReinUsesLisp:
Do you have Nvidia's driver installed?

yzct12345:
No, I don't have a Nvidia GPU. I have a laptop without only Intel

PacoAvelar:
Bf

ReinUsesLisp:
Sorry for the delay. I don't like taking hiding the error without investigating why it's happening exactly, and fixing that on yuzu or ANV. If it turns out to be a driver bug, then add the workaround for those specific driver versions.

Running ANV on my local machine I noticed we hit an assertion about the semaphore type. I don't have a log at hand.

yzct12345:
The problem is this block in the ANV driver https://github.com/mesa3d/mesa/blob/dff0d9911d176802b54890c796e19f56c50f24e1/src/intel/vulkan/anv_wsi.c#L303-L341 Device drivers and kernel code are much past my ability so I decided to work around it in yuzu instead.

@PacoAvelar If you don't want to build you can set `ANV_QUEUE_THREAD_DISABLE` to see something but it crashes later on.

yzct12345:
This affects multiples users on Discord. How about "do-no-merge" and "early-access-merge" like the other PRs that hide the error?

v1993:
I think "do-not-merge" and "mainline-merge" would be more appropriate.

ameerj:
Superseded by #6953

PR #6688, opened 2021-07-20
render_vulkan: Fix validation errors on less compatible Intel GPUs

yzct12345:
This fixes some log spam when the Vulkan validation layers are enabled on many Intel GPUs. I haven't noticed any changes in game.

Many Intel GPU don't have enough vertex input bindings or attributes. 32 is required but I only have 28. It's a good thing that most games don't use all of them, so we can get away with having just 28 when we don't have 32. By avoiding calling Vulkan when we know it will fail, we see less error messages.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_buffer_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_buffer_cache.cpp
@@ -285,6 +285,9 @@ void BufferCacheRuntime::BindQuadArrayIndexBuffer(u32 first, u32 count) {
 
 void BufferCacheRuntime::BindVertexBuffer(u32 index, VkBuffer buffer, u32 offset, u32 size,
                                           u32 stride) {
+    if (index >= device.GetMaxVertexInputBindings()) {
+        return;
+    }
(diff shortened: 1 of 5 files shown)


src/video_core/renderer_vulkan/vk_buffer_cache.cpp
```diff
@@ -182,6 +182,9 @@ void BufferCacheRuntime::BindQuadArrayIndexBuffer(u32 first, u32 count) {
 
 void BufferCacheRuntime::BindVertexBuffer(u32 index, VkBuffer buffer, u32 offset, u32 size,
                                           u32 stride) {
+    [[unlikely]] if (index >= device.GetMaxVertexInputBindings()) {
```
lioncash, line 185:
```suggestion
    if (index >= device.GetMaxVertexInputBindings()) {
```

It's preferable for this to only be used on paths where it's been benchmarked that it provides notable benefits

Moonlacer:
Is this still a valid PR, even now? It seems yzct has disappeared, so I was wondering if it could still be merged even without the CLA

liamwhite (approved):

PR #6689, opened 2021-07-20
texture_cache: Allow disabling preemptive texture downloads

yzct12345:
Preemptive texture downloads is causing me 10 FPS of lag. With it disabled, Animal Crossing finally runs at 30 FPS.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -54,6 +54,8 @@ void LogSettings() {
     log_setting("Renderer_GPUAccuracyLevel", values.gpu_accuracy.GetValue());
     log_setting("Renderer_UseAsynchronousGpuEmulation",
                 values.use_asynchronous_gpu_emulation.GetValue());
+    log_setting("Renderer_EnablePreemptiveDownloads",
+                values.enable_preemptive_downloads.GetValue());
     log_setting("Renderer_UseNvdecEmulation", values.use_nvdec_emulation.GetValue());
     log_setting("Renderer_AccelerateASTC", values.accelerate_astc.GetValue());
     log_setting("Renderer_UseVsync", values.use_vsync.GetValue());
@@ -137,6 +139,7 @@ void RestoreGlobalState(bool is_powered_on) {
     values.use_disk_shader_cache.SetGlobal(true);
     values.gpu_accuracy.SetGlobal(true);
     values.use_asynchronous_gpu_emulation.SetGlobal(true);
+    values.enable_preemptive_downloads.SetGlobal(true);
     values.use_nvdec_emulation.SetGlobal(true);
     values.accelerate_astc.SetGlobal(true);
     values.use_vsync.SetGlobal(true);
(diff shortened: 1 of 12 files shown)


v1993:
I'm not sure yuzu team will be happy with exposing such rather low-level setting to users. Is this a driver-specific issue?

yzct12345:
This improves the FPS for both my Intel integrated laptop and my AMD+AMD desktop.

I made this a setting to not force people to use this in case this made yuzu slower for other people. The preemptive download feature is also a bit big and it would be rude for me to just delete all of it. If you can confirm, I could do this instead.

FernandoS27:
hmn, we'll consider it but maybe not in this way.

v1993:
I haven't tested this much yet, but so far (Nvidia with project hades, vulkan) this seems to destabilize framerate a tiny bit (112-114 FPS instead of flat-line 113 FPS), but that's it. May be because I'm mostly CPU capped, however.

yzct12345:
> this seems to destabilize framerate a tiny bit (112-114 FPS instead of flat-line 113 FPS), but that's it

Well, if it doesn't improve things for your setup, then that might justify the setting to keep it enabled.

On the other hand, we could also interpret that another way. Since it's only 1 FPS difference, this "preemptive texture download" feature has quite a big maintenance cost in comparison to a little gain. Removing it should ease the burden for you.

> I'm mostly CPU capped

My AMD desktop is also CPU capped. It went from around 9 FPS to around 17 FPS.

> project hades

After merging project hades but not this PR, I get 26 FPS instead of 19 FPS at the title screen. I see large spikes in "Wait for the GPU" in Microprofile together with the lag. It was at 17.0 average. With this PR, it averages at 0 and I get 30 FPS, and I can even underclock to 2 Ghz CPU and 600 Mhz GPU to stop my laptop from throttling after 10 minutes. What happens if you try that? If that doesn't cause this problem, then it might be my PCI bandwidth is too low.

I am using Vulkan, async, multicore and dock. I first noticed the problem when I saw turning off async surprisingly makes my game run faster. This PR removes the surprise.

FernandoS27:
I'm planning in the future to replace Extreme for Custom mode which allows you to decide exactly which heavy features you want.

There's 2 kinds of users the ones that don't want to deal with settings and those that want full control. For those that don't want to deal with settings: High will be good enough for everything.

v1993:
I'm not sure what are you asking me to check. Since I'm CPU capped, lowering CPU frequency will lead to decrease in performance and underclocking GPU won't. It does seem like a good guess that you're getting bandwidth-limited here tho. Also, just to be clear, I was not testing AC: NH (which I don't have) but rather other games which probably don't use this feature as much (but do use it since there is a measurable effect).

yzct12345:
> I'm planning in the future to replace Extreme for Custom mode which allows you to decide exactly which heavy features you want.

Closing as it seems like you would like to fix the problem in a different way.

> I'm not sure what are you asking me to check

Never mind. As the problem seems to affect nobody else in the community, I will just keep this as a patch on my computer locally.

PR #6698, opened 2021-07-23
input_common: Improve SDL joystick and hide toggle option

### LINKED ISSUE #6455: Joycons drift when selecting input device Dual joy-con 0?

### LINKED ISSUE #6555: [BUG REPORT] Pro Controller Left Stick is offset a bit to the left and down
Yuzu 676
The Pro Controller Left stick is offset for a bit.
Yuzu Controller
image
Taken at the same time on Cemu (Not Drifting/ Haven't Touched Controller between Images)
image

I can recreate this for both my pro controllers and even my fake Wii U pro controller, with the exact same result.
While you might want to ride this off as joystick drift, I have two new pro controllers (From Nintendo) and have already calibrated them on the switch and using windows 10 Xbox calibration software. This issue is not seen in steam games, or in other emulators like Cemu. I do remember the Cemu emulator having the same issue, but was fixed in an update (https://github.com/Davidobot/BetterJoy/issues/2) that was the issue. Might be the same issue with Yuzu.

EDIT: I already know a Higher dead zone will fix this

german77:
Some QoL changes:
- Hides the toggle button on all buttons that don't support this setting
- Avoids resetting the SDL controller every time the mappings are edited
- Adds a property to set a desired SDL joystick center
- The auto mapper will set the joystick center automatically

Closes #6555
Closes #6455

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -82,6 +82,12 @@ public:
         state.buttons.insert_or_assign(button, value);
     }
 
+    void PreSetButton(int button) {
+        if (!state.buttons.contains(button)) {
+            SetButton(button, false);
+        }
+    }
+
     void SetMotion(SDL_ControllerSensorEvent event) {
         constexpr float gravity_constant = 9.80665f;
         std::lock_guard lock{mutex};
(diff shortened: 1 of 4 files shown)


src/input_common/sdl/sdl_impl.cpp
```diff
@@ -555,6 +571,8 @@ class SDLAnalog final : public Input::AnalogDevice {
     const bool invert_y;
     const float deadzone;
     const float range;
+    const float offset_x;
+    const float offset_y;
```
ameerj, line 575: I'd prefer at this point we have a `CoordPair` struct for pairs of x and y values, to more compactly pass around these values.
(feel free to use a different name that's more appropriate for the general usage)

bunnei (approved):

PR #6705, opened 2021-07-24
yuzu qt: Disable compositor on Linux

yzct12345:
SDL and most professional games do this to improve performance. This makes a difference especially on Kubuntu with a lot of effects enabled competing for the GPU. It works basically like how fullscreen on Windows removes the compositor middleman and lets the game access the screen directly.


behunin:
Shouldn't this only happen while yuzu is Fullscreen?
What about Wayland?

yzct12345:
> Shouldn't this only happen while yuzu is Fullscreen?

The SDL games I played did this even in windowed mode. I noticed only when my desktop effects stopped working when the game is open. If they only did this in fullscreen I wouldn't have know about this. In fact, some compositors might even choose to do this by default when an application is fullscreen, making this unnecessary for those users.

> What about Wayland?

Oops, forgot a line check to prevent a crash there. Fixed it.

Wayland kind of insists on doing compositing all the time. Even though gamers have complained, they still won't add this feature. They will just get the old and slow behavior then.

> Could not find a package configuration file provided by "Qt5X11Extras" with
  any of the following names

Can I get some help on fixing the build script please?

CMakeLists.txt
```diff
@@ -361,6 +361,14 @@ if(ENABLE_QT)
 
         set(YUZU_QT_NO_CMAKE_SYSTEM_PATH "NO_CMAKE_SYSTEM_PATH")
     endif()
+
+    if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
```
lioncash, line 365: This should probably go into the Qt-specific CMakeLists file and link directly to the frontends target, so we aren't making X11 and Qt's X11 extras a link dependency on every library in the project.

yzct12345: > This should probably go into the Qt frontend-specific CMakeLists file and link directly to the frontend target

Fixed

> making X11

I tried doing that but I realized that I wrote the future #6713 PR with this still merged locally. It needs X11 to be included or else yuzu-cmd fails to link with some FFmpeg error.

yzct12345: Fully fixed

dnmodder:
@yzct12345 At least in Nvidia, I have a way to know if the compositor is enabled or disabled while I have Yuzu in full screen, with the option "Enable Graphics API Visual Indicator", when the compositor is enabled it says "BLIT" and when I disable it change to "FLIP", I use KDE Plasma (KWin), I don't know if everyone with Nvidia could reproduce this.

v1993:
@dnmodder I can't reproduce this, most likely because I have two screens. Seeing if compositor is on or off under KDE is still fairly easy since a lot of effects break.

iMonZ:
> Shouldn't this only happen while yuzu is Fullscreen?
> What about Wayland?

For me, Wayland just gives me a translucent screen with this error:
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()

Any information on why this is happening and how this can be fixed?
And I found this: 
https://github.com/flameshot-org/flameshot/issues/619
https://github.com/flameshot-org/flameshot/pull/1272

yzct12345:
> GetAvailableNetworkInterfaces returned no interfaces

Yes. @dnmodder Just turn on desktop cube, wobbly windows, and transparent Konsole.

> qt.qpa.wayland: Wayland does not support QWindow::requestActivate()

Are you sure if you did `QT_QPA_PLATFORM=wayland ./yuzu`? Otherwise does anything show up if you run an "emulated" wayland. For example if you use X11 and run `weston` then run `yuzu` inside that.

And is does this problem occur on Mainline/EA or does it only happen with my changes? My changes should only activate if it detects yuzu is using X11.

iMonZ:
> > GetAvailableNetworkInterfaces returned no interfaces
> 
> Yes. @dnmodder Just turn on desktop cube, wobbly windows, and transparent Konsole.
> 
> > qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
> 
> Are you sure if you did `QT_QPA_PLATFORM=wayland ./yuzu`? Otherwise does anything show up if you run an "emulated" wayland. For example if you use X11 and run `weston` then run `yuzu` inside that.
> 
> And is does this problem occur on Mainline/EA or does it only happen with my changes? My changes should only activate if it detects yuzu is using X11.

Using xWayland yuzu works (kinda) 
But on full Wayland, I get this warning.
And the yuzu app either runs the game but with a translucent view or just crashes.

iMonZ:
I found here a fix for the Wayland issue: https://github.com/RPCS3/rpcs3/commit/4953e79588efd69b4ffe6f3afff46f4d3ee7cf2a

Tatsh:
I used to do this with a Gamemode hook but since tweaking my Nvidia driver settings and Plasma, I find it entirely unnecessary. In particular:

```
 $ cat /home/tatsh/.config/plasma-workspace/env/gl-yield.sh
export __GL_YIELD=USLEEP
# export KWIN_COMPOSE=O2
# export KWIN_OPENGL_INTERFACE=egl
export KWIN_TRIPLE_BUFFER=1
export __GL_MaxFramesAllowed=1
# export KWIN_DRM_USE_EGL_STREAMS=1
```

```
 $ cat /etc/X11/xorg.conf.d/99-nvidia.conf
Section "Device"
    Identifier     "Device0"
    Driver         "nvidia"
    VendorName     "NVIDIA Corporation"
    BoardName      "NVIDIA GeForce RTX 3080 Ti"
    Option         "TripleBuffer" "On"
    Option         "RegistryDwords"  "RMUseSwI2c=0x01; RMI2cSpeed=100"
EndSection
```

I would prefer this be a setting than automatic. Switching compositing on and off is simply error-prone and I try to avoid it.

bunnei:
@yzct12345 any plans for this one? Tatsh posted a suggestion above that might be relevant. Thanks!

noneistaken:
> What about Wayland?
Nobody cares, on Wayland it's impossible to turn the compositor off in any degree.

PR #6713, opened 2021-07-25
nvdec: Implement VA-API hardware video acceleration

yzct12345:
This off-loads the expensive video operations to the GPU. The Animal Crossing, Pokemon Let's Go, and both Kirby Star Allies intro movies now play very smoothly at or above 30 FPS.


yzct12345:
Depends on CMakeLists changes from #6705

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,6 +42,48 @@ Codec::~Codec() {
     av_frame_unref(av_frame);
     av_free(av_frame);
     avcodec_close(av_codec_ctx);
+    av_buffer_unref(&av_hw_device);
+}
+
+// Hardware acceleration code from FFmpeg/doc/examples/hw_decode.c
+// under MIT license
+#if defined(__linux__)
+static enum AVPixelFormat get_hw_format(AVCodecContext* ctx, const enum AVPixelFormat* pix_fmts) {
```
lioncash, line 51:
```suggestion
static AVPixelFormat GetHWFormat(AVCodecContext* ctx, const AVPixelFormat* pix_fmts) {
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,6 +42,48 @@ Codec::~Codec() {
...
+    if (Settings::values.nvdec_emulation.GetValue() != Settings::NvdecEmulation::Vaapi) {
+        return;
+    }
+
+#if defined(__linux__)
+    const int hwdevice_error =
+        av_hwdevice_ctx_create(&av_hw_device, AV_HWDEVICE_TYPE_VAAPI, nullptr, nullptr, 0);
+    if (hwdevice_error < 0) {
+        LOG_ERROR(Service_NVDRV, "av_hwdevice_ctx_create failed {}", hwdevice_error);
+        goto fail;
```
lioncash, line 72: It's fine to inline the fail code in both if statements here instead of goto, particularly given no resource needs to be freed or managed.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -99,12 +142,32 @@ void Codec::Decode() {
     packet.data = frame_data.data();
     packet.size = static_cast<s32>(frame_data.size());
 
-    avcodec_send_packet(av_codec_ctx, &packet);
+    int ret = avcodec_send_packet(av_codec_ctx, &packet);
```
lioncash, line 145:
```suggestion
    const int ret = avcodec_send_packet(av_codec_ctx, &packet);
```

And make the other assignment use its own local variable, since the results aren't related to one another.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -99,12 +142,32 @@ void Codec::Decode() {
...
+    ASSERT_MSG(!ret, "avcodec_send_packet error {}", ret);
 
     if (!vp9_hidden_frame) {
         // Only receive/store visible frames
-        AVFramePtr frame = AVFramePtr{av_frame_alloc(), AVFrameDeleter};
-        avcodec_receive_frame(av_codec_ctx, frame.get());
+        AVFrame* hw_frame = av_frame_alloc();
```
lioncash, line 150: This doesn't seem like it gets freed on non-linux platforms based off a quick glance of everything.

yzct12345: Isn't this this original `av_frame_alloc()` left side of the `AVFramePtr frame = AVFramePtr{av_frame_alloc(), AVFrameDeleter}`, which gets freed by `AVFrameDeleter`?

There are 1 or 2 frames here depending on whether hardware acceleration is enabled or disabled. First we `hw_frame = av_frame_alloc()` the first one. If hardware acceleration is disabled, it becomes `sw_frame` directly, and we never allocate another one. `AVFrameDeleter` frees this first frame in this case. If it is enabled on the other hand, we treat `hw_frame` as a temporary buffer. We allocate another frame named `sw_frame`. The second frame is needed as the first frame is in the wrong place because the first frame is on the GPU instead of the CPU. The first frame `hw_frame` is then freed because it's a temporary in this case. Here `AVFrameDeleter` frees the second frame instead. This is exactly what the FFmpeg example did.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: Applied separately since this is a name change

src/video_core/command_classes/codecs/codec.cpp
yzct12345: Inlined separately since this involves multiple lines. Also simplified a bit after thinking about this

src/video_core/command_classes/codecs/codec.cpp
yzct12345: Fixed the 3 places with a separate commit

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,6 +42,44 @@ Codec::~Codec() {
...
+#endif
+
+void Codec::InitializeHwdec() {
+    if (Settings::values.nvdec_emulation.GetValue() != Settings::NvdecEmulation::Vaapi) {
+        return;
+    }
+
+#if defined(__linux__)
+    const int hwdevice_error =
+        av_hwdevice_ctx_create(&av_hw_device, AV_HWDEVICE_TYPE_VAAPI, nullptr, nullptr, 0);
```
behunin, line 69: [device cannot be nullptr](https://ffmpeg.org/doxygen/3.1/hwcontext_8c.html#a21fbd088225e4e25c4d9a01b3f5e8c51)

yzct12345: I don't see anything in that link saying it can't be nullptr. I ported this code from the example at https://github.com/FFmpeg/FFmpeg/blob/f7958d0883286479a6c9cd91e68e506dfa02ed7c/doc/examples/hw_decode.c#L54

CMakeLists.txt
```diff
@@ -595,18 +605,18 @@ if (YUZU_USE_BUNDLED_FFMPEG)
                     --disable-avformat
                     --disable-doc
                     --disable-everything
-                    --disable-ffmpeg
```
v1993, line 598: This seems like an unrelated change? What's the reason for it?

CMakeLists.txt
```diff
@@ -584,7 +584,17 @@ if (YUZU_USE_BUNDLED_FFMPEG)
...
         # `configure` parameters builds only exactly what yuzu needs from FFmpeg
-        # `--disable-{vaapi,vdpau}` is needed to avoid linking issues
+        # `--disable-vdpau` is needed to avoid linking issues
+
+        if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+            pkg_check_modules(LIBVA REQUIRED)
+            pkg_check_modules(LIBVA-DRM REQUIRED)
+            pkg_check_modules(LIBVA-X11 REQUIRED)
+            link_libraries(va va-drm va-x11)
+            set(FFmpeg-hwaccel-flags --enable-hwaccel=h264_vaapi --enable-hwaccel=vp9_vaapi)
```
v1993, line 594: This should be `FFmpeg_HWACCEL_FLAGS` to stay consistent with naming.

v1993 (commented):
This is a really nice addition! Next step is probably to use host NVDEC which works both on Windows and Linux Nvidia cards.

CMakeLists.txt
yzct12345: Oops. I accidentally left in some debug stuff. It helped me find out that I need to delete the build directory when I change FFmpeg flags. Fixed.

CMakeLists.txt
yzct12345: Fixed in 4 places.

v1993: Not quite fixed - please replace dashes with underscores (feel free to `git commit --amend` and then force push, no need for new commit). Dashes are used in target names, underscores are for variables.

yzct12345: > please replace dashes with underscores

I'm so blind. Hopefully I actually fixed it this time.

> feel free to git commit --amend and then force push

Unfortunately my git setup is somewhat broken and I'm stuck using the web version. You can consider squashing all my commits when merging instead.

ameerj:
Thanks for taking the time to implement this. I do have a few questions.
 - Last I checked, the FFmpeg documentation stated that HW acceleration through VA API is only supported on Intel GPUs. Has this changed to work on AMD/NVIDIA GPUs as well?
 - Did you test titles which write RGB frames? The titles you listed all write YUV420 if memory serves right. Link's Awakening and Paper Mario The Origami King both write RGB frames, which is worth testing with this change.
 - Lastly, what are your hardware specs and yuzu configurations? and what was the performance of these videos before this change? All of the users on the yuzu discord that reported low performance decoding video, had even lower performance in-game. And our own internal testing of HW accelerated video decoding showed little to no performance difference, indicating the bottleneck was elsewhere in emulation (granted this was a year ago and things have changed)

v1993:
> Last I checked, the FFmpeg documentation stated that HW acceleration through VA API is only supported on Intel GPUs. Has this changed to work on AMD/NVIDIA GPUs as well?

Not for Nvidia (I'll get on it once this gets merged tho, I hope), but [it does work for AMD](https://wiki.archlinux.org/title/Hardware_video_acceleration).

yzct12345:
> HW acceleration through VA API is only

A detailed matrix is at https://trac.ffmpeg.org/wiki/HWAccelIntro

> Did you test titles which write RGB frames

I tested Link's Awakening and it was a slideshow both before and after. Sadly I don't have Paper Mario The Origami King.

> Lastly, what are your hardware specs and yuzu configurations? and what was the performance of these videos before this change

I will unmark as draft after more detailed benchmarking and looking into async NVDEC.

> indicating the bottleneck was elsewhere in emulation

The bottleneck for Link's Awakening is indeed somewhere else. If I turn off NVDEC it's still just as slow but it crashes.

> low performance

My laptop is neither CPU nor GPU bottlenecked. The problem is that it's thermal throttling. I need to underclock my CPU to 2 Ghz or else it will underclock itself to 400 Mhz after 10 minutes and everything suddenly becomes 4 FPS. I mainly wrote this PR to reduce CPU power usage so I can increase the clock to 2.5 Ghz and let everything play faster.

For Kirby, I get these results:

| What | Software | VA-API |
|-|-|-|
| avcodec_send_packet | 9 ms | 0 ms |
| avcodec_receive_frame | 0 ms | 0 ms |
| av_hwframe_transfer_data | N/A | 4 ms |
| Vic::Execute | 1 ms | 0 ms |

This is on my laptop with an integrated GPU. CPU-GPU latency makes it worse on a dedicated GPU as expected:

| What | Software | VA-API |
|-|-|-|
| avcodec_send_packet | 15 ms | 0 ms |
| avcodec_receive_frame | 0 ms | 0 ms |
| av_hwframe_transfer_data | N/A | 6 ms |
| Vic::Execute | 3 ms | 3 ms |

yzct12345:
Here are the benchmark results. They are listed as a range with outliers removed. I am using Vulkan, async, multicore with Normal GPU and Unsafe CPU. I will not be using dock as it makes Link's Awakening much slower.

**A certain Intel Ice Lake i7 mobile CPU only:**

I use this for gameplay, but I underclock it or else it will overheat. I got 30 FPS for Kirby with 2.5 Ghz, but I will use just 2 Ghz so that the results are comparable for both software and VA-API. I will also have the GPU set to 300 Mhz but that shouldn't matter.

| What | Software | VA-API | Off
|-|-|-|-|
| Let's Go Pikachu! | 25-27 FPS | 27-29 FPS | 27-29 FPS |
| Kirby Let's Be Friends | 23-28 FPS | 27-29 FPS | 28-29 FPS |
| Kirby Starfall: A Rude Awakening | 18-27 FPS | 26-29 FPS | 28-30 FPS |
| Link's Awakening | 9-19 FPS | 10-18 FPS | 7-9 FPS |
| ACNH | 24-26 FPS | 25-28 FPS | 26-28 FPS |

Link's Awakening's FPS on the bottom left doesn't seem to be accurate. It looks much lower in reality. With VA-API, the video looks a bit smoother. In both cases, I see blank frames inserted often. I also see less corrupt video in ACNH and Let's Go.

**A certain AMD Kaveri desktop CPU and a certain AMD GCN 4th gen GPU:**

Not a single game is playably fast enough on this desktop. I'm only using it to test AMD support. It also doesn't support VP9 VA-API.

| What | Software | VA-API | Off |
|-|-|-|-|
| Let's Go Pikachu! | 20-22 FPS | 22-26 FPS (no video) | 23-24 FPS |
| Kirby Let's Be Friends | 16-20 FPS | 20-22 FPS | 23-24 FPS |
| Kirby Starfall: A Rude Awakening | 13-20 FPS | 20-22 FPS | 22-24 FPS |
| Link's Awakening | 5-14 FPS | 5-7 FPS (no video) | 4-6 FPS |
| ACNH | 16-19 FPS then Unmapped Read32 and 0 FPS | 18-21 FPS (no video) then arm_interface abort and 30 FPS | 14-23 FPS |

Another reason the bottleneck is somewhere else is that we drop frames after 10 pile up. That hides any bottleneck and turns it into dropped frames. I see a few black frames in Kirby Star Allies with software and no black frames with VA-API.

> had even lower performance in-game

This shouldn't happen. If I put `abort()` in `Vic::Execute()` and `Codec::Decode()` they are not run when in-game. Therefore I can't see how the changes in this PR could have caused the low in-game performance.

v1993:
> This shouldn't happen. If I put abort() in Vic::Execute() and Codec::Decode() they are not run when in-game. Therefore I can't see how the changes in this PR could have caused the low in-game performance.

I'm fairly sure Ameer was talking about old tests and meant "if video decoding was slow, game were unplayably slow anyways".

yzct12345:
I made `Codec::Decode()` async.

29 FPS becomes 30 FPS more often for VA-API. Software decoding is around 4 FPS faster and the range is exactly the same as the one for VA-API.

Link's Awakening still crashes occassionally but is 10 FPS slower with the async. It spends 10 milliseconds in `Vic::Execute()` instead of 0 or 1 milliseconds. The bottleneck is whatever's calling `Vic::Execute` isn't calling it frequently enough. This causes a lot of decoded frames to be dropped and wasted. Making it async might help but I don't know how to move the `gpu.MemoryManager().Read<u64>(config_struct_address + 0x20)` part to another thread safely.

> hardware specs

Also try commenting out the `sw_frame->format = AV_PIX_FMT_NV12`. If that breaks video it means your hardware prefers a different format and we should implement it in vic.cpp.

> if video decoding was slow, game were unplayably slow anyways

Ah, I'm right on the boundary between playable and unplayable. I get 30 FPS or just below that most of the time. Each small optimization will make it playable more and more often.

behunin:
Looking at the [firefox implementation](https://searchfox.org/mozilla-central/source/dom/media/platforms/ffmpeg/FFmpegLibWrapper.cpp#300) va-api is behind a wayland ifdef.
Ubuntu is moving to wayland by default in 21.10 so this is not far off for Nvidia users.

yzct12345:
> va-api is behind a wayland ifdef

Wow I just tested with Wayland and was surprised it actually works even though I didn't specify libva-wayland in CMakeLists. I am on Ubuntu 21.10 but I always use X11 because Wayland breaks half of my software

v1993:
@behunin VA-API requires driver-level support. Just moving to wayland won't make Nvidia cards VA-API capable.

behunin:
> @behunin VA-API requires driver-level support. Just moving to wayland won't make Nvidia cards VA-API capable.

470.xx will support xwayland and rumors are that GBM support is in progress.

ameerj (changes requested):
> Each small optimization will make it playable more and more often.

To an extent. This PR is growing far beyond an integration of VA-API accelerated decoding. I would suggest you move any change not specifically related to hardware accelerated decoding to another PR, not only are they out of scope, they also make regressions harder to catch.

I tested the artifacts of this build:
- Mario Golf had a color shifted video on first launch that I haven't been able to reproduce in subsequent runs. Likely due to the async changes.
- Similarly, Paper Mario's intro video was flipped and grayscale in one launch, but it's hard to reproduce.
- Resident Evil 6's intro video, which suffered from off-by-one errors when writing its final frame during nvdec development, now has a flickering video.

Keep in mind this testing was done on Windows with an Nvidia GPU. So these are regressions to the already existing code.

I'm also very uneasy about drivers that partially/do not support VA-API, and videos simply not rendering as a result. I'd prefer a fallback to CPU decoding when an unsupported driver/codec is being used with VA-API.

v1993:
@behunin 470 driver is out by now (it's not even in beta). Hardware-accelerated video decoding via VA-API is very, very different from accelerated rendering support, be that xwayland or GBM. I'm not aware of any relation between Wayland and VA-API (VA-API works on X11 just fine, as far as I'm aware), actually, but please correct me if I'm wrong.

behunin:
> @behunin 470 driver is out by now (it's not even in beta). Hardware-accelerated video decoding via VA-API is very, very different from accelerated rendering support, be that xwayland or GBM. I'm not aware of any relation between Wayland and VA-API (VA-API works on X11 just fine, as far as I'm aware), actually, but please correct me if I'm wrong.
> 
> Additionally, Nvidia cards support accelerated video decoding via NVDEC and VDPAU. Later isn't well-maintained and lacks formats yuzu needs, as far as I'm aware. NVDEC, however, is supported both on Linux and Windows (and Switch, which is how we ended up here in the first place) and is actively maintained.

Your correct there is no relation.
Va-api and wayland support are coming this year, though that has not officially been promised yet. Xwayland is just a seeping stone for now but with the steam deck doing both wayland and va-api on AMD, nvidia will be feeling the pressure. That's what I have heard.

`vainfo` gives me:
```
libva info: VA-API version 1.10.0
libva info: Trying to open /usr/lib/x86_64-linux-gnu/dri/nouveau_drv_video.so
libva info: Found init function __vaDriverInit_1_10
libva info: va_openDriver() returns 0
vainfo: VA-API version: 1.10 (libva 2.10.0)
vainfo: Driver version: Mesa Gallium driver 21.0.1 for NV124
vainfo: Supported profile and entrypoints
      VAProfileNone                   :	VAEntrypointVideoProc
```
I am on X11 with the 470 drivers, running a GTX 980
I don't think va-api works on X11 with 470 drivers, for me at least.

NVDEC works for Maxwell (GM206) and newer cards, so for me (GM204) it's a no go. For verification I followed [this guide](https://medium.com/@multi.flexi/how-to-get-info-about-nvdec-and-nvenc-capabilities-of-your-gpu-6ecc9a5d4dac) and got errors like no supported gpu.

v1993:
@behunin Even if Nvidia decides to support VA-API (so, leaving speculations aside):

1. NVDEC works on Windows, VA-API does not. There are cross-vendor APIs on windows as well, but that's a different story.
2. VA-API is just that, an API to hardware features. If NVDEC does not work on your card for some format VA-API also won't.

src/video_core/command_classes/vic.cpp
```diff
@@ -131,38 +131,57 @@ void Vic::Execute() {
...
-            const std::size_t dst = y * aligned_width;
-            for (std::size_t x = 0; x < frame_width; ++x) {
-                luma_buffer[dst + x] = luma_ptr[src + x];
+        {
+            const u8* src = frame->data[0];
+            u8* dst = luma_buffer.data();
+            for (std::size_t y = frame_height; y--;) {
+                memcpy(dst, src, frame_width);
+                src += stride;
+                dst += aligned_width;
```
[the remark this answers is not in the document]
yzct12345: I reverted this.

> bugged with off-by-one errors

I experienced the same problems applying and reverting this change.

> almost no difference on a competent compiler

Somehow I measured a difference. With this change this function takes 0-1 milliseconds. After reverting it things take 2-3 milliseconds. This will be a bottleneck if we ever want to attempt 60 FPS.

yzct12345:
> artifacts of this build

I hope that is fixed now the async is reverted. I'll wait before making a new PR in case you want to squash.

> do not support VA-API

`av_hwdevice_ctx_create` should fail and things will already fall back to CPU decoding. I don't have a Nvidia GPU, so I can't test directly. What I did is try replacing `AV_HWDEVICE_TYPE_VAAPI` with `AV_HWDEVICE_TYPE_VDPAU` and the fallback to CPU worked properly.

> partially

Fixed. I spent a few hours on this and thought it was hard but it turned out to be a single line of code.

> if Nvidia decides to support VA-API

There's another thing about VA-API I didn't mention. It doesn't have to be the same GPU as the one for rendering. I'm not even specifying which GPU to use as it shouldn't matter. We can do this because we are copying the data to host memory and back anyway. It might even be faster on some setups if the GPU is the bottleneck.

Most people have Intel+Nvidia. Enabling the integrated GPU just for video should be an option.

@behunin Are you also thermal throttling? If not you don't even need any video acceleration at all. The async changes I just reverted make the CPU decoding framerate just as fast as VA-API. They will come back in the next PR.

NVDEC will still be necessary for Windows however.

behunin:
I am just evaluating my situation with VA-API and speculating on when it might be supported for nvidia users. I think there needs to be a check for nvidia devices along with the setting check. We will see when and if nvidia supports VA-API.

NVDEC will be necessary for nvidia users on any OS.

CMakeLists.txt
```diff
@@ -584,7 +584,17 @@ if (YUZU_USE_BUNDLED_FFMPEG)
...
         # `configure` parameters builds only exactly what yuzu needs from FFmpeg
-        # `--disable-{vaapi,vdpau}` is needed to avoid linking issues
+        # `--disable-vdpau` is needed to avoid linking issues
+
+        if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+            pkg_check_modules(LIBVA REQUIRED)
+            pkg_check_modules(LIBVA-DRM REQUIRED)
+            pkg_check_modules(LIBVA-X11 REQUIRED)
+            link_libraries(va va-drm va-x11)
+            set(FFmpeg_HWACCEL_FLAGS --enable-hwaccel=h264_vaapi --enable-hwaccel=vp9_vaapi)
```
lat9nq, line 594: We didn't need libva before, and we have a software fallback. Why are we requiring it now?

yzct12345: libva is VA-API. It's the whole purpose of this PR.

CMakeLists.txt
```diff
@@ -584,7 +584,17 @@ if (YUZU_USE_BUNDLED_FFMPEG)
...
 
         # `configure` parameters builds only exactly what yuzu needs from FFmpeg
-        # `--disable-{vaapi,vdpau}` is needed to avoid linking issues
+        # `--disable-vdpau` is needed to avoid linking issues
+
+        if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+            pkg_check_modules(LIBVA REQUIRED)
+            pkg_check_modules(LIBVA-DRM REQUIRED)
+            pkg_check_modules(LIBVA-X11 REQUIRED)
+            link_libraries(va va-drm va-x11)
```
lat9nq, line 593: This should be appended to the FFmpeg_LIBRARIES variable, not generally linked to everything.

I would replace FFmpeg_LIBRARIES with another variable, FFmpeg_BUILD_LIBRARIES and then set FFmpeg_LIBRARIES to `${LIBVA_DRM_LIBRARIES} ${LIBVA_X11_LIBRARIES} ${LIBVA_LIBRARIES} ${X11_LIBRARIES} ${FFmpeg_BUILD_LIBRARIES}`... not necessarily in that order.

CMakeLists.txt
```diff
@@ -584,7 +584,17 @@ if (YUZU_USE_BUNDLED_FFMPEG)
             CACHE PATH "Path to FFmpeg headers" FORCE)
 
         # `configure` parameters builds only exactly what yuzu needs from FFmpeg
-        # `--disable-{vaapi,vdpau}` is needed to avoid linking issues
+        # `--disable-vdpau` is needed to avoid linking issues
+
+        if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+            pkg_check_modules(LIBVA REQUIRED)
+            pkg_check_modules(LIBVA-DRM REQUIRED)
+            pkg_check_modules(LIBVA-X11 REQUIRED)
```
lat9nq, line 592: This requires the x11 package but it is not linked.

lat9nq, line 592: `pkg_check_modules` requires using FindPkgConfig but it is not checked for before use.

lat9nq, line 592: libva-x11 requires x11 but it is not linked.

```suggestion
            pkg_check_modules(LIBVA libva)
            pkg_check_modules(LIBVA_DRM libva-drm)
            pkg_check_modules(LIBVA_X11 libva-x11)
            pkg_check_modules(X11 x11)
```

Then use `<prefix>-FOUND` and `<prefix>-LIBRARIES` as appropriate.

CMakeLists.txt
lat9nq: Sorry, i was using `Add Single Comment` when I was trying to do a review.

The point is that this requires libva -- I understand that this is the point of the PR but we should not require libva on people who wish to build yuzu or cannot build yuzu without libva installed.

lat9nq:
This is using the software fallback on my system, regardless of the primary renderer. It appears that it's trying to use the first NVIDIA GPU then fails (as I believe the 980 Ti is `/dev/dri/renderD128`), ignoring the other two capable GPUs here.
```
System:    Host: Qykopi Kernel: 5.10.53-183-tkg-pds x86_64 bits: 64 compiler: gcc v: 11.1.0 Desktop: GNOME 40.3 
           tk: GTK 3.24.30 wm: gnome-shell dm: GDM, LightDM Distro: Arch Linux 
CPU:       Info: 6-Core model: 11th Gen Intel Core i5-11600K bits: 64 type: MT MCP arch: Rocket Lake rev: 1 cache: 
           L2: 12 MiB 
           flags: avx avx2 lm nx pae sse sse2 sse3 sse4_1 sse4_2 ssse3 vmx bogomips: 93975 
           Speed: 3177 MHz min/max: 800/5000 MHz Core speeds (MHz): 1: 3177 2: 4827 3: 4843 4: 4846 5: 4633 6: 4857 
           7: 4856 8: 4878 9: 4896 10: 4861 11: 4841 12: 4890 
Graphics:  Device-1: Intel RocketLake-S GT1 [UHD Graphics 750] vendor: ASUSTeK driver: i915 v: kernel 
           bus-ID: 00:02.0 chip-ID: 8086:4c8a 
           Device-2: NVIDIA GM200 [GeForce GTX 980 Ti] vendor: ZOTAC driver: nvidia v: 470.42.01 bus-ID: 01:00.0 
           chip-ID: 10de:17c8 
           Device-3: Advanced Micro Devices [AMD/ATI] Tonga PRO [Radeon R9 285/380] vendor: XFX Pine driver: amdgpu 
           v: kernel bus-ID: 04:00.0 chip-ID: 1002:6939 
           Device-4: NVIDIA GK208B [GeForce GT 730] vendor: PNY driver: nvidia v: 470.42.01 bus-ID: 06:00.0 
           chip-ID: 10de:1287 
           Display: wayland server: X.Org 1.21.1.2 compositor: gnome-shell driver: loaded: amdgpu,modesetting,nvidia 
           alternate: ati,fbdev,intel,nouveau,nv,vesa resolution: 1: 1920x1200~60Hz 2: 2560x1440~144Hz s-dpi: 96 
           OpenGL: renderer: Mesa Intel Graphics (RKL GT1) v: 4.6 Mesa 21.1.5 direct render: Yes 
```

CMakeLists.txt
yzct12345: Fixed

CMakeLists.txt
yzct12345: I have vlc installed and it pulled libva in so I thought everyone had it.

I fixed this and it's now optional.

CMakeLists.txt
yzct12345: So that's why LIBVA_LIBRARIES wasn't working for me before.

Fixed.

yzct12345:
> It appears that it's trying to use the first NVIDIA GPU then fails

I think I fixed this. I set it to drm instead of x11 and it should try 8 GPUs.

I set it to prefer Intel because it's integrated and will avoid a PCI copy. This benefits Intel integrated, Intel-AMD, Intel-Nvidia, AMD integrated and AMD-Nvidia users. Checking for AMD-AMD will take too much extra code so it'll be a random GPU.

I would have added a setting for the device path instead but the yuzu team doesn't seem to want more.

src/common/settings.h
```diff
@@ -341,7 +347,7 @@ struct Values {
     Setting<bool> use_disk_shader_cache{true, "use_disk_shader_cache"};
     Setting<GPUAccuracy> gpu_accuracy{GPUAccuracy::High, "gpu_accuracy"};
     Setting<bool> use_asynchronous_gpu_emulation{true, "use_asynchronous_gpu_emulation"};
-    Setting<bool> use_nvdec_emulation{true, "use_nvdec_emulation"};
+    Setting<NvdecEmulation> nvdec_emulation{NvdecEmulation::Software, "nvdec_emulation"};
```
ameerj, line 350: I would prefer this to remain as a boolean toggle, and hardware acceleration to be used whenever supported by the host hardware without any user intervention.
The project prefers having things "just work" out of the box with as little tinkering as possible.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+
+// Hardware acceleration code from FFmpeg/doc/examples/hw_decode.c under MIT license
+#if defined(LIBVA_FOUND)
+static AVPixelFormat GetHwFormat(AVCodecContext*, const AVPixelFormat* pix_fmts) {
+    for (const AVPixelFormat* p = pix_fmts; *p != AV_PIX_FMT_NONE; ++p) {
+        if (*p == AV_PIX_FMT_VAAPI) {
+            return AV_PIX_FMT_VAAPI;
+        }
+    }
+    LOG_ERROR(Service_NVDRV, "VA-API does not support this format, falling back to CPU");
```
ameerj, line 55:
```suggestion
    LOG_INFO(Service_NVDRV, "Could not find compatible GPU AV format, falling back to CPU");
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
     av_frame_unref(av_frame);
     av_free(av_frame);
     avcodec_close(av_codec_ctx);
+    av_buffer_unref(&av_hw_device);
+}
+
+// Hardware acceleration code from FFmpeg/doc/examples/hw_decode.c under MIT license
+#if defined(LIBVA_FOUND)
+static AVPixelFormat GetHwFormat(AVCodecContext*, const AVPixelFormat* pix_fmts) {
```
ameerj, line 49: Prefer using an anonymous namespace over `static` functions to maintain style consistency with other parts of video_core.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+        const int hwdevice_error = av_hwdevice_ctx_create(av_hw_device, AV_HWDEVICE_TYPE_VAAPI,
+                                                          nullptr, hwdevice_options, 0);
+        if (hwdevice_error >= 0) {
+            LOG_INFO(Service_NVDRV, "Using VA-API with {}", driver);
+            av_dict_free(&hwdevice_options);
+            return;
+        }
+        LOG_DEBUG(Service_NVDRV, "av_hwdevice_ctx_create failed {}", hwdevice_error);
+    }
+    LOG_ERROR(Service_NVDRV, "av_hwdevice_ctx_create failed for all drivers");
```
ameerj, line 78:
```suggestion
    LOG_WARNING(Service_NVDRV, "av_hwdevice_ctx_create failed for all drivers");
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -96,33 +210,20 @@ void Codec::Decode() {
...
-        // Limit queue to 10 frames. Workaround for ZLA decode and queue spam
-        if (av_frames.size() > 10) {
-            av_frames.pop();
-        }
-    }
+    RawFrame raw_frame{
+        .frame_data = frame_data,
+        .vp9_hidden_frame = vp9_hidden_frame,
+    };
+    // TODO async
```
ameerj, line 217: what's this in reference to?

src/video_core/command_classes/codecs/codec.h
```diff
@@ -55,18 +54,27 @@ class Codec {
...
+    AVBufferRef* av_hw_device{nullptr};
     AVCodecContext* av_codec_ctx{nullptr};
 
     GPU& gpu;
     const NvdecCommon::NvdecRegisters& state;
     std::unique_ptr<Decoder::H264> h264_decoder;
     std::unique_ptr<Decoder::VP9> vp9_decoder;
 
-    std::queue<AVFramePtr> av_frames{};
+    boost::lockfree::spsc_queue<AVFrame*, boost::lockfree::capacity<10>> av_frames{};
```
ameerj, line 77: spsc_queue is not needed. Multiple threads never interact with this queue.
And FYI the common project holds a spsc queue implementation.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+    case AV_PIX_FMT_YUV420P:
+    case AV_PIX_FMT_NV12:
+        break;
+    default:
+        UNIMPLEMENTED_MSG("Unknown video format from host graphics: {}", sw_frame->format);
+        av_frame_free(&sw_frame);
+        return;
+    }
+
+    if (!av_frames.push(sw_frame)) {
```
ameerj, line 151: Why is `av_frames` a queue of `AVFrame*`? Revert it back to being in an `AVFramePtr`.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+        LOG_TRACE(Service_NVDRV, "av_frames.push overflow dropped frame");
+        av_frame_free(&sw_frame);
```
ameerj, line 153: The intention of this queue was to remove the oldest frame, not the newest frame.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+        av_buffer_unref(&av_hw_device);
+        return;
+    }
+    av_codec_ctx->get_format = GetHwFormat;
+#else
+    UNIMPLEMENTED_MSG("Hardware acceleration without Linux");
+#endif
+}
+
+void Codec::ActuallyDecode(RawFrame& raw_frame) {
```
ameerj, line 101:
```suggestion
void Codec::DecodeImpl(const RawFrame& raw_frame) {
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+    AVPacket packet{};
+    av_init_packet(&packet);
+    packet.data = raw_frame.frame_data.data();
+    packet.size = static_cast<s32>(raw_frame.frame_data.size());
+
+    const int send_packet_ret = avcodec_send_packet(av_codec_ctx, &packet);
+    ASSERT_MSG(!send_packet_ret, "avcodec_send_packet error {}", send_packet_ret);
```
ameerj, line 108: An assert is not appropriate here. Due to the limitations of how we decode videos, it is very common for this to fail. Please remove this.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+
+    // Only receive/store visible frames
+    if (raw_frame.vp9_hidden_frame) {
+        return;
+    }
+
```
ameerj, line 114: unnecessary newline

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+    AVFrame* hw_frame = av_frame_alloc();
```
ameerj, line 115: This should only be allocated if GPU acceleration is being used.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +41,121 @@ Codec::~Codec() {
...
+    AVFrame* sw_frame = hw_frame;
+    ASSERT_MSG(hw_frame, "av_frame_alloc hw_frame failed");
+    const int receive_frame_ret = avcodec_receive_frame(av_codec_ctx, hw_frame);
+    ASSERT_MSG(!receive_frame_ret, "avcodec_receive_frame error {}", receive_frame_ret);
```
ameerj, line 119: Again, due to limitations with our decoding, errors like this are common.

src/video_core/command_classes/vic.cpp
```diff
@@ -83,16 +79,17 @@ void Vic::Execute() {
             sws_freeContext(scaler_ctx);
             scaler_ctx = nullptr;
 
-            // FFmpeg returns all frames in YUV420, convert it into expected format
-            scaler_ctx =
-                sws_getContext(frame->width, frame->height, AV_PIX_FMT_YUV420P, frame->width,
-                               frame->height, target_format, 0, nullptr, nullptr, nullptr);
+            // Software return YUV420 and VA-API returns NV12
+            // Convert it to RGB
```
ameerj, line 83:
```suggestion
            // Frames are decoded into either YUV420 or NV12 formats. Convert to desired format
```

src/video_core/command_classes/vic.cpp
```diff
@@ -83,16 +79,17 @@ void Vic::Execute() {
...
+            scaler_ctx = sws_getContext(frame->width, frame->height,
+                                        static_cast<AVPixelFormat>(frame->format), frame->width,
+                                        frame->height, target_format, 0, nullptr, nullptr, nullptr);
 
             scaler_width = frame->width;
             scaler_height = frame->height;
         }
         // Get Converted frame
-        const std::size_t linear_size = frame->width * frame->height * 4;
+        const std::size_t linear_size = static_cast<u32>(frame->width * frame->height * 4);
```
ameerj, line 92:
```suggestion
        const auto linear_size = static_cast<std::size_t>(frame->width * frame->height * 4);
```

src/video_core/command_classes/vic.cpp
```diff
@@ -131,41 +129,57 @@ void Vic::Execute() {
...
+            for (std::size_t y = 0; y < half_height; ++y) {
+                const std::size_t src = y * half_stride;
+                const std::size_t dst = y * aligned_width;
+
+                for (std::size_t x = 0; x < half_width; ++x) {
+                    chroma_buffer[dst + x * 2] = chroma_b_src[src + x];
+                    chroma_buffer[dst + x * 2 + 1] = chroma_r_src[src + x];
+                }
+            }
+        } else {
```
ameerj, line 170: I would prefer an explicit `else if (frame->format == AV_PIX_FMT_NV12)

src/video_core/command_classes/vic.h
```diff
@@ -19,11 +19,11 @@ class Vic {
 public:
     enum class Method : u32 {
         Execute = 0xc0,
-        SetControlParams = 0x1c1,
+        // SetControlParams = 0x1c1,
```
ameerj, line 22: Either remove the line or revert the diff

src/video_core/command_classes/vic.h
```diff
@@ -19,11 +19,11 @@ class Vic {
...
         SetConfigStructOffset = 0x1c2,
         SetOutputSurfaceLumaOffset = 0x1c8,
-        SetOutputSurfaceChromaUOffset = 0x1c9,
-        SetOutputSurfaceChromaVOffset = 0x1ca
+        SetOutputSurfaceChromaOffset = 0x1c9,
+        // SetOutputSurfaceChromaUnusedOffset = 0x1ca
```
ameerj, line 26: ditto

src/video_core/command_classes/vic.h
```diff
@@ -44,8 +44,8 @@ class Vic {
     union VicConfig {
         u64_le raw{};
         BitField<0, 7, u64_le> pixel_format;
-        BitField<7, 2, u64_le> chroma_loc_horiz;
-        BitField<9, 2, u64_le> chroma_loc_vert;
+        // BitField<7, 2, u64_le> chroma_loc_horiz;
+        // BitField<9, 2, u64_le> chroma_loc_vert;
```
ameerj, line 48: Keep these for completeness sake

ameerj (changes requested):

src/common/settings.h
yzct12345: The enum will be soon necessary for switching between NVDEC, VA-API and Software depending on which is faster for the user's computer.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: It refers to the async worker thread you told me to move to the next PR because of regressions.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: It's immediately renamed to `sw_frame` if GPU acceleration is not used. It's not actually on the GPU when it's software decoding. Only one frame is allocated. This is all copied from the official FFmpeg example.

src/video_core/command_classes/vic.cpp
yzct12345: `Clang-Tidy: Either cast from 'int' to 'size_t' (aka 'unsigned long') is ineffective, or there is a loss of precision before the conversion`

src/video_core/command_classes/codecs/codec.h
yzct12345: Reverted. This will become necessary again for the async worker thread.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: It was probably something with the boost queue. Reverted.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: I changed it away from an ASSERT_MSG to a LOG_DEBUG.

> very common for this to fail

I have observed the opposite. I only ever see this failing at the beginning and end of videos. In the beginning I suppose EAGAIN is normal. I ran Link's Awakening, Let's Go and Kirby. The log was quiet in the middle of videos. In fact the only time I saw log spam was when I was figuring out the hwdevice setup code and the messages were very useful.

src/video_core/command_classes/vic.cpp
yzct12345: Fixed as a switch.

src/video_core/command_classes/vic.h
yzct12345: Reverted all of these. I guess I'll just have to ignore my IDE saying this file has 4 warnings.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: Oops. Fixed.

src/common/settings.h
ameerj: That's exactly what I want to avoid completely. The code should be smart enough to choose the best GPU decoder with no user configuration/intervention, and fallback to CPU decoding if no compatible decoders are found.

src/video_core/command_classes/codecs/codec.cpp
ameerj: It's expected to have logging for debug, but this isn't appropriate to merge. An ASSERT should be used to verify a specific predicate that must be (or the code assumes to be) true, and if it is not true then it is a bug in the code. That's not necessarily the case here.

src/common/settings.h
yzct12345: Ok, reverted.

ameerj:
I'm early access tagging this for further testing.

ameerj:
This conflicts with a change some users needed for this PR to work. Please rebase when you get the chance and we will retag this.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -31,8 +29,9 @@ Codec::~Codec() {
     if (!initialized) {
         return;
     }
+
```
ameerj, line 32: unneeded newline

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
...
+    for (const AVPixelFormat* p = pix_fmts; *p != AV_PIX_FMT_NONE; ++p) {
+        if (*p == AV_PIX_FMT_VAAPI) {
+            return AV_PIX_FMT_VAAPI;
+        }
+    }
+    LOG_INFO(Service_NVDRV, "Could not find compatible GPU AV format, falling back to CPU");
+    return *pix_fmts;
+}
+
+constexpr std::array<const char*, 2> VAAPI_DRIVERS = {
```
ameerj, line 59: prefer declaring constants above functions

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
     av_frame_unref(av_frame);
     av_free(av_frame);
     avcodec_close(av_codec_ctx);
+    av_buffer_unref(&av_hw_device);
+}
+
+// Hardware acceleration code from FFmpeg/doc/examples/hw_decode.c under MIT license
+#if defined(LIBVA_FOUND)
+namespace {
```
ameerj, line 48: Move the anonymous namespace to be right after `namespace Tegra {`

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -31,8 +29,9 @@ Codec::~Codec() {
     if (!initialized) {
         return;
     }
+
     // Free libav memory
-    AVFrame* av_frame{nullptr};
+    AVFrame* av_frame;
```
ameerj, line 34: I believe the explicit `nullptr` initialization was intentional here

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
...
+
+constexpr std::array<const char*, 2> VAAPI_DRIVERS = {
+    "i915",
+    "amdgpu",
+};
+
+bool CreateVaapiHwdevice(AVBufferRef** av_hw_device) {
+    AVDictionary* hwdevice_options = nullptr;
+    av_dict_set(&hwdevice_options, "connection_type", "drm", 0);
+    for (auto driver : VAAPI_DRIVERS) {
```
ameerj, line 67:
```suggestion
    for (const auto& driver : VAAPI_DRIVERS) {
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
...
+    if (CreateVaapiHwdevice(&av_hw_device)) {
+        const auto hw_device_ctx = av_buffer_ref(av_hw_device);
+        ASSERT_MSG(hw_device_ctx, "av_buffer_ref failed");
+        av_codec_ctx->hw_device_ctx = hw_device_ctx;
+        av_codec_ctx->get_format = GetHwFormat;
+        return;
+    }
+#endif
+
+    // TODO NVDEC, but integrated GPU should be used first to avoid PCI
```
ameerj, line 96:
```suggestion
    // TODO more GPU accelerated decoders
    // Prioritize integrated GPU to mitigate bandwidth bottlenecks
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
...
+
+    if (const int ret = avcodec_send_packet(av_codec_ctx, &packet); ret) {
+        LOG_DEBUG(Service_NVDRV, "avcodec_send_packet error {}", ret);
+        return nullptr;
+    }
+
+    // Only receive/store visible frames
+    if (raw_frame.vp9_hidden_frame) {
+        return nullptr;
+    }
```
ameerj, line 113: This chunk of code can remain in the calling function. `DecodeImpl()` does not need the `RawFrame`, and can be renamed to `GetDecodedFrame()` instead.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
...
+    AVFrame* hw_frame = av_frame_alloc();
+    AVFrame* sw_frame = hw_frame;
+    ASSERT_MSG(hw_frame, "av_frame_alloc hw_frame failed");
+    if (const int ret = avcodec_receive_frame(av_codec_ctx, hw_frame); ret) {
+        LOG_DEBUG(Service_NVDRV, "avcodec_receive_frame error {}", ret);
+        av_frame_free(&hw_frame);
+        return nullptr;
+    }
+
```
ameerj, line 122: Unneeded newline

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
...
+    if (!hw_frame->width || !hw_frame->height) {
+        LOG_WARNING(Service_NVDRV, "Zero width or height in frame");
+        av_frame_free(&hw_frame);
+        return nullptr;
+    }
+
```
ameerj, line 128: Unneeded newline

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
...
+        ASSERT_MSG(sw_frame, "av_frame_alloc sw_frame failed");
+        // Can't use AV_PIX_FMT_YUV420P and share code with software decoding in vic.cpp
+        // because Intel drivers crash unless using AV_PIX_FMT_NV12
+        sw_frame->format = AV_PIX_FMT_NV12;
+        const int transfer_data_ret = av_hwframe_transfer_data(sw_frame, hw_frame, 0);
+        ASSERT_MSG(!transfer_data_ret, "av_hwframe_transfer_data error {}", transfer_data_ret);
+        av_frame_free(&hw_frame);
+    }
+#endif
+
```
ameerj, line 142: Unneeded newline

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -41,10 +40,121 @@ Codec::~Codec() {
...
+    switch (sw_frame->format) {
+    case AV_PIX_FMT_YUV420P:
+    case AV_PIX_FMT_NV12:
```
ameerj, line 145: Afaik these are the only two formats that get returned by the decoders. This can be reduced to a simple `if` condition.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -53,22 +163,27 @@ void Codec::Initialize() {
...
+    }
 
     const auto av_error = avcodec_open2(av_codec_ctx, av_codec, nullptr);
     if (av_error < 0) {
         LOG_ERROR(Service_NVDRV, "avcodec_open2() Failed.");
         avcodec_close(av_codec_ctx);
+        av_buffer_unref(&av_hw_device);
         return;
     }
+
```
ameerj, line 185: Unneeded newline

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -96,18 +209,15 @@ void Codec::Decode() {
...
-        // Limit queue to 10 frames. Workaround for ZLA decode and queue spam
-        if (av_frames.size() > 10) {
+    RawFrame raw_frame{
+        .frame_data = frame_data,
+        .vp9_hidden_frame = vp9_hidden_frame,
+    };
+    // TODO async
+    AVFrame* sw_frame = DecodeImpl(raw_frame);
+    if (sw_frame) {
+        if (av_frames.push(AVFramePtr{sw_frame, AVFrameDeleter}); av_frames.size() > 10) {
```
ameerj, line 219:
```suggestion
        av_frames.push(AVFramePtr{sw_frame, AVFrameDeleter});
        if (av_frames.size() > 10) {
```
No benefit to having the queue push in the condition statement. Much easier to read breaking this up

src/video_core/command_classes/vic.cpp
ameerj: I'm not a fan of casting from an int -> u32 -> size_t right here. We already have an unsigned casted frame width/height declared below. Move those to above the switch case and use the unsigned variants here.

ameerj (commented):
Few more nits

src/video_core/command_classes/codecs/codec.cpp
yzct12345: That was causing a warning in my IDE. I might as well combine the declaration and assignment.

yzct12345:
> Unneeded newline

> Unneeded newline

> Unneeded newline

> Unneeded newline

> Unneeded newline

At this rate you might as well configure clang-format or something to ban all empty lines if this is the code style. I was using empty lines to separate logical sections and improve readability.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: This change is a step to solving a major 15 millisecond wait in `avcodec_send_packet` on platforms without VA-API like Windows.

This chunk of code in particular needs to be in DecodeImpl to move it from the dynarmic thread to the async gpu thread. That thread will arrive in the next PR. The new `RawFrame` struct is specifically introduced so this chunk can be moved here.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: I removed all the empty lines inside the functions. I hope that makes this PR acceptable to you.

src/video_core/command_classes/vic.cpp
yzct12345: Turns out clang-tidy didn't like two things. It was both the widening cast and the signed conversion. I fixed it by removing the cast.

I tried moving them just below "Writing RGB Frame" but softscaler wants signed.

src/video_core/command_classes/codecs/codec.cpp
ameerj: Just to clarify, this code doesn't run on the dynarmic thread, even with multicore/async gpu disabled. Each HLE service runs on its own thread, with nvdec running on the nvdrv service thread.
For proper async nvdec emulation, `GPU::PushCommandBuffer` would need to push to the async gpu thread. This change will require #6799 to be merged, and will make the entire decoding process asynchronous. So there is little reason to keep the `RawFrame` remnants for the purposes of this PR.

ameerj:
Regarding the unneeded newline comments. I agree that newlines can be very helpful in structuring code. I left this comment particularly when there's a newline introduced that wasn't in the code previously, and the code is otherwise unchanged, or it was immediately after a closing curly brace. 

It's especially off-putting in cases such as this:
```c++
if (cond) {
    return;
}

if (other_cond) {
    LOG();
    return;
}

// Do something
```
Tiny indented scopes are already naturally structured, with the indentation and the block ending with a closing curly brace on a line of its own. 

I'm not saying your code did exactly this, but these are the cases where I would point out unneeded use of newlines, and you're 100% free to ignore a review comment if you think it's not the best course of action.

src/video_core/command_classes/codecs/codec.cpp
yzct12345: > GPU::PushCommandBuffer

It seems like you have a better implementation. I will revert drop my locally stashed changed for my thread.

`RawFrame` and `DecodeImpl` have disappeared.

ameerj (approved):
Thank you for this contribution. Looking forward to seeing more from you!

Issue #6710, opened 2021-07-25
Few unpolished things from Project Hades


v1993:
Opening an issue since Hades PR was merged. Those were reviews/comments there.

1. "General -> Debug -> General" tab is squeezed with addition of all new stuff (gtk2 theme):
![Screenshot_20210710_180257](https://user-images.githubusercontent.com/16255906/125167468-6713b700-e1a9-11eb-955f-1446b483c065.png)
Probably move graphics options into a separate tab? Oddly, Debugging section ("General -> Debug -> General -> Debugging" is a weird one) is not squeezed in the same manner for whatever reason.
2. https://github.com/yuzu-emu/yuzu/blob/7e272d3cd81656b65b21f5a569fc9a2d76cac758/src/shader_recompiler/ir_opt/global_memory_to_storage_buffer_pass.cpp#L108-L109
Comment is clearly copied from previous function and needs an update.
3. https://github.com/yuzu-emu/yuzu/blob/7e272d3cd81656b65b21f5a569fc9a2d76cac758/src/shader_recompiler/frontend/ir/opcodes.inc#L5
Second `arg4` is most surely intended to be an `arg5`.

FernandoS27:
oh thankfully this r just typos. Nothing serious.

v1993:
Yeah (with exception of squeezed GUI?), that's why those are "unpolished" rather than "unfinished" or "broken" or "regressed".

PR #6714, opened 2021-07-26
config: Make debug and input pages scrollable

yzct12345:
The debug and input pages are squeezed on my laptop screen. Others agrees, such as in number 1 of #6710. This PR makes the config easier to use for those who don't have a larger monitor.


Morph1984:
Could you screenshot what the contrroller input dialog now looks like? Making it a scroll area may not make it look as nice.

yzct12345:
Here are the screenshots:

yuzu1
yuzu2
yuzu3
yuzu4

Morph1984:
I see, on linux the element sizes are different

v1993:
Not that it's specifically on Linux, those are just different Qt themes.

bunnei:
@yzct12345 Can you rebase this? Thank you!

PR #6736, opened 2021-07-26
Config-graphics: reword GLASM option

CaptV0rt3x:
* Change wording to explain that GLASM is actually short for Assembly Shaders


Morph1984 (approved):

Morph1984:
Such an important change needs 10 weeks of testing 🤣

PR #6762, opened 2021-07-28
ui: make game list dir icons smaller

maltejur:
I don't know if this is intentional, but the icons for the folders in the game list tree view seem really big and are thus also a bit pixelated. In my opinion they would look a lot better if they were about the size of the text.

Before:
Screenshot_20210728_210256

After:
Screenshot_20210728_210218

### DIFF
--- a/src/yuzu/game_list_p.h
+++ b/src/yuzu/game_list_p.h
@@ -233,8 +233,7 @@ public:
         UISettings::GameDir* game_dir = &directory;
         setData(QVariant(UISettings::values.game_dirs.indexOf(directory)), GameDirRole);
 
-        const int icon_size =
-            std::min(static_cast<int>(UISettings::values.icon_size.GetValue()), 24);
+        const int icon_size = 24;
         switch (dir_type) {
         case GameListItemType::SdmcDir:
             setData(
@@ -295,8 +294,7 @@ public:
     explicit GameListAddDir() {
         setData(type(), TypeRole);
 
-        const int icon_size =
-            std::min(static_cast<int>(UISettings::values.icon_size.GetValue()), 24);
+        const int icon_size =  24;
         setData(QIcon::fromTheme(QStringLiteral("plus"))
                     .pixmap(icon_size)
                     .scaled(icon_size, icon_size, Qt::IgnoreAspectRatio, Qt::SmoothTransformation),
@@ -318,8 +316,7 @@ public:
     explicit GameListFavorites() {
         setData(type(), TypeRole);
 
-        const int icon_size =
-            std::min(static_cast<int>(UISettings::values.icon_size.GetValue()), 24);
+        const int icon_size = 24;
         setData(QIcon::fromTheme(QStringLiteral("star"))
                     .pixmap(icon_size)
                     .scaled(icon_size, icon_size, Qt::IgnoreAspectRatio, Qt::SmoothTransformation),
--- a/src/yuzu/game_list_p.h
+++ b/src/yuzu/game_list_p.h
@@ -80,7 +80,7 @@ public:
         setData(qulonglong(program_id), ProgramIdRole);
         setData(game_type, FileTypeRole);
 
-        const u32 size = UISettings::values.icon_size.GetValue();
+        const u32 size = UISettings::values.game_icon_size.GetValue();
 
         QPixmap picture;
         if (!picture.loadFromData(picture_data.data(), static_cast<u32>(picture_data.size()))) {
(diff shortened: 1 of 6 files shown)


Morph1984:
After some internal discussion, we would suggest that the folder icon size be 3/4ths of the selected game icon size, with a maximum of 64px.

Having it at 16px is too small, and we have determined that 48px at the standard size of 64x64 is more optimal (see below)

image

maltejur:
I don't think scaling the folder icons with the user selected game icon size is the right idea. The user probably would want to make the game icons **a)** smaller, when he wants more space in the game list, or **b)** bigger, when he wants to see the icon better because it you can't really see any details on the game icons on a smaller size.

The thing is, the folder icons don't need to be bigger because you should recognize them at almost any size. That also means if the user makes the game icons bigger, he probably doesn't want to make the folder icons bigger as well.

If 18px is too small, what about 24px? That would look like this:

Screenshot_20210728_222245

CaptV0rt3x:
Actually, the reason the team came up with this approach was because this PR in its current form (with max of 18) would make the icons look really small on higher resolution displays, where users might choose 128x128 icon size or 256x256. 

With our suggested approach, the icons will remain at 3/4th size of icon size for until 64x64 and beyond that it'd remain at a maximum of 64px. 

If a user really chooses 128 or 256 icon size, then I don't think 64px folder icon size would look out of place to them.

maltejur:
> would make the icons look really small on higher resolution displays

To be fair, on higher resolution displays without some sort of scaling, everything except the game icons looks small. 
(Also if you have time could you make a screenshot of the 24px version on a high resolution display, just curious)

> Also, forgot to mention that the folder icons have a right-click context menu that allows you to remove the folder or move it up/down. So, having it too small will impact that.

I don't really understand how that would impact the context menu.

maltejur:
If you want I will implement your approach, but I am still not convinced that that would look better.

G-Spawn:
I hope its fair to say that you must also consider that not everyone has the same level of clarity in their eyesight (glasses or not or even with nice screens). At least from a user standpoint personally anything below the standard 32px can just look like a glorified blob with not enough quality to be properly seen across the room and only when I'm sitting in front of my pc. This will just force me to rely on the existing larger icon settings / the filter function and a keyboard to search or to just turn off the icons entirely. 

I understand the desire for more flexibility but I'm likely not going to be the only one affected if said game - folder icons are made even smaller. Even as an added option I likely wouldn't be able to comfortably approach it. As this is how it can look even on my 1080p 40+ inch tv screen to me with the smallest 32x32px icons atm with default windows scaling afaik below.

image

now imagine scrolling through 100+ of games with even smaller  sized icons :p. None of this is meant in offense to you in any way of course. Just chiming in my thoughts on the matter.

gidoly:
hello my name is gidoly.  yuzu discord server regular and doing tests for yuzu 
i appreciate to see you trying to improve quality of life but i am here to provide opinion about this change as user. i used yuzu for long time. i liked original looking because i have bad eye sight and using glasses. it will be tire to look at small text on screen, i think other people who uses computer for long time also agree with that since it can make eye tire more quickly
in my opinion it could be added for seperate setting instead replace original looking entirely
hope you consider this user opinion and again thanks for your effort

maltejur:
Okay what about this: By default the game icons are 64px and the folder icons are 48px (like suggested), but you configure both independently from each other in the settings.

Screenshot_20210729_143040

This would address the accessibility concerns with the default 48px while also allowing users to make the icons smaller or larger to their liking.

Also a little bit off topic, you can use the `QT_SCALE_FACTOR` environment variable to configure the scaling of yuzu independent from the scaling your os if other things in yuzu are too small for you.

src/yuzu/game_list_p.h
```diff
@@ -294,7 +294,8 @@ class GameListAddDir : public GameListItem {
     explicit GameListAddDir() {
         setData(type(), TypeRole);
 
-        const int icon_size = 24;
+        const int icon_size = UISettings::values.folder_icon_size.GetValue();
+        ;
```
lioncash, line 298: seems like an unnecessary semicolon slipped through here and below

maltejur: Yes somehow missed that, thanks

src/yuzu/uisettings.h
```diff
@@ -91,7 +91,8 @@ struct Values {
 
     // Game List
     Settings::BasicSetting<bool> show_add_ons{true, "show_add_ons"};
-    Settings::BasicSetting<uint32_t> icon_size{64, "icon_size"};
+    Settings::BasicSetting<uint32_t> game_icon_size{48, "game_icon_size"};
+    Settings::BasicSetting<uint32_t> folder_icon_size{64, "folder_icon_size"};
```
Morph1984, line 95: The values here are flipped

Morph1984 (approved):

Issue #6603, opened 2021-07-11
Street Fighter 30th Anniversary - Upside Down


codbugpatrol:
Game: Street Fighter 30th Anniversary

The game is upside down.
Using Early Access 1868. Mainline has the same problem.

Host CPU: AMD Ryzen 5 3600 6-Core Processor | AVX2 | FMA
Host OS: Windows 10 (10.0)
Host RAM: 15.93 GB
Host Swap: 19.53 GB
Driver: NVIDIA 466.77.0.0
Device: NVIDIA GeForce GTX 670
Vulkan: 1.2.168

![upsidedown_streetfighter_30_anniversary](https://user-images.githubusercontent.com/85196112/125183308-f5a92800-e1eb-11eb-92cb-bc76e9f4fd43.png)

Immersion95:
I have an AMD card and it's working great, SOR4 unfortunately is upside down though.

codbugpatrol:
@Immersion95, Axion Verge, Bastion have the same problem in my testing.

codbugpatrol:
No longer happens on Yuzu EA 2074 and Nvidia driver 471.41.

PR #6765, opened 2021-07-29
vk_rasterizer: Flip viewport on Y_NEGATE

### LINKED ISSUE #2703: Proper viewport flip emulation

ReinUsesLisp:
Matches OpenGL's behavior. I don't believe this register flips geometry,
but we have to try to match behavior on both backends.

Closes #2703.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -61,11 +61,16 @@ struct DrawParams {
 VkViewport GetViewportState(const Device& device, const Maxwell& regs, size_t index) {
     const auto& src = regs.viewport_transform[index];
     const float width = src.scale_x * 2.0f;
-    const float height = src.scale_y * 2.0f;
+    float y = src.translate_y - src.scale_y;
+    float height = src.scale_y * 2.0f;
+    if (regs.screen_y_control.y_negate) {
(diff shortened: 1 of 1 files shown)


codbugpatrol:
@ReinUsesLisp Celeste still partially upside down. Street fighther 30th is fixed, and Bastion is still upside down with EA 1944.

codbugpatrol:
Fixes https://github.com/yuzu-emu/yuzu/issues/6603

ReinUsesLisp:
@codbugpatrol that's unrelated to this PR. That happens when drivers lack `GL_NV_viewport_swizzle` or `VK_NV_viewport_swizzle`.

codbugpatrol:
@ReinUsesLisp, ok. But in the just released Vulkan version in other emulator, it works perfect. No upside down in Vulkan. So it is not something lacking in my driver.

v1993:
@codbugpatrol I guesss riujinx uses some other method of flipping that does not use those extensions. While fallback to the same codepath would probably be nice, it's still an unrelated issue.

ameerj (approved):

PR #6768, opened 2021-07-30
cmake: Disable PIE for performance

yzct12345:
This makes yuzu faster by disabling PIE. PIE is used for shared libraries and the ASLR security feature. We are an executable not a shared library and we don't need extra security because we are dumping games trusted by Nintendo. This PR also makes the addr2line debugging tool work after-the-fact without gdb.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,7 @@ else()
 endif()
 
 if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
-    add_compile_options(-fno-pie -no-pie)
+    add_compile_options(-fno-pie)
     add_link_options(-no-pie)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,6 +144,11 @@ else()
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
 endif()
 
+if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    add_compile_options(-fno-pie -no-pie)
+    add_link_options(-no-pie)
+endif()
+
 # Output binaries to bin/
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 


v1993:
Making `addr2line` work sounds nice, but:

1. For performance, would you mind providing some basic benchmarking results? Just to see that it does indeed make difference and how significant it is.
2. As for `addr2line`, don't core dumps allows to do the same?

v1993:
Additionally, if you will actually measure some increase in performance, consider using CMake's mechanism for turning PIE off instead of adding flags directly. You'll also want policy CMP0083 to be set to new.

yzct12345:
> performance, would you mind providing some basic benchmarking results

The performance improvement was less than 1 FPS. I might have seen a 33 FPS to 34 FPS average increase in the Kirby intro movie but I think it's so insignificant it's a placebo effect.

Theoretically this should improve performance and I can't see why it would decrease it.

> don't core dumps allows to do the same

This is the whole reason why I want this change. Core dumps are just like gdb as I have to remember to do `ulimit -c unlimited` before I run yuzu. The problem is yuzu only crashes when I'm enjoying something and never when I looking for bugs. The GPU cache garbage collection often crashes in the Link's Awakening intro but never when gdb is attached. Also the last time I tried to open a core dump it none of the symbols loaded probably because of PIE. I only have the line in dmesg to use and this PR removes the PIE randomization so addr2line actually works. It was showing "?:?" before.

> CMake's mechanism for turning PIE off instead of adding flags directly. You'll also want policy CMP0083 to be set to new.

I tried that already and it doesn't work.

v1993:
I surely wouldn't expect this to lower performance, it's just that both PR title and description talk about performance increase as a main factor, so I wondered if it's actually that notable (it is not, it seems), given that no other emulators I've checked force disable/enable PIE. Also, symbol lookup in core dumps works just fine for me in "RelWithDebInfo" build type which is still built with PIE. As for core dump creation, you can avoid need for `ulimit` by editing `/etc/security/limits.conf`.

lioncash:
> given that no other emulators I've checked force disable/enable PIE

Indeed. Given this, I'm mainly leaning toward not particularly wanting to merge this, given the perf gains seem to be negligible and the core dump creation can be mitigated.

yzct12345:
Looks like another one to put into my set of local-only patches.

v1993:
Actually, this does not need to be a patch if you're using it locally. You can pass flags to cmake from command line to do this.

PR #6776, opened 2021-07-30
settings: Implement settings ranges

lat9nq:
Guards against scenarios where a user may manually input invalid settings to the config files then try to run yuzu.

This is achieved by adding two new settings classes, BasicRangedSetting and RangedSetting. They require a minimum and maximum value during initialization. During SetValue or assignment, it clamps the value within the range.

Need to note that RangedSetting is exhibiting diamond inheritance. For now it is handled well with C++ virtual base classes.

Also I don't like how system settings like language_index are handled. Maybe it's time to give all of these an enum class, or find where one is already defined in the source? Would like input on this.

### DIFF
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <array>
 #include <atomic>
 #include <chrono>
@@ -168,15 +169,7 @@ public:
      * @param value The desired value
      */
     void SetValue(const Type& value) override {
-        Type temp;
-        if (value < minimum) {
-            temp = minimum;
-        } else if (value > maximum) {
-            temp = maximum;
-        } else {
-            temp = value;
-        }
-        std::swap(this->global, temp);
+        this->global = std::clamp(value, minimum, maximum);
     }
 
     /**
@@ -186,15 +179,7 @@ public:
      * @returns A reference to the setting's value
      */
     const Type& operator=(const Type& value) override {
-        Type temp;
-        if (value < minimum) {
-            temp = minimum;
-        } else if (value > maximum) {
-            temp = maximum;
-        } else {
-            temp = value;
-        }
-        std::swap(this->global, temp);
+        this->global = std::clamp(value, minimum, maximum);
         return this->global;
     }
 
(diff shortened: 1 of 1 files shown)


src/common/settings.h
```diff
@@ -141,6 +141,67 @@ class BasicSetting {
...
+
+    /**
+     * Like BasicSetting's SetValue, except value is clamped to the range of the setting.
+     *
+     * @param value The desired value
+     */
+    void SetValue(const Type& value) override {
+        Type temp;
+        if (value < minimum) {
+            temp = std::move(minimum);
```
lioncash, line 173: std::move shouldn't be used here. Were the member variables not const to prevent std::move from doing anything, calling the setter twice would be undefined behavior.

Ditto for the moves in `operator=`

src/common/settings.h
```diff
@@ -234,18 +301,87 @@ class Setting final : public BasicSetting<Type> {
...
+    /**
+     * Like BasicSetting's SetValue, except value is clamped to the range of the setting. Sets the
+     * appropriate value depending on the global state.
+     *
+     * @param value The desired value
+     */
+    void SetValue(const Type& value) override {
+        Type temp;
+        if (value < this->minimum) {
+            temp = std::move(this->minimum);
```
lioncash, line 347: Ditto here  and in operator=

lioncash: This isn't resolved

src/common/settings.h
```diff
@@ -141,6 +141,67 @@ class BasicSetting {
...
+        Type temp;
+        if (value < minimum) {
+            temp = minimum;
+        } else if (value > maximum) {
+            temp = maximum;
+        } else {
+            temp = value;
+        }
+        std::swap(this->global, temp);
+        return this->global;
```
PixelyIon, line 198: This can just be a single line (Albeit, I recommend breaking it up for readability):
```cpp
return global = std::clamp(value, minimum, maximum);
````
I don't quite understand why you utilize `std::swap` when you don't utilize both the swapped values after that point, just use assignment and I'm not sure if this is convention in *yuzu* but why utilize explicit `this` here?

The points I bring up here do apply to `BasicRangedSetting::SetValue` and `RangedSetting` as well so you might want to change that up too, I should add that I'm not a maintainer/contributor from *yuzu* so I could be incorrect about code-style here.

lioncash: Nah, this is good to bring up. But yeah, I'd opt for separating the assignment from the return statement.

lat9nq: `this->` is needed because the parent class owns the fields. Cannot access those without specifying it.

bunnei (approved):

PR #6779, opened 2021-07-31
main: Make use of unique_ptr for the QtSoftwareKeyboardDialog

Morph1984:


lioncash:
Wouldn't this result in deletion issues, since we pass the parent owning dialog into the software keyboard dialog, resulting in the parent attempting to delete the pointer when it gets destroyed?

ameerj:
@Morph1984 what's the status of this?

PR #6787, opened 2021-07-31
GPU: Prequeue frame swaps.

FernandoS27:
The idea of this change is to fix an issue that was phantom added since I introduced the fence manager. Originally frame swaps were sent exactly when the fence was met. However, now fence release is delayed until the host gpu catches up, this causes swaps to be delayed further than they should.

This commit aims to solve that issue by prequeuing frames. 

This solves the bad frame pacing on MHR.

### DIFF
--- a/src/core/hle/service/nvflinger/nvflinger.cpp
+++ b/src/core/hle/service/nvflinger/nvflinger.cpp
@@ -274,8 +274,6 @@ void NVFlinger::Compose() {
             continue;
         }
 
-        const auto& igbp_buffer = buffer->get().igbp_buffer;
-
         if (!system.IsPoweredOn()) {
             return; // We are likely shutting down
         }
@@ -289,21 +287,29 @@ void NVFlinger::Compose() {
         }
         guard->lock();
 
+        system.GetPerfStats().EndSystemFrame();
         MicroProfileFlip();
-
-        // Now send the buffer to the GPU for drawing.
-        // TODO(Subv): Support more than just disp0. The display device selection is probably based
-        // on which display we're drawing (Default, Internal, External, etc)
-        auto nvdisp = nvdrv->GetDevice<Nvidia::Devices::nvdisp_disp0>("/dev/nvdisp_disp0");
-        ASSERT(nvdisp);
-
-        nvdisp->flip(igbp_buffer.gpu_buffer_id, igbp_buffer.offset, igbp_buffer.format,
-                     igbp_buffer.width, igbp_buffer.height, igbp_buffer.stride,
-                     buffer->get().transform, buffer->get().crop_rect);
+        system.SpeedLimiter().DoSpeedLimiting(system.CoreTiming().GetGlobalTimeUs());
+        system.GetPerfStats().BeginSystemFrame();
 
         swap_interval = buffer->get().swap_interval;
         buffer_queue.ReleaseBuffer(buffer->get().slot);
     }
+}
+
+void NVFlinger::PrequeueFrame(u32 buffer_queue_id, u32 slot) {
+    auto& buffer_queue = *FindBufferQueue(buffer_queue_id);
+    const auto& buffer = buffer_queue.AccessBuffer(slot);
+    const auto& igbp_buffer = buffer.igbp_buffer;
+
+    // Now send the buffer to the GPU for drawing.
+    // TODO(Subv): Support more than just disp0. The display device selection is probably based
+    // on which display we're drawing (Default, Internal, External, etc)
+    auto nvdisp = nvdrv->GetDevice<Nvidia::Devices::nvdisp_disp0>("/dev/nvdisp_disp0");
+    ASSERT(nvdisp);
+    nvdisp->flip(igbp_buffer.gpu_buffer_id, igbp_buffer.offset, igbp_buffer.format,
+                 igbp_buffer.width, igbp_buffer.height, igbp_buffer.stride, buffer.transform,
+                 buffer.crop_rect, buffer.multi_fence);
 }
 
 s64 NVFlinger::GetNextTicks() const {
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -114,10 +114,17 @@ void GPU::WaitFence(u32 syncpoint_id, u32 value) {
     });
 }
 
+void GPU::IncrementSyncPointGuest(const u32 syncpoint_id) {
+    std::lock_guard lock{pre_sync_mutex};
+    auto& syncpoint = pre_syncpoints.at(syncpoint_id);
+    syncpoint++;
+    ProcessFrameRequests(syncpoint_id, syncpoint);
+}
+
 void GPU::IncrementSyncPoint(const u32 syncpoint_id) {
+    std::lock_guard lock{sync_mutex};
     auto& syncpoint = syncpoints.at(syncpoint_id);
     syncpoint++;
-    std::lock_guard lock{sync_mutex};
     sync_cv.notify_all();
     auto& interrupt = syncpt_interrupts.at(syncpoint_id);
     if (!interrupt.empty()) {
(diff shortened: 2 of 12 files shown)


src/core/hle/service/nvflinger/nvflinger.cpp
```diff
@@ -289,23 +287,31 @@ void NVFlinger::Compose() {
...
-                     buffer->get().transform, buffer->get().crop_rect);
+        system.SpeedLimiter().DoSpeedLimiting(system.CoreTiming().GetGlobalTimeUs());
+        system.GetPerfStats().BeginSystemFrame();
 
         swap_interval = buffer->get().swap_interval;
         buffer_queue.ReleaseBuffer(buffer->get().slot);
     }
 }
 
+void NVFlinger::PrequeeFrame(u32 buffer_queue_id, u32 slot) {
```
Morph1984, line 300:
```suggestion
void NVFlinger::PrequeueFrame(u32 buffer_queue_id, u32 slot) {
```

src/video_core/gpu.cpp
```diff
@@ -162,25 +169,127 @@ bool GPU::CancelSyncptInterrupt(const u32 syncpoint_id, const u32 value) {
...
+                    [searching_id](const FrameQueue& item) { return item.id == searching_id; });
+                ASSERT(it != frame_queue_items.end());
+                frame_info = it->frame_info;
+                frame_queue_items.erase(it);
+            }
+            renderer->SwapBuffers(&frame_info);
+            break;
+        }
+        default: {
+            LOG_ERROR(HW_GPU, "Unknown, work request");
```
Morph1984, line 220: May be worth logging the unknown request type if it occurs, also break for the default case
```suggestion
            LOG_ERROR(HW_GPU, "Unknown work request type={}", request.type);
            break;
```

v1993:
This PR causes flickering with black in HW: DE, audio gets glitchy as well. Vulkan, GTX 1650 Super.

FernandoS27:
> This PR causes flickering with black in HW: DE, audio gets glitchy as well. Vulkan, GTX 1650 Super.

I'll investigate.

Morph1984:
We have another report, this time with Fire Emblem Three Houses

FernandoS27:
All regressions should be fixed now.

src/video_core/gpu.cpp
```diff
@@ -162,25 +169,121 @@ bool GPU::CancelSyncptInterrupt(const u32 syncpoint_id, const u32 value) {
...
+void GPU::QueueFrame(const Tegra::FramebufferConfig* framebuffer,
+                     const Service::Nvidia::MultiFence& fences) {
+    std::unique_lock<std::mutex> lock(frame_requests_mutex);
+    if (fences.num_fences == 0) {
+        u64 new_queue_id = frame_queue_ids++;
+        FrameQueue item{
+            .frame_info = *framebuffer,
+            .id = new_queue_id,
+        };
+        frame_queue_items.push_back(item);
```
ameerj, line 238: For consistency with other queue pushes in this file.
```suggestion
        frame_queue_items.emplace_back(*framebuffer, new_queue_id);
```

src/video_core/gpu.cpp
```diff
@@ -162,25 +169,121 @@ bool GPU::CancelSyncptInterrupt(const u32 syncpoint_id, const u32 value) {
...
+    };
+    std::unique_lock lck{pre_sync_mutex};
+    for (size_t i = 0; i < fences.num_fences; i++) {
+        auto& fence = fences.fences[i];
+        if (pre_syncpoints[fence.id].load(std::memory_order_relaxed) < fence.value) {
+            const FrameTrigger trigger{
+                .id = new_id,
+                .sync_point_value = fence.value,
+            };
+            frame_triggers[fence.id].push_back(trigger);
```
ameerj, line 256:
```suggestion
            frame_triggers[fence.id].emplace_back(new_id, fence.value);
```

src/video_core/gpu.h
```diff
@@ -392,27 +426,50 @@ class GPU final {
...
-    std::list<FlushRequest> flush_requests;
-    std::atomic<u64> current_flush_fence{};
-    u64 last_flush_fence{};
-    std::mutex flush_request_mutex;
+    struct WorkRequest {
+        explicit WorkRequest(u64 fence_, VAddr addr_, std::size_t size_)
+            : fence{fence_}, type{RequestType::Flush} {
+            flush.addr = addr_;
+            flush.size = size_;
+        }
```
ameerj, line 450:
```suggestion
            : fence{fence_}, type{RequestType::Flush}, flush{addr_, size_} {}
```

src/video_core/gpu.h
```diff
@@ -392,27 +426,50 @@ class GPU final {
...
+
+        explicit WorkRequest(u64 fence_, u64 id) : fence{fence_}, type{RequestType::QueueFrame} {
+            queue_frame.id = id;
+        }
+        u64 fence;
```
ameerj, line 455: missing newline

src/video_core/gpu.cpp
```diff
@@ -162,25 +169,121 @@ bool GPU::CancelSyncptInterrupt(const u32 syncpoint_id, const u32 value) {
...
+    if (request.count == 0) {
+        lck.unlock();
+        gpu_thread.SwapBuffers(framebuffer);
+        return;
+    }
+    frame_requests.emplace(new_id, request);
+}
+
+void GPU::ProcessFrameRequests(u32 syncpoint_id, u32 new_value) {
+    auto& list = frame_triggers[syncpoint_id];
```
ameerj, line 269: Is this threadsafe?

v1993:
This is still causing flickering in HW: DE, but of very different manner:

1. You must have "unlimited framerate" option on, which is experimental and does not work nice for this game (still, it's a regression introduced by this PR)
2. Only title screen video is flickering (with white this time), menus are fine
3. This doesn't happen on Vulkan, only on OpenGL

FernandoS27:
> 1. unlimited framerate

That's a hack no matter what. It's not considered a regression if it does not affect normal behavior. Games can act very differently under a "disabled vsync" which is what that option does. There's no guarantees about it.

v1993:
Ah, I see then. Is there any explanation why this happens on OpenGL but not on Vulkan? Lucky timing differences?

codbugpatrol:
For me, after this commit Dragon Fighter Z, which was crashing badly, now besides crash it also seems to jumping frames. 
This is really noticible using a special attack, like gohan's kamehameha. Really noticible that is losing frames.

ameerj (changes requested):
This PR causes a deadlock when booting homebrew

codbugpatrol:
@FernandoS27 , any news on this PR ? Did you also noticed the crashes after a while in DBFZ ? To reproduce, enter in practice mode, enter the match and start using special attacks, it will launch device loss.
[ 133.107316] Render.Vulkan <Critical> video_core\vulkan_common\vulkan_device.cpp:ReportLoss:669: Device loss occured!

v1993:
This PR is intended to be superseded by another one, at which point this one will be closed. Please don't report issues that happen without PR in its discussion.

Timmy5942:
> > 1. unlimited framerate
> 
> That's a hack no matter what. It's not considered a regression if it does not affect normal behavior. Games can act very differently under a "disabled vsync" which is what that option does. There's no guarantees about it.
> 
> What this PR does, is make the host send swap commands exactly where the game asks to use a buffer on. In the case of HW:DE it does not respect fencing and sends swaps before the command list is done, expecting to have enough time before vSync kicks in. The old behavior worked because we over delayed the swaps, this new behavior fixes games on normal behavior as it does the swas where it should do them.

Timmy5942:
> [FernandoS27](https://github.com/FernandoS27)@ I wonder why when the GPU is set on High it fixes this problem?
> 
> #7945 "Zelda link's awakening: bowWow not rendering"
> 
> but slows down depending on video card used....
> 
> Maybe some think like a medium high, What happens in the Source code when GPU High is Enabled?
> 
> Might be an interesting look, What code you guys come up with?, I don't mind trying a test in a Manual Mingw build!
> 
> EDIT: Issue seams to be around "use asynchronous GPU emulation" when disabled slow bowWow rendering fine, Enabled Not fine, Disabled + Normal GPU still fine

Only way to do it while on GPU High

If to change nvflinger.cpp

s64 NVFlinger::GetNextTicks() const {
    static constexpr s64 max_hertz = 120LL;

to

s64 NVFlinger::GetNextTicks() const {
    static constexpr s64 max_hertz = 240LL;

This will keep game game speed in GPU High as is it was set on Normal, Normally it slower on High

Just not FPS.....

FernandoS27:
superseded by YFC

PR #4836, opened 2020-10-27
nvdec: Defer VIC frame texture cache invalidation until end of video stream

ameerj:
This PR aims to address nvdec custscene flickering found in `The Legend of Zelda: Link's Awakening` and `Onimusha Warlords` by deferring VIC frame texture cache invalidation until the video stream is complete. 

Please do not merge as this PR currently does not formally address the issue, rather, it is a temporary workaround.

PR #6790, opened 2021-08-01
decoders: Double the performance

yzct12345:
This makes UnswizzleTexture two times faster. It is the main bottleneck in video decoding.

| What | Before | After |
|-|-|-|
| Pokemon intro movie | 27 FPS | 30 FPS |
| Kirby intro movies | 29 FPS | 44 FPS |
| Kirby first world | 25 FPS | 30 FPS |
| Kirby first level | 22 FPS | 27 FPS |
| Link's Awakening intro movie | 12 FPS | 21 FPS |
| Animal crossing set-up | 25 FPS | 30 FPS |
| Animal crossing intro movie | 29 FPS | 30 FPS |
| Animal crossing in-game | 18 FPS | 23 FPS |

### DIFF
--- a/src/video_core/textures/decoders.cpp
+++ b/src/video_core/textures/decoders.cpp
@@ -74,39 +74,7 @@ void SwizzleImpl(std::span<u8> output, std::span<const u8> input, u32 width, u32
                 u8* const dst = &output[TO_LINEAR ? swizzled_offset : unswizzled_offset];
                 const u8* const src = &input[TO_LINEAR ? unswizzled_offset : swizzled_offset];
 
-                switch (BYTES_PER_PIXEL) {
-                case 1: {
-                    *dst = *src;
-                    break;
-                }
-                case 2: {
-                    const auto dst2 = reinterpret_cast<u16* const>(dst);
-                    const auto src2 = reinterpret_cast<const u16* const>(src);
-                    *dst2 = *src2;
-                    break;
-                }
-                case 4: {
-                    const auto dst4 = reinterpret_cast<u32* const>(dst);
-                    const auto src4 = reinterpret_cast<const u32* const>(src);
-                    *dst4 = *src4;
-                    break;
-                }
-                case 8: {
-                    const auto dst8 = reinterpret_cast<u64* const>(dst);
-                    const auto src8 = reinterpret_cast<const u64* const>(src);
-                    *dst8 = *src8;
-                    break;
-                }
-                case 16: {
-                    const auto dst16 = reinterpret_cast<u128* const>(dst);
-                    const auto src16 = reinterpret_cast<const u128* const>(src);
-                    *dst16 = *src16;
-                    break;
-                }
-                default: {
-                    UNREACHABLE();
-                }
-                }
+                std::memcpy(dst, src, BYTES_PER_PIXEL);
             }
         }
     }
@@ -116,34 +84,23 @@ template <bool TO_LINEAR>
 void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixel, u32 width,
              u32 height, u32 depth, u32 block_height, u32 block_depth, u32 stride_alignment) {
     switch (bytes_per_pixel) {
-    case 1: {
-        SwizzleImpl<TO_LINEAR, 1>(output, input, width, height, depth, block_height, block_depth,
-                                  stride_alignment);
-        break;
-    }
-    case 2: {
-        SwizzleImpl<TO_LINEAR, 2>(output, input, width, height, depth, block_height, block_depth,
-                                  stride_alignment);
-        break;
-    }
-    case 4: {
-        SwizzleImpl<TO_LINEAR, 4>(output, input, width, height, depth, block_height, block_depth,
-                                  stride_alignment);
-        break;
-    }
-    case 8: {
-        SwizzleImpl<TO_LINEAR, 8>(output, input, width, height, depth, block_height, block_depth,
-                                  stride_alignment);
-        break;
-    }
-    case 16: {
-        SwizzleImpl<TO_LINEAR, 16>(output, input, width, height, depth, block_height, block_depth,
-                                   stride_alignment);
-        break;
-    }
-    default: {
+    case 1:
+        return SwizzleImpl<TO_LINEAR, 1>(output, input, width, height, depth, block_height,
+                                         block_depth, stride_alignment);
+    case 2:
+        return SwizzleImpl<TO_LINEAR, 2>(output, input, width, height, depth, block_height,
+                                         block_depth, stride_alignment);
+    case 4:
+        return SwizzleImpl<TO_LINEAR, 4>(output, input, width, height, depth, block_height,
+                                         block_depth, stride_alignment);
+    case 8:
+        return SwizzleImpl<TO_LINEAR, 8>(output, input, width, height, depth, block_height,
+                                         block_depth, stride_alignment);
+    case 16:
+        return SwizzleImpl<TO_LINEAR, 16>(output, input, width, height, depth, block_height,
+                                          block_depth, stride_alignment);
+    default:
         UNIMPLEMENTED_MSG("Unknown bytes_per_pixel {}", bytes_per_pixel);
-    }
     }
 }
 } // Anonymous namespace
--- a/src/video_core/textures/decoders.cpp
+++ b/src/video_core/textures/decoders.cpp
@@ -18,9 +18,9 @@
 
 namespace Tegra::Texture {
 namespace {
-template <bool TO_LINEAR>
-void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixel, u32 width,
-             u32 height, u32 depth, u32 block_height, u32 block_depth, u32 stride_alignment) {
+template <bool TO_LINEAR, u32 BYTES_PER_PIXEL>
+void SwizzleImpl(std::span<u8> output, std::span<const u8> input, u32 width, u32 height, u32 depth,
+                 u32 block_height, u32 block_depth, u32 stride_alignment) {
     // The origin of the transformation can be configured here, leave it as zero as the current API
     // doesn't expose it.
     static constexpr u32 origin_x = 0;
@@ -28,9 +28,9 @@ void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixe
     static constexpr u32 origin_z = 0;
 
     // We can configure here a custom pitch
-    // As it's not exposed 'width * bpp' will be the expected pitch.
-    const u32 pitch = width * bytes_per_pixel;
-    const u32 stride = Common::AlignUpLog2(width, stride_alignment) * bytes_per_pixel;
+    // As it's not exposed 'width * BYTES_PER_PIXEL' will be the expected pitch.
+    const u32 pitch = width * BYTES_PER_PIXEL;
+    const u32 stride = Common::AlignUpLog2(width, stride_alignment) * BYTES_PER_PIXEL;
 
     const u32 gobs_in_x = Common::DivCeilLog2(stride, GOB_SIZE_X_SHIFT);
     const u32 block_size = gobs_in_x << (GOB_SIZE_SHIFT + block_height + block_depth);
@@ -54,14 +54,14 @@ void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixe
                                  ((block_y & block_height_mask) << GOB_SIZE_SHIFT);
 
             for (u32 column = 0; column < width; ++column) {
-                const u32 x = (column + origin_x) * bytes_per_pixel;
+                const u32 x = (column + origin_x) * BYTES_PER_PIXEL;
                 const u32 offset_x = (x >> GOB_SIZE_X_SHIFT) << x_shift;
 
                 const u32 base_swizzled_offset = offset_z + offset_y + offset_x;
                 const u32 swizzled_offset = base_swizzled_offset + table[x % GOB_SIZE_X];
 
                 const u32 unswizzled_offset =
-                    slice * pitch * height + line * pitch + column * bytes_per_pixel;
+                    slice * pitch * height + line * pitch + column * BYTES_PER_PIXEL;
 
                 if (const auto offset = (TO_LINEAR ? unswizzled_offset : swizzled_offset);
                     offset >= input.size()) {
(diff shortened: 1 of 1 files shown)


src/video_core/textures/decoders.cpp
```diff
@@ -73,11 +73,79 @@ void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixe
...
                 const u8* const src = &input[TO_LINEAR ? unswizzled_offset : swizzled_offset];
-                std::memcpy(dst, src, bytes_per_pixel);
+
+                switch (BYTES_PER_PIXEL) {
+                case 1: {
+                    *dst = *src;
+                    break;
+                }
+                case 2: {
+                    const auto dst2 = reinterpret_cast<u16* const>(dst);
```
lioncash, line 83: These casts invoke undefined behavior, since they violate strict aliasing, as well as alignment requirements.

lioncash: Consider using `std::memcpy` with the `BYTES_PER_PIXEL` value, which will make it well-defined and also allow the compiler to elide the memcpy.

ReinUsesLisp:
Does the performance difference stay if you just call std::memcpy but leave BYTES_PER_PIXEL as a constant? That way you can avoid the UB. Consider using assume_aligned if that doesn't work.

src/video_core/textures/decoders.cpp
yzct12345: The point of this PR is to remove the `std::memcpy` and replace it with a switch.

There are two switches. The inner switch is in the template and should be optimized out.

src/video_core/textures/decoders.cpp
```diff
@@ -73,11 +73,79 @@ void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixe
...
         }
     }
 }
+
+template <bool TO_LINEAR>
+void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixel, u32 width,
+             u32 height, u32 depth, u32 block_height, u32 block_depth, u32 stride_alignment) {
+    switch (bytes_per_pixel) {
+    case 1: {
+        SwizzleImpl<TO_LINEAR, 1>(output, input, width, height, depth, block_height, block_depth,
```
ReinUsesLisp, line 120:
```cpp
switch (bytes_per_pixel) {
case 1:
  return SwizzleImpl<TO_LINEAR, 1>(...);
case 2:
  return SwizzleImpl<TO_LINEAR, 2>(...);
...
}
```

src/video_core/textures/decoders.cpp
lioncash: `memcpy` can be optimized out if the copy size is known (which is the case here), while *still being well-defined*, which the current code is *not* <- extremely key detail.

lioncash: condensed [example](https://godbolt.org/z/Px7orn9en) it gets optimized out.

yzct12345: Fixed.

yzct12345:
> std::memcpy but leave BYTES_PER_PIXEL

Oh, so that's what @lioncash meant.

> memcpy can be optimized out

Wow!

> Does the performance difference

The 2x improvement is still there. Specifically I measured an insignificant improvement in my unit test benchmark but an insignificant worsening of Link's Awakening.

> alignment requirements

I tested with `if (((size_t) dst | (size_t) src) & 3) abort()` and it didn't crash. This means the actual buffers we use are aligned. Additionally on x86 but not ARM unaligned access doesn't crash but is much slower and I would have noticed that.

> assume_aligned

I tested the first suggestion and it works fine. The disassembly looks about the same as my version.

gidoly:
world ends with you video test
current early access 25 fps ish with laggyness
yuzu_dX9T7pe9F5
with this pr
video_+_astc_EUOFR26mnX

ameerj:
By the way, I'm not sure if you're aware, but you need #4836 locally merged to make ZLA videos bearable

src/video_core/textures/decoders.cpp
```diff
@@ -73,11 +73,36 @@ void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixe
...
+                                         block_depth, stride_alignment);
+    case 4:
+        return SwizzleImpl<TO_LINEAR, 4>(output, input, width, height, depth, block_height,
+                                         block_depth, stride_alignment);
+    case 8:
+        return SwizzleImpl<TO_LINEAR, 8>(output, input, width, height, depth, block_height,
+                                         block_depth, stride_alignment);
+    case 16:
+        return SwizzleImpl<TO_LINEAR, 16>(output, input, width, height, depth, block_height,
+                                          block_depth, stride_alignment);
```
Morph1984, line 101: Add cases for the following:
3 bytes per pixel for 8_8_8 formats
6 bytes per pixel for 16_16_16 formats
12 bytes per pixel for 32_32_32 formats

src/video_core/textures/decoders.cpp
```diff
@@ -73,11 +73,36 @@ void Swizzle(std::span<u8> output, std::span<const u8> input, u32 bytes_per_pixe
...
+    default:
+        UNIMPLEMENTED_MSG("Unknown bytes_per_pixel {}", bytes_per_pixel);
```
Morph1984, line 103:
```suggestion
        UNREACHABLE_MSG("Invalid bytes_per_pixel={}", bytes_per_pixel);
```

src/video_core/textures/decoders.cpp
yzct12345: Fixed.

By the way which game and where are those cases used?

goldenx86:
@yzct12345 can I mention your PRs in the next Progress Report of yuzu?

goldenx86:
I'll take that as a yes, please confirm me if using your GitHub name is fine. Thanks!

yzct12345:
> locally merged to make ZLA

So that's where the flickering was coming from! Thank you! I thought I merged all of early access but I must have missed that one while rebasing.

> please confirm me if using your GitHub name is fine

Yes, I confirm it's fine.

src/video_core/textures/decoders.cpp
Morph1984: These are precautionary measures as there are 24, 48 and 96 bit pixel formats

Morph1984: The only 96 bit pixel format currently in use is R32G32B32_FLOAT

FernandoS27:
@yzct12345 do you have a discord account/email or whatever. I'de like to speak with you on the fantastic work you've done so far.

bylaws:
It's worth checking if the LUT for the Morton swizzle is actually beneficial, at least for us on ARM64 using a LUT decreased the speed heavily as bitfield ops were much cheaper. No idea if this will be the case on x86 though

yzct12345:
I tried removing the LUT and optimizing out the multiplications but performance was worse. My unit test lost 20% performance from 4 seconds to 5 seconds.

bylaws:
Oh okay, it would be cool if you could upload your test somewhere to help others with tuning

yzct12345:
It's not complicated. I simply call it over and over.

```
diff --git a/src/video_core/CMakeLists.txt b/src/video_core/CMakeLists.txt
index 1eb67c051..82cbfdcf8 100644
--- a/src/video_core/CMakeLists.txt
+++ b/src/video_core/CMakeLists.txt
@@ -229,6 +229,10 @@ endif()
 target_include_directories(video_core PRIVATE ${FFmpeg_INCLUDE_DIR})
 target_link_libraries(video_core PRIVATE ${FFmpeg_LIBRARIES})
 
+add_executable(swizzlebench textures/swizzlebench.cpp)
+
+target_link_libraries(swizzlebench PUBLIC video_core)
+
 add_dependencies(video_core host_shaders)
 target_include_directories(video_core PRIVATE ${HOST_SHADERS_INCLUDE})
 target_include_directories(video_core PRIVATE sirit ../../externals/Vulkan-Headers/include)
diff --git a/src/video_core/textures/swizzlebench.cpp b/src/video_core/textures/swizzlebench.cpp
new file mode 100644
index 000000000..d2220b76d
--- /dev/null
+++ b/src/video_core/textures/swizzlebench.cpp
@@ -0,0 +1,29 @@
+#include <iostream>
+
+/*
+static u32 st(u32 x, u32 y) {
+    return ((x & 32) << 3) | ((x & 16) << 1) | (x & 15) | ((y & 6) << 5) | ((y & 1) << 4);
+}
+
+using Tegra::Texture::SWIZZLE_TABLE;
+*/
+int main() {
+    std::vector<u8> input(8294400);
+    std::vector<u8> output(8355840);
+    for (int i = 1000; --i;) {
+        Tegra::Texture::UnswizzleTexture(output, input, 4, 1920, 1080, 1, 1, 0, 4);
+    }
+    //    for (u32 y = 0; y < SWIZZLE_TABLE.size(); y++) {
+    //        auto table = SWIZZLE_TABLE[y];
+    //        for (u32 x = 0; x < table.size(); x++) {
+    //            auto expect = table[x];
+    //            auto got = st(x, y);
+    //            if (expect != got) {
+    //                LOG_CRITICAL(Debug, "Expect {} got {}", expect, got);
+    //                abort();
+    //            }
+    //        }
+    //    }
+    return 0;
+}
```

Downclock CPU to 2 Ghz and run `time build/bin/swizzlebench`.

Wunkolo:
Just to add a data-point of something I just tried: I replaced the Morton-LUT with a [BMI2-based](https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set#BMI2_(Bit_Manipulation_Instruction_Set_2)) implementation using only two [pdep](https://www.felixcloutier.com/x86/pdep)s and an `or`.

The pdep implementation is equivalent to `SWIZZLE_TABLE`.

```cpp
inline u32 TegraMorton2D(u8 X, u8 Y)
{
        const u16 SpreadX = _pdep_u32(X, 0b100101111);
        const u16 SpreadY = _pdep_u32(Y, 0b011010000);
        return SpreadX | SpreadY;
}
```

Using your `swizzlebench` above this is what I got on mainline
```
hyperfine ./bin/swizzlebench
Benchmark #1: ./bin/swizzlebench
  Time (mean ± σ):      7.365 s ±  0.014 s    [User: 7.412 s, System: 0.026 s]
  Range (min … max):    7.351 s …  7.393 s    10 runs
```

this is what I got after checking out this PR:
```
hyperfine ./bin/swizzlebench
Benchmark #1: ./bin/swizzlebench
  Time (mean ± σ):      2.684 s ±  0.011 s    [User: 2.729 s, System: 0.025 s]
  Range (min … max):    2.661 s …  2.696 s    10 runs
```

This is what I got after checking out this PR and replacing all uses of `SWIZZLE_TABLE` with my `BMI2`-based implementation:
```
hyperfine ./bin/swizzlebench
Benchmark #1: ./bin/swizzlebench
  Time (mean ± σ):      2.679 s ±  0.013 s    [User: 2.723 s, System: 0.029 s]
  Range (min … max):    2.653 s …  2.694 s    10 runs
```

It's just a bit faster.

Another thing to possibly check out is if an [AVX2-based scatter-gather](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=gather&techs=AVX2) would speed up the inner `column` loop by doing all the LUT-lookups, offset calculations, and memcpys in parallel simd-sized batches.

Depends how the codebase feels about having arch-specific intrinsics.

Morph1984 (approved):

Wunkolo:
Couple more data points for reference later:

<details>
  <summary><b>(Click to expand)</b> | An AVX2 implementation, batching all the address calculations 8 at a time</summary>
```diff
-            for (u32 column = 0; column < width; ++column) {
+            u32 column = 0;
+
+            // Eight at a time
+            for (; column + 8 < width; column += 8) {
+                const __m256i column_vec = _mm256_add_epi32(
+                    _mm256_set1_epi32(column), _mm256_set_epi32(7, 6, 5, 4, 3, 2, 1, 0));
+                const __m256i x_vec =
+                    _mm256_mul_epu32(_mm256_add_epi32(column_vec, _mm256_set1_epi32(origin_x)),
+                                     _mm256_set1_epi32(BYTES_PER_PIXEL));
+                const __m256i offset_x_vec =
+                    _mm256_slli_epi32(_mm256_srli_epi32(x_vec, GOB_SIZE_X_SHIFT), x_shift);
+
+                const __m256i base_swizzled_offset =
+                    _mm256_add_epi32(offset_x_vec, _mm256_set1_epi32(offset_y + offset_z));
+
+                // AVX2 gather for parallel table look-up
+                const __m256i swizzled_offset = _mm256_add_epi32(
+                    base_swizzled_offset,
+                    _mm256_i32gather_epi32(
+                        reinterpret_cast<const int*>(table.data()),
+                        _mm256_and_si256(x_vec, _mm256_set1_epi32(GOB_SIZE_X - 1)), 4));
+
+                const __m256i unswizzled_offset = _mm256_add_epi32(
+                    _mm256_set1_epi32(slice * pitch * height + line * pitch),
+                    _mm256_mul_epu32(column_vec, _mm256_set1_epi32(BYTES_PER_PIXEL)));
+
+                // batch all reads with gather, in the case that `BYTES_PER_PIXEL` is 4
+                // const __m256i src_vec =
+                //     _mm256_i32gather_epi32(reinterpret_cast<const int*>(input.data()),
+                //                         TO_LINEAR ? unswizzled_offset : swizzled_offset, 4);
+
+                std::array<u32, 8> srcOffsets;
+                _mm256_store_si256(reinterpret_cast<__m256i*>(srcOffsets.data()),
+                                   TO_LINEAR ? unswizzled_offset : swizzled_offset);
+
+                std::array<u32, 8> dstOffsets;
+                _mm256_store_si256(reinterpret_cast<__m256i*>(dstOffsets.data()),
+                                   TO_LINEAR ? swizzled_offset : unswizzled_offset);
+
+                for (u8 i = 0; i < 8; ++i) {
+                    std::memcpy(output.data() + dstOffsets[i], input.data() + srcOffsets[i],
+                                BYTES_PER_PIXEL);
+                }
+            }
+
+            for (; column < width; ++column) {
```
</details>

On my 7900x:
```
hyperfine ./bin/swizzlebench    
Benchmark #1: ./bin/swizzlebench
  Time (mean ± σ):      1.970 s ±  0.013 s    [User: 2.003 s, System: 0.023 s]
  Range (min … max):    1.951 s …  1.997 s    10 runs
```

<details>
  <summary><b>(Click to expand)</b> An AVX512 implementation, 16 address calculations at a time, and scatter-gather</summary>
```diff
-            for (u32 column = 0; column < width; ++column) {
+            u32 column = 0;
+
+            // 16 at a time
+            for (; column + 16 < width; column += 16) {
+                const __m512i column_vec = _mm512_add_epi32(
+                    _mm512_set1_epi32(column),
+                    _mm512_set_epi32(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0));
+                const __m512i x_vec =
+                    _mm512_mul_epu32(_mm512_add_epi32(column_vec, _mm512_set1_epi32(origin_x)),
+                                     _mm512_set1_epi32(BYTES_PER_PIXEL));
+                const __m512i offset_x_vec =
+                    _mm512_slli_epi32(_mm512_srli_epi32(x_vec, GOB_SIZE_X_SHIFT), x_shift);
+
+                const __m512i base_swizzled_offset =
+                    _mm512_add_epi32(offset_x_vec, _mm512_set1_epi32(offset_y + offset_z));
+
+                // AVX2 gather for parallel table look-up
+                const __m512i swizzled_offset =
+                    _mm512_add_epi32(base_swizzled_offset,
+                                     _mm512_i32gather_epi32(
+                                         _mm512_and_si512(x_vec, _mm512_set1_epi32(GOB_SIZE_X - 1)),
+                                         reinterpret_cast<const int*>(table.data()), 4));
+
+                const __m512i unswizzled_offset = _mm512_add_epi32(
+                    _mm512_set1_epi32(slice * pitch * height + line * pitch),
+                    _mm512_mul_epu32(column_vec, _mm512_set1_epi32(BYTES_PER_PIXEL)));
+
+                if constexpr (BYTES_PER_PIXEL == 4) {
+                    const __m512i src_vec =
+                        _mm512_i32gather_epi32(TO_LINEAR ? unswizzled_offset : swizzled_offset,
+                                               reinterpret_cast<const int*>(input.data()), 1);
+                    _mm512_i32scatter_epi32(reinterpret_cast<int*>(output.data()),
+                                            TO_LINEAR ? swizzled_offset : unswizzled_offset,
+                                            src_vec, 1);
+                } else {
+                    std::array<u32, 16> srcOffsets;
+                    _mm512_store_si512(reinterpret_cast<__m512i*>(srcOffsets.data()),
+                                       TO_LINEAR ? unswizzled_offset : swizzled_offset);
+
+                    std::array<u32, 16> dstOffsets;
+                    _mm512_store_si512(reinterpret_cast<__m512i*>(dstOffsets.data()),
+                                       TO_LINEAR ? swizzled_offset : unswizzled_offset);
+
+                    for (u8 i = 0; i < 16; ++i) {
+                        std::memcpy(output.data() + dstOffsets[i], input.data() + srcOffsets[i],
+                                    BYTES_PER_PIXEL);
+                    }
+                }
+            }
+
+            for (; column < width; ++column) {
```
</details>

On my 7900x:
```
hyperfine ./bin/swizzlebench
Benchmark #1: ./bin/swizzlebench
  Time (mean ± σ):      1.811 s ±  0.019 s    [User: 1.841 s, System: 0.026 s]
  Range (min … max):    1.785 s …  1.848 s    10 runs
```

The same AVX512 implementation on my 11900k:
```
hyperfine ./swizzlebench
Benchmark #1: ./swizzlebench
  Time (mean ± σ):      1.169 s ±  0.017 s    [User: 1.172 s, System: 0.016 s]
  Range (min … max):    1.144 s …  1.208 s    10 runs
```

SIMD provides yet another 50% in performance, from `2.684` to `1.811` on my 7900x, perhaps for a later PR.

yzct12345:
> do you have a discord account/email or whatever

@FernandoS27 I joined as yzct12345#3833 but I don't have permission for #development

> AVX512

Sadly my AMD Kaveri desktop doesn't even have AVX2. I was hoping to improve yuzu until things got playable. My Intel Ice Lake laptop has but yuzu already runs fast enough on it. If you add this please don't make AVX512 mandatory.

> 50% in performance

Nice! You should open a PR.

On my side I'm not currently working on this function. I don't think it's the bottleneck anymore and I'm trying to optimize the other things the GPU thread is doing. If it still not fast enough I might look into moving the unswizzle into a Vulkan shader.

> AVX2

Shouldn't the compiler automatically do this if you allow it? This I mean passing `-march=native -mtune=native` as a flag.

Wunkolo:
> > AVX512
> 
> Sadly my AMD Kaveri desktop doesn't even have AVX2. I was hoping to improve yuzu until things got playable. My Intel Ice Lake laptop has but yuzu already runs fast enough on it. If you add this please don't make AVX512 mandatory.

I feel that anything above baseline x64(which mandates up to SSE2) should be optional. AVX512 is of course pretty exotic and should to be an opt-in feature that a user either enables for a runtime-dispatch or a build-flag to allow for the compiler to emit those instructions(checking for `#ifdef __AVX512F__`).

> > 50% in performance
> 
> Nice! You should open a PR.
> 
> On my side I'm not currently working on this function. I don't think it's the bottleneck anymore and I'm trying to optimize the other things the GPU thread is doing. If it still not fast enough I might look into moving the unswizzle into a Vulkan shader.


> > AVX2
> 
> Shouldn't the compiler automatically do this if you allow it? This I mean passing `-march=native -mtune=native` as a flag.

Yeah, if you do `-march=native` on an AVX2 machine it will define the `__AVX2__` preprocessor. [MSVC supports this as well with the](https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-160) `/arch` flag for AVX/AVX2/AVX512.

A discussion to be had though, is what pattern we want to implement for dispatching or selecting these particular arch-features. Either runtime-dispatch or compile-time build flags(and possibly having different versions of yuzu based on particular microarchitecture feature sets).

PR #6795, opened 2021-08-01
yuzu-cmd: hide mouse cursor when started fullscreen

sankasan:
Exposed `SDL_ShowCursor` to `EmuWindow` as `ShowCursor(bool)`. Both the derived classes (GL and VK) call this function when started with the fullscreen flag.

### DIFF
--- a/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
+++ b/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
@@ -122,6 +122,10 @@ void EmuWindow_SDL2::OnResize() {
     UpdateCurrentFramebufferLayout(width, height);
 }
 
+void EmuWindow_SDL2::ShowCursor(bool show_cursor) {
+    SDL_ShowCursor(show_cursor ? SDL_ENABLE : SDL_DISABLE);
+}
+
 void EmuWindow_SDL2::Fullscreen() {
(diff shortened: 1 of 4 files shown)


sankasan:
> license/cla Pending — Contributor License Agreement is not signed yet. 

Is this something I need to act on? If so could you point me to some docs about this?

v1993:
No, there is no need to do anything.

bunnei (approved):

goldenx86:
@sankasan Can I mention you for this PR in the next progress report?

sankasan:
> @sankasan Can I mention you for this PR in the next progress report?

But of course!

oblitum:
Folks, could something like this be done for the normal `yuzu` command, but when started with `-f` flag? I don't use `yuzu-cmd` as I have faced many issues with it regarding controller configurations and crashes, so `yuzu -f -g <game>` works just fine, but, on mainline 992, by default the cursor hides itself when using it normally, but for some reason, it sticks when I play yuzu from steam remote play, this doesn't happen with other emulators that have options to hide cursor, for example, cemu or dolphin have an option to hide on inactivity or to hide always, if I chose to hide on inactivity, it sticks, but if I pick hide always then I have no problems while playing remotely, yuzu is the sole one emulator which I'm still facing this (on Linux).

PR #6801, opened 2021-08-02
network: fix ternary operator in Socket::SendTo

spholz:
The ternary operator in core/network/network.cpp is written the wrong way round. The addrlen parameter for sendto should be non-zero if addr isn't a nullptr. Otherwise this call to sendto will generate an EINVAL error.


Morph1984 (approved):
Nice catch

PR #6809, opened 2021-08-03
cmake: Take advantage of FMA support

yzct12345:
yuzu's hardware requirements updated to require FMA https://github.com/yuzu-emu/yuzu-emu.github.io/blob/9e5ec00ba327c46d032a8e01b599a3dd17af8a9d/site/content/help/quickstart/index.md?plain=1#L33. If we tell the compiler to use those instructions yuzu will be faster.

The flags I put will support haswell and higher and bdver3 and higher. On the AMD side the wiki requires znver1 but I set it lower because I'm working on optimizing and lowering the requirements.

This PR is inspired by https://github.com/yuzu-emu/yuzu/pull/6790#issuecomment-891210862.


goldenx86:
Recent changes made FMA-less CPUs perform very well. Note that we recommend FMA support, but we don't enforce it.

yzct12345:
What would you use instead? Supporting older CPUs prevents the compiler from even using AVX. Would sandybridge be a better target?

yzct12345:
> apply them dynamically

I've never heard of any compiler except Intel's do that.

> Core 2 Duo and Phenom

Ok I will close this PR. I'm using march=native myself anyway.

goldenx86:
We could consider enforcing higher requirements after Windows 10 is out of support. But for now, enforcing FMA leaves too many Sandy Bridge users out, it's incredible the amount there are.

Moonlacer:
Sandy Bridge deserves it 👀

yzct12345:
I actually have a very old sandybridge laptop in some box in my basement. I never ever thought yuzu would run on it.

goldenx86:
Some optimizations to Dynarmic and fastmem made them very viable in most games. The exception being the big ones like BotW.

PR #6813, opened 2021-08-04
common: uuid: Add hex string to UUID constructor

Morph1984:
This allows for easily converting a hex string into a Common::UUID, which is backed by a 128 bit unsigned integer.

### DIFF
--- a/src/common/uuid.cpp
+++ b/src/common/uuid.cpp
@@ -6,9 +6,63 @@
 
 #include <fmt/format.h>
 
+#include "common/assert.h"
 #include "common/uuid.h"
 
 namespace Common {
+
+namespace {
+
+bool IsHexDigit(char c) {
+    return (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F');
+}
+
+u8 HexCharToByte(char c) {
+    if (c >= '0' && c <= '9') {
+        return static_cast<u8>(c - '0');
+    }
+    if (c >= 'a' && c <= 'f') {
+        return static_cast<u8>(c - 'a' + 10);
+    }
+    if (c >= 'A' && c <= 'F') {
+        return static_cast<u8>(c - 'A' + 10);
+    }
+    ASSERT_MSG(false, "{} is not a hexadecimal digit!", c);
+    return u8{0};
+}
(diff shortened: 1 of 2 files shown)


src/common/uuid.h
```diff
@@ -31,6 +35,20 @@ struct UUID {
...
+     * The hex string can be formatted in lowercase or uppercase, with or without the "0x" prefix.
+     *
+     * If the hex string is more than 32 hex characters long, only the last 32 characters are kept.
+     * For example: 0x12345abcde12345abcde12345abcde12345abcde is 40 characters long.
+     * This is converted into 0xde12345abcde12345abcde12345abcde, discarding 12345abc.
+     *
+     * If the hex string contains invalid (non-hex) characters, these are interpreted as 0.
+     * For example: 0xefgabc is converted into 0xef0abc.
+     */
+    [[nodiscard]] constexpr u128 HexStringToU128(std::string_view hex_string) const;
```
lioncash, line 50: This can't be used in a constexpr context due to the separation of implementation in the cpp file, and by extension the constructor above cant be constexpr either.

src/common/uuid.cpp
```diff
@@ -10,6 +10,45 @@
 
 namespace Common {
 
+namespace {
+
+constexpr u8 HexCharToByte(char c) {
+    if (c >= '0' && c <= '9') {
+        return c - '0';
```
lioncash, line 17: These returns need to be casted to u8, otherwise these are implicitly truncating an `int`

src/common/uuid.cpp
```diff
@@ -10,6 +10,45 @@
...
+    }
+    if (c >= 'A' && c <= 'F') {
+        return c - 'A' + 10;
+    }
+    return 0;
+}
+
+} // Anonymous namespace
+
+constexpr u128 UUID::HexStringToU128(std::string_view hex_string) const {
```
lioncash, line 30: Shouldn't this function assert if a given string has enough bytes in it that they exceed the 128-bit range?

lioncash: I don't think silently discarding characters in the string is a good idea, because it can hide genuine logic bugs.

bunnei (approved):

PR #6817, opened 2021-08-04
Add description to fast gpu time option

gidoly:


goldenx86:
I would change max to "maximum".
Also I think @FernandoS27 needs to input here, I think this may be ignored on a game level.

goldenx86:
Yeah it's main goal is to improve stability and RAM use by avoiding constant redraws. It increases resolution as a bonus.

Shahil-Ayato:
> Yeah it's main goal is to improve stability and RAM use by avoiding constant redraws. It increases resolution as a bonus.

I think you mean it just keeps the resolution set to the highest native resolution?. But yeah avoiding constant resolution changes is something that’s very good to do as you want to always disable it because recreating framebuffers is super expensive while on actual hardware it has zero cost. Sadly it doesn’t work with luigis mansion since it does some stuff differently I think.

src/yuzu/configuration/configure_graphics_advanced.ui
```diff
@@ -88,6 +88,9 @@
         </item>
         <item>
          <widget class="QCheckBox" name="use_fast_gpu_time">
+          <property name="toolTip">
+            <string>Enable Fast GPU Time. This option will force games to run at their max native resolution.</string>
```
CrusadingNinja, line 92:
```suggestion
            <string>Enables Fast GPU Time. This option will force most games to run at their maximum native resolution.</string>
```

CrusadingNinja: Minor grammar nit, adds disclaimer that not all games work with this option, and incorporates Golden's suggestion.

Shahil-Ayato: Maybe Instead do `highest` instead of `maximum` ?

FernandoS27:
Well kinda. It's hack either way. we are just reporting the GPU time as going on a lot faster than it really is so. I'de love that you add [HACK] to the name. As this option and "Asynchronous shaders" are the only 2 hack options.

bunnei (approved):

PR #6823, opened 2021-08-05
memory: Clean up code

yzct12345:
This deduplicates code, fixes all warnings and removes unused code.

This also fixes some logging bugs. I plan to use the new `WalkBlock` function later.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -5,6 +5,10 @@
 #include <algorithm>
 #include <cstring>
 
+#define BOOST_HANA_CONFIG_ENABLE_STRING_UDL
+#include <boost/hana/string.hpp>
+#undef BOOST_HANA_CONFIG_ENABLE_STRING_UDL
+
 #include "common/assert.h"
 #include "common/atomic_ops.h"
 #include "common/common_types.h"
@@ -18,6 +22,8 @@
 #include "core/hle/kernel/k_process.h"
 #include "core/memory.h"
 #include "video_core/gpu.h"
+
+using namespace boost::hana::literals;
 
 namespace Core::Memory {
 
@@ -66,18 +72,6 @@ struct Memory::Impl {
         }
 
         return system.DeviceMemory().GetPointer(paddr) + vaddr;
-    }
-
-    [[nodiscard]] u8* GetPointer(const VAddr vaddr) const {
-        const uintptr_t raw_pointer = current_page_table->pointers[vaddr >> PAGE_BITS].Raw();
-        if (u8* const pointer = Common::PageTable::PageInfo::ExtractPointer(raw_pointer)) {
-            return pointer + vaddr;
-        }
-        const auto type = Common::PageTable::PageInfo::ExtractType(raw_pointer);
-        if (type == Common::PageType::RasterizerCachedMemory) {
-            return GetPointerFromRasterizerCachedMemory(vaddr);
-        }
-        return nullptr;
     }
 
     u8 Read8(const VAddr addr) {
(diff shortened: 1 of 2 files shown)


src/core/memory.cpp
```diff
@@ -517,7 +393,6 @@ struct Memory::Impl {
                 case Common::PageType::Unmapped:
                     // It is not necessary for a process to have this region mapped into its address
                     // space, for example, a system module need not have a VRAM mapping.
-                    break;
```
lioncash, line 520: It's preferable to keep these two memory categories separated from one another, rather than falling through, since they're significantly different categories from one another.

src/core/memory.cpp
```diff
@@ -789,12 +664,11 @@ void Memory::UnmapRegion(Common::PageTable& page_table, VAddr base, u64 size) {
...
 }
 
-bool Memory::IsValidVirtualAddress(const Kernel::KProcess& process, const VAddr vaddr) const {
-    return impl->IsValidVirtualAddress(process, vaddr);
-}
-
 bool Memory::IsValidVirtualAddress(const VAddr vaddr) const {
-    return impl->IsValidVirtualAddress(vaddr);
+    const Kernel::KProcess& process = *system.CurrentProcess();
+    const auto& pageTable = process.PageTable().PageTableImpl();
```
lioncash, line 669:
```suggestion
    const auto& page_table = process.PageTable().PageTableImpl();
```

src/core/memory.h
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014 Citra Emulator Project
+// Copyright 2021 Citra Emulator Project
```
lioncash, line 1: This doesn't need to be updated. This is to establish a creation year, not a "last updated" year.

src/core/memory.cpp
```diff
@@ -230,199 +211,98 @@ struct Memory::Impl {
...
-                u8* const dest_ptr = pointer + page_offset + (page_index << PAGE_BITS);
-                std::memcpy(dest_ptr, src_buffer, copy_amount);
-                break;
-            }
-            case Common::PageType::RasterizerCachedMemory: {
-                u8* const host_ptr{GetPointerFromRasterizerCachedMemory(current_vaddr)};
+            },
+            [&system = system, &src_buffer](const VAddr current_vaddr,
+                                            const std::size_t copy_amount, u8* const host_ptr) {
+                if (!UNSAFE) {
```
lioncash, line 271:
```suggestion
                if constexpr (!UNSAFE) {
```

src/core/memory.cpp
```diff
@@ -230,199 +211,98 @@ struct Memory::Impl {
...
                 std::memcpy(dest_buffer, src_ptr, copy_amount);
-                break;
-            }
-            case Common::PageType::RasterizerCachedMemory: {
-                const u8* const host_ptr{GetPointerFromRasterizerCachedMemory(current_vaddr)};
+            },
+            [&system = system, &dest_buffer](const VAddr current_vaddr,
+                                             const std::size_t copy_amount,
+                                             const u8* const host_ptr) {
+                if (!UNSAFE) {
```
lioncash, line 238:
```suggestion
                if constexpr (!UNSAFE) {
```

src/core/memory.cpp
```diff
@@ -1,11 +1,9 @@
-// Copyright 2015 Citra Emulator Project
+// Copyright 2021 Citra Emulator Project
```
lioncash, line 1: Ditto, this doesn't need to be updated.

src/core/memory.cpp
yzct12345: `Clang-Tidy: Switch has 2 consecutive identical branches`

lioncash: Just because clang-tidy suggests it doesn't mean it needs to be done.

yzct12345: I changed it to "NOLINT(bugprone-branch-clone)"

src/core/memory.cpp
```diff
@@ -453,51 +447,106 @@ struct Memory::Impl {
...
-     *           is undefined.
-     *
-     * @returns The instance of T read from the specified virtual address.
+     * @tparam NAME The caller name like "Read"_s or "Write"_s.
+     * @tparam BYTES The number of bits written. 0 is for read and sizeof(T) is for write.
+     * @tparam SUFFIX A suffix. ""_s is for read and " = 0x{:016X}" is for write.
      */
-    template <typename T>
-    T Read(VAddr vaddr) {
+    template <boost::hana::string NAME, int BYTES, boost::hana::string SUFFIX>
```
lioncash, line 456: This seems like overkill. Especially given it adds a dependency on hana just to print out a logging message.

yzct12345: I thought you like `constexpr` :)

bunnei:
Going to briefly EA-tag this to make sure there are no issues. Will merge if things look good after a day or so. Thanks!

bunnei (approved):

PR #6829, opened 2021-08-06
settings_ui: Disable async shaders by a simple comment addition

Moonlacer:
Disables async shaders in both global and custom configuration by making anything regarding them a comment, so as not to break anything being worked on (I hope)

### DIFF
--- a/src/yuzu/configuration/configure_graphics_advanced.cpp
+++ b/src/yuzu/configuration/configure_graphics_advanced.cpp
@@ -23,11 +23,11 @@ ConfigureGraphicsAdvanced::~ConfigureGraphicsAdvanced() = default;
 void ConfigureGraphicsAdvanced::SetConfiguration() {
     const bool runtime_lock = !Core::System::GetInstance().IsPoweredOn();
     ui->use_vsync->setEnabled(runtime_lock);
-    ui->use_asynchronous_shaders->setEnabled(runtime_lock);
+    //ui->use_asynchronous_shaders->setEnabled(runtime_lock);
     ui->anisotropic_filtering_combobox->setEnabled(runtime_lock);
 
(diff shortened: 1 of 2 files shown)


Moonlacer:
Clang is annoying, but I won't be able to try to fix it much until I get back from my trip

yzct12345:
What's wrong with async shaders?

Moonlacer:
> What's wrong with async shaders?

It breaks many games in yuzu, atm

yzct12345:
> many

That means some games don't get broken. For those games you will be taking away the freedom of the user to use this feature. If async shaders are breaking all of your games you can simply uncheck the checkbox on your computer instead of on everyone's computers.

Moonlacer:
> > many
> 
> That means some games don't get broken. For those games you will be taking away the freedom of the user to use this feature. If async shaders are breaking all of your games you can simply uncheck the checkbox on your computer instead of on everyone's computers.

For the games it doesn't break, there is no benefit to them being enabled

v1993:
I'm curious if this is indeed true. Any word from devs?

Moonlacer:
> I'm curious if this is indeed true. Any word from devs?

Maybe @goldenx86 could give you a clearer answer on this, this is just what I have tested and seen from being in the Discord server

Moonlacer:
> I'm curious if this is indeed true. Any word from devs?

Also @v1993, if you are in the Discord, you should be able to look at this for a bit more info: https://discord.com/channels/398318088170242053/668532415257509958/863215731172048927

v1993:
That discord message was related to bug in Hades before its merge. It was ultimately fixed before getting into master.

Moonlacer:
I think I may just open a different PR adding onto #6817 that shows that async may break things

PR #6832, opened 2021-08-07
kernel: Various improvements to scheduler

bunnei:
The primary goal of this PR is to continue to improve the accuracy of our scheduler implementation. There are a few changes here:
* Implement more complete tracking of KThread dispatch count through KScopedDisableDispatch.
* Improved KScheduler EnableScheduling/DisableScheduling routines as a result.
* Overall simplification of a few KScheduler implementations.
* Use std::jthread for CPU threads.
* Improved emulation of host-only HLE threads as simulated "guest threads" on core 3.
* Improved management of some KEvents in services. This is not strictly related to this change, but I noticed some issues with these and this change, so I incorporated this improvement.

With overall better tracking of kernel objects, I have noticed several known shutdown crashes being fixed with these changes.

### DIFF
--- a/src/core/cpu_manager.cpp
+++ b/src/core/cpu_manager.cpp
@@ -21,34 +21,25 @@ namespace Core {
 CpuManager::CpuManager(System& system_) : system{system_} {}
 CpuManager::~CpuManager() = default;
 
-void CpuManager::ThreadStart(CpuManager& cpu_manager, std::size_t core) {
-    cpu_manager.RunThread(core);
+void CpuManager::ThreadStart(std::stop_token stop_token, CpuManager& cpu_manager,
+                             std::size_t core) {
+    cpu_manager.RunThread(stop_token, core);
 }
 
 void CpuManager::Initialize() {
     running_mode = true;
     if (is_multicore) {
         for (std::size_t core = 0; core < Core::Hardware::NUM_CPU_CORES; core++) {
-            core_data[core].host_thread =
-                std::make_unique<std::thread>(ThreadStart, std::ref(*this), core);
+            core_data[core].host_thread = std::jthread(ThreadStart, std::ref(*this), core);
         }
     } else {
-        core_data[0].host_thread = std::make_unique<std::thread>(ThreadStart, std::ref(*this), 0);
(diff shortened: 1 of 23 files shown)


src/core/hle/kernel/k_thread.h
```diff
@@ -752,4 +771,16 @@ class KThread final : public KAutoObjectWithSlabHeapAndContainer<KThread, KSynch
     }
 };
 
+class KScopedDisableDispatch {
+public:
+    explicit KScopedDisableDispatch(KernelCore& kernel_) : kernel{kernel_} {
```
lioncash, line 776:
```suggestion
    [[nodiscard]] explicit KScopedDisableDispatch(KernelCore& kernel_) : kernel{kernel_} {
```

Will allow the compiler to flag logic bugs where an instance is constructed and destroyed immediately.

e.g. accidentally doing

```cpp
{
    KScopedDisableDispatch{kernel}
    // more code...
}
```

instead of:

```cpp

{
    KScopedDisableDispatch dd{kernel}
    // more code...
}
```

src/core/cpu_manager.cpp
```diff
@@ -21,34 +21,25 @@ namespace Core {
...
     running_mode = true;
     if (is_multicore) {
         for (std::size_t core = 0; core < Core::Hardware::NUM_CPU_CORES; core++) {
-            core_data[core].host_thread =
-                std::make_unique<std::thread>(ThreadStart, std::ref(*this), core);
+            core_data[core].host_thread = std::jthread(ThreadStart, std::ref(*this), core);
         }
     } else {
-        core_data[0].host_thread = std::make_unique<std::thread>(ThreadStart, std::ref(*this), 0);
+        core_data[0].host_thread = std::jthread(ThreadStart, std::ref(*this), 0);
```
FernandoS27, line 36: core 0 is valid, change this to -1 or something else.

bunnei:
Re-tagging, as the reported freezes have been fixed

FernandoS27:
untagging, it breaks SC.

lioncash (approved):

PR #6833, opened 2021-08-07
cmake: allow unbundling some external libraries

abouvier:
* cubeb
* inih
* xbyak


abouvier:
Oops, `dynarmic` [also needs to be edited](https://github.com/MerryMage/dynarmic/pull/639) to fix the building of `yuzu` when the external `xbyak` is used.

CMakeLists.txt
```diff
@@ -27,6 +27,12 @@ option(YUZU_USE_BUNDLED_LIBUSB "Compile bundled libusb" OFF)
 
 option(YUZU_USE_BUNDLED_FFMPEG "Download/Build bundled FFmpeg" "${WIN32}")
 
+option(YUZU_USE_BUNDLED_XBYAK "Compile bundled xbyak" ON)
+
+option(YUZU_USE_BUNDLED_INIH "Compile bundled inih" ON)
+
+option(YUZU_USE_BUNDLED_CUBEB "Compile bundled cubeb" ON)
```
Tachi107, line 34: Adding an option for every library isn't always ideal in my opinion. For small libraries I think that using the system's one if possible and falling back to the subprojects otherwise is preferable

t's already done with most libraries, see
https://github.com/yuzu-emu/yuzu/blob/3592628302409dce264bb017d70880ecee74173c/CMakeLists.txt#L169-L174 and
https://github.com/yuzu-emu/yuzu/blob/3592628302409dce264bb017d70880ecee74173c/externals/CMakeLists.txt#L119-L123

liushuyu:
I think this would be superseded by #7610

Tatsh:
> I think this would be superseded by #7610

No longer. I hope this PR gets merged instead. My #7610 only covers Opus now.

CMakeLists.txt
Tatsh: On the latter one, my PR does it now with an explicit option to enable system Opus. The reason to have it be explicit is to disallow opportunist linking to system Opus and to avoid using the bundled code if this is not desired.

bunnei:
@abouvier can you rebase this? Sorry for the delay, but I think we can merge this now.

bunnei:
(also, some of these maybe can use conan instead of externals?)

abouvier:
I rebased and also added support for `dynarmic` and `httplib`.
Sorry about `conan` but I don't use it and I'm not sure how it works. I'm only interested in using system libraries.

CMakeLists.txt
abouvier: I don't understand why the size of the library matters here. The only condition I see is the good support of `find_package()` from the library.

liushuyu (approved):
LGTM (although the CMake scripts are a bit more verbose now, I think that couldn't be helped)

CMakeLists.txt
Tachi107: With size I mean the complexity of the library, not the size in bytes. If a library is complex, has a lot of optional features, and linking to the system one can create issues in some cases then having an option to force its bundling is necessary, but for simple cases like `inih` I think that having an option is a bit overkill.

I mean, it doesn't certainly cause big issues, but I believe that if some code/logic can be avoided it should be preferable to do so

Tatsh:
Do we really want Dynarmic like this? It seems like between Citra and Yuzu, Dynarmic HEAD rarely works with both. Usually fails to compile with one.

Tachi107:
Until Dynarmic starts tagging releases more frequently (i.e. semantic versioning) it will be hard to reliably use a shared version of it

Tatsh:
> Until Dynarmic starts tagging releases more frequently (i.e. semantic versioning) it will be hard to reliably use a shared version of it

I'm all for that but I do not expect it right now due to it being heavily worked on. In my packaging, Yuzu and Citra both get their respective bundled versions of Dynarmic.

CMakeLists.txt
```diff
@@ -538,6 +548,59 @@ endif()
...
+    if (dynarmic_FOUND)
+        add_library(dynarmic ALIAS dynarmic::dynarmic)
+    else()
+        message(STATUS "dynarmic not found, falling back to externals")
+        set(YUZU_USE_BUNDLED_DYNARMIC ON)
+    endif()
+endif()
+
+if (ENABLE_WEB_SERVICE AND NOT YUZU_USE_BUNDLED_HTTPLIB)
+    find_package(httplib)
```
Tachi107, line 595: Some distros (like [Debian](https://packages.debian.org/testing/amd64/libcpp-httplib-dev/filelist)) only ship a pkg-config file instead of a CMake config file. You could handle this by writing a custom `Findhttplib.cmake` that first runs `find_dependency(httplib)` and then tries [`pkg_search_module(cpp-httplib IMPORTED_TARGET cpp-httplib)`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_search_module). If placed in `externals/find-modules` it should automatically work.

abouvier: Wouldn't it be better to patch the Debian package to include the existing upstream cmake config file instead of creating a `Findhttplib.cmake` file in every project using `cpp-httplib`? Or maybe adding this `Findhttplib.cmake` file directly in the `cpp-httplib` repo?

Tachi107: I've looked into that, but upstream's config file makes use of `install(EXPORT)`, a feature that is available only when building with CMake, and the Debian package is built with Meson.

I could submit a patch upstream that installs a pre-generated CMake config file, but it would be only included in new releases. Some distros (notably Ubuntu 22.04 LTS) won't ship it.

abouvier: OK, let's try :p

abouvier:
Rebased again and also added support for discord-rpc.

externals/find-modules/FindDiscordRPC.cmake
```diff
@@ -0,0 +1,34 @@
...
+  NAMES discord_rpc.h
+)
+
+find_library(DiscordRPC_LIBRARY
+  NAMES discord-rpc
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(DiscordRPC
+  FOUND_VAR DiscordRPC_FOUND
```
Tachi107, line 12: FOUND_VAR is deprecated, ignored, and shouldn't be use, see https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html

The DiscordRPC_FOUND variable is automatically set to true when all the REQUIRED_VARS are set to true. Unless you need some other special variable to be defined, it should be sufficient to add DiscordRPC_FOUND to the REQUIRED_VARS, like I did in the FindMbedTLS.cmake and FindSoundTouch.cmake modules.

Tachi107:
Hey @abouvier, could you please rebase this again? The patch can't be cleanly applied since #8134 was merged

externals/find-modules/FindDiscordRPC.cmake
abouvier: You're right, this option is now obsolete. But you're wrong too, because `find_package_handle_standard_args` is supposed to define the `<PackageName>_FOUND` variable and not validate it.

Tachi107: I mean, since the module internally uses `DiscordRPC_FOUND` to internally mark the module as found, you could simply use that variable to tell `find_package_handle_standard_args` when the module is to be considered found. Or you could drop `DiscordRPC_FOUND` entirely

abouvier:
> Hey @abouvier, could you please rebase this again? The patch can't be cleanly applied since #8134 was merged

Done! See you in six months for the next rebase 😅

Tachi107 (changes requested):

externals/find-modules/Findxbyak.cmake
```diff
@@ -0,0 +1,38 @@
+find_package(PkgConfig QUIET)
```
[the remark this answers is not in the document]
abouvier: I don't find it in the conan package :$

german77:
Dynarmic changes frequently. Often a mismatch breaks the build. I don't think is good to not use the one in externals

Tachi107:
That's true, but recently Dynarmic started tagging releases more often, 
thus making it easier to stick to a specific version.

abouvier:
Currently the builds are failing because of https://github.com/conan-io/conan-center-index/pull/11690 and https://github.com/conan-io/conan-center-index/issues/11718.

abouvier:
OK I think it's good enough to be merged now :p

Tatsh:
Why are you removing `pkg_check_modules` for zstd? My system does not have a CMake config for zstd 1.5.2.

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find zstd: found neither zstdConfig.cmake nor zstd-config.cmake
  (Required is at least version "1.5")
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:269 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:579 (_FPHSA_HANDLE_FAILURE_CONFIG_MODE)
  externals/find-modules/Findzstd.cmake:7 (find_package_handle_standard_args)
  CMakeLists.txt:193 (find_package)
```

abouvier:
> Why are you removing `pkg_check_modules` for zstd? My system does not have a CMake config for zstd 1.5.2.

Because upstream zstd provides a CMake config file since nearly 2 years, so I thought that every system will have one. But zstd can be compiled/installed without using CMake, so I was wrong :p
I will update the find modules.

Tachi107:
>Because upstream zstd provides a CMake config file since nearly 2 years, so I thought that every system will have one. 

As far as I understand their "main" build system is Make, so Meson, CMake, and Visual Studio solutions are second class citizens

Tatsh:
How about an update to allow system cpp-jwt?

abouvier:
It's already supported :p
There is no find module for cpp-jwt because it use CMake by default, so there is a system CMake config file installed.

CMakeLists.txt
```diff
@@ -163,24 +181,55 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 # System imported libraries
 # =======================================================================
 
-find_package(fmt 8.0.1 REQUIRED CONFIG)
-find_package(nlohmann_json 3.8 REQUIRED CONFIG)
+find_package(fmt 8.0.1 REQUIRED)
+find_package(lz4 1.8 REQUIRED)
+find_package(nlohmann_json 3.8 REQUIRED)
```
Tachi107, line 186: Why did you remove `CONFIG` from fmt and nlohmann_json?

abouvier: Because it's useless. There is no more find modules for these two libs in yuzu repo, so system/vcpkg config files will be used anyway.

externals/find-modules/FindVulkanHeaders.cmake
```diff
@@ -0,0 +1,47 @@
+# SPDX-FileCopyrightText: 2022 Alexandre Bouvier <contact@amb.tf>
```
Tachi107, line 1: Instead of using a custom find module, I think it would be better to use the one used by upstream Vulkan: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/cmake/FindVulkanHeaders.cmake

abouvier: Mine is better :p
It creates an imported target with the same name that the [(future?) export target](https://github.com/KhronosGroup/Vulkan-Headers/pull/207), and it detects the version (allowing version constraint in `find_package`).

Tachi107: If so, upstream it :)

abouvier: This find module will be useless if/when the PR adding CMake config files to Vulkan-Headers is merged. Update it seems a wasted effort.

Tachi107:
Fair enough :)

-- 
OpenPGP key: 66DE F152 8299 0C21 99EF  A801 A8A1 28A8 AB1C EE49

abouvier:
Is there something blocking the merge of this PR? I'm ready to amend anything, just tell me!

Tachi107:
> Is there something blocking the merge of this PR? I'm ready to amend anything, just tell me!

I'm still not convinced about adding a large number of options to toggle the use of bundled subprojects. It's more code, hence more fragile, and goes against what the rest of yuzu's build script does: using system deps when available, and falling back to vcpkg/externals/whatever otherwise.

I've also recently realised that explicitly defining options is not needed, because in CMake you can define [`CMAKE_DISABLE_FIND_PACKAGE_<PackageName>`](https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html) to make `find_package()` skip dependency lookups for a given package.

If I were to write patch with the same goal of this one, I would do something like this:

```cmake
find_package(Opus 1.3)
if (NOT Opus_found)
    add_subdirectory(opus EXCLUDE_FROM_ALL)
endif()
```

This way yuzu would use the bundled Opus version only if not installed on the host, or if the user explicitly requires to use the bundled version by invoking `cmake` with the argument `-DCMAKE_DISABLE_FIND_PACKAGE_Opus=true`

CMakeLists.txt
```diff
@@ -175,24 +173,39 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
...
-find_package(ZLIB 1.2 REQUIRED)
+find_package(enet 1.3.13)
+find_package(fmt 9 REQUIRED)
+find_package(inih 52)
+find_package(lz4 1.9.3 REQUIRED)
+find_package(nlohmann_json 3.10.5 REQUIRED)
+find_package(Opus 1.3.1)
+find_package(VulkanHeaders 1.3.213)
+find_package(ZLIB 1.2.12 REQUIRED)
+find_package(zstd 1.5.2 REQUIRED)
```
Tachi107, line 184: Are these strict versions requirements needed? For example, why zlib 1.2.12 and not simply 1.2?

externals/find-modules/Findzstd.cmake
```diff
@@ -1,19 +1,31 @@
...
+            add_library(zstd::zstd ALIAS zstd::libzstd_static)
+        endif()
+    endif()
+else()
+    find_package(PkgConfig QUIET)
+    if (PKG_CONFIG_FOUND)
+        pkg_search_module(ZSTD QUIET IMPORTED_TARGET GLOBAL libzstd)
+        if (ZSTD_FOUND)
+            add_library(zstd::zstd ALIAS PkgConfig::ZSTD)
+            set(zstd_VERSION ${ZSTD_VERSION})
```
Tachi107, line 24: Doing this variable parkour is not needed anymore, as I've fixed the CMake logging issue upstream: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7527

externals/find-modules/Findzstd.cmake
```diff
@@ -1,19 +1,31 @@
...
+        endif()
     endif()
+    find_package_handle_standard_args(zstd
+        REQUIRED_VARS ZSTD_LINK_LIBRARIES
```
Tachi107, line 28: I think that listing the version variable in the list of required variables is still needed, but I may be wrong

externals/CMakeLists.txt
```diff
@@ -74,25 +79,31 @@ if (YUZU_USE_EXTERNAL_SDL2)
...
+    add_library(DiscordRPC::discord-rpc ALIAS discord-rpc)
 endif()
 
 # Sirit
-add_subdirectory(sirit)
+add_subdirectory(sirit EXCLUDE_FROM_ALL)
 
-if (ENABLE_WEB_SERVICE)
+# cpp-httplib
+if (ENABLE_WEB_SERVICE AND NOT TARGET httplib::httplib)
```
Tachi107, line 106: I think that it may be better to move the `AND NOT TARGET httplib::httplib` part at line 112/123

CMakeLists.txt
abouvier: It's the exact version of libs pulled by the current `vcpkg` submodule (or directly the version of submodules in the externals directory). It's to be sure that the system uses the minimum versions required/tested by yuzu.

externals/find-modules/Findzstd.cmake
abouvier: Nice, but not everyone will have the last version of CMake anytime soon :p

externals/find-modules/Findzstd.cmake
abouvier: Looking at `FindPkgConfig.cmake` it seems that `<XXX>_LINK_LIBRARIES` variable is only set when the module is already found and `<XXX>_FOUND` is defined.

externals/CMakeLists.txt
abouvier: Why? I don't understand :p
OpenSSL is only needed for the external httplib.

CMakeLists.txt
eli-schwartz: This sounds like a good way to drop support for even slightly older systems without any particular belief that it shouldn't work.

abouvier: I know, but otherwise the project devs complain that they will receive zillion of bug reports about incompatible libraries. It's generally their main excuse to bundle everything in the first place :p

eli-schwartz: So just make the bundled versions the default then? What is even the point of allowing system libraries as an ***option*** if the option is useless because it can never actually be used anywhere except for the two distros that package software released less than a month ago?

Why do you even suspect they are incompatible in the first place?

Tachi107: Version constraints are necessary in general, but I don't see why requiring, for example, zlib 1.2.12 instead of simply 1.2, as patch releases are all compatible between each other and won't cause incompatibilities.

Unless the package maintainer has done some crazy patching with it, the system library won't introduce any compatibility issues.

externals/find-modules/Findzstd.cmake
Tachi107: Yeah but it doesn't really cause issues to remove the workaround, it's only a minor cosmetic logging bug, not worth adding more lines of code just to work around it in my opinion. It would also kinda undo #8660 :)

externals/CMakeLists.txt
Tachi107: Oh ok, I thought it was used somewhere else

CMakeLists.txt
abouvier: You are preaching to the choir, but then you have answers [like this](https://github.com/cemu-project/Cemu/pull/136#issuecomment-1234221809).

externals/find-modules/Findzstd.cmake
Tachi107: Yep, you're right

CMakeLists.txt
Tachi107: That's unrelated, as what they say there is that they only want to use submodules. yuzu already uses system libraries, and anyway we are talking about whether requiring 1.2.12 instead of 1.2 makes sense or not. Since vcpkg already doesn't use a fixed dependency version, it doesn't.

Tachi107: And when I say "Unless the package maintainer has done some crazy patching with it, the system library won't introduce any compatibility issues", this also includes vcpkg packages, as they are patched by the vcpkg package maintainers too.

externals/CMakeLists.txt
```diff
@@ -29,12 +21,10 @@ endif()
...
 add_subdirectory(glad)
 
-# inih
-add_subdirectory(inih)
-
 # mbedtls
 add_subdirectory(mbedtls EXCLUDE_FROM_ALL)
-target_include_directories(mbedtls PUBLIC ./mbedtls/include)
+target_include_directories(mbedcrypto PUBLIC ./mbedtls/include)
+add_library(MbedTLS::mbedcrypto ALIAS mbedcrypto)
```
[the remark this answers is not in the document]
Tachi107: I've rebased #7044 and also incorporated the `mbedtls` -> `MbedTLS::mbedcrypto` change in that PR. As this doesn't really add the possibility to use MbedTLS as a system library, could you please drop these changes? This way this patch and the one from #6833 becomes easier.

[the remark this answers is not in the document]
abouvier: Done!

abouvier:
I removed the module `FindVulkanHeaders` because CMake already provides the module `FindVulkan` which exports many imported targets, including `Vulkan::Headers`.

liamwhite (approved):

lioncash:
Note that this breaks performing clean builds on Windows. e.g. lz4 handled by vcpkg is no longer found (along with a few other libraries)

image

Reverting this causes CMake generation on Windows to resume working as normal

abouvier:
Probably needs some tweaking in [Findlz4.cmake](https://github.com/yuzu-emu/yuzu/blob/master/externals/find-modules/Findlz4.cmake). What is the name of the pc/cmake file exported by vcpkg for lz4 on Windows?

Tachi107:
Strange. As shown in the snippet below, lz4 is searched using CONFIG mode just like before. Could you please remove the QUIET options from the find module to see what is failing exactly?

https://github.com/yuzu-emu/yuzu/blob/3b19f741bd6a19f603858e6cdf8db41516c7075f/externals/find-modules/Findlz4.cmake#L6

lioncash:
This is what is logged out with `QUIET` removed:

image

abouvier:
The cmake file exported by vcpkg contains a erroneous version string ("unknown" instead of e.g. "1.9.4"), so the version check fails.

lioncash:
I mean, I get that. But how is it going to be fixed? We don't have control over what vcpkg supplies and we currently use vcpkg as our package manager on Windows for building. Leaving it as is isn't an option, otherwise the patch needs to be reverted.

abouvier:
Well, you can just remove the version constraint on lz4 until vcpkg is fixed: `find_pacakge(lz4 REQUIRED)`

eli-schwartz:
> We don't have control over what vcpkg supplies

Why can't you just open a bug report for vcpkg stating that its lz4 package is currently broken and needs to be fixed in order to be usable?

It's not like vcpkg is proprietary trade secrets sold under NDA via twice-a-decade code dumps plus a reseller. You have a great deal of power even if you don't have the ability to personally merge a fix into vcpkg.

> Leaving it as is isn't an option, otherwise the patch needs to be reverted.

This sacrifices correctness and the ability to guarantee a sane, working build in the general case, in the name of preserving the status quo.

It's certainly a valid opinion to hold, but that doesn't automatically mean that all project maintainers agree the correct thing to do is be wrong in order to work around a solvable vcpkg bug

lioncash:
> Why can't you just open a bug report for vcpkg stating that its lz4 package is currently broken and needs to be fixed in order to be usable?

I did. When I wrote the post I meant in terms of "what are we going to do until the fix lands?"

> This sacrifices correctness and the ability to guarantee a sane, working build in the general case, in the name of preserving the status quo. ....

I dunno why you opted to post like an ass here, but go off king. Anyway, moving on.

> I wonder who added that instead of submitting a vcpkg bug report. :)

I haven't touched the build system since the transition to vcpkg, so it's anyone's guess =)

PR #6834, opened 2021-08-07
Respect Vulkan bufferImageGranularity

### LINKED ISSUE #6764: Super Mario Odyssey - Vertex explosion after breaking "cube"
After breaking a [moon rock](https://www.mariowiki.com/Moon_Rock#/media/File:Mysterious_Cube_SMO.jpg) on desert kingdom, I'm seeing exploding vertices. I can provide a save if you want. The issue isn't 100% reliably reproducible but I did manage to reproduce it a second time with that save without too many attempts.
[Once I also ended up with some minor flickering instead.](https://drive.google.com/file/d/1J4KUHJIKEXMwZjwAowvFa_bHFXO0ACEn/view?usp=sharing)

yuzu bug

Super Mario Odyssey 1.3
Nvidia GTX 1070, driver: 470.57.02
Yuzu commit b2b3fcdccda8399735a0307114a0ad5c2f118421, Linux build
It does not seem to happen with the OpenGL renderer.

I took a look at it in Renderdoc and it seems like the Index buffer is messed up, causing out of bounds reads.
image

On the other hand, the mesh looks fine (I think) in the VS In tab of the mesh viewer and the issue only shows up after the vertex shader.

image
image

image
The data in the index buffer only seems sensible starting with index 128,.

I compared the same draw call in another capture where I didn't run into the bug and the first 128 indices are correct. So it's either not copying the first 256 bytes (16bit int indices) or it's copying garbage into it.

I already talked to Rodrigo on Discord, and sent him both the Renderdoc capture and the save, I'm opening this regardless to keep track of the bug.

The moon rock is located here:
image
but you should be able to see it from the spawn location.

K0bin:
Closes #6764

The Vulkan spec says:
> bufferImageGranularity is the granularity, in bytes, at which buffer or linear image resources, and optimal image resources can be bound to adjacent offsets in the same VkDeviceMemory object without aliasing.

Yuzu did not respect this.
In the case of Mario Odyssey it caused image copies to write into the index buffer placed right after that image in host GPU memory. 

The value of this is 1024 on Nvidia GPUs. It does not appear to be a problem on AMD GPUs as the value for those is either 1 or 64 depending on the driver.

This probably impacts a bunch of other games on Nvidia GPUs.

### DIFF
--- a/src/video_core/vulkan_common/vulkan_memory_allocator.cpp
+++ b/src/video_core/vulkan_common/vulkan_memory_allocator.cpp
@@ -228,7 +228,9 @@ void MemoryCommit::Release() {
 
 MemoryAllocator::MemoryAllocator(const Device& device_, bool export_allocations_)
     : device{device_}, properties{device_.GetPhysical().GetMemoryProperties()},
-      export_allocations{export_allocations_} {}
+      export_allocations{export_allocations_},
+      buffer_image_granularity{
+          device_.GetPhysical().GetProperties().limits.bufferImageGranularity} {}
 
 MemoryAllocator::~MemoryAllocator() = default;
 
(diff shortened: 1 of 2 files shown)


FernandoS27 (approved):

Joshua-Ashton (approved):

FernandoS27:
I'll leave this testing for a few days and see if it breaks anything else. Afterwards, I'll merge it.

FernandoS27:
Seems good enough, thank you for your contribution!

merged

PR #6838, opened 2021-08-08
vic: Specify sws_scale height stride.

ameerj:
Silences a sws_scale runtime warning about unaligned strides.


src/video_core/command_classes/vic.cpp
```diff
@@ -96,12 +96,11 @@ void Vic::Execute() {
         if (!converted_frame_buffer) {
             converted_frame_buffer = AVMallocPtr{static_cast<u8*>(av_malloc(linear_size)), av_free};
         }
-
-        const int converted_stride{frame->width * 4};
+        const std::array converted_stride{frame->width * 4, frame->height * 4, 0, 0};
```
Morph1984, line 99: It might be useful to explicitly specify that this is an `int` array so cast errors can be identified

bunnei (approved):

PR #6841, opened 2021-08-08
Garbage Collection: enable as default, eliminate option.

FernandoS27:
Thanks to @K0bin the graphic corruption in SMO was fixed. Therefore, it's completely safe to have garbage collection always active.

This PRR eliminates the option and makes GC always be activated.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -59,7 +59,6 @@ void LogSettings() {
     log_setting("Renderer_UseVsync", values.use_vsync.GetValue());
     log_setting("Renderer_ShaderBackend", values.shader_backend.GetValue());
     log_setting("Renderer_UseAsynchronousShaders", values.use_asynchronous_shaders.GetValue());
-    log_setting("Renderer_UseGarbageCollection", values.use_caches_gc.GetValue());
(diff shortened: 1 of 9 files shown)


yzct12345:
I had some crashes a while ago while working on #6713. Link's Awakening crashed during the intro movie only when garbage collection was on. Are those fixed too?

FernandoS27:
> I had some crashes a while ago while working on #6713. Link's Awakening crashed during the intro movie only when garbage collection was on. Are those fixed too?

uhm that shouldn't happen. I was not aware of that one.

jacky400:
Less option is good . Wonder do we still need the option of use NVDEC emulation unless there are some games that broken wthout this option .

v1993:
> Less option is good .

Hard disagree (by me, anyways). Less options that you never actually need to change is good for sure.

> Wonder do we still need the option of "Use NVDEC emulation" unless there are some games that broken wthout this option .

There are still some games having issues with it enabled: #6525 (see comments).

jacky400:
I see. Just tested Mario Golf again and the in-game movie background playback is broken .

FernandoS27:
I've disabled this in the meantime. I'll investigate current issues

v1993:
This should be closed as option is gone with #6897 anyways.

Issue #6681, opened 2021-07-19
DELETE ME


IanHollow:

PacoAvelar:
Can you post the driver version?
 Task manager, gpu. Should hace the driver version

IanHollow:
sorry just edited it

RinMaru:
yea vulkan has issues at least its not this bad. I guess ill post this here since its generally VK issues

Windows 10
AMD RX 480 4GB Adrenalin 21.7.1
20GB DDR4 RAM
Ryzen 5 5600 6Cores 12 Threads

![Screenshot 2021-07-19 094331](https://user-images.githubusercontent.com/15071652/126169591-4f270de5-40ad-4ecb-84fc-b8d56631ccc0.png)

cam1170:
I don't know if this is the same issue, but I currently still have the weird eyes, as well as a display corruption of pink squares, even after quitting yuzu, the first time leaving Link's room. It goes away (mostly) when I turn my monitor off and on.
Windows 11
AMD RX 6600 XT 8 GB Adrenalin 23.4.3
16 GB DDR4 RAM
Ryzen 5600 6 Cores 12 Threads
Yuzu 1430

PR #6846, opened 2021-08-08
nvdec: Add GPU video decoding for all capable drivers and platforms

ameerj:
As a follow-up to #6713, this change introduces the option to use FFmpeg's GPU video decoding for more GPUs and platforms, with graceful software decoding fallback if the driver/codec combination is not supported. 

The matrix for the implemented GPU decoders is as follows:
| GPU    | Windows       | Linux       |
|--------|---------------|-------------|
| NVIDIA | NVDEC/D3D11VA | NVDEC/VDPAU |
| AMD    | D3D11VA       | VA-API/VDPAU       |
| INTEL  | D3D11VA       | VA-API      |

Note that the GPU selected for decoding will not necessarily be the GPU used by yuzu for rendering. 

On Windows, NVDEC is preferred over D3D11VA for supported hardware, as I experienced lower performance using D3D11VA due to it defaulting to using my integrated Intel graphics.

On Linux, users reported having better performance using VA-API with integrated graphics, so it is preferred over the discrete GPU decoder if supported.

Huge thanks to @lat9nq for Linux testing and CMake fixes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -518,6 +518,10 @@ set(FFmpeg_COMPONENTS
     avutil
     swscale)
 
+if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    Include(FindPkgConfig REQUIRED)
+    pkg_check_modules(LIBVA libva)
+endif()
 if (NOT YUZU_USE_BUNDLED_FFMPEG)
     # Use system installed FFmpeg
     find_package(FFmpeg QUIET COMPONENTS ${FFmpeg_COMPONENTS})
@@ -583,10 +587,6 @@ if (YUZU_USE_BUNDLED_FFMPEG)
             "${FFmpeg_PREFIX};${FFmpeg_BUILD_DIR}"
             CACHE PATH "Path to FFmpeg headers" FORCE)
 
-        if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
-            Include(FindPkgConfig REQUIRED)
-            pkg_check_modules(LIBVA libva)
-        endif()
         if(LIBVA_FOUND)
             pkg_check_modules(LIBDRM libdrm REQUIRED)
             find_package(X11 REQUIRED)
@@ -670,7 +670,7 @@ if (YUZU_USE_BUNDLED_FFMPEG)
         endif()
     else() # WIN32
         # Use yuzu FFmpeg binaries
-        set(FFmpeg_EXT_NAME "ffmpeg-4.3.1")
+        set(FFmpeg_EXT_NAME "ffmpeg-4.4")
         set(FFmpeg_PATH "${CMAKE_BINARY_DIR}/externals/${FFmpeg_EXT_NAME}")
         download_bundled_external("ffmpeg/" ${FFmpeg_EXT_NAME} "")
         set(FFmpeg_FOUND YES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -544,6 +544,9 @@ endif()
 
 if (YUZU_USE_BUNDLED_FFMPEG)
     if (NOT WIN32)
+        # TODO(lat9nq): Move this to externals/ffmpeg/CMakeLists.txt (and move externals/ffmpeg to
+        # externals/ffmpeg/ffmpeg)
+
         # Build FFmpeg from externals
         message(STATUS "Using FFmpeg from externals")
 
@@ -583,16 +586,23 @@ if (YUZU_USE_BUNDLED_FFMPEG)
                 CACHE PATH "Paths to FFmpeg libraries" FORCE)
         endforeach()
 
-        set(FFmpeg_INCLUDE_DIR
-            "${FFmpeg_PREFIX};${FFmpeg_BUILD_DIR}"
-            CACHE PATH "Path to FFmpeg headers" FORCE)
+        Include(FindPkgConfig REQUIRED)
+        pkg_check_modules(LIBVA libva)
+        pkg_check_modules(CUDA cuda)
+        pkg_check_modules(FFNVCODEC ffnvcodec)
+        pkg_check_modules(VDPAU vdpau)
+
+        set(FFmpeg_HWACCEL_LIBRARIES)
+        set(FFmpeg_HWACCEL_FLAGS)
+        set(FFmpeg_HWACCEL_INCLUDE_DIRS)
+        set(FFmpeg_HWACCEL_LDFLAGS)
 
         if(LIBVA_FOUND)
             pkg_check_modules(LIBDRM libdrm REQUIRED)
             find_package(X11 REQUIRED)
             pkg_check_modules(LIBVA-DRM libva-drm REQUIRED)
             pkg_check_modules(LIBVA-X11 libva-x11 REQUIRED)
-            set(FFmpeg_LIBVA_LIBRARIES
+            list(APPEND FFmpeg_HWACCEL_LIBRARIES
                 ${LIBDRM_LIBRARIES}
                 ${X11_LIBRARIES}
                 ${LIBVA-DRM_LIBRARIES}
@@ -602,9 +612,54 @@ if (YUZU_USE_BUNDLED_FFMPEG)
                 --enable-hwaccel=h264_vaapi
                 --enable-hwaccel=vp9_vaapi
                 --enable-libdrm)
+            list(APPEND FFmpeg_HWACCEL_INCLUDE_DIRS
+                ${LIBDRM_INCLUDE_DIRS}
+                ${X11_INCLUDE_DIRS}
+                ${LIBVA-DRM_INCLUDE_DIRS}
+                ${LIBVA-X11_INCLUDE_DIRS}
+                ${LIBVA_INCLUDE_DIRS}
+            )
             message(STATUS "VA-API found")
         else()
             set(FFmpeg_HWACCEL_FLAGS --disable-vaapi)
+        endif()
+
+        if (FFNVCODEC_FOUND AND CUDA_FOUND)
+            list(APPEND FFmpeg_HWACCEL_FLAGS
+                --enable-cuvid
+                --enable-ffnvcodec
+                --enable-nvdec
+                --enable-hwaccel=h264_nvdec
+                --enable-hwaccel=vp9_nvdec
+                --extra-cflags=-I${CUDA_INCLUDE_DIRS}
+            )
+            list(APPEND FFmpeg_HWACCEL_LIBRARIES
+                ${FFNVCODEC_LIBRARIES}
+                ${CUDA_LIBRARIES}
+            )
+            list(APPEND FFmpeg_HWACCEL_INCLUDE_DIRS
+                ${FFNVCODEC_INCLUDE_DIRS}
+                ${CUDA_INCLUDE_DIRS}
+            )
+            list(APPEND FFmpeg_HWACCEL_LDFLAGS
+                ${FFNVCODEC_LDFLAGS}
+                ${CUDA_LDFLAGS}
+            )
+            message(STATUS "ffnvcodec libraries version ${FFNVCODEC_VERSION} found")
+        endif()
+
+        if (VDPAU_FOUND)
+            list(APPEND FFmpeg_HWACCEL_FLAGS
+                --enable-vdpau
+                --enable-hwaccel=h264_vdpau
+                --enable-hwaccel=vp9_vdpau
+            )
+            list(APPEND FFmpeg_HWACCEL_LIBRARIES ${VDPAU_LIBRARIES})
+            list(APPEND FFmpeg_HWACCEL_INCLUDE_DIRS ${VDPAU_INCLUDE_DIRS})
+            list(APPEND FFmpeg_HWACCEL_LDFLAGS ${VDPAU_LDFLAGS})
+            message(STATUS "vdpau libraries version ${VDPAU_VERSION} found")
+        else()
+            list(APPEND FFmpeg_HWACCEL_FLAGS --disable-vdpau)
         endif()
 
         # `configure` parameters builds only exactly what yuzu needs from FFmpeg
@@ -624,7 +679,6 @@ if (YUZU_USE_BUNDLED_FFMPEG)
                     --disable-network
                     --disable-postproc
                     --disable-swresample
-                    --disable-vdpau
                     --enable-decoder=h264
                     --enable-decoder=vp9
                     --cc="${CMAKE_C_COMPILER}"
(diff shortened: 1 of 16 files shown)


src/common/settings.h
```diff
@@ -47,6 +47,12 @@ enum class FullscreenMode : u32 {
     Exclusive = 1,
 };
 
+enum class NvdecEmulation : u32 {
```
yzct12345, line 50: https://github.com/yuzu-emu/yuzu/pull/6713#discussion_r680098877

> I would prefer this to remain as a boolean toggle, and hardware acceleration to be used whenever supported by the host hardware without any user intervention.
The project prefers having things "just work" out of the box with as little tinkering as possible.
> That's exactly what I want to avoid completely. The code should be smart enough to choose the best GPU decoder with no user configuration/intervention, and fallback to CPU decoding if no compatible decoders are found.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -16,108 +17,190 @@ extern "C" {
...
+        case NvdecCommon::VideoCodec::Vp9:
+            return vp9_test;
+        default:
+            UNIMPLEMENTED_MSG("Unknown codec {}", current_codec);
+            return std::array<u8, 48>{};
+        }
+    }();
+    AVPacketPtr packet{av_packet_alloc(), AVPacketDeleter};
+    if (!packet) {
+        LOG_ERROR(Service_NVDRV, "av_packet_alloc failed");
```
yzct12345, line 151: We don't need error handling for av_packet_alloc. It will only fail on out of memory and that is already fatal for the rest of the emulator. We can just ASSERT.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -16,108 +17,190 @@ extern "C" {
...
+    avcodec_free_context(&av_codec_ctx);
+    av_buffer_unref(&av_gpu_decoder);
+}
+
+bool Codec::CreateGpuAvDevice() {
+#if defined(LIBVA_FOUND)
+    static constexpr std::array<const char*, 2> VAAPI_DRIVERS = {
+        "i915",
+        "iHD",
+    };
```
yzct12345, line 61: Why is `amdgpu` gone? I was using VA-API with an AMD gpu.

src/common/settings.h
ameerj: You got me. If you look at the commit history and the message for the commit that introduced this, you'll see that I needed to this change because some hardware configurations report as being capable of decoding, but caused visual artifacting and repeated frames. So this is a band-aid workaround until these issues can be ironed out.


https://user-images.githubusercontent.com/52414509/128651145-5555f848-6634-4e74-8120-db36bb872543.mp4

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -16,108 +17,190 @@ extern "C" {
...
         }
         LOG_DEBUG(Service_NVDRV, "VA-API av_hwdevice_ctx_create failed {}", hwdevice_error);
     }
     LOG_DEBUG(Service_NVDRV, "VA-API av_hwdevice_ctx_create failed for all drivers");
     av_dict_free(&hwdevice_options);
+#endif
+    static constexpr auto HW_CONFIG_METHOD = AV_CODEC_HW_CONFIG_METHOD_HW_DEVICE_CTX;
+    static constexpr std::array GPU_DECODER_TYPES{
+        AV_HWDEVICE_TYPE_CUDA,
+        AV_HWDEVICE_TYPE_D3D11VA,
```
yzct12345, line 82:
```suggestion
#ifdef _WIN32
        AV_HWDEVICE_TYPE_D3D11VA,
#endif
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -16,108 +17,190 @@ extern "C" {
...
             return true;
         }
         LOG_DEBUG(Service_NVDRV, "VA-API av_hwdevice_ctx_create failed {}", hwdevice_error);
     }
     LOG_DEBUG(Service_NVDRV, "VA-API av_hwdevice_ctx_create failed for all drivers");
     av_dict_free(&hwdevice_options);
+#endif
+    static constexpr auto HW_CONFIG_METHOD = AV_CODEC_HW_CONFIG_METHOD_HW_DEVICE_CTX;
+    static constexpr std::array GPU_DECODER_TYPES{
+        AV_HWDEVICE_TYPE_CUDA,
```
yzct12345, line 81: There should be `#ifdef` for all three of these.

src/video_core/command_classes/codecs/codec.cpp
ameerj: It seemed that vdpau was the better decoder for AMD. If this is not the case I can reintroduce it.

yzct12345: I thought vdpau was obsolete. It isn't really maintained anymore https://forums.developer.nvidia.com/t/whats-the-situation-with-vdpau-vaapi-nvdec/61031/5 . It's also created by Nvidia who is AMD's competitor. I doubt AMD would want to support it as much as VA-API which is somewhat more open.

I've also heard of broken VP9 support on vdpau.

- https://www.phoronix.com/forums/forum/linux-graphics-x-org-drivers/open-source-amd-linux/1023075-amdgpu-hardware-decoding-encoding-vdpau-or-vaapi
- https://github.com/yuzu-emu/yuzu/pull/6713#issuecomment-887817085

Unless there are actual users using vdpau where NVDEC and VA-API are broken, I would suggest just removing vdpau support.

bibitocarlos8: Please use Vaapi with Amd, vdpau has been deprecated for a long time and never works correctly, at least for me (raven ridge)

jacky400:
GPU decoding seems to be not working with MK8 using nvidia card.

ameerj:
Untagging as this is shown to regress a number of titles.

ameerj:
> MK8 that shows black screen 
@jacky400 Thanks for the report. I confirmed this was a bug within libavcodec and submitted a patch. No idea if/when it will get merged.

jacky400:
@ameerj thanks . By the way , i have intel+nvidia . How i can know if it is using the Nvidiai card to decode the video ?

ameerj:
> By the way , i have intel+nvidia . How can I know if it is using the Nvidiai card to decode the video instead of iGPU?

You'll need to monitor the GPU activity to see which GPU is active when video is being decoded. On Windows, this is easily visible in the Task Manager as it has a dedicated `Video Decode` monitor for most GPUs. Not so familiar with Linux tools that can show comparable info.

ameerj:
Retagging as the reported regressions seem to be fixed.

yzct12345:
> How can I know if it is using the Nvidiai card

You can search the log for `Using {} GPU decoder`.

# Windows

Sadly it'll use Nvidia by default.

# Linux

It'll try Intel then Nvidia. You seem to want it to use the Nvidia instead of the iGPU. That is not a good thing.

If you use Nvidia the video will go like: CPU ->(slow PCIe copy) Nvidia GPU ->(fast decode) Nvidia GPU ->(slow PCIe copy) CPU ->(fast memcpy) CPU ->(slow PCIe copy) Nvidia GPU.

If you use Intel it'll go like: CPU ->(fast memcpy) Intel iGPU ->(fast decode) Intel iGPU ->(fast memcpy) CPU ->(fast memcpy) CPU ->(slow PCIe copy) Nvidia GPU.

You don't want the data copied in and out of the CPU through slow PCI so you want it to use the iGPU first.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -16,108 +17,146 @@ extern "C" {
...
-    return false;
-}
-} // namespace
 #endif
-
-void AVFrameDeleter(AVFrame* ptr) {
-    av_frame_free(&ptr);
+    static constexpr auto HW_CONFIG_METHOD = AV_CODEC_HW_CONFIG_METHOD_HW_DEVICE_CTX;
+    static constexpr std::array GPU_DECODER_TYPES{
+        AV_HWDEVICE_TYPE_CUDA,
```
yzct12345, line 86: Is it safe to call it with `AV_HWDEVICE_TYPE_CUDA` without `#ifdef CUDA_FOUND`?

jacky400:
> You can search the log for `Using {} GPU decoder`.
 
Found it . Using cuda GPU decoder

> You don't want the data copied in and out of the CPU through slow PCI so you want it to use the iGPU first.

On windows , how i can set it to use iGPU then ?

yzct12345:
> cuda GPU

This means it wants to use Nvidia

> Could not find compatible GPU AV format, falling back to CPU

This means the GPU video decoding is not working for some reason and it reverted to what it's like before this PR.

> Can expose the Intel iGPU to select in a drop down list ?

Yuzu developers don't like adding settings.

> , how i can set it to use iGPU then

That's probably not going to happen. Programmers have decided it's not worth the effort. It'll take a few hours of coding and research for barely any gain.

src/video_core/command_classes/codecs/codec.cpp
ameerj: Absolutely. The gpu context creation will simply fail gracefully and be handled, proceeding to the next element in this array.

ameerj:
> You don't want the data copied in and out of the CPU through slow PCI so you want it to use the iGPU first.

@yzct12345 I would recommend you test on more hardware configurations. My testing, along with a few internal testers, has shown that Nvidia GPU decoding is on par with UHD 750 decoding (atm basically the best performing iGPU) in the worst case, with it generally being far more performant than iGPUs (with UHD630 being slower than CPU decoding, as I stated in the PR description). 

While in theory, it would seem that iGPUs should perform better than dGPUs for decoding. Reality is often not so cut and dry. The "slow" PCI memory transactions are likely negligible compared to the task of decoding, and a competent dGPU is better suited for decoding than internal graphics.

> On windows , how i can set it to use iGPU then

@jacky400 There's no control for specifying exactly which GPU is used for decoding for the Windows decoders. d3d11va doesn't seem to expose any method to specify a preferred GPU, and selects what it thinks is best from what I can tell.
Although I did not look too deeply into this as NVDEC capable GPUs will almost always outperform the iGPUs based on our testing. 
I'll investigate further why the 1060 is reporting as incompatible.

yzct12345:
> recommend you test on more hardware configurations

You're right. I measured 0 milliseconds on my Intel and 1 millisecond on my AMD dGPU and came up with the theory on the spot.

> generally being far more performant than iGPUs

I was assuming that both GPUs would decode exactly the same way. The rest of the GPU I know they can scale up and down the number of compute units. I have been thinking that they would use the same IP block in both iGPU and dGPU. That is unless they used an IP block with a lower profile and level.

> PCI

I would like to change my claim. As with #6857 I plan to explore the performance if I move all of `vic.cpp` to a Vulkan shader. Even without thinking about PCI, swscale with RGB frames is around 12 milliseconds while the NV12 conversion is 0 milliseconds so we need to explore the shader anyway.

behunin:
This depends on [nv-codec-headers](https://github.com/FFmpeg/nv-codec-headers) for Linux.  Add a check for that lib?

ameerj:
@jacky400 Another dev on the team tested his GTX 1060 and it was able to decode H264 and VP9 videos, with visual confirmation in task manager. Can you try on the latest build?
@behunin the CMake has an explicit check for `pkg_check_modules(FFNVCODEC ffnvcodec)`, which is the same library you linked.

behunin:
Ah I see, I think a note on the linux build wiki should be added. Just something to let you know that lib is required for Nvidia users to have hardware decoding.

jacky400:
Tested with latest commit witj my 1060 on windows 10

1. MK8D - Fall back to CPU decoding but renders ok
2. ZLA - Use CUDA decode ok and task manager shows 7% usage .

CMakeLists.txt
```diff
@@ -602,11 +612,57 @@ if (YUZU_USE_BUNDLED_FFMPEG)
...
+        if (FFNVCODEC_FOUND AND CUDA_FOUND)
+            list(APPEND FFmpeg_HWACCEL_FLAGS
+                --enable-cuvid
+                --enable-ffnvcodec
+                --enable-nvdec
+                --enable-cuda-llvm
+                --enable-hwaccel=h264_nvdec
+                --enable-hwaccel=vp9_nvdec
+                --extra-cflags=-I${CUDA_INCLUDE_DIRS}
+            )
```
behunin, line 636:
```suggestion
           list(APPEND FFmpeg_HWACCEL_FLAGS
                --enable-nonfree
                --enable-cuvid
                --enable-ffnvcodec
                --enable-nvdec
                --enable-cuda-nvcc
                --enable-hwaccel=h264_nvdec
                --enable-hwaccel=vp9_nvdec
                --extra-cflags=-I${CUDA_INCLUDE_DIRS}
                --nvccflags='-gencode arch=compute_52,code=sm_52 -O2'
            )
```

This is what I had to do to get a successful build on GCC 11.1 Ubuntu 20.10
I believe cuda_llvm is just clang builds.
the default nvcc flags are found [here.](https://github.com/FFmpeg/FFmpeg/blob/115f5e803502451ac64fa698730923299e5acc4e/configure#L4376)

[This explains the different gencodes.](https://arnon.dk/matching-sm-architectures-arch-and-gencode-for-various-nvidia-cards/)

jacky400:
@yzct12345 Just retested with latest commit and this is the result.

- MK8D - Use CUDA decode ok now and task manager shows 2-3% usage
- ZLA - Use CUDA decode ok and task manager shows 7% usage but FPS dropped sigificantly from previous 30-40 to 20-25

yzct12345:
> MK8D - Use CUDA decode ok now and task manager shows 2-3% usage

:tada:

> 30-40 to 20-25

Apply #6861 :smiley:

jacky400:
@yzct12345 sorry my fault . My PCI link width somehow runs x1 instead x16 . Reboot the OS and it back to good FPS.

CMakeLists.txt
lat9nq: We cannot set `--enable-nonfree` -- see https://www.ffmpeg.org/legal.html part 1.

We should just axe `--enable-cuda-*` since it doesn't seem to make a difference in yuzu on my end.

I don't understand why we need to specify Maxwell v2 when the FFmpeg's flags default to Kepler.

Maybe unrelated, or maybe it is related since it could affect your ability to build yuzu, but I recommend you use a supported operating system. See https://wiki.ubuntu.com/Releases

behunin: I was following [this guide](https://docs.nvidia.com/video-technologies/video-codec-sdk/ffmpeg-with-nvidia-gpu/)  and tried to compile without the nonfree flag but it complained that nvcc required it so that's sad.

The gencode I used is specific for my card (GTX 980) but the page I linked explains that the cuda toolkit 11 supports up to maxwell and that it will be deprecated in 12. The article also explains that it is a good idea to target the arch and gencode for the gpu it will run on for performance reasons.

I am actually on 21.04 but my brain is used to typing 20.10.... sorry about that

behunin:
```cmake
list(APPEND FFmpeg_HWACCEL_FLAGS
                --enable-cuvid
                --enable-ffnvcodec
                --enable-nvdec
                --enable-hwaccel=h264_nvdec
                --enable-hwaccel=vp9_nvdec
                --extra-cflags=-I${CUDA_INCLUDE_DIRS}
            )
```

Tested with success, what a relief.
Thanks for the heads up on the nonfree problem.

jacky400:
Btw , what is the rational behind that GPU decoing is slower than CPU decoding  for example ZLA intro movie.

v1993:
> Btw , what is the rational behind that GPU decoing is slower than CPU decoding for example ZLA intro movie ?

I'm not sure if it's right, but since #6861 is said to fix it it's probably that GPU provides frames in format that requires extra work on CPU. Optimizing conversions on CPU makes things as they should be.

yzct12345:
> GPU decoing is slower than CPU decoding

Ignore what I said about iGPU vs discrete GPU. I need to do more work about that.

> 6861

Let me explain this in simple terms.

1. There's a video decoding step. It converts H264 or VP9 to images. This step can happen either on the CPU or the GPU whether integrated or discrete.
2. There's a YUV -> RGB conversion that nobody has optimized yet so let's not talk about it.
3. There's a memory alignment step. The switch is designed to have a few zeroes between every row of pixels or something. It also puts two colors next to each other but the brightness in some other part of memory. We need to go over all the memory and convert to this format.

I started optimizing step 1 in #6713 but it only supports my setup. The current PR by ameerj completes support for all computers.

#6790 optimizes step 3 for YUV frames which are used basically by anything except ZLA and it's merged. The not-yet-merged #6681 just does the same thing for RGB frames which speeds up ZLA.

FernandoS27:
the fish(diggi) told me this is mergeable. If you guys suffer, well you know who to blame.

PR #6853, opened 2021-08-10
Update year and names of many .cpp and .h files

Moonlacer:
A simple comment edit to years and names of the copyrighted files


lioncash:
I appreciate the gusto for contributing, however, these are to establish creation dates. They don't need to be updated on an annual basis.

PR #6820, opened 2021-08-05
texture_cache: Split out template definitions

yzct12345:
It takes as much as three minutes to incrementally compile yuzu when I edit any of the template functions in `texture_cache.h`. It is compiling about 20 other unrelated files. This was already frustrating for me when writing #6790 and it gets worse debugging trickier problems. By splitting out the definitions the compile merely takes 30 seconds.

PR #6857, opened 2021-08-11
Vulkan Fastmem 3: No more loading stutters

yzct12345:
This moves UnswizzleTexture which can take as much as 20 milliseconds to the GPU. Just like CPU fastmem I use memory mapping to let the GPU see the emulated memory. The function now takes less than 1 millisecond.

Link's Awakening's intro video plays at 50 FPS maximum instead of 44 FPS. Kirby Star Allies doesn't stutter any more while loading such as when pressing A after starting emulation.

### DIFF
--- a/src/core/device_memory.cpp
+++ b/src/core/device_memory.cpp
@@ -2,11 +2,19 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include "common/logging/log.h"
 #include "core/device_memory.h"
 
 namespace Core {
 
-DeviceMemory::DeviceMemory() : buffer{DramMemoryMap::Size, 1ULL << 39} {}
+DeviceMemory::DeviceMemory() : buffer{DramMemoryMap::Size, 1ULL << 39} {
+    auto ptr = reinterpret_cast<std::size_t>(buffer.BackingBasePointer());
+    if (ptr & 0xfff || !ptr) {
+        LOG_CRITICAL(HW_Memory, "Unaligned DeviceMemory");
+        abort();
+    }
+}
+
 DeviceMemory::~DeviceMemory() = default;
 
 } // namespace Core
--- a/src/core/device_memory.h
+++ b/src/core/device_memory.h
@@ -12,7 +12,9 @@ namespace Core {
 namespace DramMemoryMap {
 enum : u64 {
     Base = 0x80000000ULL,
-    Size = 0x100000000ULL,
+    GiB = 0x40000000ULL,
+    GiBs = 4,
+    Size = GiB * GiBs,
     End = Base + Size,
     KernelReserveBase = Base + 0x60000,
     SlabHeapBase = KernelReserveBase + 0x85000,
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -65,7 +65,11 @@ struct Memory::Impl {
             return {};
         }
 
-        return system.DeviceMemory().GetPointer(paddr) + vaddr;
+        u8* test = system.DeviceMemory().GetPointer(paddr);
+        //        LOG_CRITICAL(Debug, "{:016X} {:016X} {:016X} va {:016X} {:016X}",
+        //                     (size_t)system.DeviceMemory().buffer.BackingBasePointer(),
+        //                     (size_t)test, paddr, vaddr, (size_t)(test + vaddr));
+        return test + vaddr;
     }
 
     u8 Read8(const VAddr addr) {
@@ -238,6 +242,42 @@ struct Memory::Impl {
 
     void ReadBlockUnsafe(const VAddr src_addr, void* dest_buffer, const std::size_t size) {
         ReadBlockImpl<true>(*system.CurrentProcess(), src_addr, dest_buffer, size);
+    }
+
+    void ReadBlockPointersUnsafe(const Kernel::KProcess& process, const VAddr src_addr,
+                                 ReadPointers& result, const std::size_t size) {
+        const auto end = &result.data[0];
+        const auto base_ptr = system.DeviceMemory().buffer.BackingBasePointer();
+        auto& tail = result.tail;
+        WalkBlock(
+            process, src_addr, size,
+            [src_addr, size, &tail](const std::size_t copy_amount, const VAddr current_vaddr) {
+                LOG_ERROR(
+                    HW_Memory,
+                    "Unmapped ReadBlockPointers @ 0x{:016X} (start address = 0x{:016X}, size = {})",
+                    current_vaddr, src_addr, size);
+                tail->backing_offset = DramMemoryMap::Size;
+            },
+            [&tail, base_ptr](const std::size_t copy_amount, const u8* const src_ptr) {
+                tail->backing_offset = src_ptr - base_ptr;
+            },
+            [&tail, base_ptr](const VAddr current_vaddr, const std::size_t copy_amount,
+                      const u8* const host_ptr) {
+                tail->backing_offset = host_ptr - base_ptr;
+            },
+            [&tail, end](const std::size_t copy_amount) {
+                tail->copy_amount = static_cast<u32>(copy_amount);
+                if (tail == end) {
+                    LOG_CRITICAL(Debug, "Trying to read too much???");
+                    abort();
+                }
+                --tail;
+            });
+    }
+
+    void ReadBlockPointersUnsafe(const VAddr src_addr, ReadPointers& result,
+                                 const std::size_t size) {
+        ReadBlockPointersUnsafe(*system.CurrentProcess(), src_addr, result, size);
     }
 
     template <bool UNSAFE>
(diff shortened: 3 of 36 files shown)


FernandoS27:
Please don't use direct GPU memory mapping just for this. It's best to just upload the memory to the GPU in a stagging buffer.

Direct GPU mapping will fuck up due to memory protections on guest memory. You need to do a sepparate CPU mirror on it that's not protected. You may also run out of handles on AMD and it will fuck up on Windows NVIDIA on sparse textures.

yzct12345:
> stagging buffer

This is the main bottleneck I identified in my profiling of the ZLA intro. There is a staging buffer and another temporary buffer and that is the problem. `ReadBlockUnsafe` takes 3 milliseconds and that's a whopping 20% of the frame time if we ever want to reach 60 FPS. After optimizing `UnswizzleTexture` both functions take nearly the same time. Moving both to the GPU is critical for performance.

If you are concerned the data will get modified in the meantime I can `vkCmdCopyBuffer` from direct GPU mapped memory to a staging buffer. `memcpy` must be avoided at all costs and that's what VK_EXT_external_memory_host allows us to do.

> due to memory protections on guest memory

Can you explain this more please? It would be nice to have an example of what could get broken.

> sepparate CPU mirror on it that's not protected

I forgot about the `RasterizerCachedMemory` case. Thanks for pointing that out. I'll think about how to handle it.

> run out of handles on AMD

This only adds 4 new buffers.

> will fuck up on Windows NVIDIA on sparse textures

I will have to learn about sparse textures first.

yzct12345:
> Please don't use direct GPU memory mapping just for this. It's best to just upload the memory to the GPU in a stagging buffer.

I finally got to looking at @ReinUsesLisp's code. Doesn't it use exactly the same extension VK_EXT_EXTERNAL_MEMORY_HOST?

# Credit

People on discord have accused me of [stealing Rodrigo's branch](https://discord.com/channels/398318088170242053/402002111241388032/874930526799167518). First of all the commits are completely different. The approach and targets are also different.

The only similarity is that we both use VK_EXT_EXTERNAL_MEMORY_HOST. Unfortunately I can't prove that I came up with that idea myself. Rodrigo's first commit is on [June 29](https://github.com/ReinUsesLisp/yuzu/commit/d951c716315fa1f048e261cc6d1ac89da0793774) and the last commit is on [July 13](https://github.com/ReinUsesLisp/yuzu/commit/d08be6412280a96e0b884448548579fbe23a3a35) and the push is on [2021-08-01T22:07:17Z](https://api.github.com/repos/reinuseslisp/yuzu/events). The Discord announcement is on the [7th](https://discord.com/channels/398318088170242053/402002111241388032/873571040381534278).

However there is a logical series of steps that led to mine. I located and fixed the problem with `UnswizzleTexture` called by `Image::UploadMemory` in #6790 at Aug 1, 2021, 12:05 AM UTC. This is 10 hours before Rodrigo's push. The later #6820 and #6823 on August 5 were split out of my work-in-progress fastmem 3 which continues work on the same problem. That's 2 days before the Discord announcement. While I can't prove that I didn't look at the push on August 1, it's unlikely I would know of it without Discord. I worked to get fastmem 3 working until publishing on August 11. Upon checking Discord I [informed Rodrigo of this](https://discord.com/channels/398318088170242053/402002111241388032/874103396548890644).

Furthermore we are optimizing different part. Rodrigo is targeting the [buffer cache](https://discord.com/channels/398318088170242053/402002111241388032/873572508819922985). As you know from NVDEC I'm targeting the texture cache and hoping it also improves in-game performance. If not, then at least get rid of lag spikes.

We are also trying to do the opposite thing. Rodrigo brings "MASSIVE Accuracy advantages at a cost of a lil perf" while I don't care about accuracy that much and you know from my PRs how much I care about performance.

Blink said gpu fastmem 2 only works well on ["NVIDIA- Linux. For Windows, we are waiting for an nvidia fix"](https://discord.com/channels/398318088170242053/402002111241388032/873572268364681267). Mine only works on Intel Linux. Though I haven't tried I don't see why mine wouldn't work on Nvidia Windows.

I cannot claim I invented using VK_EXT_EXTERNAL_MEMORY_HOST for emulators. Another emulator added that [a year ago](https://www.libretro.com/index.php/parallel-rdp-rewritten-from-scratch-available-in-parallel-n64-right-now-for-retroarch/). But Rodrigo's and my fastmem work differently. I just map all the physical memory and have whatever dispatches the compute shaders use `ReadPointers` to quickly do the lookup there. You don't see `GPUVAddr` and `VAddr` in my mapping code. Rodrigo maps the virtual memory and the method is `void HostMemoryAllocator::Map(GPUVAddr gpu_addr, VAddr cpu_addr, size_t size_bytes, bool cpu_cached, bool is_remap)`. As shown by the `is_remap` variable Rodrigo's memory can get remapped at runtime. I would never do that:

http://lkml.iu.edu/hypermail/linux/kernel/0004.0/0728.html

> People love mmap() and other ways to play with the page tables to
> optimize away a copy operation, and sometimes it is worth it.
> 
> HOWEVER, playing games with the virtual memory mapping is very expensive
> in itself. It has a number of quite real disadvantages that people tend
> to ignore because memory copying is seen as something very slow, and
> sometimes optimizing that copy away is seen as an obvious improvment.
> 
> Downsides to mmap:
>  - quite noticeable setup and teardown costs. And I mean _noticeable_.
>    It's things like following the page tables to unmap everything
>    cleanly. It's the book-keeping for maintaining a list of all the
>    mappings. It's The TLB flush needed after unmapping stuff.
>  - page faulting is expensive. That's how the mapping gets populated,
>    and it's quite slow.
> 
> Upsides of mmap: 
> - if the data gets re-used over and over again (not relevant for emulators at least for my UploadTexture)
> 
> - if the data is large (not relevant for emulators)
> 
> Linus  (Torvalds)

I would like to claim I invented using this for yuzu:

- Mapping physical instead of virtual memory
- Mapping the memory just once at startup

# Summary

In summary they use the same VK_EXT_EXTERNAL_MEMORY_HOST but for different purposes.

|What|Rodrigo|mine|
|-|-|-|
|Target|Buffer cache|Texture cache|
|Improves|Accuracy|Performance|
|Worsens|Performance|Accuracy?|
|Driver|Nvidia-Linux|Intel-Linux|
|mapped when?|During gameplay?|Only four times at startup|
|Memory|Virtual|Physical|

I don't plan to touch the buffer cache unless profiling leads to it or if dealing with `RasterizerCachedMemory`requires it, so there shouldn't be a code conflict. What I'm worried about [is this](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_external_memory_host.html#_issues). Some compat might be needed between fastmem 2 and fastmem 3:

> 6) Can the same host allocation be imported multiple times into a given physical device?
> 
> RESOLVED: No, at least not guaranteed by this extension. Some platforms do not allow locking the same physical pages for device access multiple times, so attempting to do it may result in undefined behavior.

v1993:
This is a great technical description, but

> People on discord have accused me of [stealing Rodrigo's branch](https://discord.com/channels/398318088170242053/402002111241388032/874930526799167518).

is just plain wrong. They have literally just asked a question if it's building on previous work or something entirely new (and even if you were building on previous work, it would hardly be stealing anyways).

Morph1984:
It's more of people being confused of what this "fastmem" is, they assume its the same as Rodrigo's, but its not. Perhaps the description could've earlier stated that it isn't the same as users aren't familiar with how this fastmem differs.

FernandoS27:
I wouldn't call this stealing. If someone calls you that, let me know, they'll have to deal with my anger. This just does the same idea of direct host memory import of gpu memory but for a very specific task. Maybe you want to change the name, it really gave people a bad impression of what this is as it's not what we normally call "fastmem" or what that thing was supposed to do.

Personally I'm against using GPU memory mapping just to reduce the need to swizzle/unswizzle. I do believe that this could actually be part of a full GPU memory mapping based GPU emulation (such as Rodrigo's). just on the premise that GPU memory mapping is heavily unstable right now and very recent tech. If the texture is within 2 windows memory placeholders, nvidia fucks up. If the memory is protected (as usual with caching) then the gpu fucks up if it writes it (all vendors). Intel and AMD easily run out of memory handles.

I think you should add this to Rodrigo's fastmem and continue working on that in a future. It's a very experimental technology but I do believe it will yield nice fruits in the future. It's obviously counterproductive for Discrete GPUs as they must constantly cache/uncache this memory from host memory but it will be important in the future for integrated SoCs and unfied memory. It also increases accuracy by a thresshold.

Another thing is that this optimization sacrifices accuracy and that's not acceptable.

Moonlacer:
If people are claiming that you've stolen Rodrigo's code, they have no idea what they're talking about :) @yzct12345 
Keep up the good work 👍

v1993:
> If people are claiming that you've stolen Rodrigo's code, they have no idea what they're talking about :) @yzct12345
> Keep up the good work

We yet have to see any such claims (or, at least, quotations of them here). The original attribution of them seems like a simple case of misunderstanding - happens to the best of us.

zeltermann:
Seems interesting, unfortunately no updates after 2 years.

PR #6860, opened 2021-08-11
settings: Fix MSVC issues

lat9nq:
According to https://stackoverflow.com/questions/469508, we run into a MSVC bug (since VS 2005) when using diamond inheritance for RangedSetting.

This explicitly implements those functions in RangedSetting. GetValue is implemented as just calling the inherited version. The explicit converson operator is reimplemented. I opted for this over ignoring the warning with a pragma since this specifies the inherited behavior, and I have now less faith in MSVC to pick the right one.

In addition, we mark destructors as virtual to silence what I believe is a fair MSVC compilation error.


src/common/settings.h
```diff
@@ -318,7 +318,22 @@ class RangedSetting final : public BasicRangedSetting<Type>, public Setting<Type
         : BasicSetting<Type>{default_val, name},
           BasicRangedSetting<Type>{default_val, min_val, max_val, name}, Setting<Type>{default_val,
                                                                                        name} {}
-    ~RangedSetting() = default;
+    virtual ~RangedSetting() = default;
+
+    // The following are needed to avoid a MSVC bug
+    // (source: https://stackoverflow.com/questions/469508)
+    [[nodiscard]] virtual const Type& GetValue() const {
```
lioncash, line 325: shouldn't these implementations be using `override` instead of `virtual`?

PR #6861, opened 2021-08-12
decoders: Optimize memcpy for the other functions

yzct12345:
Applies the speed doubling in #6790 to the rest of the functions in that file.

Thanks to @ameerj for telling me that the problem is here.

ZLA plays at 60 FPS instead of 50 FPS!


bunnei (approved):

bunnei:
Awesome change, thank you!

PR #6862, opened 2021-08-12
input_common: Disable sdl raw input mode

german77:
SDL dies when a web applet opens on yuzu by unknown reasons if JOYSTICK_RAWINPUT is enabled. We don't need this feature.


yzct12345:
> RAWINPUT

Will removing this cause unwanted mouse acceleration?

german77:
We never used rawinput, SDL enabled this feature by default on `2.0.16`.  Mouse input is handled by Qt.

bunnei (approved):

PR #6863, opened 2021-08-12
Fix LAN Play

### LINKED ISSUE #6807: bsd: blocking sockets don't seem to be working correctly
Trying to start the LAN mode in games like Mario Kart 8 Deluxe seems to correctly set up 3 UDP sockets. However after creating a room only one call to RecvFrom is logged. I'm not entirely sure why that is the case, but it seems that the service thread is blocked due to the game not requesting a non-blocking socket. But if the whole service is blocking the game isn't able to send broadcast messages announcing the room (Maybe the game starts multiple threads each controlling a blocking socket?). When I used a real switch connected to the same network I noticed it sent some broadcast packets. These packets seemed to stop the bsd service from blocking and the log showed that the game called RecvFrom again and even did two calls to SendTo. The real switch seems to have stopped the thread block. With this knowledge I wrote a small [script](https://gist.github.com/spholz/9480c72249dd6c93b58aedf80ab5da97) in python (note: this script is for linux only. I don't know if and how you can get the size of the receive queue on Windows) that constantly checks if one of the ports used by the game has an empty receive queue and fills it up with zeroes so that the game can continue handling other events and even announce the room correctly. This script indeed made Mario Kart 8 Deluxe running inside yuzu able to communicate with a real switch. After applying a small patch to IGeneralService::GetCurrentIpConfigInfo that hardcodes the local IP address of my PC I was even able to start a race. This patch combined with the script made the LAN mode of Mario Tennis Aces work, too. However, the Splatoon 2 LAN mode hangs when creating or joining a room.

spholz:
This PR fixes the LAN Play mode found in some games (e.g. Mario Kart 8 Deluxe).
Although this PR mostly consists of small bug fixes to make LAN Play work, a new "Network Interface" combobox is added in the configuration window (located in the "Network" tab).

Tested games:
- Mario Kart 8 Deluxe
- Mario Tennis Aces
- Splatoon 2 (needs further patches in IAudioIn)

This PR was tested both on Windows and on Linux.

Fixes #6807


src/core/hle/service/nifm/nifm.cpp
```diff
@@ -322,12 +322,15 @@ class IGeneralService final : public ServiceFramework<IGeneralService> {
     void GetCurrentIpAddress(Kernel::HLERequestContext& ctx) {
         LOG_WARNING(Service_NIFM, "(STUBBED) called");
 
-        const auto [ipv4, error] = Network::GetHostIPv4Address();
-        UNIMPLEMENTED_IF(error != Network::Errno::SUCCESS);
+        auto ipv4 = Network::GetHostIPv4Address();
+        if (!ipv4) {
+            LOG_CRITICAL(Service_NIFM, "Couldn't get host IPv4 address, defaulting to 0.0.0.0");
```
lioncash, line 327:
```suggestion
            LOG_ERROR(Service_NIFM, "Couldn't get host IPv4 address, defaulting to 0.0.0.0");
```

since this doesn't threaten stability of the program

src/core/network/network.cpp
```diff
@@ -353,27 +360,27 @@ NetworkInstance::~NetworkInstance() {
...
-    if (ent->h_length != sizeof(in_addr)) {
-        UNIMPLEMENTED_MSG("Unexpected size={} in hostent->h_length", ent->h_length);
-    }
+    const auto res = std::ranges::find_if(network_interfaces,
+                                          [&selected_network_interface](const auto& interface) {
+                                              return interface.name == selected_network_interface;
+                                          });
+
+    if (res != network_interfaces.end()) {
+        char ip_addr[16];
```
lioncash, line 375:
```suggestion
        char ip_addr[16] = {};
```

src/core/hle/service/nifm/nifm.cpp
```diff
@@ -322,12 +322,15 @@ class IGeneralService final : public ServiceFramework<IGeneralService> {
...
+            ipv4.emplace(Network::IPv4Address{0, 0, 0, 0});
+        }
 
         IPC::ResponseBuilder rb{ctx, 3};
         rb.Push(ResultSuccess);
-        rb.PushRaw(ipv4);
+        rb.PushRaw(ipv4.value());
```
lioncash, line 333:
```suggestion
        rb.PushRaw(*ipv4);
```

src/core/hle/service/nifm/nifm.cpp
```diff
@@ -322,12 +322,15 @@ class IGeneralService final : public ServiceFramework<IGeneralService> {
     void GetCurrentIpAddress(Kernel::HLERequestContext& ctx) {
         LOG_WARNING(Service_NIFM, "(STUBBED) called");
 
-        const auto [ipv4, error] = Network::GetHostIPv4Address();
-        UNIMPLEMENTED_IF(error != Network::Errno::SUCCESS);
+        auto ipv4 = Network::GetHostIPv4Address();
```
lioncash, line 325:
```suggestion
        const auto ipv4 = Network::GetHostIPv4Address();
```

src/core/hle/service/nifm/nifm.cpp
```diff
@@ -354,10 +357,16 @@ class IGeneralService final : public ServiceFramework<IGeneralService> {
         static_assert(sizeof(IpConfigInfo) == sizeof(IpAddressSetting) + sizeof(DnsSetting),
                       "IpConfigInfo has incorrect size.");
 
+        auto ipv4 = Network::GetHostIPv4Address();
```
lioncash, line 360:
```suggestion
        const auto ipv4 = Network::GetHostIPv4Address();
```

src/core/hle/service/nifm/nifm.cpp
```diff
@@ -354,10 +357,16 @@ class IGeneralService final : public ServiceFramework<IGeneralService> {
...
+        if (!ipv4) {
+            LOG_CRITICAL(Service_NIFM, "Couldn't get host IPv4 address, defaulting to 0.0.0.0");
```
lioncash, line 362:
```suggestion
            LOG_ERROR(Service_NIFM, "Couldn't get host IPv4 address, defaulting to 0.0.0.0");
```

src/core/hle/service/nifm/nifm.cpp
```diff
@@ -354,10 +357,16 @@ class IGeneralService final : public ServiceFramework<IGeneralService> {
...
+            ipv4.emplace(Network::IPv4Address{0, 0, 0, 0});
+        }
+
         const IpConfigInfo ip_config_info{
             .ip_address_setting{
                 .is_automatic{true},
-                .current_address{192, 168, 1, 100},
+                .current_address{ipv4.value()},
```
lioncash, line 369:
```suggestion
                .current_address{*ipv4},
```

src/core/network/network_interface.cpp
```diff
@@ -0,0 +1,111 @@
...
+    for (auto ifa = ifaddr; ifa != nullptr; ifa = ifa->ifa_next) {
+        if (ifa->ifa_addr == nullptr) {
+            continue;
+        }
+
+        if (ifa->ifa_addr->sa_family != AF_INET) {
+            continue;
+        }
+
+        if (!(ifa->ifa_flags & IFF_UP) || ifa->ifa_flags & IFF_LOOPBACK) {
```
lioncash, line 95: Prefer making arithmetic to bool conversions explicit.

```suggestion
        if ((ifa->ifa_flags & IFF_UP) == 0 || (ifa->ifa_flags & IFF_LOOPBACK) != 0) {
```

src/core/network/network_interface.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    // retry up to 5 times
+    for (int i = 0; i < 5 && ret == ERROR_BUFFER_OVERFLOW; i++) {
+        ret = GetAdaptersAddresses(AF_INET, GAA_FLAG_SKIP_MULTICAST | GAA_FLAG_SKIP_DNS_SERVER,
+                                   nullptr, adapter_addresses, &buf_size);
+
+        if (ret == ERROR_BUFFER_OVERFLOW) {
+            adapter_addresses_raw.resize(buf_size);
+            adapter_addresses =
+                reinterpret_cast<PIP_ADAPTER_ADDRESSES>(adapter_addresses_raw.data());
```
lioncash, line 40: This is undefined behavior to dereference. Consider making the contents of the vector contain `IP_ADAPTER_ADDRESSES` instances to begin with instead of type punning from u8.

src/core/network/network_interface.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <ifaddrs.h>
+#include <net/if.h>
+#endif
+
+namespace Network {
+
+#ifdef _WIN32
+
+std::vector<NetworkInterface> GetAvailableNetworkInterfaces() {
+    std::vector<NetworkInterface> result;
```
lioncash, line 25: This can be moved down into the `if (ret == NO_ERROR)` case down below to narrow its scope. The error case can do `return {};`

src/yuzu/configuration/configure_network.cpp
```diff
@@ -47,29 +48,42 @@ ConfigureService::ConfigureService(QWidget* parent)
     ui->bcat_source->addItem(QStringLiteral("Boxcat"), QStringLiteral("boxcat"));
 #endif
 
+    ui->network_interface->addItem(QStringLiteral("None"));
```
lioncash, line 51:
```suggestion
    ui->network_interface->addItem(tr("None"));
```

Since this is a user-facing string.

v1993:
Fixes #6807, I presume? Now only if I had friends to test this with...

src/core/hle/service/nifm/nifm.cpp
spholz: This cannot be const because if ipv4 does not contain a value it is set to 0.0.0.0. However, std::optional<T>::value_or could be used here

src/core/network/network_interface.cpp
spholz: so do you suggest doing `adapter_addresses_raw.resize((buf_size / sizeof(IP_ADAPTER_ADRESSES)) + 1);`? to get the right size?

spholz:
Your CI doesn't seem to like std::bit_cast (gcc only supports it starting from version 11, clang doesn't support it at all). Should I use your implementation in common/bit_cast.h (I noticed only now that yuzu has its own implementation)?

german77:
Yes. We are using clang 10

src/yuzu/configuration/configure_network.h
```diff
@@ -8,16 +8,18 @@
 #include <QFutureWatcher>
 #include <QWidget>
 
+#include "core/network/network_interface.h"
```
lioncash, line 11: Missed this in the initial review, but this should be moved into the cpp file, since nothing in the header depends on this.

src/core/network/network_interface.cpp
lioncash: Yup. Anything that doesn't require the `reinterpret_cast`s.

src/core/network/network.cpp
```diff
@@ -353,27 +360,27 @@ NetworkInstance::~NetworkInstance() {
...
-std::pair<IPv4Address, Errno> GetHostIPv4Address() {
-    std::array<char, 256> name{};
-    if (gethostname(name.data(), static_cast<int>(name.size()) - 1) == SOCKET_ERROR) {
-        return {IPv4Address{}, GetAndLogLastError()};
-    }
+std::optional<IPv4Address> GetHostIPv4Address() {
+    const std::string& selected_network_interface = Settings::values.network_interface.GetValue();
+    const auto network_interfaces = Network::GetAvailableNetworkInterfaces();
+    ASSERT_MSG(network_interfaces.size() > 0,
+               "GetAvailableNetworkInterfaces returned no interfaces");
```
yzct12345, line 367: Please don't assert here. Assert should only be used for things that should logically never happen.

I run third-party software such as any yuzu build except my own in a network sandbox on Linux. This ensures the software sees 0 network interfaces. That will make me hit this assert.

```suggestion
    if (network_interfaces.size() == 0) {
        LOG_ERROR(Network, "GetAvailableNetworkInterfaces returned no interfaces");
        return {};
    }
```

src/core/network/network_interface.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#ifdef _WIN32
+
+std::vector<NetworkInterface> GetAvailableNetworkInterfaces() {
+    std::vector<u8> adapter_addresses_raw;
+    auto adapter_addresses = reinterpret_cast<PIP_ADAPTER_ADDRESSES>(adapter_addresses_raw.data());
+    DWORD ret = ERROR_BUFFER_OVERFLOW;
+    DWORD buf_size = 0;
+
+    // retry up to 5 times
```
yzct12345, line 31: Should you wait between the retries?

vonchenplus:
So what's the root case that you couldn't connect in the Mario Kart 8 Deluxe before?

german77:
I was able to play MK8 with no issues. Good job.

src/core/network/network_interface.cpp
FernandoS27: I gave him this code and nope, he does not have to. It's just how (stupid) the windows API works. There's no function to get the size, you try with a size and if it's wrong, it changes it for you and you have to retry.

spholz: Sorry, totally forgot to mention that you gave me this code.

src/core/hle/service/nifm/nifm.cpp
spholz: This also cannot be const

spholz:
There is no single root case that prevented you from creating LAN Play games. All of the bug fixes are needed (and the new Network Interface UI).

somewhereintherainbow:
Amazing feature. I tried Mario Kart 8 Deluxe but only allows max 2 players. Everytime we tried to add the third player game keeps on "Please wait" Then Disconnected from server.

FernandoS27:
> Amazing feature. I tried Mario Kart 8 Deluxe but only allows max 2 players. Everytime we tried to add the third player game keeps on "Please wait" Then Disconnected from server. Used ZeroTier One. We tried with Europe + South america = Same error. We tried all three players on South America = Same error.

Have in mind LAN mode is designed for very short distances (literally within a building). It's normal that most games may have way lower timeout/latency tolerance. The best is that you try different games.

spholz:
Yeah it seems like LAN Play games have very low tolerances. I even noticed a LAN game getting out of sync. My switch ran without any stutters, of course. But even the slightest lag on yuzu caused the games to have a very noticeable transfer delay.

somewhereintherainbow:
> 
> 
> > Amazing feature. I tried Mario Kart 8 Deluxe but only allows max 2 players. Everytime we tried to add the third player game keeps on "Please wait" Then Disconnected from server. Used ZeroTier One. We tried with Europe + South america = Same error. We tried all three players on South America = Same error.
> 
> Have in mind LAN mode is designed for very short distances (literally within a building). It's normal that most games may have way lower timeout/latency tolerance. The best is that you try different games and for long distances play, you may have to wait for LDN to be implermented,

Hey! I played Mario Tennis Aces with a guy from Chile, South America. It is working fine. It has like the same delay as Smash Bros Ultimate has on real switch.

FernandoS27:
> > > Amazing feature. I tried Mario Kart 8 Deluxe but only allows max 2 players. Everytime we tried to add the third player game keeps on "Please wait" Then Disconnected from server. Used ZeroTier One. We tried with Europe + South america = Same error. We tried all three players on South America = Same error.
> > 
> > 
> > Have in mind LAN mode is designed for very short distances (literally within a building). It's normal that most games may have way lower timeout/latency tolerance. The best is that you try different games and for long distances play, you may have to wait for LDN to be implermented,
> 
> Hey! I played Mario Tennis Aces with a guy from Chile, South America. It is working fine. It has like the same delay as Smash Bros Ultimate has on real switch.

depends on the game but expect most to have low tolerance.

FernandoS27 (approved):

FernandoS27:
thank you for your contribution, merged.

PR #6868, opened 2021-08-13
threadsafe_queue: Fix deadlock

yzct12345:
This fixes a lost wakeup in SPSCQueue. If the reader is in just the right position, the writer's notification will be lost and this will be a problem if the writer then does something to wait on the reader.

This was discovered to affect my upcoming stacktrace PR. I don't think any performance decrease will be noticeable because an uncontended mutex is smart enough to skip the syscall. This PR might also resolve some rare deadlocks but I don't know of any examples.

### DIFF
--- a/src/common/threadsafe_queue.h
+++ b/src/common/threadsafe_queue.h
@@ -46,15 +46,13 @@ public:
         ElementPtr* new_ptr = new ElementPtr();
         write_ptr->next.store(new_ptr, std::memory_order_release);
         write_ptr = new_ptr;
+        ++size;
 
-        const size_t previous_size{size++};
-
-        // Acquire the mutex and then immediately release it as a fence.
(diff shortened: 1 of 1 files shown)


FernandoS27:
Tbh due to bad use(ahem @bunnei ) the SPSCQueue is no longer SPSC (Single Producer Single Consumer) nor lockfree.

Keeping it is useless, best is to replace its instances for std::deque or srd::queue.

yzct12345:
> Keeping it is useless, best is to replace

This PR is supposed to be a hotfix.

Cleaning up such an important component like a queue and such a hard thing like concurrency is going to take at least a day or two.

FernandoS27:
it's cool, I'll accept it.

bunnei (approved):

PR #6869, opened 2021-08-13
logging: Simplify and make thread-safe

yzct12345:
This simplifies the logging system.

This also fixes some lost messages on startup.

The simplification is simple. I removed unused functions and moved most things in the `.h` to the `.cpp`. I replaced the unnecessary linked list with its contents laid out as three member variables. Anything that went through the linked list now directly accesses the backends. Generic functions are replaced with those for each specific use case and there aren't many. This change increases coupling but we gain back more KISS and encapsulation.

With those changes it was easy to make it thread-safe. I just removed the mutex and turned a boolean atomic. I was planning to use this thread-safety in my next PR about stacktraces. It was actually async-signal-safety at first but I ended up using a different approach. Anyway getting rid of the linked list is important for that because have the list of backends constantly changing complicates things.

### DIFF
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -2,13 +2,9 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
-#include <algorithm>
 #include <atomic>
 #include <chrono>
 #include <climits>
-#include <condition_variable>
-#include <memory>
-#include <mutex>
 #include <thread>
 #include <vector>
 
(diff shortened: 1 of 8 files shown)


src/core/core.h
```diff
@@ -396,7 +401,7 @@ class System {
     struct Impl;
     std::unique_ptr<Impl> impl;
 
-    static System s_instance;
+    inline static std::unique_ptr<System> s_instance{};
```
lioncash, line 404: This doesn't seem like a related change to the logging system.

src/common/logging/backend.cpp
```diff
@@ -2,118 +2,237 @@
...
+        static_cast<void>(FS::RenameFile(filename, old_filename));
+
+        file = std::make_unique<FS::IOFile>(filename, FS::FileAccessMode::Write,
+                                            FS::FileType::TextFile);
+    }
+
+    ~FileBackend() override = default;
+
+    void Write(const Entry& entry) override {
+        if (!~bytes_written) {
```
lioncash, line 100:
```suggestion
        if (~bytes_written == 0) {
```

We prefer explicit comparisons with integral values.

src/common/logging/backend.cpp
```diff
@@ -2,118 +2,237 @@
...
+
+        using namespace Common::Literals;
+        // Prevent logs from exceeding a set maximum size in the event that log entries are spammed.
+        const auto write_limit = Settings::values.extended_logging ? 1_GiB : 100_MiB;
+        const bool write_limit_exceeded = bytes_written > write_limit;
+        if (entry.log_level >= Level::Error || write_limit_exceeded) {
+            if (write_limit_exceeded) {
+                // Stop writing after the write limit is exceeded.
+                // Don't close the file so we can print a stacktrace if necessary
+                bytes_written = ~0U;
```
lioncash, line 114:
```suggestion
                bytes_written = UINT32_MAX;
```

or if it was meant to affect the whole value, this should be SIZE_MAX

src/common/logging/backend.cpp
```diff
@@ -135,100 +254,51 @@ class Impl {
...
+    void ForEachBackend(auto lambda) {
+        lambda(static_cast<Backend&>(debugger_backend));
+        lambda(static_cast<Backend&>(color_console_backend));
+        lambda(static_cast<Backend&>(file_backend));
+    }
 
-void FileBackend::Write(const Entry& entry) {
-    if (!file->IsOpen()) {
-        return;
+    static void Deleter(Impl* ptr) {
```
lioncash, line 263: This doesn't really do anything that the default unique_ptr deleter wouldn't do anyway.

lioncash (commented):
Consider placing the PR comment into the main logging commit message. This is info that should be stored within the repository itself. `logging: Simplify` doesn't add any context or explain anything about the changes by itself.

src/common/logging/backend.cpp
yzct12345: The default unique_ptr deleter can't see the private destructorp

src/core/core.h
yzct12345: If this change is undone yuzu will crash on launch because I'm now enforcing that we don't lose log messages on startup. Just delete the `Common::Log::InitializeLogging()` in `main.cpp` and run with gdb to see where we're losing messages.

System's constructor ends up indirectly logging some messages. Before this PR the logging silently fails. Now I delay and make the System initialization explicit instead of surprisingly happening before `main`. This rescues the early log messages.

src/common/logging/backend.cpp
yzct12345: `~0U` was a trick that I thought would optimize well. It just uses an unlikely value to store a boolean in the same variable. Since you don't think this is readable I might as well make it an actual boolean.

lioncash: Yeah, there's no real need to be 'clever' like this.

lioncash:
LGTM once the commit message is amended.

lioncash:
The commit message needs to include the description in the PR comment (the part that explains what the simplification actually is).

ameerj:
Looks like changes made in this PR might be ignoring the configured log filter at boot.

Morph1984:
Untagging as this causes homebrew to softlock

src/core/core.h
Morph1984: I think its fine if we lose messages on startup for the time being, because we are going to change how the system instance is constructed after #6774 is merged. I'd suggest to keep the changes small in scope to avoid potential weird behavior

german77:
Current EA still have the same issues with homebrew. Adding To EA again since this PR is not the cause of that issue

bunnei (approved):

bunnei:
LGTM, thank you!

PR #6878, opened 2021-08-15
kernel: Optimize GetHostThreadID

BreadFish64:
I profiled the *Fire Emblem: Three Houses* loading screen and found that `GetHostThreadID` was taking up a fair chunk of CPU time. It seems like `thread_local` initialization and the optional parameter for ID allocation hurt the chances of inlining and eliminating some checks in each path.
Here's a profile of the old implementation:
image
And this implementation:
image

### DIFF
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -267,20 +267,23 @@ struct KernelCore::Impl {
         }
     }
 
-    /// Creates a new host thread ID, should only be called by GetHostThreadId
-    u32 AllocateHostThreadId(std::optional<std::size_t> core_id) {
-        if (core_id) {
-            // The first for slots are reserved for CPU core threads
-            ASSERT(*core_id < Core::Hardware::NUM_CPU_CORES);
-            return static_cast<u32>(*core_id);
-        } else {
-            return next_host_thread_id++;
+    static inline thread_local u32 host_thread_id = UINT32_MAX;
+
+    /// Gets the host thread ID for the caller, allocating a new one if this is the first time
+    u32 GetHostThreadId(std::size_t core_id) {
+        if (host_thread_id == UINT32_MAX) {
+            // The first four slots are reserved for CPU core threads
+            ASSERT(core_id < Core::Hardware::NUM_CPU_CORES);
+            host_thread_id = static_cast<u32>(core_id);
         }
+        return host_thread_id;
     }
 
     /// Gets the host thread ID for the caller, allocating a new one if this is the first time
-    u32 GetHostThreadId(std::optional<std::size_t> core_id = std::nullopt) {
-        const thread_local auto host_thread_id{AllocateHostThreadId(core_id)};
+    u32 GetHostThreadId() {
+        if (host_thread_id == UINT32_MAX) {
+            host_thread_id = next_host_thread_id++;
+        }
         return host_thread_id;
     }
 


src/core/hle/kernel/kernel.cpp
```diff
@@ -267,20 +267,23 @@ struct KernelCore::Impl {
         }
     }
 
-    /// Creates a new host thread ID, should only be called by GetHostThreadId
-    u32 AllocateHostThreadId(std::optional<std::size_t> core_id) {
-        if (core_id) {
+    static inline thread_local u32 host_thread_id = static_cast<u32>(-1);
```
lioncash, line 270:
```suggestion
    static inline thread_local u32 host_thread_id = UINT32_MAX
```

src/core/hle/kernel/kernel.cpp
```diff
@@ -267,20 +267,23 @@ struct KernelCore::Impl {
...
     }
 
-    /// Creates a new host thread ID, should only be called by GetHostThreadId
-    u32 AllocateHostThreadId(std::optional<std::size_t> core_id) {
-        if (core_id) {
+    static inline thread_local u32 host_thread_id = UINT32_MAX;
+
+    /// Gets the host thread ID for the caller, allocating a new one if this is the first time
+    u32 GetHostThreadId(std::size_t core_id) {
+        if (static_cast<s32>(host_thread_id) < 0) {
```
v1993, line 274: Any specific reason why this isn't a simple equality check?

src/core/hle/kernel/kernel.cpp
```diff
@@ -267,20 +267,23 @@ struct KernelCore::Impl {
...
+            host_thread_id = static_cast<u32>(core_id);
         }
+        return host_thread_id;
     }
 
     /// Gets the host thread ID for the caller, allocating a new one if this is the first time
-    u32 GetHostThreadId(std::optional<std::size_t> core_id = std::nullopt) {
-        const thread_local auto host_thread_id{AllocateHostThreadId(core_id)};
+    u32 GetHostThreadId() {
+        if (static_cast<s32>(host_thread_id) < 0) {
```
v1993, line 284: Ditto

src/core/hle/kernel/kernel.cpp
```diff
@@ -267,20 +267,23 @@ struct KernelCore::Impl {
...
 
-    /// Creates a new host thread ID, should only be called by GetHostThreadId
-    u32 AllocateHostThreadId(std::optional<std::size_t> core_id) {
-        if (core_id) {
+    static inline thread_local u32 host_thread_id = UINT32_MAX;
+
+    /// Gets the host thread ID for the caller, allocating a new one if this is the first time
+    u32 GetHostThreadId(std::size_t core_id) {
+        if (static_cast<s32>(host_thread_id) < 0) {
             // The first for slots are reserved for CPU core threads
```
v1993, line 275: While we're at it, probably also fix typo

```suggestion
            // First four slots are reserved for CPU core threads
```

src/core/hle/kernel/kernel.cpp
BreadFish64: The intent was to save a few bytes. This is just `test eax, eax` while `cmp eax, 0xFFFFFFFF` needs extra code size for the literal. I double checked it and it looks `-1` is compacted to a single byte so I'll change it for readability's sake. 👍

ameerj (approved):

PR #6886, opened 2021-08-19
applet_error: Fix 64-bit error code conversion

Morph1984:
Thanks to @v1993 for pointing this out!

### DIFF
--- a/src/core/hle/service/am/applets/applet_error.cpp
+++ b/src/core/hle/service/am/applets/applet_error.cpp
@@ -16,6 +16,30 @@
 
 namespace Service::AM::Applets {
 
+struct ErrorCode {
+    u32 error_category{};
+    u32 error_number{};
+
+    static constexpr ErrorCode FromU64(u64 error_code) {
+        return {
+            .error_category{static_cast<u32>(error_code >> 32)},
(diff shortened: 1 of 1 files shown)


src/core/hle/service/am/applets/applet_error.cpp
```diff
@@ -16,6 +16,30 @@
 
 namespace Service::AM::Applets {
 
+struct ErrorCode {
+    u32 error_category{};
+    u32 error_number{};
+
+    static constexpr ErrorCode FromU64(u64 error_code) {
+        return ErrorCode{
```
lioncash, line 24:
```suggestion
        return {
```

src/core/hle/service/am/applets/applet_error.cpp
```diff
@@ -16,6 +16,30 @@
...
+            .error_category{error_code >> 32},
+            .error_number{error_code & 0xFFFFFFFF},
+        };
+    }
+
+    static constexpr ErrorCode FromResultCode(ResultCode result) {
+        return ErrorCode{
```
lioncash, line 31:
```suggestion
        return {
```

lioncash (approved):

PR #6890, opened 2021-08-19
Replace QPoint with QPointF where applicable

v1993:
Previously, floats were implicitly cast to integers. If this turns out to break things, just code them as integers instead.

As advised by @german77, opening as draft first to test.


v1993:
Confirmed to be fine in its current form.

lioncash (approved):

PR #6900, opened 2021-08-21
structured_control_flow: Add DemoteCombinationPass

### LINKED ISSUE #6636: Pokémon Sword - Graphical Issue
I don't know why it is happening. I've already changed the settings many times and had no results. it happens just with NPCs and the player.

image

image

image

image

image

image

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/6811935/yuzu_log.txt)

[yuzu_log.txt.old.txt](https://github.com/yuzu-emu/yuzu/files/6811934/yuzu_log.txt.old.txt)

ameerj:
This change adds a new pass to reorder Demote instructions to the end of the program. This change came about after noticing that some drivers don't behave nicely when the following usage pattern was encountered:

```glsl
bool cond = x > y;
float attr_val = in_attr0.x;
if (cond) {
    demote;
}
// proceed to use attr_val
```
Querying attribute values near a `demote` / `discard` seemed to cause some drivers to misread the input attributes. 

By deferring demotion to the end of the program, this issue was resolved.
This fixes the "wireframe" bug found in `Pokemon Sword and Shield`.

Before |  After
pkmn_bug | pkmn_fix
image | image

closes #6636

### DIFF
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -325,6 +325,8 @@ PipelineCache::PipelineCache(RasterizerVulkan& rasterizer_, Tegra::Engines::Maxw
     host_info = Shader::HostTranslateInfo{
         .support_float16 = device.IsFloat16Supported(),
         .support_int64 = device.IsShaderInt64Supported(),
+        .needs_demote_reorder = driver_id == VK_DRIVER_ID_AMD_PROPRIETARY_KHR ||
+                                driver_id == VK_DRIVER_ID_AMD_OPEN_SOURCE_KHR,
     };
 }
 
--- a/src/shader_recompiler/frontend/maxwell/structured_control_flow.cpp
+++ b/src/shader_recompiler/frontend/maxwell/structured_control_flow.cpp
@@ -20,6 +20,7 @@
 #include "shader_recompiler/frontend/maxwell/decode.h"
 #include "shader_recompiler/frontend/maxwell/structured_control_flow.h"
 #include "shader_recompiler/frontend/maxwell/translate/translate.h"
+#include "shader_recompiler/host_translate_info.h"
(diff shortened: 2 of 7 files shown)


goldenx86:
"Some" drivers, I wonder which ones.
FineWine my butt.

v1993:
Is there any (meaningful) performance penalty on less broken drivers with this change?

goldenx86:
From my understanding, it should be _very_ small.

ameerj:
@v1993 the intent is to enable this pass only when needed on the drivers that suffer from the bug before merging. but having all users participate in testing this in EA will uncover bugs quicker than if it were limited to a specific driver.

goldenx86:
bestgirl

Seems to still affect best girl, I'm trying to find a save to easily reproduce.

goldenx86:
[0100ABF008968000.zip](https://github.com/yuzu-emu/yuzu/files/7026295/0100ABF008968000.zip)
Save provided by gidoly.

ameerj:
Marnie's face should be fixed with latest commits. Had to rethink the workaround.

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -325,6 +325,7 @@ PipelineCache::PipelineCache(RasterizerVulkan& rasterizer_, Tegra::Engines::Maxw
     host_info = Shader::HostTranslateInfo{
         .support_float16 = device.IsFloat16Supported(),
         .support_int64 = device.IsShaderInt64Supported(),
+        .needs_demote_reorder = driver_id == VK_DRIVER_ID_AMD_PROPRIETARY_KHR,
```
FernandoS27, line 328: you should also add AMD Open Source Driver (Not RADV, the other one). Since it mostly shares isues with the propietary driver.

bunnei (approved):

PR #6950, opened 2021-09-02
input_common: Add advanced setting for 8 player support

german77:
I feel like experimenting today.  

Add a setting to enable 8 player support with xinput controllers since this was disabled because input stops working if the web applet opens. This setting will effectively disable the web applets.

Replace SDL config strings with the equivalent.

Log important settings that might affect gameplay

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -72,6 +72,9 @@ void LogSettings() {
     log_setting("Debugging_ProgramArgs", values.program_args.GetValue());
     log_setting("Services_BCATBackend", values.bcat_backend.GetValue());
     log_setting("Services_BCATBoxcatLocal", values.bcat_boxcat_local.GetValue());
+    log_setting("Input_EnableMotion", values.motion_enabled.GetValue());
+    log_setting("Input_EnableVibration", values.vibration_enabled.GetValue());
+    log_setting("Input_EnableRawInput", values.enable_raw_input.GetValue());
 }
 
 bool IsConfiguringGlobal() {
--- a/src/yuzu/configuration/configure_input_advanced.ui
+++ b/src/yuzu/configuration/configure_input_advanced.ui
@@ -2672,6 +2672,22 @@
                  </property>
                 </widget>
                </item>
+               <item row="9" column="0">
+                 <widget class="QCheckBox" name="enable_raw_input">
+                   <property name="toolTip">
+                     <string>Requires restarting yuzu</string>
+                   </property>
+                   <property name="minimumSize">
+                     <size>
+                       <width>0</width>
+                       <height>23</height>
+                     </size>
+                   </property>
+                   <property name="text">
+                     <string>Enable XInput 8 player support (disables web applet)</string>
+                   </property>
+                 </widget>
+               </item>
               </layout>
              </widget>
             </item>
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -16,7 +16,6 @@
 
 #include "common/common_types.h"
 #include "common/settings_input.h"
-#include "input_common/udp/client.h"
 
 namespace Settings {
 
@@ -498,14 +497,15 @@ struct Values {
 
     Setting<bool> use_docked_mode{true, "use_docked_mode"};
 
+    BasicSetting<bool> enable_raw_input{false, "enable_raw_input"};
+
     Setting<bool> vibration_enabled{true, "vibration_enabled"};
     Setting<bool> enable_accurate_vibrations{false, "enable_accurate_vibrations"};
 
     Setting<bool> motion_enabled{true, "motion_enabled"};
     BasicSetting<std::string> motion_device{"engine:motion_emu,update_period:100,sensitivity:0.01",
                                             "motion_device"};
-    BasicSetting<std::string> udp_input_servers{InputCommon::CemuhookUDP::DEFAULT_SRV,
-                                                "udp_input_servers"};
+    BasicSetting<std::string> udp_input_servers{"127.0.0.1:26760", "udp_input_servers"};
 
     BasicSetting<bool> mouse_panning{false, "mouse_panning"};
     BasicRangedSetting<u8> mouse_panning_sensitivity{10, 1, 100, "mouse_panning_sensitivity"};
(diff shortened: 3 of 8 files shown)


src/yuzu/configuration/configure_input_advanced.ui
```diff
@@ -2672,6 +2672,19 @@
...
+               <item row="9" column="0">
+                 <widget class="QCheckBox" name="enable_raw_input">
+                   <property name="minimumSize">
+                     <size>
+                       <width>0</width>
+                       <height>23</height>
+                     </size>
+                   </property>
+                   <property name="text">
+                     <string>Enable xinput for 8 player support (Disables web applet)</string>
```
lioncash, line 2684:
```suggestion
                     <string>Enable 8 player support with XInput (disables web applet)</string>
```

src/yuzu/configuration/configure_input_advanced.ui
```diff
@@ -2672,6 +2672,19 @@
...
+               <item row="9" column="0">
+                 <widget class="QCheckBox" name="enable_raw_input">
+                   <property name="minimumSize">
+                     <size>
+                       <width>0</width>
+                       <height>23</height>
+                     </size>
+                   </property>
+                   <property name="text">
+                     <string>Enable Xinput for 8 player support (disables web applet)</string>
```
lioncash, line 2684:
```suggestion
                     <string>Enable Xinput 8 player support (disables web applet)</string>
```

unuruu5760:
this is really cool ,Thank you for your effort.

K900:
The Steam Controller driver isn't enabled in SDL by default, it's only on for mobile targets.

K900:
To clarify, when I say "not enabled", I mean "not compiled in at all", and you'll actually have to patch SDL to make it build the Steam Controller driver on desktop targets. It also doesn't support rumble or motion at all, so there's no benefit to using it over Steam Input or a third party mapper. Also, the last time I tried to build SDL with it, at least on Linux, it just segfaulted at launch.

german77:
I see motion data here https://github.com/libsdl-org/SDL/blob/main/src/joystick/hidapi/SDL_hidapi_steam.c#L851
Since SDL is great at not explaining features I will just wait until a couple of persons confirms that it's indeed not working, we don't loose anything.

K900:
It is read, but not reported - other drivers call `SDL_PrivateJoystickSensor` with the motion data, and this one doesn't. I've actually filed this as https://github.com/libsdl-org/SDL/issues/4577 . Also, it still crashes.

german77:
If that's the case I will simply remove it from supported controllers. With no motion there is no real motive to try to make it work

K900:
Also as far as I can see, yuzu is using upstream SDL, which means the driver won't even get compiled in anyway.

german77:
I can fix the driver but I don't own the controller yet XD

K900:
There's comments like these in the source: https://github.com/libsdl-org/SDL/blob/main/src/joystick/hidapi/SDL_hidapi_steam.c#L1037

So I'm not sure if a better driver is something they're interested in? At least that's what held me back from more experimenting with it.

src/common/settings.cpp
```diff
@@ -72,6 +72,9 @@ void LogSettings() {
     log_setting("Debugging_ProgramArgs", values.program_args.GetValue());
     log_setting("Services_BCATBackend", values.bcat_backend.GetValue());
     log_setting("Services_BCATBoxcatLocal", values.bcat_boxcat_local.GetValue());
+    log_setting("Input_EnableMotion", values.motion_enabled.GetValue());
+    log_setting("Input_EnableVibrations", values.vibration_enabled.GetValue());
```
Morph1984, line 76:
```suggestion
    log_setting("Input_EnableVibration", values.vibration_enabled.GetValue());
```

src/yuzu/configuration/configure_input_advanced.ui
```diff
@@ -2672,6 +2672,22 @@
...
+                     <string>Requires restarting yuzu</string>
+                   </property>
+                   <property name="minimumSize">
+                     <size>
+                       <width>0</width>
+                       <height>23</height>
+                     </size>
+                   </property>
+                   <property name="text">
+                     <string>Enable Xinput 8 player support (disables web applet)</string>
```
Morph1984, line 2687:
```suggestion
                     <string>Enable XInput 8 player support (disables web applet)</string>
```

Morph1984 (approved):
Some very minor nits

v1993:
I think you should hide this option on Linux since it does nothing there (and xinput is a rather unrelated thing on Linux).

bunnei (approved):

austinarchibald:
How do you get more than 4 controllers to work? I have enabled the advanced feature in yuzu, but cannot get it to work. Windows will only recognize 4 controllers. I have tried all sorts of combinations, all xinput, all dinput, 4 xinput + additional dinput, wireless, wired, wired/wireless combo, etc. Nothing works. I read elsewhere you may need a special dll? Or use xboxce? Would love any suggestions please, thank you!

tsquillario:
I'm seeing duplicate entries in the Input Device dropdown when connecting more than 5 controllers. 

See https://github.com/yuzu-emu/yuzu/issues/13022

austinarchibald:
> I'm seeing duplicate entries in the Input Device dropdown when connecting more than 5 controllers.
> 
> See #13022

Same here.

TrevorSquillario:
> How do you get more than 4 controllers to work? I have enabled the advanced feature in yuzu, but cannot get it to work. Windows will only recognize 4 controllers. I have tried all sorts of combinations, all xinput, all dinput, 4 xinput + additional dinput, wireless, wired, wired/wireless combo, etc. Nothing works. I read elsewhere you may need a special dll? Or use xboxce? Must I install openxinput? If so, how? (the github page is confusing to me). Would love any suggestions please, thank you! @german77

I could only get 4+ controllers to work using an Xbox Wireless Adapter. It uses 2.4Ghz wireless. I could never get Bluetooth working past 2.

mirh:
[Allegedly](https://github.com/atar-axis/xpadneo/issues/34) that's a limitation of the windows drivers.

german77:
This is not the place for discussion as it's unrelated to the PR

PR #6953, opened 2021-09-02
renderer_vulkan: Wait on present semaphore at queue submit

ameerj:
The present semaphore is being signalled by the call to acquire the
swapchain image. This semaphore is meant to be waited on when rendering
to the swapchain image. Currently it is waited on when presenting, which 
happens after rendering, but moving its usage to be waited on in the command
buffer submission allows for proper usage of this semaphore.

Fixes the device lost when launching titles on the Intel Linux Mesa driver.

Supersedes #6687

### DIFF
--- a/src/video_core/renderer_vulkan/renderer_vulkan.cpp
+++ b/src/video_core/renderer_vulkan/renderer_vulkan.cpp
@@ -164,7 +164,8 @@ void RendererVulkan::SwapBuffers(const Tegra::FramebufferConfig* framebuffer) {
         blit_screen.Recreate();
     }
     const VkSemaphore render_semaphore = blit_screen.DrawToSwapchain(*framebuffer, use_accelerated);
-    scheduler.Flush(render_semaphore);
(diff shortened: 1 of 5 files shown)


PacoAvelar:
@ameerj 
MHR keeps crashing at boot with same 'Device loss' error on log. 
OpenGL boots the game fine.


Is there someway I can provide more information to debug ?

ameerj:
> 
> 
> MHR keeps crashing at boot with same 'Device loss' error on log.

@PacoAvelar a device loss can happen for many different (and hard to track) reasons. If you can confirm this happens on the MHR demo as well I can investigate it.

PacoAvelar:
@ameerj 
I can confirm that this issue happens also on the MHR demo.

But here is a list of other games that also encounter that issue, not actually at boot but at loading screen (in-game) 

    - Astral Chain
    - Destroy All Humans 
    
    
 Will continue updating this list

PR #6962, opened 2021-09-03
renderer_vulkan: Spirv support glsl  legacy attribute

vonchenplus:
This PR aims to support the gl_Color and gl_TexCoord attributes of glsl in spirv.

Solved similar DQ3 black screen issue in vulkan mode.

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_context.cpp
@@ -6,7 +6,6 @@
 #include <array>
 #include <climits>
 #include <string_view>
-#include <queue>
 
 #include <fmt/format.h>
 
@@ -430,6 +429,16 @@ Id DescType(EmitContext& ctx, Id sampled_type, Id pointer_type, u32 count) {
 }
 
 constexpr size_t NUM_FIXEDFNCTEXTURE = 10;
+
+size_t FindFistUnUsedLocation(const auto& used_location) {
+    if (used_location.all()) {
+        throw RuntimeError("Unable to get an unused location");
+    }
+    size_t location = 0;
+    while (location < used_location.size() && used_location.test(location))
+        ++location;
+    return location;
+}
 } // Anonymous namespace
 
 void VectorTypes::Define(Sirit::Module& sirit_ctx, Id base_type, std::string_view name) {
@@ -1230,15 +1239,19 @@ void EmitContext::DefineInputs(const IR::Program& program) {
         loads[IR::Attribute::TessellationEvaluationPointV]) {
         tess_coord = DefineInput(*this, F32[3], false, spv::BuiltIn::TessCoord);
     }
-    std::queue<u32> ununsed_location;
+    std::bitset<IR::NUM_GENERICS> used_location;
     for (size_t index = 0; index < IR::NUM_GENERICS; ++index) {
         const AttributeType input_type{runtime_info.generic_input_types[index]};
-        if (!runtime_info.previous_stage_stores.Generic(index) || !loads.Generic(index) ||
-            input_type == AttributeType::Disabled) {
-            ununsed_location.push(static_cast<u32>(index));
+        if (!runtime_info.previous_stage_stores.Generic(index)) {
             continue;
         }
-
+        if (!loads.Generic(index)) {
+            continue;
+        }
+        if (input_type == AttributeType::Disabled) {
+            continue;
+        }
+        used_location.set(index);
         const Id type{GetAttributeType(*this, input_type)};
         const Id id{DefineInput(*this, type, true)};
         Decorate(id, spv::Decoration::Location, static_cast<u32>(index));
@@ -1265,22 +1278,16 @@ void EmitContext::DefineInputs(const IR::Program& program) {
         }
     }
     if (loads.AnyComponent(IR::Attribute::ColorFrontDiffuseR)) {
-        if (ununsed_location.empty()) {
-            throw RuntimeError("Unable to get an unused location");
-        }
-        u32 location = ununsed_location.front();
-        ununsed_location.pop();
+        size_t location = FindFistUnUsedLocation(used_location);
+        used_location.set(location);
         const Id id{DefineInput(*this, F32[4], true)};
         Decorate(id, spv::Decoration::Location, location);
         input_front_color = id;
     }
     for (size_t index = 0; index < NUM_FIXEDFNCTEXTURE; ++index) {
         if (loads.AnyComponent(IR::Attribute::FixedFncTexture0S + index * 4)) {
-            if (ununsed_location.empty()) {
-                throw RuntimeError("Unable to get an unused location");
-            }
-            u32 location = ununsed_location.front();
-            ununsed_location.pop();
+            size_t location = FindFistUnUsedLocation(used_location);
+            used_location.set(location);
             const Id id{DefineInput(*this, F32[4], true)};
             Decorate(id, spv::Decoration::Location, location);
             input_fixed_fnc_textures[index] = id;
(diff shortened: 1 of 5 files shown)


src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -85,6 +91,14 @@ std::optional<OutAttr> OutputAttrPointer(EmitContext& ctx, IR::Attribute attr) {
         const Id element_id{ctx.Const(element)};
         return OutputAccessChain(ctx, ctx.output_f32, ctx.output_position, element_id);
     }
+    case IR::Attribute::ColorFrontDiffuseR:
+    case IR::Attribute::ColorFrontDiffuseG:
+    case IR::Attribute::ColorFrontDiffuseB:
+    case IR::Attribute::ColorFrontDiffuseA: {
+        const u32 element{static_cast<u32>(attr) % 4};
```
v1993, line 98: How about using `TxtCoordAttributeElement` here?

vonchenplus: My naming scheme here is the same as IR::Attribute::PositionX

v1993: I'm not talking about naming here. It just looks like `TxtCoordAttributeElement` exists specifically to do this conversion (it's the same code for sure).

vonchenplus: Oh, yes, I misunderstood. Actually calculating AttributeElement is the same in several places, and I can't quite pinpoint this thing, so for now I'll probably leave this piece alone for now.

src/shader_recompiler/backend/spirv/emit_context.cpp
```diff
@@ -1227,17 +1228,15 @@ void EmitContext::DefineInputs(const IR::Program& program) {
         loads[IR::Attribute::TessellationEvaluationPointV]) {
         tess_coord = DefineInput(*this, F32[3], false, spv::BuiltIn::TessCoord);
     }
+    std::queue<u32> ununsed_location;
```
ameerj, line 1231: Rather than use a std::queue, I'd prefer the use of a std::bitset to represent the availability of all generics location.

src/shader_recompiler/backend/spirv/emit_context.cpp
```diff
@@ -1227,17 +1228,15 @@ void EmitContext::DefineInputs(const IR::Program& program) {
...
-        if (!loads.Generic(index)) {
-            continue;
-        }
-        if (input_type == AttributeType::Disabled) {
+        if (!runtime_info.previous_stage_stores.Generic(index) || !loads.Generic(index) ||
+            input_type == AttributeType::Disabled) {
+            ununsed_location.push(static_cast<u32>(index));
             continue;
         }
+
```
ameerj, line 1239: Unneeded newline

src/shader_recompiler/backend/spirv/emit_context.cpp
```diff
@@ -1227,17 +1228,15 @@ void EmitContext::DefineInputs(const IR::Program& program) {
...
         const AttributeType input_type{runtime_info.generic_input_types[index]};
-        if (!runtime_info.previous_stage_stores.Generic(index)) {
-            continue;
-        }
-        if (!loads.Generic(index)) {
-            continue;
-        }
-        if (input_type == AttributeType::Disabled) {
+        if (!runtime_info.previous_stage_stores.Generic(index) || !loads.Generic(index) ||
+            input_type == AttributeType::Disabled) {
```
ameerj, line 1235: I'd revert this change and keep the conditions separated, seeing as the conditions are not entirely related.

src/shader_recompiler/backend/spirv/emit_context.cpp
```diff
@@ -1313,9 +1334,34 @@ void EmitContext::DefineOutputs(const IR::Program& program) {
         viewport_mask = DefineOutput(*this, TypeArray(U32[1], Const(1u)), std::nullopt,
                                      spv::BuiltIn::ViewportMaskNV);
     }
+    std::queue<u32> ununsed_location;
```
ameerj, line 1337: ditto about the bitset

src/shader_recompiler/frontend/ir/attribute.h
```diff
@@ -222,8 +222,15 @@ enum class Attribute : u64 {
     FrontFace = 255,
 };
 
+constexpr size_t NUM_FIXEDFNCTEXTURE = 10;
```
ameerj, line 225: Since this is currently only used in the `spirv/emit_context.cpp`, it would be more appropriate to declare it in the anonymous namespace of that file

src/shader_recompiler/frontend/ir/attribute.h
```diff
@@ -222,8 +222,15 @@ enum class Attribute : u64 {
...
 constexpr size_t NUM_GENERICS = 32;
 
+[[nodiscard]] bool IsFixedFncTexture(Attribute attribute);
+
+[[nodiscard]] u32 FixedFncTextureAttributeIndex(Attribute attribute);
+
+[[nodiscard]] u32 FixedFncTextureAttributeElement(Attribute attribute);
+
```
ameerj, line 233: These should be moved to the anonymous namespace of `spirv/emit_spirv_context_get_set.cpp`

ameerj (commented):
Few nits

src/shader_recompiler/backend/spirv/emit_context.cpp
```diff
@@ -427,6 +427,18 @@ Id DescType(EmitContext& ctx, Id sampled_type, Id pointer_type, u32 count) {
         return pointer_type;
     }
 }
+
+constexpr size_t NUM_FIXEDFNCTEXTURE = 10;
+
+size_t FindFistUnUsedLocation(const auto& used_location) {
+    if (used_location.all()) {
```
ameerj, line 434: Since this case is unlikely, and may need to iterate the entire bitset to evaluate, we can save ourselves from looping the bitset twice by refactoring this function to be a for loop through all indices, returning early if a free index is found, otherwise throwing at the end.

src/shader_recompiler/backend/spirv/emit_context.cpp
```diff
@@ -427,6 +427,18 @@ Id DescType(EmitContext& ctx, Id sampled_type, Id pointer_type, u32 count) {
...
+        throw RuntimeError("Unable to get an unused location");
```
ameerj, line 435:
```suggestion
        throw RuntimeError("Unable to get an unused location for legacy attribute");
```

src/shader_recompiler/backend/spirv/emit_context.cpp
```diff
@@ -1227,6 +1239,7 @@ void EmitContext::DefineInputs(const IR::Program& program) {
         loads[IR::Attribute::TessellationEvaluationPointV]) {
         tess_coord = DefineInput(*this, F32[3], false, spv::BuiltIn::TessCoord);
     }
+    std::bitset<IR::NUM_GENERICS> used_location;
```
ameerj, line 1242:
```suggestion
    std::bitset<IR::NUM_GENERICS> used_locations{};
```

src/shader_recompiler/backend/spirv/emit_context.cpp
```diff
@@ -1263,6 +1277,22 @@ void EmitContext::DefineInputs(const IR::Program& program) {
...
+    if (loads.AnyComponent(IR::Attribute::ColorFrontDiffuseR)) {
+        size_t location = FindFistUnUsedLocation(used_location);
+        used_location.set(location);
+        const Id id{DefineInput(*this, F32[4], true)};
+        Decorate(id, spv::Decoration::Location, location);
+        input_front_color = id;
+    }
+    for (size_t index = 0; index < NUM_FIXEDFNCTEXTURE; ++index) {
+        if (loads.AnyComponent(IR::Attribute::FixedFncTexture0S + index * 4)) {
+            size_t location = FindFistUnUsedLocation(used_location);
```
ameerj, line 1289: I'd suggest keeping a running index of the available position in the bitset, and pass that in as an argument to this function to find the next available position, rather than repeating the work of looking for the position from index 0 everytime.

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -43,6 +43,25 @@ Id AttrPointer(EmitContext& ctx, Id pointer_type, Id vertex, Id base, Args&&...
...
 }
 
+bool IsFixedFncTexture(IR::Attribute attribute) {
+    return attribute >= IR::Attribute::FixedFncTexture0S &&
+           attribute <= IR::Attribute::FixedFncTexture9Q;
+}
+
+u32 FixedFncTextureAttributeIndex(IR::Attribute attribute) {
+    if (!IsFixedFncTexture(attribute)) {
+        throw InvalidArgument("Attribute is not fixedfnctexture {}", attribute);
```
ameerj, line 53:
```suggestion
        throw InvalidArgument("Attribute {} is not a FixedFncTexture", attribute);
```

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -43,6 +43,25 @@ Id AttrPointer(EmitContext& ctx, Id pointer_type, Id vertex, Id base, Args&&...
...
+    }
+    return (static_cast<u32>(attribute) - static_cast<u32>(IR::Attribute::FixedFncTexture0S)) / 4u;
+}
+
+u32 FixedFncTextureAttributeElement(IR::Attribute attribute) {
+    if (!IsFixedFncTexture(attribute)) {
+        throw InvalidArgument("Attribute is not fixedfnctexture {}", attribute);
+    }
+    return static_cast<u32>(attribute) % 4;
```
ameerj, line 62:
```suggestion
    return static_cast<u32>(attribute) % 4u;
```

src/shader_recompiler/backend/spirv/emit_context.cpp
vonchenplus: The reason for this is that the use of broken locations is not continuous, for example, only 0 and 2 were used before, and the location 1 was left vacant

ameerj: This won't be an issue. In your example, the first call to the function will check location 0, find it is not available, check 1, return that as the index of an available location. The next time the function is called, there is no reason to check indices 0 and 1 again as they have already been checked and determined to be used, so you begin searching from index 2.

vonchenplus: Well, I guess it's not a performance bottleneck here, doing so leads to an increase in code complexity?

ameerj (commented):
Last few nits

bunnei (approved):

bunnei:
LGTM, thank you!

PR #6964, opened 2021-09-03
Revert PR #6922

bunnei:
Reverts a revert (PR #6922) which reintroduces #6832. This PR was reverted while I tracked down a crash in Pokémon Sword/Shield. This no longer appears to be happening, so I am reintroducing these changes. Since the original PR was approved, I will likely quickly merge this once others have tested and confirmed that the issue is gone.


bunnei:
This has since been fixed and merged with #7462

PR #6979, opened 2021-09-07
input_common: Add alternative string for joycons

german77:
Linux have another name for joycons. We simply add a check for the other string

### DIFF
--- a/src/input_common/sdl/sdl_impl.cpp
+++ b/src/input_common/sdl/sdl_impl.cpp
@@ -254,11 +254,25 @@ public:
     }
 
     bool IsJoyconLeft() const {
-        return std::strstr(GetControllerName().c_str(), "Joy-Con Left") != nullptr;
+        const std::string controller_name = GetControllerName();
+        if (std::strstr(controller_name.c_str(), "Joy-Con Left") != nullptr) {
+            return true;
+        }
+        if (std::strstr(controller_name.c_str(), "Joy-Con (L)") != nullptr) {
+            return true;
+        }
+        return false;
     }
 
     bool IsJoyconRight() const {
-        return std::strstr(GetControllerName().c_str(), "Joy-Con Right") != nullptr;
+        const std::string controller_name = GetControllerName();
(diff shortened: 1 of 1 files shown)


Whovian9369:
Can confirm the dual Joycon option (`Nintendo Dual Joy-Con 0`) appears in input option.
Though from my limited testing, actual control with left Joycon acts as if controller is sideways?
(For example, "Right" on the joystick actually inputs "Up", etc.)

Tested with Azure build CI for this PR on Ubuntu 21.04.

src/input_common/sdl/sdl_impl.cpp
```diff
@@ -254,11 +254,23 @@ class SDLJoystick {
     }
 
     bool IsJoyconLeft() const {
-        return std::strstr(GetControllerName().c_str(), "Joy-Con Left") != nullptr;
+        if (std::strstr(GetControllerName().c_str(), "Joy-Con Left") != nullptr) {
```
lioncash, line 257: The `GetControllerName()` call can be moved to a const local, to avoid potentially constructing the string twice in a row

Ditto for `IsJoyconRight()`

german77:
> Though from my limited testing, actual control with left Joycon acts as if controller is sideways?
> (For example, "Right" on the joystick actually inputs "Up", etc.)

As explained on discord the mapping issue is more driver related could be fixed with a custom mapping. For now it at least detects properly the dual joycon

v1993:
On what driver does this happen? HIDAPI, the one we want to use, provides same name across all platforms.

german77:
He did said it did not work properly. If there's a special setup. It must be worth it to make a guide on the webpage

v1993:
You need to add udev rule that allows users to access hidraw devices in RW mode. It varies a little bit per distro, but should be fairly easy to set up.

I'll look into making a proper guide. In the meantime, feel free to ask me on discord about getting it to work (and yeah, close this PR since it doesn't really solve any issues).

german77:
I'm going to keep this PR open. Even if the mappings are wrong a workaround can arrive later or the user can fix the mappings

bunnei (approved):

PR #6980, opened 2021-09-07
Fix blend equation enum error

### LINKED ISSUE #6707: Texture processing abnormal after starting a new game in DQ3
Hello Guys,

Recently, I’m trying to solve the problem that displays incorrect graphics at the beginning of DQ3.

Abnormal:
image

Normal:
image

After my debugging, I finally found that there was a problem with the background texture, Under normal circumstances this texture should look like this:
image

But the result given by yuzu is like this:
image

We can find that there is a piece in the middle that should be transparent, but the data actually obtained is not transparent.

I debugged the code that handles the texture, but in the end I didn’t find the cause of the problem. Can anyone provide me with a little help? Thanks!

vonchenplus:
Close #6707,

Before：
image

After：
image

Finally, I think all the issue about DQ3 should have been solved, so is it possible to consider updating the compatibility list?

### DIFF
--- a/src/video_core/engines/maxwell_3d.h
+++ b/src/video_core/engines/maxwell_3d.h
@@ -475,10 +475,10 @@ public:
 
                 // These values are used by Nouveau and some games.
                 AddGL = 0x8006,
-                SubtractGL = 0x8007,
-                ReverseSubtractGL = 0x8008,
-                MinGL = 0x800a,
-                MaxGL = 0x800b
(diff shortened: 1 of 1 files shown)


ameerj:
Good catch! Gonna put this in EA for testing.

FernandoS27 (approved):

FernandoS27:
I can verify it's correct according to nouveau's docs.

PR #6987, opened 2021-09-08
common: Move error handling functions out of common_funcs

Morph1984:
This avoids implicitly including `<string>` (and by extension `<string_view>`) in all files including common_funcs.h
While we're at it, we can also replace `<algorithm>` with `<iterator>` for `std::size` in common_funcs.h

### DIFF
--- a/src/common/error.h
+++ b/src/common/error.h
@@ -1,0 +1,21 @@
+// Copyright 2013 Dolphin Emulator Project / 2014 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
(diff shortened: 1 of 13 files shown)


src/common/error.h
```diff
@@ -0,0 +1,17 @@
...
+
+#pragma once
+
+#include <string>
+
+// Generic function to get last error message.
+// Call directly after the command or use the error num.
+// This function might change the error code.
+// Defined in error.cpp.
+[[nodiscard]] std::string GetLastErrorMsg();
```
lioncash, line 13: These can be put in the Common namespace

PR #6997, opened 2021-09-12
main: Apply confirm exit setting in exit locked scenarios

ameerj:
Some titles set an exit lock through HLE, which prompts an exit confirmation when stopping emulation if the system is locked.

This change allows bypassing this confirmation if the setting to confirm exits has been disabled by the user.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -3174,12 +3174,11 @@ std::optional<u64> GMainWindow::SelectRomFSDumpTarget(const FileSys::ContentProv
 }
 
 bool GMainWindow::ConfirmClose() {
-    if (emu_thread == nullptr || !UISettings::values.confirm_before_closing)
+    if (emu_thread == nullptr || !UISettings::values.confirm_before_closing) {
         return true;
-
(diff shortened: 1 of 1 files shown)


src/yuzu/main.cpp
```diff
@@ -3174,12 +3174,11 @@ std::optional<u64> GMainWindow::SelectRomFSDumpTarget(const FileSys::ContentProv
...
 bool GMainWindow::ConfirmClose() {
-    if (emu_thread == nullptr || !UISettings::values.confirm_before_closing)
+    if (emu_thread == nullptr || !UISettings::values.confirm_before_closing) {
         return true;
-
-    QMessageBox::StandardButton answer =
-        QMessageBox::question(this, tr("yuzu"), tr("Are you sure you want to close yuzu?"),
-                              QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
+    }
+    const QMessageBox::StandardButton answer =
```
Morph1984, line 3180:
```suggestion
    const auto answer =
```

Morph1984 (approved):

PR #7001, opened 2021-09-12
vk_rasterizer: Fix dynamic StencilOp updating when two faces are enabled

ameerj:
This function was incorrectly using the stencil_two_side_enable register when dynamically updating the StencilOp.

Fixes missing graphics in `WarioWare: Get It Together!` when using the Vulkan API

Before |  After
wario-bug | wario-fix

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -765,12 +765,7 @@ void RasterizerVulkan::UpdateStencilOp(Tegra::Engines::Maxwell3D::Regs& regs) {
     const Maxwell::StencilOp zpass = regs.stencil_front_op_zpass;
     const Maxwell::ComparisonOp compare = regs.stencil_front_func_func;
     if (regs.stencil_two_side_enable) {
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -764,13 +764,15 @@ void RasterizerVulkan::UpdateStencilOp(Tegra::Engines::Maxwell3D::Regs& regs) {
     const Maxwell::StencilOp zfail = regs.stencil_front_op_zfail;
     const Maxwell::StencilOp zpass = regs.stencil_front_op_zpass;
     const Maxwell::ComparisonOp compare = regs.stencil_front_func_func;
-    if (regs.stencil_two_side_enable) {
+    if (!regs.stencil_two_side_enable) {
```
lioncash, line 767: Maybe invert the conditional bodies so that the negation isn't necessary

ameerj: Didn't want to pad the diff unnecessarily since the fix was just an inversion of the condition.

lioncash (approved):

FernandoS27 (approved):

PR #7003, opened 2021-09-13
vk_swapchain: Use immediate present mode when mailbox is unavailable and FPS is unlocked

ameerj:
Allows supported drivers that do not support `VK_PRESENT_MODE_MAILBOX_KHR` the ability to present at a framerate higher than the monitor's refresh rate when the FPS is unlocked.

Note that this may cause visible tearing as frames are always presented as they are complete without waiting for a vertical blank.


goldenx86:
Will have to test this with enhanced sync enabled.

bunnei (approved):

PR #7007, opened 2021-09-13
Build System: Build with JCC Erratum Mitigation

FernandoS27:


v1993:
Should this be unconditionally enabled everywhere? I get that it makes sense for distributable (official) builds, but there's no need to have this enabled for non-affected users compiling Yuzu themselves.

FernandoS27:
> Should this be unconditionally enabled everywhere? I get that it makes sense for distributable (official) builds, but there's no need to have this enabled for non-affected users compiling Yuzu themselves.

It has no negative effect on non affected CPUs.

v1993:
*Then why not just have it as compiler default?* Anyways, looks like GCC used for builds doesn't support it. Also, libraries from externals aren't using this flag (not sure if you want them to),

dnmodder:
c++: error: unrecognized command-line option ‘-mbranches-within-32B-boundaries’

ameerj:
You should add an ELI5 what JCC Erratum is, why care about mitigating it, what are the tradeoffs being made, etc. Otherwise it's not clear why this PR matters at all.

FernandoS27:
A simple google search tells you:
https://devblogs.microsoft.com/cppblog/jcc-erratum-mitigation-in-msvc/

PR #7009, opened 2021-09-14
core: Destroy main_process during shutdown

ameerj:
The main_process was never being cleaned up, causing a noticeable memory leak on subsequent launches. 
This change cleans up the memory during Core Shutdown, mitigating the leak.

For reference, here's the memory usage after 4 launches of a relatively light homebrew:

Before:
leak

After:
destroyed

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -83,6 +83,12 @@ FileSys::StorageId GetStorageIdForFrontendSlot(
     }
 }
 
+void KProcessDeleter(Kernel::KProcess* process) {
+    process->Destroy();
+}
+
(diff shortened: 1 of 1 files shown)


src/core/core.cpp
```diff
@@ -83,6 +83,14 @@ FileSys::StorageId GetStorageIdForFrontendSlot(
     }
 }
 
+void KProcessDeleter(Kernel::KProcess* process) {
+    if (process) {
```
lioncash, line 87: A unique_ptr's deleter will only ever be called if its pointer isn't null

bunnei (approved):
LGTM, thanks for tracking this down

PR #7013, opened 2021-09-14
Detect depth mode by viewport

vonchenplus:
This PR aims to solve the blurred of ZLA in opengl mode.

By capturing the data using Renderdoc, the cause of the blurring is that one frame of data tried to drawing through 4 textures (see below), but there was an error in the depth texture that caused the blurred texture copy area to be calculated incorrectly.

image

According to my guess there should be a problem with the depth mode in the register. Currently I dynamically determine the value of the depth mode by comparing the values of viewport transform and viewport.

Before:
image

After:
image

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -566,10 +566,8 @@ void RasterizerOpenGL::SyncViewport() {
         if (regs.screen_y_control.y_negate != 0) {
             flip_y = !flip_y;
         }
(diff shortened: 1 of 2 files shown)


FernandoS27:
A. This likely causes regressions somewhere else.

B. the correct solution is to use unrestricted depth range, sadly only NVIDIA supports it in OGL and it's broken for D24S8 textures.

There's another solution though. Which is to rescale the depth on the vertex shader to fit 0,1 depth range from the original one. Look at how we rescale the depth on Vulkan when it's 0,1

v1993:
Linked issue specifically talks about Vulkan, so this doesn't look like it would fix it.

PR #7017, opened 2021-09-15
Spir-V: Rescale the frag depth to 0,1 mode when -1,1 mode is used in Vulkan.

FernandoS27:
This fixes Shadows in UE4 games on Vulkan.

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
@@ -477,7 +477,13 @@ void EmitSetSampleMask(EmitContext& ctx, Id value) {
 }
 
 void EmitSetFragDepth(EmitContext& ctx, Id value) {
-    ctx.OpStore(ctx.frag_depth, value);
+    if (!ctx.runtime_info.convert_depth_mode) {
+        ctx.OpStore(ctx.frag_depth, value);
+        return;
+    }
+    const Id unit{ctx.Const(0.5f)};
+    const Id new_depth{ctx.OpFma(ctx.F32[1], value, unit, unit)};
+    ctx.OpStore(ctx.frag_depth, new_depth);
 }
 
 void EmitGetZFlag(EmitContext&) {


src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -477,6 +477,14 @@ void EmitSetSampleMask(EmitContext& ctx, Id value) {
 }
 
 void EmitSetFragDepth(EmitContext& ctx, Id value) {
+    if (ctx.runtime_info.convert_depth_mode) {
+        const Id type{ctx.F32[1]};
+        const Id unit{ctx.Constant(type, 0.5f)};
```
ameerj, line 482:
```suggestion
        const Id half{ctx.Const(0.5f)};
```

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -477,6 +477,14 @@ void EmitSetSampleMask(EmitContext& ctx, Id value) {
...
+        const Id half_depth{ctx.OpFMul(type, value, unit)};
+        const Id new_depth{ctx.OpFAdd(type, half_depth, unit)};
```
ameerj, line 484: maybe `ctx.OpFma`?

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -477,6 +477,14 @@ void EmitSetSampleMask(EmitContext& ctx, Id value) {
...
+        ctx.OpStore(ctx.frag_depth, new_depth);
+        return;
+    }
     ctx.OpStore(ctx.frag_depth, value);
```
ameerj, line 488: Just a nit, I'd invert the condition. This way, only the single instruction is in the body of the condition, and the more complex condition is outside the braces.

FernandoS27: sadly it's the most common case.

ameerj (approved):

PR #7042, opened 2021-09-18
Hide XInput bypass on non-Windows OSes

v1993:
Follow-up to #6950. This option is a no-op on other OSes and only serves to spread confusion there.


v1993:
Well, technically, it isn't a no-op. If someone copies their config from Windows with this option set, they will have web engine disabled without apparent reasons.

ameerj (approved):

PR #7045, opened 2021-09-19
Tas configure ui nits

behunin:
Text looked cramped on my pc (Ubuntu 21.04). Re-flowed text as well for nicer read.
Before | After
--------- | --------
Screenshot from before | Screenshot from after

### DIFF
--- a/src/yuzu/configuration/configure_tas.ui
+++ b/src/yuzu/configuration/configure_tas.ui
@@ -2,37 +2,23 @@
 <ui version="4.0">
   <class>ConfigureTas</class>
   <widget class="QDialog" name="ConfigureTas">
-    <property name="geometry">
-      <rect>
-        <x>0</x>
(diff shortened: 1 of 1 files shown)


src/yuzu/configuration/configure_tas.ui
```diff
@@ -6,8 +6,8 @@
       <rect>
         <x>0</x>
         <y>0</y>
-        <width>800</width>
-        <height>300</height>
+        <width>900</width>
+        <height>506</height>
```
v1993, line 10: Those are rather nonstandard values and will vary per font used. Please consider making it scrollable instead.

v1993 (changes requested):

lioncash:
Can you post a pair of before/after screenshots so the visual changes can be seen?

bunnei (approved):

PR #7082, opened 2021-09-24
service: bsd: Implement Read

Morph1984:
- Used by Diablo II: Resurrected


Morph1984:
Seems like this requires other BSD functions to be implemented to work properly.

bunnei (approved):

bunnei:
Stub is fine here -- I'd prefer to avoid perma-tagged fixes

PR #7090, opened 2021-09-25
(Minor Change) Hint (?) Button Removal and Spacing Change to the TAS Window

Moonlacer:
Before (EA 2083)| After
------------ | -------------
image | image

Notice where "OK" and "Cancel" are located after stretching the window down
Instead of it being in the middle of the blank space as can be seen on the left picture, it will be at the bottom of the window at all sizes

Also removes the useless "?" button on the top right


bunnei (approved):
LGTM, thank you!

PR #7091, opened 2021-09-25
core: Fix memory leak

vonchenplus:
1. Fix KScopedAutoObject leak when SendSyncRequest
  **Because the KScopedSchedulerLock triggers a fiber switch. when a suspend been triggered, the current KScopedAutoObject reference count may not be released**
2. Fix SharedMemory leak

Known Issue:
1. svc WaitSynchronization objects may not be released correctly
**KSynchronizationObject::Wait will trigger a fiber switch, causing the object not to be released**
2. svc StartThread thread leak
**It is possible that the shutdown was performed without calling ExitThread**
3. HLE Service's event leak
**The event object is treated as a member parameter, when the class is released, the event object will be destructured and needs to be changed to event pointer object**

I tried to solve the three problems above, but I found that sometimes it would run crash, so I'll put it aside for now

### DIFF
--- a/src/core/hle/kernel/k_process.cpp
+++ b/src/core/hle/kernel/k_process.cpp
@@ -23,6 +23,7 @@
 #include "core/hle/kernel/k_scheduler.h"
 #include "core/hle/kernel/k_scoped_resource_reservation.h"
 #include "core/hle/kernel/k_shared_memory.h"
+#include "core/hle/kernel/k_shared_memory_info.h"
 #include "core/hle/kernel/k_slab_heap.h"
 #include "core/hle/kernel/k_thread.h"
 #include "core/hle/kernel/kernel.h"
@@ -254,10 +255,26 @@ ResultCode KProcess::AddSharedMemory(KSharedMemory* shmem, [[maybe_unused]] VAdd
     // Lock ourselves, to prevent concurrent access.
     KScopedLightLock lk(state_lock);
 
-    // TODO(bunnei): Manage KSharedMemoryInfo list here.
+    // Try to find an existing info for the memory.
+    KSharedMemoryInfo* shemen_info = nullptr;
+    const auto iter = std::find_if(
+        shared_memory_list.begin(), shared_memory_list.end(),
+        [shmem](const KSharedMemoryInfo* info) { return info->GetSharedMemory() == shmem; });
+    if (iter != shared_memory_list.end()) {
+        shemen_info = *iter;
+    }
 
-    // Open a reference to the shared memory.
+    if (shemen_info == nullptr) {
+        shemen_info = KSharedMemoryInfo::Allocate(kernel);
+        R_UNLESS(shemen_info != nullptr, ResultOutOfMemory);
+
+        shemen_info->Initialize(shmem);
+        shared_memory_list.push_back(shemen_info);
+    }
+
+    // Open a reference to the shared memory and its info.
     shmem->Open();
+    shemen_info->Open();
 
     return ResultSuccess;
 }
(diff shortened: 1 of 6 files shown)


FernandoS27:
Let me know when it's ready for testing.

vonchenplus:
> Let me know when it's ready for testing.

it's ready for testing

FernandoS27:
Which source are you using for this ? It seems inaccurate. Best is to use the main source:
https://github.com/Atmosphere-NX/Atmosphere/tree/master/libraries/libmesosphere

we try to keep everything as close as possible since mesosphere is Hardware tested.

vonchenplus:
My changes do not add any third-party libraries

src/core/hle/service/acc/async_context.cpp
```diff
@@ -7,11 +7,10 @@
 #include "core/hle/service/acc/async_context.h"
 
 namespace Service::Account {
-IAsyncContext::IAsyncContext(Core::System& system_)
-    : ServiceFramework{system_, "IAsyncContext"}, compeletion_event{system_.Kernel()} {
+IAsyncContext::IAsyncContext(Core::System& system_) : ServiceFramework{system_, "IAsyncContext"} {
 
-    Kernel::KAutoObject::Create(std::addressof(compeletion_event));
-    compeletion_event.Initialize("IAsyncContext:CompletionEvent");
+    compeletion_event = Kernel::KEvent::Create(system_.Kernel());
```
bunnei, line 12: This and all others should use `KernelHelpers::ServiceContext::CreateEvent`

src/core/hle/kernel/k_process.cpp
```diff
@@ -254,10 +255,28 @@ ResultCode KProcess::AddSharedMemory(KSharedMemory* shmem, [[maybe_unused]] VAdd
...
-    // Open a reference to the shared memory.
+        shemen_info->Initialize(shmem);
+        shared_memory_list.push_back(shemen_info);
+    }
+
+    // Open a reference to the shared memory and its info.
     shmem->Open();
+    shemen_info->Open();
+
+    // shared_memory_list.push_back(shmem);
```
bunnei, line 279: Remove

src/core/hle/kernel/k_process.cpp
```diff
@@ -412,6 +444,24 @@ void KProcess::Finalize() {
     // Finalize the handle table and close any open handles.
     handle_table.Finalize();
 
+    /* Free all shared memory infos. */
```
bunnei, line 447: Comment style is incorrect (see one ones above)

src/core/hle/service/acc/async_context.cpp
```diff
@@ -25,12 +24,17 @@ IAsyncContext::IAsyncContext(Core::System& system_)
     RegisterHandlers(functions);
 }
 
+IAsyncContext::~IAsyncContext() {
```
bunnei, line 27: Here and elsewhere below should use `KernelHelpers::ServiceContext::CloseEvent`

bunnei (changes requested):
Looks mostly fine -- thank you for tracking down these leaks!

The main comment I have is to use KernelHelpers::ServiceManager helper code for managing events for the HLE services.

Morph1984:
In addition to the comments above - please fixup the `Keep share memory as close` commit with the `Fix share memory object leak` commit

Morph1984:
I've replaced all `KAutoObject::Create(event)` with `KernelHelpers::ServiceContext::CreateEvent` in #7103

vonchenplus:
OK, I revert the relevant changes.

bunnei (approved):

ameerj (approved):

PR #7102, opened 2021-09-29
Remove Boxcat BCAT backend

Morph1984:
This PR formally removes the existing Boxcat BCAT backend implementation, as well as the frontend UI for selecting it.

Additionally, this removes the libzip external as there are no remaining usages of libzip outside of the Boxcat implementation.

The current implementation of BoxCat as it stands is non-functional due to the reliance on a server providing BCAT files.
This implementation will eventually be replaced with one that allows the use of local BCAT files dumped from a Nintendo Switch.


lioncash:
this should probably explain *why* the backend is being removed

Morph1984:
Added an explanation of why the backend is being removed, and that a replacement is being planned to be re-added.

bunnei (approved):

PR #7103, opened 2021-09-29
service: Replace all service event creation with ServiceContext::CreateEvent

Morph1984:
The service context helps to manage all created events and allows us to close them upon destruction.


Morph1984:
Right - I only saw that the KProcess was destructed automatically

bunnei (approved):

PR #7110, opened 2021-09-30
fix extract offline romefs error

vonchenplus:
Fix Ni no Kuni Wrath of the White Witch read offline html document faild


Morph1984 (changes requested):
This will break other web applets since they need to be patched

vonchenplus:
Can you tell me more about it?

Morph1984:
For example, in Super Smash Bros. Ultimate, new updates change the web applet manual romfs, so the update needs to be patched to the base romfs to return the correct web applet.

Morph1984:
Which is what the `else` block should do

src/core/hle/service/am/applets/applet_web_browser.cpp
```diff
@@ -122,6 +123,14 @@ FileSys::VirtualFile GetOfflineRomFS(Core::System& system, u64 title_id,
...
 
         if (nca == nullptr) {
+            if (nca_type == FileSys::ContentRecordType::HtmlDocument) {
+                auto& loader{system.GetAppLoader()};
+                FileSys::VirtualFile romfs;
+                loader.ReadManualRomFS(romfs);
+                if (romfs != nullptr)
+                    return romfs;
+            }
+
```
Morph1984, line 133:
```suggestion
            if (nca_type == FileSys::ContentRecordType::HtmlDocument) {
                LOG_WARNING(Service_AM, "Falling back to AppLoader to get the RomFS");
                FileSys::VirtualFile romfs;
                system.GetAppLoader().ReadManualRomFS(romfs);
                if (romfs != nullptr) {
                    return romfs;
                }
            }

```

Morph1984:
@vonchenplus could you rename the commit to the one I wrote for the title of this PR?

PR #7113, opened 2021-10-01
network: Do not log IP address

Morph1984:
Logging this may be a privacy concern for some users.


german77:
Should logs be removed for UDP controllers as well? It usually links to localhost or devices in the local network

G-Spawn:
At least it was just local ip and not corresponding MAC's as well :p

bunnei (approved):

PR #7118, opened 2021-10-02
gpu: Migrate implementation to the cpp file

ameerj:
This PR mainly aims to reduce the amount of dependencies in `video_core/gpu.h`, as it is a very widely included header file, in an effort to reduce compiling overhead. The definition and implementation of `Tegra::GPU` has moved to the cpp file, with the header now serving only as an interface to its public methods. As a result, many unusued headers were trimmed.

Through this migration, two other changes were needed: the `GPU::LockSync` method's single usage was refactored, and the VSync thread was converted to a `std::jthread` to avoid a deadlock on stopping emulation. Further testing is needed to ensure this refactoring did not introduce other anomalies.

### DIFF
--- a/src/video_core/texture_cache/texture_cache_base.h
+++ b/src/video_core/texture_cache/texture_cache_base.h
@@ -4,13 +4,12 @@
 
 #pragma once
 
-#include <array>
 #include <mutex>
 #include <span>
 #include <type_traits>
 #include <unordered_map>
-#include <unordered_set>
 #include <vector>
+#include <queue>
 
 #include "common/common_types.h"
 #include "common/literals.h"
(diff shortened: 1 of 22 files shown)


src/video_core/texture_cache/texture_cache_base.h
```diff
@@ -4,13 +4,13 @@
...
 #pragma once
 
-#include <array>
 #include <mutex>
 #include <span>
 #include <type_traits>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
+#include <queue>
```
lioncash, line 13: Shouldn't this be below `<mutex>`?

ameerj: clang format says no 🤷

Morph1984: can confirm, clang format somehow places <queue> at the bottom

FernandoS27: I'll just accept this, no idea why clang format is wrong there.

FernandoS27 (approved):

PR #7130, opened 2021-10-05
CMakeLists: Update requirement to CMake 3.18

Morph1984:
Requires https://github.com/yuzu-emu/build-environments/pull/42

Edit: Seems like another issue has been uncovered.
For the MinGW build, it is failing with
```
CMake Error at externals/CMakeLists.txt:75 (add_library):
  add_library cannot create ALIAS target "SDL2" because another target with
  the same name already exists.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.18)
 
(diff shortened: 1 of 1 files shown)


Morph1984:
This will be re-considered when it is needed, as the dynarmic project has been updated to fix the issues pre-3.18. Closing for now.

PR #6702, opened 2021-07-24
yuzu_cmd, yuzu qt: Use SDL to disable the screen saver

lat9nq:
This disables the screen saver when yuzu is running a game via SDL2's screen saver functions. Adds an option to allow or disable the screen saver in General options.

I went and finished this, got ready to make the PR, and noticed #6421 already exists. I thought I propose this, anyway, since (a) this implements it in 1/3 of the code, (b) doesn't add any new dependencies, and (c) SDL2's screen saver inhibition is platform-agnostic --- we implement it here for Linux, and it'll work on Android and macOS and anything else that SDL2's screen saver functions are implemented for, too.

Closes #6262

PR #7211, opened 2021-10-22
Logging: Impl refactor

behunin:
This addresses a few problems I have had with the logger.

Moving the backend thread out of the constructor gives the queue some time to build some entrys before starting to PopWait them off. I would occasionally crash on booting yuzu due to collision in the queue.

Moving to a stop_source ensures that there is a clean shutdown. I would have to manually kill my gdb session because the logger would be stuck waiting on an entry to pop.


bunnei:
Conflicts with https://github.com/yuzu-emu/yuzu/pull/6702 -- Sorry for the delay, should get this tested soon

bunnei:
@behunin now that there isn't a conflict, can you rebase when you get a chance? Thank you!

PR #7212, opened 2021-10-22
Move to MPMCQueue

behunin:
from: https://github.com/rigtorp/MPMCQueue

supersedes: #6874
Thanks to @yzct12345 for the inspiration.

I set the queue size high for now, we will tune it down.

Clear is not thread safe, I just ripped it out so further testing will need to be done but I have not had any problems.

### DIFF
--- a/src/common/atomic_threadsafe_queue.h
+++ b/src/common/atomic_threadsafe_queue.h
@@ -25,18 +25,9 @@ SOFTWARE.
 #include <atomic>
 #include <cassert>
 #include <cstddef> // offsetof
-#include <limits>
 #include <memory>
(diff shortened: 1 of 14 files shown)


german77:
It's a good thing that the input rewrite will remove all the queues from the input side. Only one file will contain the queue of all input events. So tuning the value down will be easier. Don't worry about the rewrite, I will rebase if this PR gets merged sooner.

behunin:
Open world BOTW goes from high 20s to a solid 30 fps.
Using atomics for data.
MPMC opens things up for more speed gains in the future when things are designed for it.

PR #2860, opened 2019-09-15
Implement Native Resolution Rescaling

FernandoS27:
Native Resolution Rescaling in Emulators used to be an almost trivial process for old console GPUs where only 1 render target existed, memory wasn't unified (in the GPU sense, not on SoC sense) and textures weren't anywhere as complex as they are now. Native Resolution Rescaling for a modern GPU as Maxwell 2nd Gen is quite a challenge due to many factors: More complex texture types can be partially or entirely rendered onto: Cube textures, Array Textures, 3d Textures, etc; 8 different render targets with shared viewports and scissor tests, unified memories with uncalled readbacks and hardcoded shaders created for fixed resolutions (glFragCoord & TexelFetch). This PR introduces a set of heuristics and mechanisms to detect possible candidate render targets for rescaling based on a simple evolutive AI algorithm and performs render target rescaling based on a generated database of rescalable rendertargets (rescale profile) plus its corrections on state and shader code.

PR #7219, opened 2021-10-24
Project A.R.T. Advanced Rendering Techniques

### LINKED ISSUE #6847: [Feature Request] Consider to add AMD FidelityFX Super Resolution (FSR) support in Yuzu ?
Hi Yuzu team !! 

AMD FSR (FidelityFX Super Resolution) add the possibility for any harware platform/OS to a good upscaling with less CPU/GPU ans so much less energy than a native resolution (HD, 4K...)
AMD FSR use spatialization (Nvidia DLSS use vectorisation), for example, RPCS3 emulator have implemented it recently. [https://wccftech.com/new-tools-allow-amd-fsr-injection-in-rpcs3-emulator-and-potentially-in-every-game/]

It could be a good solution for better performances on high resolution for many devices, and less energy consumption on laptop, or others devices with battery.

FernandoS27:
Project A.R.T.  (Advanced Rendering Techniques) implements the following features:

- A new base resolution scaler with support for both upscaling and downscaling.
- A new set of window sampling filters for post processing of the game's final frame prior to presentation to the render window.
- A new internal presampling Anti-Aliasing pass.
- Various fixes for regressions introduced by TCR.
- Fix depth/stencil blits In Vulkan for AMD/Intel.

**The Resolution Scaler:**

Unlike the original scaler (#2860), which was inspired by Cemu’s profile-based scaling, the new resolution scaler uses a rating system. This rating system works by having a set of rules dictating which render targets can be scaled and having a rating for each texture. We couldn’t simply use a binary “YES/NO” system because some games may render to textures only once. Scaling these would be wasteful and has the potential to break them.

Textures can only earn 1 credit per frame. And to earn that 1 credit, the texture itself as well as any other textures in that renderpass, all need to satisfy our set rules. After accruing 2 or more credits, the texture will be scaled and now all rendering will be done in the scaled resolution for that texture.

The textures keep earning more credits as frames pass. But if in any frame, even a single texture within a renderpass fails to satisfy the set rules, all of texture scores will be reset to 0. If a texture interacts with other textures, its new rating will either be based on the current rating of those textures or will be increased to the maximum possible rating, if one of those textures is already scaled.

**Window Sampling Filters**

Project A.R.T. also introduces a set of new sampling methods: 
- Nearest Neighbor
- Bilinear (standard filtering used prior to A.R:T.)
- Bicubic
- Gaussian
- ScaleForce
- AMD's FidelityFX™️ Super Resolution (only works on Vulkan).

**Others**

We also added FXAA in an Anti-Aliasing pass, and fixed some of TCR's regressions like blur the in `Bayonetta 2`'s cutscenes, blur in `Pokemon Sword/Shield` cutscenes on AMD/Intel drivers, and `Yu-Gi-Oh: Legacy of the Duelist`'s cards not rendering.

**Trivia: The internal name of A.R.T.**

Internally, A.R.T. was referred to as Aristotle's Right Testicle as an inside joke. Back during the development of the Prometheus project, some guy called it aristotles-left-testicle. I don't know why or how he came to such silly conclusion but we were laughing our asses for a week or so. we would do comments like "are we showing the aristotle's testicle to everyone yet (Prometheus' release)". It was good fun.

**Credits:** 
- Developers:  @ameerj, @ReinUsesLisp, @BreadFish64, @FernandoS27
- Testing: @BSoDGamingYT, @german77, @gidoly, @goldenx86, @G-Spawn, @lat9nq, @Morph1984, @MysticExile, @Norstyrm, @theboy181
- Advice: @degasus from Dolphin.

Closes #6847


src/video_core/renderer_opengl/gl_texture_cache.cpp
```diff
@@ -1201,7 +1201,14 @@ Sampler::Sampler(TextureCacheRuntime& runtime, const TSCEntry& config) {
...
 
     if (GLAD_GL_ARB_texture_filter_anisotropic || GLAD_GL_EXT_texture_filter_anisotropic) {
-        glSamplerParameterf(handle, GL_TEXTURE_MAX_ANISOTROPY, config.MaxAnisotropy());
+        const f32 setting_anisotropic =
+            static_cast<f32>(1U << Settings::values.max_anisotropy.GetValue());
+        const f32 game_anisotropic = std::clamp(config.MaxAnisotropy(), 1.0f, 16.0f);
+        const bool aument_anisotropic =
+            game_anisotropic > 1.0f || config.mipmap_filter == TextureMipmapFilter::Linear;
+        const f32 max_anisotropy =
+            aument_anisotropic ? std::max(game_anisotropic, setting_anisotropic) : game_anisotropic;
```
v1993, line 1210: Perhaps move this into a function, since logic is the same for both video backends.

SpecialDarkwolf999:
@FernandoS27 do you know why the upscaller causes stuttering when u use upscale on 2x or higher.
if you use 1x then this problem doesnt excists and this is with several games.

RinMaru:
Atelier and Spyro have weird upscaling artifacts akin to dithering. Filters make no diff

FSR

TheBlackGamer:
Pokemon Sword running on yuzu early access 2189
Some weird shadow glitch
https://imgur.com/a/BFzkmVy

[yuzu_log.zip](https://github.com/yuzu-emu/yuzu/files/7490018/yuzu_log.zip)
Log File

FernandoS27:
> Pokemon Sword running on yuzu early access 2189 Some weird shadow glitch https://imgur.com/a/BFzkmVy
> 
> [yuzu_log.zip](https://github.com/yuzu-emu/yuzu/files/7490018/yuzu_log.zip) Log File

use theboy181's outline mods instead of the old one. The old one causes those issues.

RinMaru:
0100c3c012718000_2021-11-16_20-35-55-142
GTA games do not upscale at all

bunnei (approved):

Morph1984 (approved):
Nice

PR #7227, opened 2021-10-26
Fix memory leak v2

vonchenplus:
1. Fix svc WaitSynchronization objects may not be released correctly.
2. Fix svc StartThread thread leak.
3. Fix dangling kernel objects when exiting.
4. Revert #7009


bunnei (approved):

bunnei:
LGTM, thank you!

PR #7259, opened 2021-10-31
nvhost_ctrl: Stub out IocCtrlFreeEventBatch

vonchenplus:
Used by Dragon Quest Builders

Many thanks to the Skyline Team and Contributors for their help!

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_ctrl.cpp
@@ -2,6 +2,10 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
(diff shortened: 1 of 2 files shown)


FernandoS27:
after discussing with Bylaws. The copyright of this is perfectly Skyline's as far as you used nvdec's branch of skyline emulator (please verify this).

The ryujinx copyright is because the code has been verified against their implementation.

You must select the MPL 2.0 license for this (remove MIT or) as it's the closest to GPLv2.

vonchenplus:
> after discussing with Bylaws from Skyline. The copyright of this is perfectly Skyline's as far as you used nvdec's branch of skyline emulator (please verify this).
> 
For IocCtrlFreeEventBatch, the  nvdec's branch only modifies one compilation warning and I have done it.

PR #7262, opened 2021-10-31
ShaderCache: Order Phi Arguments from farthest away to nearest.

FernandoS27:
This fixes the post processing in Mario Party SuperStars.

The idea of this is so the tracking algorithm, gets the nearest instances first rather than the oldest instances of a value.

### DIFF
--- a/src/shader_recompiler/frontend/ir/value.h
+++ b/src/shader_recompiler/frontend/ir/value.h
@@ -182,6 +182,9 @@ public:
     /// Add phi operand to a phi instruction.
     void AddPhiOperand(Block* predecessor, const Value& value);
 
+    /// Orders the Phi arguments from farthest away to nearest.
+    void OrderPhiArgs();
+
     void Invalidate();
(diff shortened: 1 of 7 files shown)


src/shader_recompiler/frontend/ir/value.h
```diff
@@ -182,6 +182,9 @@ class Inst : public boost::intrusive::list_base_hook<> {
     /// Add phi operand to a phi instruction.
     void AddPhiOperand(Block* predecessor, const Value& value);
 
+    /// Orders the Phi arguments.
```
lioncash, line 185: Perhaps go into more detail about how it orders the arguments, otherwise this is just repeating what the function name already indicates

FernandoS27: xD true

theboy181: NIce!

bunnei (approved):

PR #7274, opened 2021-11-02
Skip checking if a u64 is non-negative

fpdotmonkey:
It is sufficient to allow the compiler to check if num_handles in
svc.cpp:WaitSynchronization is non-negative because it is unsigned, so
let's not put code (and compiler warnings!) towards that.


SciresM:
This is an inaccuracy and should not be merged.

the value in question is an int32_t received from userland.

a correct fix is to change the value from u64 to s32.

Morph1984:
Superseded by #7278

PR #7278, opened 2021-11-03
svc: Correct WaitSynchronization num_handles param type

Morph1984:
num_handles is a s32

Supersedes #7274


src/core/hle/kernel/svc.cpp
```diff
@@ -409,7 +409,7 @@ static ResultCode GetProcessId32(Core::System& system, u32* out_process_id_low,
 
 /// Wait for the given handles to synchronize, timeout after the specified nanoseconds
 static ResultCode WaitSynchronization(Core::System& system, s32* index, VAddr handles_address,
-                                      u64 num_handles, s64 nano_seconds) {
+                                      s32 num_handles, s64 nano_seconds) {
```
lioncash, line 412: The loop in the SCOPE_EXIT below should have the signedness of `i` changed, otherwise there's a mismatched sign comparison

lioncash (approved):

FernandoS27 (approved):

PR #7283, opened 2021-11-04
service: acc: Stub acc:u0 '160'

Morph1984:
- Used by Animal Crossing: New Horizons v2.0.0

Since the name is currently unknown, '160' is used as a placeholder.


FernandoS27 (approved):

src/core/hle/service/acc/acc_u0.cpp
```diff
@@ -34,6 +34,7 @@ ACC_U0::ACC_U0(std::shared_ptr<Module> module_, std::shared_ptr<ProfileManager>
         {140, &ACC_U0::InitializeApplicationInfoRestricted, "InitializeApplicationInfoRestricted"}, // 6.0.0+
         {141, &ACC_U0::ListQualifiedUsers, "ListQualifiedUsers"}, // 6.0.0+
         {150, &ACC_U0::IsUserAccountSwitchLocked, "IsUserAccountSwitchLocked"}, // 6.0.0+
+        {160, &ACC_U0::Unknown160, "Unknown160"},
```
vonchenplus, line 37: The name of 160 is InitializeApplicationInfoV2, as you can see by switchbrew.

FernandoS27 (approved):

src/core/hle/service/acc/acc_u0.cpp
FernandoS27: oh oops, i didn't refresh this and there wasn't a comment before merging.

PR #7291, opened 2021-11-05
Add miss vulkan image format

vonchenplus:
Used by Dragon Quest Builders


german77:
Any game that uses this format?

Morph1984 (approved):

PR #7293, opened 2021-11-05
service/pctl: Stub EndFreeCommunication

german77:
Used by Just Dance 2022


Morph1984:
amend the commit message to include the game that this is used by

PR #7294, opened 2021-11-05
Fix image update/download error when width too small

vonchenplus:
Currently, there is a restriction on the level when create an image, but there is no such restriction when update and download, opengl report an error when calling the corresponding method.

### DIFF
--- a/src/video_core/renderer_opengl/gl_texture_cache.h
+++ b/src/video_core/renderer_opengl/gl_texture_cache.h
@@ -219,6 +219,7 @@ private:
     GLenum gl_internal_format = GL_NONE;
     GLenum gl_format = GL_NONE;
     GLenum gl_type = GL_NONE;
+    GLsizei gl_num_levels{};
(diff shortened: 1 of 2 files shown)


Morph1984:
Currently conflicts with #7219

FernandoS27:
@vonchenplus could you rebase?

vonchenplus:
I will rebase after the A.R.T been merged.

src/video_core/renderer_opengl/gl_texture_cache.h
```diff
@@ -185,6 +185,11 @@ class Image : public VideoCommon::ImageBase {
         return gl_type;
     }
 
+    GLsizei NumLevels() const noexcept {
```
ameerj, line 188: This should be a private method, whose implementation is in the cpp file, 
or a private member variable that's initialized once and reused.

bunnei (approved):

PR #7302, opened 2021-11-08
[input_common] Fixed thread hang

VPeruS:
Partially reverted b01aa72, thread was unable to change status to "Completed".  Calibration is supposed to have multiple request-response exchanges, because of that we cannot store calibration's temporary data inside of callback's local variables.
Added small test for verification.

### DIFF
--- a/src/input_common/drivers/udp_client.cpp
+++ b/src/input_common/drivers/udp_client.cpp
@@ -536,8 +536,6 @@ CalibrationConfigurationJob::CalibrationConfigurationJob(
     std::function<void(u16, u16, u16, u16)> data_callback) {
 
     std::thread([=, this] {
-        constexpr u16 CALIBRATION_THRESHOLD = 100;
-
         u16 min_x{UINT16_MAX};
         u16 min_y{UINT16_MAX};
         u16 max_x{};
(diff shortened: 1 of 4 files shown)


FernandoS27:
@german77 could you review this?

vonchenplus:
This looks like only a partial rollback of PR #7207, have you tested the compilation of VS 2022?

VPeruS:
No, I haven't tested compilation with VS2022

german77:
This fixes issues with mapping touch. But we need to verify that compiles on vs2022. I will approve once I manage to do that.

german77:
Could you rebase. Files changed a bit

german77 (approved):

src/input_common/drivers/udp_client.cpp
```diff
@@ -536,42 +536,46 @@ CalibrationConfigurationJob::CalibrationConfigurationJob(
     std::function<void(u16, u16, u16, u16)> data_callback) {
 
     std::thread([=, this] {
+        constexpr u16 CALIBRATION_THRESHOLD = 100;
```
ameerj, line 539: This needs to be declared in the lambda body to avoid a VS2022 compilation error.

VPeruS: Done

ameerj (approved):

bunnei (approved):

bunnei:
LGTM, thank you!

PR #7303, opened 2021-11-08
applets/swkbd: Skip text checking if the text has been confirmed

### LINKED ISSUE #6712: Strange chinese error when entering name in famicom detective club
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/6876887/yuzu_log.txt)

изображение

Morph1984:
Confirm means that the text has already been checked by the application to be correct, but is asking the user for confirmation.
This also fixes the corrupted text in the confirmation dialog, which is caused by using the incorrect encoding.

Fixes #6712 

Fixes the software keyboard in Famicom Detective Club: The Missing Heir


Morph1984:
The latest commit fixes the text corruption, it turns out it can be encoded in UTF-8.
image

FernandoS27 (approved):

PR #7330, opened 2021-11-14
Replace "Light" theme by "Default"

MightyCreak:
This reflects the current behavior: Light = System default. If your
system is set to dark theme, then Light = Dark, which is a bit confusing
for the end user.

In this PR, I propose to change "Light" with "Default". This way, the
user has "Default" and "Default Colorful", which will apply the system
theme. Now that the Flatpak respects the system theme, I think this
makes much more sense.

I also simplified the theme update. Before the code was branching
between the default theme and the others, but I think we can have
something simpler by forcing the default theme if no theme is defined in
the settings, or if the selected theme doesn't exist. And if there's an
error, tell the theme name in the error message.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -3381,36 +3381,38 @@ void GMainWindow::filterBarSetChecked(bool state) {
 }
 
 void GMainWindow::UpdateUITheme() {
-    const QString default_icons = QStringLiteral("default");
-    const QString& current_theme = UISettings::values.theme;
-    const bool is_default_theme = current_theme == QString::fromUtf8(UISettings::themes[0].second);
+    const QString default_theme = QStringLiteral("default");
+    QString current_theme = UISettings::values.theme;
     QStringList theme_paths(default_theme_paths);
 
-    if (is_default_theme || current_theme.isEmpty()) {
-        const QString theme_uri(QStringLiteral(":default/style.qss"));
-        QFile f(theme_uri);
-        if (f.open(QFile::ReadOnly | QFile::Text)) {
-            QTextStream ts(&f);
-            qApp->setStyleSheet(ts.readAll());
-            setStyleSheet(ts.readAll());
-        } else {
-            qApp->setStyleSheet({});
-            setStyleSheet({});
-        }
-        QIcon::setThemeName(default_icons);
-    } else {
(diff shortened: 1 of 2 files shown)


MightyCreak:
I tested locally on Fedora 34 with my system set to the light and dark theme.

I don't know what happens when running on KDE, I suppose the old "Light" theme was behaving the same as in GNOME... Maybe someone would like to test that?

src/yuzu/main.cpp
```diff
@@ -3381,36 +3381,38 @@ void GMainWindow::filterBarSetChecked(bool state) {
...
     QStringList theme_paths(default_theme_paths);
 
-    if (is_default_theme || current_theme.isEmpty()) {
-        const QString theme_uri(QStringLiteral(":default/style.qss"));
+    if (current_theme.isEmpty()) {
+        current_theme = default_theme;
+    }
+
+    if (current_theme != default_theme) {
+        QString theme_uri(QLatin1Char{':'} + current_theme + QStringLiteral("/style.qss"));
```
lioncash, line 3393:
```suggestion
        QString theme_uri{QStringLiteral(":%1/style.qss").arg(current_theme)};
```

Seems like it would be a little more concise

src/yuzu/main.cpp
```diff
@@ -3381,36 +3381,38 @@ void GMainWindow::filterBarSetChecked(bool state) {
...
-            setStyleSheet({});
+        if (!f.open(QFile::ReadOnly | QFile::Text)) {
+            LOG_ERROR(Frontend, "Unable to open style \"{}\", fallback to the default theme",
+                      UISettings::values.theme.toStdString());
+            current_theme = default_theme;
         }
-        QIcon::setThemeName(default_icons);
+    }
+
+    QString theme_uri(QLatin1Char{':'} + current_theme + QStringLiteral("/style.qss"));
```
lioncash, line 3402: Ditto

src/yuzu/main.cpp
MightyCreak: Done!

src/yuzu/main.cpp
MightyCreak: Done!

bunnei (approved):

Issue #6479, opened 2021-06-18
Resolve Thread Sanitizer issues


lioncash:
![Discord Image](https://user-images.githubusercontent.com/712067/122512671-2de78d00-cfd7-11eb-9c51-ee4996b372a2.png)

We have several cases where Clang and GCC's thread sanitizer flag races in our code. Ideally, these should be investigated.

- Worst-case scenario: some bug vectors get cleaned up and it makes our code more reliable.
- Best-case scenario: it fixes some bugs and funky behaviors that are encountered in some games.

These are both positive outcomes, so any investigation into these would be appreciated.

ReinUsesLisp:
I'll try collecting the known errors in this message.

- [ ] Data race in `owner_thread`
Reader: https://github.com/yuzu-emu/yuzu/blob/0485b8e84bacbf7fd56081822faca46114cbeb85/src/core/hle/kernel/k_scheduler_lock.h#L23 Writer: https://github.com/yuzu-emu/yuzu/blob/0485b8e84bacbf7fd56081822faca46114cbeb85/src/core/hle/kernel/k_scheduler_lock.h#L57

ForceFaction:
Has anyone investigated the last race reported in the list? I am facing segfaults and even though many races got fixed in the meantime this seems to be close to what i'm seeing.

If no one has investigated it yet: How do i myself enable tsan properly with yuzu? i tried but i cannot mange to get around yuzu segfaulting before the GUI even starts...

PR #7339, opened 2021-11-15
Fixes for races in owner_thread and native_clock

wtzhang23:
Partially fixes the first two race conditions listed in issue #6479.


FernandoS27:
This seems to cause a system hang.

FernandoS27:
Personally from my own experience, those "race conditions" are false positive. I have never seen a thread sanatizer detect atomic race conditions correctly.

wtzhang23:
I think I agree. The hang probably occurs in the native_clock changes. The changes in the other file fix a race condition that ultimately doesn't matter since the IsLockedByCurrentThread returning true will only occur if it owns the spinlock. I think I will close this.

PR #7341, opened 2021-11-15
core: Fix crash when exit app

vonchenplus:
When a kernel object construct, it is registered into the KernelCore and unregistered after destruct.

A KThread kernel object is generated for each thread by GetHostDummyThread, and saved by thread local, including the main thread.

So when the main thread exits, the KThread object will be destructured, but by this time, the KernelCore object no longer exists and will cause a crash every time it exits.

**The current solution is to temporarily open a thread to load the game to prevent the main thread creating a thread-local KThread instance.**

This problem was actually introduced by #7182, because now the System instance is no longer global. @Morph1984

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1273,8 +1273,14 @@ bool GMainWindow::LoadROM(const QString& filename, u64 program_id, std::size_t p
         std::make_unique<QtWebBrowser>(*this),         // Web Browser
     });
 
-    const Core::SystemResultStatus result{
-        system->Load(*render_window, filename.toStdString(), program_id, program_index)};
+    Core::SystemResultStatus result{};
+    auto load_thread = std::jthread(
+        [this, filename, program_id, program_index](Core::SystemResultStatus& result) {
+            result =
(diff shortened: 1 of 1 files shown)


lioncash (changes requested):
We've worked extremely hard over several years to remove global state from core. Please don't reintroduce it.

bunnei:
Thanks for tracking down the problem, but +1 with what Lioncash said -- we cannot be moving this state into globals. Can you find a better way to fix this? Thank you!

src/core/hle/kernel/kernel.cpp
```diff
@@ -1110,4 +1110,4 @@ const Core::System& KernelCore::System() const {
     return impl->system;
 }
 
-} // namespace Kernel
```
v1993, line 1113: Unrelated changle

bunnei:
@vonchenplus Thank you for tracking down the issue and providing a valid fix. After discussion with the team, we think the solution with #7359 is better, because it fixes the issue in the emulator core, instead of the Qt frontend (note -- SDL, and possibly other future frontends might have this problem).

PR #7359, opened 2021-11-17
Fix crash on exit due to static scoped dummy threads

heinermann:
`thread_local` created a static variable which is destroyed on application exit (after `KernelCore` is destroyed), which caused a crash on exit while the emulation was running, because `KThread`s have references to `KernelCore`. This solution causes the dummy `KThread`s to be destroyed earlier, when `KernelCore` gets destroyed.

### DIFF
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -300,15 +300,16 @@ struct KernelCore::Impl {
     // Gets the dummy KThread for the caller, allocating a new one if this is the first time
     KThread* GetHostDummyThread() {
         auto make_thread = [this]() {
-            std::unique_ptr<KThread> thread = std::make_unique<KThread>(system.Kernel());
+            std::lock_guard lk(dummy_thread_lock);
+            auto& thread = dummy_threads.emplace_back(std::make_unique<KThread>(system.Kernel()));
             KAutoObject::Create(thread.get());
             ASSERT(KThread::InitializeDummyThread(thread.get()).IsSuccess());
             thread->SetName(fmt::format("DummyThread:{}", GetHostThreadId()));
-            return thread;
+            return thread.get();
         };
 
-        thread_local auto thread = make_thread();
-        return thread.get();
+        thread_local KThread* saved_thread = make_thread();
+        return saved_thread;
     }
 
     /// Registers a CPU core thread by allocating a host thread ID for it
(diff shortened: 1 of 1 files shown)


heinermann:
Could be related to #6479 but I'm not sure?

vonchenplus:
Essentially, we should not save the dummy thread in the KernelCore because the dummy thread should follow the current thread lifecycle, I have provided an alternative solution in PR #7341

heinermann:
> Essentially, we should not save the dummy thread in the KernelCore because the dummy thread should follow the current thread lifecycle, I have provided an alternative solution in PR #7341.
> 
> I think you need to decide on the final solution. @bunnei @lioncash

Where does your PR deal with this? I don't understand your PR from the Files changed tab (8 lines added in main.cpp). The issue would theoretically still happen since that code hasn't been touched right? If `GetHostDummyThread` is ever called in your code it will still crash.

Are you testing the main app in `DEBUG` and closing the window after starting emulation?

vonchenplus:
Please take a careful look at my PR description, the reason why the problem has occurred.

heinermann:
Yes it's the same problem but I don't understand what it is doing or why it fixes the issue.

vonchenplus:
As I described in the PR, I avoid the main thread trigger GetHostDummyThread because the main thread only calls GetHostDummyThread when it loads rom.

src/core/hle/kernel/kernel.cpp
```diff
@@ -725,10 +732,6 @@ struct KernelCore::Impl {
     std::unordered_set<KServerSession*> server_sessions;
     std::unordered_set<KAutoObject*> registered_objects;
     std::unordered_set<KAutoObject*> registered_in_use_objects;
-    std::mutex server_ports_lock;
-    std::mutex server_sessions_lock;
-    std::mutex registered_objects_lock;
-    std::mutex registered_in_use_objects_lock;
```
Morph1984, line 731: These should likely stay in place, with `dummy_thread_lock` moved below the `dummy_threads` vector

lioncash:
I feel like I'd be more inclined to go with this solution, since it centralizes the solution in the core instead of being in the frontend (so the same problem wouldn't exist if we were to add another frontend in the future, for whatever reason).

bunnei (approved):

src/core/hle/kernel/kernel.cpp
heinermann: I moved them up so that they are destroyed last in Kernel's destruction process. A lot of other things take on Kernel references, and mutexes have no dependencies. If one of these gets destroyed prior to a random other member with a kernel reference, it can lead to undefined behaviour if the member's destructor calls a Kernel function such as `UnregisterKernelObject` which is the case with `KAutoObject`.

Sorry I wasn't able to address this prior to it being merged.

goldenx86:

heinermann:
> Hi, I'm in charge of writing the monthly progress reports, could we include your work there?

Yep go ahead

Issue #7052, opened 2021-09-19
Accelerate DMA causes a hard crash on Dragon Quest Builders


ameerj:
Booting Dragon Quest Builders with the OpenGL backend causes a hard memory crash. This seems to be a regression caused by #6629. 

To reproduce, boot Dragon Quest Builders with the OpenGL backend. Make sure to save all your work beforehand since the memory crash can corrupt opened programs.

By reverting the dependent fixup 84b9c42 and original merge commit, c53b688 , The crashes no longer occur.

NineKain:
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7288573/yuzu_log.txt)

PR #7375, opened 2021-11-19
Convert all legacy attributes to generic attributes

### LINKED ISSUE #6548: Dragon Quest III with white screen on AMD Video Card and possible corrupted graphics and such...
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/6749704/yuzu_log.txt)

Hello to all, I am experiencing on Linux Mint 20.1 Linux OS with an AMD Radeon 5700XT Video Card with Dragon Quest III on OpenGL a white screen with possible corrupted graphics and on Vulkan a Black Screen with nothing on it and audio working properly as well..  on Latest to greatest Mainline release as of Today also.. Thanks for listening and Have a great day ahead and stay safe also..

vonchenplus:
Rewrote the handling of legacy attributes, convert all legacy attributes to generic attributes.

- Fix ARB_separate_shader_objects extension is not compatible with gl_FogFragCoord.
- Fix Dragon Quest Builders shader build error.
- Resolve legacy attributes not supported by some graphics card drivers.

close #6548


ameerj (changes requested):
GLSL provides compatibility to use legacy attributes. There is no need to take up generic attribute slots for their usage.
`ARB_separate_shader_objects` simply requires that the legacy attributes in use are declared in the input/output blocks in each shader.

vonchenplus:
> GLSL provides compatibility to use legacy attributes. There is no need to take up generic attribute slots for their usage. 

In fact, the legacy attributes are no longer supported on some drivers, such as issue #6548.

> `ARB_separate_shader_objects` simply requires that the legacy attributes in use are declared in the input/output blocks in each shader.

The gl_FogFragCoord attribute will still fail to link even if you declare it, you can use Dragon Quest Builders to test.

ameerj:
> The gl_FogFragCoord attribute will still fail to link even if you declare it, you can use Dragon Quest Builders to test.

This looks to be a very stupid driver bug where it confuses `gl_FogFragCoord` because it has the same name in the vertex shader and the fragment shader.
I resolved the link issues here: https://github.com/ameerj/yuzu/commit/4877ca9f76ed56bfb959a00dcdc0811aa8829991
but I get a driver panic and can't test further at the moment due to the issue outlined here: #7052 

As for legacy attributes not being supported on some drivers. I tested Nvidia and AMD proprietary drivers and they had no issues. It's likely that radeonsi requires some configuration/environment variable to be set if they don't support legacy OGL by default. but I do not know enough about mesa to be certain about that.

vonchenplus:
I tested your code and it looks like the link has no errors, but I can't confirm if it's working properly.

In the long term, we should remove compatibility mode support, because it is hard to guarantee that one day, the driver will no longer support compatibility mode.

But I will respect your opinion, if you insist on keeping the GLSL compatibility mode.

FernandoS27 (approved):
I'm totally fine with dropping native legacy output support considering some OGL implementations don't support them (like those that don't support compatibility profiles) and I'de love to set a goal for ourselves to get rid of compatibility profile in the future.

FernandoS27:
@vonchenplus could you rebase?

Issue #7392, opened 2021-11-21
[Feature Request] Add support for Skyline modding framework


jam1garner:
Spinning this off from #6076 as it's unspecific and doesn't provided any technical details about what it'd entail.

### Motivation

[Skyline](https://github.com/skyline-dev/skyline), for those who aren't familiar, is a framework for modifying the code of games. It acts as a subsdk (overriding subsdk9, as no current game uses it) and uses a plugin system in order to allow users to load their own code, and provides an API for users to patch existing game code in order to redirect into user re-implementations of functions.

It's desirable to support due to its usage across a variety of games, most notably Smash Ultimate, which requires it to do [any serious patching of game files](https://github.com/Raytwo/ARCropolis), but it is also the framework used for moveset edits among other things. It's also been used some in the {Animal Crossing New Horizon, Pokemon SwSh/BdSp, Spelunky, etc} modding scenes, and other developers seem to be showing interest in it. (For example Pokemon Mystery Dungeon has shown interest but needs emulator support first: https://github.com/tech-ticks/hyperbeam/issues/3#issuecomment-967349620)

### Outstanding Issues

* **SVC Support**
  * [x] Support `MapProcessMemory` SVC
    * Implementation status: https://github.com/yuzu-emu/yuzu/pull/7519
  * [x] Support `UnmapProcessMemory` SVC
    * Implementation status: https://github.com/yuzu-emu/yuzu/pull/7519
  * [x] Support `CreateCodeMemory` SVC
    * Implementation status: https://github.com/yuzu-emu/yuzu/pull/7519
  * [x] Support `ControlCodeMemory` SVC
    * Implementation status: https://github.com/yuzu-emu/yuzu/pull/7519
  * [x] Support `SetProcessMemoryPermission` SVC
    * Implementation status: https://github.com/yuzu-emu/yuzu/pull/7394  
* **CPU Emulation Level Support**
  * [x] Support Instruction Cache Clearing
[...]

PR #7394, opened 2021-11-21
kernel: svc: Implement SetProcessMemoryPermission 

Morph1984:
- Used by Skyline modding framework

See #7392 for progress on the support of this modding framework

### DIFF
--- a/src/core/hle/kernel/k_page_table.h
+++ b/src/core/hle/kernel/k_page_table.h
@@ -41,7 +41,7 @@ public:
     ResultCode MapPages(VAddr addr, KPageLinkedList& page_linked_list, KMemoryState state,
                         KMemoryPermission perm);
     ResultCode UnmapPages(VAddr addr, KPageLinkedList& page_linked_list, KMemoryState state);
-    ResultCode SetCodeMemoryPermission(VAddr addr, std::size_t size, KMemoryPermission perm);
+    ResultCode SetProcessMemoryPermission(VAddr addr, std::size_t size, KMemoryPermission perm);
     KMemoryInfo QueryInfo(VAddr addr);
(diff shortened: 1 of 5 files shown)


src/core/hle/kernel/k_page_table.h
```diff
@@ -64,6 +64,15 @@ class KPageTable final : NonCopyable {
         return page_table_impl;
     }
 
+    constexpr bool Contains(VAddr addr) {
```
lioncash, line 67:
```suggestion
    bool Contains(VAddr addr) const {
```

This can't be used in a non-constexpr context

src/core/hle/kernel/k_page_table.h
```diff
@@ -64,6 +64,15 @@ class KPageTable final : NonCopyable {
...
+        return address_space_start <= addr && addr <= address_space_end - 1;
+    }
+
+    constexpr bool Contains(VAddr addr, size_t size) {
```
lioncash, line 71: Ditto

src/core/hle/kernel/svc.cpp
```diff
@@ -1183,6 +1183,18 @@ constexpr bool IsValidRemoteSharedMemoryPermission(Svc::MemoryPermission perm) {
     return IsValidSharedMemoryPermission(perm) || perm == Svc::MemoryPermission::DontCare;
 }
 
+constexpr bool IsValidProcessMemoryPermission(Svc::MemoryPermission perm) {
```
lioncash, line 1186:
```suggestion
static constexpr bool IsValidProcessMemoryPermission(Svc::MemoryPermission perm) {
```

bunnei (approved):

PR #7368, opened 2021-11-19
Fix ART Blit detection regression and add D24S8 <-> RGBA8 conv to Vulkan

FernandoS27:
This fixes many issues in Pokemon Brilliant Diamond. Including, Vulkan & OGL crashes.

It's not in a good state to merge though.

PR #7395, opened 2021-11-21
general: Resolve comments in PR #7368

Morph1984:
Resolves some of the comments in PR #7368

### DIFF
--- a/src/video_core/renderer_vulkan/blit_image.cpp
+++ b/src/video_core/renderer_vulkan/blit_image.cpp
@@ -749,8 +749,9 @@ void BlitImageHelper::ConvertColorToDepthPipeline(vk::Pipeline& pipeline, VkRend
     });
 }
 
-void BlitImageHelper::ConvertPipelineColorTargetEx(vk::Pipeline& pipeline, VkRenderPass renderpass,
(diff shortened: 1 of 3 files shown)


Morph1984:
@FernandoS27 There are still 2 comments that need resolving:
> Looks like all usages of ConvertPipelineDepthTargetEx pass in true as the last param. Is that needed?
and
> Looks like all usages of ConvertPipelineColorTargetEx pass in false as the last param

src/video_core/renderer_vulkan/vk_buffer_cache.cpp
```diff
@@ -62,7 +62,7 @@ std::array<T, 6> MakeQuadIndices(u32 quad, u32 first) {
 }
 
 vk::Buffer CreateBuffer(const Device& device, u64 size) {
-    VkBufferUsageFlags flags =
+    static constexpr VkBufferUsageFlags flags =
```
lioncash, line 65: won't this fail to compile because the flags are ORed into below this in a conditional?

Morph1984: Nice catch, yea I did not see that.

bunnei (approved):

PR #7399, opened 2021-11-21
video_core: Refactoring post A.R.T. merge

ameerj:
The refactors a bit of the code that was introduced in #7219 and #7368. Mainly to reduce code duplication and unneeded arguments passed along between functions.

It also fixes an assertion failure that occurs when ConvertImage is called on rescaled textures.


ameerj:
supersedes #7395

FernandoS27:
This should be merged after I finish foxing the conversions.

FernandoS27 (approved):

PR #7406, opened 2021-11-22
Added TAS controls to the menu under Tools

heinermann:
This PR improves the discoverability and accessibility of TAS controls by adding them to a TAS submenu under Tools. They were previously only accessible through hotkey configuration.

2021-11-21 17_45_24-yuzu _ master-71313509f _ Metroid Dread (64-bit) _ 1 0 0 _ NVIDIA

2021-11-21 17_47_50-yuzu

#### Notes

- The alternatives to the changes in `GRenderWindow` are either to turn `Tas` into a `QObject`, or have `Tas` track a callback function.
- I think there is a bug where you can play and record a TAS at the same time, this PR is not addressing that.
- The hotkeys in the screenshots above are non-default.

### DIFF
--- a/src/yuzu/bootmanager.cpp
+++ b/src/yuzu/bootmanager.cpp
@@ -320,7 +320,7 @@ GRenderWindow::~GRenderWindow() {
 
 void GRenderWindow::OnFrameDisplayed() {
     input_subsystem->GetTas()->UpdateThread();
-    TasInput::TasState new_tas_state = std::get<0>(input_subsystem->GetTas()->GetStatus());
+    const TasInput::TasState new_tas_state = std::get<0>(input_subsystem->GetTas()->GetStatus());
 
     if (!first_frame) {
         last_tas_state = new_tas_state;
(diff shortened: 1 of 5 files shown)


src/yuzu/bootmanager.cpp
```diff
@@ -319,10 +320,18 @@ GRenderWindow::~GRenderWindow() {
 
 void GRenderWindow::OnFrameDisplayed() {
     input_subsystem->GetTas()->UpdateThread();
+    TasInput::TasState new_tas_state = std::get<0>(input_subsystem->GetTas()->GetStatus());
```
lioncash, line 323:
```suggestion
    const TasInput::TasState new_tas_state = std::get<0>(input_subsystem->GetTas()->GetStatus());
```

Since we try to make mutability explicit if possible

src/yuzu/main.cpp
```diff
@@ -2821,6 +2819,32 @@ void GMainWindow::OnConfigureTas() {
...
+    }
+    input_subsystem->GetTas()->StartStop();
+    OnTasStateChanged();
+}
+
+void GMainWindow::OnTasRecord() {
+    if (!emulation_running) {
+        return;
+    }
+    bool is_recording = input_subsystem->GetTas()->Record();
```
lioncash, line 2834:
```suggestion
    const bool is_recording = input_subsystem->GetTas()->Record();
```

lioncash:
Looks good other than the two nits 👍

german77:
Looks fine. I will test once the artifact is out. 

Recording while TAS is running is not a bug. I made it this way to be able to record my controller while TAS is running with the second player.

german77 (approved):

lioncash (approved):

PR #7407, opened 2021-11-22
Add InstructionCacheOperationRaised handler for 'ic ivau' instruction

jam1garner:
Used by Skyline modding framework, see #7392 for tracking issue on skyline support.

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic_64.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic_64.cpp
@@ -98,6 +98,9 @@ public:
             return;
 
         case Dynarmic::A64::InstructionCacheOperation::InvalidateAllToPoU:
+            parent.ClearInstructionCache();
+            return;
+
         case Dynarmic::A64::InstructionCacheOperation::InvalidateAllToPoUInnerSharable:
         default:
             LOG_DEBUG(Core_ARM, "Unprocesseed instruction cache operation");
--- a/src/core/arm/dynarmic/arm_dynarmic_64.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic_64.cpp
@@ -86,6 +86,24 @@ public:
                   num_instructions, MemoryReadCode(pc));
     }
 
+    void InstructionCacheOperationRaised(Dynarmic::A64::InstructionCacheOperation op,
+                                         VAddr value) override {
+        constexpr u64 ICACHE_LINE_SIZE = 64;
+        u64 cache_line_start;
+
+        switch (op) {
+        case Dynarmic::A64::InstructionCacheOperation::InvalidateByVAToPoU:
+            cache_line_start = value & ~(ICACHE_LINE_SIZE - 1);
+            parent.InvalidateCacheRange(cache_line_start, ICACHE_LINE_SIZE);
+            return;
+
+        case Dynarmic::A64::InstructionCacheOperation::InvalidateAllToPoU:
+        case Dynarmic::A64::InstructionCacheOperation::InvalidateAllToPoUInnerSharable:
+        default:
+            LOG_DEBUG(Core_ARM, "Unprocesseed instruction cache operation");
+        }
+    }
+
     void ExceptionRaised(u64 pc, Dynarmic::A64::Exception exception) override {
         switch (exception) {
         case Dynarmic::A64::Exception::WaitForInterrupt:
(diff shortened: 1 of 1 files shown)


src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -86,6 +86,24 @@ class DynarmicCallbacks64 : public Dynarmic::A64::UserCallbacks {
                   num_instructions, MemoryReadCode(pc));
     }
 
+    void InstructionCacheOperationRaised(Dynarmic::A64::InstructionCacheOperation op,
+                                         VAddr value) override {
+        constexpr u64 ICACHE_LINE_SIZE = 64;
+        u64 cache_line_start;
```
lioncash, line 92: This can be turned into a const variable with a narrower scope by bracing the only case statement it's used in and then joining the declaration and assignment. The variable above it can be moved into it as well.

jam1garner:
Went ahead and added `ic iallu` too, but I don't understand the semantics of `ic ialluis` to feel comfortable implementing it.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -86,6 +86,27 @@ class DynarmicCallbacks64 : public Dynarmic::A64::UserCallbacks {
...
+            parent.InvalidateCacheRange(cache_line_start, ICACHE_LINE_SIZE);
+            return;
+
+        case Dynarmic::A64::InstructionCacheOperation::InvalidateAllToPoU:
+            parent.ClearInstructionCache();
+            return;
+
+        case Dynarmic::A64::InstructionCacheOperation::InvalidateAllToPoUInnerSharable:
+        default:
+            LOG_DEBUG(Core_ARM, "Unprocesseed instruction cache operation");
```
lioncash, line 106:
```suggestion
            LOG_DEBUG(Core_ARM, "Unprocessed instruction cache operation: {}", op);
```

Just so the value of the unsupported op is available in logs if we need to care about it in the future

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -86,6 +86,27 @@ class DynarmicCallbacks64 : public Dynarmic::A64::UserCallbacks {
                   num_instructions, MemoryReadCode(pc));
     }
 
+    void InstructionCacheOperationRaised(Dynarmic::A64::InstructionCacheOperation op,
+                                         VAddr value) override {
+        switch (op) {
+        case Dynarmic::A64::InstructionCacheOperation::InvalidateByVAToPoU: {
+            constexpr u64 ICACHE_LINE_SIZE = 64;
+
+            u64 cache_line_start = value & ~(ICACHE_LINE_SIZE - 1);
```
lioncash, line 95:
```suggestion
            const u64 cache_line_start = value & ~(ICACHE_LINE_SIZE - 1);
```

Since we try to make mutability explicit

lioncash (commented):
LGTM aside from the one nit

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -86,6 +86,27 @@ class DynarmicCallbacks64 : public Dynarmic::A64::UserCallbacks {
...
+            parent.InvalidateCacheRange(cache_line_start, ICACHE_LINE_SIZE);
+            return;
```
Morph1984, line 97: Looks good, just minor comments:
- change `return` to `break`
- `break` on the default case as well
- remove the newlines separating the case(s)

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -86,6 +86,27 @@ class DynarmicCallbacks64 : public Dynarmic::A64::UserCallbacks {
                   num_instructions, MemoryReadCode(pc));
     }
 
+    void InstructionCacheOperationRaised(Dynarmic::A64::InstructionCacheOperation op,
+                                         VAddr value) override {
+        switch (op) {
+        case Dynarmic::A64::InstructionCacheOperation::InvalidateByVAToPoU: {
+            constexpr u64 ICACHE_LINE_SIZE = 64;
```
Morph1984, line 93: Should this be `static constexpr`?

jam1garner: double checked the assembly because I couldn't remember which is better in this situation and seems like you're right.

lioncash:
Thanks again for the patch

lioncash (approved):

goldenx86:

jam1garner:
Sure @goldenx86

PR #7431, opened 2021-11-25
video_core/codecs: fix multiple decoding issues on Linux

liushuyu:
This Pull Request addresses the issues in the following scenarios:

- When someone installed Intel video drivers on an AMD system, the decoder will select the Intel VA-API decoding driver and yuzu will crash due to incorrect driver selection; the fix will check if the currently about-to-use driver is loaded in the kernel.
- When using NVIDIA driver on Linux with a ffmpeg that does not have CUDA capability enabled, the decoder will crash; the fix simply makes the decoder prefers the VDPAU driver over CUDA on Linux.
- When you are using Flatpak.

Related issues:

- https://github.com/flathub/org.yuzu_emu.yuzu/issues/218
- https://github.com/flathub/org.yuzu_emu.yuzu/issues/200

### DIFF
--- a/src/video_core/command_classes/codecs/codec.cpp
+++ b/src/video_core/command_classes/codecs/codec.cpp
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <cstdio>
 #include <fstream>
 #include <vector>
 #include "common/assert.h"
@@ -59,6 +60,36 @@ Codec::~Codec() {
     av_buffer_unref(&av_gpu_decoder);
 }
 
+#ifdef LIBVA_FOUND
+// List all the currently loaded Linux modules
+static std::vector<std::string> ListLinuxKernelModules() {
+    std::vector<std::string> modules{};
+    auto module_listing = fopen("/proc/modules", "rt");
+    char* buffer = nullptr;
+    size_t buf_len = 0;
+    if (!module_listing) {
+        LOG_WARNING(Service_NVDRV, "Could not open /proc/modules to collect available modules");
+        return modules;
+    }
+    while (getline(&buffer, &buf_len, module_listing) != -1) {
+        // format for the module listing file (sysfs)
+        // <name> <module_size> <depended_by_count> <depended_by_names> <status> <load_address>
+        auto line = std::string(buffer);
+        // we are only interested in module names
+        auto name_pos = line.find_first_of(" ");
+        if (name_pos == std::string::npos) {
+            continue;
+        }
+        modules.push_back(line.erase(name_pos + 1));
+    }
+    if (buffer) {
+        free(buffer);
+    }
+    fclose(module_listing);
+    return modules;
+}
+#endif
+
 bool Codec::CreateGpuAvDevice() {
 #if defined(LIBVA_FOUND)
     static constexpr std::array<const char*, 3> VAAPI_DRIVERS = {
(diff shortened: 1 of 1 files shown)


src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -67,8 +98,21 @@ bool Codec::CreateGpuAvDevice() {
         "amdgpu",
     };
     AVDictionary* hwdevice_options = nullptr;
+    auto loaded_modules = ListLinuxKernelModules();
     av_dict_set(&hwdevice_options, "connection_type", "drm", 0);
     for (const auto& driver : VAAPI_DRIVERS) {
+        bool found = false;
```
lioncash, line 104: this loop can be collapsed into a `std::any_of` check

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -67,8 +98,21 @@ bool Codec::CreateGpuAvDevice() {
         "amdgpu",
     };
     AVDictionary* hwdevice_options = nullptr;
+    auto loaded_modules = ListLinuxKernelModules();
```
lioncash, line 101:
```suggestion
    const auto loaded_modules = ListLinuxKernelModules();
```

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -59,6 +60,36 @@ Codec::~Codec() {
     av_buffer_unref(&av_gpu_decoder);
 }
 
+#ifdef LIBVA_FOUND
+// List all the currently loaded Linux modules
+static std::vector<std::string> ListLinuxKernelModules() {
+    std::vector<std::string> modules{};
```
lioncash, line 66: A few of these variables can be moved further down closer to where they're used.

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -59,6 +60,36 @@ Codec::~Codec() {
...
+    auto module_listing = fopen("/proc/modules", "rt");
```
lioncash, line 67:
```suggestion
    using FILEPtr = std::unique_ptr<FILE, decltype(&std::fclose)>;
    auto module_listing = FILEPtr{std::fopen("/proc/modules", "rt"), std::fclose};
```

Which allows the closing to be handled automatically

src/video_core/command_classes/codecs/codec.cpp
```diff
@@ -59,6 +60,36 @@ Codec::~Codec() {
...
+        // <name> <module_size> <depended_by_count> <depended_by_names> <status> <load_address>
+        auto line = std::string(buffer);
+        // we are only interested in module names
+        auto name_pos = line.find_first_of(" ");
+        if (name_pos == std::string::npos) {
+            continue;
+        }
+        modules.push_back(line.erase(name_pos + 1));
+    }
+    if (buffer) {
```
lioncash, line 85: This check can be removed. It's well-defined to call free on a null pointer.

liushuyu:
@lioncash Thank you! I have addressed all of your comments.

liushuyu:
Ah, sorry, I misread the `std::string::erase` documentation. There was a off-by-one mistake, I have fixed it.

v1993:
I'm a bit skeptical on preferring VDPAU over CUDA. Doesn't VDPAU lack support for some codecs Yuzu needs and CUDA implements?

liushuyu:
> I'm a bit skeptical on preferring VDPAU over CUDA. Doesn't VDPAU lack support for some codecs Yuzu needs and CUDA implements?

This is a fallback order: if VDPAU does not support the codec that the game is using (VP8), then `ffmpeg` will fall back to C  UDA.

Also, most of the Linux distro does not have CUDA enabled in FFmpeg (since CUDA libraries are probably not in their "main" repository due to legal issues), and Flatpak for some reason has a broken ffmpeg that would attempt CUDA decoding when CUDA is unavailable. This fix is an attempt to work around the issue by simply adjusting the fallback order and reducing the chance of CUDA being selected.

bunnei (approved):

liushuyu:
For the record:

> ^ Uhh, I'm still not sure that was a good idea
VDPAU is not maintained by Nvidia anymore
And FFMPEG is built by yuzu anyways

VDPAU is now maintained by FD.o and most of the Linux distro has support for it.
As you can see, this Pull Request addresses issues from Flatpak mostly, where you are required to use bundled runtime (not built by Yuzu) due to portal requirements when accessing hardware from inside the Flatpak sandbox. Also, there are some Linux/BSD distributions packaging non-AppImage versions of Yuzu where they prefer system libraries over bundled ones for maintainability (and security concerns). (Reference: https://repology.org/project/yuzu/versions)

v1993:
Alright, I see. Does yuzu still crash if VDPAU does not implement required codec?

liushuyu:
> Alright, I see. Does yuzu still crash if VDPAU does not implement required codec?

According to the FLatpak version users, it will now fall back to software decoding instead of crashing (Flatpak runtime does not include CUDA support, although I am currently attempting to make Flatpak people add CUDA/NPP support).

PR #7452, opened 2021-11-28
yuzu: Implement basic controller UI navigation

german77:
This is one of the most requested features. Adds the ability to navigate through the game list, profile selector and controller applet.

The controller applet still needs some work since it's a bit more complex.

### DIFF
--- a/src/yuzu/game_list.h
+++ b/src/yuzu/game_list.h
@@ -24,6 +24,7 @@
 #include "uisettings.h"
 #include "yuzu/compatibility_list.h"
 
+class ControllerNavigation;
 class GameListWorker;
 class GameListSearchField;
 class GameListDir;
@@ -88,6 +89,9 @@ public:
     void SaveInterfaceLayout();
     void LoadInterfaceLayout();
 
+    /// Disables events from the emulated controller
+    void UnloadController();
+
     static const QStringList supported_file_extensions;
 
 signals:
@@ -143,6 +147,7 @@ private:
     QStandardItemModel* item_model = nullptr;
     GameListWorker* current_worker = nullptr;
     QFileSystemWatcher* watcher = nullptr;
+    ControllerNavigation* controller_navigation = nullptr;
     CompatibilityList compatibility_list;
 
     friend class GameListSearchField;
--- a/src/yuzu/util/controller_navigation.h
+++ b/src/yuzu/util/controller_navigation.h
@@ -1,0 +1,51 @@
+// Copyright 2021 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included
+
+#pragma once
+
+#include <QKeyEvent>
+#include <QObject>
+
+#include "common/input.h"
+#include "common/settings_input.h"
+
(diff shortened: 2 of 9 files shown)


src/yuzu/controller_navigation.cpp
```diff
@@ -0,0 +1,160 @@
...
+                                         Settings::NativeButton::Values native_button,
+                                         Qt::Key key) {
+    if (buttons[native_button].value && !button_values[native_button].locked) {
+        emit TriggerKeyboardEvent(key);
+    }
+}
+
+void ControllerNavigation::ControllerUpdateEvent(Core::HID::ControllerTriggerType type) {
+    if (type == Core::HID::ControllerTriggerType::Button) {
+        const auto controller_type = controller->GetNpadStyleIndex();
```
lioncash, line 41: Maybe move the contents of this if into a `ControllerUpdateButton()` member function and call it here, and do the same for the sticks as well. That way their behavior is kept separate from one another

src/yuzu/controller_navigation.h
```diff
@@ -0,0 +1,42 @@
...
+
+private:
+    void TriggerButton(const Core::HID::ButtonValues& buttons,
+                       Settings::NativeButton::Values native_button, Qt::Key key);
+    void ControllerUpdateEvent(Core::HID::ControllerTriggerType type);
+
+    Core::HID::ButtonValues button_values{};
+    Core::HID::SticksValues stick_values{};
+
+    int callback_key;
```
lioncash, line 39:
```suggestion
    int callback_key{};
```

even if it's overwritten in the constructor, so it has a consistent value, regardless of which phase the constructor is in for debugging purposes.

src/yuzu/controller_navigation.h
```diff
@@ -0,0 +1,42 @@
...
+// Refer to the license.txt file included
+
+#pragma once
+
+#include <QKeyEvent>
+#include <QObject>
+
+#include "common/input.h"
+#include "common/settings.h"
+#include "core/hid/emulated_controller.h"
```
Morph1984, line 12: This can be a forward declaration

src/yuzu/game_list.h
```diff
@@ -23,6 +23,7 @@
 #include "common/common_types.h"
 #include "uisettings.h"
 #include "yuzu/compatibility_list.h"
+#include "yuzu/controller_navigation.h"
```
Morph1984, line 26: This can be a forward declaration

Morph1984:
I would suggest moving controller_navigation to the `util` folder

src/yuzu/util/controller_navigation.h
```diff
@@ -0,0 +1,48 @@
...
+// Refer to the license.txt file included
+
+#pragma once
+
+#include <QKeyEvent>
+#include <QObject>
+
+#include "common/input.h"
+#include "common/settings.h"
+#include "core/hid/emulated_controller.h"
```
Morph1984, line 12: This can be a forward declaration

Morph1984 (approved):

Issue #7149, opened 2021-10-09
Metroid Dread - All videos glitch in specific room


megaboyx90:
Mainline Build - 33c453503
Title - Metroid Dread
Issue: When loading into this specific room, all video files break including: game over screen, power up menu videos, and
background videos. Issue doesn't persist when room is exited. Happens with OpenGL & Vulkan renderers.
![image](https://user-images.githubusercontent.com/45728868/136645979-6ab94fa2-1e9a-4e32-9702-86760901e34d.png)
![image](https://user-images.githubusercontent.com/45728868/136646008-64425e52-a052-4123-9628-c0babd1de688.png)
![image](https://user-images.githubusercontent.com/45728868/136646017-56c63b4a-497f-4642-8eeb-44264e14b698.png)
Specs: 
- Ryzen 5 2600
- GTX 1050 ti 
- 16gb DDR4 ram

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7315047/yuzu_log.txt)

G-Spawn:
Could you provide a save file and try auto cpu accuracy?

smiglo5:
I can confirm that
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7315117/yuzu_log.txt)

Here is the log, and here is the savefile, its the room below yellow saveroom:
[010093801237C000.zip](https://github.com/yuzu-emu/yuzu/files/7315119/010093801237C000.zip)

smiglo5:
https://user-images.githubusercontent.com/30460684/136672643-b53f35a0-1106-416a-a6d8-70cf9f859c34.mp4

Here is a video showing this bug

Hootek:
same, in cutscene too

vasishath:
This bug is also triggered on viewing any video after viewing a shuttle travelling cutscene. Even the game over screen video is affected. It can also cause the game/emulator to crash/freeze as well if u die a few times.

G-Spawn:
This should be fixed on updated yuzu. Please reopen should it still happen.

PR #7488, opened 2021-12-02
Support multiple videos playing

### LINKED ISSUE #7196: Super Robot Wars 30 DEMO Intermission Flickering
The same problem occurs in both OpenGL and Vulkan.

Displayed correctly in ryujinx
ryujinx_capture_2021-10-17_08-30-41

Version: yuzu 781
CPU: Intel(R) Core(TM) i3-8300 CPU @ 3.70GHz | AVX2 | FMA
OS: Windows 10 Version 2009
RAM: 15.94 GB
Device: NVIDIA GeForce GTX 1660 SUPER
Driver: NVIDIA 472.12.0.0
Vulkan: 1.2.175

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7358406/yuzu_log.txt)

save file
[0100136015090000.zip](https://github.com/yuzu-emu/yuzu/files/7358415/0100136015090000.zip)

vonchenplus:
Currently, when there are multiple videos that need to be played at the same time, there is video flicker. Because only one video decoder is supported internally, when there are other videos that need to be played at the same time, it causes decoder to receive video frames that are sent from different video sources.

The current solution is to generate multiple video decoders to parse different videos according to different fd.

Fixed video confusion when selecting 3d and 2d scenes when starting a new game in Dragon Quest XI.
Before:
image

After:
image

Fix #7196 
Before:
image

After:
image

There should be other similar scenes fixed.

### DIFF
--- a/src/core/hle/service/nvdrv/devices/nvhost_nvdec.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_nvdec.cpp
@@ -21,7 +21,7 @@ NvResult nvhost_nvdec::Ioctl1(DeviceFD fd, Ioctl command, const std::vector<u8>&
     case 0x0:
         switch (command.cmd) {
         case 0x1:
-            return Submit(input, output);
+            return Submit(fd, input, output);
         case 0x2:
             return GetSyncpoint(input, output);
         case 0x3:
@@ -62,11 +62,16 @@ NvResult nvhost_nvdec::Ioctl3(DeviceFD fd, Ioctl command, const std::vector<u8>&
     return NvResult::NotImplemented;
 }
 
-void nvhost_nvdec::OnOpen(DeviceFD fd) {}
+void nvhost_nvdec::OnOpen(DeviceFD fd) {
+    static u32 next_id{};
+    fd_to_id[fd] = next_id++;
+}
 
 void nvhost_nvdec::OnClose(DeviceFD fd) {
     LOG_INFO(Service_NVDRV, "NVDEC video stream ended");
-    system.GPU().ClearCdmaInstance();
+    if (fd_to_id.find(fd) != fd_to_id.end()) {
+        system.GPU().ClearCdmaInstance(fd_to_id[fd]);
+    }
 }
 
 } // namespace Service::Nvidia::Devices
--- a/src/core/hle/service/nvdrv/devices/nvhost_nvdec.cpp
+++ b/src/core/hle/service/nvdrv/devices/nvhost_nvdec.cpp
@@ -20,8 +20,12 @@ NvResult nvhost_nvdec::Ioctl1(DeviceFD fd, Ioctl command, const std::vector<u8>&
     switch (command.group) {
     case 0x0:
         switch (command.cmd) {
-        case 0x1:
+        case 0x1: {
+            if (!fd_to_id.contains(fd)) {
+                fd_to_id[fd] = next_id++;
+            }
             return Submit(fd, input, output);
+        }
         case 0x2:
             return GetSyncpoint(input, output);
         case 0x3:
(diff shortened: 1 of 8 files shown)


FernandoS27:
hmn interesting this was going to be part of YFC. It's not fully accurate but good enough.

smiglo5:
This should also fix glitched rooms and tutorial videos in Metroid Dread, great job!

bylaws:
Rather than going by FD it's probably cleaner to go by the history map address in the nvmap host1x struct, that makes it a bit more accurate too and it should be trivial to impl

src/core/hle/service/nvdrv/devices/nvhost_nvdec.cpp
```diff
@@ -62,11 +62,16 @@ NvResult nvhost_nvdec::Ioctl3(DeviceFD fd, Ioctl command, const std::vector<u8>&
     return NvResult::NotImplemented;
 }
 
-void nvhost_nvdec::OnOpen(DeviceFD fd) {}
+void nvhost_nvdec::OnOpen(DeviceFD fd) {
+    static u32 next_id{};
```
lioncash, line 66: We don't use mutable static state (this is a global variable in terms of lifetimes). Please consider a design that doesn't require this.

src/core/hle/service/nvdrv/devices/nvhost_nvdec.cpp
```diff
@@ -62,11 +62,16 @@ NvResult nvhost_nvdec::Ioctl3(DeviceFD fd, Ioctl command, const std::vector<u8>&
...
+    fd_to_id[fd] = next_id++;
+}
 
 void nvhost_nvdec::OnClose(DeviceFD fd) {
     LOG_INFO(Service_NVDRV, "NVDEC video stream ended");
-    system.GPU().ClearCdmaInstance();
+    if (fd_to_id.find(fd) != fd_to_id.end()) {
```
lioncash, line 72: This code is hashing into the unordered_map twice to retrieve the same thing. Consider storing the result of the find in a const local variable, and then dereferencing the iterator in the `ClearCdmaInstance()` call

src/core/hle/service/nvdrv/devices/nvhost_vic.cpp
```diff
@@ -62,10 +62,15 @@ NvResult nvhost_vic::Ioctl3(DeviceFD fd, Ioctl command, const std::vector<u8>& i
     return NvResult::NotImplemented;
 }
 
-void nvhost_vic::OnOpen(DeviceFD fd) {}
+void nvhost_vic::OnOpen(DeviceFD fd) {
+    static u32 next_id{};
```
lioncash, line 66: Ditto about the fact we don't use global variables anymore for service code.

src/core/hle/service/nvdrv/devices/nvhost_vic.cpp
```diff
@@ -62,10 +62,15 @@ NvResult nvhost_vic::Ioctl3(DeviceFD fd, Ioctl command, const std::vector<u8>& i
...
+    fd_to_id[fd] = next_id++;
+}
 
 void nvhost_vic::OnClose(DeviceFD fd) {
-    system.GPU().ClearCdmaInstance();
+    if (fd_to_id.find(fd) != fd_to_id.end()) {
```
lioncash, line 71: Ditto about hashing into the map twice instead of once.

src/video_core/gpu.cpp
```diff
@@ -338,25 +328,26 @@ struct GPU::Impl {
...
-        cdma_pusher->ProcessEntries(std::move(entries));
+        cdma_pushers[id]->ProcessEntries(std::move(entries));
     }
 
     /// Frees the CDMAPusher instance to free up resources
-    void ClearCdmaInstance() {
-        cdma_pusher.reset();
+    void ClearCdmaInstance(u32 id) {
+        if (cdma_pushers.find(id) != cdma_pushers.end()) {
+            cdma_pushers.erase(id);
```
lioncash, line 349: You can store the result of the find into a const local variable and pass the iterator directly to `erase()`, eliminating a second lookup into the map.

src/video_core/gpu.cpp
```diff
@@ -338,25 +328,26 @@ struct GPU::Impl {
...
     /// Push GPU command buffer entries to be processed
-    void PushCommandBuffer(Tegra::ChCommandHeaderList& entries) {
+    void PushCommandBuffer(u32 id, Tegra::ChCommandHeaderList& entries) {
         if (!use_nvdec) {
             return;
         }
 
-        if (!cdma_pusher) {
-            cdma_pusher = std::make_unique<Tegra::CDmaPusher>(gpu);
+        if (cdma_pushers.find(id) == cdma_pushers.end()) {
```
lioncash, line 336:
```suggestion
        if (!cdma_pushers.contains(id)) {
            cdma_pushers.insert_or_assign(id, std::make_unique<Tegra::CDmaPusher>(gpu));
        }
```

bunnei:
Untagged, seems to break Metroid Dread videos

vonchenplus:
> Untagged, seems to break Metroid Dread videos

Any specific information?

german77:
yuzu_625gq3pnI6

vonchenplus:
Fixed Metroid Dread videos

image

gidoly:
fixes #7149

bunnei (approved):

PR #7503, opened 2021-12-05
core/hid: Ensure only valid npad are connected

### LINKED ISSUE #7500: Ring fit adventure screen goes black at the beginning of the game
After the game is loaded, you can hear the footsteps that should be emitted when entering the game, but the screen is always black.

log
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7654545/yuzu_log.txt)

Thanks and best wishes

german77:
SwapNpadAssignment generated a SVC break when it tried to swap a controller that is invalid.
Fixes #7500

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -866,7 +866,50 @@ void EmulatedController::SetLedPattern() {
     }
 }
 
+void EmulatedController::SetSupportedNpadStyleTag(NpadStyleTag supported_styles) {
+    supported_style_tag = supported_styles;
+    if (!is_connected) {
+        return;
+    }
+    if (!IsControllerSupported()) {
+        LOG_ERROR(Service_HID, "Controller type {} is not supported. Disconnecting controller",
+                  npad_type);
+        Disconnect();
+    }
+}
+
+bool EmulatedController::IsControllerSupported() const {
+    switch (npad_type) {
+    case NpadStyleIndex::ProController:
+        return supported_style_tag.fullkey;
+    case NpadStyleIndex::JoyconDual:
+        return supported_style_tag.joycon_dual;
+    case NpadStyleIndex::JoyconLeft:
+        return supported_style_tag.joycon_left;
+    case NpadStyleIndex::JoyconRight:
+        return supported_style_tag.joycon_right;
+    case NpadStyleIndex::GameCube:
+        return supported_style_tag.gamecube;
+    case NpadStyleIndex::Pokeball:
+        return supported_style_tag.palma;
+    case NpadStyleIndex::NES:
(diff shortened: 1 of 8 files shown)


src/core/hid/hid_types.h
```diff
@@ -256,6 +256,8 @@ enum class NpadStyleSet : u32 {
     Lager = 1U << 11,
     SystemExt = 1U << 29,
     System = 1U << 30,
+    All = Fullkey + Handheld + JoyDual + JoyLeft + JoyRight + Gc + Palma + Lark + HandheldLark +
```
lioncash, line 259:
```suggestion
    All = Fullkey + Handheld | JoyDual | JoyLeft | JoyRight | Gc | Palma | Lark | HandheldLark |
```

etc for the other identifiers on the line below

src/core/hid/emulated_controller.cpp
```diff
@@ -866,7 +866,49 @@ void EmulatedController::SetLedPattern() {
...
+    }
+    if (!IsControllerSupported()) {
+        LOG_ERROR(Service_HID, "Controller type is not supported. Disconnecting controller");
+        Disconnect();
+    }
+}
+
+bool EmulatedController::IsControllerSupported() const {
+    switch (npad_type) {
+    case Core::HID::NpadStyleIndex::ProController:
```
lioncash, line 882:
```suggestion
    case NpadStyleIndex::ProController:
```

Given this class is in the same namespace. Ditto for the other entries in the switch

src/core/hid/emulated_controller.cpp
```diff
@@ -866,7 +866,49 @@ void EmulatedController::SetLedPattern() {
...
+    case Core::HID::NpadStyleIndex::SegaGenesis:
+        return supported_style_tag.lager;
+    default:
+        return false;
+    }
+}
+
 void EmulatedController::Connect() {
+    if (!IsControllerSupported()) {
+        LOG_ERROR(Service_HID, "Controller type is not supported");
```
lioncash, line 909: This should likely log out the controller type, so it gives useful information in the log.

src/core/hid/hid_types.h
```diff
@@ -256,6 +256,8 @@ enum class NpadStyleSet : u32 {
     Lager = 1U << 11,
     SystemExt = 1U << 29,
     System = 1U << 30,
+    All = Fullkey | Handheld | JoyDual | JoyLeft | JoyRight | Gc | Palma | Lark | HandheldLark |
```
Morph1984, line 259: I would just set this to `0xFFFFFFFFU` instead, so you don't need to update this if new controllers are added

bunnei (approved):

PR #7551, opened 2021-12-10
Fix blit image/view not compatible

vonchenplus:
This change fixes the incompatibility image/view between compressed and uncompressed textures format when blit image.

This should be a regression introduced by several recent texture cache changes.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -1137,8 +1137,13 @@ typename TextureCache<P>::BlitImages TextureCache<P>::GetBlitImages(
     } while (has_deleted_images);
     const ImageBase& src_image = slot_images[src_id];
     const ImageBase& dst_image = slot_images[dst_id];
+    const bool native_bgr = runtime.HasNativeBgr();
     if (GetFormatType(dst_info.format) != GetFormatType(dst_image.info.format) ||
-        GetFormatType(src_info.format) != GetFormatType(src_image.info.format)) {
+        GetFormatType(src_info.format) != GetFormatType(src_image.info.format) ||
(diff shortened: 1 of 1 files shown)


FernandoS27:
This is overboard, if the view formats don't match there's a bug somewhere else.

vonchenplus:
Yes, very much agree with you. It is possible that several recent modifications have caused the issue to be exposed.

FernandoS27:
> Yes, very much agree with you. It is possible that several recent modifications have caused the issue to be exposed.
> 
> In fact there are quite a few games that have this problem. For example Dragon Quest XI、Taxi Chaos and so on.

could you DM me what they are printing?

vonchenplus:
> > Yes, very much agree with you. It is possible that several recent modifications have caused the issue to be exposed.
> > In fact there are quite a few games that have this problem. For example Dragon Quest XI、Taxi Chaos and so on.
> 
> could you DM me what they are printing?

```
[ 107.474212] Debug <Critical> video_core\texture_cache\image_view_base.cpp:operator ():29: Assertion Failed!
Image view format R32G32B32A32_FLOAT is incompatible with image format ASTC_2D_6X6_UNORM
[ 107.478741] Debug <Critical> video_core\texture_cache\image_view_base.cpp:operator ():29: Assertion Failed!
Image view format R32G32B32A32_FLOAT is incompatible with image format ASTC_2D_6X6_SRGB
[ 107.484147] Debug <Critical> video_core\texture_cache\image_view_base.cpp:operator ():29: Assertion Failed!
Image view format R32G32B32A32_FLOAT is incompatible with image format BC5_UNORM
```

FernandoS27:
> 

oh that's new. It's trying to blit into a compressed format, specially ASTC, damn.

bunnei (approved):

PR #7558, opened 2021-12-11
common/cpu_detect: Remove CPU family and model

Morph1984:
We currently do not make use of these fields, remove them for now.


bunnei:
I'd vote for just removing -- we can always use history if we would like to add this back.

Morph1984:
Removed

lioncash (approved):

PR #7593, opened 2021-12-16
core/hid: Cancel any vibration after the test

german77:
Some games test the device rumble all the time. Resulting in an unpleasant continuous rumble. To avoid this we set the amplitude to zero after the test

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -843,23 +843,18 @@ bool EmulatedController::SetVibration(std::size_t device_index, VibrationValue v
 }
 
 bool EmulatedController::TestVibration(std::size_t device_index) {
-    if (device_index >= output_devices.size()) {
-        return false;
-    }
-    if (!output_devices[device_index]) {
-        return false;
-    }
-
-    // Send a slight vibration to test for rumble support
-    constexpr Common::Input::VibrationStatus status = {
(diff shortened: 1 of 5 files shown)


src/core/hid/emulated_controller.cpp
```diff
@@ -843,23 +843,18 @@ bool EmulatedController::SetVibration(std::size_t device_index, VibrationValue v
...
-    if (device_index >= output_devices.size()) {
-        return false;
-    }
-    if (!output_devices[device_index]) {
-        return false;
-    }
-
-    // Send a slight vibration to test for rumble support
-    constexpr Common::Input::VibrationStatus status = {
+    constexpr VibrationValue test_vibration = {
```
Morph1984, line 846:
```suggestion
    static constexpr VibrationValue test_vibration = {
```

src/core/hid/hid_types.h
```diff
@@ -496,6 +496,13 @@ struct VibrationValue {
 };
 static_assert(sizeof(VibrationValue) == 0x10, "VibrationValue has incorrect size.");
 
+static constexpr VibrationValue DEFAULT_VIBRATION_VALUE{
```
Morph1984, line 499:
```suggestion
constexpr VibrationValue DEFAULT_VIBRATION_VALUE{
```

Morph1984 (approved):

PR #7596, opened 2021-12-16
externals/CMakeLists: fix detection/init of Switch controllers in SDL 2.0.18

Tatsh:
The latest Linux builds do not see my Switch Pro Controller. @v1993 also confirms that joycons are not seen. It shows ` input_common/drivers/sdl_driver.cpp:InitJoystick:287: Failed to open joystick 0`.

Enable `SDL_THREADS` and `SDL_ATOMIC`. They are now required explicitly since 2.0.18  to have HIDAPI and libusb work.

Set `SDL_WAYLAND=OFF` due to [build issue](https://github.com/libsdl-org/SDL/commit/e2ade2bfc46d915cd306c63c830b81d800b2575f) (patch is not available in 2.0.18).

[Related issue](https://github.com/libsdl-org/SDL/issues/5095)

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -52,12 +52,12 @@ endif()
 # SDL2
 if (YUZU_USE_EXTERNAL_SDL2)
     if (NOT WIN32)
-        # Yuzu itself needs: Events Joystick Haptic Sensor Timers Audio Threads Atomic
-        # Since 2.0.18 Threads/Atomic required for HIDAPI/libusb (see https://github.com/libsdl-org/SDL/issues/5095)
+        # Yuzu itself needs: Atomic Audio Events Joystick Haptic Sensor Threads Timers
+        # Since 2.0.18 Atomic+Threads required for HIDAPI/libusb (see https://github.com/libsdl-org/SDL/issues/5095)
         # Yuzu-cmd also needs: Video (depends on Loadso/Dlopen)
         set(SDL_UNUSED_SUBSYSTEMS
-            Render Power File CPUinfo
-            Filesystem Locale)
+            CPUinfo File Filesystem
+            Locale Power Render)
         foreach(_SUB ${SDL_UNUSED_SUBSYSTEMS})
           string(TOUPPER ${_SUB} _OPT)
           option(SDL_${_OPT} "" OFF)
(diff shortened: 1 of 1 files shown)


german77:
Does this removes SDL support to wayland? I would rather update SDL to that commit rather than adding a workaround.

Tatsh:
I can add that to this PR if you want, or just not do the Wayland thing here.

Tatsh:
Let me test my build against that commit and I'll report back.

Tatsh:
Works as far as I can tell.

https://github.com/libsdl-org/SDL/commit/e2ade2bfc46d915cd306c63c830b81d800b2575f with the following:

* SMO, SM3DW
* Tested motion controls including shaking and accelerometer
* Pro Controller + joycond + hid_nintendo
* The device can be utilised by joycond when Yuzu exits without having to re-attach (was not the case with SDL 2.0.16)

german77 (approved):

Tatsh:
Technically that version of SDL is 2.0.19. I wasn't sure if you wanted me to update that in the root CMakeLists.

```
 $ ag '2\.0\.18' CMakeLists.txt
373:            set(SDL2_VER "SDL2-2.0.18")
393:        find_package(SDL2 2.0.18 REQUIRED)
```

This will only affect those who wish to build against a system copy and the Windows builder when it looks for binaries.

german77:
Windows uses precompiled binaries and those are set to 2.0.18. Since there isn't any meaningful change to windows and this commit is very close to 2.0.18. I don't see why they will need to be updated.

v1993:
Are you sure there's no change to windows? Lack of thread support might as well be causing some odd issues there, unless prebuilt libraries aren't using options passed in this build.

externals/CMakeLists.txt
```diff
@@ -52,11 +52,12 @@ endif()
 # SDL2
 if (YUZU_USE_EXTERNAL_SDL2)
     if (NOT WIN32)
-        # Yuzu itself needs: Events Joystick Haptic Sensor Timers Audio
+        # Yuzu itself needs: Events Joystick Haptic Sensor Timers Audio Threads Atomic
```
v1993, line 55: A nitpick, but - this was supposed to be an alphabetically sorted list, and (I think) so was one in `SDL_UNUSED_SUBSYSTEMS`. It would be nice to have it properly sorted again.

german77:
The windows builds use the default settings from SDL. So it should have threads already on

bunnei (approved):

bunnei:
LGTM, thank you!

PR #7608, opened 2021-12-21
Allow overriding SCM version info

Tatsh:
If the build is from a non-repository (like an archive), the `git_*` functions will return empty. This patch allows using define arguments to CMake to set version info such as `-DGIT_BRANCH=master`.


bunnei (approved):

abouvier:
These functions will fail too, even in a git repository, when building out of tree because `src/common/CMakeLists.txt` will run `CMakeModules/GenerateSCMRev.cmake` from the build directory instead of the source directory (where `.git/` is found).
Currently the `.git/` directory must be copied inside the future build directory before calling cmake to fix that.

> When run in -P script mode, CMake sets the variables CMAKE_BINARY_DIR, CMAKE_SOURCE_DIR, CMAKE_CURRENT_BINARY_DIR and CMAKE_CURRENT_SOURCE_DIR to the current working directory.

PR #7610, opened 2021-12-21
Add support for linking against system Opus

Tatsh:
Adds support for the following system libs with CMake defines:

* opus - `YUZU_USE_BUNDLED_OPUS=OFF`

~~I have been using this patch since about the beginning of this year with over 99 different builds spanning every yuzu-mainline tag since then. I have not experienced any issues using system libs for these vs the bundled.~~

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -116,8 +116,8 @@ if (ENABLE_WEB_SERVICE)
 endif()
 
 # Opus
-find_package(opus 1.3)
-if (NOT opus_FOUND)
-    message(STATUS "opus 1.3 or newer not found, falling back to externals")
+if (YUZU_USE_BUNDLED_OPUS)
     add_subdirectory(opus EXCLUDE_FROM_ALL)
+else()
+    find_package(opus 1.3 REQUIRED)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ option(ENABLE_CUBEB "Enables the cubeb audio backend" ON)
 
 option(USE_DISCORD_PRESENCE "Enables Discord Rich Presence" OFF)
 
+option(YUZU_USE_BUNDLED_OPUS "Compile bundled opus" ON)
+
 # Default to a Release build
 get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
 if (NOT IS_MULTI_CONFIG AND NOT CMAKE_BUILD_TYPE)


src/core/CMakeLists.txt
```diff
@@ -721,3 +721,8 @@ if (ARCHITECTURE_x86_64)
     )
     target_link_libraries(core PRIVATE dynarmic)
 endif()
+
+if (USE_SYSTEM_OPUS)
+    find_path(OPUS_H_PATH opus.h PATH_SUFFIXES opus)
+    target_include_directories(core PUBLIC ${OPUS_H_PATH})
+endif()
```
liushuyu, line 728: This is strange. Why a search of `opus.h` is needed? I would guess there are some distros place `opus.h` in a different location?

externals/CMakeLists.txt
```diff
@@ -8,8 +9,12 @@ include(DownloadExternals)
 if (ARCHITECTURE_x86 OR ARCHITECTURE_x86_64)
     add_library(xbyak INTERFACE)
     file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/xbyak/include)
-    file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/xbyak/xbyak DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/xbyak/include)
-    target_include_directories(xbyak SYSTEM INTERFACE ${CMAKE_CURRENT_BINARY_DIR}/xbyak/include)
+    if (USE_SYSTEM_XBYAK)
+        target_include_directories(xbyak SYSTEM INTERFACE /usr/include/xbyak)
```
liushuyu, line 13: Instead of hardcoding `/usr/include/xbyak` I think it's probably better to use `find_path` like you already did in the later sections.

liushuyu (commented):
I would guess this change is for packaging purposes? If so, I think we could add another option like `ENABLE_PACKAGING` or something like that to conveniently switch on all the `ENABLE_SYSTEM_*` variables.

Tatsh:
> I would guess this change is for packaging purposes? If so, I think we could add another option like `ENABLE_PACKAGING` or something like that to conveniently switch on all the `ENABLE_SYSTEM_*` variables.

Yes this is for packaging. My ebuild (Gentoo) uses these patches: https://github.com/Tatsh/tatsh-overlay/blob/master/games-emulation/yuzu/yuzu-0.853.ebuild

src/core/CMakeLists.txt
Tatsh: Yes. At some point the search didn't work in the past without these lines. Maybe now `FindOpus` works to set the include path. My `opus.h` is at `/usr/include/opus/opus.h`.

Tatsh: Removed as it is not necessary.

Tatsh:
Also we could do the flag to enable the system package options but I will have to keep using them separately because of SDL on Gentoo.

Yuzu should only be built with SDL 2.0.18 (on Linux) and even then many system versions of SDL may lack the proper libusb/hidapi support (which makes Nintendo controllers' motion work properly). This is the case on Gentoo.

liushuyu:
> > I would guess this change is for packaging purposes? If so, I think we could add another option like `ENABLE_PACKAGING` or something like that to conveniently switch on all the `ENABLE_SYSTEM_*` variables.
> 
> Yes this is for packaging. My ebuild (Gentoo) uses these patches: https://github.com/Tatsh/tatsh-overlay/blob/master/games-emulation/yuzu/yuzu-0.853.ebuild

Ah, I see.

Although I am not a Gentoo user myself, I still would like to make a few suggestions:

- Probably add `USE` flags for `discord-presence` and `compatibility-reporting` since some people may want to enable Discord integrations or disable compatibility reporting. Note that `compatibility-reporting` depends on the `web-service` flag.
- The `use-cubeb` flag is a bit questionable: `cubeb` will still be enabled regardless, just not using the system library.
- In the `DEPEND` section, `ffmpeg` needs to be version `>=4.3` for better NVDEC emulation.

Tatsh:
> > > I would guess this change is for packaging purposes? If so, I think we could add another option like `ENABLE_PACKAGING` or something like that to conveniently switch on all the `ENABLE_SYSTEM_*` variables.
> > 
> > 
> > Yes this is for packaging. My ebuild (Gentoo) uses these patches: https://github.com/Tatsh/tatsh-overlay/blob/master/games-emulation/yuzu/yuzu-0.853.ebuild
> 
> Ah, I see.
> 
> Although I am not a Gentoo user myself, I still would like to make a few suggestions:
> 
> * Probably add `USE` flags for `discord-presence` and `compatibility-reporting` since some people may want to enable Discord integrations or disable compatibility reporting. Note that `compatibility-reporting` depends on the `web-service` flag.
> * The `use-cubeb` flag is a bit questionable: `cubeb` will still be enabled regardless, just not using the system library.
> * In the `DEPEND` section, `ffmpeg` needs to be version `>=4.3` for better NVDEC emulation.

Thanks for the tips. I removed the cubeb USE and made cubeb a hard requirement. I also set the constraint on ffmpeg version.

I will add compatibility-reporting.

As for Discord I guess it can be done but nobody using my repo has asked for it. I avoided using it because the discord-rpc project is deprecated and has been for ~2 years.

liushuyu:
> Thanks for the tips. I removed the cubeb USE and made cubeb a hard requirement. I also set the constraint on ffmpeg version.
> 
> I will add compatibility-reporting.
> 
> As for Discord I guess it can be done but nobody using my repo has asked for it. I avoided using it because the discord-rpc project is deprecated and has been for ~2 years.

Thank you! In which case I think that will do. Thank again for your help, I don't have merge permissions in this repo, but I will inform other team members about this.

Tatsh:
Do you know why the CI skipped in https://github.com/yuzu-emu/yuzu/pull/7608 ?

liushuyu:
> Do you know why the CI skipped in #7608 ?

Seems like a CI bug, I have restarted the check for you.

Tatsh:
Thanks.

Tachi107:
My comment on #6833 also applies here: https://github.com/yuzu-emu/yuzu/pull/6833#discussion_r767850622

Tatsh:
I don't want to step on someone else's work. It looks like #6833 would do most of the same things and I like how it uses aliases.

I think I will just make this PR about using the system version of Opus since #6833 does not cover that.

Tatsh:
With regards to tiny libraries like inih and xbyak, I think it would be okay to not use those as submodules and instead copy the content directly into the project. This would make the source archive already contain the files. Tiny libraries are unlikely to need any changes but you do gain that flexibility.

liushuyu (approved):

bunnei (approved):

PR #7618, opened 2021-12-22
Increase boost requirement to 1.78.0

goldenx86:
Liu's finding, this allows to build yuzu on VS 2022.
Ignore at will.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,7 +229,7 @@ elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR YUZU_USE_BUNDLED_BOOST)
     include_directories(SYSTEM "${Boost_INCLUDE_DIRS}")
 else()
     message(STATUS "Boost 1.73.0 or newer not found, falling back to Conan")
-    list(APPEND CONAN_REQUIRED_LIBS "boost/1.73.0")
+    list(APPEND CONAN_REQUIRED_LIBS "boost/1.78.0")
 endif()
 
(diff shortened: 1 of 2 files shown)


german77:
Did Conan added support to 1.78.0?

neico:
If you use the new https://center.conan.io as remote, then yes.

FearlessTobi:
Will this cause any problems when other compilers are using different boost versions? 
It seems like we're using 1.73, 1.75 and 1.78 at the same time now.

german77:
I noticed that as well. We don't use any new feature. But 1.78 is required to compile in VS2022. Maybe update requirements to only above 1.73 with default at 1.78

liushuyu:
> Will this cause any problems when other compilers are using different boost versions? It seems like we're using 1.73, 1.75 and 1.78 at the same time now.

You can find a compiler compatibility list at the bottom of this page:
https://www.boost.org/users/history/version_1_78_0.html

bunnei:
`##[warning]cl(0,0): Warning D9025: overriding '/Ob2' with '/Ob3'`

Looks like the warning settings are bleeding into our build?

`##[error]src\shader_recompiler\backend\spirv\emit_spirv.cpp(35,0): Error C2220: the following warning is treated as an error`

... Untagging again.

goldenx86:
Conflicts somehow with Early Access, clossing.

goldenx86:
Boss says to reopen, we reopen.

bunnei (approved):

PR #7624, opened 2021-12-24
vk_texture_cache: Use 3D scale helpers for MSAA texture scaling on Intel Windows drivers

ameerj:
Fixes a crash when scaling MSAA textures in titles such as Sonic Colors Ultimate.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_texture_cache.h
+++ b/src/video_core/renderer_vulkan/vk_texture_cache.h
@@ -148,6 +148,8 @@ public:
 private:
     bool BlitScaleHelper(bool scale_up);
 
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_vulkan/vk_texture_cache.h
```diff
@@ -148,6 +148,8 @@ class Image : public VideoCommon::ImageBase {
 private:
     bool BlitScaleHelper(bool scale_up);
 
+    bool NeedsScaleHelper();
```
lioncash, line 151:
```suggestion
    bool NeedsScaleHelper() const;
```

 perhaps?

PR #7629, opened 2021-12-25
shaders: Add fixes for NVIDIA drivers 495+

ameerj:
After running the new driver through our suite of [unit tests](https://github.com/ReinUsesLisp/nxgpucatch), The two main issues encountered by the newer NVIDIA drivers are negation of floating point values, and bitwise conversions of input attributes. This PR addresses both.

When negating floating point values, prior drivers allowed the following usage:
```glsl
float f1 = 42.0f;
float f2 = -f1;    // f2 equals -42.0f
```
However, the new drivers ignored negation in these cases. The work around is to simply subtract the value from `0.0` to achieve the same result.

The bitwise conversion bug was more subtle. Due to the `GetAttribute` OpCode's return value being an F32, integer attribute values are bitwise converted to F32's, which are then converted back to U32 values when they are being used, resulting in the following pattern:
```glsl
float f1 = intBitsToFloat(gl_VertexID);
uint instance_id = floatBitsToUint(f1);
```

On the new drivers, `instance_id` was always 0 for the integer shader inputs as tested in GLSL. The fix for this was to simply collapse the attribute retrieval to simply be `uint instance_id = uint(gl_VertexID);`. 

These issues mainly affected the GLSL backend, but the bitwise conversion fix has been added as an optimization pass within the IR and is applied to all backends as a result.

Please test the new driver thoroughly and report any more regressions with reproduction steps on all backends.

EDIT:
Another negation related bug, for integers. This was an edge case in some shaders:
```glsl
// cbuf[0] = 1.0f;
uint u0 = uint(cbuf[0]);  // u0 = 1
uint u1 = -u0;            // u1 = 0xFFFFFFFF
u1 = u1 << 1u             // u1 should equal 0xFFFFFFFE. instead it equals 2.
```
This was resolved by negating the integers in a similar fashion to the floating point numbers.

The `floatBitsToUint` bug seems to be that when converting a `float` value to a `uint` variable, TRUNC.U is emitted, which truncates the floating point value before storing into the uint value.

### DIFF
--- a/src/shader_recompiler/backend/glasm/emit_glasm_context_get_set.cpp
+++ b/src/shader_recompiler/backend/glasm/emit_glasm_context_get_set.cpp
@@ -126,6 +126,22 @@ void EmitGetAttribute(EmitContext& ctx, IR::Inst& inst, IR::Attribute attr, Scal
     }
 }
 
+void EmitGetAttributeU32(EmitContext& ctx, IR::Inst& inst, IR::Attribute attr, ScalarU32) {
+    switch (attr) {
+    case IR::Attribute::PrimitiveId:
+        ctx.Add("MOV.S {}.x,primitive.id;", inst);
+        break;
+    case IR::Attribute::InstanceId:
+        ctx.Add("MOV.S {}.x,{}.instance;", inst, ctx.attrib_name);
+        break;
+    case IR::Attribute::VertexId:
+        ctx.Add("MOV.S {}.x,{}.id;", inst, ctx.attrib_name);
+        break;
+    default:
+        throw NotImplementedException("Get U32 attribute {}", attr);
+    }
+}
+
 void EmitSetAttribute(EmitContext& ctx, IR::Attribute attr, ScalarF32 value,
                       [[maybe_unused]] ScalarU32 vertex) {
     const u32 element{static_cast<u32>(attr) % 4};
--- a/src/shader_recompiler/profile.h
+++ b/src/shader_recompiler/profile.h
@@ -65,6 +65,8 @@ struct Profile {
     bool has_gl_component_indexing_bug{};
     /// The precise type qualifier is broken in the fragment stage of some drivers
     bool has_gl_precise_bug{};
+    /// Some drivers do not properly support floatBitsToUint when used on cbufs
+    bool has_gl_cbuf_ftou_bug{};
     /// Ignores SPIR-V ordered vs unordered using GLSL semantics
     bool ignore_nan_fp_comparisons{};
 
--- a/src/shader_recompiler/profile.h
+++ b/src/shader_recompiler/profile.h
@@ -67,6 +67,8 @@ struct Profile {
     bool has_gl_precise_bug{};
     /// Some drivers do not properly support floatBitsToUint when used on cbufs
     bool has_gl_cbuf_ftou_bug{};
+    /// Some drivers poorly optimize boolean variable references
+    bool has_gl_bool_ref_bug{};
     /// Ignores SPIR-V ordered vs unordered using GLSL semantics
     bool ignore_nan_fp_comparisons{};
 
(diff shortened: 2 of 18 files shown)


Tatsh:
I'll be trying this out later today.

gidoly:
yuzu_kiyCIemeGk
aoc loading screen renders dark atm with last driver

gidoly:
https://user-images.githubusercontent.com/66776795/147382746-eb6e9d30-b7ef-45a7-b93d-b1b833fa2028.mp4

broken filter on astralchin it looks blocky atm

gidoly:
glsl
 botw icon broken on pouch screen
yuzu_Z3lGLIvvKI

gidoly:
glsl links awakening
yuzu_dQQ6WXQeQU
broken over world gfx

gidoly:
glsl pokemon sword broken
yuzu_lgS5hZnewM

gidoly:
glsl luigi mansion
yuzu_ibA4JFTTmo

gidoly:
glsl smtv
yuzu_GNcvmXod2r

gidoly:
glsl

https://user-images.githubusercontent.com/66776795/147383205-55b035fd-71e5-453b-9f7b-b81493fd3548.mp4

broken grass on xenoblade titles

DonelBueno:
I'm also getting tons of regressions with this PR.

src/shader_recompiler/backend/glasm/emit_glasm_context_get_set.cpp
```diff
@@ -126,6 +126,22 @@ void EmitGetAttribute(EmitContext& ctx, IR::Inst& inst, IR::Attribute attr, Scal
...
+        ctx.Add("MOV.S {}.x,primitive.id;", inst);
+        break;
+    case IR::Attribute::InstanceId:
+        ctx.Add("MOV.S {}.x,{}.instance;", inst, ctx.attrib_name);
+        break;
+    case IR::Attribute::VertexId:
+        ctx.Add("MOV.S {}.x,{}.id;", inst, ctx.attrib_name);
+        break;
+    default:
+        throw NotImplementedException("Get attribute {}", attr);
```
v1993, line 141: Probably better update error message in order to differentiate between `EmitGetAttributeU32` and `EmitGetAttribute` failures.

Ditto for other backends

v1993 (changes requested):

gidoly:
> I'm also getting tons of regressions with this PR (Nvidia 497.29).

This pr removes auto detection of nvidia driver version
Which switch to glasm if you use recent driver. So it is glsl nvidia driver issue rather then regression caused by this pr

gidoly:
while emit_glsl_integer: Use negation work around commit fixes xenoblade grass it breaks brightness of title screen 

yuzu_SyTEvvQ8nV

save:
[bfsgame00.zip](https://github.com/yuzu-emu/yuzu/files/7776387/bfsgame00.zip)

gidoly:
additional report kirby star allies not render character models correctly

yuzu_U1JbzqwzDv

yuzu_r4ECG054W7

LordVulkan:
Hi, did you report this bug to NVIDIA? I have this strange feeling that they don't give a fuck about their OpenGL driver anymore, it has been broken for 2 months and they don't even recognize this as a bug in their release notes.

FernandoS27:
> Hi, did you report this bug to NVIDIA? I have this strange feeling that they don't give a fuck about their OpenGL driver anymore, it has been broken for 2 months and they don't even recognize this as a bug in their release notes.

They are fully aware but it seems it's hard for them to find it for some reason.

src/shader_recompiler/profile.h
```diff
@@ -65,6 +65,8 @@ struct Profile {
     bool has_gl_component_indexing_bug{};
     /// The precise type qualifier is broken in the fragment stage of some drivers
     bool has_gl_precise_bug{};
+    /// Some drivers do not properly support floatBitsToUint when used on cbufs
```
v1993, line 68: This comment's structure seems to be somewhat different to surrounding ones

gidoly:
new issue found on glsl backend with last driver

glsl

yuzu_ppxBnITjew

vulkan

yuzu_lsTujcIayw

PR #7648, opened 2021-12-30
core: hle: kernel: Implement thread pinning.

bunnei:
- We largely had the mechanics in place for thread pinning, this change hooks these up.
- Validated with tests https://github.com/Atmosphere-NX/Atmosphere/blob/master/tests/TestSvc/source/test_thread_pinning.cpp.

### DIFF
--- a/src/core/hle/kernel/k_interrupt_manager.h
+++ b/src/core/hle/kernel/k_interrupt_manager.h
@@ -1,0 +1,17 @@
+// Copyright 2021 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include "common/common_types.h"
+
+namespace Kernel {
+
+class KernelCore;
+
+namespace KInterruptManager {
+void HandleInterrupt(KernelCore& kernel, s32 core_id);
+}
+
+} // namespace Kernel
--- a/src/core/hle/kernel/k_process.cpp
+++ b/src/core/hle/kernel/k_process.cpp
@@ -220,30 +220,28 @@ bool KProcess::ReleaseUserException(KThread* thread) {
     }
 }
 
-void KProcess::PinCurrentThread() {
+void KProcess::PinCurrentThread(s32 core_id) {
     ASSERT(kernel.GlobalSchedulerContext().IsLocked());
 
     // Get the current thread.
-    const s32 core_id = GetCurrentCoreId(kernel);
-    KThread* cur_thread = GetCurrentThreadPointer(kernel);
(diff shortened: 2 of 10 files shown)


src/core/hle/kernel/k_interrupt_manager.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include "common/common_types.h"
+
+namespace Kernel {
+
+class KernelCore;
+
+namespace KInterruptManager {
+void HandleInterrupt(KernelCore& kernel, s32 core_id);
+};
```
lioncash, line 15:
```suggestion
}
```

src/core/hle/kernel/k_thread.cpp
```diff
@@ -63,6 +64,13 @@ namespace Kernel {
 
 namespace {
 
+struct ThreadLocalRegion {
+    static constexpr std::size_t MessageBufferSize = 0x100;
+    std::array<u32, MessageBufferSize / sizeof(u32)> message_buffer;
+    volatile u16 disable_count;
```
lioncash, line 70: These should be atomics rather than volatile variables, since we're not accessing any memory that we don't ultimately control (such as MMIO or anything). Doubly so if they're used in a threading setting, since volatile != thread-safe.

src/core/hle/kernel/k_scheduler.cpp
```diff
@@ -53,6 +54,13 @@ void KScheduler::RescheduleCores(KernelCore& kernel, u64 cores_pending_reschedul
         }
         cores_pending_reschedule &= ~(1ULL << core);
     }
+
+    for (std::size_t core = 0; core < Core::Hardware::NUM_CPU_CORES; ++core) {
+        if (kernel.PhysicalCore(core).IsInterrupted()) {
+            KInterruptManager::HandleInterrupt(kernel, core);
```
lioncash, line 60: An implicit sign conversion is happening here, since the parameter taking `core` is signed. Maybe make the cast explicit here.

src/core/hle/kernel/k_process.cpp
```diff
@@ -220,30 +220,28 @@ bool KProcess::ReleaseUserException(KThread* thread) {
...
-    const s32 core_id = GetCurrentCoreId(kernel);
-    KThread* cur_thread = GetCurrentThreadPointer(kernel);
+    KThread* cur_thread = kernel.Scheduler(core_id).GetCurrentThread();
 
     // If the thread isn't terminated, pin it.
     if (!cur_thread->IsTerminationRequested()) {
         // Pin it.
         PinThread(core_id, cur_thread);
-        cur_thread->Pin();
+        cur_thread->Pin(core_id);
```
lioncash, line 233: This will result in an implicit sign conversion, since the parameter expects an s32.

PR #7658, opened 2022-01-01
video_core/memory_manager: Fixes for sparse memory management

ameerj:
This fixes a few oversights in memory management for sparse (non-contiguous) GPU backed memory. 

The main issue fixed was in `GetSubmappedRange` which did not account for page offsets in the ranges returned in the result.

It also avoids reading/writing to memory whose CPU backed address is 0, as these are the non-resident portions of the sparse memory regions.

This should fix memory related issues that occur in titles that heavily make use of sparse textures, notably UE4 games.


FernandoS27:
This is just a workaround for an old existing hack. The "address 0" is what happens on a special mapBufferEx /Remap cases.

FernandoS27:
I have reconsidered it.

FernandoS27 (approved):

PR #7659, opened 2022-01-01
texture_cache/util: Fix s32 overflow when resolving overlaps

ameerj:
This addresses a miscalculation that occurs when resolving overlaps that are relatively distant apart in gpu memory space. 

The signed `diff` variable became negative for large differences, which resulted in the `if (diff > new_size)` check to fail under these circumstances, leading to large textures being allocated and crashes to ensue.

Fixes a VRAM spike+crash which most notably occurred in `Bravely Default 2`


FernandoS27 (approved):
nice catch, feel free to merge once everything passes.

PR #7660, opened 2022-01-01
Update translations (2022-01-01)

zhaobot:
This PR is generated by [zhaowenlan1779/txbot](https://github.com/zhaowenlan1779/txbot).


lioncash:
@zhaowenlan1779 Seems like this autogenerated PR is touching ffmpeg, which doesn't seem right

Morph1984:
Superseded by #7675

PR #7663, opened 2022-01-02
controller_applet: Only populate supported controllers

german77:
There is no need to display all controller types if game only allows a few.

### DIFF
--- a/src/yuzu/applets/qt_controller.cpp
+++ b/src/yuzu/applets/qt_controller.cpp
@@ -400,36 +400,66 @@ void QtControllerSelectorDialog::SetSupportedControllers() {
 }
 
 void QtControllerSelectorDialog::SetEmulatedControllers(std::size_t player_index) {
+    const auto npad_style_set = system.HIDCore().GetSupportedStyleTag();
     auto& pairs = index_controller_type_pairs[player_index];
 
     pairs.clear();
     emulated_controllers[player_index]->clear();
 
-    pairs.emplace_back(emulated_controllers[player_index]->count(),
-                       Core::HID::NpadStyleIndex::ProController);
-    emulated_controllers[player_index]->addItem(tr("Pro Controller"));
+    const auto add_item = [&](Core::HID::NpadStyleIndex controller_type,
+                              const QString& controller_name) {
+        pairs.emplace_back(emulated_controllers[player_index]->count(), controller_type);
+        emulated_controllers[player_index]->addItem(controller_name);
+    };
 
-    pairs.emplace_back(emulated_controllers[player_index]->count(),
-                       Core::HID::NpadStyleIndex::JoyconDual);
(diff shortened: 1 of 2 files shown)


Morph1984 (approved):

src/yuzu/applets/qt_controller.cpp
```diff
@@ -400,36 +400,66 @@ void QtControllerSelectorDialog::SetSupportedControllers() {
 }
 
 void QtControllerSelectorDialog::SetEmulatedControllers(std::size_t player_index) {
+    Core::HID::NpadStyleTag npad_style_set = system.HIDCore().GetSupportedStyleTag();
```
Morph1984, line 403:
```suggestion
    const auto npad_style_set = system.HIDCore().GetSupportedStyleTag();
```

src/yuzu/applets/qt_controller.cpp
```diff
@@ -400,36 +400,66 @@ void QtControllerSelectorDialog::SetSupportedControllers() {
...
     auto& pairs = index_controller_type_pairs[player_index];
 
     pairs.clear();
     emulated_controllers[player_index]->clear();
 
-    pairs.emplace_back(emulated_controllers[player_index]->count(),
-                       Core::HID::NpadStyleIndex::ProController);
-    emulated_controllers[player_index]->addItem(tr("Pro Controller"));
+    const auto add_item = [&](Core::HID::NpadStyleIndex controller_type,
+                              const QString& controller_name) {
```
Morph1984, line 410:
```suggestion
                              QString&& controller_name) {
```

src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -911,74 +911,59 @@ void ConfigureInputPlayer::SetConnectableControllers() {
     index_controller_type_pairs.clear();
     ui->comboControllerType->clear();
 
+    const auto add_item = [&](Core::HID::NpadStyleIndex controller_type,
+                              const QString& controller_name) {
```
Morph1984, line 915:
```suggestion
                              QString&& controller_name) {
```

lioncash: There's no real benefit to using an rvalue-reference here, since `addItem()` takes strings by const reference. It's also slightly more misleading in terms of intent given that.

src/yuzu/applets/qt_controller.cpp
lioncash: Ditto wrt my comment on the other similar review comment

src/yuzu/configuration/configure_input_player.cpp
Morph1984: Ah right, my brain fried looking at `emplace_back` and not at `addItem`, my bad 😅

bunnei (approved):

PR #7664, opened 2022-01-02
core/hid: Add fallback to fullkey controllers

german77:
Fullkey controllers fallback to pro controllers in real HW. We currently don't do that. To avoid this issue we reconnect the controller as Pro controller to avoid having to open the config window.

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -879,10 +879,36 @@ void EmulatedController::SetSupportedNpadStyleTag(NpadStyleTag supported_styles)
     if (!is_connected) {
         return;
     }
-    if (!IsControllerSupported()) {
-        LOG_ERROR(Service_HID, "Controller type {} is not supported. Disconnecting controller",
-                  npad_type);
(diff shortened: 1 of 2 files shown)


src/core/hid/emulated_controller.cpp
```diff
@@ -879,10 +879,36 @@ void EmulatedController::SetSupportedNpadStyleTag(NpadStyleTag supported_styles)
...
+    if (IsControllerSupported()) {
+        return;
+    }
+
+    Disconnect();
+
+    // Fallback fullkey controllers to Pro controllers
+    if (IsControllerFullkey() && supported_style_tag.fullkey) {
+        LOG_WARNING(Service_HID, "Reconnecting controller type {} as Pro controller", npad_type);
+        SetNpadStyleIndex(Core::HID::NpadStyleIndex::ProController);
```
Morph1984, line 891:
```suggestion
        SetNpadStyleIndex(NpadStyleIndex::ProController);
```

bunnei (approved):

PR #7668, opened 2022-01-03
gpu: Use std::stop_token in WaitFence for VSync thread

ameerj:
Fixes a hang that may occur when stopping emulation and the VSync thread is blocked on the syncpoint condition variable.

### DIFF
--- a/src/core/hle/service/nvflinger/nvflinger.cpp
+++ b/src/core/hle/service/nvflinger/nvflinger.cpp
@@ -266,10 +266,11 @@ void NVFlinger::Compose() {
 
         auto& gpu = system.GPU();
         const auto& multi_fence = buffer->get().multi_fence;
+        const auto stop_token = vsync_thread.get_stop_token();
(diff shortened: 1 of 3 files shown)


bunnei (approved):

FernandoS27:
Please redo this. a Stop token in WaitFence invades it's design and forces coupling towards only being of use with helper jthreads. This shouldn't be a case, the Event pattern should be used with a Set on deconstructor.

ameerj:
@FernandoS27 the stop token argument is only populated by its usage in nvflinger, all other usages are not forced to be helper jthreads or provide a stop token.

FernandoS27:
> @FernandoS27 the stop token argument is only populated by its usage in nvflinger, all other usages are not forced to be helper jthreads or provide a stop token.

Please don't this method is being coupled into GPU design. Use the Event pattern instead. The GPU design shouldn't be aware of the use by the client. This violates depenmdencies.

PR #7670, opened 2022-01-04
 gpu: Add shut down method to synchronize threads before destruction 

ameerj:
Alternative approach to #7668. Same goal of avoiding hangs on the vsync thread when stopping emulation.

### DIFF
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -312,6 +312,12 @@ struct GPU::Impl {
         cpu_context->MakeCurrent();
     }
 
+    void NotifyShutdown() {
+        std::unique_lock lk{sync_mutex};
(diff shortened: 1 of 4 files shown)


src/video_core/gpu.cpp
```diff
@@ -307,6 +312,11 @@ struct GPU::Impl {
         cpu_context->MakeCurrent();
     }
 
+    void NotifyShutdown() {
+        shutting_down.store(true, std::memory_order::relaxed);
+        sync_cv.notify_all();
```
FernandoS27, line 317: you must capture the lock here: "std::unique_lock lock{sync_mutex};"

FernandoS27 (commented):
Pretty good, just 1 thing.

FernandoS27 (approved):

PR #7674, opened 2022-01-05
configure_per_game: Initialize tabs after loading custom configuration

lat9nq:
Tab initialization happens before the game's custom configuration was loaded by the program. The original solution was to use the tab's SetConfiguration function to let the UI correctly, but visual highlighting was done in the UI setup and fixing that directly would just lead to a snowball of 'fixes'.

Changes tab initialization to happen after the configuration is loaded, which means that it no longer happens in member initializers in the `ConfigurePerGame` constructor. Removes the cluster of `??_tab->SetConfiguration`'s that I added earlier to get around this issue initially.

Fixes a regression in #6774


german77 (approved):

bunnei (approved):

bunnei:
Build failure is unrelated

PR #7683, opened 2022-01-08
logging: adapt to changes in fmt 8.1

liushuyu:
`fmtlib/fmt` upstream stated they fixed a bug introduced in `fmt` 6.0.0 where `enum class` was erroneously implicitly converted to `int`. They stated that this bug has been fixed in `fmt` 8.1.

Now that this "bug" has been patched, it broke yuzu's fundamental assumptions about the `fmt` library. A workaround was kindly provided by a third party on the issues page (see https://github.com/fmtlib/fmt/issues/2704) and is used in this pull request.

This "fix" should not break compatibilities with older `fmt`(< 8.1) but compilation performance may take a hit since a larger `fmt` header must be included in `logging/log.h` in order to apply the workaround.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,7 +169,7 @@ macro(yuzu_find_packages)
     set(REQUIRED_LIBS
     #    Cmake Pkg Prefix  Version     Conan Pkg
         "Catch2            2.13.7      catch2/2.13.7"
-        "fmt               8.0.1       fmt/8.0.1"
+        "fmt               8.0.1       fmt/8.1.1"
         "lz4               1.8         lz4/1.9.2"
         "nlohmann_json     3.8         nlohmann_json/3.8.0"
         "ZLIB              1.2         zlib/1.2.11"
--- a/src/common/logging/log.h
+++ b/src/common/logging/log.h
@@ -6,10 +6,23 @@
 
 #include <algorithm>
 #include <string_view>
+#include <type_traits>
 
-#include <fmt/core.h>
+#include <fmt/format.h>
 
 #include "common/logging/types.h"
+
+// adapted from https://github.com/fmtlib/fmt/issues/2704
+// a generic formatter for enum classes (<= 32 bits)
+#if FMT_VERSION >= 80100
+template <typename T>
+struct fmt::formatter<T, std::enable_if_t<std::is_enum_v<T>, char>> : formatter<u32> {
+    template <typename FormatContext>
+    auto format(const T& value, FormatContext& ctx) -> decltype(ctx.out()) {
+        return fmt::formatter<u32>::format(static_cast<u32>(value), ctx);
+    }
+};
+#endif
 
 namespace Common::Log {
 
--- a/src/common/logging/log.h
+++ b/src/common/logging/log.h
@@ -6,25 +6,11 @@
 
 #include <algorithm>
 #include <string_view>
-#include <type_traits>
 
 #include <fmt/format.h>
 
+#include "common/logging/formatter.h"
 #include "common/logging/types.h"
-
-// adapted from https://github.com/fmtlib/fmt/issues/2704
-// a generic formatter for enum classes (<= 32 bits)
-#if FMT_VERSION >= 80100
-template <typename T>
-struct fmt::formatter<T, std::enable_if_t<std::is_enum_v<T>, char>>
-    : formatter<std::underlying_type_t<T>> {
-    template <typename FormatContext>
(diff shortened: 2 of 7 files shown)


bunnei (approved):

lioncash:
While fmt stuff is being touched, this line [here](https://github.com/yuzu-emu/yuzu/blob/master/CMakeLists.txt#L172) should be changed to:

```cmake
        "fmt               8.1.1       fmt/8.1.1"
```

So that we're uniformly using the updated version

liushuyu:
> While fmt stuff is being touched, this line [here](https://github.com/yuzu-emu/yuzu/blob/master/CMakeLists.txt#L172) should be changed to:
> 
> ```cmake
>         "fmt               8.1.1       fmt/8.1.1"
> ```
> 
> So that we're uniformly using the updated version

I didn't make this change when I opened this pull request because Conan did not update `fmt` to 8.1.1 at that time. But it seems like it does now.

I have kept version requirements to `8.0.1` because I noticed some Linux distributions started to package yuzu and I would like to relax the dependency limitations for them (since they didn't upgrade `fmt` due to the same reason why this pull request is needed).

liushuyu:
For the record, @merryhime suggested I change hardcoded types to `std::underlying_type_t<T>`.

It seems like if we eventually use C++23, we could use the `to_underlying<T>(...)` function to replace the `static_cast` in the future.

CMakeLists.txt
```diff
@@ -169,7 +169,7 @@ macro(yuzu_find_packages)
     set(REQUIRED_LIBS
     #    Cmake Pkg Prefix  Version     Conan Pkg
         "Catch2            2.13.7      catch2/2.13.7"
-        "fmt               8.0.1       fmt/8.0.1"
+        "fmt               8.0.1       fmt/8.1.1"
```
lioncash, line 172:
```suggestion
        "fmt               8.1.1       fmt/8.1.1"
```

src/common/logging/log.h
```diff
@@ -6,11 +6,26 @@
...
 #include <string_view>
+#include <type_traits>
 
-#include <fmt/core.h>
+#include <fmt/format.h>
 
 #include "common/logging/types.h"
 
+// adapted from https://github.com/fmtlib/fmt/issues/2704
+// a generic formatter for enum classes (<= 32 bits)
```
Morph1984, line 16:
```suggestion
// a generic formatter for enum classes (<= 32 bits)
```
Since it now uses std::underlying_type_t, this comment no longer applies

Morph1984:
and fwiw casting to u32 would make signed enums lose their signed-ness

liushuyu:
> and fwiw casting to u32 would make signed enums lose their signed-ness, and larger enums would lose their higher bits

:+1: Good to know.

Although, according to this logic,

https://github.com/yuzu-emu/yuzu/pull/7683/files#diff-de5b457f285722a54f02ec7740228251032bc283e2f004bbbfdf456829685775R16

I should be able to remove the `static_cast` here. But if I remove this cast, compilers (tries Clang/GCC) will complain that it/they can't find a formatter implementation for `enum class Patch`. Am I missing something here or I am misunderstanding something?

Morph1984:
> that it/they can't find a formatter implementation for enum class Patch

Perhaps providing the same formatter implementation that you did in log.h in `shader_recompiler/exception.h` should do the trick?

liushuyu:
> > that it/they can't find a formatter implementation for enum class Patch
> 
> Perhaps providing the same formatter implementation that you did in log.h in `shader_recompiler/exception.h` should do the trick?

I think your suggestion is better since it affects fewer files.

I solve the issue by directly including `common/logging/log.h` inside `shader_recompiler/exception.h`. Duplicating the code would create a slew of re-definition errors.

src/shader_recompiler/exception.h
```diff
@@ -9,7 +9,7 @@
 #include <string_view>
 #include <utility>
 
-#include <fmt/format.h>
+#include "common/logging/log.h"
```
Morph1984, line 12: It may be better to just copy/paste the template into this file instead of bringing in unnecessary baggage from the log header

liushuyu: However, this would also cause quite a few files to have redefinition errors (included both `log.h` and `exception.h`. I think we need to add some `ifdef`s here?

Morph1984: Hmm, perhaps create a new file with the definition of the formatter, then include that in both files

liushuyu: Done. Moved to a separate file.

Morph1984 (approved):

Issue #7686, opened 2022-01-08
Loading Invalid TAS Recording Causes Crash


VD-15:
## Description

Loading an invalid or malformed TAS script with TAS features enabled will cause Yuzu to crash and may subsequently cause yuzu to crash on launch. I found this happens with TAS scripts with one or more frame numbers missing midway through the file, I imagine there are other edge cases that could cause this, too.

## Steps to Reproduce

1. In a file explorer, open yuzu's user directory and in the `tas` subdirectory, create or edit the file `script0-1.txt` with the following content (note the whitespace at the beginning of line 4):

```
0 NONE 0;0 0;0
1 NONE 0;0 0;0
2 NONE 0;0 0;0
  NONE 0;0 0;0
```

2. In yuzu, go to `Tools -> TAS -> Configure TAS...` and check `Enable TAS features`
3. Launch a game
4. Go to `Tools -> TAS -> Reset` (Or press Ctrl+F6) and yuzu will crash

Yuzu will also fail to launch after crashing this way. Presumably due to it attempting to load the malformed file on startup. Fixing the syntax error in `tas/script0-1.txt` or deleting it will then allow yuzu to launch normally.

## System Information

Issue was confirmed present on yuzu-mainline release 872 (commit 4e9911a) installed via AUR package.

- OS: Arch Linux
- Kernel: x86_64 Linux 5.15.13-arch1-1
- CPU: AMD Ryzen 7 3800XT 8-Core @ 16x 3.9GHz
- GPU: AMD Radeon RX 6700 XT (NAVY_FLOUNDER, DRM 3.42.0, 5.15.13-arch1-1, LLVM 13.0.0)

## Logfiles and Backtraces

I've attached yuzu's logfile generated from my session reproducing the crash. I was not able to generate a backtrace from this session but I was able to generate one when Yuzu was crashing on launch, which i have also attached:

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7833647/_yuzu_log.txt)
[backtrace.txt](https://github.com/yuzu-emu/yuzu/files/7833649/backtrace.txt)

german77:
Thanks for reporting I will fix it soon

MonsterDruide1:
There are probably a lot of similar issues to this one, regarding misformatted files. The current implementation does not have many checks in place to avoid crashing from broken files, so fixing this one might be better done in a complete overhaul of the file loading code, introducing checks for ... _everything_.

german77:
[...]

PR #7687, opened 2022-01-08
input_common: Handle errors on TAS scripts

german77:
It's posible to crash the emulator with misshaped TAS scripts.  This PR adds some error handling to prevent crashing from string to int/float conversions.

fixes: #7686

### DIFF
--- a/src/input_common/drivers/tas_input.cpp
+++ b/src/input_common/drivers/tas_input.cpp
@@ -105,10 +105,16 @@ void Tas::LoadTasFile(size_t player_index, size_t file_index) {
             continue;
         }
 
-        const auto num_frames = std::stoi(seg_list[0]);
-        while (frame_no < num_frames) {
-            commands[player_index].emplace_back();
-            frame_no++;
+        try {
+            const auto num_frames = std::stoi(seg_list[0]);
+            while (frame_no < num_frames) {
+                commands[player_index].emplace_back();
+                frame_no++;
+            }
+        } catch (const std::invalid_argument&) {
+            LOG_ERROR(Input, "Invalid argument: '{}' at command {}", seg_list[0], frame_no);
+        } catch (const std::out_of_range&) {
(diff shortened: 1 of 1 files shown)


src/input_common/drivers/tas_input.cpp
```diff
@@ -105,10 +105,16 @@ void Tas::LoadTasFile(size_t player_index, size_t file_index) {
...
-        while (frame_no < num_frames) {
-            commands[player_index].emplace_back();
-            frame_no++;
+        try {
+            const auto num_frames = std::stoi(seg_list[0]);
+            while (frame_no < num_frames) {
+                commands[player_index].emplace_back();
+                frame_no++;
+            }
+        } catch (std::invalid_argument) {
```
lioncash, line 114:
```suggestion
        } catch (const std::invalid_argument&) {
```

Otherwise this copies the exception

src/input_common/drivers/tas_input.cpp
```diff
@@ -105,10 +105,16 @@ void Tas::LoadTasFile(size_t player_index, size_t file_index) {
...
+            LOG_ERROR(Input, "Invalid argument: '{}' at command {}", seg_list[0], frame_no);
+        } catch (std::out_of_range) {
```
lioncash, line 116:
```suggestion
        } catch (const std::out_of_range&) {
```

src/input_common/drivers/tas_input.cpp
```diff
@@ -233,10 +239,21 @@ TasAnalog Tas::ReadCommandAxis(const std::string& line) const {
...
+        LOG_ERROR(Input, "Invalid axis data: '{}'", line);
+        return {};
+    }
 
-    return {x, y};
+    try {
+        const float x = std::stof(seg_list.at(0)) / 32767.0f;
+        const float y = std::stof(seg_list.at(1)) / 32767.0f;
+        return {x, y};
+    } catch (std::invalid_argument) {
```
lioncash, line 251: Ditto for these two

Morph1984 (approved):

PR #7689, opened 2022-01-09
video_core: changes in garbage collector config, added item to GUI

sayanss2:
1) Changed the modes of operation of the garbage collector (GC) texture cache:
- safe - always enabled if VRAM <= 4GiB, ticks and number of iterations have been changed
- high - call when 40% VRAM is exceeded, ticks and number of iterations have been changed
- aggressive - call when 70% VRAM is exceeded, ticks and number of iterations changed

2) Changed the size of the buffer cache (frequent GC calls were noticed in a number of games)

3) Added GUI settings to call GC only when 70% VRAM is exceeded (aggressive mode - disabled default)

As a result, when tested in the following games:
- ASTRAL CHAIN
- Xenoblade Chronicles 2
- No more heroes 3
managed to achieve smoother work when feeding frames, in particular when using scaling.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -61,6 +61,7 @@ void LogSettings() {
     log_setting("Renderer_UseVsync", values.use_vsync.GetValue());
     log_setting("Renderer_ShaderBackend", values.shader_backend.GetValue());
     log_setting("Renderer_UseAsynchronousShaders", values.use_asynchronous_shaders.GetValue());
+    log_setting("Renderer_UseGarbageCollectionAdvanced", values.use_caches_gc_adv.GetValue());
     log_setting("Renderer_AnisotropicFilteringLevel", values.max_anisotropy.GetValue());
     log_setting("Audio_OutputEngine", values.sink_id.GetValue());
     log_setting("Audio_OutputDevice", values.audio_device_id.GetValue());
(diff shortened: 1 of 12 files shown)


FernandoS27:
hmn I don't like the idea of adding an option like that. The more options, the harder is to use the emulator.

FernandoS27:
ok here's we can make it work without an option:

we'll have agresive threshold instead of being 40% VRAM, will be max(VRAM - 60% of 4Gb, 1.25 Gb)
we'll have critical threshold instead of being 70% VRAM, will be max(VRAM - 30% of 4Gb, 1.65 Gb)

Now ASTC should only be removed on high/critical, we can add a flag for them on register.
https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/texture_cache/texture_cache.h#L1414
add a new flag called "GCProtected"
https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/texture_cache/image_base.h#L35
Then check here if it's "GCProtected", it should only keep forward if in high priority mode
https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/texture_cache/texture_cache.h#L80

With this stuff, the option won't be needed anymore and it will be both fair for people with low-end cards and people with high-end cards.

sayanss2:
I'm closing this PR because a fix is ​​pending as part of PR #7720

PR #7695, opened 2022-01-11
common: bit_util: Add IsPow2 helper function

Morph1984:


src/common/bit_util.h
```diff
@@ -45,6 +45,12 @@ template <typename T>
     return static_cast<u32>(log2_f + static_cast<u64>((value ^ (1ULL << log2_f)) != 0ULL));
 }
 
+template <typename T>
+requires std::is_integral_v<T>
+[[nodiscard]] constexpr bool IsPow2(T value) {
```
lioncash, line 50: This is the same as `std::has_single_bit` in the `<bit>` header

src/common/bit_util.h
```diff
@@ -45,6 +45,12 @@ template <typename T>
...
+    return std::has_single_bit(static_cast<std::make_unsigned_t<T>>(value));
```
lioncash, line 51: I'm not particularly sure it's wise to allow passing in signed values in this function and silently convert them to unsigned vaalues, given this'll give incorrect results for negative values, which is why even the stdlib function doesn't allow passing in signed types

Morph1984: Hmm, @BreadFish64 suggested this one. Would the previous solution which tests for positivity be a better solution instead of using `std::has_single_bit`

lioncash: I would just constrain the function to unsigned values and any cases that make use of signed values can explicitly cast. Behaviorally, the functionality being tested for literally only makes sense with unsigned values. Implicitly casting to unsigned can hide bugs in call-sites.

Morph1984: The other method was `value > 0 && (value & (value - 1)) == 0`, whereas `std::has_single_bit` is functionally equivalent to `value != 0 && (value & (value - 1)) == 0`

Morph1984: The reason why I'd need support for signed integers is that Atmosphere implements an equivalent function using `value > 0 && (value & (value - 1)) == 0`, accepting signed integers.

bunnei (approved):

PR #7716, opened 2022-01-15
yuzu: Add volume hotkeys

german77:
Since the idea is to do everything with the gamepad It was decided to add hotkeys to control the volume. I removed the emulation speed hotkeys mainly since they only work on single core and it's not recommended anymore to use single core.

As an extra I noticed the volume % was a little far from ideal. So I changed the formula to something better.

### DIFF
--- a/src/audio_core/stream.cpp
+++ b/src/audio_core/stream.cpp
@@ -79,8 +79,8 @@ static void VolumeAdjustSamples(std::vector<s16>& samples, float game_volume) {
         return;
     }
 
-    // Implementation of a volume slider with a dynamic range of 60 dB
-    const float volume_scale_factor = volume == 0 ? 0 : std::exp(6.90775f * volume) * 0.001f;
+    // Perceived volume is not the same as the volume level
+    const float volume_scale_factor = (0.85f * ((volume * volume) - volume)) + volume;
     for (auto& sample : samples) {
         sample = static_cast<s16>(sample * volume_scale_factor);
     }
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1008,25 +1008,6 @@ void GMainWindow::InitializeHotkeys() {
             ToggleFullscreen();
         }
     });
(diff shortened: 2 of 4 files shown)


src/audio_core/stream.cpp
```diff
@@ -79,8 +79,8 @@ static void VolumeAdjustSamples(std::vector<s16>& samples, float game_volume) {
         return;
     }
 
-    // Implementation of a volume slider with a dynamic range of 60 dB
-    const float volume_scale_factor = volume == 0 ? 0 : std::exp(6.90775f * volume) * 0.001f;
+    // Perceived volume is not the same as the volume level
+    const float volume_scale_factor = (std::exp(volume) - 1) * 0.58198f;
```
Morph1984, line 83:
```suggestion
    const float volume_scale_factor = (std::exp(volume) - 1) * 0.5819767f;
```

src/yuzu/main.cpp
```diff
@@ -1035,6 +1016,22 @@ void GMainWindow::InitializeHotkeys() {
     });
     connect_shortcut(QStringLiteral("Mute Audio"),
                      [] { Settings::values.audio_muted = !Settings::values.audio_muted; });
+    connect_shortcut(QStringLiteral("Volume Down"), [] {
```
lioncash, line 1019: Any reason these aren't just:

```cpp
const auto current_volume = static_cast<int>(Settings::values.volume.GetValue());
const auto new_volume = std::clamp(current_volume - 5, 0, 100);
Settings::values.volume.SetValue(static_cast<u8>(new_volume));
```

ditto for volume up:

```cpp
const auto current_volume = static_cast<int>(Settings::values.volume.GetValue());
const auto new_volume = std::clamp(current_volume + 5, 0, 100);
Settings::values.volume.SetValue(static_cast<u8>(new_volume));
```

german77: I was trying to avoid negative values. But that works too

bunnei (approved):

PR #7719, opened 2022-01-16
Change default theme to Dark colorful

gidoly:
We did patreon poll about what theme should be new default theme instead white theme and dark colorful theme won


gidoly:
For note
Value 3 is dark colorful theme

bunnei (approved):

PR #7725, opened 2022-01-17
input_common: Reintroduce motion from mouse and use button names

german77:
We now have the necessary code to map the mouse as a motion device. This time with the z axis as the wheel input. On top of that use mouse button names since these are well defined.

### DIFF
--- a/src/input_common/drivers/mouse.cpp
+++ b/src/input_common/drivers/mouse.cpp
@@ -16,6 +16,7 @@ constexpr int mouse_axis_x = 0;
 constexpr int mouse_axis_y = 1;
 constexpr int wheel_axis_x = 2;
 constexpr int wheel_axis_y = 3;
+constexpr int motion_wheel_y = 4;
 constexpr int touch_axis_x = 10;
 constexpr int touch_axis_y = 11;
 constexpr PadIdentifier identifier = {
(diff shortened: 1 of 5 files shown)


src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -102,6 +102,16 @@ QString GetButtonName(Common::Input::ButtonNames button_name) {
         return QObject::tr("Share");
     case Common::Input::ButtonNames::Options:
         return QObject::tr("Options");
+    case Common::Input::ButtonNames::ButtonMouseWheel:
+        return QObject::tr("Wheel");
```
lioncash, line 106:
```suggestion
        return QObject::tr("Wheel", "Indicates the mouse wheel");
```

Since `"Wheel"` can be quite ambiguous in isolation to a translator

src/input_common/drivers/mouse.cpp
```diff
@@ -171,13 +176,39 @@ AnalogMapping Mouse::GetAnalogMappingForDevice(
     return mapping;
 }
 
+Common::Input::ButtonNames Mouse::GetUIButtonName(const Common::ParamPackage& params) const {
+    MouseButton button = static_cast<MouseButton>(params.Get("button", 0));
```
lioncash, line 180:
```suggestion
    const auto button = static_cast<MouseButton>(params.Get("button", 0));
```

since the type is already on the right hand side.

bunnei (approved):

PR #7726, opened 2022-01-17
service/hid: Initialize applet_resource on SetNpadAnalogStickUseCenterClamp

german77:
According to RE SetNpadAnalogStickUseCenterClamp is able to initialize applet_resource if it wasn't initialized before.

Fixes crash at boot on `パワプロクンポケットR`

### DIFF
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -1176,7 +1176,8 @@ void Hid::SetNpadAnalogStickUseCenterClamp(Kernel::HLERequestContext& ctx) {
 
     const auto parameters{rp.PopRaw<Parameters>()};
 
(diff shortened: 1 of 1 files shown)


src/core/hle/service/hid/hid.cpp
```diff
@@ -1176,13 +1176,17 @@ void Hid::SetNpadAnalogStickUseCenterClamp(Kernel::HLERequestContext& ctx) {
 
     const auto parameters{rp.PopRaw<Parameters>()};
 
-    applet_resource->GetController<Controller_NPad>(HidController::NPad)
```
Morph1984, line 1179: Just simply replacing applet_resource with `GetAppletResource()` should suffice

german77: I didn't notice there was a function for this. Ty

Morph1984 (approved):

PR #7728, opened 2022-01-17
hid: fix std::transform call

v1993:
Found by static analysis with PVS-Studio.


Moonlacer:
I find it weird that PVS tries to put end before begin, but I don't really know how it works

v1993:
Check out [function's signature](https://en.cppreference.com/w/cpp/algorithm/transform) (or surrounding lines) - this is indeed a valid fix.

german77 (approved):

lioncash (approved):

PR #7731, opened 2022-01-17
shader_recompiler: fix potential OOB access

v1993:
Found by static analysis with PVS-Studio. Original check wasn't actually checking for OOB and would segfault in case of it.

### DIFF
--- a/src/shader_recompiler/backend/glsl/glsl_emit_context.cpp
+++ b/src/shader_recompiler/backend/glsl/glsl_emit_context.cpp
@@ -458,9 +458,10 @@ void EmitContext::DefineGenericOutput(size_t index, u32 invocations) {
         std::string definition{fmt::format("layout(location={}", index)};
         const u32 remainder{4 - element};
         const TransformFeedbackVarying* xfb_varying{};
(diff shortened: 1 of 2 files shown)


src/shader_recompiler/backend/glsl/glsl_emit_context.cpp
```diff
@@ -458,9 +458,10 @@ void EmitContext::DefineGenericOutput(size_t index, u32 invocations) {
         std::string definition{fmt::format("layout(location={}", index)};
         const u32 remainder{4 - element};
         const TransformFeedbackVarying* xfb_varying{};
-        if (!runtime_info.xfb_varyings.empty()) {
-            xfb_varying = &runtime_info.xfb_varyings[base_index + element];
-            xfb_varying = xfb_varying && xfb_varying->components > 0 ? xfb_varying : nullptr;
+        size_t xfb_varying_index{base_index + element};
```
lioncash, line 461:
```suggestion
        const size_t xfb_varying_index{base_index + element};
```

For consistency with the vars above it. Ditto for the other file as well

bunnei (approved):

PR #7732, opened 2022-01-17
kernel: remove no-op code

v1993:
Found by static analysis with PVS-Studio. Nobody seems to really know what was it doing there.

### DIFF
--- a/src/core/hle/kernel/k_scheduler.cpp
+++ b/src/core/hle/kernel/k_scheduler.cpp
@@ -49,8 +49,6 @@ void KScheduler::RescheduleCores(KernelCore& kernel, u64 cores_pending_reschedul
         if (!must_context_switch || core != current_core) {
             auto& phys_core = kernel.PhysicalCore(core);
             phys_core.Interrupt();
-        } else {
-            must_context_switch = true;
         }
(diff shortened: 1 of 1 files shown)


CeruleanSky:
From best I can tell it seems to of been introduced by FernandoS27 from here:

[General: Recover Prometheus project from harddrive failure commit e31425d line 331](https://github.com/yuzu-emu/yuzu/pull/3955/commits/e31425df3877636c098ec7426ebd2067920715cb#diff-9760c83bc13ad24de6c0729f7d45ad9894ada7a8d5151f4d10b58a1d7212433cR331)
`This commit: Implements CPU Interrupts, Replaces Cycle Timing for Host 
Timing, Reworks the Kernel's Scheduler, Introduce Idle State and 
Suspended State, Recreates the bootmanager, Initializes Multicore 
system.`

Maybe this might jog the memory on what they were attempting.

v1993:
Thanks for tracing this back, that indeed looks like the original source of this code. Maybe @FernandoS27 can figure out what was the original intent?

FernandoS27:
it's fine I believe. That variable is only to a workaround to make Single Core work.

bunnei (approved):

PR #7735, opened 2022-01-17
input_common: Report battery for UDP controllers

german77:
Display the battery level for UDP controllers

### DIFF
--- a/src/input_common/drivers/udp_client.cpp
+++ b/src/input_common/drivers/udp_client.cpp
@@ -192,6 +192,25 @@ std::size_t UDPClient::GetClientNumber(std::string_view host, u16 port) const {
     return MAX_UDP_CLIENTS;
 }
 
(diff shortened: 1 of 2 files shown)


src/input_common/drivers/udp_client.cpp
```diff
@@ -192,6 +192,25 @@ std::size_t UDPClient::GetClientNumber(std::string_view host, u16 port) const {
...
+    case Response::Battery::Low:
+        return BatteryLevel::Critical;
+    case Response::Battery::Medium:
+        return BatteryLevel::Low;
+    case Response::Battery::High:
+        return BatteryLevel::Medium;
+    case Response::Battery::Full:
+        return BatteryLevel::Full;
+    case Response::Battery::Charging:
+    case Response::Battery::Charged:
```
v1993, line 208: It's probably a good idea to display "Charged" as "Full", informing user that charging is complete now.

bunnei (approved):

PR #7737, opened 2022-01-18
Various fixes to HLE service thread management

bunnei:
This change includes various fixes and improvements for HLE service thread management. Primarily, this was motivated by a KThread leak of dummy threads allocated to back service host threads, which are used for grabbing the kernel locks. Previously, we would not release these dummy threads when the service threads were destroyed. For games that create/destroy service interfaces frequently (e.g. Pokémon Sword and Shield), this would result in a pretty severe leak over long gameplay sessions.

Furthermore, previously, these dummy threads could inadvertently get scheduled on core 3 under certain conditions. These aren't threads that can actually be scheduled by the guest emulation, and this could result in a crash (e.g. in Pokémon Sword and Shield after many hours of gameplay). We add some checks to ensure this never happens, as well as decrease the dummy thread priority such that the idle thread is always preferred.

### DIFF
--- a/src/core/hle/kernel/service_thread.cpp
+++ b/src/core/hle/kernel/service_thread.cpp
@@ -12,6 +12,7 @@
 #include "common/scope_exit.h"
 #include "common/thread.h"
 #include "core/hle/kernel/k_session.h"
+#include "core/hle/kernel/k_thread.h"
 #include "core/hle/kernel/kernel.h"
 #include "core/hle/kernel/service_thread.h"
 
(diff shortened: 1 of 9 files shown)


src/core/hle/kernel/service_thread.cpp
```diff
@@ -50,6 +51,10 @@ ServiceThread::Impl::Impl(KernelCore& kernel, std::size_t num_threads, const std
 
             kernel.RegisterHostThread();
 
+            // Ensure the dummy thread allocated for this host thread is closed on exit.
+            auto* dummy_thread = kernel.GetCurrentEmuThread();
+            SCOPE_EXIT({ dummy_thread->Close(); });
```
vonchenplus, line 56: dummy threads are currently stored in the kernel's dummy_threads object, which may result in multiple destructions

bunnei: Good catch -- I forgot we had this. Will address.

bunnei:
Clang failures are unrelated

PR #7756, opened 2022-01-21
service: Update function tables

lioncash:
Updates a few function tables, either adding missing entries or changing some unknown function names into their proper names based off information available on SwitchBrew.


bunnei (approved):

TK3600:
What is the significance of this?

PR #7760, opened 2022-01-22
yuzu: Add modifiers for keyboard

### LINKED ISSUE #7759: Add the ability to invert the control stick modifier key.
This is a feature request to basically mimic the "!" (or "NOT") logic function that you can apply to keys within the Dolphin emulator. For the context of the modifier key, the idea is that you can invert the logic of the modifier key so that instead of decreasing the stick's position to a percentage, it will instead increase the stick's position to a percentage.
This would allow players who prefer walking to be the default movement when using a keyboard, and requiring the pressing down or toggle of the modifier key to start running (rather than running by default).

Here's an image below regarding how this is implemented within Dolphin. Normally to set this up within Dolphin, you have the range of each control stick direction decreased to a value under 100 (e.g. 80), and have the modifier key's value set to 200 (i.e. doubling the range from 80 to 160).
image

Thank you.

german77:
Somehow this wasn't added on kraken. Add the button modifiers to invert and toggle the input on keyboard

Closes #7759

### DIFF
--- a/src/yuzu/configuration/configure_input_player.cpp
+++ b/src/yuzu/configuration/configure_input_player.cpp
@@ -147,7 +147,7 @@ QString ConfigureInputPlayer::ButtonToText(const Common::ParamPackage& param) {
     // Retrieve the names from Qt
     if (param.Get("engine", "") == "keyboard") {
         const QString button_str = GetKeyName(param.Get("code", 0));
-        return QObject::tr("%1%2").arg(toggle, button_str);
+        return QObject::tr("%1%2%3").arg(toggle, inverted, button_str);
     }
 
     if (common_button_name == Common::Input::ButtonNames::Invalid) {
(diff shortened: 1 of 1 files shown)


bunnei (approved):

pianoboy8:
Testing this out, It's a bit hard to figure out if this is working as it's supposed to. Mainly the best way to test if this works/can replicate what the original #7759 request is if there was an additional range feature for left/right sticks and if the modifier itself could go past a value of 100.

For reference, for the intended mechanic to work (walk when pressing arrow keys, run when pressing arrow keys + modifier key), the arrow key/control stick should have a range value of 50-75, and the modifier key should be inverted and set to a value of 200. This would then make it so when holding down the modifier key, you would have a range value of 100-150.

pianoboy8:
@german77 by any chance are you planning to look into implementing the above?

german77:
I did implement just that on this PR. Let me make a short demostration of what you are trying to do. Unless I'm understanding all wrong

pianoboy8:
Sure thing. Maybe I missed a feature inside yuzu to make this change or something.

german77:
My bad. You are correct. The button is not working correctly. Give me a few minutes to do a proper fix

PR #7765, opened 2022-01-23
hle: kernel: KThread: Improve Increment/Decrement RunningThreadCount.

bunnei:
- Previously implementation was incorrect, and would occasionally underflow.


OZtistic:
little typo in src/core/hle/kernel/k_process.cpp - line 177
UNIMPLEMENTED_MSG("KProcess termination is not implemennted!");

PR #7773, opened 2022-01-24
input_common/udp_client: Replace deprecated from_string()/to_ulong() functions

lioncash:
These are deprecated and `make_address` variants and `to_uint()` should be used instead.

Just prevents the UDP client from breaking in the event the deprecated functions get removed at some point.


lioncash:
Error seems to be CI-related

Morph1984 (approved):

PR #7774, opened 2022-01-24
input_common/main: Pass MappingData by const reference in callbacks

lioncash:
Avoids creating unnecessary 168 byte copies per callback invocation.

Also does some minor touchups while in the area.


lioncash:
Failure seems CI related

Moonlacer:
Yes, builds are broken currently

Morph1984 (approved):

PR #7784, opened 2022-01-27
input_common: Add DS5 to HD rumble list

german77:
Dual sense controllers support HD rumble. So they should use a linear amplitude instead of the exponential for motors. This makes the controller vibrate with less force


Morph1984 (approved):

Mutcholoko:
Oh, okay, that makes sense.

PR #7791, opened 2022-01-27
wall_clock: Use standard wall clock if rtsc frequency is too low

german77:
If the rtsc frequency is lower than 1,000,000,000hz it will cause udiv128 to overflow. To prevent this issue we fallback to the StandardWallClock.

Fixes issue where yuzu crash at boot with no logs.

### DIFF
--- a/src/common/wall_clock.cpp
+++ b/src/common/wall_clock.cpp
@@ -72,7 +72,9 @@ std::unique_ptr<WallClock> CreateBestMatchingClock(u32 emulated_cpu_frequency,
     if (caps.invariant_tsc) {
         rtsc_frequency = EstimateRDTSCFrequency();
     }
-    if (rtsc_frequency == 0) {
+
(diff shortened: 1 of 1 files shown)


src/common/wall_clock.cpp
```diff
@@ -72,7 +72,7 @@ std::unique_ptr<WallClock> CreateBestMatchingClock(u32 emulated_cpu_frequency,
     if (caps.invariant_tsc) {
         rtsc_frequency = EstimateRDTSCFrequency();
     }
-    if (rtsc_frequency == 0) {
+    if (rtsc_frequency <= 1000000000) {
```
lioncash, line 75: There should probably be a comment here explaining why 1000000000 is used.

german77:
CI failure seems related to other event. I can trigger another build if needed

Morph1984 (approved):

PR #7800, opened 2022-01-29
spirv_atomic: Define U32x2 storage buffers for 64-bit storage atomics

ameerj:
Some drivers do not support 64-bit atomics, and fallback to atomically modifying U32x2 vectors. This change ensures that U32x2 storage vectors are defined in the spir-v shader when 64-bit atomics are used.

Fixes a hang when booting `Pokemon Legends Arceus` on some devices, notably Intel GPUs,.


Morph1984 (approved):

Morph1984:
Sad that Intel is the only major vendor not to support this on Windows

PR #7812, opened 2022-01-31
BufferCache: Find direction of the stream buffer increase.

FernandoS27:
This avoids doubling the buffer but instead increase it only where it's needed.

t should reduce the memory used in PLA.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -1464,19 +1464,27 @@ typename BufferCache<P>::OverlapResult BufferCache<P>::ResolveOverlaps(VAddr cpu
         overlap_ids.push_back(overlap_id);
         overlap.Pick();
         const VAddr overlap_cpu_addr = overlap.CpuAddr();
-        if (overlap_cpu_addr < begin) {
+        const bool expands_left = overlap_cpu_addr < begin;
+        if (expands_left) {
             cpu_addr = begin = overlap_cpu_addr;
         }
-        end = std::max(end, overlap_cpu_addr + overlap.SizeBytes());
-
+        const VAddr overlap_end = overlap_cpu_addr + overlap.SizeBytes();
+        const bool expands_right = overlap_end > end;
+        if (overlap_end > end) {
+            end = overlap_end;
+        }
         stream_score += overlap.StreamScore();
         if (stream_score > STREAM_LEAP_THRESHOLD && !has_stream_leap) {
             // When this memory region has been joined a bunch of times, we assume it's being used
             // as a stream buffer. Increase the size to skip constantly recreating buffers.
             has_stream_leap = true;
-            begin -= PAGE_SIZE * 256;
-            cpu_addr = begin;
-            end += PAGE_SIZE * 256;
+            if (expands_right) {
+                begin -= PAGE_SIZE * 256;
+                cpu_addr = begin;
+            }
+            if (expands_left) {
+                end += PAGE_SIZE * 256;
+            }
         }
     }
     return OverlapResult{


src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -1464,19 +1464,29 @@ typename BufferCache<P>::OverlapResult BufferCache<P>::ResolveOverlaps(VAddr cpu
         overlap_ids.push_back(overlap_id);
         overlap.Pick();
         const VAddr overlap_cpu_addr = overlap.CpuAddr();
+        bool goes_left = false;
```
Morph1984, line 1467:
```suggestion
        bool goes_left = overlap_cpu_addr < begin;
```

src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -1464,19 +1464,29 @@ typename BufferCache<P>::OverlapResult BufferCache<P>::ResolveOverlaps(VAddr cpu
...
         if (overlap_cpu_addr < begin) {
+            goes_left = true;
             cpu_addr = begin = overlap_cpu_addr;
         }
-        end = std::max(end, overlap_cpu_addr + overlap.SizeBytes());
-
+        const VAddr overlap_end = overlap_cpu_addr + overlap.SizeBytes();
+        bool goes_right = false;
```
Morph1984, line 1473:
```suggestion
        bool goes_right = overlap_end > end;
```

Morph1984 (commented):
These make the comparisons more concise

ameerj (changes requested):
This is a naive and improper approach. You are not taking into account the historical expansion direction into account. 

Let's say we have a buffer that kept getting expanded to the right, but at the instance where its `stream_score` is above the threshold, it happens to be expanding to the left. Your logic will only expand it further to the left, despite being far more biased to expanding to the right.

Also, the existing logic isn't "doubling the buffer" size. it increases the buffer size by `PAGE_SIZE * 256` on both ends, which is 16MiB. This size increase is arbitrary though and can be tweaked.

src/video_core/buffer_cache/buffer_cache.h
```diff
@@ -1464,19 +1464,29 @@ typename BufferCache<P>::OverlapResult BufferCache<P>::ResolveOverlaps(VAddr cpu
...
-            begin -= PAGE_SIZE * 256;
-            cpu_addr = begin;
-            end += PAGE_SIZE * 256;
+            if (goes_right) {
+                begin -= PAGE_SIZE * 256;
+                cpu_addr = begin;
+            }
+            if (goes_left) {
+                end += PAGE_SIZE * 256;
+            }
```
ameerj, line 1489: Your logic here is swapped
```suggestion
            if (goes_left) {
                begin -= PAGE_SIZE * 256;
                cpu_addr = begin;
            }
            if (goes_right) {
                end += PAGE_SIZE * 256;
            }
```

ameerj (changes requested):

src/video_core/buffer_cache/buffer_cache.h
ameerj: these can also be made `const` and moved to the body of the if condition where they are actually being used.

src/video_core/buffer_cache/buffer_cache.h
FernandoS27: you are confused, it's the other way around. The real overlap is the "find range", what it's called overlap in that function are the buffers that existed before it.

ameerj: Then use more intuitive variable names for these cases. `expand_left` and `expand_right`

bunnei (approved):

PR #7816, opened 2022-01-31
input_common: Fix G-Shark 702 faulty mapping

german77:
This fixes a bug on SDL mappings where the Y axis is not detected at all on G-Shark 702 controllers. It's supposed to be fixed in 2.0.18 but since this is not the case I added the fix manually.

Needs confirmation from someone who has this controller before merge since I can't test.


german77:
This only fixes default mappings. Not the actual issue

PR #7829, opened 2022-02-01
common/param_package: Allow heterogenous key lookup 

lioncash:
Allows the functions that only query or retrieve from the underlying map to be able to handle keys in various string formats.

This means we can make these bits of the API non-allocating by default.


lioncash:
Unfortunate, looks like our builders don't use more recent compilers

PR #7832, opened 2022-02-01
input_common: Use recursive mutex

german77:
Fixes some crashes due calling the mutex twice while using hotkeys. 

Issue:
You press the button -> set the mutex ->triggers hotkey->hotkey opens another other window-> on out of focus releases all keys -> calls the mutex again

### DIFF
--- a/src/input_common/input_engine.h
+++ b/src/input_common/input_engine.h
@@ -217,7 +217,7 @@ private:
                                 int index) const;
 
     mutable std::mutex mutex;
-    mutable std::mutex mutex_callback;
+    mutable std::recursive_mutex mutex_callback;
     bool configuring{false};
     const std::string input_engine;
     int last_callback_key = 0;


src/input_common/input_engine.h
```diff
@@ -216,8 +216,8 @@ class InputEngine {
                                 const PadIdentifier& identifier, EngineInputType type,
                                 int index) const;
 
-    mutable std::mutex mutex;
-    mutable std::mutex mutex_callback;
+    mutable std::recursive_mutex mutex;
```
lioncash, line 219: This seems like it's papering over a threading issue elsewhere in our codebase, based off a cursory glance of how the InputEngine class operates

german77: I do agree that this isn't the nicest of the solutions. But I don't like more to have a know crash method. I'm still looking into a better solution to avoid having a recursive_mutex to begin with

Morph1984:
Superseded by #8249

PR #7852, opened 2022-02-05
common: Revise and fix the UUID implementation

Morph1984:
This is a fixed and revised implementation of UUID that uses an array of bytes as its internal representation of a UUID instead of a u128 (which was an array of 2 u64s).
In addition to this, the generation of RFC 4122 Version 4 compliant UUIDs is also implemented.


src/common/uuid.h
```diff
@@ -1,79 +1,122 @@
...
+
+    constexpr UUID& operator=(const UUID&) noexcept = default;
+    constexpr UUID& operator=(UUID&&) noexcept = default;
+
+    /**
+     * Returns whether the stored UUID is valid or not.
+     *
+     * @returns True if the stored UUID is valid, false otherwise.
+     */
+    constexpr bool IsValid() const {
```
lioncash, line 52: I'd preserve the `IsInvalid()` function with the new interface.

``if (IsInvalid())` is harder to misread compared to `if (!IsValid())`

src/common/uuid.cpp
```diff
@@ -1,23 +1,27 @@
...
 namespace {
 
-bool IsHexDigit(char c) {
-    return (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F');
-}
+constexpr size_t RawStringSize = sizeof(UUID) * 2;
+constexpr size_t FormattedStringSize = RawStringSize + 4;
 
-u8 HexCharToByte(char c) {
+u8 HexCharToByte(char c, bool& is_invalid) {
```
lioncash, line 21: wouldn't it make sense to just return an optional here instead of using an out parameter? This is putting stateful behavior into the function that it shouldn't even be caring about.

src/common/uuid.cpp
```diff
@@ -28,60 +32,172 @@ u8 HexCharToByte(char c) {
...
 }
 
-} // Anonymous namespace
+std::array<u8, 0x10> ConstructFromRawString(std::string_view raw_string) {
+    std::array<u8, 0x10> uuid;
+
+    bool is_invalid = false;
+
+    for (size_t i = 0; i < RawStringSize; i += 2) {
+        uuid[i / 2] = static_cast<u8>((HexCharToByte(raw_string[i], is_invalid) << 4) |
```
lioncash, line 45: Seems weird to keep continuing the loop even in the event something is invalid instead of early exiting.

bunnei (approved):

PR #7519, opened 2021-12-05
kernel: svc: Implement ProcessMemory and CodeMemory SVCs

itsmeft24:
Used by Skyline modding framework, see #7392 for tracking issue on skyline support.

PR #7870, opened 2022-02-09
hle: kernel: KCodeMemory: Correct m_page_group number of pages.

bunnei:
Credits to @xerpi for finding this issue and pointing it out on #7519.


itsmeft24:
sweet! i wasnt able to submit another PR yesterday so thx for doing it!

PR #7877, opened 2022-02-11
audio_core: Update current process revision

lat9nq:
Update CURRENT_PROCESS_REVISION from REV9 to REVA.

Used by Nintendo Entertainment System - Nintendo Switch Online 6.0.0 and Super Nintendo Entertainment System - Nintendo Switch Online 3.0.0.

### DIFF
--- a/src/audio_core/common.h
+++ b/src/audio_core/common.h
@@ -15,7 +15,9 @@ constexpr ResultCode ERR_INVALID_PARAMETERS{ErrorModule::Audio, 41};
 constexpr ResultCode ERR_SPLITTER_SORT_FAILED{ErrorModule::Audio, 43};
 } // namespace Audren
 
-constexpr u32_le CURRENT_PROCESS_REVISION = Common::MakeMagic('R', 'E', 'V', '9');
+constexpr u8 BASE_REVISION = '0';
+constexpr u32_le CURRENT_PROCESS_REVISION =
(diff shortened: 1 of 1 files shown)


src/audio_core/common.h
```diff
@@ -15,7 +15,7 @@ constexpr ResultCode ERR_INVALID_PARAMETERS{ErrorModule::Audio, 41};
 constexpr ResultCode ERR_SPLITTER_SORT_FAILED{ErrorModule::Audio, 43};
 } // namespace Audren
 
-constexpr u32_le CURRENT_PROCESS_REVISION = Common::MakeMagic('R', 'E', 'V', '9');
+constexpr u32_le CURRENT_PROCESS_REVISION = Common::MakeMagic('R', 'E', 'V', 'A');
```
Morph1984, line 18:
```suggestion
constexpr u32_le CURRENT_PROCESS_REVISION = Common::MakeMagic('R', 'E', 'V', ':');
```
Nintendo increments this byte with the ASCII value, so '9' + 1 = ':'

lat9nq: Do you think I should change this to 0x3A instead of ':'? It's the reason I was confused, and I think having the byte written out would be more clear as to what's going here anyway.

lioncash: you would still need to cast in the case of the last parameter, since `u8 + int -> int`

Morph1984 (approved):

bunnei (approved):

Issue #7337, opened 2021-11-14
Splatoon 2 Octo Expansion Level 4: Roll Out Station not Beatable


DavidBeh:
I am not able to roll on the wall after the 3rd dash track up.
The ball slides down the wall into the void leaving a ink trail behind.

Version: 809
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7534618/yuzu_log_splatoonbug.txt)

lat9nq:
I am able to reproduce this issue on RADV only on Mainline 910. I do not have hardware to test the latest AMDVLK with, which the original issue was made with. This looks to be local to Vulkan on AMD devices.

I can complete Station A04 with NVIDIA's OpenGL and Vulkan drivers, as well as RadeonSI, iris, and ANV.

benderscruffy:
Is this still an issue ??

Issue #4982, opened 2020-11-24
Splatoon 2 Ink bug


VasyaPupkins:
Im using Yuzu EA 1144, but this bug exists for long time and still not fixed.
Some surfaces clean or inked is behaving like there is enemy ink on it. Even if i cover it with my ink. Some levels unplayable because of it. Tryed every possible settings but it not helps. In other emulator (ryujinx) this game runs without any issues.
Look at screenshots, hope it explains what im trying to say. You can face this problem in Sector 2, Sector 2 Level 6, Sector 3 Level 14:

![splatoon2-ink-bug-sector2-clean-plate](https://user-images.githubusercontent.com/74937848/100052197-860b4600-2e71-11eb-858c-9d5bd4832016.png)
![splatoon2-ink-bug-sector2-inked-plate](https://user-images.githubusercontent.com/74937848/100052321-c5399700-2e71-11eb-8820-637bfb97355c.png)
![splatoon2-ink-bug-sector3-level14](https://user-images.githubusercontent.com/74937848/100052432-0f227d00-2e72-11eb-9535-449c1310c501.png)

Here is log file:
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5587606/yuzu_log.txt)

slashiee:
I'm unable to replicate this issue.
<img width="961" alt="2020-11-23_22-19-07_yuzu" src="https://user-images.githubusercontent.com/15317421/100056091-e6e73f80-2dd9-11eb-9ebe-44761b30da37.png">

Tabby4675:
For me, it happened after beating Level 1 of Sector 1, the whole Hub world acted as covered in enemy ink, but my ink worked just fine.

Try restarting the emulator after you've compiled the shaders (use disk shader cache) and see if the problem persists, since the problem only happened to me the first time around and fixed after a restart of the emulator.

funtimeandrefoxy:
This error also happens to me when I use vulkan.

PeachMangoes:
This issue is still occurring exactly as the author says. I'm using yuzu 580 with Splatoon 2 v5.4.0.

VasyaPupkins:
> 
> 
> I'm unable to replicate this issue.
> 



> 
> 
> For me, it happened after beating Level 1 of Sector 1, the whole Hub world acted as covered in enemy ink, but my ink worked just fine.
> 
[...]

PR #7884, opened 2022-02-12
Vulkan Rasterizer: Move counter updates just before the draw.

FernandoS27:
Fixes Splatoon 2 Ink when there's multiple floors on Vulkan.

Long story short: there were empty queries continuously created due to "OutOfRenderPass" operations on the scheduler, causing counters to be filled with nothing.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -183,8 +183,6 @@ void RasterizerVulkan::Draw(bool is_indexed, bool is_instanced) {
     SCOPE_EXIT({ gpu.TickWork(); });
     FlushWork();
 
-    query_cache.UpdateCounters();
-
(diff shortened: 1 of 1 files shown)


lat9nq:
This PR alone does not fix #7337 or #4982

From my testing, neither NVIDIA Vulkan/Maxwell v2 nor ANV/RKL are affected (they do not have the issue in the first place). 

RADV/GCN 3.0 and NVIDIA Vulkan/Kepler are affected. I heard from other members that NVIDIA Vulkan/Turing is also affected.

funtimeandrefoxy:
It remains the same for me both for my pc of these specs (see image) and for my pc intel hd graphcis 520.
Also this bug behaves differently for both pcs.
And I think there are all sorts of bugs related to paint like for example surfaces that I can walk in squid mode even though there's no paint there or I can't climb walls even though there's paint there.
This correction may have given some results but it did not correct the problems that continued to have.
image

FernandoS27:
hmn, likely it seemed fixed.

PR #7903, opened 2022-02-16
cmake: improve ffmpeg/boost detection

voidanix:


voidanix:
Closing because this was mostly centered around Linux distros that do not support `BoostConfig.cmake`

Issue #7023, opened 2021-09-16
ANV: Mario Kart 8 Deluxe rendering issues


lat9nq:
MK8D's title screen and character select has an odd effect when using ANV. Additionally, the character select is missing the primary light source. The emulator crashes immediately after selecting a character.

<img src="https://user-images.githubusercontent.com/22451773/133535881-22d54f0b-cea8-47d3-a996-e12f4b648f3b.png" width=640/>

<img src="https://user-images.githubusercontent.com/22451773/133535889-164fe99a-716c-4362-a484-be0d83fcd163.png" width=640/>

[yuzu_log.txt.tar.gz](https://github.com/yuzu-emu/yuzu/files/7174327/yuzu_log.txt.tar.gz)

Output of `inxi -GSC -xx`:
```
System:    Host: Qykopi Kernel: 5.13.10-arch1-2-vfio x86_64 bits: 64 compiler: gcc v: 11.1.0 Desktop: Xfce 4.16.0 
           tk: Gtk 3.24.29 wm: xfwm dm: LightDM Distro: Arch Linux 
CPU:       Info: 6-Core model: 11th Gen Intel Core i5-11600K bits: 64 type: MT MCP arch: Rocket Lake rev: 1 cache: 
           L2: 12 MiB 
           flags: avx avx2 lm nx pae sse sse2 sse3 sse4_1 sse4_2 ssse3 vmx bogomips: 93924 
           Speed: 800 MHz min/max: 800/5000 MHz Core speeds (MHz): 1: 800 2: 800 3: 1377 4: 884 5: 884 6: 854 7: 805 
           8: 805 9: 923 10: 861 11: 1426 12: 1591 
Graphics:  Device-1: Intel RocketLake-S GT1 [UHD Graphics 750] vendor: ASUSTeK driver: i915 v: kernel 
           bus-ID: 00:02.0 chip-ID: 8086:4c8a 
           Device-2: Advanced Micro Devices [AMD/ATI] Tonga PRO [Radeon R9 285/380] vendor: XFX Pine driver: amdgpu 
           v: kernel bus-ID: 03:00.0 chip-ID: 1002:6939 
           Display: x11 server: X.Org 1.20.13 compositor: picom driver: loaded: amdgpu,modesetting 
           alternate: ati,fbdev,intel,vesa resolution: 1: 1920x1200~60Hz 2: 2560x1440 s-dpi: 96 
[...]

PR #7913, opened 2022-02-18
vulkan_device: fix missing format in ANV

### LINKED ISSUE #6958: [Iris Xe][Mesa][ANV][Vulkan] Bowsers Fury - Crash after selecting Bowsers Fury
Game:Bowsers Fury - Crash after selecting Bowsers Fury
Yuzu Version : EA 2031
CPU : i7 1165G7
RAM : 16GB
GPU : Iris Xe
GPU Driver : Mesa 21.2.1
OS: Debian 11
Kernel : 5.13.13-xanmod1

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7103156/yuzu_log.txt)

voidanix:
Currently Mesa's ANV driver [does not support VK_FORMAT_B5G6R5_UNORM_PACK16](https://gitlab.freedesktop.org/mesa/mesa/-/blob/21.3/src/intel/vulkan/anv_formats.c#L152) which results in a crash; implement an alternative for it.

Fixes #6958 and part of #7023

### DIFF
--- a/src/video_core/renderer_vulkan/vk_texture_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_texture_cache.cpp
@@ -554,10 +554,12 @@ void CopyBufferToImage(vk::CommandBuffer cmdbuf, VkBuffer src_buffer, VkImage im
     };
 }
 
-[[nodiscard]] bool IsFormatFlipped(PixelFormat format) {
+[[nodiscard]] bool IsFormatFlipped(PixelFormat format, bool emulate_bgr565) {
     switch (format) {
     case PixelFormat::A1B5G5R5_UNORM:
         return true;
+    case PixelFormat::B5G6R5_UNORM:
+        return emulate_bgr565;
     default:
         return false;
     }
@@ -1488,7 +1490,7 @@ ImageView::ImageView(TextureCacheRuntime& runtime, const VideoCommon::ImageViewI
     };
     if (!info.IsRenderTarget()) {
         swizzle = info.Swizzle();
-        if (IsFormatFlipped(format)) {
+        if (IsFormatFlipped(format, device->MustEmulateBGR565())) {
             std::ranges::transform(swizzle, swizzle.begin(), SwapBlueRed);
         }
         if ((aspect_mask & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT)) != 0) {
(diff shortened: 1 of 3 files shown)


FernandoS27:
hmn, it's a good solution even though I would have preffered that ANV got that format back n track as the official driver supports.

Considering the alternative format is not hard to emulate. This is good enough.

FernandoS27 (approved):

voidanix:
There seems to be more than one format working, currently testing and will add them soon

src/video_core/vulkan_common/vulkan_device.cpp
```diff
@@ -39,6 +39,14 @@ constexpr std::array DEPTH16_UNORM_STENCIL8_UINT{
     VK_FORMAT_D32_SFLOAT_S8_UINT,
     VK_FORMAT_UNDEFINED,
 };
+
+constexpr std::array B5G6R5_UNORM_PACK16{
+    VK_FORMAT_R5G6B5_UNORM_PACK16,
+    VK_FORMAT_R5G5B5A1_UNORM_PACK16,
+    VK_FORMAT_A1R5G5B5_UNORM_PACK16,
+    VK_FORMAT_R4G4B4A4_UNORM_PACK16,
```
FernandoS27, line 47: this 3 are not good alternatives

voidanix:
Thanks for the link :)

btw thanks to MK8 Deluxe making it obvious, swizzling **is** required.

> this 3 are not good alternatives

yup, fixed by reverting to what it was like before.

It now seems to work correctly both on ANV and RADV, although sometimes swapBlueRed gets applied when it shouldn't for some reason.

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -554,10 +554,16 @@ void CopyBufferToImage(vk::CommandBuffer cmdbuf, VkBuffer src_buffer, VkImage im
...
+[[nodiscard]] bool IsFormatFlipped(PixelFormat format, bool emulate_bgr) {
     switch (format) {
     case PixelFormat::A1B5G5R5_UNORM:
         return true;
+    case PixelFormat::B5G6R5_UNORM:
+        if (emulate_bgr) {
+            return true;
+        } else {
+            return false;
+        }
```
ameerj, line 566:
```suggestion
        return emulate_bgr;
```

ameerj (commented):
I would also suggest renaming the instances of "emulate BGR" to "emulate BGR565" since the logic applies to that specific format only. Otherwise it could be misconstrued to apply to 32-bit BGR formats as well i.e. `B8G8R8A8`

FernandoS27:
just that last snip and it's good to go.

bunnei (approved):

PR #7932, opened 2022-02-21
Add extended memory layout (6GB) support and improve KResourceLimit management

bunnei:
The main purpose of this change is to add a setting to emulate an extended memory layout (6GB), which has been a highly requested feature for allowing use of certain mods (high resolution and texture mods, etc.) that require the extra memory. This is a memory layout that is available on Switch for developer hardware that has the extra memory. This is something that has been part of yuzu for a while now, but not exposed as a setting. Note, Switch also supports an 8GB layout, but this is not useful to expose because it does not allow any more memory to the emulated process than the 6GB mode.

As part of this, we needed to improve our KResourceLimit management. Previously, yuzu would use the system (kernel) KResourceLimit for all emulated processes (the main userland process + the processes created for each HLE'd service). Not only is this incorrect, but it exposed a problem where we would misreport available physical memory for different layouts. This only worked before due to a hack that fudged the available memory for 4GB. With these changes, we can remove this hack.

This new setting can be found under the "General" section of yuzu Configuration:

image

It is also available as a per-game setting. We might want to find a better place for this.

### DIFF
--- a/src/core/device_memory.cpp
+++ b/src/core/device_memory.cpp
@@ -3,10 +3,13 @@
 // Refer to the license.txt file included.
 
 #include "core/device_memory.h"
+#include "hle/kernel/board/nintendo/nx/k_system_control.h"
 
 namespace Core {
 
-DeviceMemory::DeviceMemory() : buffer{DramMemoryMap::Size, 1ULL << 39} {}
+DeviceMemory::DeviceMemory()
+    : buffer{Kernel::Board::Nintendo::Nx::KSystemControl::Init::GetIntendedMemorySize(),
+             1ULL << 39} {}
 DeviceMemory::~DeviceMemory() = default;
 
 } // namespace Core
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -167,6 +167,7 @@ void RestoreGlobalState(bool is_powered_on) {
 
     // Core
     values.use_multi_core.SetGlobal(true);
+    values.use_extended_memory_layout.SetGlobal(true);
 
     // CPU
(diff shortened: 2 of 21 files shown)


Moonlacer:
Could this option be put in the CPU tab? I also think "Multicore CPU Emulation could be moved there as well

src/core/device_memory.cpp
```diff
@@ -3,10 +3,13 @@
...
 
 #include "core/device_memory.h"
+#include "hle/kernel/board/nintendo/nx/k_system_control.h"
 
 namespace Core {
 
-DeviceMemory::DeviceMemory() : buffer{DramMemoryMap::Size, 1ULL << 39} {}
+DeviceMemory::DeviceMemory()
+    : buffer{Kernel::Board::Nintendo::Nx::KSystemControl::Init::GetIntendedMemorySize(),
+             1ULL << 39} {}
```
Wunkolo, line 12: Might be a good moment to use the `512_GiB` literal here over in `literals.h`.

Morph1984:
> Could this option be moved to the CPU tab? I also think "Multicore CPU Emulation" could be moved there as well

This isn't really a CPU option

src/core/device_memory.cpp
bunnei: To ack -- 39 here reflects Switch's max 39-bit virtual address space, which may be more obvious than 512_GB

SilentStorm-1986:
Does anyone have a virtual memory map of the 6GB layout? (ie. https://wiiubrew.org/wiki/Cafe_OS#Virtual_Memory_Map)

PR #7933, opened 2022-02-22
service: am: Update enum names to match documentation

german77:
Purely visual change.

### DIFF
--- a/src/core/hle/service/apm/apm_controller.h
+++ b/src/core/hle/service/apm/apm_controller.h
@@ -32,15 +32,18 @@ enum class PerformanceConfiguration : u32 {
     Config16 = 0x9222000C,
 };
 
+// This is nn::oe::CpuBoostMode
 enum class CpuBoostMode : u32 {
-    Disabled = 0,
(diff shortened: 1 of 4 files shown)


src/core/hle/service/apm/apm_controller.h
```diff
@@ -32,15 +32,18 @@ enum class PerformanceConfiguration : u32 {
...
+    Normal = 0,   // Boost mode disabled
+    FastLoad = 1, // CPU + GPU -> Config 13, 14, 15, or 16
+    Partial = 2,  // GPU Only -> Config 15 or 16
 };
 
-enum class PerformanceMode : u8 {
-    Handheld = 0,
-    Docked = 1,
+// This is nn::oe::PerformanceMode
+enum class PerformanceMode : s32 {
```
lioncash, line 43: Note that changing this to s32 isn't a visual change. This will push 4 bytes instead of 1 inside `APM::GetPerformanceMode` is that intended?

german77: `APM::GetPerformanceMode` Should return a 4 byte value

lioncash (approved):

PR #7935, opened 2022-02-24
logging: Convert `backend_thread` into an `std::jthread`

Wunkolo:
Was getting an unhandled `invalid_argument` [exception](https://en.cppreference.com/w/cpp/thread/thread/join) during shutdown on my linux
machine. This function is called by the deconstructor and
may be called before `StartBackendThread` gets a chance to run, causing it to try and join a null thread.

### DIFF
--- a/src/common/logging/backend.cpp
+++ b/src/common/logging/backend.cpp
@@ -218,19 +218,17 @@ private:
     Impl(const std::filesystem::path& file_backend_filename, const Filter& filter_)
         : filter{filter_}, file_backend{file_backend_filename} {}
 
-    ~Impl() {
-        StopBackendThread();
-    }
+    ~Impl() = default;
 
     void StartBackendThread() {
-        backend_thread = std::thread([this] {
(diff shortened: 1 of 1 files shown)


src/common/logging/backend.cpp
```diff
@@ -246,7 +246,9 @@ class Impl {
 
     void StopBackendThread() {
         stop.request_stop();
-        backend_thread.join();
+        if (backend_thread.joinable()) {
```
lioncash, line 249: Would it be beneficial to turn the thread itself into a `std::jthread` so that this can happen automatically on destruction (since the only use of StopBackendThread is in the destructor)

Wunkolo: I'll try that. It looks like we are basically just re-implementing a `std::jthread` when it can just use it directly.

bunnei (approved):

bunnei:
LGTM, thanks!

PR #7936, opened 2022-02-24
cpu_detect: Refactor detection of processor features

Wunkolo:
These are some updates and fixes for `cpu_detect`'s `CPUCaps` structure to make it more robust for later feature-detection additions.

### DIFF
--- a/src/common/x64/cpu_detect.h
+++ b/src/common/x64/cpu_detect.h
@@ -1,12 +1,13 @@
-// Copyright 2013 Dolphin Emulator Project / 2015 Citra Emulator Project
-// Licensed under GPLv2 or any later version
-// Refer to the license.txt file included.
+// Copyright 2013 Dolphin Emulator Project / 2015 Citra Emulator Project / 2022 Yuzu Emulator
+// Project Project Licensed under GPLv2 or any later version Refer to the license.txt file included.
 
 #pragma once
 
+#include "common/common_types.h"
+
 namespace Common {
 
-enum class Manufacturer : u32 {
+enum class Manufacturer : u8 {
     Intel = 0,
     AMD = 1,
     Hygon = 2,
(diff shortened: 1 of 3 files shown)


src/common/x64/cpu_detect.h
```diff
@@ -1,42 +1,50 @@
...
-    bool invariant_tsc;
+    char brand_string[13];
+
+    char cpu_string[48];
+
     u32 base_frequency;
     u32 max_frequency;
     u32 bus_frequency;
+
+    bool sse : 1;
```
lioncash, line 32: Seems odd to optimize for size when there's only going to be one instance of it active (and even then, the codegen difference isn't even that big, and usually adds an extra and to accesses).

Wunkolo: This was more of a "why not" from me since I planned to add more ISA feature-flags without each `bool` costing a whole byte but I am finding that the compact bitfields allows for multiple flags to be checked in parallel a bit faster than if it were byte-sized `bool`s.

https://godbolt.org/z/Gxxbfhzv9
With bitfields:
```asm
CheckABF(BunchaFlags const&):             # @CheckABF(BunchaFlags const&)
        mov     al, byte ptr [rdi]
        not     al
        test    al, 35
        sete    al
        ret
```

https://godbolt.org/z/jKeT6srzb
Without bitfields:
```asm
CheckABF(BunchaFlags const&):             # @CheckABF(BunchaFlags const&)
        xor     eax, eax
        cmp     byte ptr [rdi], 0
        je      .LBB0_3
        cmp     byte ptr [rdi + 1], 0
        je      .LBB0_3
        cmp     byte ptr [rdi + 5], 0
        setne   al
.LBB0_3:
        ret
```

lioncash: Right, but in most use cases, capabilities would only be accessed from external code to check for a single feature (iirc, the only occurrence of checks in sequence we have currently are when we're using it to construct strings, which makes the codegen more or less negligible)

Wunkolo: In cases of checking for one feature or multiple in-sequence, I'm seeing better assembly with bitfields.
Like loading all the compacted-flags at once into a register up-front with just one memory load and then testing against it(with the `test`-instruction) in-place, rather than loading-in and checking a new byte of data for each individual flag.

Not seeing any kind of net-loss anywhere. But I can remove it back to byte-based bools if I'm overlooking a regression somewhere.

Here's a unit case.
https://godbolt.org/z/oWWKnobb6

lioncash: The problem with this test is it is assuming the internal static will be visible to external code.

Consider: https://godbolt.org/z/9En6E7zef which treats it like how the API is currently set up.

Though, given it generates roughly equivalent code (only diff being test/cmp), I'll leave it up to you with whichever you wanna go with.

bunnei (approved):

PR #7969, opened 2022-03-01
gl_graphics_pipeline: Improve shader builder synchronization using fences

ameerj:
This PR aims to alleviate reported issues with async shaders on OpenGL where graphical issues persist during the session.

The challenge with async shaders on OpenGL is that there are two levels of asynchronous execution happening:
- The CPU shader builder thread runs concurrently, and has to coordinate with the main thread to inform it when the thread completed its work
- OGL commands are asynchronous by nature, and run concurrently with the CPU code.

So, even though the shader builder CPU thread completes its CPU work, it doesn't imply that the OGL commands it issued have been completed by the OGL context it shares with with main thread. 

This PR improves on async shaders by making use of `GLsync` fences to better coordinate when the builder thread has completed the OpenGL commands in its context. This is an improvement from before since this ensures the GL commands have been fully executed once the `GLsync` fence was reached, and these changes have been propagated to the contexts which `Wait` on the fence.

### DIFF
--- a/src/video_core/renderer_opengl/gl_graphics_pipeline.cpp
+++ b/src/video_core/renderer_opengl/gl_graphics_pipeline.cpp
@@ -229,10 +229,9 @@ GraphicsPipeline::GraphicsPipeline(
     if (key.xfb_enabled && device.UseAssemblyShaders()) {
         GenerateTransformFeedbackState();
     }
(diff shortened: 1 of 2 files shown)


FernandoS27:
Squahs those commits into 1, too many for such few lines.

PR #7973, opened 2022-03-02
host_memory: Fix fastmem crashes in debug builds

Morph1984:
It is possible for virtual_offset to not be 0 when the iterator is at the beginning, and thus, `std::prev(it)` may be evaluated, leading to a crash in debug mode.

### DIFF
--- a/src/common/host_memory.cpp
+++ b/src/common/host_memory.cpp
@@ -327,8 +327,8 @@ private:
     bool IsNiechePlaceholder(size_t virtual_offset, size_t length) const {
         const auto it = placeholders.upper_bound({virtual_offset, virtual_offset + length});
         if (it != placeholders.end() && it->lower() == virtual_offset + length) {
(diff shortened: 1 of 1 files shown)


FernandoS27 (approved):

src/common/host_memory.cpp
```diff
@@ -327,8 +327,8 @@ class HostMemory::Impl {
     bool IsNiechePlaceholder(size_t virtual_offset, size_t length) const {
```
lioncash, line 327: What even is a "nieche" placeholder? Is this supposed to be "niche"?

Morph1984: I was pondering this as well, @FernandoS27 is it supposed to be "niche"?

FernandoS27: ehm ehm ehm it's like that austrian philosopher, I think..... (no idea, I did not write the original code)

well whatever, just change it if you feel like.

PR #7975, opened 2022-03-03
hle: service: ldr: Use deterministic addresses when mapping NROs.

bunnei:
- Instead of randomization, choose in-order addresses for where to map NROs into memory.
- This results in predictable behavior when debugging and consistent behavior when reproducing issues.

### DIFF
--- a/src/core/hle/service/ldr/ldr.cpp
+++ b/src/core/hle/service/ldr/ldr.cpp
@@ -288,7 +288,7 @@ public:
     }
 
     bool ValidateRegionForMap(Kernel::KPageTable& page_table, VAddr start, std::size_t size) const {
-        constexpr std::size_t padding_size{4 * Kernel::PageSize};
+        const std::size_t padding_size{page_table.GetNumGuardPages() * Kernel::PageSize};
         const auto start_info{page_table.QueryInfo(start - 1)};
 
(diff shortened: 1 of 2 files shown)


src/core/hle/service/ldr/ldr.cpp
```diff
@@ -308,31 +308,69 @@ class RelocatableObject final : public ServiceFramework<RelocatableObject> {
...
-                   (Kernel::KSystemControl::GenerateRandomRange(0, end_pages) << Kernel::PageBits);
-        } while (!page_table.IsInsideAddressSpace(addr, size) ||
-                 page_table.IsInsideHeapRegion(addr, size) ||
-                 page_table.IsInsideAliasRegion(addr, size));
-        return addr;
+    ResultCode GetAvailableMapRegion(Kernel::KPageTable& page_table, u64 size, VAddr& out_addr) {
+        size = Common::AlignUp(size, Kernel::PageSize);
+        size += page_table.GetNumGuardPages() * Kernel::PageSize * 4;
+
+        auto is_region_available = [&](VAddr addr) {
```
lioncash, line 315:
```suggestion
        const auto is_region_available = [&](VAddr addr) {
```

lioncash (commented):
LGTM otherwise

PR #7976, opened 2022-03-03
loader: log the type of mismatching file-extension

BytesGalore:
This change additionally logs the deduced type of a mismatching type/file-extension combination.
For instance loading a ```my_cool.xci``` that is actually a NSP type will log out:
```[   0.148175] Loader <Warning> core/loader/loader.cpp:GetLoader:264: my_cool.xci has a different type (NSP) than its extension.```


Morph1984:
I don't think that logging the type is particularly useful here, since an unknown type is just `FileType::Unknown`

BytesGalore:
I just thought it might be handy to see whats going on without the need to set debug loglevel.

bunnei (approved):

Issue #7412, opened 2021-11-22
Add support to rotate inputs by 90 degrees counter-clockwise


Reichmann-M:
It would be very useful to add support to the above mentioned feature.
Because if you want to emulate a single joy con (e.g. the left one) by using a DS4 / XBOX controller you awkwardly have to physically rotate your controller in your hands by 90 degrees to correctly use the single joy con in games they require it. (e.g. Super Mario Party).

Yes, you can remap all the buttons by 90 degrees. Unfortunately, you cannot remap every direction of the analog sticks. And the "invert axis" functionality only swaps the axes.

Dolphin has a checkmark for "horizontal holding" for the wiimote.
Thank you in advance :)

v1993:
This sounds pretty useful indeed. As an additional feature, correctly translating motion would also be nice.

On the hand, "invert axis" should be functioning properly, that sounds like a separate bug.

Reichmann-M:
With some trial and error by editing an input config it now works:

`lstick\default=false
lstick="axis_x:1,engine:sdl,invert_x:-,port:0,guid:030000005e0400008e02000000007801,range:0.950000,pad:0,threshold:0.500000,axis_y:0,deadzone:0.150000"`

But it would still be better to have an extra button in settings.

german77:
If you map your sticks inverted (up then left) they will be swapped.

Reichmann-M:
> With some trial and error by editing an input config it now works:
> 
> `lstick\default=false lstick="axis_x:1,engine:sdl,invert_x:-,port:0,guid:030000005e0400008e02000000007801,range:0.950000,pad:0,threshold:0.500000,axis_y:0,deadzone:0.150000"`
> 
> But it would still be better to have an extra button in settings.

This is actually not correct. This is exactly the opposite way...

The right code is here:

`lstick\default=false
lstick="engine:sdl,axis_x:1,port:0,invert_x:+,guid:030000005e0400008e02000000007801,axis_y:0,offset_x:-0.000000,offset_y:-0.000031,invert_y:-"`

Papaj-kremufka:
how do i use the code

german77:
I'm aware. The reason it's pretty simple. When you rotate by 90° the Y axis ends up facing left instead of right. This means the x axis needs to be inverted to be mapped properly.
![image](https://user-images.githubusercontent.com/5944268/156800962-1a2ddfe2-4dd9-4d18-acd2-a9fff3f472c9.png)

PR #7978, opened 2022-03-04
input_common: Map sticks correctly when mapped sideways

german77:
Some people like to use single joycons and not turning their xbox controller sideways. Originally trying to manually map a controller in this configuration resulted in inverted axis that had some extra steps to get it working properly.

This PR adds the necessary logic to correct the axis direction to ensure it always goes to where you want it to go.

solves #7412

### DIFF
--- a/src/input_common/drivers/sdl_driver.cpp
+++ b/src/input_common/drivers/sdl_driver.cpp
@@ -934,4 +934,37 @@ u8 SDLDriver::GetHatButtonId(const std::string& direction_name) const {
     return direction;
 }
 
+bool SDLDriver::IsStickInverted(const Common::ParamPackage& params) {
+    if (!params.Has("guid") || !params.Has("port")) {
+        return false;
+    }
+    const auto joystick = GetSDLJoystickByGUID(params.Get("guid", ""), params.Get("port", 0));
+    if (joystick == nullptr) {
+        return false;
+    }
+    auto* controller = joystick->GetSDLGameController();
+    if (controller == nullptr) {
+        return false;
+    }
+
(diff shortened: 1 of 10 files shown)


src/input_common/drivers/sdl_driver.cpp
```diff
@@ -259,6 +259,20 @@ std::shared_ptr<SDLJoystick> SDLDriver::GetSDLJoystickByGUID(const std::string&
     return joystick_map[guid].emplace_back(std::move(joystick));
 }
 
+std::shared_ptr<SDLJoystick> SDLDriver::GetSDLJoystickByGUID(const std::string& guid,
+                                                             int port) const {
+    const auto it = joystick_map.find(guid);
```
lioncash, line 264: Shouldn't the mutex for the joystick map be locked before accessing it?

src/input_common/drivers/sdl_driver.h
```diff
@@ -47,6 +47,7 @@ class SDLDriver : public InputEngine {
      * tie it to a SDLJoystick with the same guid and that port
      */
     std::shared_ptr<SDLJoystick> GetSDLJoystickByGUID(const std::string& guid, int port);
+    std::shared_ptr<SDLJoystick> GetSDLJoystickByGUID(const std::string& guid, int port) const;
```
lioncash, line 50: I'd strongly suggest renaming one of these to something else, since one creates the joystick if it doesn't exist and the other one doesn't, so the behavior isn't exactly consistent between the two.

bunnei (approved):

PR #7985, opened 2022-03-07
shader: rewrite LOP3.LUT

asLody:
This change rewrites the **LOP3.LUT** instruction.

At IR level, this instruction now no longer generates a lot of BitwiseAnd & BitwiseOr, but preserves the original semantics of the shader.

The new approach not only **improves shader performance**, but is necessary for **vulkan games such as hades**, which tend to use multiple LOP3 instructions before `BindlessImage*`. The original implementation would not be able to match the correct texture handle and sampler handle.

### DIFF
--- a/src/shader_recompiler/frontend/maxwell/translate/impl/logic_operation_three_input.cpp
+++ b/src/shader_recompiler/frontend/maxwell/translate/impl/logic_operation_three_input.cpp
@@ -2,6 +2,30 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+// This files contains code from Ryujinx
+// A copy of the code can be obtained from https://github.com/Ryujinx/Ryujinx
+// The sections using code from Ryujinx are marked with a link to the original version
+
+// MIT License
+//
+// Copyright (c) Ryujinx Team and Contributors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+// associated documentation files (the "Software"), to deal in the Software without restriction,
+// including without limitation the rights to use, copy, modify, merge, publish, distribute,
+// sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
+// NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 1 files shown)


src/shader_recompiler/frontend/ir/ir_emitter.h
```diff
@@ -225,6 +225,7 @@ class IREmitter {
     [[nodiscard]] U32 BitReverse(const U32& value);
     [[nodiscard]] U32 BitCount(const U32& value);
     [[nodiscard]] U32 BitwiseNot(const U32& value);
+    [[nodiscard]] U32 BitwiseNot(const U32& value, bool inverted);
```
v1993, line 228: This overload is a pretty confusing one. IMO it would be a good idea to drop it and just use ternary operator in three places where it's currently invoked (or make it a function-local lambda there, given how similar all invocations are).

asLody: Now it's better.

FernandoS27:
Even though it follows their license considering yuzu is always shipped along the source code. I would prefer to avoid using their optimization.

We can actually use a K'Map for this and/or Quine–McCluskey_algorithm. It will cost some compute power but I'm sure it will drop down to something very simple.

ameerj:
Please add some test cases in our [shader unit tests](https://github.com/ReinUsesLisp/nxgpucatch/blob/master/src/tests/shader/lop3.cpp) that mimic the inputs which cause incorrect behavior.

src/shader_recompiler/frontend/maxwell/translate/impl/logic_operation_three_input.cpp
```diff
@@ -2,6 +2,30 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+// This files contains code from Ryujinx
+// A copy of the code can be obtained from https://github.com/Ryujinx/Ryujinx
+// The sections using code from Ryujinx are marked with a link to the original version
+
+// MIT License
```
lioncash, line 9: This can be shortened to:

```cpp
// Copyright (c) Ryujinx Team and Contributors
// SPDX-License-Identifier: MIT
```

To make it a little less noisy

degasus:
I acutally wonder about the `The original implementation would not be able to match the correct texture handle and sampler handle.` statement. Is this patch a pure performance optimization or does it modify the behavior as well?

About the algorithm itself, the KMap is even better than Gray as Fernando suggested. However we're not targeting `AND,OR,NOT`, we're targeting `AND,OR,XOR` and perhaps also `AND NOT, NOT AND, ...` (depends on the host GPU). Especially the XOR makes a large difference here, so there is a good chance that we can get a lot better optimizations.

vonchenplus:
Yes, I also noticed to Bindless texture not be able to match the correct texture handle and sampler handle.

@degasus Did your patch fix this problem?

asLody:
### Let me explain in more detail:
Optimising LOP3 just **outputs a better looking IR**, it doesn't change the behaviour.
But this will be a prerequisite for the future improvement of texture pass.

degasus:
Honestly, I have no clue if my patch fixes anything. But it should provide at least the same kind of optimization as this PR here. It would be good if someone with better testing abilities is able to verfy it.

@asLody Do you have a preference of which patch is better?

asLody:
> Honestly, I have no clue if my patch fixes anything. But it should provide at least the same kind of optimization as this PR here. It would be good if someone with better testing abilities is able to verfy it.
> 
> @asLody Do you have a preference of which patch is better?

Your PR is doing better and I will close this.

degasus:
Thanks for the testing, and a lot of thanks for pointing to the issue. Debugging whe source of such issues is often harder than fixing them. So regardless of which patch which be used in the long term, you made the larger part of this contribution.

I'm looking forward for your next pull request!

PR #7986, opened 2022-03-07
core, video_core: Fix two crashes when failing to create the emulated GPU instance

lat9nq:
This PR aims to fix two crashes when yuzu fails to create the emulated GPU instance (e.g. no Vulkan driver is present). The first is when `CreateGPU` fails yuzu would try and shutdown the GPU instance regardless of whether any instance was actually created. 

Secondly, when `CreateRenderer` fails, the `GraphicsContext` that was `std::move`'d into it is destroyed before the `Scoped` that was created to manage its currency. In that case, the `GraphicsContext::Scoped` will still call its destructor at the ending of the function. And because the `context` is destroyed, the `Scoped` will cause a crash as it attempts to call a destroyed object's `DoneCurrent` function.

~~The first issue if fixed by checking for `nullptr` before calling those methods. The second is fixed by avoiding using any `Scoped` and instead manage the context currency manually through a functor.~~ See commit messages

For Vulkan, `DoneCurrent` is not redefined so this is to manage an empty function -- it appears that it's purely for compatibility with OpenGL.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -326,7 +326,9 @@ struct System::Impl {
         is_powered_on = false;
         exit_lock = false;
 
-        gpu_core->NotifyShutdown();
+        if (gpu_core != nullptr) {
+            gpu_core->NotifyShutdown();
+        }
 
         services.reset();
         service_manager.reset();
--- a/src/core/frontend/emu_window.h
+++ b/src/core/frontend/emu_window.h
@@ -42,11 +42,20 @@ public:
             context.MakeCurrent();
         }
         ~Scoped() {
-            context.DoneCurrent();
+            if (active) {
+                context.DoneCurrent();
+            }
+        }
+
+        /// In the event that context was destroyed before the Scoped is destroyed, this provides a
+        /// mechanism to prevent calling a destroyed object's method during the deconstructor
+        void Cancel() {
+            active = false;
         }
 
     private:
         GraphicsContext& context;
+        bool active{true};
     };
 
     /// Calls MakeCurrent on the context and calls DoneCurrent when the scope for the returned value
--- a/src/video_core/video_core.cpp
+++ b/src/video_core/video_core.cpp
@@ -50,6 +50,7 @@ std::unique_ptr<Tegra::GPU> CreateGPU(Core::Frontend::EmuWindow& emu_window, Cor
         gpu->BindRenderer(std::move(renderer));
         return gpu;
     } catch (const std::runtime_error& exception) {
+        scope.Cancel();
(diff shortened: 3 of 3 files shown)


src/core/core.cpp
```diff
@@ -336,7 +338,9 @@ struct System::Impl {
         time_manager.Shutdown();
         core_timing.Shutdown();
         app_loader.reset();
-        gpu_core.reset();
+        if (gpu_core != nullptr) {
```
lioncash, line 341: this check isn't necessary. reset() already does this check internally to determine whether or not to destroy the object

src/core/frontend/emu_window.h
```diff
@@ -36,6 +38,11 @@ class GraphicsContext {
     /// Releases (dunno if this is the "right" word) the context from the caller thread
     virtual void DoneCurrent() {}
 
+    /// Returns a functor that calls DoneCurrent, for when Scoped cannot be used
+    std::function<void()> GetDoneCurrent() {
```
lioncash, line 42:
```suggestion
    std::function<void()> GetDoneCurrent() const {
```

src/video_core/video_core.cpp
```diff
@@ -44,10 +45,14 @@ std::unique_ptr<Tegra::GPU> CreateGPU(Core::Frontend::EmuWindow& emu_window, Cor
     const bool use_async = Settings::values.use_asynchronous_gpu_emulation.GetValue();
     auto gpu = std::make_unique<Tegra::GPU>(system, use_async, use_nvdec);
     auto context = emu_window.CreateSharedContext();
-    auto scope = context->Acquire();
+    // Cannot use a GraphicsContext::Scoped as `context` may be destroyed in CreateRenderer before
```
lioncash, line 48: Would you not just be able to call `Cancel()` on the scope guard itself in the failure case instead?

lat9nq: > Would you not just be able to call Cancel() or add the ability to cancel it to the Scoped interface

Nope, that works. This was just a case of me over-complicating things again.

src/core/frontend/emu_window.h
```diff
@@ -42,11 +42,20 @@ class GraphicsContext {
...
         }
         ~Scoped() {
-            context.DoneCurrent();
+            if (active) {
+                context.DoneCurrent();
+            }
+        }
+
+        /// In the event that context was destroyed before the Scoped is destroyed, this provides a
+        /// mechanism to prevent calling a destroyed object's method during
```
lioncash, line 51: Seems like the second line got cut off at the end

bunnei (approved):

PR #7989, opened 2022-03-07
shader_recompiler/LOP3: Use brute force python results within switch/case.

degasus:
Thanks to @asLody for optimizing this function within #7985. This raised the focus that this function should be optimized more.

The current table assumes that the host GPU is able to invert for free, so only AND,OR,XOR are accumulated in the performance metrik.

Performance results:

Instructions
0: 8
1: 30
2: 114
3: 80
4: 24

Latency
0: 8
1: 30
2: 194
3: 24

### DIFF
--- a/src/shader_recompiler/frontend/maxwell/translate/impl/logic_operation_three_input.cpp
+++ b/src/shader_recompiler/frontend/maxwell/translate/impl/logic_operation_three_input.cpp
@@ -13,59 +13,535 @@ namespace {
 // Emulate GPU's LOP3.LUT (three-input logic op with 8-bit truth table)
 IR::U32 ApplyLUT(IR::IREmitter& ir, const IR::U32& a, const IR::U32& b, const IR::U32& c,
                  u64 ttbl) {
-    IR::U32 r{ir.Imm32(0)};
-    const IR::U32 not_a{ir.BitwiseNot(a)};
-    const IR::U32 not_b{ir.BitwiseNot(b)};
-    const IR::U32 not_c{ir.BitwiseNot(c)};
-    if (ttbl & 0x01) {
-        // r |= ~a & ~b & ~c;
-        const auto lhs{ir.BitwiseAnd(not_a, not_b)};
-        const auto rhs{ir.BitwiseAnd(lhs, not_c)};
-        r = ir.BitwiseOr(r, rhs);
+    switch (ttbl) {
+        // generated code, do not edit manually
+    case 0:
+        return ir.Imm32(0);
+    case 1:
+        return ir.BitwiseNot(ir.BitwiseOr(a, ir.BitwiseOr(b, c)));
+    case 2:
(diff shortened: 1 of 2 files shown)


src/shader_recompiler/frontend/maxwell/translate/impl/logic_operation_three_input.cpp
```diff
@@ -14,56 +14,794 @@ namespace {
...
-    }
-    if (ttbl & 0x80) {
-        // r |= a & b & c;
-        const auto lhs{ir.BitwiseAnd(a, b)};
-        const auto rhs{ir.BitwiseAnd(lhs, c)};
-        r = ir.BitwiseOr(r, rhs);
+    switch (ttbl) {
+    // automatically generated code, do not edit manually
+    case 0:
+        r = ir.Imm32(0);
```
lioncash, line 20: Given how nothing is done after the switch, you can make this slightly more compact by returning in the case instead of assigning then breaking out.

Morph1984:
Supersedes #7985

src/shader_recompiler/frontend/maxwell/translate/impl/logic_operation_three_input.cpp
degasus: done

src/shader_recompiler/frontend/maxwell/translate/impl/logic_operation_three_input_lut3.py
```diff
@@ -0,0 +1,79 @@
...
+                    max(latency_a, latency_b) + 1) # max latency + 1
+    if registered == 0:
+        # No update at all? So terminate
+        break
+
+# Hacky output. Please improve me to output valid C++ instead.
+s = """    case {imm}:
+        return {op};"""
+for imm in range(256):
+    print(s.format(imm=imm, op=optimized_calls[imm][0]))
```
v1993, line 79: Missing newline at EOF

merryhime (approved):

merryhime:
I note a lot of the cases could be improved by assigning a cost to `not`, e.g.: For 0x08 I would probably prefer `ir.BitwiseAnd(ir.BitwiseNot(a), ir.BitwiseAnd(b, c)));`, but *shrug*.

degasus:
@merryhime I've added support for variable amount of arguments to the instructions and added cost(NEG) = 0.1. So it is now trying to minimize the latency and the amount if NEG calls as long as it doesn't increase the cost of AND,OR,XOR.
In case there are still two with equal cost, it prefers the one with the shorter code string.

merryhime:
@degasus NPN classification is the idea that given three transformations (**n**egating the output / **p**ermuting the inputs / **n**egating the inputs), you can classify luts into families of related circuits; it's a well known result that there are 14 such families for 3-luts.

PR #7997, opened 2022-03-10
cpu_detect: Add additional x86 flags and telemetry

Wunkolo:
Adds detection of additional CPU flags to cpu_detect and additions to telemetry output.

This is not exhaustive but guided by features that [dynarmic utilizes](https://github.com/merryhime/dynarmic/blob/bcfe377aaa5138af740e90af5be7a7dff7b62a52/src/dynarmic/backend/x64/host_feature.h#L12-L33) as well as features that are currently utilized but not reported to telemetry(invariant_tsc). This is intended to guide future optimizations.

AVX512 in particular is broken up into its individual subsets and some other processor features such as [sha](https://en.wikipedia.org/wiki/Intel_SHA_extensions) and [gfni](https://en.wikipedia.org/wiki/AVX-512#GFNI) are added to have some forward-facing data-points.

What used to be a single `CPU_Extension_x64_AVX512` telemetry field
is also broken up into individual `CPU_Extension_x64_AVX512{F,VL,CD,...}` fields.

This needs to be handled so that previous telemetry data can map somewhat to the new fields(it can mean Skylake-X level AVX512 like the [current implementation](https://github.com/yuzu-emu/yuzu/blob/9a97ef46478c999fc1220f0963847667dae795ce/src/common/x64/cpu_detect.cpp#L119-L124) or just AVX512**F** since that is the baseline of having _any_ sort of AVX512).

### DIFF
--- a/src/common/telemetry.cpp
+++ b/src/common/telemetry.cpp
@@ -55,22 +55,50 @@ void AppendBuildInfo(FieldCollection& fc) {
 
 void AppendCPUInfo(FieldCollection& fc) {
 #ifdef ARCHITECTURE_x86_64
-    fc.AddField(FieldType::UserSystem, "CPU_Model", Common::GetCPUCaps().cpu_string);
-    fc.AddField(FieldType::UserSystem, "CPU_BrandString", Common::GetCPUCaps().brand_string);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_AES", Common::GetCPUCaps().aes);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_AVX", Common::GetCPUCaps().avx);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_AVX2", Common::GetCPUCaps().avx2);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_AVX512", Common::GetCPUCaps().avx512);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_BMI1", Common::GetCPUCaps().bmi1);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_BMI2", Common::GetCPUCaps().bmi2);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_FMA", Common::GetCPUCaps().fma);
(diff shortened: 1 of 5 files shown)


src/common/telemetry.cpp
```diff
@@ -55,22 +55,42 @@ void AppendBuildInfo(FieldCollection& fc) {
...
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_FMA4", Common::GetCPUCaps().fma4);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_SSE", Common::GetCPUCaps().sse);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_SSE2", Common::GetCPUCaps().sse2);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_SSE3", Common::GetCPUCaps().sse3);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_SSSE3", Common::GetCPUCaps().ssse3);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_SSE41", Common::GetCPUCaps().sse4_1);
-    fc.AddField(FieldType::UserSystem, "CPU_Extension_x64_SSE42", Common::GetCPUCaps().sse4_2);
+
+    const auto& caps = Common::GetCPUCaps();
+    fc.AddField(FieldType::UserSystem, "CPU_Model", caps.cpu_string);
```
lioncash, line 60: Thoughts on wrapping a local lambda (or internally linked function) somewhat like:

```cpp
const auto add_field = [&fc](std::string_view feature_name, bool has_feature) {
    fc.AddField(FieldType::UserSystem, feature_name, has_feature);
};
```

so that `FieldType::UserSystem` doesn't need to be repeated several times. It's make adding to this more convenient in the future.

Wunkolo: This looks like a good idea, to remove redundant arguments. I'll also update `AddField` to take a `string_view` while I'm at it too.

Wunkolo:
Just to put it in writing, was told to keep maintaining the previous `CPU_Extension_x64_AVX512` telemetry-field to mean "Skylake-X/SP level AVX512" along-side the broken-up ones.

bunnei (approved):
LGTM, thank you!

PR #7974, opened 2022-03-03
Kernel Memory Updates (Part 5): Revamp MapCodeMemory and UnmapCodeMemory.

bunnei:
This is a continuation of my work to bring our kernel memory management up to speed with recent system updates and the latest known updates (see previous PRs #7956, https://github.com/yuzu-emu/yuzu/pull/7701, https://github.com/yuzu-emu/yuzu/pull/7698, and https://github.com/yuzu-emu/yuzu/pull/7684 for a history of this effort).

- This makes these functions more accurate to the real HOS implementations.
- Fixes memory access issues in Super Smash Bros. Ultimate that occur when un/mapping NROs.

PR #7956, opened 2022-02-26
Kernel Memory Updates (Part 4): Revamp KMemoryManager & other fixes

bunnei:
This is a continuation of my work to bring our kernel memory management up to speed with recent system updates and the latest known updates (see previous PRs #7701, #7698, and #7684 for a history of this effort).

The primary focus of this round of changes is to fully revamp the KMemoryManager implementation. Beyond making this more accurate, this fixes a major issue where we were misreporting memory pool sizes by not taking into consideration the two memory managers used for the application pool. Furthermore, we move the memory layout to the kernel state, allowing this to be accessed elsewhere without deriving it again. Lastly, there are several other misc. fixes in support of and/or related to these changes.

PR #8005, opened 2022-03-12
Kernel Memory Updates (Part 6): Use guest memory for slab heaps & update TLS.

bunnei:
This is a continuation of my work to bring our kernel memory management up to speed with recent system updates and the latest known information (see previous PRs #7974, https://github.com/yuzu-emu/yuzu/pull/7956, https://github.com/yuzu-emu/yuzu/pull/7701, https://github.com/yuzu-emu/yuzu/pull/7698, and https://github.com/yuzu-emu/yuzu/pull/7684 for a history of this effort).

With this change, we:
* Migrate slab heaps for guest kernel objects from host heap memory to emulated guest memory. We now accurately track overall system limits for these. Because they are now allocated in guest memory, it is impossible for these to cause host leaks; we would simply run out of space and abort.
* With the above, we've fixed several more kernel object leaks that affected a few games, but were largely unnoticed due to the previous implementation allowing unlimited allocations.
* Rewrite thread local storage (TLS) to be accurate based on latest known HOS behavior, which was a dependency on the above.
* Refresh our implementations of BSD tree, intrusive red/black tree, etc.
* Remove various heap allocation tracking hacks, as we can know guest memory is always freed on shutdown, which kernel objects now use. 

In addition to simply being more accurate to HOS behavior, the goal of these changes is to improve memory usage and overall emulation stability.

### DIFF
--- a/src/core/hle/kernel/k_thread_local_page.h
+++ b/src/core/hle/kernel/k_thread_local_page.h
@@ -1,0 +1,121 @@
+// Copyright 2022 yuzu Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
+
+#pragma once
+
+#include <array>
+
+#include "common/alignment.h"
+#include "common/assert.h"
+#include "common/common_types.h"
+#include "common/intrusive_red_black_tree.h"
+#include "core/hle/kernel/k_page_buffer.h"
+#include "core/hle/kernel/memory_types.h"
+#include "core/hle/kernel/slab_helpers.h"
+#include "core/hle/result.h"
+
+namespace Kernel {
+
+class KernelCore;
+class KProcess;
+
+class KThreadLocalPage final : public Common::IntrusiveRedBlackTreeBaseNode<KThreadLocalPage>,
+                               public KSlabAllocated<KThreadLocalPage> {
+public:
+    static constexpr size_t RegionsPerPage = PageSize / Svc::ThreadLocalRegionSize;
+    static_assert(RegionsPerPage > 0);
+
+public:
+    explicit KThreadLocalPage(VAddr addr = {}) : m_virt_addr(addr) {
+        for (size_t i = 0; i < m_is_region_free.size(); i++) {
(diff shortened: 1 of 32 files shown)


src/core/hle/kernel/k_thread_local_page.h
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    ResultCode Initialize(KernelCore& kernel, KProcess* process);
+    ResultCode Finalize();
+
+    VAddr Reserve();
+    void Release(VAddr addr);
+
+    bool IsAllUsed() const {
+        for (size_t i = 0; i < RegionsPerPage; i++) {
```
lioncash, line 47: This seems like it can be compressed into a `std::ranges::all_of`

src/core/hle/kernel/k_thread_local_page.h
```diff
@@ -0,0 +1,121 @@
...
+            if (m_is_region_free[i]) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    bool IsAllFree() const {
+        for (size_t i = 0; i < RegionsPerPage; i++) {
```
lioncash, line 56: Ditto

src/core/hle/kernel/k_thread_local_page.h
```diff
@@ -0,0 +1,121 @@
...
+
+class KThreadLocalPage final : public Common::IntrusiveRedBlackTreeBaseNode<KThreadLocalPage>,
+                               public KSlabAllocated<KThreadLocalPage> {
+public:
+    static constexpr size_t RegionsPerPage = PageSize / Svc::ThreadLocalRegionSize;
+    static_assert(RegionsPerPage > 0);
+
+public:
+    explicit KThreadLocalPage(VAddr addr = {}) : m_virt_addr(addr) {
+        for (size_t i = 0; i < m_is_region_free.size(); i++) {
```
lioncash, line 31:
```cpp
m_is_region_free.fill(true);
```

PR #8009, opened 2022-03-12
config: Write dynarmic exclusive memory configs

ameerj:
Ensures the configs are written and saved between boots


merryhime (approved):
thanks, missed this

PR #8010, opened 2022-03-12
vs2022

vricosti:
I have removed my VS2017/VS2019 from my pc to install VS2022 and I was curious to compile yuzu with it.

I made a pull request but it's more to show where the problems are than to be integrated inside master because my fixes might not be the right ones.
The first issue is inside externals/libressl (not in my pull request because I didin't want to fork externals/libressl just for that) and especially with the following define inside externals/libresslCMakeLists.txt:

```
if(WIN32)
	add_definitions(-Drestrict) 
```
because Windows headers files now seems to include the macro _CRTRESTRICT:

```
_Check_return_ _Ret_maybenull_ _Post_writable_byte_size_(_Count * _Size)
_ACRTIMP _CRTALLOCATOR _CRTRESTRICT
void* __cdecl _calloc_base(
    _In_ size_t _Count,
    _In_ size_t _Size
    );
```

The macro is defined like that:

```
#if defined _CRT_SUPPRESS_RESTRICT || defined _CORECRT_BUILD
    #define _CRTRESTRICT
#else
    #define _CRTRESTRICT __declspec(restrict)
#endif
```

So you caneither define _CRT_SUPPRESS_RESTRICT  or remove the add_definitions(-Drestrict).
Personally I did this:

```
if(WIN32)
	add_definitions(-D_CRT_SUPPRESS_RESTRICT)
	add_definitions(-Drestrict)
```

[src/input_common/CMakeLists.txt]
Inside boost 1.78 there is a warning inside include\boost\crc.hpp about potentially uninitialized local variable 'result' used so I have added /wd4701

### DIFF
--- a/src/input_common/CMakeLists.txt
+++ b/src/input_common/CMakeLists.txt
@@ -38,6 +38,7 @@ if (MSVC)
         /we4244 # 'conversion': conversion from 'type1' to 'type2', possible loss of data
         /we4245 # 'conversion': conversion from 'type1' to 'type2', signed/unsigned mismatch
         /we4254 # 'operator': conversion from 'type1:field_bits' to 'type2:field_bits', possible loss of data
+        /wd4701 # potentially uninitialized local variable 'result' used
     )
 else()
     target_compile_options(input_common PRIVATE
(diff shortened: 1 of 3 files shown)


merryhime:
>  Inside boost 1.78 there is a warning inside include\boost\crc.hpp about potentially uninitialized local variable 'result' used so I have added /wd4701 

Preferred correction would be to update boost.
(Locally I've just been wrapping the `#include`s with the relevant `#pragma warning`s)

lioncash:
Yeah, I don't think we should disable uninitialized variable warnings for the entire input_common library, just disabling it around the relevant includes is preferable.

PR #8012, opened 2022-03-13
externals: Update to experimental dynarmic branch acctype

merryhime:
dynarmic's A64 frontend determined access types but the x64 backend ignored these. 

Here we use `xadd`/`xchg`/`mfence` to (over?)-enforce memory ordering.

In theory this should provide general system stability improvements to enhance the user's experience

If this fixes weird threading issues I'll consider tidying this up and making this a proper PR.

More things for me to consider:

- [ ] Ensure 128-bit accesses are atomic.
- [ ] Add to A32 frontend.
- [x] re: xchg + read-only values


merryhime:
Will re-introduce when on dynarmic `master`.

PR #8013, opened 2022-03-13
Kernel Memory Updates (Part 6): Use guest memory for slab heaps & update TLS.

bunnei:
This is a continuation of my work to bring our kernel memory management up to speed with recent system updates and the latest known information (see previous PRs #7974, https://github.com/yuzu-emu/yuzu/pull/7956, https://github.com/yuzu-emu/yuzu/pull/7701, https://github.com/yuzu-emu/yuzu/pull/7698, and https://github.com/yuzu-emu/yuzu/pull/7684 for a history of this effort).

With this change, we:
* Migrate slab heaps for guest kernel objects from host heap memory to emulated guest memory. We now accurately track overall system limits for these. Because they are now allocated in guest memory, it is impossible for these to cause host leaks; we would simply run out of space and abort.
* With the above, we've fixed several more kernel object leaks that affected a few games, but were largely unnoticed due to the previous implementation allowing unlimited allocations.
* Rewrite thread local storage (TLS) to be accurate based on latest known HOS behavior, which was a dependency on the above.
* Refresh our implementations of BSD tree, intrusive red/black tree, etc.
* Remove various heap allocation tracking hacks, as we can know guest memory is always freed on shutdown, which kernel objects now use. 

In addition to simply being more accurate to HOS behavior, the goal of these changes is to improve memory usage and overall emulation stability.

### DIFF
--- a/src/common/intrusive_red_black_tree.h
+++ b/src/common/intrusive_red_black_tree.h
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include "common/alignment.h"
+#include "common/common_funcs.h"
 #include "common/parent_of_member.h"
 #include "common/tree.h"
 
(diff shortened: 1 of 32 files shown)


lioncash:
Most of the comments made in https://github.com/yuzu-emu/yuzu/pull/8005 still apply

src/common/intrusive_red_black_tree.h
```diff
@@ -252,16 +262,18 @@ namespace impl {
 } // namespace impl
 
 template <typename T, typename Default>
-using LightCompareType = std::remove_pointer_t<decltype(impl::GetLightCompareType<T, Default>())>;
+using RedBlackKeyType =
+    typename std::remove_pointer_t<decltype(impl::GetRedBlackKeyType<T, Default>())>;
```
lioncash, line 266:
```suggestion
    std::remove_pointer_t<decltype(impl::GetRedBlackKeyType<T, Default>())>;
```

SInce the typename isn't necessary with the _t variant

src/core/hle/kernel/k_port.cpp
```diff
@@ -57,7 +57,12 @@ ResultCode KPort::EnqueueSession(KServerSession* session) {
     R_UNLESS(state == State::Normal, ResultPortClosed);
 
     server.EnqueueSession(session);
-    server.GetSessionRequestHandler()->ClientConnected(server.AcceptSession());
+
+    if (auto session_ptr = server.GetSessionRequestHandler().lock(); session_ptr) {
```
lioncash, line 61:
```suggestion
    if (auto session_ptr = server.GetSessionRequestHandler().lock()) {
```

src/core/hle/kernel/k_thread_local_page.h
```diff
@@ -0,0 +1,112 @@
...
+            return 1;
+        }
+    }
+
+private:
+    constexpr VAddr GetRegionAddress(size_t i) {
+        return this->GetAddress() + i * Svc::ThreadLocalRegionSize;
+    }
+
+    constexpr bool Contains(VAddr addr) {
```
lioncash, line 95:
```suggestion
    constexpr bool Contains(VAddr addr) const {
```

lioncash (commented):
Did a second pass

src/core/hle/kernel/k_thread_local_page.h
```diff
@@ -0,0 +1,112 @@
...
+public:
+    static constexpr size_t RegionsPerPage = PageSize / Svc::ThreadLocalRegionSize;
+    static_assert(RegionsPerPage > 0);
+
+public:
+    explicit KThreadLocalPage(VAddr addr = {}) : m_virt_addr(addr) {
+        m_is_region_free.fill(true);
+    }
+
+    constexpr VAddr GetAddress() const {
```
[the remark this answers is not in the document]
bunnei: ctor marked as constexpr

PR #8015, opened 2022-03-13
Shader decompiler: Fix storage tracking in deko3d.

FernandoS27:
as requested by xerpi. This fixes an small bug on tracking global mem in deko3d's shaders that use use nouveau's compiler.


xerpi:
Confirmed to be working!

bunnei (approved):

PR #8023, opened 2022-03-14
maxwell_3d: Implement a safer CB data upload

ameerj:
This makes constant buffer uploads safer and more accurate by updating the GPU memory as soon as the CB Data method is invoked. The previous implementation was deferring the updates until a different maxwell 3d method was detected, then writing all CB data at once.

This fixes the geometry pop-in in `Kirby and the Forgotten Land`, graphical corruptions in `Monster Hunter Rise` and `Diablo II`, and likely other titles with long-standing graphical issues.

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -195,7 +195,7 @@ void Maxwell3D::ProcessMethodCall(u32 method, u32 argument, u32 nonshadow_argume
     case MAXWELL3D_REG_INDEX(const_buffer.cb_data) + 13:
     case MAXWELL3D_REG_INDEX(const_buffer.cb_data) + 14:
     case MAXWELL3D_REG_INDEX(const_buffer.cb_data) + 15:
-        return StartCBData(method);
+        return ProcessCBData(argument);
     case MAXWELL3D_REG_INDEX(cb_bind[0]):
         return ProcessCBBind(0);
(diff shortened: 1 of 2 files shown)


src/video_core/engines/maxwell_3d.cpp
```diff
@@ -587,61 +579,24 @@ void Maxwell3D::ProcessCBBind(size_t stage_index) {
...
     const GPUVAddr buffer_address = regs.const_buffer.BufferAddress();
     ASSERT(buffer_address != 0);
 
     // Don't allow writing past the end of the buffer.
     ASSERT(regs.const_buffer.cb_pos <= regs.const_buffer.cb_size);
 
-    const GPUVAddr address{buffer_address + cb_data_state.start_pos};
-    const std::size_t size = regs.const_buffer.cb_pos - cb_data_state.start_pos;
+    const GPUVAddr address{buffer_address + regs.const_buffer.cb_pos};
+    const u32 copy_size = amount * sizeof(u32);
```
lioncash, line 591:
```suggestion
    const size_t copy_size = amount * sizeof(u32);
```

Otherwise this is implicitly truncating to u32

FernandoS27 (approved):

gandalfthewhite19890404:
Diablo II is not fixed even after that PR was merged, all graphical bugs are in place-vertex explosions and missing text.
Tested with latest NVIDIA drivers, with both backends - opengl and vulkan.

PR #8027, opened 2022-03-15
emu_window_sdl2: Set window size to display dimensions for exclusive fullscreen

lat9nq:
Since SDL2 does not automatically resize the canvas when entering fullscreen mode, currently we resize the window to whichever native Switch display resolution we are using depending on the docked mode. Instead, resize the window to desktop display dimensions. This allows the different scaling filters as well as resolution scaling effects to appear to work properly in exclusive fullscreen.

### DIFF
--- a/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
+++ b/src/yuzu_cmd/emu_window/emu_window_sdl2.cpp
@@ -123,14 +123,15 @@ void EmuWindow_SDL2::ShowCursor(bool show_cursor) {
 }
 
 void EmuWindow_SDL2::Fullscreen() {
+    SDL_DisplayMode display_mode;
     switch (Settings::values.fullscreen_mode.GetValue()) {
     case Settings::FullscreenMode::Exclusive:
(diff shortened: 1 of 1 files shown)


dkoluris:
> Since SDL2 does not automatically resize the canvas when entering fullscreen mode, currently we resize the window to whichever native Switch display resolution we are using depending on the docked mode. Instead, resize the window to desktop display dimensions. This allows the different scaling filters as well as resolution scaling effects to appear to work properly in exclusive fullscreen.

Maybe one of the reasons why FSR did not appear to offer a benefit on certain configurations. Seems very promising change.

v1993:
> Maybe one of the reasons why FSR did not appear to offer a benefit on certain configurations. Seems very promising change.

While this technically affects FSR on certain setups, next to no people use yuzu-cmd. As such, it's very unlikely to cause any major changes - a far more likely explanation for that is that FSR is no silver bullet and is indeed pretty useless (if not outright harmful) in many cases.

bunnei (approved):

Issue #5973, opened 2021-02-21
Keyboard input / Gamecube controller input not detected in some web applets


Morph1984:
Some games' web applets exclusively use the [HTML5 Gamepad API](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API) to detect input. As a result, they cannot be navigated or accepted with keyboard input or controllers that are not detected by this API such as Gamecube controllers that are not in xinput mode.

Examples of games that use this are:
- Super Smash Bros. Ultimate
- Bioshock Trilogy

I'm not sure how someone would go about solving this issue, since supported controllers would have to be used.

As a result I am heavily considering a (per-game) setting to disable the web applet.

benderscruffy:
is this still an issue ???

Issue #5044, opened 2020-11-30
web applet: menubar and statusbar disappears in fullscreen and does not reappear


lat9nq:
In yuzu, the menubar and statusbar disappear if yuzu is already in fullscreen and the web applet is opened, and they do not reappear upon exiting fullscreen.

At least in Linux, this causes extra problems as the web applet cannot be closed via pressing ENTER nor via the web applet itself due to #5043. 

Reproduction steps:
1. Start any game
2. Enter fullscreen
3. Open a web applet
4. Exit fullscreen

Issue #5043, opened 2020-11-30
web applet: (Linux) no response to controller inputs


lat9nq:
yuzu's web applet does not respond to controller inputs in Linux.

On Mainline, these errors are found in the terminal (not the log) when pressing controller buttons, indicating that yuzu is sending something to the web engine:
`js: Uncaught TypeError: yuzu_key_callbacks[#] is not a function`

benderscruffy:
is this stil an issue ???

PR #8035, opened 2022-03-17
yuzu qt: Disable the web applet by default

lat9nq:
The web applet causes issues with the rest of the application. See #5973, #5044, and #5043. Additionally, crashes are reported, but I didn't find any documentation to link here.

Make the option save with the application, disable it by default, and add a debug option to re-enable it until a proper solution can be found.


lat9nq:
Another option would be to just set YUZU_USE_QT_WEB_ENGINE=OFF for the time being, maybe in addition to this.

bunnei (approved):

PR #8039, opened 2022-03-18
general: Reduce core.h includes

ameerj:
This reduces the dependency on `core.h` in some common header files to reduce some compile time churn.


src/core/CMakeLists.txt
```diff
@@ -210,6 +210,7 @@ add_library(core STATIC
     hle/kernel/k_memory_region_type.h
     hle/kernel/k_page_bitmap.h
     hle/kernel/k_page_buffer.h
+    hle/kernel/k_page_buffer.cpp
```
Morph1984, line 213: cpp above h

Morph1984 (approved):

PR #8048, opened 2022-03-19
general: Reduce unused includes across the project

ameerj:
This change aims to reduce compile-time churn caused by includes which go unused. 

This was achieved with the help of [Resharper C++](https://www.jetbrains.com/resharper-cpp/).


src/core/crypto/key_manager.h
```diff
@@ -16,6 +15,15 @@
 #include "common/common_types.h"
 #include "core/crypto/partition_data_manager.h"
 
+#ifdef _MSC_VER
+namespace std::filesystem {
```
lioncash, line 19: The standard explicitly forbids forward declaring standard types (except when provided, e.g. `<iosfwd>` and there's no guarantee this will always work (since these types can be moved into inline namespaces for versioning purposes and then exported in the interface depending on multiple factors).

I'd strongly suggest not doing this.

lioncash (changes requested):

src/shader_recompiler/frontend/maxwell/translate/impl/half_floating_point_set.cpp
```diff
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include "common_funcs.h"
```
lioncash, line 5:
```suggestion
#include "common/common_funcs.h"
```

src/shader_recompiler/frontend/maxwell/translate/impl/half_floating_point_set_predicate.cpp
```diff
@@ -2,6 +2,7 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include "common_funcs.h"
```
lioncash, line 5:
```suggestion
#include "common/common_funcs.h"
```

lioncash (commented):
just two nits, but looks good otherwise

src/shader_recompiler/frontend/maxwell/translate/impl/half_floating_point_set.cpp
ameerj: Thanks, good catch!

ameerj: Ah wait sorry this was including a different `common_funcs.h`
https://github.com/yuzu-emu/yuzu/blob/master/src/shader_recompiler/frontend/maxwell/translate/impl/common_funcs.h

lioncash (approved):

bunnei (approved):

PR #8050, opened 2022-03-19
Rewrite of the NVFlinger implementation

bunnei:
This is a rewrite of "NVFlinger", our HLE implementation of the service component responsible for managing buffers used for frame presentation. Our previous implementation was largely based on guesswork early on in the project (2018-2019). The new implementation is based on the Android Open Source Project (AOSP), which Switch also uses for this functionality. See individual files for citations to their AOSP counterparts.

The referenced Android code is licensed under the Apache License, Version 2.0, which is not compatible with GPLv2, but is compatible with GPLv3. Because of this, the project must be licensed as GPLv3+ going forward, assuming this change is merged. Because yuzu historically has been GPLv2 or any later version (GPLv2+), and so was Citra before that, this is a fairly benign licensing change, and grants us more flexibility. As far as I can tell, there are no GPLv2-only dependencies that prevent us from doing this, and no major drawbacks to making this change.

In addition to being much more accurate to the Switch implementation, the main change that you will see with this is improvements to overall smoothness of emulation. Our previous implementation would block the calling thread until a frame was ready for presentation, which not only is incorrect, but could cause stuttering.


src/core/hle/service/nvflinger/binder.h
```diff
@@ -0,0 +1,42 @@
...
+#pragma once
+
+#include "common/common_types.h"
+
+namespace Kernel {
+class HLERequestContext;
+class KReadableEvent;
+} // namespace Kernel
+
+namespace android {
```
lioncash, line 16: Given this (and several other files) are added to the services, this namespace should still be within the overarching `Service` namespace

src/core/hle/service/nvflinger/buffer_queue_consumer.cpp
```diff
@@ -0,0 +1,225 @@
...
+                                          u64 max_frame_number) {
+    s32 num_dropped_buffers{};
+
+    std::shared_ptr<IProducerListener> listener;
+    {
+        std::unique_lock lock(core->mutex);
+
+        // Check that the consumer doesn't currently have the maximum number of buffers acquired.
+        s32 num_acquired_buffers{};
+        for (const auto& slot : slots) {
```
lioncash, line 30: this seems like a `count_if`

src/core/hle/service/nvflinger/buffer_queue_consumer.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include "core/hle/service/nvflinger/producer_listener.h"
+
+namespace android {
+
+BufferQueueConsumer::BufferQueueConsumer(std::shared_ptr<BufferQueueCore> core_)
+    : core{std::move(core_)}, slots{core->slots} {}
+
+BufferQueueConsumer::~BufferQueueConsumer() = default;
+
+Status BufferQueueConsumer::AcquireBuffer(BufferItem* out_buffer, s64 expected_presenst_ns,
```
lioncash, line 20: Any reason to not use `<chrono>` here to encode the unit in the type as opposed to using a raw primitive?

src/core/hle/service/nvflinger/buffer_queue_core.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+#include "common/assert.h"
+
+#include "core/hle/service/nvflinger/buffer_queue_core.h"
+
+namespace android {
+
+BufferQueueCore::BufferQueueCore() : lock{mutex} {
+    // This is locked on creation, so unlock.
+    lock.unlock();
```
lioncash, line 15: This should be able to be removed by initializing the lock like:

```cpp
lock{mutex, std::defer_lock}
```

src/core/hle/service/nvflinger/buffer_queue_core.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    for (s32 slot = 0; slot < BufferQueueDefs::NUM_BUFFER_SLOTS; ++slot) {
+        free_slots.insert(slot);
+    }
+}
+
+void BufferQueueCore::NotifyShutdown() {
+    std::unique_lock<std::mutex> lk(mutex);
```
lioncash, line 23:
```suggestion
    std::unique_lock lk(mutex);
```

src/core/hle/service/nvflinger/buffer_queue_core.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+bool BufferQueueCore::StillTracking(const BufferItem* item) const {
+    const BufferSlot& slot = slots[item->slot];
+
+    return (slot.graphic_buffer != nullptr) && (item->graphic_buffer == slot.graphic_buffer);
+}
+
+void BufferQueueCore::WaitWhileAllocatingLocked() const {
+    while (is_allocating) {
+        std::unique_lock<std::mutex> lk(mutex);
```
lioncash, line 130:
```suggestion
        std::unique_lock lk(mutex);
```

src/core/hle/service/nvflinger/buffer_queue_core.cpp
```diff
@@ -0,0 +1,135 @@
...
+void BufferQueueCore::FreeAllBuffersLocked() {
+    queue.clear();
+    buffer_has_been_queued = false;
+
+    for (s32 slot = 0; slot < BufferQueueDefs::NUM_BUFFER_SLOTS; ++slot) {
+        FreeBufferLocked(slot);
+    }
+}
+
+bool BufferQueueCore::StillTracking(const BufferItem* item) const {
```
lioncash, line 122: Shouldn't this be taking the BufferItem by reference, since based off the code below, it can't be null.

src/core/hle/service/nvflinger/buffer_queue_core.h
```diff
@@ -0,0 +1,98 @@
...
+class IProducerListener;
+
+class BufferQueueCore final {
+    friend class BufferQueueProducer;
+    friend class BufferQueueConsumer;
+
+public:
+    static constexpr s32 INVALID_BUFFER_SLOT = BufferItem::INVALID_BUFFER_SLOT;
+
+    BufferQueueCore();
```
lioncash, line 34: Perhaps also default the destructor in the cpp file, given the large amount of objects that need to be destructed.

src/core/hle/service/nvflinger/buffer_queue_producer.cpp
```diff
@@ -0,0 +1,939 @@
...
+            }
+        }
+    }
+
+    return Status::NoError;
+}
+
+Status BufferQueueProducer::DequeueBuffer(s32* out_slot, Fence* out_fence, bool async, u32 width,
+                                          u32 height, PixelFormat format, u32 usage) {
+    { BufferQueueCore::AutoLock lock(core); }
```
lioncash, line 232: Perhaps provide a comment as to why this is locked and then immediately unlocked

src/core/hle/service/nvflinger/buffer_queue_producer.cpp
```diff
@@ -0,0 +1,939 @@
...
+
+    LOG_DEBUG(Service_NVFlinger, "async={} w={} h={} format={}, usage={}", async ? "true" : "false",
+              width, height, format, usage);
+
+    if ((width && !height) || (!width && height)) {
```
lioncash, line 237:
```suggestion
    if ((width != 0 && height == 0) || (width == 0 && height != 0)) {
```

To be explicit about these being numeric types

src/core/hle/service/nvflinger/consumer_base.h
```diff
@@ -0,0 +1,59 @@
...
+
+    virtual void OnFrameAvailable(const BufferItem& item) override;
+    virtual void OnFrameReplaced(const BufferItem& item) override;
+    virtual void OnBuffersReleased() override;
+    virtual void OnSidebandStreamChanged() override;
+
+    void FreeBufferLocked(s32 slot_index);
+    Status AcquireBufferLocked(BufferItem* item, u64 present_when_ns, u64 max_frame_number = 0);
+    Status ReleaseBufferLocked(s32 slot, const std::shared_ptr<GraphicBuffer> graphic_buffer);
+    bool StillTracking(s32 slot, const std::shared_ptr<GraphicBuffer> graphic_buffer);
```
lioncash, line 39:
```suggestion
    bool StillTracking(s32 slot, const std::shared_ptr<GraphicBuffer> graphic_buffer) const;
```

src/core/hle/service/nvflinger/graphic_buffer_producer.h
```diff
@@ -0,0 +1,78 @@
...
+    NativeWindowTransform transform{};
+    u32 sticky_transform{};
+    s32 async{};
+    s32 swap_interval{};
+    Fence fence{};
+};
+#pragma pack(pop)
+static_assert(sizeof(QueueBufferInput) == 84, "QueueBufferInput has wrong size");
+
+#pragma pack(push, 1)
```
lioncash, line 51: This pragma isn't necessary, since there's no unused bits that would be padded

src/core/hle/service/nvflinger/parcel.h
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        return {};
+    }
+
+    std::u16string ReadInterfaceToken() {
+        [[maybe_unused]] const u32 unknown = Read<u32>();
+        const u32 length = Read<u32>();
+
+        std::u16string token{};
```
lioncash, line 92: perhaps reserve the space for the string before entering the loop to append each character

src/core/hle/service/nvflinger/ui/rect.h
```diff
@@ -0,0 +1,75 @@
...
+#pragma once
+
+#include <cstdint>
+#include <utility>
+
+#include "common/common_types.h"
+
+namespace android {
+
+class Rect final {
```
lioncash, line 16: Wouldn't it make sense to just extend the Rectangle class we have in common instead of adding another variant?

src/core/hle/service/nvflinger/buffer_queue_consumer.cpp
ameerj, line 20:
```suggestion
Status BufferQueueConsumer::AcquireBuffer(BufferItem* out_buffer, s64 expected_present_ns,
```

src/core/hle/service/nvflinger/buffer_queue_producer.cpp
ameerj: This can be simplified to an xor
```suggestion
    if (width != 0 ^ height != 0) {
```

src/core/hle/service/nvflinger/buffer_queue_core.h
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+    private:
+        std::shared_ptr<BufferQueueCore>& core;
+    };
+
+private:
+    mutable std::mutex mutex;
+    mutable std::unique_lock<std::mutex> lock;
+    bool is_abandoned{};
```
ameerj, line 69: It's not clear to me where `is_abandoned` is ever assigned outside this default initialization

src/core/hle/service/nvflinger/consumer_base.h
```diff
@@ -0,0 +1,59 @@
...
+    struct Slot final {
+        std::shared_ptr<GraphicBuffer> graphic_buffer;
+        Fence fence;
+        u64 frame_number{};
+    };
+
+protected:
+    std::array<Slot, BufferQueueDefs::NUM_BUFFER_SLOTS> slots;
+
+    bool is_abandoned{};
```
ameerj, line 52: It's not clear to me where `is_abandoned` is ever assigned outside the default initialization

src/core/hle/service/vi/display/vi_display.cpp
```diff
@@ -13,14 +13,34 @@
...
+    std::shared_ptr<android::BufferQueueCore> core;
+    std::unique_ptr<android::BufferQueueProducer> producer;
+    std::unique_ptr<android::BufferQueueConsumer> consumer;
+};
+
+static BufferQueue CreateBufferQueue(KernelHelpers::ServiceContext& service_context) {
+    auto buffer_queue_core = std::make_shared<android::BufferQueueCore>();
+    return {buffer_queue_core,
+            std::make_unique<android::BufferQueueProducer>(service_context, buffer_queue_core),
+            std::make_unique<android::BufferQueueConsumer>(buffer_queue_core)};
```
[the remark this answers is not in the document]
bunnei: buffer_queue_core is an arg for each

src/core/hle/service/nvflinger/buffer_queue_producer.cpp
lioncash: This is significantly less readable or practical to the general reader

src/core/hle/service/nvflinger/buffer_queue_core.h
bunnei: Removed in  later commit

src/core/hle/service/nvflinger/consumer_base.h
bunnei: Removed in a later commit

src/common/math_util.h
```diff
@@ -20,37 +21,76 @@ struct Rectangle {
...
         return Rectangle{left, top + y, right, bottom + y};
     }
 
-    [[nodiscard]] Rectangle<T> Scale(const float s) const {
+    [[nodiscard]] constexpr Rectangle<T> Scale(const float s) const {
         return Rectangle{left, top, static_cast<T>(static_cast<float>(left + GetWidth()) * s),
                          static_cast<T>(static_cast<float>(top + GetHeight()) * s)};
     }
+
+    [[nodiscard]] constexpr bool operator==(const Rectangle<T>& rhs) const {
```
lioncash, line 78: This should be able to be defaulted:

```suggestion
    [[nodiscard]] constexpr bool operator==(const Rectangle& rhs) = default
```

src/core/hle/service/nvflinger/consumer_base.h
```diff
@@ -0,0 +1,61 @@
...
+
+class BufferItem;
+class BufferQueueConsumer;
+
+class ConsumerBase : public IConsumerListener, public std::enable_shared_from_this<ConsumerBase> {
+public:
+    void Connect(bool controlled_by_app);
+
+protected:
+    ConsumerBase(std::unique_ptr<BufferQueueConsumer> consumer_);
```
lioncash, line 29:
```suggestion
    explicit ConsumerBase(std::unique_ptr<BufferQueueConsumer> consumer_);
```

src/core/hle/service/vi/vi.cpp
```diff
@@ -57,447 +60,24 @@ struct DisplayInfo {
...
 private:
-    u32_le value;
+    const u32 magic = 2;
+    const u32 process_id = 1;
+    const u32 id;
+    const INSERT_PADDING_WORDS(3);
+    const std::array<u8, 8> dispdrv = {'d', 'i', 's', 'p', 'd', 'r', 'v', '\0'};
+    const INSERT_PADDING_WORDS(2);
 };
+static_assert(sizeof(NativeWindow) == 0x28, "ParcelData has wrong size");
```
lioncash, line 75:
```suggestion
static_assert(sizeof(NativeWindow) == 0x28, "NativeWindow has wrong size");
```

FernandoS27:
@bunnei you forgot to change the license file, including adapting the license exception we gave Skyline.

goldenx86:
Possible cause for crashes on Pokémon Sword/Shield on Extreme GPU accuracy. Mainline is not affected, nor are other Pokémon titles.

bunnei:
No repro on Pokemon Sword/Shield issues, so we are chalking that up to something else.

PR #8055, opened 2022-03-20
Maxwell3D: Use vertex array limit calculator instead of estimate.

FernandoS27:
This is a proper implementation of a vertex array size calculator instead of using an approximation. Therefore, it's mostly accurate (except flushing). This accelerates SM3DAS without hacking at an small cost. Currently we only do it on "small indices" as it may be more expensive on other 

TODO: figure what's wrong in linux and fastmem.

Replaces: https://github.com/yuzu-emu/yuzu/pull/8033

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -2,9 +2,13 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <algorithm>
 #include <cstring>
 #include <optional>
+
+#include "common/alignment.h"
 #include "common/assert.h"
+#include "common/settings.h"
 #include "core/core.h"
 #include "core/core_timing.h"
 #include "video_core/dirty_flags.h"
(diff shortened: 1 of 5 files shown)


src/video_core/engines/maxwell_3d.cpp
```diff
@@ -667,4 +672,71 @@ void Maxwell3D::ProcessClearBuffers() {
...
+        ptr = reinterpret_cast<u8*>(vn_state.index_buffer_cache.data());
+        memory_manager.ReadBlockUnsafe(start_address + offset, ptr, expected_size);
+    }
+    vn_state.current_max_index = index_count;
+
+    u32 new_num_vertices{};
+    switch (regs.index_array.format) {
+    case Regs::IndexFormat::UnsignedByte: {
+        std::span<const u8> span{ptr, num_indices};
+        const auto max = std::max_element(span.begin(), span.end());
```
lioncash, line 710: `<algorithm>` needs to be included for this

FernandoS27:
don't merge as there's  a better solution than this.

Morph1984:
Closing in favor of #8074

PR #8066, opened 2022-03-22
codec: Fix GPU decoder detection on Windows and a memory leak

ameerj:
It was reported that decoders with the `HW_FRAMES_CTX` config  method causes crashes on certain Linux decoding backends, hence the check to avoid it. This check subsequently caused Windows GPU decoders to never be selected and always fall back to CPU decoding. 

This PR disables `HW_FRAMES_CTX` config check on Windows for now.

This PR also addresses a memory leak relating to the allocation of GPU decoder contexts for the desired decoders which was not being freed.

### DIFF
--- a/src/video_core/command_classes/codecs/codec.cpp
+++ b/src/video_core/command_classes/codecs/codec.cpp
@@ -98,6 +98,8 @@ bool Codec::CreateGpuAvDevice() {
             LOG_DEBUG(Service_NVDRV, "{} explicitly unsupported", av_hwdevice_get_type_name(type));
             continue;
         }
+        // Avoid memory leak from not cleaning up after av_hwdevice_ctx_create
+        av_buffer_unref(&av_gpu_decoder);
         const int hwdevice_res = av_hwdevice_ctx_create(&av_gpu_decoder, type, nullptr, nullptr, 0);
         if (hwdevice_res < 0) {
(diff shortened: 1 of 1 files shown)


v1993 (commented):
This is a step in the right direction, but I'm a little skeptical about issues on Unix (and if they indeed exist, affected users can just pick CPU decoding, instead of forcing everyone to use it). Can you pass @goldenx86 build without protection against such decoders for testing?

ameerj:
@v1993 I haven't had the chance to confirm if linux users also always fallback to CPU decoding as a result of the safety check. But there are plans to implement proper support when a config method uses `HW_FRAMES_CTX` in the hopes of alleviating crashes and CPU fallback.

v1993:
Software fallback happens on Nvidia on Linux, that's something I can personally confirm. Also, from my understanding (docs are not very clear on this matter), `HW_FRAMES_CTX` refers to a *capability of decoder* rather than hard requirement to support it on our side of things. The fact that it works on Windows and, in most (all?) cases, on Linux, suggests that my interpretation is correct.

FernandoS27 (approved):

vasishath:
This change is causing my system to always fallback on software decoding. But, is also preventing CUDA backend from being used, since I have an optimus GPU which does not have an NVDECODE chip.

I had to manually bypass this check for `AV_HWDEVICE_TYPE_VAAPI` to be able to use my integrated intel for video decode.

PR #8068, opened 2022-03-22
shader_recompiler/dead_code_elimination: Add DeadBranchElimination pass

ameerj:
This adds a pass to eliminate `if(false) {...}` branches within the shader code.

This change is mainly to alleviate dead condition blocks from appearing in shader code during debug inspection. 
But may potentially improve shader code size and/or compilation time.

### DIFF
--- a/src/shader_recompiler/frontend/ir/microinstruction.cpp
+++ b/src/shader_recompiler/frontend/ir/microinstruction.cpp
@@ -326,6 +326,11 @@ void Inst::AddPhiOperand(Block* predecessor, const Value& value) {
     phi_args.emplace_back(predecessor, value);
 }
 
+void Inst::ErasePhiOperand(size_t index) {
+    const auto operand_it{phi_args.begin() + static_cast<ptrdiff_t>(index)};
+    phi_args.erase(operand_it);
+}
+
 void Inst::OrderPhiArgs() {
     if (op != Opcode::Phi) {
         throw LogicError("{} is not a Phi instruction", op);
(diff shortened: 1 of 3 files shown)


src/shader_recompiler/frontend/ir/microinstruction.cpp
```diff
@@ -326,6 +326,11 @@ void Inst::AddPhiOperand(Block* predecessor, const Value& value) {
     phi_args.emplace_back(predecessor, value);
 }
 
+void Inst::ErasePhiOperand(size_t index) {
+    const auto operand_it{phi_args.begin() + static_cast<u32>(index)};
```
lioncash, line 330: Is the cast here necessary? Shouldn't this work regardless?

ameerj: Clang didn't like it:
```
/yuzu/src/shader_recompiler/frontend/ir/microinstruction.cpp:330:46: error: implicit conversion changes signedness: 'size_t' (aka 'unsigned long') to 'boost::container::vec_iterator<std::pair<Shader::IR::Block *, Shader::IR::Value> *, false>::difference_type' (aka 'long') [-Werror,-Wsign-conversion]
    const auto operand_it{phi_args.begin() + index};
```

lioncash: ah, you may want to change the cast to `ptrdiff_t` instead, since this cast is currently truncating it and implicitly casting to `long` (weird that it doesn't warn about this, since it's still technically a sign conversion).

FernandoS27 (approved):

PR #8076, opened 2022-03-24
Vulkan: Use 3D helpers for MSAA scaling on NV drivers 510+

ameerj:
NVIDIA Vulkan drivers 510+ crash when blitting MSAA images. Fall-back to 3D scale helpers for MSAA image scaling on these drivers.

Fixes crashes when scaling was enabled on titles such as `Monster Hunter Rise`, `Rune Factory 5`, `Sonic Colors Ultimate`, among others.

### DIFF
--- a/src/video_core/renderer_vulkan/blit_image.cpp
+++ b/src/video_core/renderer_vulkan/blit_image.cpp
@@ -367,17 +367,14 @@ BlitImageHelper::BlitImageHelper(const Device& device_, VKScheduler& scheduler_,
           PipelineLayoutCreateInfo(two_textures_set_layout.address()))),
       full_screen_vert(BuildShader(device, FULL_SCREEN_TRIANGLE_VERT_SPV)),
       blit_color_to_color_frag(BuildShader(device, VULKAN_BLIT_COLOR_FLOAT_FRAG_SPV)),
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_vulkan/blit_image.cpp
```diff
@@ -367,17 +367,14 @@ BlitImageHelper::BlitImageHelper(const Device& device_, VKScheduler& scheduler_,
           PipelineLayoutCreateInfo(two_textures_set_layout.address()))),
       full_screen_vert(BuildShader(device, FULL_SCREEN_TRIANGLE_VERT_SPV)),
       blit_color_to_color_frag(BuildShader(device, VULKAN_BLIT_COLOR_FLOAT_FRAG_SPV)),
+      blit_depth_stencil_frag(BuildShader(device, VULKAN_BLIT_DEPTH_STENCIL_FRAG_SPV)),
```
FernandoS27, line 370: this shader does not work in NVIDIA since it cannot write stencil....

FernandoS27: Please only limit yourself to Resolves and Depth/stencil resolves.

FernandoS27 (changes requested):

bunnei (approved):

jacky400:
@ameerj Just tested a recent game that it crash with 512.59 driver with vullkan backend and 2x scaler.

PR #8091, opened 2022-03-26
build: use system SoundTouch when available

Tachi107:
Checking for both a CMake config file and a pkg-config file is needed because some distros only ship the latter.

Related to #6833 and #7044

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -69,7 +69,12 @@ if (YUZU_USE_EXTERNAL_SDL2)
 endif()
 
 # SoundTouch
-add_subdirectory(soundtouch)
+find_package(SoundTouch MODULE)
+if (NOT SoundTouch_FOUND)
+    message(STATUS "SoundTouch not found, falling back to externals")
+    add_subdirectory(soundtouch)
+    add_library(SoundTouch::SoundTouch ALIAS SoundTouch)
+endif()
 
 # Cubeb
 if(ENABLE_CUBEB)
--- a/externals/find-modules/FindSoundTouch.cmake
+++ b/externals/find-modules/FindSoundTouch.cmake
@@ -1,0 +1,63 @@
+# SPDX-FileCopyrightText: 2022 Andrea Pappacoda <andrea@pappacoda.it>
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+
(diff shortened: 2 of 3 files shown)


abouvier:
The SoundTouch fork vendored by yuzu use 16bit integer sample data instead of the default floating point samples. As most distros package the vanilla library with float samples, using system package won't work. You need a check, like for the CMAC support in Mbed TLS.

Or maybe yuzu can use float samples with minor modifications?

Tachi107:
Un theory simply defining `SOUNDTOUCH_INTEGER_SAMPLES` should work, as the STTypes.h header only sets `SOUNDTOUCH_FLOAT_SAMPLES` is neither are set.

In practice it doesn't work in e.g. Debian, where they ship a `soundtouch_config.h` header that overrides every other setting that sets `SOUNDTOUCH_FLOAT_SAMPLES`, resulting in both the sample times set and a compile time error.

I've submitted a patch upstream that adds a `SOUNDTOUCH_NO_CONFIG` option that disables the inclusion of the config header, but that also means that I should also add some additional check here.

abouvier:
The time stretcher code used by `cubeb` is [disabled since nearly two years](https://github.com/yuzu-emu/yuzu/commit/d68856ab124adefd0ff52f6c120ecc03aee60365) 😅
So `yuzu` doesn't currently need `SoundTouch` at all.

Tachi107:
> The time stretcher code used by `cubeb` is [disabled since nearly two years](https://github.com/yuzu-emu/yuzu/commit/d68856ab124adefd0ff52f6c120ecc03aee60365) 
> So `yuzu` doesn't currently need `SoundTouch` at all.

Uhm, interesting. But I believe that it is still needed to compile the project, right?

german77:
The library is still included. But the code is never used. Unless you actually want to fully remove it the library is still needed to compile

Tachi107:
> The library is still included. But the code is never used. Unless you actually want to fully remove the code the library is still needed to compile

Thanks, I've disabled the time stretcher code in my [packaging branch](https://salsa.debian.org/tachi/yuzu-mainline/-/blob/debian/latest/debian/patches/soundtouch-disable.patch) and everything works fine. I'll probability submit a PR later :)

lioncash:
Yeah, given how long its been disabled, I think it's fine to remove the relevant code and the dependency on SoundTouch

PR #8093, opened 2022-03-27
video_core/memory_manager: Mark memory range as remapped when mapping

ameerj:
This change reduces video memory usage, especially in `Kirby and the Forgotten Land`

### DIFF
--- a/src/video_core/memory_manager.cpp
+++ b/src/video_core/memory_manager.cpp
@@ -40,6 +40,9 @@ GPUVAddr MemoryManager::UpdateRange(GPUVAddr gpu_addr, PageEntry page_entry, std
 }
 
 GPUVAddr MemoryManager::Map(VAddr cpu_addr, GPUVAddr gpu_addr, std::size_t size) {
+    // Mark any pre-existing rasterizer memory in this range as remapped
+    rasterizer->ModifyGPUMemory(gpu_addr, size);
+
     const auto it = std::ranges::lower_bound(map_ranges, gpu_addr, {}, &MapRange::first);
     if (it != map_ranges.end() && it->first == gpu_addr) {
         it->second = size;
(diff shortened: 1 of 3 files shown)


src/video_core/memory_manager.cpp
```diff
@@ -39,7 +39,22 @@ GPUVAddr MemoryManager::UpdateRange(GPUVAddr gpu_addr, PageEntry page_entry, std
...
+        if (!cpu_vaddr) {
+            continue;
+        }
+        rasterizer->UnmapMemory(*cpu_vaddr, map_size);
+    }
+}
+
 GPUVAddr MemoryManager::Map(VAddr cpu_addr, GPUVAddr gpu_addr, std::size_t size) {
+    // Unmap any pre-existing rasterizer memory in this range
+    UnmapSubmappedRanges(gpu_addr, size);
```
FernandoS27, line 56: uhm, this should only be done if the GPU address had different backing cpu_address as the new one.

ameerj: No, because especially with the texture cache, the old data will be considered "overlapping" with the new data despite being replaced, and is not evicted. This is what Kirby does.

FernandoS27: > No, because especially with the texture cache, the old data will be considered "overlapping" with the new data despite being replaced, and is not evicted. This is what Kirby does.

Then it's not safe as you are marking dirty unnecessarily and probably overwriting memory that's in the GPU cache.

BSoDGamingYT:
This change does indeed reduce VRAM usage on Nvidia GPUs, but also causes small GFX flashes/flickers when loading new areas, on top of extreme stutters when entering new areas/when travelling though areas on Kirby and the Forgotten Land

src/video_core/memory_manager.cpp
ameerj: > Then it's not safe as you are marking dirty unnecessarily and probably overwriting memory that's in the GPU cache.

Reworked the changes.

BSoDGamingYT:
Users are reporting game flickering/flashing and crashes in game menus when graphical issues occur. I tracked the menu graphical issues to this PR

See issue and repro below:

Load game, open menu, open pokemon screen

https://user-images.githubusercontent.com/38402209/161447317-b8c20319-4800-4464-a793-cc0b61f72639.mp4

PR #8105, opened 2022-03-28
native_clock: Use AtomicLoad128

merryhime:
We implement an atomic 128-bit load using `cmpxchg16b`. (This is the only guaranteed way to have an atomic 128-bit load; `movaps` is not guaranteed to be atomic.)

We use this in `native_clock` to fix an issue with the `time_point` read not being atomic.

### DIFF
--- a/src/common/atomic_ops.h
+++ b/src/common/atomic_ops.h
@@ -46,6 +46,13 @@ namespace Common {
                                           reinterpret_cast<__int64*>(expected.data())) != 0;
 }
 
+[[nodiscard]] inline u128 AtomicLoad128(volatile u64* pointer) {
+    u128 result{};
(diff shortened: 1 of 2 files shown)


merryhime:
Added write-back variants for CAS functions to avoid the need to atomically load value (when CAS already kindly loads the new value for us).

src/common/atomic_ops.h
```diff
@@ -46,6 +46,50 @@ namespace Common {
                                           reinterpret_cast<__int64*>(expected.data())) != 0;
 }
 
+[[nodiscard]] inline bool AtomicCompareAndSwap(volatile u8* pointer, u8 value, u8 expected,
+                                               u& actual) {
```
Morph1984, line 50: u&?

merryhime: thanks

bunnei (approved):

PR #8108, opened 2022-03-29
Kernel: Improve shutdown by attempting to close open object references

bunnei:
Currently, when we tear down emulation, we do not close objects that still have open references. Since these are backed by guest memory, there is not a memory leak, but it means that finalize procedures are not invoked. With this change, we attempt to close any dangling objects on shutdown, fixing an issue where some services (e.g. HID) might not be fully torn down.


lioncash (approved):

german77:
This PR causes crashes when yuzu is closed and logs multiple critical errors on game stop.

PR #8110, opened 2022-03-29
gpu_thread: Block on FlushRegion for all GPU accuracies

ameerj:
This addresses a synchronization issue that causes an SVC break in `Kirby and the Forgotten Land`.

My own benchmarking showed the block incurs 1-2 milliseconds of overhead every 5-10 frames or so, so relatively unnoticeable. Further performance testing would be appreciated.

### DIFF
--- a/src/video_core/gpu_thread.cpp
+++ b/src/video_core/gpu_thread.cpp
@@ -90,6 +90,10 @@ void ThreadManager::FlushRegion(VAddr addr, u64 size) {
         return;
     }
     if (!Settings::IsGPULevelExtreme()) {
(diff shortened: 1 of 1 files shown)


FernandoS27:
uhm if we have to block then it means something was not flushed in time or we signaled a fence too early.

FernandoS27:
Try this: Separate this method into 2. 1 that's called by the CPU directly that does not do the sync and 1 that always does the sync and flushes as well. If this fixes the issue in High, then it's likely the flush in unmap.

ameerj:
Root cause was identified and fixed in #8116

PR #4165, opened 2020-06-25
MacOS compatibility

VHRanger:
This PR revives PR [#3593](https://github.com/yuzu-emu/yuzu/pull/3593) and updates the code to up-to-date to changes in the intervening months. There is still work to do for MacOS to be playable especially on 3d games:

- Fix macOS SDL frontend
- Much more testing needed on games.

Here are some tips to build on a MacOS machine:

- OSX version 10.14 or higher (this could be moved to 10.13)

Get [Homebrew](https://brew.sh/) and run the following on command line:

```
brew install mbedtls
brew install zstd
brew install openssl
brew install gpm
brew install pkgconfig
brew install sdl2
brew install qt5 # note if you have a QT erorr do brew remove qt than brew install qt5
brew install cmake
brew install conan
git clone --recursive https://github.com/yuzu-emu/yuzu
cd yuzu
git submodule update --init --recursive
```

When building (from `yuzu` folder):

```
export Qt5_DIR=$(brew --prefix)/opt/qt5
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Release
make -j4
```

PR #8113, opened 2022-03-29
Fix macOS build on Clang13

sguo35:
Reviving https://github.com/yuzu-emu/yuzu/pull/4165 to make yuzu work on macOS. This PR only fixes the build to compile and work on macOS, the UI frequently crashes/freezes and no games work yet. I plan to tackle the crash issues and MoltenVK integration next if this gets merged. The backcompat hacks are also a bit of a mess, would appreciate tips on how to fix it.

Build instructions:

If you're on M1, follow https://stackoverflow.com/a/68443301 to dual-install both x86 and Arm brew environments, else you will have to deal with many compiler path headaches. Make sure to run as x86 shell: `arch -x86_64 zsh`

```
# install dependencies
brew install zstd openssl pkgconfig sdl2 cmake conan autoconf automake glslang libtool qt5 ninja nasm
# brew installs libtoolize as glibtoolize so do a soft link
ln -s /usr/local/bin/glibtoolize /usr/local/bin/libtoolize
git clone --recursive git@github.com:sguo35/yuzu.git
cd yuzu
git checkout macos
git submodule update --init --recursive
```

When building make sure you have in your bash environment:
```
 export Qt5_DIR=$(brew --prefix)/opt/qt5
 export SDKROOT=$(xcrun --sdk macosx --show-sdk-path)
 export LIBRARY_PATH="$LIBRARY_PATH:/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib"
 export LIBRARY_PATH="$LIBRARY_PATH:/usr/local/opt/zstd/lib"
 export PATH="/usr/local/opt/openssl@1.1/bin:/usr/local/opt/zstd/bin:$PATH"
 export LDFLAGS="-L/usr/local/opt/openssl@1.1/lib -L/usr/local/opt/zstd/lib"
 export CPPFLAGS="-I/usr/local/opt/openssl@1.1/include -I/usr/local/opt/zstd/include"
 export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig -L/usr/local/opt/zstd/lib/pkgconfig"
 export OPENSSL_ROOT_DIR="/usr/local/opt/openssl@1.1"
 export OPENSSL_INCLUDE_DIR="/usr/local/opt/openssl@1.1/include"
```

Build:

```
mkdir build
cd build
cmake .. -GNinja
ninja
```


lioncash:
Personally, I'd rather wait until macOS' build toolchain is up to date so that patching and injecting library facilities into the standard library namespace isn't necessary

That, or require the version of gcc provided by Homebrew, which should already be up to date in terms of library support (especially for other things we may want to use in the future like atomic_ref, etc). Given how lacking clang is on macOS (or in general in terms of library support at the moment), it'd be a step down in terms of development to support it as is in my opinion.

yisonPylkita:
I'm trying to follow your steps but I'm failing miserably at linking stage. I'm getting so many errors that I can't scroll through them. Here are some
```
  "_sws_getContext", referenced from:
      Tegra::Vic::WriteRGBFrame(AVFrame const*, Tegra::VicConfig const&) in libvideo_core.a(vic.cpp.o)
  "_sws_scale", referenced from:
      Tegra::Vic::WriteRGBFrame(AVFrame const*, Tegra::VicConfig const&) in libvideo_core.a(vic.cpp.o)
ld: symbol(s) not found for architecture x86_64
```

I'm running this on M1 Air 16GB with Rosetta 2 installed. I tried both native Terminal and Terminal with Rosetta 2. I have installed Brew dependencies for both x64 and arm64. No success yet.

Please, can you package these building steps into something like `macos_install_dependencies_and_build.sh`?

sguo35:
> That, or require the version of gcc provided by Homebrew, which should already be up to date in terms of library support (especially for other things we may want to use in the future like atomic_ref, etc). Given how lacking clang is on macOS (or in general in terms of library support at the moment), it'd be a step down in terms of development to support it as is in my opinion.

Can you help me reconcile this guidance with https://github.com/yuzu-emu/yuzu/pull/4165#issuecomment-868207956? I've done some research and it doesn't seem like there are any big landmines with gcc in macOS. Is there a particular preference on this from your side? If not I can go ahead and move this to gcc.

Also for my own learning purposes, if I understand you correctly the reasons for preferring gcc over macOS clang are 1) ease of staying up to date with new C++20/23 features and 2) cleanliness of code (injecting compat patches everywhere and polluting std)?

lioncash:
> Can you help me reconcile this guidance with https://github.com/yuzu-emu/yuzu/pull/4165#issuecomment-868207956? I've done some research and it doesn't seem like there are any big landmines with gcc in macOS. Is there a particular preference on this from your side? If not I can go ahead and move this to gcc.

I haven't used GCC on macOS extensively, but if it's doable, it's generally much preferable (currently anyways), to clang from feature support alone.

> Also for my own learning purposes, if I understand you correctly the reasons for preferring gcc over macOS clang are 1) ease of staying up to date with new C++20/23 features and 2) cleanliness of code (injecting compat patches everywhere and polluting std)?

For the time being, until Apple decides to, frankly, care about the C++ ecosystem, it actively makes it hard to support macOS with their variant of clang, given how far behind it is. For comparison you can find compiler feature support [here](https://en.cppreference.com/w/cpp/compiler_support). Apple's version of clang is usually behind mainline clang in terms of support, whereas GCC generally is mostly in line with MSVC, save for a few areas, which makes it more desirable (at least until clang catches up).

sguo35:
After hacking at it for a while... long story short, gcc is not viable. Apple decided to put tons of non-standard Objective C code ([example](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78352)) in their framework headers so if you try to link to anything that uses them (e.g. IOKit) you need to be able to compile Apple ObjC. Old gcc Apple version did this, and apple-clang does this too. But OSS gcc doesn't have those ObjC extensions so even if you pass `-lobjc` you still will fail to compile.

Unfortunately if back-porting stuff with patches to std is not an option and we insist on not holding back on using new stdlib features then I think this means macOS support is non-viable as apple-clang will never be at parity with gcc and MSVC because it's always behind by a long amount of time.

FernandoS27:
Hmn, it's a tough decision. I personally would love macOS support to come back as I want to work on ARM support but @lioncash  has valid concerns. The real question here is if we are willing to sacrifice maintainability for this? I'll try discussing it with maintainers.

Ella-0:
Might be worth noting that some of these issues apply to LLVM based Linux distributions as `libc++` is still missing some required c++ 20 features

lockieluke:
please merge

lockieluke:
[](https://github.com/fujitsu/xbyak_aarch64)

this might be useful for building native binaries for M1 or ARM machines since yuzu is using an x86 version of the assembler which is not supported on ARM machines, I'll try implementing it

lioncash:
Note that just because it is implemented doesn't mean it will be supported in master (yet). Ideally we (or, in particular I, and a few other team members) would like to freeze on C++23 if we support macOS going forward, since it brings a lot of good things to the table and Apple moves at it's abysmally glacial pace when it comes to releasing updated compilers as part of its Xcode package which is a huge downside for other platforms we support.

lockieluke:
still working on fixing errors when compiling for arm64, i added a header file to fix constexpr issues cuz apple clang somehow doesn't like constexpr

lockieluke:
I'm trying to compile yuzu with LLVM's Clang instead, I'll see if it works

lockieluke:
image
there's something wrong with either the fmt library or M1 itself

patriciocs1979:
Hello I am not a developer (I was about two decades ago when I programmed some business applications in C, VisualVB and system utilities in x86 assembler) but I have a M1 Mac mini where I can do some tests. Also if you give me some directions I would be happy to contribute in any way on this Apple Silicon build.

lockieluke:
[It's alive OpenGL 4.5 on Mac OS Metal](https://www.reddit.com/r/opengl/comments/qtxgph/its_alive_opengl_45_on_mac_os_metal/)

someone implemented OpenGL 4.5 on Metal

ImFilou:
I’m not a developer, but I’m highly supporting your works. Is Metal 3 able to make MoltenVK viable to port Yuzu to MacOS? I saw this https://github.com/KhronosGroup/MoltenVK/discussions/1616
I really don’t know much about development, I just want to bring information, sorry if you were already aware of that, have a nice day

lockieluke:
I think it's absolutely possible but apple CLANG 13 won't compile some of the newer C++ features we have to add polyfills

CMakeLists.txt
```diff
@@ -6,7 +6,14 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/externals/find-module
...
 
-project(yuzu)
+if (APPLE)
+    # Hack for M1. Currently fails when compiling for arm64
+    set(CMAKE_OSX_ARCHITECTURES "x86_64" CACHE STRING "")
+    set(CMAKE_CROSSCOMPILING TRUE)
+    project(yuzu)
+else()
+    project(yuzu)
+endif()
```
Tachi107, line 17: Can't `project(yuzu)` stay out of the `if`?

```cmake
if (APPLE)
    set(CMAKE_OSX_ARCHITECTURES "x86_64" CACHE STRING "")
    set(CMAKE_CROSSCOMPILING TRUE)
endif()
project(yuzu)
```

lockieluke: Yeah

StupidRepo:
Is @sguo35 still updating this? It'd be a shame if it was abandoned.

sguo35:
> I don't think so but I'm constantly working on this

I don't want to be the bearer of bad news here, but until the team decides to standardize on a C++ version or limit their usage of new C++ features, Mac support won't happen. Clang will (probably) never be at feature parity with gcc, so supporting it is a maintainability nightmare. Not to mention MoltenVK will also be a maintainability nightmare. I can understand why the team doesn't want to support that.

Closing this PR now.

Tachi107:
> I don't want to be the bearer of bad news here, but until the team decides to standardize on a C++ version or limit their usage of new C++ features, Mac support won't happen. Apple Clang will (probably) never be at feature parity with gcc, so supporting it is a maintainability nightmare. Not to mention MoltenVK will also be a maintainability nightmare. I can understand why the team doesn't want to support that.

Why is this a problem? GCC and classic Clang are supported on Macs, right?

sguo35:
> Why is this a problem? GCC and classic Clang are supported on Macs, right?

I've only tested gcc, but in order to compile with Apple system framework headers you need to be able to compile Apple's proprietary objc extensions which OSS gcc doesn't have.

rpcs3 is fully compilable using Apple Clang.

Tachi107:
Oh right, the compiler is not enough... I don't really understand why Apple Clang is always so behind compared to regular Clang, but whatever :/

lockieluke:
> > I don't think so but I'm constantly working on this
> 
> 
> 
> I don't want to be the bearer of bad news here, but until the team decides to standardize on a C++ version or limit their usage of new C++ features, Mac support won't happen. Apple Clang will (probably) never be at feature parity with gcc, so supporting it is a maintainability nightmare. Not to mention MoltenVK will also be a maintainability nightmare. I can understand why the team doesn't want to support that.
> 
> 
> 
> Closing this PR now.

we could open a separate branch or fork specifically for macOS which might not get the latest updates, we just want macOS support I personally wouldn't mind.  If the yuzu team is not supporting this, people with the necessary knowledge would eventually open a separate fork to work on this.  You can see that Dolphin Emulator is constantly adding mac specific features, like Metal as a new graphical backend, recently.  There's certainly an increasing demand for this due to Apple Silicon making mac computers more affordable(perhaps?) and more performant at a lower price point, we could make this happen together

rcaridade145:
Why not just checkout clang from brew and use that?

lockieluke:
> Why not just checkout clang from brew and use that?

I've tried using brew's clang but didn't work due to apple's proprietary ObjC framework code

rcaridade145:
Maybe some path issues?  The idea is to exclude everything from Apple's clang

For instance (from RPCS3)

export CXX=clang++
export CC=clang
export Qt5_DIR="/usr/local/opt/qt@5/lib/cmake/Qt5"
export PATH="/usr/local/opt/llvm/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Library/Apple/usr/bin"
export LDFLAGS="-L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib"
export CXXFLAGS="-I/usr/local/opt/llvm/include -msse -msse2 -mcx16 -no-pie"
export CPLUS_INCLUDE_PATH="/usr/local/opt/molten-vk/include"
export VULKAN_SDK="/usr/local/opt/molten-vk"
export VK_ICD_FILENAMES="$VULKAN_SDK/share/vulkan/icd.d/MoltenVK_icd.json"

DarthNihilus1:
> > > I don't think so but I'm constantly working on this
> > 
> > 
> > I don't want to be the bearer of bad news here, but until the team decides to standardize on a C++ version or limit their usage of new C++ features, Mac support won't happen. Apple Clang will (probably) never be at feature parity with gcc, so supporting it is a maintainability nightmare. Not to mention MoltenVK will also be a maintainability nightmare. I can understand why the team doesn't want to support that.
> > Closing this PR now.
> 
> we could open a separate branch or fork specifically for macOS which might not get the latest updates, we just want macOS support I personally wouldn't mind. If the yuzu team is not supporting this, people with the necessary knowledge would eventually open a separate fork to work on this. You can see that Dolphin Emulator is constantly adding mac specific features, like Metal as a new graphical backend, recently. There's certainly an increasing demand for this due to Apple Silicon making mac computers more affordable(perhaps?) and more performant at a lower price point, we could make this happen together

I wouldn't mind losing out on the newest features if we can get some sort of macOS support. Some support is better than none and maybe someday in the future things will get better.

DarthNihilus1:
Could this be used by any chance to help mac support on the graphics side of things?
https://github.com/conversy/MGL/

lockieluke:
Yes absolutely but I don't have the necessary knowledge to fix the arm code converter thing I couldn't get it working

UltraHDR:
Does clang 14 in https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes help? Thanks

jimmywarting:
Keeping my hopes up for a Mac M1 port too

MichaelGDev48:
instead of brew wouldnt macports be better
Macports supports universal2 building so that will help with getting dependencies for a universal2 build (x86_64 and arm64)

PR #8122, opened 2022-03-31
Improve usage of service host threads

bunnei:
Previously, we would allocate a host thread for every HLE service interface. This was necessary because some service routines need to block, and we not support guest rescheduling from HLE host threads. This could result in dozens of host threads being created for services, which may result in weird behavior, especially on systems that have more constrained resources.

With this change, we introduce a single "default service thread" that is used for the 99% of service methods that are non-blocking. For more time-sensitive or blocking services, we still allow a thread to be created (e.g., audio, FS, nvdrv, BSD).

This should reduce service threads from ~25-50 down to ~4. The goal of this change is to improve stability and consistency, especially across systems with lower core counts. Ideally, this should not regress perf, but it is possible I missed a service that should still have a dedicated thread created.

### DIFF
--- a/src/core/hle/kernel/kernel.h
+++ b/src/core/hle/kernel/kernel.h
@@ -271,14 +271,24 @@ public:
     void ExitSVCProfile();
 
     /**
-     * Creates an HLE service thread, which are used to execute service routines asynchronously.
-     * While these are allocated per ServerSession, these need to be owned and managed outside
-     * of ServerSession to avoid a circular dependency.
+     * Creates a host thread to execute HLE service requests, which are used to execute service
+     * routines asynchronously. While these are allocated per ServerSession, these need to be owned
+     * and managed outside of ServerSession to avoid a circular dependency. In general, most
+     * services can just use the default service thread, and not need their own host service thread.
+     * See GetDefaultServiceThread.
      * @param name String name for the ServerSession creating this thread, used for debug
      * purposes.
      * @returns The a weak pointer newly created service thread.
      */
     std::weak_ptr<Kernel::ServiceThread> CreateServiceThread(const std::string& name);
+
+    /**
+     * Gets the default host service thread, which executes HLE service requests. Unless service
+     * requests need to block on the host, the default service thread should be used in favor of
+     * creating a new service thread.
+     * @returns The a weak pointer for the default service thread.
+     */
+    std::weak_ptr<Kernel::ServiceThread> GetDefaultServiceThread() const;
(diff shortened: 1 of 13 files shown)


src/core/hle/kernel/kernel.h
```diff
@@ -271,15 +271,25 @@ class KernelCore {
...
      */
     std::weak_ptr<Kernel::ServiceThread> CreateServiceThread(const std::string& name);
 
+    /**
+     * Gets the default host service thread, which executes HLE service requests. Unless service
+     * requests need to block on the host, the default service thread should be used in favor of
+     * creating a new service thread.
+     * @returns The a weak pointer for the default service thread.
+     */
+    std::weak_ptr<Kernel::ServiceThread> GetDefaultServiceThread();
```
lioncash, line 291:
```suggestion
    std::weak_ptr<Kernel::ServiceThread> GetDefaultServiceThread() const;
```

src/core/hle/service/service.h
```diff
@@ -176,14 +177,18 @@ class ServiceFramework : public ServiceFrameworkBase {
...
      * @param service_name_ Name of the service.
-     * @param max_sessions_ Maximum number of sessions that can be
-     *                      connected to this service at the same time.
+     * @param create_service_thread_ Creates a service thread for this if true, otherwise use the
+     * default service thread.
+     * @param max_sessions_ Maximum number of sessions that can be connected to this service at the
+     * same time.
      */
     explicit ServiceFramework(Core::System& system_, const char* service_name_,
+                              bool create_service_thread_ = false,
```
lioncash, line 188: Perhaps make this an enum instead?

That way boolean literals don't need to be followed by a comment indicating what the parameter is

src/core/hle/service/service.h
```diff
@@ -114,7 +114,8 @@ class ServiceFrameworkBase : public Kernel::SessionRequestHandler {
                            Kernel::HLERequestContext& ctx);
 
     explicit ServiceFrameworkBase(Core::System& system_, const char* service_name_,
-                                  u32 max_sessions_, InvokerFn* handler_invoker_);
+                                  bool create_service_thread_, u32 max_sessions_,
```
lioncash, line 117: Ditto

src/core/hle/service/service.h
Morph1984: Perhaps we can make it into a template class with enum `ServiceThreadType` set to a default value?

PR #8127, opened 2022-03-31
vk_staging_buffer_pool: Rework stream buffer usage

ameerj:
This PR aims to fix random vertex explosions and other graphical issues in titles such as `Kirby and the Forgotten Land`
when using the Vulkan backend.

The previous implementation was susceptible to off-by-one errors such as checking 
if `Region(i + 1)` is available but then consuming `Region(i)` or vice versa, or 
begin/end iterators being equal causing their usage in std algorithms to be no-ops.

This PR attempts to account for these issues by simplifying the implementation. 
The stream buffer is now divided into `NUM_STREAM_REGIONS`, with each region 
having an associated master semaphore tick indicating which tick it was being used in.

The one downside compared to the previous implementation is that each stream 
buffer region (currently 4K in size) can only be used by one upload request, no matter how small.

The previous implementation was seemingly able to pack multiple uploads into a single region,
although it's unclear if this was intentional or a consequence of the aforementioned indexing errors.


BSoDGamingYT:
I'm seeing a 15-20% performance loss in Zelda BOTW with this change, similar smaller 5-6% drops in other games like SMO and Bowsers Fury

Issue #8121, opened 2022-03-31
Pr 7720 new garbage collector regression


gidoly:
several user reported that game such as xenoblade 2, age of calamity got regression
so @goldenx86  and i track downed which pr caused it. it was gc pr as title said
while we revert pr we decide to document issue while ferdinado get time to adress it

xenoblade 2:

it got huge ram leak on open gl. while vulkan not exceed limit of vram it uses more resource then it used to be
not shown on posted picture but commited memory ewent up to 50gb

early access 2622

![yuzu_zKYku2fhI4](https://user-images.githubusercontent.com/66776795/160952548-a9673c00-babf-41ed-a6d9-b41044bd7386.jpg)

same behavior on #7720 pr artifact from github

![yuzu_sZKPNVzgtm](https://user-images.githubusercontent.com/66776795/160952644-16ca21c5-722f-4a94-a172-3da1b7f01af9.jpg)

mainline 960 which is before pr merge

![yuzu_Qke2EdsfOy](https://user-images.githubusercontent.com/66776795/160953403-1d22aa05-44a3-45bc-8a4f-9601041af009.jpg)

save of that area runing around this area for while then it will start to leak ram

[bf2option.zip](https://github.com/yuzu-emu/yuzu/files/8385054/bf2option.zip)

age of calamity:

this game got regression on getting frequent graphic break that used to rarely happend

example of broken graphic:

![yuzu_FdQ2xTKJ2B](https://user-images.githubusercontent.com/66776795/160954292-a79e7669-4218-44a4-9b6f-23528ad14064.jpg)

(not always looks same but color of game will not render properly or render all stuff black)

example of correct graphic:

![yuzu_ZtGtLKx3fb](https://user-images.githubusercontent.com/66776795/160953792-af10668a-e5eb-4f40-afe5-53643f3dff09.jpg)

on #7720 artifact it happens in 1/10 rate

while mainline 960 (before pr merge) have chance to render graphic properly 10/10

save of above pic:

[01002B00111A2000.zip](https://github.com/yuzu-emu/yuzu/files/8385060/01002B00111A2000.zip)

select  red moon icon on middle of map to enter same place with above picture

system:
5800x
32gb ram 
rtx2080s

PR #8128, opened 2022-03-31
GPU Garbage Collection: Fix regressions.

FernandoS27:
Fixes [HW: AoC Regression](https://github.com/yuzu-emu/yuzu/issues/8121)

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -282,7 +282,7 @@ void main() {
 
 u64 Device::GetCurrentDedicatedVideoMemory() const {
     GLint cur_avail_mem_kb = 0;
-    glGetIntegerv(GL_GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX, &cur_avail_mem_kb);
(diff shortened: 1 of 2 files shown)


Morph1984:
Temporarily untagged for testing

src/video_core/renderer_opengl/gl_device.cpp
```diff
@@ -282,7 +282,7 @@ void main() {
 
 u64 Device::GetCurrentDedicatedVideoMemory() const {
     GLint cur_avail_mem_kb = 0;
-    glGetIntegerv(GL_GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX, &cur_avail_mem_kb);
+    glGetIntegerv(GL_GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX, &cur_avail_mem_kb);
```
v1993, line 285: This change looks odd at the first glance - maybe it would be good to add a comment why is total memory reported instead of dedicated memory?

FernandoS27: it was actually a regression, I was testing "GL_GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX" while addressing feedback and left it there by accident.

PR #8136, opened 2022-04-02
core: hid: Remove service callbacks on game close

german77:
Currently not all services are destroyed and callbacks are never released for the same reason. To prevent triggering a callback on a partially dead npad service we manually remove all the input callbacks related to services.

#8108 tries to fix the issue but it's not doing it cleanly. For now this PR ensures issues like this never happen.


german77:
Resolved by #8165

PR #8143, opened 2022-04-02
native_clock: Use lfence with rdtsc

merryhime:
**Why lfence before rdtsc?**

mfence is not guaranteed by Intel to serialize the instruction stream, whereas lfence is, therefore lfence is the correct choice here. It's also more lightweight.

**Why lfence after rdtsc?**

rdtsc is not a serialising instruction, therefore instructions after it may be reordered to execute before it. [The manual actually explicitly mentions all of this](https://www.felixcloutier.com/x86/rdtsc), asking you to execute lfence after rdtsc(p) if correct ordering is required.

**Why is _ReadWriteBarrier required?**

lfence is a CPU-level fence, and does not necessarily disable compiler optimizations that may reorder instructions at compile time. A compiler barrier is required. (_ReadWriteBarrier does not emit any instructions.)

**Why is inline assembly required for non-MSVC compilers?**

It is the only way to express a compile-time memory barrier, so you might as well implement the whole thing in an inline assembly block.

**What about rdtscp?**

I would generally prefer rdtscp where possible due to clearer semantics. It is however not supported on processors earlier than Nehalem and the K8 Hammer; which leads to the natural question: do we support those CPUs?

---

MSVC does correctly inline this at inline level /Ob2.

Verify codegen: https://godbolt.org/z/9Tna8Kz39

### DIFF
--- a/src/common/x64/native_clock.cpp
+++ b/src/common/x64/native_clock.cpp
@@ -10,25 +10,47 @@
 #include "common/uint128.h"
 #include "common/x64/native_clock.h"
 
+#ifdef _MSC_VER
+#include <intrin.h>
+#endif
+
 namespace Common {
+
+inline u64 FencedRDTSC() {
+#ifdef _MSC_VER
+    _mm_lfence();
+    _ReadWriteBarrier();
+    const u64 result = __rdtsc();
+    _mm_lfence();
+    _ReadWriteBarrier();
+    return result;
+#else
+    u64 result;
+    asm volatile("lfence\n\t"
+                 "rdtsc\n\t"
+                 "shl $32, %%rdx\n\t"
+                 "or %%rdx, %0\n\t"
+                 "lfence"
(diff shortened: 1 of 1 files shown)


src/common/x64/native_clock.cpp
```diff
@@ -10,25 +10,47 @@
 #include "common/uint128.h"
 #include "common/x64/native_clock.h"
 
+#ifdef _MSC_VER
+#include <intrin.h>
+#endif
+
 namespace Common {
 
+inline u64 FencedRDTSC() {
```
lioncash, line 19:
```suggestion
static u64 FencedRDTSC() {
```

So this has internal linkage

bunnei:
(tag removed just because this merge conflicts)

v1993:
Specifically, git fails to merge this in EA due to #8105.

FernandoS27 (approved):

PR #8154, opened 2022-04-04
texture_cache/util: Remove unneeded ReadBlockUnsafe

ameerj:
This call is reading GPU memory into the dst buffer, which is then overwritten by the SwizzleTexture call on the next line.


bunnei (approved):

FernandoS27:
This change is dangerous. as the destination is not fully written by the swizzle, thus some of the original memory used as left out of boundaries will be rewritten with different data.

PR #8157, opened 2022-04-05
kernel: Fix some data races

lat9nq:
Used [ThreadSanitizer](https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual) on yuzu and found a few races. This just fix the ones that yuzu-cmd encounters in `core/hle/kernel` while booting [es2gears](https://github.com/switchbrew/switch-examples/tree/master/graphics/opengl/es2gears), minus a `Kernel::SessionRequestManager` one that I'm not sure how to fix.

### DIFF
--- a/src/core/hle/kernel/board/nintendo/nx/k_system_control.cpp
+++ b/src/core/hle/kernel/board/nintendo/nx/k_system_control.cpp
@@ -148,9 +148,9 @@ u64 GenerateUniformRange(u64 min, u64 max, F f) {
 } // Anonymous namespace
 
 u64 KSystemControl::GenerateRandomU64() {
-    static std::random_device device;
-    static std::mt19937 gen(device());
-    static std::uniform_int_distribution<u64> distribution(1, std::numeric_limits<u64>::max());
+    std::random_device device;
+    std::mt19937 gen(device());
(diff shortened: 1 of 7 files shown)


src/core/hle/kernel/board/nintendo/nx/k_system_control.cpp
```diff
@@ -148,6 +149,9 @@ u64 GenerateUniformRange(u64 min, u64 max, F f) {
 } // Anonymous namespace
 
 u64 KSystemControl::GenerateRandomU64() {
+    static std::mutex mutex;
+    std::lock_guard lock{mutex};
+
     static std::random_device device;
```
lioncash, line 155: Pretty sure all of these can be made non-static to get rid of the need to use a mutex here

lat9nq: I don't really understand how the random number generator stuff works. If we initialize a new generator every time we call this function, is it still random enough to use? Do we care? (This is why I left it static.)

lioncash: Yeah, it should be fine, since random_device is being used.

Even then, implementations are allowed to implement random_device with a PRNG instead of a true RNG source, so for our purposes, it's fine to have it just be sufficiently random

PR #8158, opened 2022-04-05
build: remove -fconcepts

Tachi107:
It was needed on GCC versions not supporting `-std=c++20`, but GCC 10 and newer (required to compile yuzu) don't need it anymore

Edit: added in https://github.com/yuzu-emu/yuzu/commit/dd2ff236217cb02275cbfa66f50c265a578d1f1e, CC: @lioncash

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -103,12 +103,6 @@ else()
         -Wno-unused-parameter
     )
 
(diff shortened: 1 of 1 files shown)


lioncash (approved):

lioncash:
Thanks!

PR #8167, opened 2022-04-07
fix: remove #pragma once in .cpp file

Tachi107:


Moonlacer:
Was there a bug or something this PR fixes? A description is most likely necessary for this to be merged, if so

Tachi107:
I didn't add one as this is a trivial fix, but having `#pragma once` in a .cpp file throws a warning on some compilers, and should only be present in headers anyway

PR #8170, opened 2022-04-07
a workaround if the mouse cursor leaves the rendering window in panning mode

IamSanjid:
If the mouse DPI is high enough or the rendering window size is small enough the mouse cursor can leave the window, which can cause some annoying behavior, just in case a workaround for that issue.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -3595,7 +3595,15 @@ void GMainWindow::dragMoveEvent(QDragMoveEvent* event) {
 
 void GMainWindow::leaveEvent(QEvent* event) {
     if (Settings::values.mouse_panning) {
-        mouse_center_timer.start();
+        const QRect& rect = geometry();
+        QPoint position = QCursor::pos();
+
+        qint32 x = qBound(rect.left(), position.x(), rect.right());
+        qint32 y = qBound(rect.top(), position.y(), rect.bottom());
+        /* only start the timer if the mouse has left the window bound, the leave event is also triggered when the window looses focus */
+        if (x != position.x() || y != position.y()) {
+            mouse_center_timer.start();
+        }
         event->accept();
     }
 }
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -3306,16 +3306,9 @@ void GMainWindow::ShowMouseCursor() {
 }
 
 void GMainWindow::CenterMouseCursor() {
-    static bool logged = false;
     if (emu_thread == nullptr || !Settings::values.mouse_panning) {
         mouse_center_timer.stop();
-        logged = false;
-        LOG_INFO(Debug, "Stopped centering cursor!");
         return;
-    }
-    if (!logged) {
-        LOG_INFO(Debug, "Started centering cursor!");
-        logged = true;
     }
     const int center_x = render_window->width() / 2;
     const int center_y = render_window->height() / 2;
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -3310,6 +3310,10 @@ void GMainWindow::CenterMouseCursor() {
         mouse_center_timer.stop();
         return;
     }
+   if (!this->isActiveWindow()) {
+        mouse_center_timer.stop();
+        return;
+    }
     const int center_x = render_window->width() / 2;
     const int center_y = render_window->height() / 2;
 
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -152,7 +152,8 @@ __declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
 }
 #endif
 
-constexpr int default_mouse_timeout = 2500;
+constexpr int default_mouse_hide_timeout = 2500;
+constexpr int default_mouse_center_timeout = 10;
 
 /**
  * "Callouts" are one-time instructional messages shown to the user. In the config settings, there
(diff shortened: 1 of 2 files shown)


german77:
Seems like a complex solution for a simple problem. wouldn't it be way easier to just to reset the mouse position on `leaveEvent`

IamSanjid:
> Seems like a complex solution for a simple problem. wouldn't it be way easier to just to reset the mouse position on `leaveEvent`

Well, I tried that but the DPI was too high I guess and it simply didn't work. Also tried the QEvent::HoverLeave same result.

IamSanjid:
> Seems like a complex solution for a simple problem. wouldn't it be way easier to just to reset the mouse position on `leaveEvent`

Or maybe it's a QT thing, it seems the `QCursor::setPos` function actually fails to trigger the `enterEvent` or `QEvent::HoverEnter` sometimes if the DPI is high or if the mouse moved too far away from the window bound. So the behavior is kind of
mouse leaves the window -> leaveEvent -> setPos -> enterEvent -> but the dpi is too high so the mouse leaves again -> leaveEvent -> but setPos fails this time (maybe QT's internal mouse pos don't get updated in time so it thinks it already reached the desired position)

image
image
image

german77:
On bootmanager.cpp I have the code to reset the position on mouseMoveEvent If the mouse is able to scape it's because Qt updates too slow. This is why the mouse can scape to begin with.  The same event probably triggers enterEvent/leaveEvent.

I will give it a test to your code to see how well it performs. I have a feeling it will start to ignore mouse inputs if you move too fast.

IamSanjid:
> On bootmanager.cpp I have the code to reset the position on mouseMoveEvent If the mouse is able to scape it's because Qt updates too slow. The same event probably triggers enterEvent/leaveEvent.
> 
> I will give it a test to your code to see how well it performs. I have a feeling it will start to ignore mouse inputs if you move too fast.

The idea was if the panning is enabled we start a timer after the `leaveEvent` is called and reset the mouse position to center until QT detects the `HoverMove` event.

Or maybe we could completely ditch the QT's mouse move event. Then use a `QTimer` to monitor the mouse position every 10ms(let the user move the cursor a little bit to pan around) save the last position and check if the position was changed then we call our mouseMoveEvent of bootmanager.cpp in this way we won't have to care if the mouse position gets out of the window coz our monitor function gonna call the mouseMoveEvent everytime if it detects the position was changed since the last frame.

german77:
Just tested and confirmed the issues. Your solution works nice. I only found one major problem. If something popups like the controller applet or a warning your mouse is still trapped. Please disable auto center if the window looses focus 
image

src/yuzu/main.cpp
```diff
@@ -3301,10 +3305,22 @@ void GMainWindow::ShowMouseCursor() {
     }
 }
 
+void GMainWindow::CenterMouseCursor() {
+    if (emu_thread == nullptr || !Settings::values.mouse_panning) {
+        mouse_center_timer.stop();
+        return;
+    }
+    const int center_x = render_window->width() / 2;
```
german77, line 3313:
```suggestion    
   if (!this->isActiveWindow()) {
        mouse_center_timer.stop();
        return;
    }
    const int center_x = render_window->width() / 2;
```

src/yuzu/main.cpp
```diff
@@ -3577,6 +3593,22 @@ void GMainWindow::dragMoveEvent(QDragMoveEvent* event) {
...
+
+        qint32 x = qBound(rect.left(), position.x(), rect.right());
+        qint32 y = qBound(rect.top(), position.y(), rect.bottom());
+        // Only start the timer if the mouse has left the window bound.
+        // The leave event is also triggered when the window looses focus.
+        if (x != position.x() || y != position.y()) {
+            mouse_center_timer.start();
+        }
+        event->accept();
+    }
```
german77, line 3609:
```suggestion
    if (!this->isActiveWindow()) {
        return;
    }

    if (Settings::values.mouse_panning) {
        mouse_center_timer.start();
        event->accept();
    }
```

german77 (commented):
We can use `isActiveWindow` to easily know if the current window is on focus

IamSanjid:
> We can use `isActiveWindow` to easily know if the current window is on focus

uunfortunately Qt is kind of annoying the `isisActiveWindow` window is set to true even if a pop window is shown somehow.

german77:
I tested both changes. Both worked correctly. What did you use to make it fail?

IamSanjid:
> I tested both changes. Both worked correctly. What did you use to make it fail?

ow i think for me on leaveEvent that function returned true probably slow qt update but since u also call on the timer elapsed function that time it returned false may be, i couldn't test the elapsed function change, if you're okay with everything i'll be committing ur changes

german77:
I think just adding the change that I left unresolved should be enough to add it on EA or mainline. I still think it's a bit complicated solution but given the slow update Qt has it should be ok.

IamSanjid:
> I think just adding the change that I left unresolved should be enough to add it on EA or mainline. I still think it's a bit complicated solution but given the slow update Qt has it should be ok.

okay, this should be it then. yeah i can understand you doing all those stuff just to clip the mouse to the center feels a little bit overkill, but it is what it is.

Morph1984 (approved):

goldenx86:

IamSanjid:
> Hi, can I mention you and this PR in the next progress report?

sure

PR #8171, opened 2022-04-07
Improvements for game modding with Skyline, DNS resolution

tech-ticks:
- Allows loading LayeredExeFs patches from `sdmc:/atmosphere/contents/titleid/exefs`
  - Makes mod distribution easier since users can just copy the `atmosphere` folder to the SD card root like on hardware
  - Required for self-updating Skyline plugins that assume that `subsdk9` is writable in this directory
- Adds support for the `SO_KEEPALIVE` socket option
  - Used by the Skyline TCP logger (attaching the logger still crashes Yuzu though)
- Implements DNS address resolution in the `sfdnsres` service (`GetAddrInfoRequest` is now functional and `GetAddrInfoRequestWithOptions` was added)
  - Required for Skyline plugins that use HTTP requests (but requests still seem to get stuck after the DNS resolution)

### DIFF
--- a/src/core/hle/service/sockets/sfdnsres.cpp
+++ b/src/core/hle/service/sockets/sfdnsres.cpp
@@ -2,8 +2,24 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <string_view>
+#include <utility>
+#include <vector>
+
+#include "common/string_util.h"
+#include "common/swap.h"
+#include "core/core.h"
 #include "core/hle/ipc_helpers.h"
 #include "core/hle/service/sockets/sfdnsres.h"
+#include "core/memory.h"
+
+#ifdef _WIN32
+#include <ws2tcpip.h>
+#elif YUZU_UNIX
+#include <arpa/inet.h>
+#include <netdb.h>
+#include <sys/socket.h>
+#endif
 
 namespace Service::Sockets {
 
@@ -21,7 +37,7 @@ SFDNSRES::SFDNSRES(Core::System& system_) : ServiceFramework{system_, "sfdnsres"
         {9, nullptr, "CancelRequest"},
         {10, nullptr, "GetHostByNameRequestWithOptions"},
         {11, nullptr, "GetHostByAddrRequestWithOptions"},
-        {12, nullptr, "GetAddrInfoRequestWithOptions"},
+        {12, &SFDNSRES::GetAddrInfoRequestWithOptions, "GetAddrInfoRequestWithOptions"},
         {13, nullptr, "GetNameInfoRequestWithOptions"},
         {14, nullptr, "ResolverSetOptionRequest"},
         {15, nullptr, "ResolverGetOptionRequest"},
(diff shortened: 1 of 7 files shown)


src/core/hle/service/sockets/sfdnsres.cpp
```diff
@@ -41,12 +98,136 @@ void SFDNSRES::GetAddrInfoRequest(Kernel::HLERequestContext& ctx) {
...
+                const auto addr = *reinterpret_cast<sockaddr_in6*>(current->ai_addr);
+                serialized_addr.sin6_family = htons(addr.sin6_family);
+                serialized_addr.sin6_port = htons(addr.sin6_port);
+                serialized_addr.sin6_flowinfo = htonl(addr.sin6_flowinfo);
+                serialized_addr.sin6_scope_id = htonl(addr.sin6_scope_id);
+                std::memcpy(serialized_addr.sin6_addr, &addr.sin6_addr, sizeof(u8) * 16);
+                std::memcpy(header_ptr + header_size, &serialized_addr, sizeof(SockAddrIn6));
+
+                char addr_string_buf[64];
+                inet_ntop(AF_INET6, &addr.sin6_addr, addr_string_buf, 64);
```
lioncash, line 207:
```suggestion
                inet_ntop(AF_INET6, &addr.sin6_addr, addr_string_buf, std::size(addr_string_buf));
```

src/core/hle/service/sockets/sfdnsres.cpp
```diff
@@ -41,12 +98,136 @@ void SFDNSRES::GetAddrInfoRequest(Kernel::HLERequestContext& ctx) {
...
+                    u32 sin6_scope_id;
+                };
+
+                SockAddrIn6 serialized_addr = {};
+                const auto addr = *reinterpret_cast<sockaddr_in6*>(current->ai_addr);
+                serialized_addr.sin6_family = htons(addr.sin6_family);
+                serialized_addr.sin6_port = htons(addr.sin6_port);
+                serialized_addr.sin6_flowinfo = htonl(addr.sin6_flowinfo);
+                serialized_addr.sin6_scope_id = htonl(addr.sin6_scope_id);
+                std::memcpy(serialized_addr.sin6_addr, &addr.sin6_addr, sizeof(u8) * 16);
```
lioncash, line 203:
```suggestion
                std::memcpy(serialized_addr.sin6_addr, &addr.sin6_addr, sizeof(SockAddrIn6::sin6_addr)));
```

src/core/hle/service/sockets/sfdnsres.cpp
```diff
@@ -41,12 +98,136 @@ void SFDNSRES::GetAddrInfoRequest(Kernel::HLERequestContext& ctx) {
...
+                std::memcpy(header_ptr + header_size, &serialized_addr, sizeof(SockAddrIn6));
+
+                char addr_string_buf[64];
+                inet_ntop(AF_INET6, &addr.sin6_addr, addr_string_buf, 64);
+                LOG_INFO(Service, "Resolved host '{}' to IPv6 address {}", host, addr_string_buf);
+                break;
+            }
+            default:
+                std::memcpy(header_ptr + header_size, current->ai_addr, addr_size);
```
lioncash, line 212: Generally we add breaks, even in the default case, unless it's a default signified as unreachable

src/core/hle/service/sockets/sfdnsres.cpp
```diff
@@ -41,12 +98,136 @@ void SFDNSRES::GetAddrInfoRequest(Kernel::HLERequestContext& ctx) {
...
+                };
+
+                SockAddrIn serialized_addr = {};
+                const auto addr = *reinterpret_cast<sockaddr_in*>(current->ai_addr);
+                serialized_addr.sin_port = htons(addr.sin_port);
+                serialized_addr.sin_family = htons(addr.sin_family);
+                serialized_addr.sin_addr = htonl(addr.sin_addr.s_addr);
+                std::memcpy(header_ptr + header_size, &serialized_addr, sizeof(SockAddrIn));
+
+                char addr_string_buf[64];
```
lioncash, line 183:
```suggestion
                char addr_string_buf[64]{};
```

src/core/hle/service/sockets/sfdnsres.cpp
```diff
@@ -41,12 +98,136 @@ void SFDNSRES::GetAddrInfoRequest(Kernel::HLERequestContext& ctx) {
...
+                inet_ntop(AF_INET, &addr.sin_addr, addr_string_buf, 64);
```
lioncash, line 184:
```suggestion
                inet_ntop(AF_INET, &addr.sin_addr, addr_string_buf, std::size(addr_string_buf));
```

src/core/hle/service/sockets/sfdnsres.cpp
```diff
@@ -41,12 +98,136 @@ void SFDNSRES::GetAddrInfoRequest(Kernel::HLERequestContext& ctx) {
...
+                SockAddrIn6 serialized_addr = {};
+                const auto addr = *reinterpret_cast<sockaddr_in6*>(current->ai_addr);
+                serialized_addr.sin6_family = htons(addr.sin6_family);
+                serialized_addr.sin6_port = htons(addr.sin6_port);
+                serialized_addr.sin6_flowinfo = htonl(addr.sin6_flowinfo);
+                serialized_addr.sin6_scope_id = htonl(addr.sin6_scope_id);
+                std::memcpy(serialized_addr.sin6_addr, &addr.sin6_addr, sizeof(u8) * 16);
+                std::memcpy(header_ptr + header_size, &serialized_addr, sizeof(SockAddrIn6));
+
+                char addr_string_buf[64];
```
lioncash, line 206:
```suggestion
                char addr_string_buf[64]{};
```

src/core/hle/service/sockets/sfdnsres.h
```diff
@@ -19,6 +20,11 @@ class SFDNSRES final : public ServiceFramework<SFDNSRES> {
 
 private:
     void GetAddrInfoRequest(Kernel::HLERequestContext& ctx);
+    void GetAddrInfoRequestWithOptions(Kernel::HLERequestContext& ctx);
+
+    std::pair<u32, s32> GetAddrInfoRequestImpl(Kernel::HLERequestContext& ctx);
+    const std::vector<u8> SerializeAddrInfo(const addrinfo* addrinfo, s32 result_code,
```
lioncash, line 26:
```suggestion
    std::vector<u8> SerializeAddrInfo(const addrinfo* addrinfo, s32 result_code,
```

Also `<string_view>`, `<utility>` and `<vector>` need to be included

src/core/hle/service/sockets/sfdnsres.h
```diff
@@ -19,6 +20,11 @@ class SFDNSRES final : public ServiceFramework<SFDNSRES> {
 
 private:
     void GetAddrInfoRequest(Kernel::HLERequestContext& ctx);
+    void GetAddrInfoRequestWithOptions(Kernel::HLERequestContext& ctx);
+
+    std::pair<u32, s32> GetAddrInfoRequestImpl(Kernel::HLERequestContext& ctx);
```
lioncash, line 25: Is this and the function below dependent on any class state? It seems like they're only called by member functions and can be kept internal to the cpp file

src/core/hle/service/sockets/sfdnsres.h
```diff
@@ -19,6 +20,11 @@ class SFDNSRES final : public ServiceFramework<SFDNSRES> {
...
+    const std::vector<u8> SerializeAddrInfo(const addrinfo* addrinfo, s32 result_code,
+                                            const std::string& host) const;
```
lioncash, line 27:
```suggestion
                                            std::string_view host) const;
```

src/core/hle/service/sockets/sfdnsres.cpp
```diff
@@ -41,12 +98,136 @@ void SFDNSRES::GetAddrInfoRequest(Kernel::HLERequestContext& ctx) {
...
+
+        constexpr auto HEADER_MAGIC = 0xBEEFCAFE;
+        header.magic = htonl(HEADER_MAGIC);
+        header.family = htonl(current->ai_family);
+        header.flags = htonl(current->ai_flags);
+        header.socket_type = htonl(current->ai_socktype);
+        header.protocol = htonl(current->ai_protocol);
+        header.address_length = current->ai_addr ? htonl((u32)current->ai_addrlen) : 0;
+
+        const auto header_ptr = data.data() + last_size;
```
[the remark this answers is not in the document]
tech-ticks: I had to remove the `const` since it's a `memcpy` destination, seems like this is ignored when `auto` is used instead of `auto*`

src/core/hle/service/sockets/sfdnsres.h
tech-ticks: No, but the logging macro needs a class so I got a compile error when I tried to move it out of the class. I could remove the logging though, idk if it's useful

FernandoS27:
@tech-ticks hi thank you very much for your contribution! May I tag it to EA or would you prefer dedicated testing and then go to mainline?

src/core/hle/service/sockets/sfdnsres.h
lioncash: I don't get what you mean. We have several places in the codebase where we log in a function that isn't part of a class

tech-ticks:
> @tech-ticks hi thank you very much for your contribution! May I tag it to EA or would you prefer dedicated testing and then go to mainline?

EA is fine for me, yeah

src/core/hle/service/sockets/sfdnsres.h
tech-ticks: I guess I missed something then, I'll take another look.

tech-ticks: Yeah it works just fine, seems like I got confused by something else. Sorry for that, resolved.

src/core/hle/service/sockets/sfdnsres.h
```diff
@@ -4,11 +4,16 @@
 
 #pragma once
 
+#include <string_view>
```
lioncash, line 7: These can be moved into the cpp file now that the function definitions are internally linked

src/core/hle/service/sockets/sfdnsres.h
```diff
@@ -9,6 +9,7 @@
 namespace Core {
 class System;
 }
+struct addrinfo;
```
lioncash, line 12: Sorry for missing this in the previous review, but this can be removed, since the functions were made internally linked as well.

Everything looks good after

FernandoS27 (approved):

PR #8172, opened 2022-04-07
hle: kernel: Use std::mutex instead of spin locks for most kernel locking.

bunnei:
This is an experimental change to see if using host mutexes over our own spin-lock implementation improves how the emulator runs, particularly on systems with low core counts. Preliminary testing has shown that this makes emulator much more usable on 4T systems. Marking as do-not-merge for now until we get broader testing.

Credits to @FernandoS27 for the idea.

We still use spinlocks for thread context guards, as for some reason this appears to be very sensitive to timing changes. I'm still investigating the root problem here.

### DIFF
--- a/src/common/fiber.cpp
+++ b/src/common/fiber.cpp
@@ -2,9 +2,10 @@
 // Licensed under GPLv2 or any later version
 // Refer to the license.txt file included.
 
+#include <mutex>
+
 #include "common/assert.h"
 #include "common/fiber.h"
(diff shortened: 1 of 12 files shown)


DiegoMadGlez:
This PR also fixes issues with stuttering or 0fps while having yuzu as priority window.
It improves sound and emulator is smoother in every way.
In some games I see improvements in fps and stability in general.

FernandoS27 (approved):

PR #8178, opened 2022-04-08
hle: kernel: Invalidate entire icache in UnmapProcessMemory and UnmapCodeMemory (fixes #8174)

### LINKED ISSUE #8174: Corrupted "nested" hooks in Skyline plugins
Related to #7392

The Skyline modding framework allows loading plugin NROs that can register dynamic hooks. The NROs load fine, but hooks get corrupted if the following sequence occurs (the error can probably also reproduced in other ways):
1. Skyline loads the plugin NRO after the game mounts the romfs and executes the plugin's main entrypoint function
2. The plugin hooks a function from the main executable (in the attached example, it's a function that is executed every frame)
3. *Inside* that hook, another hook is registered, this time for a function in the SDK

Code of the Skyline plugin (`libyuzu_repro.nro`): https://github.com/tech-ticks/skyline_yuzu_repro_plugin/blob/main/src/lib.rs

## Expected behaviour

The plugin registers a hook to `nn::fs::OpenFile` that logs every accessed file and then calls the original function. The log output on console looks like this:

```
...
[PluginManager] Loading plugins...
[PluginManager] Loaded 'rom:/skyline/plugins/libyuzu_repro.nro'
[PluginManager] Running `main` for rom:/skyline/plugins/libyuzu_repro.nro
[repro_plugin] main() in yuzu repro plugin. Installing hook for game function 0x182050 (GroundManager_Update)..
[PluginManager] Finished running `main` for 'rom:/skyline/plugins/libyuzu_repro.nro' (0x0)
[skyline_main] loaded plugins
[repro_plugin] Inside GroundManager_Update, installing hook for SDK function 'nn::fs::OpenFile'

-- Yuzu prints the same lines until here --

[repro_plugin] !!! nn::fs::OpenFile called with path rom:/Data/Resources/unity_builtin_extra
[repro_plugin] !!! nn::fs::OpenFile called with path rom:/Data/Resources/unity default resources
[repro_plugin] !!! nn::fs::OpenFile called with path rom:/Data/resources.assets.resS
...
```

## Actual behaviour

The hook that is supposed to log file access is broken on Yuzu. On my machine, one of two things happen at random:
- The game hangs during startup
- The game loads, but the hook is never called. The original `nn::fs::OpenFile` seems to be called instead, no logs show up.

## Reproduction steps

Yuzu build: `master-2db5076ec`

Copy this folder to `$appdata/yuzu/load/01003D200BAA2000`:
[skyline_plugin_hook_repro.zip](https://github.com/yuzu-emu/yuzu/files/8453791/skyline_plugin_hook_repro.zip)

and run **Pokèmon Mystery Dungeon Rescue Team DX** (version **1.0.2**). An offset that depends on this version is used, so it won't work with other game versions.

## Other Resources

Yuzu log: [yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/8453674/yuzu_log.txt)
Skyline branch of the included build (uses `svcOutputDebugString` instead of TCP logging): https://github.com/tech-ticks/skyline/tree/yuzu-repro

tech-ticks:
Thanks to @bunnei for figuring this out!

- Adds icache invalidation to `UnmapProcessMemory`
- Forces invalidation of the entire icache in `UnmapCodeMemory`
  - icache invalidation is still constrained to a range when unloading NROs to avoid potential performance regressions in Super Smash Bros. Ultimate

Fixes #8174

### DIFF
--- a/src/core/hle/kernel/k_page_table.cpp
+++ b/src/core/hle/kernel/k_page_table.cpp
@@ -346,7 +346,8 @@ ResultCode KPageTable::MapCodeMemory(VAddr dst_address, VAddr src_address, std::
     return ResultSuccess;
 }
 
-ResultCode KPageTable::UnmapCodeMemory(VAddr dst_address, VAddr src_address, std::size_t size) {
+ResultCode KPageTable::UnmapCodeMemory(VAddr dst_address, VAddr src_address, std::size_t size,
+                                       ICacheInvalidationStrategy icache_invalidation_strategy) {
     // Validate the mapping request.
     R_UNLESS(this->CanContain(dst_address, size, KMemoryState::AliasCode),
              ResultInvalidMemoryRegion);
@@ -396,7 +397,11 @@ ResultCode KPageTable::UnmapCodeMemory(VAddr dst_address, VAddr src_address, std
     bool reprotected_pages = false;
     SCOPE_EXIT({
         if (reprotected_pages && any_code_pages) {
-            system.InvalidateCpuInstructionCacheRange(dst_address, size);
+            if (icache_invalidation_strategy == ICacheInvalidationStrategy::InvalidateRange) {
+                system.InvalidateCpuInstructionCacheRange(dst_address, size);
+            } else {
+                system.InvalidateCpuInstructionCaches();
+            }
         }
     });
 
@@ -562,6 +567,8 @@ ResultCode KPageTable::UnmapProcessMemory(VAddr dst_addr, std::size_t size,
     // Apply the memory block update.
     block_manager->Update(dst_addr, num_pages, KMemoryState::Free, KMemoryPermission::None,
                           KMemoryAttribute::None);
+
+    system.InvalidateCpuInstructionCaches();
 
     return ResultSuccess;
 }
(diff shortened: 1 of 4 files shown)


src/core/hle/kernel/k_page_table.cpp
```diff
@@ -346,7 +346,8 @@ ResultCode KPageTable::MapCodeMemory(VAddr dst_address, VAddr src_address, std::
     return ResultSuccess;
 }
 
-ResultCode KPageTable::UnmapCodeMemory(VAddr dst_address, VAddr src_address, std::size_t size) {
+ResultCode KPageTable::UnmapCodeMemory(VAddr dst_address, VAddr src_address, std::size_t size,
+                                       bool invalidate_entire_icache) {
```
lioncash, line 350: I'd use an enum class here instead of a bool for readability.

e.g.

```cpp
page_table.UnmapCodeMemory(addr + nro_size, bss_addr, bss_size, false);
```

is much more opaque at the usage-site compared to something like:

```
page_table.UnmapCodeMemory(addr + nro_size, bss_addr, bss_size, ICache::Preserve);
```

bunnei (approved):

Issue #8185, opened 2022-04-09
physical_core: Data races


lat9nq:
Caused by #8148, @merryhime you may want to take a look

[ThreadSanitizer](https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual) now reports a few data races while booting [es2gears](https://github.com/switchbrew/switch-examples/tree/master/graphics/opengl/es2gears) via yuzu-cmd.

[dynarm_kern_tsan.txt](https://github.com/yuzu-emu/yuzu/files/8457910/dynarm_kern_tsan.txt)

PR #8188, opened 2022-04-10
dynarmic: Fix race when switching page tables

merryhime:
Addresses #8185.

Zero performance impact. (`std::atomic::load` is free.)

A proper fix would be to make it a core-option but that's awaiting refactor.

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic_32.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic_32.cpp
@@ -88,7 +88,7 @@ public:
 
     void CallSVC(u32 swi) override {
         parent.svc_swi = swi;
-        parent.jit->HaltExecution(svc_call);
+        parent.jit.load()->HaltExecution(svc_call);
     }
 
     void AddTicks(u64 ticks) override {
@@ -150,6 +150,13 @@ std::shared_ptr<Dynarmic::A32::Jit> ARM_Dynarmic_32::MakeJit(Common::PageTable* 
     // Code cache size
     config.code_cache_size = 512_MiB;
     config.far_code_offset = 400_MiB;
+
+    // null_jit
+    if (!page_table) {
+        // Don't waste too much memory on null_jit
+        config.code_cache_size = 8_MiB;
+        config.far_code_offset = 4_MiB;
+    }
 
     // Safe optimizations
     if (Settings::values.cpu_debug_mode) {
@@ -228,7 +235,7 @@ std::shared_ptr<Dynarmic::A32::Jit> ARM_Dynarmic_32::MakeJit(Common::PageTable* 
 
 void ARM_Dynarmic_32::Run() {
     while (true) {
-        const auto hr = jit->Run();
+        const auto hr = jit.load()->Run();
         if (Has(hr, svc_call)) {
             Kernel::Svc::Call(system, svc_swi);
         }
@@ -239,7 +246,7 @@ void ARM_Dynarmic_32::Run() {
 }
 
 void ARM_Dynarmic_32::Step() {
-    jit->Step();
+    jit.load()->Step();
 }
 
 ARM_Dynarmic_32::ARM_Dynarmic_32(System& system_, CPUInterrupts& interrupt_handlers_,
@@ -249,24 +256,24 @@ ARM_Dynarmic_32::ARM_Dynarmic_32(System& system_, CPUInterrupts& interrupt_handl
       cb(std::make_unique<DynarmicCallbacks32>(*this)),
       cp15(std::make_shared<DynarmicCP15>(*this)), core_index{core_index_},
       exclusive_monitor{dynamic_cast<DynarmicExclusiveMonitor&>(exclusive_monitor_)},
-      jit(MakeJit(nullptr)) {}
+      null_jit{MakeJit(nullptr)}, jit{null_jit.get()} {}
 
 ARM_Dynarmic_32::~ARM_Dynarmic_32() = default;
 
 void ARM_Dynarmic_32::SetPC(u64 pc) {
-    jit->Regs()[15] = static_cast<u32>(pc);
+    jit.load()->Regs()[15] = static_cast<u32>(pc);
 }
 
 u64 ARM_Dynarmic_32::GetPC() const {
-    return jit->Regs()[15];
+    return jit.load()->Regs()[15];
 }
 
 u64 ARM_Dynarmic_32::GetReg(int index) const {
-    return jit->Regs()[index];
+    return jit.load()->Regs()[index];
 }
 
 void ARM_Dynarmic_32::SetReg(int index, u64 value) {
-    jit->Regs()[index] = static_cast<u32>(value);
+    jit.load()->Regs()[index] = static_cast<u32>(value);
 }
 
 u128 ARM_Dynarmic_32::GetVectorReg(int index) const {
@@ -276,11 +283,11 @@ u128 ARM_Dynarmic_32::GetVectorReg(int index) const {
 void ARM_Dynarmic_32::SetVectorReg(int index, u128 value) {}
 
 u32 ARM_Dynarmic_32::GetPSTATE() const {
-    return jit->Cpsr();
+    return jit.load()->Cpsr();
 }
 
 void ARM_Dynarmic_32::SetPSTATE(u32 cpsr) {
-    jit->SetCpsr(cpsr);
+    jit.load()->SetCpsr(cpsr);
 }
 
 u64 ARM_Dynarmic_32::GetTlsAddress() const {
@@ -301,7 +308,7 @@ void ARM_Dynarmic_32::SetTPIDR_EL0(u64 value) {
 
 void ARM_Dynarmic_32::SaveContext(ThreadContext32& ctx) {
     Dynarmic::A32::Context context;
-    jit->SaveContext(context);
+    jit.load()->SaveContext(context);
     ctx.cpu_registers = context.Regs();
     ctx.extension_registers = context.ExtRegs();
     ctx.cpsr = context.Cpsr();
@@ -314,27 +321,27 @@ void ARM_Dynarmic_32::LoadContext(const ThreadContext32& ctx) {
     context.ExtRegs() = ctx.extension_registers;
     context.SetCpsr(ctx.cpsr);
     context.SetFpscr(ctx.fpscr);
-    jit->LoadContext(context);
+    jit.load()->LoadContext(context);
 }
 
 void ARM_Dynarmic_32::PrepareReschedule() {
-    jit->HaltExecution(break_loop);
+    jit.load()->HaltExecution(break_loop);
 }
 
 void ARM_Dynarmic_32::SignalInterrupt() {
-    jit->HaltExecution(break_loop);
+    jit.load()->HaltExecution(break_loop);
 }
 
 void ARM_Dynarmic_32::ClearInstructionCache() {
-    jit->ClearCache();
+    jit.load()->ClearCache();
 }
 
 void ARM_Dynarmic_32::InvalidateCacheRange(VAddr addr, std::size_t size) {
-    jit->InvalidateCacheRange(static_cast<u32>(addr), size);
+    jit.load()->InvalidateCacheRange(static_cast<u32>(addr), size);
 }
 
 void ARM_Dynarmic_32::ClearExclusiveState() {
-    jit->ClearExclusiveState();
+    jit.load()->ClearExclusiveState();
 }
 
 void ARM_Dynarmic_32::PageTableChanged(Common::PageTable& page_table,
@@ -345,13 +352,14 @@ void ARM_Dynarmic_32::PageTableChanged(Common::PageTable& page_table,
     auto key = std::make_pair(&page_table, new_address_space_size_in_bits);
     auto iter = jit_cache.find(key);
     if (iter != jit_cache.end()) {
-        jit = iter->second;
+        jit.store(iter->second.get());
         LoadContext(ctx);
         return;
     }
-    jit = MakeJit(&page_table);
+    std::shared_ptr new_jit = MakeJit(&page_table);
+    jit.store(new_jit.get());
     LoadContext(ctx);
-    jit_cache.emplace(key, jit);
+    jit_cache.emplace(key, std::move(new_jit));
 }
 
 } // namespace Core
--- a/src/core/arm/dynarmic/arm_dynarmic_64.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic_64.cpp
@@ -109,7 +109,7 @@ public:
             break;
         }
 
-        parent.jit->HaltExecution(Dynarmic::HaltReason::CacheInvalidation);
+        parent.jit.load()->HaltExecution(Dynarmic::HaltReason::CacheInvalidation);
     }
 
     void ExceptionRaised(u64 pc, Dynarmic::A64::Exception exception) override {
@@ -130,7 +130,7 @@ public:
 
     void CallSVC(u32 swi) override {
         parent.svc_swi = swi;
-        parent.jit->HaltExecution(svc_call);
+        parent.jit.load()->HaltExecution(svc_call);
     }
 
     void AddTicks(u64 ticks) override {
@@ -212,6 +212,13 @@ std::shared_ptr<Dynarmic::A64::Jit> ARM_Dynarmic_64::MakeJit(Common::PageTable* 
     config.code_cache_size = 512_MiB;
     config.far_code_offset = 400_MiB;
 
+    // null_jit
+    if (!page_table) {
+        // Don't waste too much memory on null_jit
+        config.code_cache_size = 8_MiB;
+        config.far_code_offset = 4_MiB;
+    }
+
     // Safe optimizations
     if (Settings::values.cpu_debug_mode) {
         if (!Settings::values.cpuopt_page_tables) {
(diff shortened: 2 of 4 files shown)


src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -88,7 +88,7 @@ class DynarmicCallbacks32 : public Dynarmic::A32::UserCallbacks {
 
     void CallSVC(u32 swi) override {
         parent.svc_swi = swi;
-        parent.jit->HaltExecution(svc_call);
+        parent.jit.load()->HaltExecution(svc_call);
```
FernandoS27, line 91: std::memory_order_acquire release/sequence are not required.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -228,7 +235,7 @@ std::shared_ptr<Dynarmic::A32::Jit> ARM_Dynarmic_32::MakeJit(Common::PageTable*
 
 void ARM_Dynarmic_32::Run() {
     while (true) {
-        const auto hr = jit->Run();
+        const auto hr = jit.load()->Run();
```
FernandoS27, line 238: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -249,24 +256,24 @@ ARM_Dynarmic_32::ARM_Dynarmic_32(System& system_, CPUInterrupts& interrupt_handl
...
       cp15(std::make_shared<DynarmicCP15>(*this)), core_index{core_index_},
       exclusive_monitor{dynamic_cast<DynarmicExclusiveMonitor&>(exclusive_monitor_)},
-      jit(MakeJit(nullptr)) {}
+      null_jit{MakeJit(nullptr)}, jit{null_jit.get()} {}
 
 ARM_Dynarmic_32::~ARM_Dynarmic_32() = default;
 
 void ARM_Dynarmic_32::SetPC(u64 pc) {
-    jit->Regs()[15] = static_cast<u32>(pc);
+    jit.load()->Regs()[15] = static_cast<u32>(pc);
```
FernandoS27, line 264: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -249,24 +256,24 @@ ARM_Dynarmic_32::ARM_Dynarmic_32(System& system_, CPUInterrupts& interrupt_handl
...
 }
 
 u64 ARM_Dynarmic_32::GetPC() const {
-    return jit->Regs()[15];
+    return jit.load()->Regs()[15];
```
FernandoS27, line 268: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -249,24 +256,24 @@ ARM_Dynarmic_32::ARM_Dynarmic_32(System& system_, CPUInterrupts& interrupt_handl
...
 }
 
 u64 ARM_Dynarmic_32::GetReg(int index) const {
-    return jit->Regs()[index];
+    return jit.load()->Regs()[index];
```
FernandoS27, line 272: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -249,24 +256,24 @@ ARM_Dynarmic_32::ARM_Dynarmic_32(System& system_, CPUInterrupts& interrupt_handl
...
 }
 
 void ARM_Dynarmic_32::SetReg(int index, u64 value) {
-    jit->Regs()[index] = static_cast<u32>(value);
+    jit.load()->Regs()[index] = static_cast<u32>(value);
```
FernandoS27, line 276: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -276,11 +283,11 @@ u128 ARM_Dynarmic_32::GetVectorReg(int index) const {
 void ARM_Dynarmic_32::SetVectorReg(int index, u128 value) {}
 
 u32 ARM_Dynarmic_32::GetPSTATE() const {
-    return jit->Cpsr();
+    return jit.load()->Cpsr();
```
FernandoS27, line 286: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -276,11 +283,11 @@ u128 ARM_Dynarmic_32::GetVectorReg(int index) const {
...
 }
 
 void ARM_Dynarmic_32::SetPSTATE(u32 cpsr) {
-    jit->SetCpsr(cpsr);
+    jit.load()->SetCpsr(cpsr);
```
FernandoS27, line 290: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -301,7 +308,7 @@ void ARM_Dynarmic_32::SetTPIDR_EL0(u64 value) {
 
 void ARM_Dynarmic_32::SaveContext(ThreadContext32& ctx) {
     Dynarmic::A32::Context context;
-    jit->SaveContext(context);
+    jit.load()->SaveContext(context);
```
FernandoS27, line 311: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -314,27 +321,27 @@ void ARM_Dynarmic_32::LoadContext(const ThreadContext32& ctx) {
     context.ExtRegs() = ctx.extension_registers;
     context.SetCpsr(ctx.cpsr);
     context.SetFpscr(ctx.fpscr);
-    jit->LoadContext(context);
+    jit.load()->LoadContext(context);
```
FernandoS27, line 324: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -314,27 +321,27 @@ void ARM_Dynarmic_32::LoadContext(const ThreadContext32& ctx) {
...
 }
 
 void ARM_Dynarmic_32::InvalidateCacheRange(VAddr addr, std::size_t size) {
-    jit->InvalidateCacheRange(static_cast<u32>(addr), size);
+    jit.load()->InvalidateCacheRange(static_cast<u32>(addr), size);
 }
 
 void ARM_Dynarmic_32::ClearExclusiveState() {
-    jit->ClearExclusiveState();
+    jit.load()->ClearExclusiveState();
```
FernandoS27, line 344: ditto.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -345,13 +352,14 @@ void ARM_Dynarmic_32::PageTableChanged(Common::PageTable& page_table,
     auto key = std::make_pair(&page_table, new_address_space_size_in_bits);
     auto iter = jit_cache.find(key);
     if (iter != jit_cache.end()) {
-        jit = iter->second;
+        jit.store(iter->second.get());
```
FernandoS27, line 355: std::memory_order_release. acquire is nor required nor sequencing.

src/core/arm/dynarmic/arm_dynarmic_32.cpp
```diff
@@ -345,13 +352,14 @@ void ARM_Dynarmic_32::PageTableChanged(Common::PageTable& page_table,
...
         LoadContext(ctx);
         return;
     }
-    jit = MakeJit(&page_table);
+    std::shared_ptr new_jit = MakeJit(&page_table);
+    jit.store(new_jit.get());
```
FernandoS27, line 360: ditto.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -109,7 +109,7 @@ class DynarmicCallbacks64 : public Dynarmic::A64::UserCallbacks {
             break;
         }
 
-        parent.jit->HaltExecution(Dynarmic::HaltReason::CacheInvalidation);
+        parent.jit.load()->HaltExecution(Dynarmic::HaltReason::CacheInvalidation);
```
FernandoS27, line 112: acquire.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -130,7 +130,7 @@ class DynarmicCallbacks64 : public Dynarmic::A64::UserCallbacks {
 
     void CallSVC(u32 swi) override {
         parent.svc_swi = swi;
-        parent.jit->HaltExecution(svc_call);
+        parent.jit.load()->HaltExecution(svc_call);
```
FernandoS27, line 133: acquire.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -289,7 +296,7 @@ std::shared_ptr<Dynarmic::A64::Jit> ARM_Dynarmic_64::MakeJit(Common::PageTable*
 
 void ARM_Dynarmic_64::Run() {
     while (true) {
-        const auto hr = jit->Run();
+        const auto hr = jit.load()->Run();
```
FernandoS27, line 299: ditto.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -300,7 +307,7 @@ void ARM_Dynarmic_64::Run() {
 }
 
 void ARM_Dynarmic_64::Step() {
-    jit->Step();
+    jit.load()->Step();
```
FernandoS27, line 310: ditto.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -309,40 +316,40 @@ ARM_Dynarmic_64::ARM_Dynarmic_64(System& system_, CPUInterrupts& interrupt_handl
...
 }
 
 u32 ARM_Dynarmic_64::GetPSTATE() const {
-    return jit->GetPstate();
+    return jit.load()->GetPstate();
 }
 
 void ARM_Dynarmic_64::SetPSTATE(u32 pstate) {
-    jit->SetPstate(pstate);
+    jit.load()->SetPstate(pstate);
```
FernandoS27, line 352: acquire.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -362,45 +369,47 @@ void ARM_Dynarmic_64::SetTPIDR_EL0(u64 value) {
...
 }
 
 void ARM_Dynarmic_64::InvalidateCacheRange(VAddr addr, std::size_t size) {
-    jit->InvalidateCacheRange(addr, size);
+    jit.load()->InvalidateCacheRange(addr, size);
 }
 
 void ARM_Dynarmic_64::ClearExclusiveState() {
-    jit->ClearExclusiveState();
+    jit.load()->ClearExclusiveState();
```
FernandoS27, line 412: acquire

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -362,45 +369,47 @@ void ARM_Dynarmic_64::SetTPIDR_EL0(u64 value) {
...
 
 void ARM_Dynarmic_64::SaveContext(ThreadContext64& ctx) {
-    ctx.cpu_registers = jit->GetRegisters();
-    ctx.sp = jit->GetSP();
-    ctx.pc = jit->GetPC();
-    ctx.pstate = jit->GetPstate();
-    ctx.vector_registers = jit->GetVectors();
-    ctx.fpcr = jit->GetFpcr();
-    ctx.fpsr = jit->GetFpsr();
+    Dynarmic::A64::Jit* j = jit.load();
```
FernandoS27, line 372: acquire

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -362,45 +369,47 @@ void ARM_Dynarmic_64::SetTPIDR_EL0(u64 value) {
...
 
 void ARM_Dynarmic_64::LoadContext(const ThreadContext64& ctx) {
-    jit->SetRegisters(ctx.cpu_registers);
-    jit->SetSP(ctx.sp);
-    jit->SetPC(ctx.pc);
-    jit->SetPstate(ctx.pstate);
-    jit->SetVectors(ctx.vector_registers);
-    jit->SetFpcr(ctx.fpcr);
-    jit->SetFpsr(ctx.fpsr);
+    Dynarmic::A64::Jit* j = jit.load();
```
FernandoS27, line 384: 132 pokemon

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -411,13 +420,14 @@ void ARM_Dynarmic_64::PageTableChanged(Common::PageTable& page_table,
     auto key = std::make_pair(&page_table, new_address_space_size_in_bits);
     auto iter = jit_cache.find(key);
     if (iter != jit_cache.end()) {
-        jit = iter->second;
+        jit.store(iter->second.get());
```
FernandoS27, line 423: release.

src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -411,13 +420,14 @@ void ARM_Dynarmic_64::PageTableChanged(Common::PageTable& page_table,
...
         LoadContext(ctx);
         return;
     }
-    jit = MakeJit(&page_table, new_address_space_size_in_bits);
+    std::shared_ptr new_jit = MakeJit(&page_table, new_address_space_size_in_bits);
+    jit.store(new_jit.get());
```
FernandoS27, line 428: image

merryhime:
Not sure how acquire/release is appropriate here?

behunin:
> Not sure how acquire/release is appropriate here? (If anything, probably relaxed on most loads except for seq_cst on the interrupt load and seq_cst on stores would be more appropriate -- especially given that we would want a total ordering with the halt_reason rmw?)

This is what I refer to when discussing memory_order and the reasons behind each use case.

https://levelup.gitconnected.com/memory-model-basic-d8b5f8fddd5f  

> And given that that's non-trivial to reason about and requires knowledge of which threads call which member functions, the simplest thing seems to be just to acq_rel on all loads given that doing otherwise isn't even a performance advantage on x64?

Basically, yes. 
seq_cst is the default mode for store/load and it is usually not the fastest.

merryhime:
@behunin Thank you. However, please do note my comment was meant to be a polite way of saying "please kindly reevaulate your thinking process as I think you're wrong".

Just to make it clear: Acquire/Release is not appropriate in this case.

behunin:
I can see how you would think I am trying to imply that you don't understand atomics or that you are wrong in your decision making. I am sorry to have upset you or cause you any distress.

My intention was to share what I hold to be my source of understanding of each memory_order, I thought it would work nicely into my next point. The fact that seq_cst is usually slower.

I thought you had come to this conclusion already and was not sure if acquire/release was the right choice.

From the section on Sequential Consistent ordering:
> Sequential consistency ordering not only guarantees the same synchronization among threads as release-acquire ordering , but also guarantees that all threads must see the same order of memory operations. In other words, it guarantees that there is only one common order of memory operations history among different threads.

In the context of this PR, I believe that acquire/release is the right choice for most the most part, considering yuzu needs all the speed it can get.
Except for when sequential ordering is a must.

Please correct me if I am wrong.

lat9nq:
Thank you for addressing the issue so quickly, @merryhime! This solves the races I found when I test locally, too.

FernandoS27:
@merryhime it's fine. It was a suggestion. if you feel this is the way, I'll just go through with what we have.

merryhime:
> In the context of this PR, I believe that acquire/release is the right choice for most the most part, considering yuzu needs all the speed it can get. Except for when sequential ordering is a must.

Sequential ordering is required, because we require a total ordering between `jit` with `halt_reason`. Acquire/Release does not provide a total ordering, as per your own quote.

behunin:
Thank you for that detailed explanation.
> Sequential ordering is required, because we require a total ordering between jit and halt_reason.

I understand what you mean but I am struggling to see this in my mental model of this PR. 
From what I understand `const auto hr = jit.load()->Run()` is the halt_reason. It is then checked by the `Has` function a few lines down. Since the halt_reason comes from the jit being ran, I didn't see the need for ordering.

> There is very little speed benefit to weakened memory orderings on x64. Notably there is zero difference on atomic load.

I see. The only speed benefit would come from less fencing around stores.

merryhime:
There is a variable internal to dynarmic called halt_reason -- the return value of Run is actually the result of a `xchg 0` of this value when the atomic is reset. Internally to dynarmic, halt_reason is implemented in assembly as a SC atomic with `xchg` and `lock or`.

Since std::memory_order_acquire is a non-SC acquire (and thus can be reordered with SC operations of other addresses, and also can be reordered with regards to preceding non-SC release operations), it is possible for the release of jit to be reordered with a rmw of this internal value, resulting in a deadlock due to non-termination of the jit.

behunin:
@merryhime I very much appreciate the time you have taken to explain this PR and the context around atomics.

bunnei (approved):

Superstarlife:
This seems to be causing problems in games like Pokemon Arceus, at least on the steam deck, in the mainline version 985 I can play for hours without crashing or freezing problems.

FernandoS27:
> This seems to be causing problems in games like Pokemon Arceus, at least on the steam deck, in the mainline version 985 I can play for hours without crashing or freezing problems.

this shouldn't cause any issues while the game is executing as the jit is only created and changed at startup.

PR #8190, opened 2022-04-10
ui: Set Link Color when setting theme

Docteh:
Long story short, QT doesn't allow the link colors to be set via their stylesheets.

There are two ways to work with this, specify the color manually for every link (See the About dialog) The other way is to change the default palette.

IsDarkTheme is copy/pasted from src/yuzu/debugger/wait_tree.cpp

I considered setting the link color to be the same light blue as in the license dialog, but with a bit of testing I found that it wasn't a good fit. I lightened it for the dark theme, and darkened it for the light theme. 

Before
image

After
image

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -3652,6 +3652,16 @@ void GMainWindow::UpdateUITheme() {
         setStyleSheet({});
     }
 
+    QPalette new_pal(qApp->palette());
+    if (UISettings::IsDarkTheme()) {
+        new_pal.setColor(QPalette::Text, QColor(255, 255, 255, 255));
+        new_pal.setColor(QPalette::Link, QColor(0, 190, 255, 255));
+    } else {
+        new_pal.setColor(QPalette::Text, QColor(0, 0, 0, 255));
+        new_pal.setColor(QPalette::Link, QColor(0, 140, 200, 255));
+    }
+    qApp->setPalette(new_pal);
+
     QIcon::setThemeName(current_theme);
     QIcon::setThemeSearchPaths(theme_paths);
 }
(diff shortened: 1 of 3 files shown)


src/yuzu/main.cpp
```diff
@@ -3652,6 +3652,18 @@ void GMainWindow::UpdateUITheme() {
         setStyleSheet({});
     }
 
+    QPalette newPal(qApp->palette());
+    if (UISettings::IsDarkTheme()) {
+        newPal.setColor(QPalette::Text, QColor(255,255,255,255));
+        newPal.setColor(QPalette::Link, QColor(0,190,255,255));
+        //newPal.setColor(QPalette::LinkVisited, QColor(0,190,255,255));
```
lioncash, line 3659: Generally we prefer to not keep commented out code around, ditto for the else case

src/yuzu/main.cpp
```diff
@@ -3652,6 +3652,18 @@ void GMainWindow::UpdateUITheme() {
         setStyleSheet({});
     }
 
+    QPalette newPal(qApp->palette());
```
lioncash, line 3655:
```suggestion
    QPalette new_pal(qApp->palette());
```

Docteh:
hi, thanks for the review, I will check into it, and the clang formatting thing in a bit

german77:
Changing between light and dark theme will render the game list incorrectly. This doesn't happen on mainline
image

Docteh:
That is a very interesting interaction between touching QPalette::Text and the themes. 

On the plus side just touching QPalette::Link is working going from light to dark to light again.

german77 (approved):

bunnei (approved):

goldenx86:
Hi, can I mention your PRs in the next progress report?

PR #8192, opened 2022-04-11
bootmanager: Don't create another screenshot request if previous one is not done yet

german77:
It's posible to make yuzu crash if on the loading splash screen you spam the screenshot button.  This was due to the callback being destroyed before it was able to reply. To avoid this issue we avoid requesting a screenshot at all if the backend is not ready

### DIFF
--- a/src/yuzu/bootmanager.cpp
+++ b/src/yuzu/bootmanager.cpp
@@ -933,6 +933,12 @@ void GRenderWindow::CaptureScreenshot(const QString& screenshot_path) {
     auto& renderer = system.Renderer();
     const f32 res_scale = Settings::values.resolution_info.up_factor;
 
+    if (renderer.IsScreenshotPending()) {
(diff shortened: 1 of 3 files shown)


src/yuzu/bootmanager.cpp
```diff
@@ -933,6 +933,11 @@ void GRenderWindow::CaptureScreenshot(const QString& screenshot_path) {
     auto& renderer = system.Renderer();
     const f32 res_scale = Settings::values.resolution_info.up_factor;
 
+    if (renderer.IsScreenshotPending()) {
+        LOG_ERROR(Render, "A screenshot is already requested or in progress, ignoring the request");
```
lioncash, line 937: This should probably be a warning rather than an error, since the previously submitted request is going to succeed

lioncash (approved):

PR #8204, opened 2022-04-14
ui: Fix Game Compatibility list translations

Docteh:
Reported by GillianMC on Discord.  Translations of the states, such as "Perfect" and "Bad" are done, but not used.

Looks to be a small quirk in the QT API.

Technical detail
setText(QObject::tr(status.text));
bringing up QObject broke the connection to the GameListItemCompat class.

Also related, certain strings are duplicated in translation, which can confuse translators. 
QT_TR_NOOP("Perfect") -> QT_TRANSLATE_NOOP("CompatDB", "Perfect")

If you're wondering about "Not Tested" being left alone, its not possible to submit a compatibility report stating that you didn't test the game.

Also I thought removing the space from `Great` in src/yuzu/compatdb.ui would disconnect the translation but things are still Gut 

image

### DIFF
--- a/src/yuzu/configuration/configure_dialog.cpp
+++ b/src/yuzu/configuration/configure_dialog.cpp
@@ -25,6 +25,7 @@
 #include "yuzu/configuration/configure_ui.h"
 #include "yuzu/configuration/configure_web.h"
 #include "yuzu/hotkeys.h"
+#include "yuzu/uisettings.h"
 
 ConfigureDialog::ConfigureDialog(QWidget* parent, HotkeyRegistry& registry,
                                  InputCommon::InputSubsystem* input_subsystem,
@@ -169,6 +170,8 @@ void ConfigureDialog::PopulateSelectionList() {
 
 void ConfigureDialog::OnLanguageChanged(const QString& locale) {
     emit LanguageChanged(locale);
+    //  Reloading the game list is needed to force retranslation.
+    UISettings::values.is_game_list_reload_pending = true;
     // first apply the configuration, and then restore the display
     ApplyConfiguration();
     RetranslateUI();
--- a/src/yuzu/game_list_p.h
+++ b/src/yuzu/game_list_p.h
@@ -164,8 +164,8 @@ public:
         }
         const CompatStatus& status = iterator->second;
         setData(compatibility, CompatNumberRole);
-        setText(QObject::tr(status.text));
-        setToolTip(QObject::tr(status.tooltip));
+        setText(tr(status.text));
+        setToolTip(tr(status.tooltip));
         setData(CreateCirclePixmapFromColor(status.color), Qt::DecorationRole);
     }
 
--- a/src/yuzu/compatdb.ui
+++ b/src/yuzu/compatdb.ui
@@ -86,7 +86,7 @@
     <item row="4" column="0">
      <widget class="QRadioButton" name="radioButton_Great">
       <property name="text">
-       <string>Great </string>
+       <string>Great</string>
       </property>
      </widget>
     </item>


src/yuzu/game_list_p.h
```diff
@@ -147,13 +147,13 @@ class GameListItemCompat : public GameListItem {
...
+            {QStringLiteral("2"),  {QStringLiteral("#94b242"), QT_TRANSLATE_NOOP("CompatDB", "Okay"),       QT_TR_NOOP("Game functions with major graphical or audio glitches, but game is playable from start to finish with\nworkarounds.")}},
+            {QStringLiteral("3"),  {QStringLiteral("#f2d624"), QT_TRANSLATE_NOOP("CompatDB", "Bad"),        QT_TR_NOOP("Game functions, but with major graphical or audio glitches. Unable to progress in specific areas due to glitches\neven with workarounds.")}},
+            {QStringLiteral("4"),  {QStringLiteral("#FF0000"), QT_TRANSLATE_NOOP("CompatDB", "Intro/Menu"), QT_TR_NOOP("Game is completely unplayable due to major graphical or audio glitches. Unable to progress past the Start\nScreen.")}},
+            {QStringLiteral("5"),  {QStringLiteral("#828282"), QT_TRANSLATE_NOOP("CompatDB", "Won't Boot"), QT_TR_NOOP("The game crashes when attempting to startup.")}},
+            {QStringLiteral("99"), {QStringLiteral("#000000"), QT_TR_NOOP("Not Tested"),                    QT_TR_NOOP("The game has not yet been tested.")}},
```
lioncash, line 156: Shouldn't this label also be converted over to `QT_TRANSLATE_NOOP` as well as the description comments?

src/yuzu/configuration/configure_dialog.cpp
```diff
@@ -169,6 +170,8 @@ void ConfigureDialog::PopulateSelectionList() {
 
 void ConfigureDialog::OnLanguageChanged(const QString& locale) {
     emit LanguageChanged(locale);
+    //  Reloading the game list is needed to force retranslation.
+    UISettings::values.is_game_list_reload_pending.exchange(true);
```
lioncash, line 174: Since the return value from exchange isn't used, this should be able to be:

```suggestion
    UISettings::values.is_game_list_reload_pending = true;
```

Docteh: I was wondering about that. I looked up what exchange does and it apparently does a [read-modify-write](https://en.cppreference.com/w/cpp/atomic/atomic/exchange)

If you do a code search for `is_game_list_reload_pending` you'll find 4 other times where exchange is used, but the return value isn't used. 

Maybe there is some multithreaded related stuff happening under the hood?

https://github.com/yuzu-emu/yuzu/blob/46da380b571b46d85bf0a6e62ce7a74c9fdce832/src/yuzu/configuration/configure_per_game_addons.cpp#L66-L67

https://github.com/yuzu-emu/yuzu/blob/46da380b571b46d85bf0a6e62ce7a74c9fdce832/src/yuzu/configuration/configure_filesystem.cpp#L134-L139


https://github.com/yuzu-emu/yuzu/blob/46da380b571b46d85bf0a6e62ce7a74c9fdce832/src/yuzu/configuration/configure_general.cpp#L77-L89

https://github.com/yuzu-emu/yuzu/blob/46da380b571b46d85bf0a6e62ce7a74c9fdce832/src/yuzu/configuration/configure_ui.cpp#L123-L125

lioncash: It is fine to use an assignment here, since an `operator=` for an atomic will atomically set the value as well. `exchange` in all of these cases isn't required either and can just be assignments

src/yuzu/game_list_p.h
Docteh: `QT_TR_NOOP` doesn't require a specified context, but `QR_TRANSLATE_NOOP` does

QT_TR_NOOP("The game crashes when attempting to startup.")
would become
QT_TRANSLATE_NOOP("GameListItemCompat", "The game crashes when attempting to startup.")

I could convert `QT_TR_NOOP("Not Tested")` to QT_TRANSLATE_NOOP("GameListItemCompat", "Not Tested")

then at least the columns are using the same macros. 

I just tried QT_TRANSLATE_NOOP("CompatDB", "Not Tested") and it grabbed the translation from the other context, but it will add a string to the translation site
image

lioncash: > `QT_TR_NOOP` doesn't require a specified context, but `QR_TRANSLATE_NOOP` does

I am very aware how the localizing in Qt works, considering I put the initial localizing plumbing in yuzu for the event handling. This is fine for consistency. But using different contexts in the same table without a comment is bound to be confusing, so if using a different context must absolutely be done, it should have a little comment accompanying it.

Docteh:
Tried `UISettings::values.is_game_list_reload_pending = true;` and it WorksForMe

Is there ever a reason to translate string `yuzu` in like the About Dialog, or the main Window title? 

I tried looking at Qt Creator with the goal of merging the strings, and all I could find is a flag to not translate certain labels.
For example:
`<string>yuzu</string>` to `<string notr="true">yuzu</string>`

lioncash:
Qt doesn't merge translation strings (and it should never do this, ever), since assuming the same translated string will always contextually mean the same thing in different parts of the program is a very assuming and opinionated view about language and how they function as a whole (they do not obey all the rules of English or some other Latin-based languag), which is generally why you want the strings left alone, because the translator has the power to make that choice rather than having it made for them without having an opt-out.

It would also, in my opinion, look horrible code-wise if we managed that manually

Docteh:
I believe merging the 6 compatibility types makes sense as in the compatibility report form, users are selecting the option they feel should show up on the games list. But I'll check for your reply.

Yuzu Report Compatibility Form


I'm not planning on touching it, but the title for the configuration reset dialog window title is  just `yuzu`. A lot of the other little popups have titles. Like the one asking for Derivation Keys. 
image

Docteh:
I reverted the `QT_TR_NOOP("Perfect")` -> `QT_TRANSLATE_NOOP("CompatDB", "Perfect")` changes and squashed down to one commit.

lioncash (approved):

Fs00:
Can this fix be ported to Citra too?
The same bug is still present there. 🙏

PR #8212, opened 2022-04-15
dynarmic: Exit Run on every interrupt and after every service call

merryhime:
I am beginning to question the accuracy of the original code in deciding when returning is appropriate. Would it be possible for someone who has actually reverse engineered and understood HOS's rescheduling algorithm (i.e. @bunnei) to please confirm this?

Here we return from Run after any interrupt, cache invalidation, or svc call. Note that this currently also returns on every cache invalidation as well, which is excessive and clearly incorrect.


merryhime:
I need a break.

PR #8224, opened 2022-04-18
ui: translate hat directions

Docteh:
988 Yuzu UI set to Russian
image

Removing QObject:: would move the translation to ConfigureInputPlayer context, but that context is a bit busy, so I'm leaning towards not touching it.

image

image
In Russian, the last letter gets cut off for the word right.

### DIFF
--- a/src/yuzu/configuration/configure_input_player.cpp
+++ b/src/yuzu/configuration/configure_input_player.cpp
@@ -10,6 +10,7 @@
 #include <QMenu>
 #include <QMessageBox>
 #include <QTimer>
+#include "common/assert.h"
 #include "common/param_package.h"
 #include "core/hid/emulated_controller.h"
 #include "core/hid/hid_core.h"
@@ -119,6 +120,23 @@ QString GetButtonName(Common::Input::ButtonNames button_name) {
     }
 }
 
+QString GetDirectionName(const std::string& direction) {
+    if (direction == "left") {
+        return QObject::tr("Left");
+    }
+    if (direction == "right") {
+        return QObject::tr("Right");
+    }
+    if (direction == "up") {
+        return QObject::tr("Up");
+    }
+    if (direction == "down") {
+        return QObject::tr("Down");
+    }
+    UNIMPLEMENTED_MSG("Unimplemented direction name={}", direction);
+    return QString::fromStdString(direction);
+}
+
 void SetAnalogParam(const Common::ParamPackage& input_param, Common::ParamPackage& analog_param,
                     const std::string& button_name) {
     // The poller returned a complete axis, so set all the buttons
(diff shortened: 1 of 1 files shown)


merryhime:
Hey thanks, consider instead:

```const QString hat = GetDirectionName(param.Get("direction", ""));```

where:

```QString GetDirectionName(const std::string& direction) {
    if (direction == "left") {
        return QObject::tr("Left");
    }
    if (direction == "right") {
        return QObject::tr("Right");
    }
    if (direction == "up") {
        return QObject::tr("Up");
    }
    if (direction == "down") {
        return QObject::tr("Down");
    }
    return QString::fromStdString(direction);
}```

**Comment by Docteh** [comment]
That changes the translation context from ConfigureInputPlayer to QObject, but after thinking about it, it makes sense.

QOBject's "Left" is for the times code is asking for it, for these input buttons. and ConfigureInputPlayer's "Left" is the times it shows up in the UI

Let me know if this needs squishing down to 1 commit.

**Comment by Morph1984** [comment]
Yes, ideally this should squash down into 1 as the second commit completely replaces the first

**Review by merryhime — APPROVED** [review]

File: src/yuzu/configuration/configure_input_player.cpp [hunk 1]
```diff
@@ -119,6 +119,22 @@ QString GetButtonName(Common::Input::ButtonNames button_name) {
     }
 }
 
+QString GetDirectionName(const std::string& direction) {
```
lioncash, line 122:
```suggestion
QString GetDirectionName(std::string_view direction) {
```

It's preferable to use a `string_view` here

src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -119,6 +119,22 @@ QString GetButtonName(Common::Input::ButtonNames button_name) {
...
+    if (direction == "right") {
+        return QObject::tr("Right");
+    }
+    if (direction == "up") {
+        return QObject::tr("Up");
+    }
+    if (direction == "down") {
+        return QObject::tr("Down");
+    }
+    return QString::fromStdString(direction);
```
lioncash, line 135: It feels like we should have an assert or something here, so if (for whatever reason) anything else is added and not handled gets caught, as well as any typo that slips through.

lioncash (commented):
Just two minor nits, but otherwise looks good

src/yuzu/configuration/configure_input_player.cpp
Docteh: I'm not super familiar with well written errors, but something along these lines?

`ASSERT_MSG(false, "GetDirectionName \"{}\" unhandled!", direction);`

image

`UNIMPLEMENTED_MSG("Unimplemented direction name={}", direction);`
image

`Unexpected Direction: lolcats`
`Unexpected Direction "lolcats"`
`Unexpected Direction name=lolcats`

`Unexpected direction assignment name=lolcats`

lioncash: Yeah, either works

bunnei:
LGTM, thanks!

PR #8231, opened 2022-04-20
input_common: Ignore boost uninitialized local variable

german77:
Addresses one of the issues of #8010

### DIFF
--- a/src/input_common/helpers/udp_protocol.h
+++ b/src/input_common/helpers/udp_protocol.h
@@ -8,7 +8,16 @@
 #include <optional>
 #include <type_traits>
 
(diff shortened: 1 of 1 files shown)


src/input_common/helpers/udp_protocol.h
```diff
@@ -8,7 +8,10 @@
 #include <optional>
 #include <type_traits>
 
+#pragma warning(push)
```
lioncash, line 11: these pragamas need to be surrounded with an ifdef that checks for msvc otherwise this will cause warnings on other platforms

lioncash (approved):

PR #8225, opened 2022-04-18
yuzu: mention GPLv3.0 in about dialog

Tachi107:
Follow-up to 284934ebfdf5e530c960cf69969172ff76f40bea

Fixes #8218

Only tested with the Italian translation, but it should work fine for all the languages

image

PR #8233, opened 2022-04-21
yuzu: fix "GPLv3.0++" in Chinese about dialog

Tachi107:
Introduced in 24d7a027dcaa7572108c620b86ce4ec5878bbcb1

Thanks to @mzji for spotting the issue


FearlessTobi:
Thanks for your contribution.
However, the translations files should never be manually changed.
The mentioned fixes should instead be made via Transifex.

Closing.

Tachi107:
Oh, ok, thanks

Docteh:
@Tachi107 are you registered as a translator on transifex? I'm curious the strings over there have been updated after #8225

Tachi107:
> @Tachi107 are you registered as a translator on transifex? I'm curious the strings over there have been updated after #8225

Unfortunately not. I tried to register but it seemed that the page had some issues, and after a few trials I gave up. Feel free to correct my errors if you already have an account :)

Docteh:
@FearlessTobi Since the change in #8225 is just incrementing the license from `GPLv2.0` to `GPLv3.0+` would it be possible for someone registered as  say Deutsch to suggest translations for the other languages?

PR #8236, opened 2022-04-22
Changes to language order in General -> UI -> Interface Language

Docteh:
I don't know why, but the development tools embed the language files in the following order
ca da cs de el es fi fr id it nb nl pl sv vi ko_KR ja_JP tr_TR pt_BR pt_PT ru_RU zh_CN zh_TW vi_VN

Sorted alphabetically
ca cs da de el es fi fr id it ja_JP ko_KR nb nl pl pt_BR pt_PT ru_RU sv tr_TR vi vi_VN zh_CN zh_TW

Current order
image

For the Yuzu User Interface Language, I believe the options should be in native language. 

image

Note: Not touching the System Language, those have a nice enumeration order provided by "N"
image

Basically with this Draft PR, I'm asking about two things

1. Can we sort the list alphabetically by the language code?
2. Should this list be in native strings, or left as the English names of the languages?

### DIFF
--- a/src/yuzu/configuration/configure_ui.cpp
+++ b/src/yuzu/configuration/configure_ui.cpp
@@ -9,6 +9,7 @@
 #include <QDirIterator>
 #include "common/common_types.h"
 #include "common/fs/path_util.h"
+#include "common/logging/log.h"
 #include "common/settings.h"
 #include "core/core.h"
 #include "ui_configure_ui.h"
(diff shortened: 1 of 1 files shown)


src/yuzu/configuration/configure_ui.cpp
```diff
@@ -169,18 +169,40 @@ void ConfigureUi::RetranslateUI() {
     }
 }
 
+// Intentionally avoiding translation
+QString ConfigureUi::languageCodeToString(QString localeCode) {
```
lioncash, line 173: You may want to look into doing something similar to what Dolphin [does](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/DolphinQt/Settings/InterfacePane.cpp#L30) rather than hardcoding in repeated if statements on a per-language basis

src/yuzu/configuration/configure_ui.h
```diff
@@ -38,6 +38,7 @@ private slots:
     void changeEvent(QEvent*) override;
     void RetranslateUI();
 
+    QString languageCodeToString(QString localeCode);
```
lioncash, line 41: Unless it's a Qt function being overridden, functions should be cased like CamelCase, not camelCase. Our convention for variable names uses `snake_case` as well.

src/yuzu/configuration/configure_ui.cpp
```diff
@@ -169,18 +169,40 @@ void ConfigureUi::RetranslateUI() {
...
 void ConfigureUi::InitializeLanguageComboBox() {
     ui->language_combobox->addItem(tr("<System>"), QString{});
-    ui->language_combobox->addItem(tr("English"), QStringLiteral("en"));
-    QDirIterator it(QStringLiteral(":/languages"), QDirIterator::NoIteratorFlags);
-    while (it.hasNext()) {
-        QString locale = it.next();
+    ui->language_combobox->addItem(QStringLiteral("English [en]"), QStringLiteral("en"));
+    QDir languagesDir;
+    languagesDir.setPath(QStringLiteral(":/languages"));
+    foreach (const QString& file, languagesDir.entryList()) {
```
lioncash, line 200: foreach is a Qt-ism that doesn't apply anymore in C++11 and onwards. This can be

```suggestion
    for (const QString& file : languagesDir.entryList()) {
```

src/yuzu/configuration/configure_ui.cpp
```diff
@@ -169,18 +169,40 @@ void ConfigureUi::RetranslateUI() {
...
+}
+
 void ConfigureUi::InitializeLanguageComboBox() {
     ui->language_combobox->addItem(tr("<System>"), QString{});
-    ui->language_combobox->addItem(tr("English"), QStringLiteral("en"));
-    QDirIterator it(QStringLiteral(":/languages"), QDirIterator::NoIteratorFlags);
-    while (it.hasNext()) {
-        QString locale = it.next();
+    ui->language_combobox->addItem(QStringLiteral("English [en]"), QStringLiteral("en"));
+    QDir languagesDir;
```
lioncash, line 198: This and the following line can be combined like so:

```suggestion
    QDir languages_dir{QStringLiteral(":/languages")};
```

src/yuzu/configuration/configure_ui.cpp
```diff
@@ -169,18 +169,40 @@ void ConfigureUi::RetranslateUI() {
...
+    languagesDir.setPath(QStringLiteral(":/languages"));
+    foreach (const QString& file, languagesDir.entryList()) {
+        QString locale = file;
         locale.truncate(locale.lastIndexOf(QLatin1Char{'.'}));
-        locale.remove(0, locale.lastIndexOf(QLatin1Char{'/'}) + 1);
-        const QString lang = QLocale::languageToString(QLocale(locale).language());
+        QString lang = QStringLiteral("%1 [%2]").arg(languageCodeToString(locale), locale);
```
lioncash, line 203: We generally mark locals as `const` if they aren't modified

Docteh:
It might not be necessary, but I like how Yuzu's current implementation can have languages added to build without requiring code modification. 

Also looking at my notes its easily possible to make builds that completely lack translations

In Dolphin [they hardcoded the language order](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/DolphinQt/Settings/InterfacePane.cpp#L30) 

I guess as an alternative, print out the expected languages in order if they're available, and print any unexpected languages at the end?

src/yuzu/configuration/configure_ui.cpp
```diff
@@ -170,14 +170,58 @@ void ConfigureUi::RetranslateUI() {
...
+    QDir languages_dir{QStringLiteral(":/languages")};
+    for (const QString& file : languages_dir.entryList()) {
+        QString locale = file;
         locale.truncate(locale.lastIndexOf(QLatin1Char{'.'}));
-        locale.remove(0, locale.lastIndexOf(QLatin1Char{'/'}) + 1);
-        const QString lang = QLocale::languageToString(QLocale(locale).language());
+        QString language_name = QLocale::languageToString(QLocale(locale).language());
+        for (const auto& lang : languages) {
+            if (QString::fromLatin1(lang.id) == locale) {
+                language_name = lang.name;
```
lioncash, line 221: Once this matches, shouldn't it break out of the loop instead of continuing onwards?

src/yuzu/configuration/configure_ui.h
lioncash, line 41: This no longer seems to be used

src/yuzu/configuration/configure_ui.cpp
Docteh: Yes
I'm really hoping its just `break;` to do that.

Docteh:
@lioncash I'm leaving the list sorted by language code, but I still have Dolphin's sorted order kind of in the back of my mind. 

other bits are fixed up

lioncash:
I would strongly suggest sorting by language name, not the language code, since that's how the majority of applications order their language listings and I would prefer it if we maintain that convention. Most users don't care about the language code at all more than the name of the language itself.

It's also super weird UX to have Français appear after Suomi, for example. This would be no better than providing an unordered list.

Docteh:
This is where it's at, unexpected languages at the bottom, English is properly in the list. 
image
image


Tiếng Việt [vi]
Tiếng Việt [vi_VN]

Maybe this would make sense:
Tiếng Việt (Việt Nam) [vi_VN]

Docteh:
After using the language list for a bit, having the language ids in the list under normal circumstances stopped making sense.

I don't think users will be interested in seeing that Deutsch is [de]. but if somebody is tinkering and a `de_DE` shows up, it'll be at the bottom as `German [de_DE]`

image

I've disabled clang-format for the languages list, as it has a different opinion on where to align the trailing comments (AlignTrailingComments) every time a change is made in there.

lioncash (approved):

PR #8243, opened 2022-04-23
CMakeLists: Revert ad55faaa3

lat9nq:
The premise behind ad55faaa3 was due to an issue between Conan's libiconv package and compiling SDL2 from our externals. Since none of our Conan externals require libiconv any longer, though, we can remove downloading our own Boost package and just rely on Conan again.

Additionally, removing CONFIG from the find_package(boost) call fixes issues with finding Boost on Fedora and MSYS2, which was the main motivation for this. (Revert 697a2c001)

Also, remove QUIET since if something goes wrong finding Boost, this makes it harder to tell what went wrong.


lat9nq:
Reminder for myself to axe https://github.com/yuzu-emu/ext-windows-bin/pull/19 if this PR is merged.

Morph1984 (approved):

PR #8245, opened 2022-04-23
general: Convert source file copyright comments over to SPDX

### LINKED ISSUE #8242: GPL-2.0-only files (license incompatibility)
There are still some GPL-2.0-only files in the project, copyright by Atmosphère-NX:

           src/common/bit_set.h
           src/core/file_sys/fsmitm_romfsbuild.cpp
           src/core/file_sys/fsmitm_romfsbuild.h
           src/core/memory/dmnt_cheat_types.h
           src/core/memory/dmnt_cheat_vm.cpp
           src/core/memory/dmnt_cheat_vm.h

### LINKED ISSUE #8244: Convert source file copyright comments over to SPDX
A good first issue for anyone looking to contribute to yuzu, considering this is mostly a find-replace operation.

At the top of several of our C++ source and header files, we have a comment that looks like so:

```cpp
// Copyright [year] yuzu emulator team
```

e.g.

```cpp
// Copyright 2018 yuzu emulator team
```

The problem here is, since we relicensed, these comments are no longer accurate. Since they would all need to be changed anyway, this would be a good time to use [SPDX](https://spdx.dev), a standardized way to communicate licensing information (which would make doing find/replaces nicer in the future if we ever need to do that again).

So, the above comment corrected for our current license would look like:

```cpp
// Copyright 2018 yuzu emulator team
// SPDX-License-Identifier: GPL-3.0-or-later
```

One nice thing being we don't need 3 lines, just two.

Note that the year should be left as is. It establishes the creation year, not a "last updated" year.

Morph1984:
Closes #8244 
Fixes #8242 

This formats all copyright comments according to SPDX formatting guidelines.
Additionally, this resolves the remaining GPLv2 only licensed files by relicensing them to GPLv2.0-or-later.

### DIFF
--- a/src/audio_core/algorithm/filter.cpp
+++ b/src/audio_core/algorithm/filter.cpp
@@ -1,6 +1,5 @@
-// Copyright 2018 yuzu Emulator Project
-// Licensed under GPLv2 or any later version
-// Refer to the license.txt file included.
+// SPDX-FileCopyrightText: Copyright 2018 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
 
 #define _USE_MATH_DEFINES
 
--- a/src/audio_core/algorithm/filter.h
+++ b/src/audio_core/algorithm/filter.h
@@ -1,6 +1,5 @@
-// Copyright 2018 yuzu Emulator Project
-// Licensed under GPLv2 or any later version
-// Refer to the license.txt file included.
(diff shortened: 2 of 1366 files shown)


lioncash:
You may want to see @bgermann's comment about using GPLv2.0-or-later instead, since what they say is correct (also I was a wiener and completely overlooked that GPLv2+ is forward compatible for some reason)

Docteh:
1336? I wanted the scroll through the changes, but my web browser really doesn't like that :D

PR #8277, opened 2022-04-28
core: hid: Allow to reconnect all players on button input

german77:
Currently only the first player was able to turn on the controller with any button input. This PR tries to add this feature to all players that doesn't have the default keyboard mappings.


german77:
As commended by other devs. This feature might create undesired behaviors if the same controller is mapped to multiple players. I still believe these cases would be rare. Please do not merge before EA testing

PR #8293, opened 2022-05-02
ui: retranslate the network tab

Docteh:
Looks like it was just missed when the tab was added, as currently the Network Tab only has one item
RetranslateUI is used more commonly throughout the project

If you open UI in English and switch to Japanese, Network tab isn't told to re translate
Before example
image

After example
image

### DIFF
--- a/src/yuzu/configuration/configure_network.cpp
+++ b/src/yuzu/configuration/configure_network.cpp
@@ -26,7 +26,15 @@ void ConfigureNetwork::ApplyConfiguration() {
     Settings::values.network_interface = ui->network_interface->currentText().toStdString();
 }
 
-void ConfigureNetwork::RetranslateUi() {
+void ConfigureNetwork::changeEvent(QEvent* event) {
+    if (event->type() == QEvent::LanguageChange) {
+        RetranslateUI();
+    }
+
+    QWidget::changeEvent(event);
+}
+
+void ConfigureNetwork::RetranslateUI() {
     ui->retranslateUi(this);
 }
 
--- a/src/yuzu/configuration/configure_network.h
+++ b/src/yuzu/configuration/configure_network.h
@@ -18,9 +18,10 @@ public:
     ~ConfigureNetwork() override;
 
     void ApplyConfiguration();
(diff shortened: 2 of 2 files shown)


src/yuzu/configuration/configure_network.cpp
```diff
@@ -26,7 +26,14 @@ void ConfigureNetwork::ApplyConfiguration() {
     Settings::values.network_interface = ui->network_interface->currentText().toStdString();
 }
 
-void ConfigureNetwork::RetranslateUi() {
+void ConfigureNetwork::changeEvent(QEvent* event) {
+    if (event->type() == QEvent::LanguageChange) {
+        RetranslateUI();
+    }
+    QWidget::changeEvent(event);
```
Morph1984, line 33:
```suggestion
    }
    
    QWidget::changeEvent(event);
```
For consistency with other files

src/yuzu/configuration/configure_network.h
```diff
@@ -18,10 +18,11 @@ class ConfigureNetwork : public QWidget {
     ~ConfigureNetwork() override;
 
     void ApplyConfiguration();
-    void RetranslateUi();
+    void RetranslateUI();
```
Morph1984, line 21: Should be private

src/yuzu/configuration/configure_network.h
```diff
@@ -18,10 +18,11 @@ class ConfigureNetwork : public QWidget {
...
 
 private:
     void SetConfiguration();
+    void changeEvent(QEvent*) override;
```
Morph1984, line 25: Should be moved above SetConfiguration (and RetranslateUI after it's been made private)

Docteh:
@Morph1984 done (fingers crossed)

lioncash (approved):

baka0815:
Is there a reason for ``changeEvent`` starting with a lower case letter? All other methods start with an uppercase letter.

german77:
`changeEvent` is an override from Qt class. Qt always starts functions with lowercase while yuzu always use uppercase.

PR #8295, opened 2022-05-02
gc_adapter: fix libusb import on GCC11.2

behunin:
Just to silence an intermittent error.
GCC11.2 complains cannot find 'libusb.h' during a fresh build.


lioncash:
LGTM once the commit title is updated to better describe the change

lioncash:
It does seem weird that other projects (e.g. Dolphin) don't seem to be hitting this include path difference though

Morph1984:
You'd probably want to update the commit as well

Morph1984 (approved):

PR #8308, opened 2022-05-05
service: hid: Disable correctly motion input

german77:
When motion is disabled it doesn't stop updating. Instead it just returns the starting orientation. This solves issues where pokemon let's go is spamming `StopSixAxisSensor`.

While working on these changes I noticed that I totally missed the delta_time parameter. Setting this value correctly fixes the issues from #7707 and allows us to have the same refresh rate as the switch.

### DIFF
--- a/src/core/hle/service/hid/controllers/npad.cpp
+++ b/src/core/hle/service/hid/controllers/npad.cpp
@@ -3,7 +3,9 @@
 
 #include <algorithm>
 #include <array>
+#include <chrono>
 #include <cstring>
+
 #include "common/assert.h"
 #include "common/bit_field.h"
 #include "common/common_types.h"
@@ -547,7 +549,9 @@ void Controller_NPad::OnMotionUpdate(const Core::Timing::CoreTiming& core_timing
 
         const auto set_motion_state = [&](SixAxisSensorState& state,
                                           const Core::HID::ControllerMotion& hid_state) {
+            using namespace std::literals::chrono_literals;
             static constexpr SixAxisSensorState default_motion_state = {
+                .delta_time = std::chrono::nanoseconds(5ms).count(),
                 .accel = {0, 0, -1.0f},
                 .orientation =
                     {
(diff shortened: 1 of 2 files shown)


src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -510,6 +510,17 @@ void Controller_NPad::OnMotionUpdate(const Core::Timing::CoreTiming& core_timing
         return;
     }
 
+    constexpr SixAxisSensorState default_motion_state = {
```
lioncash, line 513: This looks like it can be `static constexpr inside of the lambda below, since those are its only two usages

src/core/hle/service/hid/hid.cpp
```diff
@@ -525,10 +524,10 @@ void Hid::StartSixAxisSensor(Kernel::HLERequestContext& ctx) {
...
     const auto result = controller.SetSixAxisEnabled(parameters.sixaxis_handle, true);
 
-    LOG_DEBUG(Service_HID,
-              "called, npad_type={}, npad_id={}, device_index={}, applet_resource_user_id={}",
-              parameters.sixaxis_handle.npad_type, parameters.sixaxis_handle.npad_id,
-              parameters.sixaxis_handle.device_index, parameters.applet_resource_user_id);
+    LOG_INFO(Service_HID,
+             "called, npad_type={}, npad_id={}, device_index={}, applet_resource_user_id={}",
+             parameters.sixaxis_handle.npad_type, parameters.sixaxis_handle.npad_id,
+             parameters.sixaxis_handle.device_index, parameters.applet_resource_user_id);
```
Morph1984, line 530: Perhaps move these to #8309

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -548,6 +548,7 @@ void Controller_NPad::OnMotionUpdate(const Core::Timing::CoreTiming& core_timing
         const auto set_motion_state = [&](SixAxisSensorState& state,
                                           const Core::HID::ControllerMotion& hid_state) {
             static constexpr SixAxisSensorState default_motion_state = {
+                .delta_time = 5 * 1000 * 1000, // 5ms
```
lioncash, line 551: Feels like you'd be able to use std::chrono to be able to state the time more clearly, even if the raw value is used, like:

```cpp
.delta_time = std::chrono::time_format_to_convert_to(5ms).count();
```

Morph1984 (approved):

PR #8292, opened 2022-05-01
Update translations (2022-05-01)

zhaobot:
This PR is generated by [zhaowenlan1779/txbot](https://github.com/zhaowenlan1779/txbot).

PR #8310, opened 2022-05-06
Translations Update (2022-05-05)

Docteh:
Hi, I'm not sure how urgent updating translations are, but I like the idea of the monthly schedule.

Instead of grabbing languages that are at 47% or higher, I'm merely updating the languages that are there.
Finnish is at 46.6% 🇫🇮 
`/usr/bin/tx pull -f -l "$(ls *.ts |tr '\n' ',' | sed  "s/\.ts//g")"`

#8225 changed the contents of the About Box, there are a few languages that still need retranslation
ca cs da el fi fr id it ja_JP nb nl pl sv tr_TR vi vi_VN


FearlessTobi:
Thanks for your contribution, however we already have an automatic system for updating the translation files in regular intervals.
You can see https://github.com/yuzu-emu/yuzu/pull/8292 and https://github.com/zhaowenlan1779/txbot for more details.

lioncash:
translations were updated in #8325

PR #8311, opened 2022-05-06
vk_rasterizer: fix stencil test when two faces are disabled

asLody:
It should not affect games using NVN api, as it seems that two faces are always enabled.
This may fix some ogl games, but idk exactly what is fixed.


bunnei:
Thanks!

PR #8318, opened 2022-05-08
Update some files with Qt 5.15.2 best practices in mind

Docteh:
Update some files with Qt 5.15.2 best practices in mind

There was some discussion about updating to Qt6 and I figured I would
work on some smaller parts. For Windows platform the WinMain function has moved
from the Qt5::WinMain to a new one called Qt6::EntryPointPrivate

Also Qt5 supports versionless CMake targets
https://www.qt.io/blog/versionless-cmake-targets-qt-5.15

These other changes in this commit are to support Qt6, but in ways that don't mess with Qt5.

src/yuzu/bootmanager.cpp: Qt6 complains about not being able to know to use QPoint or QPointF, picking QPoint
src/yuzu/bootmanager.h: Qt6 prefers that QStringList.h be included rather than an empty class definition
src/yuzu/configuration/configure_system.cpp: toULongLong intends to return unsigned 64 bit integer, but
   Settings::values.rng_seed is only 32 bits wide
src/yuzu/game_list.cpp: Qt6 returns a different datatype for QStringList.length than Qt5,
   it used to be int, but in Qt6 its now qsizetype
src/yuzu/loading_screen.cpp: Qt5's for QStyleOption.init say to switch to initFrom.
   The QStyleOption.init doesn't exist in Qt6
src/yuzu/main.cpp: Another QPointer and QStringList.size, lets standardize on count()

---
I did a hasty "Make Qt6 work just for me" build, and now I'm going back and the plan is to do everything nicely in a way that Qt5 and Qt6 are happy

### DIFF
--- a/src/yuzu/CMakeLists.txt
+++ b/src/yuzu/CMakeLists.txt
@@ -187,7 +187,7 @@ if (ENABLE_QT_TRANSLATION)
     # Update source TS file if enabled
     if (GENERATE_QT_TRANSLATION)
         get_target_property(SRCS yuzu SOURCES)
-        qt5_create_translation(QM_FILES
+        qt_create_translation(QM_FILES
             ${SRCS}
             ${UIS}
             ${YUZU_QT_LANGUAGES}/en.ts
@@ -203,7 +203,7 @@ if (ENABLE_QT_TRANSLATION)
     list(REMOVE_ITEM LANGUAGES_TS ${YUZU_QT_LANGUAGES}/en.ts)
 
     # Compile TS files to QM files
-    qt5_add_translation(LANGUAGES_QM ${LANGUAGES_TS})
+    qt_add_translation(LANGUAGES_QM ${LANGUAGES_TS})
 
     # Build a QRC file from the QM file list
     set(LANGUAGES_QRC ${CMAKE_CURRENT_BINARY_DIR}/languages.qrc)
@@ -215,7 +215,7 @@ if (ENABLE_QT_TRANSLATION)
     file(APPEND ${LANGUAGES_QRC} "</qresource></RCC>")
 
     # Add the QRC file to package in all QM files
-    qt5_add_resources(LANGUAGES ${LANGUAGES_QRC})
+    qt_add_resources(LANGUAGES ${LANGUAGES_QRC})
 else()
     set(LANGUAGES)
 endif()
@@ -236,7 +236,11 @@ if (APPLE)
     set_target_properties(yuzu PROPERTIES MACOSX_BUNDLE_INFO_PLIST ${CMAKE_CURRENT_SOURCE_DIR}/Info.plist)
 elseif(WIN32)
     # compile as a win32 gui application instead of a console application
-    target_link_libraries(yuzu PRIVATE Qt5::WinMain)
+    if (QT_VERSION VERSION_GREATER 6)
+        target_link_libraries(yuzu PRIVATE Qt6::EntryPointPrivate)
+    else()
+        target_link_libraries(yuzu PRIVATE Qt5::WinMain)
+    endif()
     if(MSVC)
         set_target_properties(yuzu PROPERTIES LINK_FLAGS_RELEASE "/SUBSYSTEM:WINDOWS")
     elseif(MINGW)
@@ -247,7 +251,7 @@ endif()
 create_target_directory_groups(yuzu)
 
 target_link_libraries(yuzu PRIVATE common core input_common video_core)
-target_link_libraries(yuzu PRIVATE Boost::boost glad Qt5::Widgets)
+target_link_libraries(yuzu PRIVATE Boost::boost glad Qt::Widgets)
 target_link_libraries(yuzu PRIVATE ${PLATFORM_LIBRARIES} Threads::Threads)
 
 target_include_directories(yuzu PRIVATE ../../externals/Vulkan-Headers/include)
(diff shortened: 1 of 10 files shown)


src/yuzu/CMakeLists.txt
```diff
@@ -236,7 +236,13 @@ if (APPLE)
     set_target_properties(yuzu PROPERTIES MACOSX_BUNDLE_INFO_PLIST ${CMAKE_CURRENT_SOURCE_DIR}/Info.plist)
 elseif(WIN32)
     # compile as a win32 gui application instead of a console application
-    target_link_libraries(yuzu PRIVATE Qt5::WinMain)
+    if (QT_VERSION VERSION_GREATER 6)
+        # If CXX_STANDARD is not set, MSVC is trying to compile with C++17
```
lioncash, line 240: We compile with [/std:c++latest](https://github.com/yuzu-emu/yuzu/blob/master/src/CMakeLists.txt#L44) for MSVC, so it already uses the latest available standard

Docteh: From what I can tell Qt6 appears to be "helpfully" setting it to /std:c++17 

image

Docteh:
Just noting down other items that would need to be tackled. I think the `rng_seed` is worth double checking, otherwise its just notes, and compilers are good at showing errors anyways.

The following section might not be collapsed in email. 
<details>
<summary>notes on other items</summary>

- [ ] UILayout items in qt-config.ini are not backwards compatible. 
This causes "Name" "Compatibility" "Add-ons" "File type" "Size" headers to not show after using a Qt6 build
May or may not be a problem by the time yuzu moves to Qt6

- [ ] CMakeLists.txt Something like the following might be useful to stop Qt6 from forcing /std:c++17
In Qt\6.3.0\lib\cmake\Qt6\Qt6Targets.cmake they are asking for cxx_std_17,
`set_target_properties(Qt6::Platform PROPERTIES INTERFACE_COMPILE_FEATURES "cxx_std_23")`

- [ ] src/yuzu/configuration/configure_system.cpp  toULongLong is 64bits, and toULong is 32bits, 
what is rng_seed supposed to be?

- [ ] src/yuzu/loading_screen.h 
 ¯\\\_(ツ)_/¯
//Qt6 thinks this declaration is unnecessary: explicit specialization 'QMetaTypeId<T>' has already been instantiated/defined (two separate errors)
//Qt5 needs this declared.
Q_DECLARE_METATYPE(VideoCore::LoadCallbackStage);
something to do with moc

- [ ] src/yuzu/main.cpp

QDesktopWidget obsolecense 
Qt5 docs say to stop using it https://doc.qt.io/qt-5/qdesktopwidget.html
This needs to be fixed/tried/testing on a multi monitor PC

```diff
 void GMainWindow::SetDefaultUIGeometry() {
     // geometry: 53% of the window contents are in the upper screen half, 47% in the lower half
-    const QRect screenRect = QApplication::desktop()->screenGeometry(this);
+    //Qt6 yoloing
+    const QRect screenRect = QGuiApplication::primaryScreen()->geometry();
```
Also another call in
`void GMainWindow::ShowFullscreen() {`
</details>

src/yuzu/main.cpp
```diff
@@ -1581,7 +1581,7 @@ void GMainWindow::StoreRecentFile(const QString& filename) {
 
 void GMainWindow::UpdateRecentFiles() {
     const int num_recent_files =
-        std::min(UISettings::values.recent_files.size(), max_recent_files_item);
+        std::min(int(UISettings::values.recent_files.count()), max_recent_files_item);
```
Morph1984, line 1584:
```suggestion
        std::min(static_cast<int>(UISettings::values.recent_files.size()), max_recent_files_item);
```

src/yuzu/main.cpp
```diff
@@ -3319,7 +3319,7 @@ void GMainWindow::CenterMouseCursor() {
     const int center_x = render_window->width() / 2;
     const int center_y = render_window->height() / 2;
 
-    QCursor::setPos(mapToGlobal({center_x, center_y}));
+    QCursor::setPos(mapToGlobal(QPoint(center_x, center_y)));
```
Morph1984, line 3322:
```suggestion
    QCursor::setPos(mapToGlobal(QPoint{center_x, center_y}));
```

src/yuzu/configuration/configure_system.cpp
```diff
@@ -130,8 +130,7 @@ void ConfigureSystem::ApplyConfiguration() {
         // Guard if during game and set to game-specific value
         if (Settings::values.rng_seed.UsingGlobal()) {
             if (ui->rng_seed_checkbox->isChecked()) {
-                Settings::values.rng_seed.SetValue(
-                    ui->rng_seed_edit->text().toULongLong(nullptr, 16));
+                Settings::values.rng_seed.SetValue(ui->rng_seed_edit->text().toULong(nullptr, 16));
```
Morph1984, line 133:
```suggestion
                Settings::values.rng_seed.SetValue(ui->rng_seed_edit->text().toUInt(nullptr, 16));
```

Docteh:
Build system appears to be complaining about not being able to downlead fmt/8.1.1, but https://conan.io/center/fmt seems there. 

Since the latest "Update Translations" into transifex failed also with a conan error I'm going to just leave it for now.

lioncash (approved):

PR #8325, opened 2022-05-09
Update translations (2022-05-09)

zhaobot:
This PR is generated by [zhaowenlan1779/txbot](https://github.com/zhaowenlan1779/txbot).


zhaowenlan1779:
OK this seems to be working. So right now translations falling below the threshold won't get deleted but won't get updated as well. I think I'll implement a better mechanism...

lioncash (approved):

PR #8332, opened 2022-05-13
general: Use smaller array types where applicable

Morph1984:
Using smaller array types allows for a slightly more compact executable.
This PR leads to a total reduction of 36352 bytes in the compiled executable on MSVC.

Note: There are other arrays that use larger-than-necessary types, but they do not reduce the size of the compiled executable, hence why the changes are limited to these 4 files.
Note2: The usage of u8 in BITS_PER_BLOCK_TABLE assumes that 256 bit (32 byte) formats (such as R64G64B64A64) do not exist on the Tegra X1 due to their inefficiency and impracticality.

### DIFF
--- a/src/audio_core/command_generator.cpp
+++ b/src/audio_core/command_generator.cpp
@@ -129,17 +129,17 @@ s32 ToS32(float sample) {
     return static_cast<s32>(rescaled_sample);
 }
 
-constexpr std::array<std::size_t, 20> REVERB_TAP_INDEX_1CH{0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-                                                           0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+constexpr std::array<u8, 20> REVERB_TAP_INDEX_1CH{0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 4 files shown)


Docteh:
Lowering the compression threshold on the compatibility_list would save space on the executable.

From what I can tell it just doesn't quite compress well enough for the default threshold (70)

```perl
$ /d/github/yuzu/buildmsvc/externals/qt-5.15.2-msvc2019_64/bin/rcc.exe --threshold 69 --verbose --compress 9 --compress-algo=zlib -g cpp  ./dist/compatibility_list/compatibility_list.qrc -o lol.lol
Qt resource compiler
Processing 1 files [listMode=0]
Interpreting ./dist/compatibility_list/compatibility_list.qrc
Outputting code
compatibility_list.json: note: compressed using zlib (1255189 -> 386964)
```

The 5.12.8 version of the tool is even less useful when askingit to be verbose

$ rcc --verbose --threshold 69 --compress 9 ./compatibility_list.qrc -o lol
Qt resource compiler
Processing 1 files [listMode=0]
Interpreting ./compatibility_list.qrc
Outputting code

bunnei (approved):

PR #8312, opened 2022-05-06
yuzu: fix icon name

abouvier:
The icon was renamed in https://github.com/yuzu-emu/yuzu/pull/8283.

PR #8283, opened 2022-04-28
dist: add AppStream metadata

Tachi107:
The AppStream file is mostly copied from the one already used by the [Flatpak build](https://github.com/flathub/org.yuzu_emu.yuzu/blob/62fc225acfa0178ddf61be68ad9be60841077263/org.yuzu_emu.yuzu.metainfo.xml).

As it already defines the application id as `org.yuzu_emu.yuzu` I renamed the `yuzu.desktop` and `yuzu.xml` files so that they match.

I've also made some minor tweaks to it, like fixing the capitalization of "yuzu", adding a few keys and sorting them as presented in the documentation.

Lastly, I added `PrefersNonDefaultGPU=true` to the .desktop file so that yuzu is launched with the dedicated graphics card on Linux.

I'm not an expert in freedesktop.org stuff, so I'm pinging some yuzu contributors that probably know something more than I do: @jroweboy @Tatsh @liushuyu @abouvier

PR #8339, opened 2022-05-15
about dialog: Fix yuzu icon

Docteh:
about dialog: Fix the logo in a multiplatform way

The Icon was renamed in #8283 for Linux builds, and the fix proposed in #8312 would in turn break
the icon for Windows users.

I've decided to fix the aboutdialog.ui file via qtcreator.

I'm not sure its important to have the yuzu icon inside the About dialog grabbed from the local Qt theme,
but I've reword how the code works for that, and we can just delete those lines.

I've also thrown the yuzu.png through pngcrush to remove this warning
libpng warning: iCCP: known incorrect sRGB profile

Credit to abouvier for bringing bug up.


bunnei (approved):

bunnei:
Thanks!

PR #8342, opened 2022-05-15
general: Use Common::U16StringFromBuffer in place of QString::toStdU16String

lat9nq:
This PR aims to fix linking errors when compiling yuzu against Qt 5.15.2 and libstdc++ 12 or later with Clang 13 or 14. Linking something that uses `QString::toStdU16String` throws missing symbol errors. Without investigating it, my guess is that it requires a function that was dropped or changed from previous libstdc++ versions. 

This adds and uses `Common::U16StringFromBuffer` in string_util, which effectively works the same as the Qt function but resolves linker issues.

### DIFF
--- a/src/yuzu/applets/qt_software_keyboard.cpp
+++ b/src/yuzu/applets/qt_software_keyboard.cpp
@@ -411,11 +411,11 @@ void QtSoftwareKeyboardDialog::ShowTextCheckDialog(
             break;
         }
 
-        const auto& text = ui->topOSK->currentIndex() == 1 ? ui->text_edit_osk->toPlainText()
-                                                           : ui->line_edit_osk->text();
-        std::u16string text_s = Common::U16StringFromBuffer(text.utf16(), text.size());
+        const auto text = ui->topOSK->currentIndex() == 1 ? ui->text_edit_osk->toPlainText()
+                                                          : ui->line_edit_osk->text();
+        auto text_str = Common::U16StringFromBuffer(text.utf16(), text.size());
 
-        emit SubmitNormalText(SwkbdResult::Ok, std::move(text_s), true);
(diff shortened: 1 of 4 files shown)


src/yuzu/applets/qt_software_keyboard.cpp
```diff
@@ -1282,11 +1283,11 @@ void QtSoftwareKeyboardDialog::TranslateButtonPress(Core::HID::NpadButton button
         if (is_inline) {
             emit SubmitInlineText(SwkbdReplyType::DecidedCancel, current_text, cursor_position);
         } else {
-            auto text = ui->topOSK->currentIndex() == 1
-                            ? ui->text_edit_osk->toPlainText().toStdU16String()
-                            : ui->line_edit_osk->text().toStdU16String();
+            const auto& text = ui->topOSK->currentIndex() == 1 ? ui->text_edit_osk->toPlainText()
```
lioncash, line 1286:
```suggestion
            const auto text = ui->topOSK->currentIndex() == 1 ? ui->text_edit_osk->toPlainText()
```

Both of these functions return by value, not by reference

src/yuzu/applets/qt_software_keyboard.cpp
```diff
@@ -1189,7 +1189,8 @@ void QtSoftwareKeyboardDialog::InlineKeyboardButtonClicked(QPushButton* button)
         return;
     }
 
-    InlineTextInsertString(button->text().toStdU16String());
+    InlineTextInsertString(
+        Common::U16StringFromBuffer(button->text().utf16(), button->text().size()));
```
lioncash, line 1193: Maybe extract `button->text()` to its own variable, otherwise this is constructing the QString twice

lat9nq: Thanks for the help!

liamwhite (approved):

PR #8348, opened 2022-05-17
Add control to metainfo

JakobDev:
For reference see the [specification](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-control)


Docteh:
> Add control to metainfo

I think you could be more specific about what and where.

Maybe something along these lines (These are ~~three~~ four separate ideas, and they might not be quite there):
Add supported controls to freedesktop (flatpak) metainfo file.

Add supported control types to freedesktop meta data

Add supported control types to Linux meta data

Add supported control types to flatpak meta data.

bunnei (approved):

goldenx86:
Hi, can I mention this pull request in the next progress report?

PR #8351, opened 2022-05-17
video_core: Support new VkResult

abouvier:
Added in [Vulkan-Headers v1.3.213](https://github.com/KhronosGroup/Vulkan-Headers/blob/v1.3.213/include/vulkan/vulkan_core.h#L178).


lioncash (approved):
LGTM, but in the future it's preferable to have the submodule update in its own commit to make identifying external updates much nicer (and keeps them separate from source changes)

PR #8353, opened 2022-05-18
log the MSVC runtime version when running on MSVC build

Docteh:
This might be useful information, not 100% sure.

[   0.954129] Frontend <Info> yuzu\main.cpp:GMainWindow:245: yuzu Version: yuzu Development Build | master-0b9ef3c0b-dirty
[   0.954150] Frontend <Info> yuzu\main.cpp:GMainWindow:262: MSVC Compiler: 1931 Runtime: 14.32.31326.0

--------

The Compiler number is the _MSC_VER DEFINE, and its reading `vcruntime140.dll` for the Runtime version


src/yuzu/main.cpp
```diff
@@ -243,6 +243,33 @@ GMainWindow::GMainWindow()
     const auto yuzu_build_version = override_build.empty() ? yuzu_build : override_build;
 
     LOG_INFO(Frontend, "yuzu Version: {}", yuzu_build_version);
+
+#ifdef _MSC_VER
```
lioncash, line 247: You may want to move this block into a function like `LogRuntime()` and call it here. This can then be used for other runtimes if they become useful to log in the future.

src/yuzu/main.cpp
```diff
@@ -243,6 +243,33 @@ GMainWindow::GMainWindow()
...
+    // It is possible that the name of the dll will change.
+    // vcruntime140.dll is for 2015 and onwards
+    const char* runtime_dll_name = "vcruntime140.dll";
```
lioncash, line 250:
```suggestion
    constexpr char runtime_dll_name[] = "vcruntime140.dll";
```

src/yuzu/main.cpp
```diff
@@ -243,6 +243,33 @@ GMainWindow::GMainWindow()
...
+    DWORD dwHandle, sz = GetFileVersionInfoSizeA(runtime_dll_name, &dwHandle);
+    bool runtime_version_inspection_worked = FALSE;
```
lioncash, line 252:
```suggestion
    bool runtime_version_inspection_worked = false;
```

src/yuzu/main.cpp
```diff
@@ -243,6 +243,33 @@ GMainWindow::GMainWindow()
     const auto yuzu_build_version = override_build.empty() ? yuzu_build : override_build;
 
     LOG_INFO(Frontend, "yuzu Version: {}", yuzu_build_version);
+
+#ifdef _MSC_VER
+    // It is possible that the name of the dll will change.
+    // vcruntime140.dll is for 2015 and onwards
+    const char* runtime_dll_name = "vcruntime140.dll";
+    DWORD dwHandle, sz = GetFileVersionInfoSizeA(runtime_dll_name, &dwHandle);
```
lioncash, line 251: Prefer placing variables on their own individual line, also `dw_handle`

src/yuzu/main.cpp
```diff
@@ -243,6 +243,33 @@ GMainWindow::GMainWindow()
...
+    bool runtime_version_inspection_worked = FALSE;
+    if (sz > 0) {
+        std::vector<unsigned char> buf(sz);
+        if (GetFileVersionInfoA(runtime_dll_name, dwHandle, sz, &buf[0])) {
```
lioncash, line 255:
```suggestion
        if (GetFileVersionInfoA(runtime_dll_name, dwHandle, sz, buf.data())) {
```

src/yuzu/main.cpp
```diff
@@ -243,6 +243,33 @@ GMainWindow::GMainWindow()
...
+            VS_FIXEDFILEINFO* pvi;
+            sz = sizeof(VS_FIXEDFILEINFO);
+            if (VerQueryValueA(&buf[0], "\\", (LPVOID*)&pvi, (unsigned int*)&sz)) {
+                if (pvi->dwSignature == VS_FFI_SIGNATURE) {
+                    runtime_version_inspection_worked = TRUE;
```
lioncash, line 260:
```suggestion
                    runtime_version_inspection_worked = true;
```

src/yuzu/main.cpp
```diff
@@ -243,6 +243,33 @@ GMainWindow::GMainWindow()
...
+    // vcruntime140.dll is for 2015 and onwards
+    const char* runtime_dll_name = "vcruntime140.dll";
+    DWORD dwHandle, sz = GetFileVersionInfoSizeA(runtime_dll_name, &dwHandle);
+    bool runtime_version_inspection_worked = FALSE;
+    if (sz > 0) {
+        std::vector<unsigned char> buf(sz);
+        if (GetFileVersionInfoA(runtime_dll_name, dwHandle, sz, &buf[0])) {
+            VS_FIXEDFILEINFO* pvi;
+            sz = sizeof(VS_FIXEDFILEINFO);
+            if (VerQueryValueA(&buf[0], "\\", (LPVOID*)&pvi, (unsigned int*)&sz)) {
```
lioncash, line 258:
```suggestion
            if (VerQueryValueA(buf.data(), "\\", &pvi, &sz)) {
```

If you make `sz` a `UINT` type, you don't need to cast

src/yuzu/main.cpp
```diff
@@ -198,6 +198,36 @@ static void RemoveCachedContents() {
...
+    // winver.h calls this dwHandle, but the parameter isn't used as a handle of any kind.
+    DWORD dw_handle;
+    uint sz = GetFileVersionInfoSizeA(runtime_dll_name, &dw_handle);
+    bool runtime_version_inspection_worked = false;
+    if (sz > 0) {
+        std::vector<unsigned char> buf(sz);
+        if (GetFileVersionInfoA(runtime_dll_name, dw_handle, sz, buf.data())) {
+            VS_FIXEDFILEINFO* pvi;
+            sz = sizeof(VS_FIXEDFILEINFO);
+            if (VerQueryValueA(buf.data(), "\\", (LPVOID*)&pvi, &sz)) {
```
lioncash, line 215: If the LPVOID cast is required, this should be using either `static_cast` or `reinterpret_cast`. Otherwise it should be omitted

src/yuzu/main.cpp
```diff
@@ -198,6 +198,36 @@ static void RemoveCachedContents() {
...
 }
 
+static void LogRuntimes() {
+#ifdef _MSC_VER
+    // It is possible that the name of the dll will change.
+    // vcruntime140.dll is for 2015 and onwards
+    constexpr char runtime_dll_name[] = "vcruntime140.dll";
+    // winver.h calls this dwHandle, but the parameter isn't used as a handle of any kind.
+    DWORD dw_handle;
+    uint sz = GetFileVersionInfoSizeA(runtime_dll_name, &dw_handle);
```
lioncash, line 208:
```suggestion
    UINT sz = GetFileVersionInfoSizeA(runtime_dll_name, &dw_handle);
```

is the proper Windows type alias

Docteh:
@lioncash ~~Fixes implemented~~ (I see some more just now)

The cast to `LPVOID*` is still required, I tried `void*` to make it less Windowsey and the compiler threw an error. 
minwindef.h:173:`typedef void far            *LPVOID;`

static_cast, gotcha.

I'm not sure if this is from renaming dwHandle, I've noticed that this helpful green underline stuff takes awhile to show up.
image

I can pass GetFileVersionInfoA a zero if we're concerned. Kind of amusing because the [API docs for the function](https://docs.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfoa) state that the argument is ignored...


I've named the function `LogRuntimes` so that it lines up with `GMainWindow`.
```
[   0.958068] Frontend <Info> yuzu\main.cpp:GMainWindow:275: yuzu Version: yuzu Development Build | master-0b9ef3c0b-dirty
[   0.958095] Frontend <Info> yuzu\main.cpp:LogRuntimes:220: MSVC Compiler: 1931 Runtime: 14.32.31326.0
```

src/yuzu/main.cpp
```diff
@@ -198,6 +198,36 @@ static void RemoveCachedContents() {
     Common::FS::RemoveDirRecursively(offline_system_data);
 }
 
+static void LogRuntimes() {
+#ifdef _MSC_VER
+    // It is possible that the name of the dll will change.
+    // vcruntime140.dll is for 2015 and onwards
+    constexpr char runtime_dll_name[] = "vcruntime140.dll";
+    // winver.h calls this dwHandle, but the parameter isn't used as a handle of any kind.
+    DWORD dwHandle; // Intellisense wants this variable name.
```
Docteh, line 207: @lioncash Intellisense is unhappy unless this variable is dwHandle, can we leave it as dwHandle?

lioncash: I think that is just your intellisense cache being "sticky"

As far as I'm aware, API-wise the variable can be removed entirely, since it's marked as an optional parameter on the documentation page for `GetFileVersionInfoSizeA`, so you can pass nullptr into it directly, and pass 0 into `GetFileVersionInfoA`. since that parameter is documented as unused.

Docteh:
Intellisense did eventually complain about dwHandle but it took a really really long time compared to the other variable names, not sure why. I only noticed just as I changed it to a 0.

@lioncash `nullptr` worked out well, implemented.

src/yuzu/main.cpp
```diff
@@ -198,6 +198,34 @@ static void RemoveCachedContents() {
...
 
+static void LogRuntimes() {
+#ifdef _MSC_VER
+    // It is possible that the name of the dll will change.
+    // vcruntime140.dll is for 2015 and onwards
+    constexpr char runtime_dll_name[] = "vcruntime140.dll";
+    UINT sz = GetFileVersionInfoSizeA(runtime_dll_name, nullptr);
+    bool runtime_version_inspection_worked = false;
+    if (sz > 0) {
+        std::vector<unsigned char> buf(sz);
```
Morph1984, line 209:
```suggestion
        std::vector<u8> buf(sz);
```

lioncash (approved):

lioncash:
Needs a rebase

Docteh:
@lioncash successfully rebased

PR #8359, opened 2022-05-19
shader_recompiler: Implement LowerInt16ToInt32

lat9nq:
AMD drivers 22.3.2 and later expose a bug in yuzu, where the application would submit 16-bit integer instructions to GPUs that don't support 16-bit integers, namely GCN 4 devices.

Replace any 16-bit instructions with 32-bit ones so newer AMD drivers will work with `VK_KHR_workgroup_memory_explicit_layout`.

@FernandoS27 I would appreciate if you looked over the transformations I made here.

### DIFF
--- a/src/shader_recompiler/CMakeLists.txt
+++ b/src/shader_recompiler/CMakeLists.txt
@@ -219,6 +219,7 @@ add_library(shader_recompiler STATIC
     ir_opt/global_memory_to_storage_buffer_pass.cpp
     ir_opt/identity_removal_pass.cpp
     ir_opt/lower_fp16_to_fp32.cpp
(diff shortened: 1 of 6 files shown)


Moonlacer:
Did the bug cause any noticeable issues in games? Like performance loss, stuttering, etc.?

lat9nq:
> Did the bug cause any noticeable issues in games? Like performance loss, stuttering, etc.?

The application would fail an assertion in sirit if the shader used both U16 and VK_KHR_workgroup_memory_explicit_layout, thus would abort execution. For example, The Legend of Zelda: Breath of the Wild would cause yuzu to abort immediately upon booting.

PR #8363, opened 2022-05-20
vk_texture_cache: Fix the behavior of framebuffer layers

asLody:
Some games bind render targets of a different size and expect it to work, so we'll compute the minimum size.


liamwhite:
@asLody which games does this fix?

FernandoS27 (approved):

gidoly:
it regress shin megami tensei 5

yuzu_LR0NTzzYWy

Morph1984 (changes requested):
(to ensure this does not get accidentally merged)

bunnei:
@asLody any plans to investigate the above? Thank you!

PR #8374, opened 2022-05-23
input_common: Make vibration request async

german77:
Some controllers are too slow to respond when a vibration request was made. This caused the whole emulated thread to be delayed causing some heavy stutters.

This PR pushes SDL vibrations to a queue where they can be executed on a different thread. This avoids locking the main emulation thread

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -884,18 +884,42 @@ bool EmulatedController::SetVibration(std::size_t device_index, VibrationValue v
 }
 
 bool EmulatedController::TestVibration(std::size_t device_index) {
-    static constexpr VibrationValue test_vibration = {
+    if (device_index >= output_devices.size()) {
+        return false;
+    }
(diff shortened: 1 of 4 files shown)


src/core/hid/emulated_controller.cpp
```diff
@@ -851,7 +851,8 @@ void EmulatedController::SetBattery(const Common::Input::CallbackStatus& callbac
     TriggerOnChange(ControllerTriggerType::Battery, true);
 }
 
-bool EmulatedController::SetVibration(std::size_t device_index, VibrationValue vibration) {
+bool EmulatedController::SetVibration(std::size_t device_index, VibrationValue vibration,
+                                      bool is_test) {
```
lioncash, line 855: I would avoid using a bool here and instead use an enum

```cpp
SetVibration(index, value, true);
```

doesn't actually communicate anything meaningful to the reader

bunnei (approved):

PR #8379, opened 2022-05-25
vulkan_device: Block AMDVLK's VK_KHR_push_descriptor

lat9nq:
Recent AMD Vulkan drivers (22.5.2, or 2.0.226 for specifically Vulkan) have a broken initial implementation of VK_KHR_push_descriptor that causes a crash in yuzu. Disable it for the time being.

Specifically RX 6400 and RX 6500XT devices will still crash to an unknown reason. All other cards appear to work with this as far as it's been tested.


goldenx86:
Can confirm that a 6500 XT works, no idea what the first round of debugging triggered.

bunnei (approved):

PR #8381, opened 2022-05-26
path_util: Resolve `-Wpointer-bool-conversion` warning

lat9nq:
Clang (rightfully) warns that we are checking for the existence of a pointer to something just allocated on the stack, which is always true.

Instead, check that the string is not empty.

### DIFF
--- a/src/common/fs/path_util.cpp
+++ b/src/common/fs/path_util.cpp
@@ -232,9 +232,7 @@ void SetYuzuPath(YuzuPath yuzu_path, const fs::path& new_path) {
 fs::path GetExeDirectory() {
     wchar_t exe_path[MAX_PATH];
 
-    GetModuleFileNameW(nullptr, exe_path, MAX_PATH);
-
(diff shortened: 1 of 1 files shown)


src/common/fs/path_util.cpp
```diff
@@ -234,7 +234,7 @@ fs::path GetExeDirectory() {
 
     GetModuleFileNameW(nullptr, exe_path, MAX_PATH);
 
-    if (!exe_path) {
+    if (exe_path[0] == '\0') {
```
lioncash, line 237: I think it would be sufficient to merge the above function call into the if statement like:

```cpp
if (GetModuleFileNameW(nullptr, exe_path, MAX_PATH) == 0)
```

since this function is documented as returning zero if the function fails

lat9nq: Thanks! Yeah this is much better.

lioncash (approved):

PR #8383, opened 2022-05-27
yuzu: Make variable shadowing a compile-time error

Morph1984:
Now that the entire project is free of variable shadowing, we can enforce this as a compile time error to prevent any further introduction of this logic bug.

Some things to keep in mind:
The following are treated as shadowing in GCC/Clang, but not MSVC:
```cpp
int a;
auto f = [] {
    int a; // shadows a above, even though its not in the capture list
};
```
```cpp
struct Impl {
    Impl(int a) : a{a} {} // MSVC should emit C4458 here, but doesn't for whatever reason

    void Set(int a) { // This does, though
        this->a = a;
    }

    int a;
};
```
Addendum: In the case above, GCC/Clang are inconsistent about emitting warnings regarding variable shadowing in constructors.

### DIFF
--- a/externals/microprofile/microprofile.h
+++ b/externals/microprofile/microprofile.h
@@ -1246,7 +1246,7 @@ struct MicroProfileScopeLock
 {
     bool bUseLock;
     std::recursive_mutex& m;
-    MicroProfileScopeLock(std::recursive_mutex& m) : bUseLock(g_bUseLock), m(m)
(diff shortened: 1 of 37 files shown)


Morph1984:
Pending an update to cpp-httplib to resolve their variable shadowing errors

lioncash:
Still uncovers some shadowing, but the builds almost complete

lioncash:
Ping on the status of this?

lioncash (approved):

PR #8393, opened 2022-05-30
general: Set renderer_backend's default to Vulkan

lat9nq:
I'm going to skip the premise as to why we want to default to Vulkan. It's more clear to other people, and I didn't get any sleep last night. The people to ask about that would be @goldenx86, or @bunnei who tasked me with this.

The reason this hasn't been done before though is due to crashes when opening the Vulkan loader. From my research today this seems more to do with broken Vulkan layers and not having anything to do with outdated drivers. (It may be possible a __broken driver__ will also reproduce crashes, but I don't really want to try and implement my own Vulkan driver even if it just stubs the necessary functions before forcing a segfault.) 

To make `renderer_backend`'s default to Vulkan, a few things had to be done to make that palatable:

- When yuzu boots, it just opens the Vulkan library.
  - If it works, all good and we continue with Vulkan as the default.
  - If yuzu crashes, a new file in the config directory will be left behind (this is deleted normally).
- If it's found that loading Vulkan crashes the application on next boot, `has_broken_vulkan` is set to true.
  - The first time this happens, a warning is displayed to notify the user.
  - This forces use of OpenGL, and Vulkan cannot be selected.
  - The Shader Backend selector is made accessible for use in custom configurations.
  - To disable `has_broken_vulkan`, the user needs to press a button in Graphics Configuration to manually run the Vulkan device enumeration.

I think I would prefer if we could handle the crash when loading the Vulkan library as opposed to aborting, but I don't know how we can do that if it's possible. I guess it would be something along the lines of catching a SIGSEGV? And I don't have a test case of this for Windows at the moment.

The most notable user-facing change is that new yuzu installations will use Vulkan by default. The few that have broken Vulkan layers installed will be locked out of Vulkan until they fix it themselves.

### DIFF
--- a/src/yuzu/check_vulkan.cpp
+++ b/src/yuzu/check_vulkan.cpp
@@ -1,0 +1,51 @@
+#include "video_core/vulkan_common/vulkan_wrapper.h"
+
+#include <exception>
+#include <filesystem>
+#include <fstream>
+#include "common/fs/fs.h"
+#include "common/fs/path_util.h"
+#include "common/logging/log.h"
+#include "video_core/vulkan_common/vulkan_instance.h"
+#include "video_core/vulkan_common/vulkan_library.h"
+#include "yuzu/check_vulkan.h"
+#include "yuzu/uisettings.h"
+
+constexpr char TEMP_FILE_NAME[] = "vulkan_check";
+
+bool CheckVulkan() {
+    if (UISettings::values.has_broken_vulkan) {
+        return true;
+    }
+
+    LOG_DEBUG(Frontend, "Checking presence of Vulkan");
+
+    const auto fs_config_loc = Common::FS::GetYuzuPath(Common::FS::YuzuPath::ConfigDir);
+    const auto temp_file_loc = fs_config_loc / TEMP_FILE_NAME;
+
+    if (std::filesystem::exists(temp_file_loc)) {
+        LOG_WARNING(Frontend, "Detected recovery from previous failed Vulkan initialization");
+
+        UISettings::values.has_broken_vulkan = true;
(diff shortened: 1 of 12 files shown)


src/yuzu/check_vulkan.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <fstream>
+#include "common/fs/fs.h"
+#include "common/fs/path_util.h"
+#include "common/logging/log.h"
+#include "video_core/vulkan_common/vulkan_instance.h"
+#include "video_core/vulkan_common/vulkan_library.h"
+#include "yuzu/check_vulkan.h"
+#include "yuzu/uisettings.h"
+
+const char* TEMP_FILE_NAME = "vulkan_check";
```
lioncash, line 16:
```suggestion
constexpr char TEMP_FILE_NAME[] = "vulkan_check";
```

src/yuzu/configuration/configure_graphics.cpp
```diff
@@ -57,6 +59,23 @@ ConfigureGraphics::ConfigureGraphics(const Core::System& system_, QWidget* paren
         UpdateBackgroundColorButton(new_bg_color);
     });
 
+    connect(ui->button_check_vulkan, &QAbstractButton::clicked, this, [this]() {
```
lioncash, line 62:
```suggestion
    connect(ui->button_check_vulkan, &QAbstractButton::clicked, this, [this] {
```

lioncash (commented):
Just some minor nits

bunnei (approved):

UndeadlyEye:
a video would be amazing

PR #8400, opened 2022-06-01
fix UI opening fullscreen after certain crashes

Docteh:
Sometimes when yuzu crashes, it restarts with the games list in fullscreen,
which would be fine, except there isn't an easy way to exit this.
It also doesn't occur often enough for qt-config.ini files to be in good supply.

UILayout\geometry value in qt-config.ini is the culprit,
 at least for the one provided.

Proposed fix is to simply check isFullScreen when yuzu is starting up,
and take it out of full screen immediately

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1015,6 +1015,10 @@ void GMainWindow::SetDefaultUIGeometry() {
 void GMainWindow::RestoreUIState() {
     setWindowFlags(windowFlags() & ~Qt::FramelessWindowHint);
     restoreGeometry(UISettings::values.geometry);
+    // Work-around because the games list isn't supposed to be full screen
+    if (isFullScreen()) {
+        showNormal();
+    }
     restoreState(UISettings::values.state);
     render_window->setWindowFlags(render_window->windowFlags() & ~Qt::FramelessWindowHint);
     render_window->restoreGeometry(UISettings::values.renderwindow_geometry);


Docteh:
I got sidetracked twice looking for the code that deserializes, but  sticking this into your qt-config.ini at 1920x1080 shows the bug. 
```
UILayout\geometry=@ByteArray(\x1\xd9\xd0\xcb\0\x3\0\0\0\0\0\a\0\0\0\0\0\0\ax\0\0\x4\x30\0\0\0\0\0\0\0\x17\0\0\a\x7f\0\0\x4\xf\0\0\0\0\x2\x4\0\0\a\x80\0\0\0\0\0\0\0\0\0\0\a\x7f\0\0\x4\x37)
```

german77 (approved):
Seems to work as intended.

src/yuzu/main.cpp
```diff
@@ -1015,6 +1015,11 @@ void GMainWindow::SetDefaultUIGeometry() {
 void GMainWindow::RestoreUIState() {
     setWindowFlags(windowFlags() & ~Qt::FramelessWindowHint);
     restoreGeometry(UISettings::values.geometry);
+    // Work-around because the games list isn't supposed to be full screen
+    if (isFullScreen()) {
+        LOG_INFO(Frontend, "starting up in fullscreen, previous crash?");
```
Morph1984, line 1020: The log can be omitted tbh

Docteh: it's now omitted

bunnei (approved):

PR #8403, opened 2022-06-01
gdbstub: Explicitly cast return type to u8

Morph1984:
Otherwise, the addition promotes the returned value to an int instead of keeping it as a u8.


vonchenplus:
This may result in overflow of the added value

Morph1984:
If in the case of overflow, it would've anyways since the return type is meant to be a u8

liamwhite (approved):

lioncash (approved):

PR #8405, opened 2022-06-01
ui: Status bars dock button becomes DOCKED/HANDHELD button

Docteh:
For people not used to the Yuzu UI it's not always clear if the emulated
console is docked or not.  The other items update their text when clicked,
this PR brings the DOCK button in line with this.

DOCK -> DOCK or UNDOCK

Revised to

DOCK -> DOCKED or HANDHELD

---
image
image

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -827,12 +827,11 @@ void GMainWindow::InitializeWidgets() {
 
     // Setup Dock button
     dock_status_button = new QPushButton();
-    dock_status_button->setObjectName(QStringLiteral("TogglableStatusBarButton"));
+    dock_status_button->setObjectName(QStringLiteral("DockingStatusBarButton"));
     dock_status_button->setFocusPolicy(Qt::NoFocus);
     connect(dock_status_button, &QPushButton::clicked, this, &GMainWindow::OnToggleDockedMode);
-    dock_status_button->setText(tr("DOCK"));
     dock_status_button->setCheckable(true);
-    dock_status_button->setChecked(Settings::values.use_docked_mode.GetValue());
+    UpdateDockedButton();
     statusBar()->insertPermanentWidget(0, dock_status_button);
 
     gpu_accuracy_button = new QPushButton();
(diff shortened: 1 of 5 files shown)


Kewlan:
Maybe it should say `DOCKED` and `UNDOCKED` instead, because now it seems like pressing the button will cause it to `DOCK` or `UNDOCK` which is the opposite behaviour. Greying it out might not be needed anymore either.

Otherwise, renaming the button to `DOCKED` and keeping the grey-out may be enough?

Norstyrm:
I think it'd be better to name it DOCKED and HANDHELD as it'd be more in line with Nintendo's terminology for play modes instead of UNDOCK/UNDOCKED. I also agree with @Kewlan that greying out the text when toggled would not be needed anymore as the naming would make it obvious which mode is currently selected.

Docteh:
image
image

dist/qt_themes/qdarkstyle_midnight_blue/style.qss has an unnecessary warning about how the file is created. I wonder if the style sheets should be separated into theme/yuzu specific parts

src/yuzu/main.cpp
```diff
@@ -3267,9 +3271,9 @@ void GMainWindow::UpdateAAText() {
 }
 
 void GMainWindow::UpdateStatusButtons() {
-    dock_status_button->setChecked(Settings::values.use_docked_mode.GetValue());
     renderer_status_button->setChecked(Settings::values.renderer_backend.GetValue() ==
                                        Settings::RendererBackend::Vulkan);
+    UpdateDockedButton();
```
Morph1984, line 3276: minor nit, this should be below UpdateGPUAccuracyButton

src/yuzu/main.cpp
```diff
@@ -3250,6 +3249,11 @@ void GMainWindow::UpdateFilterText() {
         break;
     }
 }
+void GMainWindow::UpdateDockedButton() {
```
Morph1984, line 3252: this should go above UpdateFilterText

src/yuzu/main.h
```diff
@@ -322,6 +322,7 @@ private slots:
                            std::string_view gpu_vendor = {});
     void UpdateFilterText();
     void UpdateAAText();
+    void UpdateDockedButton();
```
Morph1984, line 325: this should go above UpdateFilterText

lioncash (approved):

PR #8413, opened 2022-06-02
gpu_thread: Move to bounded queue

behunin:
The main advantage is no new in the hotpath thanks to the slots being allocated at initialization.

Just 1-2 more fps, but much better recovery from load spikes and such.

from: https://github.com/rigtorp/MPMCQueue

### DIFF
--- a/src/common/bounded_threadsafe_queue.h
+++ b/src/common/bounded_threadsafe_queue.h
@@ -1,0 +1,180 @@
+// SPDX-FileCopyrightText: Copyright (c) 2020 Erik Rigtorp <erik@rigtorp.se>
+// SPDX-License-Identifier: MIT
+#pragma once
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4324)
+#endif
+
+#include <atomic>
+#include <bit>
+#include <condition_variable>
+#include <memory>
+#include <mutex>
+#include <new>
+#include <stdexcept>
+#include <stop_token>
+#include <type_traits>
+#include <utility>
+
+namespace Common {
+namespace mpsc {
+#if defined(__cpp_lib_hardware_interference_size)
+constexpr size_t hardware_interference_size = std::hardware_destructive_interference_size;
+#else
+constexpr size_t hardware_interference_size = 64;
+#endif
+
+template <typename T>
+using AlignedAllocator = std::allocator<T>;
+
+template <typename T>
+struct Slot {
+    ~Slot() noexcept {
+        if (turn.test()) {
+            destroy();
+        }
+    }
+
+    template <typename... Args>
+    void construct(Args&&... args) noexcept {
+        static_assert(std::is_nothrow_constructible_v<T, Args&&...>,
+                      "T must be nothrow constructible with Args&&...");
+        std::construct_at(reinterpret_cast<T*>(&storage), std::forward<Args>(args)...);
+    }
+
+    void destroy() noexcept {
+        static_assert(std::is_nothrow_destructible_v<T>, "T must be nothrow destructible");
+        std::destroy_at(reinterpret_cast<T*>(&storage));
+    }
+
+    T&& move() noexcept {
+        return reinterpret_cast<T&&>(storage);
+    }
+
+    // Align to avoid false sharing between adjacent slots
+    alignas(hardware_interference_size) std::atomic_flag turn{};
+    struct aligned_store {
+        struct type {
+            alignas(T) unsigned char data[sizeof(T)];
+        };
+    };
+    typename aligned_store::type storage;
+};
+
+template <typename T, typename Allocator = AlignedAllocator<Slot<T>>>
+class Queue {
+public:
+    explicit Queue(const size_t capacity, const Allocator& allocator = Allocator())
+        : allocator_(allocator) {
+        if (capacity < 1) {
+            throw std::invalid_argument("capacity < 1");
+        }
+        // Ensure that the queue length is an integer power of 2
+        // This is so that idx(i) can be a simple i & mask_ insted of i % capacity
+        // https://github.com/rigtorp/MPMCQueue/pull/36
+        if (!std::has_single_bit(capacity)) {
+            throw std::invalid_argument("capacity must be an integer power of 2");
+        }
+
+        mask_ = capacity - 1;
+
+        // Allocate one extra slot to prevent false sharing on the last slot
+        slots_ = allocator_.allocate(mask_ + 2);
+        // Allocators are not required to honor alignment for over-aligned types
+        // (see http://eel.is/c++draft/allocator.requirements#10) so we verify
+        // alignment here
+        if (reinterpret_cast<uintptr_t>(slots_) % alignof(Slot<T>) != 0) {
+            allocator_.deallocate(slots_, mask_ + 2);
+            throw std::bad_alloc();
+        }
+        for (size_t i = 0; i < mask_ + 1; ++i) {
+            std::construct_at(&slots_[i]);
+        }
+        static_assert(alignof(Slot<T>) == hardware_interference_size,
+                      "Slot must be aligned to cache line boundary to prevent false sharing");
+        static_assert(sizeof(Slot<T>) % hardware_interference_size == 0,
+                      "Slot size must be a multiple of cache line size to prevent "
+                      "false sharing between adjacent slots");
+        static_assert(sizeof(Queue) % hardware_interference_size == 0,
+                      "Queue size must be a multiple of cache line size to "
+                      "prevent false sharing between adjacent queues");
+    }
+
+    ~Queue() noexcept {
+        for (size_t i = 0; i < mask_ + 1; ++i) {
+            slots_[i].~Slot();
+        }
+        allocator_.deallocate(slots_, mask_ + 2);
+    }
+
+    // non-copyable and non-movable
+    Queue(const Queue&) = delete;
+    Queue& operator=(const Queue&) = delete;
+
+    void Push(const T& v) noexcept {
+        static_assert(std::is_nothrow_copy_constructible_v<T>,
+                      "T must be nothrow copy constructible");
+        emplace(v);
+    }
+
+    template <typename P, typename = std::enable_if_t<std::is_nothrow_constructible_v<T, P&&>>>
+    void Push(P&& v) noexcept {
+        emplace(std::forward<P>(v));
+    }
+
+    void Pop(T& v, std::stop_token stop) noexcept {
+        auto const tail = tail_.fetch_add(1);
+        auto& slot = slots_[idx(tail)];
+        if (false == slot.turn.test()) {
+            std::unique_lock lock{cv_mutex};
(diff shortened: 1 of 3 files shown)


FernandoS27:
@behunin could you test if stop emulation works with thsi all the time.

Timmy5942:
> @behunin could you test if stop emulation works with thsi all the time.

On a manual compile, On shutdown / stop it freeze crash

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+#include <atomic>
+#include <memory>
+#include <new>
+#include <utility>
+
+namespace Common {
+namespace mpsc {
+static constexpr size_t hardwareInterferenceSize = 64;
+
+#define IS_POWER_OF_2(x) ((((x)-1) & (x)) == 0)
```
lioncash, line 36: This macro can be replaced with `std::has_single_bit` from the `<bit>` header.

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+#endif
+
+#include <atomic>
+#include <memory>
+#include <new>
+#include <utility>
+
+namespace Common {
+namespace mpsc {
+static constexpr size_t hardwareInterferenceSize = 64;
```
lioncash, line 34:
```suggestion
constexpr size_t hardware_interference_size = 64;
```

This constant is already provided by the standard library in the `<new>` [header](https://en.cppreference.com/w/cpp/thread/hardware_destructive_interference_size).

Given that only GCC (12.0+) and MSVC support it at the moment, you may wish to use an ifdef

behunin:
I have tested with and without this PR and here are my findings.

__Stopping__
With this pr:
>assert -- Line 161 k_auto_object

Without this pr:
>assert -- Line 161 k_auto_object

__Shutdown__
With this pr:
>assert -- Line 161 k_auto_object

Sometimes hangs

Without this pr:
>assert -- Line 161 k_auto_object


Conclusion: atomic wait is not a cool as I thought, I will refactor to mutex and condition_variable.

Side note: I didn't think a u32 could become negative so why is that assert checking for that?

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+struct Slot {
+    ~Slot() noexcept {
+        if (true == turn.test()) {
+            destroy();
+        }
+    }
+
+    template <typename... Args>
+    void construct(Args&&... args) noexcept {
+        static_assert(std::is_nothrow_constructible<T, Args&&...>::value,
```
lioncash, line 51:
```suggestion
        static_assert(std::is_nothrow_constructible_v<T, Args&&...>,
```

Note that `<type_traits>` needs to be included for this

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+
+#define IS_POWER_OF_2(x) ((((x)-1) & (x)) == 0)
+
+template <typename T>
+using AlignedAllocator = std::allocator<T>;
+
+template <typename T>
+struct Slot {
+    ~Slot() noexcept {
+        if (true == turn.test()) {
```
lioncash, line 44:
```suggestion
        if (turn.test()) {
```

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+
+    template <typename... Args>
+    void construct(Args&&... args) noexcept {
+        static_assert(std::is_nothrow_constructible<T, Args&&...>::value,
+                      "T must be nothrow constructible with Args&&...");
+        new (&storage) T(std::forward<Args>(args)...);
+    }
+
+    void destroy() noexcept {
+        static_assert(std::is_nothrow_destructible<T>::value, "T must be nothrow destructible");
```
lioncash, line 57:
```suggestion
        static_assert(std::is_nothrow_destructible_v<T>, "T must be nothrow destructible");
```

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+        if (true == turn.test()) {
+            destroy();
+        }
+    }
+
+    template <typename... Args>
+    void construct(Args&&... args) noexcept {
+        static_assert(std::is_nothrow_constructible<T, Args&&...>::value,
+                      "T must be nothrow constructible with Args&&...");
+        new (&storage) T(std::forward<Args>(args)...);
```
lioncash, line 53: This line can be replaced with `std::construct_at`

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    void destroy() noexcept {
+        static_assert(std::is_nothrow_destructible<T>::value, "T must be nothrow destructible");
+        reinterpret_cast<T*>(&storage)->~T();
```
lioncash, line 58: This can be replaced with `std::destroy_at`

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    T&& move() noexcept {
+        return reinterpret_cast<T&&>(storage);
+    }
+
+    // Align to avoid false sharing between adjacent slots
+    alignas(hardwareInterferenceSize) std::atomic_flag turn{};
+    typename std::aligned_storage<sizeof(T), alignof(T)>::type storage;
```
lioncash, line 67: aligned_storage is deprecated as of C++23. so this should probably be replaced with an `alignas()`'d buffer or something

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+};
+
+template <typename T, typename Allocator = AlignedAllocator<Slot<T>>>
+class Queue {
+public:
+    explicit Queue(const size_t capacity, const Allocator& allocator = Allocator())
+        : allocator_(allocator), head_(0), tail_(0) {
```
lioncash, line 74: `head_` and `tail_` can be directly initialized at their declaration in the class body

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+        if (capacity < 1) {
+            throw -1;
```
lioncash, line 76: These throws should probably be some meaningful exception type, rather than magic numbers

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        mask_ = capacity - 1;
+
+        // Allocate one extra slot to prevent false sharing on the last slot
+        slots_ = allocator_.allocate(mask_ + 2);
+        // Allocators are not required to honor alignment for over-aligned types
+        // (see http://eel.is/c++draft/allocator.requirements#10) so we verify
+        // alignment here
+        if (reinterpret_cast<size_t>(slots_) % alignof(Slot<T>) != 0) {
```
lioncash, line 92:
```suggestion
        if (reinterpret_cast<uintptr_t>(slots_) % alignof(Slot<T>) != 0) {
```

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+            allocator_.deallocate(slots_, mask_ + 2);
+            throw std::bad_alloc();
+        }
+        for (size_t i = 0; i < mask_ + 1; ++i) {
+            new (&slots_[i]) Slot<T>();
```
lioncash, line 97: std::construct_at

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+        }
+        allocator_.deallocate(slots_, mask_ + 2);
+    }
+
+    // non-copyable and non-movable
+    Queue(const Queue&) = delete;
+    Queue& operator=(const Queue&) = delete;
+
+    void Push(const T& v) noexcept {
+        static_assert(std::is_nothrow_copy_constructible<T>::value,
```
lioncash, line 121:
```suggestion
        static_assert(std::is_nothrow_copy_constructible_v<T>,
```

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+                      "T must be nothrow copy constructible");
+        emplace(v);
+    }
+
+    template <typename P, typename = typename std::enable_if<
```
lioncash, line 126: This can be simplified as:

```cpp
    template <typename P, typename = std::enable_if_t<
                              std::is_nothrow_constructible_v<T, P&&>>>

File: src/common/bounded_threadsafe_queue.h [hunk 15]
```diff
@@ -0,0 +1,182 @@
...
+        v = slot.move();
+        slot.destroy();
+        slot.turn.clear();
+        slot.turn.notify_one();
+    }
+
+private:
+    template <typename... Args>
+    void emplace(Args&&... args) noexcept {
+        static_assert(std::is_nothrow_constructible<T, Args&&...>::value,
```
lioncash, line 145:
```suggestion
        static_assert(std::is_nothrow_constructible_v<T, Args&&...>,
```

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,182 @@
...
+
+    size_t mask_;
+    Slot<T>* slots_;
+    [[no_unique_address]] Allocator allocator_;
+
+    // Align to avoid false sharing between head_ and tail_
+    alignas(hardwareInterferenceSize) std::atomic<size_t> head_;
+    alignas(hardwareInterferenceSize) std::atomic<size_t> tail_;
+
+    static_assert(std::is_nothrow_copy_assignable<T>::value ||
```
lioncash, line 167: This and the below assert can use the `_v` variants of the traits

src/common/bounded_threadsafe_queue.h
behunin: I tried: `std::construct_at(&slots_[i], new Slot<T>());` and it broke.

src/common/bounded_threadsafe_queue.h
```diff
@@ -0,0 +1,198 @@
...
+    template <typename... Args>
+    void construct(Args&&... args) noexcept {
+        static_assert(std::is_nothrow_constructible_v<T, Args&&...>,
+                      "T must be nothrow constructible with Args&&...");
+        std::construct_at(reinterpret_cast<T*>(&storage), std::forward<Args>(args)...);
+    }
+
+    void destroy() noexcept {
+        static_assert(std::is_nothrow_destructible_v<T>, "T must be nothrow destructible");
+        std::destroy_at(&storage);
```
lioncash, line 66: This likely needs to be:

```cpp
std::destroy_at(reinterpret_cast<T*>(&storage));
```

otherwise the destructor for T wouldn't run, since it'd treat it as calling the destructor on unsigned char

bunnei:
LGTM, thank you!

Issue #8407, opened 2022-06-01
Translation issue - french Windows game list settings section


ErixTricittey:
Translation issue: In french frontend, game icon size and folder icon size option labels have been inverted. So setting game icon size would set the folder size value instead, and setting folder icon size sets game icon size. Just changing frontend text would fix this.

![Capture d’écran 2022-06-01 085753](https://user-images.githubusercontent.com/101439928/171409815-2c349c6f-8d03-40c7-91ce-23faca87779e.png)

G-Spawn:
this should be corrected on Transifex now if I'm not mistaken. If not please discuss in the Translation channel on discord.

G-Spawn:
![image](https://user-images.githubusercontent.com/23653025/173757801-88d3e6ba-a9e5-42f6-8f51-a29ce3b568dc.png)
Seems the user who meant to change it didn't. So I did. It should be reflected in mainline-early access whenever the changes are pushed through.

PR #8418, opened 2022-06-03
Fix mixed up option descriptions

Camonophy:
#8407


Docteh:
Is this also fixed on Transifex? https://www.transifex.com/yuzu-emulator/yuzu/

Docteh:
I just pulled the latest from transifex and it looks like it needs to be updated over there. I'm going to mention this in the #translation channel over on the discord. 

Yuzu pulls the translations from transifex automatically every month.

I checked the [blame](https://github.com/yuzu-emu/yuzu/blame/master/dist/languages/fr.ts) because I was wondering how recent the mistake was. 
image

bunnei:
Thank you for the fix. However, yes, Docteh is right -- Please update this on Transifex as these changes will just be overwritten.

PR #8427, opened 2022-06-05
deprecate usage of QDesktopWidget for going fullscreen

Docteh:
This is related, but not dependent on #8318 I'm PR-ing this separately as the method might be wrong, and this also requires Linux testing. 

------

Idea works as follows, while going fullscreen we compare the current window geometry with
available screens and ask for an intersection rectangle, we go fullscreen where most of
the window is located

GuessCurrentScreen could also potentially be used to see which screen
the window is on for dynamic DPI handling

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -52,7 +52,6 @@ static FileSys::VirtualFile VfsDirectoryCreateFileWrapper(const FileSys::Virtual
 #define QT_NO_OPENGL
 #include <QClipboard>
 #include <QDesktopServices>
-#include <QDesktopWidget>
 #include <QFile>
 #include <QFileDialog>
 #include <QInputDialog>
@@ -60,6 +59,7 @@ static FileSys::VirtualFile VfsDirectoryCreateFileWrapper(const FileSys::Virtual
 #include <QProgressBar>
 #include <QProgressDialog>
 #include <QPushButton>
+#include <QScreen>
 #include <QShortcut>
 #include <QStatusBar>
 #include <QString>
@@ -1002,7 +1002,7 @@ void GMainWindow::InitializeHotkeys() {
 
 void GMainWindow::SetDefaultUIGeometry() {
     // geometry: 53% of the window contents are in the upper screen half, 47% in the lower half
-    const QRect screenRect = QApplication::desktop()->screenGeometry(this);
+    const QRect screenRect = QGuiApplication::primaryScreen()->geometry();
 
     const int w = screenRect.width() * 2 / 3;
     const int h = screenRect.height() * 2 / 3;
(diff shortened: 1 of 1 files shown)


src/yuzu/main.cpp
```diff
@@ -2580,7 +2580,22 @@ void GMainWindow::ToggleFullscreen() {
         HideFullscreen();
     }
 }
-
+static QScreen* GuessCurrentScreen(QWidget* window) {
+    // We're going to return the screen that the yuzu window has the most pixels on
+    QScreen* best_screen = QApplication::primaryScreen();
+    int best_amount = 0;
+    const QList<QScreen*> screens = QGuiApplication::screens();
+    for (const auto& screen : screens) {
```
lioncash, line 2588: This seems like it can be compressed into a usage of `std::max_element`

src/yuzu/main.cpp
```diff
@@ -2580,7 +2580,22 @@ void GMainWindow::ToggleFullscreen() {
...
+        const QRect intersection_geometry = screen->geometry().intersected(window->geometry());
+        const int intersection_area =
+            intersection_geometry.width() * intersection_geometry.height();
+        if (intersection_area > best_amount) {
+            best_amount = intersection_area;
+            best_screen = screen;
+        }
+    }
+    return best_screen;
+}
```
lioncash, line 2598: There should be a newline after this brace to separate this function from the one below it.

src/yuzu/main.cpp
```diff
@@ -2580,7 +2580,22 @@ void GMainWindow::ToggleFullscreen() {
         HideFullscreen();
     }
 }
-
```
lioncash, line 2583: This newline should be left in

src/yuzu/main.cpp
```diff
@@ -2581,6 +2581,20 @@ void GMainWindow::ToggleFullscreen() {
     }
 }
 
+// We're going to return the screen that the given window has the most pixels on
+static QScreen* GuessCurrentScreen(QWidget* window) {
+    QScreen* best_screen = QApplication::primaryScreen();
+    const QList<QScreen*> screens = QGuiApplication::screens();
+    best_screen = *std::max_element(
```
lioncash, line 2588: Wouldn't this just be able to be a direct return since the initial value gets overwritten by the `max_element()` call?

Docteh: Yes, I started off with the GuessCurrentScreen's code inside the function that calls it. I'll shift things around.

lioncash:
Looks fine. Has anyone on Linux tested this?

Docteh:
Not that I've heard. I will solicit the discord.

dejelloper:
Works on Ubuntu 22.04 w/dual monitors

lioncash (approved):

PR #8434, opened 2022-06-07
input_common: Replace usage of string guid to common uuid

german77:
At some point SDL returned a string with 33 characters which is bigger than the UUID size. This might be a bug on SDL side but s since we don't really need to use the string at all it's just easier to replace all instances with Common::UUID

This PR should have no difference for the user.

### DIFF
--- a/src/input_common/drivers/sdl_driver.cpp
+++ b/src/input_common/drivers/sdl_driver.cpp
@@ -13,11 +13,11 @@
 namespace InputCommon {
 
 namespace {
-std::string GetGUID(SDL_Joystick* joystick) {
+Common::UUID GetGUID(SDL_Joystick* joystick) {
     const SDL_JoystickGUID guid = SDL_JoystickGetGUID(joystick);
-    char guid_str[33];
-    SDL_JoystickGetGUIDString(guid, guid_str, sizeof(guid_str));
-    return guid_str;
+    std::array<u8, 16> data{};
+    std::memcpy(data.data(), guid.data, sizeof(data));
+    return Common::UUID{data};
 }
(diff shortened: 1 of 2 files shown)


src/input_common/drivers/sdl_driver.cpp
```diff
@@ -13,11 +13,11 @@
...
 namespace {
-std::string GetGUID(SDL_Joystick* joystick) {
+Common::UUID GetGUID(SDL_Joystick* joystick) {
     const SDL_JoystickGUID guid = SDL_JoystickGetGUID(joystick);
-    char guid_str[33];
-    SDL_JoystickGetGUIDString(guid, guid_str, sizeof(guid_str));
-    return guid_str;
+    std::array<u8, 16> data{};
+    std::memcpy(data.data(), guid.data, sizeof(data));
+    return Common::UUID{data};
```
vonchenplus, line 20: It might be more concise to return the RawString of the current UUID

german77: All the input side uses Common::UUID to identify the controller. SDL was the exception. 

The reason I'm not using 'SDL_JoystickGetGUIDString' it's because SDL returned an invalid string somehow from an array of 16 bytes. I don't know why or how it was even posible. But the proof is clear.

vonchenplus: I am referring to the RawString function using Common::UUI, so that the cost of code modification is minimal.

german77: That's an option too. But this way we can avoid a similar error in the future. Since the string doesn't have a defined size

PR #8447, opened 2022-06-10
Revert "buffer_cache: reset cached write bits after flushing invalidations"

jacky400:
Reverts yuzu-emu/yuzu#8074


jacky400:
Sorry , my fault that pressed the revert button . Closing it now.

@liamwhite , if there is other workaround , it would be great.

PR #8449, opened 2022-06-11
retranslate the game list placeholder

Docteh:
This is the "Double-click to add a new folder to the game list" message
that shows up when users first launch yuzu and is most likely never seen
again. Previously this message was not re-translated.

-----
Bug is triggered by changing language before adding a directory to game list (Start in English, switch to Russian)
image

With PR
image

### DIFF
--- a/src/yuzu/game_list.cpp
+++ b/src/yuzu/game_list.cpp
@@ -870,7 +870,7 @@ GameListPlaceholder::GameListPlaceholder(GMainWindow* parent) : QWidget{parent} 
     layout->setAlignment(Qt::AlignCenter);
     image->setPixmap(QIcon::fromTheme(QStringLiteral("plus_folder")).pixmap(200));
 
-    text->setText(tr("Double-click to add a new folder to the game list"));
+    RetranslateUI();
     QFont font = text->font();
     font.setPointSize(20);
     text->setFont(font);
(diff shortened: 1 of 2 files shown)


src/yuzu/game_list.h
```diff
@@ -166,6 +166,9 @@ private slots:
     void mouseDoubleClickEvent(QMouseEvent* event) override;
 
 private:
+    void changeEvent(QEvent* event);
```
lioncash, line 169: Doesn't this need to be marked as an override?

src/yuzu/game_list.cpp
```diff
@@ -870,7 +870,7 @@ GameListPlaceholder::GameListPlaceholder(GMainWindow* parent) : QWidget{parent}
     layout->setAlignment(Qt::AlignCenter);
     image->setPixmap(QIcon::fromTheme(QStringLiteral("plus_folder")).pixmap(200));
 
-    text->setText(tr("Double-click to add a new folder to the game list"));
+    GameListPlaceholder::RetranslateUI();
```
lioncash, line 873:
```suggestion
    RetranslateUI();
```

src/yuzu/game_list.h
Docteh: Yes. fixed.

lioncash (approved):

PR #8455, opened 2022-06-12
ci/windows: Use Clang for MinGW builds

lat9nq:
The reason I wanted to do this is twofold: GCC 12 appears to have optimization errors and bogus warnings, and in the future I want to move MinGW to Clang anyway for possibly releasing these builds instead of those from MSVC.

We aren't ready to release any MinGW build though due to stability issues in yuzu's VFS. However, the timing of this part wasn't entirely my choice since Arch Linux has moved their MinGW compiler to GCC 12.

I would still like to see https://github.com/yuzu-emu/yuzu/pull/8439 working before merging this, but if for some reason it doesn't seem worth it, that isn't a hard requirement.

One last thing: I wanted to include [Polly](https://polly.llvm.org/) optimizations in this PR, but I'd rather back that up with evidence showing that we benefit from it at a later time.

### DIFF
--- a/.ci/scripts/windows/docker.sh
+++ b/.ci/scripts/windows/docker.sh
@@ -5,7 +5,14 @@ cd /yuzu
 ccache -s
 
 mkdir build || true && cd build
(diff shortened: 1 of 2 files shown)


Morph1984 (approved):

Tachi107:
What's the advantage in using Clang + MinGW compared to only using native Windows Clang?

PR #8456, opened 2022-06-12
Modify if, else if, & else statement set

Velocities:
Just changing the set of if, else if, and else statement set to align it with programming best-practices.


lioncash:
`switch` statements can only be used with enum or integral types, not string types. If this does actually work, it's because of a compiler extension that we shouldn't be relying on.

While the contribution attempt is appreciated, it would be beneficial to ensure code compiles before staging it

PR #8459, opened 2022-06-14
vk_compute_pass: Silence Wextra warning

Morph1984:
Silences a warning about using enumerated and non-enumerated types in a conditional expression.


vonchenplus:
MSVC build error.

PR #8474, opened 2022-06-16
Make yuzu-cmd respect log_filter setting

DCNick3:
Because logging infrastructure initializes before the loading of the
config, it reads the default setting for log_filter and ignores the one
set in config. To change log_filter after logging initialization some
additional calls need to be made.

### DIFF
--- a/src/yuzu_cmd/yuzu.cpp
+++ b/src/yuzu_cmd/yuzu.cpp
@@ -138,6 +138,12 @@ int main(int argc, char** argv) {
 
     Config config{config_path};
 
+    // apply the log_filter setting
+    // the logger was initialized before and doesn't pick up the filter on its own
+    Common::Log::Filter filter;
+    filter.ParseFilterString(Settings::values.log_filter.GetValue());
+    Common::Log::SetGlobalFilter(filter);
+
     if (!program_args.empty()) {
         Settings::values.program_args = program_args;
     }


src/yuzu_cmd/config.cpp
```diff
@@ -331,6 +333,12 @@ void Config::ReadValues() {
     // log_filter has a different default here than from common
     Settings::values.log_filter =
         sdl2_config->Get("Miscellaneous", Settings::values.log_filter.GetLabel(), "*:Trace");
+
+    // apply the log_filter setting, the logger was initialized before
+    Common::Log::Filter filter;
+    filter.ParseFilterString(Settings::values.log_filter.GetValue());
+    Common::Log::SetGlobalFilter(filter);
```
Morph1984, line 340: I'd rather move this to main after the config is loaded rather than here

DCNick3: FYU: this is what qt version does in `ConfigureDebug::ApplyConfiguration`: https://github.com/yuzu-emu/yuzu/blob/de2f2e5140eb85311e0fd844c580d6726adf7e03/src/yuzu/configuration/configure_debug.cpp#L78

Morph1984: This is because it is in a separate file from where the configuration values are actually read, and the UI action should directly correlate with a change in the logging system. config.cpp in yuzu's file does not do this.

src/yuzu_cmd/config.cpp
```diff
@@ -20,6 +20,8 @@
 #include "common/fs/file.h"
 #include "common/fs/fs.h"
 #include "common/fs/path_util.h"
+#include "common/logging/backend.h"
+#include "common/logging/filter.h"
```
Morph1984, line 24: The changes in this file should be removed

DCNick3: Ah, right, sorry

Morph1984:
Also since the change has been entirely moved to a different file, the commits should likely be squashed to attribute the changed lines directly to the commit description

DCNick3:
Fair, git blame should be easier

Morph1984 (approved):

Morph1984:
Thanks for the fix!

PR #8485, opened 2022-06-20
cmake: Use compatibility list in source directory

nezd5553:
For [Flatpak builds](https://github.com/flathub/org.yuzu_emu.yuzu), the compatibility list is located in the source directory, because the Flatpak build system is not allowed access to the internet. In this case, CMake will copy it to `${PROJECT_BINARY_DIR}/dist/compatibility_list/`. This fixes the issue in the Flatpak build where the compatibility of all games is "not tested" as mentioned in  flathub/org.yuzu_emu.yuzu#6.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,6 +76,10 @@ if (ENABLE_COMPATIBILITY_LIST_DOWNLOAD AND NOT EXISTS ${PROJECT_BINARY_DIR}/dist
         https://api.yuzu-emu.org/gamedb/
         "${PROJECT_BINARY_DIR}/dist/compatibility_list/compatibility_list.json" SHOW_PROGRESS)
 endif()
+if (EXISTS ${PROJECT_SOURCE_DIR}/compatibility_list.json)
+    file(COPY "${PROJECT_SOURCE_DIR}/compatibility_list.json"
+        DESTINATION "${PROJECT_BINARY_DIR}/dist/compatibility_list/")
+endif()
 if (NOT EXISTS ${PROJECT_BINARY_DIR}/dist/compatibility_list/compatibility_list.json)
     file(WRITE ${PROJECT_BINARY_DIR}/dist/compatibility_list/compatibility_list.json "")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,15 +70,16 @@ endif()
 configure_file(${PROJECT_SOURCE_DIR}/dist/compatibility_list/compatibility_list.qrc
                ${PROJECT_BINARY_DIR}/dist/compatibility_list/compatibility_list.qrc
                COPYONLY)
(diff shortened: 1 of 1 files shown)


Docteh:
I think there should be a comment explaining the section, maybe even a link to https://github.com/flathub/org.yuzu_emu.yuzu if you're feeling generous 

putting the compatibility list into dist/compatibility_list/ on the ${PROJECT_SOURCE_DIR} side _might_ also be good idea. 

```
# the flatpak build provides its own copy of compatibility list as the flatpak build system is 
# not allowed internet access. flatpak is at https://github.com/flathub/org.yuzu_emu.yuzu 
```

CMakeLists.txt
```diff
@@ -76,6 +76,10 @@ if (ENABLE_COMPATIBILITY_LIST_DOWNLOAD AND NOT EXISTS ${PROJECT_BINARY_DIR}/dist
         https://api.yuzu-emu.org/gamedb/
         "${PROJECT_BINARY_DIR}/dist/compatibility_list/compatibility_list.json" SHOW_PROGRESS)
 endif()
+if (EXISTS ${PROJECT_SOURCE_DIR}/compatibility_list.json)
+    file(COPY "${PROJECT_SOURCE_DIR}/compatibility_list.json"
+        DESTINATION "${PROJECT_BINARY_DIR}/dist/compatibility_list/")
+endif()
```
Tachi107, line 82: As line 73 already checks for the existence of `${PROJECT_BINARY_DIR}/dist/compatibility_list/compatibility_list.json` and skips the download if it already exists, why not putting something like

```cmake
if (EXISTS ${PROJECT_SOURCE_DIR}/dist/compatibility_list/compatibility_list.json)
    configure_file("${PROJECT_SOURCE_DIR}/dist/compatibility_list/compatibility_list.json"
                   "${PROJECT_BINARY_DIR}/dist/compatibility_list/compatibility_list.json"
                   COPYONLY)
endif()
```

under line 72?

nezd5553: I was just copied what was in [Citra's CMakeLists.txt](https://github.com/citra-emu/citra/blob/355933218d00603c910cd0a9daead0b1cd72a3c3/CMakeLists.txt#L85), but I changed it to use the file in `${PROJECT_SOURCE_DIR}` instead of `${PROJECT_SOURCE_DIR}/dist/compatibility_list` so it would work without modifying the Flatpak.

Tachi107: Why not modifying the flatpak instead?

nezd5553: I figured it would be easier that way, but I wasn't thinking about consistency at the time.

Docteh (approved):
LGTM

MaxALEau (approved):

bunnei (approved):

PR #8486, opened 2022-06-20
CI: Use GitHub Actions to validate pull requests

liushuyu:
This pull request will add a new GitHub Actions workflow to **validate pull requests**.

**Note:** that this pull request **does not** remove the pull request validation from ADO. A team member with respective permissions needs to do that themselves.

This pull request includes some other related changes to the yuzu:

- MSVC: When the cache is enabled, Ninja is used instead of MSBuild to use the cache launcher properly
- MSVC: When the cache is enabled, debug information generation will be downgraded to make the debug info more deterministic (more cacheable)
- Linux: All Linux builds now use Ninja to speed up the build (less overhead from the build system itself)
- Linux (fresh/AppImage): When inside the container, the AppImage builder will be launched using the "extract and run" method instead of the FUSE-based approach (requires kernel-side assistance and is not available on GHA when running inside the Docker [Although there are workarounds, I think using the fallback approach is better]).


Morph1984:
Needs a rebase and a review from @lat9nq as #8455 has been merged

liushuyu:
Rebased. Although this would break ADO's CI due to work directory difference, I need to think of a good way to handle that.

.ci/scripts/common/post-upload.sh
```diff
@@ -4,8 +4,10 @@
 cp license.txt "$DIR_NAME"
 cp README.md "$DIR_NAME"
 
-tar -cJvf "${REV_NAME}-source.tar.xz" src externals CMakeLists.txt README.md license.txt
-cp "${REV_NAME}-source.tar.xz" "$DIR_NAME"
+if [[ "x${NO_SOURCE_PACK}" == "x" ]]; then
```
lat9nq, line 7:
```suggestion
if [[ -z "${NO_SOURCE_PACK}" ]]; then
```

lat9nq (commented):
Couple small things caught my eye but overall looks good.

liushuyu:
ADO breakage fixed and comments addressed.

Morph1984:
There still seems to be a build failure in ADO for the mingw run

liushuyu:
> There still seems to be a build failure in ADO for the mingw run

Ah, sorry. I have fixed it by downgrading `wget -nc` (no-clobber) to `wget -c` (continue). `wget -nc` will return an error if the file specified already exists, but `wget -c` will not.

Actually I think we should re-download the Conan's profile configuration each time to avoid outdated files?

lat9nq (approved):

PR #8499, opened 2022-06-25
Translate english plurals

Docteh:
Turns out that for Qt to properly handle plurals in English a
translation needs to be provided, otherwise the user is left with
messages such as "Building: 2 shader(s)"

Plurals for other all other languages are handled on transifex. 
I wrote the README.md to just refer to it as a translation 
collaboration site just in case we ever switch.

These translations being out of date won't pose any technical problems
so I believe it is fine to handle them manually on a "best effort"
basis.

The files are generated into the source directory so that the
relative filenames are correct. The generated file is added to
.gitignore

-------

We might be able to technobabble something decent out of Transifex so I'm submitting this as a draft. Attn: @FearlessTobi 

But in my extensive research on this the best I could find is a reference to Crowdin (a different site that handles crowd sourced translations) having difficulties in handling translations of plurals for the source language. https://wiki.mudlet.org/w/Translating_Mudlet#Plural_forms_in_.28American.29_English

You can pull the source file from transifex, and this is what comes back.

```xml
    <message numerus="yes">
        <location filename="../../src/yuzu/main.cpp" line="3262"/>
        <source>Building: %n shader(s)</source>
        <translation><numerusform>Building: %n shader(s)</numerusform><numerusform>Building: %n shader(s)</numerusform></translation>
    </message>
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ doc-build/
 
 # Generated source files
 src/common/scm_rev.cpp
-.travis.descriptor.json
+dist/english_plurals/generated_en.ts
 
 # Project/editor files
 *.swp
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -2,7 +2,8 @@ Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Comment: It is best to use this file to record copyright information about
          generated, binary and third party files
 
-Files: dist/icons/controller/*.png
+Files: dist/english_plurals/*
+       dist/icons/controller/*.png
        dist/icons/overlay/*.png
        dist/languages/*
        dist/qt_themes/*/icons/index.theme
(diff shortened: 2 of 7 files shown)


Docteh:
All of the strings that end in \n and therefore annoy me are all in the same place. 

https://github.com/yuzu-emu/yuzu/blob/c78f6d4f209b08d16d9220c0884a04e110a5373d/src/yuzu/main.cpp#L2374-L2381

Docteh:
Screenshots might be useful.

Before: Building: 3971 shader(s)
image

This pr: Building: 4384 shaders
image

Tachi107:
Can plurals be handled for other similar languages?

Docteh:
Other languages have working plurals handled via transifex. I'll see about rewording the commit message (the bit above the header line in the first post)


dist/english_plurals/README.md
dist/english_plurals/en.ts

Not sure about what copyright to specify, the en.ts will need to be in that `.reuse/dep5` file as we need to consider someone just taking the generated en.ts and processing it through the Linquist tool

bunnei (approved):

PR #8506, opened 2022-06-26
Fix get nvmap object random crash

vonchenplus:
nvdisp_disp0 will get an object from nvmap according to handle when flip, but this object may have been released early, so we need to add a reference when nvfinger is in QueueBuffer.

This change fixes random crashes on Teenage Mutant Ninja Turtles Shredder's Revenge startup.


vonchenplus:
pull request to [yfc](https://github.com/FernandoS27/yuzu/pull/4)，closed.

PR #8508, opened 2022-06-26
hle: service: nvflinger: Factor speed limit into frame time calculation.

bunnei:
- This allows the %-based "Limit Speed Percent" setting to work with MC emulation.
- This is already supported for SC emulation.


Fneufneu:
just FYI, this give me an error `conversion from ‘long int’ to ‘float’ may change value [-Werror=conversion]`

fixed by `(float)(max_hertz * fps_cap)`

FernandoS27 (approved):

gidoly:
Several stuff need to address before merge.
+ remove frame rate cap option from setting as this option completely replace it.
+ move ctrl + u hotkey to this option.

german77 (approved):

guidodi:
Would it be possible to add the frame cap back? The option is called "limit speed" but it doesnt really limit speed to 200% or whatever you set it to, what it does is make the game run at 200% speed all the time but you can still unlock fps and it will go to 200fps even with speed limit enabled.

I would like to be able to play Pokemon Sword at default speed of 30 and speed it up to 60 fps in battles (like it worked before with frame cap x2). Right now my only option is setting speed limit to 200% but I am forced to play at 60 all the time, even outside of battles and unlocking fps in battles goes to 200fps which makes the game unplayable.

PR #8510, opened 2022-06-26
input_common: sdl: lower vibration frequency and use it's own unique thread

german77:
When I moved vibrations to the poll thread it created a whole new set of issues. As now controllers would now go unresponsive rather than having stutter on the game. This PR reduces the amount of vibrations send per second as well move them out of the poll thread to avoid locking any controller input.

If issues still persist I might consider disabling vibration devices on the fly if they take too long to respond.

### DIFF
--- a/src/core/hle/service/hid/controllers/npad.cpp
+++ b/src/core/hle/service/hid/controllers/npad.cpp
@@ -820,11 +820,11 @@ bool Controller_NPad::VibrateControllerAtIndex(Core::HID::NpadIdType npad_id,
 
         const auto now = steady_clock::now();
 
-        // Filter out non-zero vibrations that are within 10ms of each other.
+        // Filter out non-zero vibrations that are within 15ms of each other.
         if ((vibration_value.low_amplitude != 0.0f || vibration_value.high_amplitude != 0.0f) &&
             duration_cast<milliseconds>(
                 now - controller.vibration[device_index].last_vibration_timepoint) <
-                milliseconds(10)) {
+                milliseconds(15)) {
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


src/input_common/drivers/sdl_driver.cpp
```diff
@@ -438,10 +438,17 @@ SDLDriver::SDLDriver(std::string input_engine_) : InputEngine(std::move(input_en
             using namespace std::chrono_literals;
             while (initialized) {
                 SDL_PumpEvents();
-                SendVibrations();
                 std::this_thread::sleep_for(1ms);
             }
         });
+        vibration_thread = std::thread([this] {
+            Common::SetCurrentThreadName("yuzu:input:SDL_vibration");
```
Morph1984, line 445:
```suggestion
            Common::SetCurrentThreadName("yuzu:input:SDL_Vibration");
```

src/core/hle/service/hid/controllers/npad.cpp
```diff
@@ -820,11 +820,11 @@ bool Controller_NPad::VibrateControllerAtIndex(Core::HID::NpadIdType npad_id,
 
         const auto now = steady_clock::now();
 
-        // Filter out non-zero vibrations that are within 10ms of each other.
+        // Filter out non-zero vibrations that are within 15ms of each other.
         if ((vibration_value.low_amplitude != 0.0f || vibration_value.high_amplitude != 0.0f) &&
             duration_cast<milliseconds>(
                 now - controller.vibration[device_index].last_vibration_timepoint) <
-                milliseconds(10)) {
+                milliseconds(15)) {
```
Morph1984, line 827: Curious why this is different to the sleep on the vibration thread

german77: If vibration is making things lag so much it's probably taking a few ms.  So if  both run with the same delay the driver will be always slower.

liamwhite (approved):

PR #8512, opened 2022-06-27
Replace multiple names with a better name

german77:
This PR will probably conflict with everything, handle with care. ResultCode and KPageLinkedList where replaced with Result and KPageGroup to match HOS names. As well it removes any VK from non Vulkan SDK classes to avoid confusions.

Functionally wise nothing was changed.


Morph1984 (approved):
yolo

PR #8518, opened 2022-06-29
Revert "vulkan_device: Block AMDVLK's VK_KHR_push_descriptor"

lat9nq:
Reverts yuzu-emu/yuzu#8379

AMD's latest driver fixes the issue worked around in #8379, but retains the same version string in the driver -- 2.0.226.

So since AMD didn't update the version string, solution here is to just revert it and tell people to update their driver if they experience crashes.


Moonlacer:
Did AMD fix the unknown crashes on the 6400 and 6500xt?

goldenx86:
Unsure, I'll confirm you once mine is back.

goldenx86 (commented):
LGTM :D

FernandoS27 (approved):

PR #8544, opened 2022-07-07
service: Update some services to 14.0.0+

german77:


src/core/hle/service/acc/acc.cpp
```diff
@@ -406,15 +406,15 @@ class IProfileCommon : public ServiceFramework<IProfileCommon> {
                       reinterpret_cast<const char*>(base.username.data()), base.username.size()),
                   base.timestamp, base.user_uuid.RawString());
 
-        if (user_data.size() < sizeof(ProfileData)) {
+        if (user_data.size() < sizeof(UserData)) {
             LOG_ERROR(Service_ACC, "ProfileData buffer too small!");
```
Docteh, line 410: Does this one get renamed too?

german77: Good catch

liamwhite (approved):

bunnei (approved):

PR #8561, opened 2022-07-10
Rework CoreTiming events

Kelebek1:
The main idea here is to make coretiming much more accurate in its callback timing, and add an automatically looping type of event, so that functions don't need to keep re-registering themselves, as almost all events are auto, and not one-shot.

The current problem with coretiming callbacks is the reliance on the callback to re-schedule itself, which necessarily means that some time is lost, between the coretiming timestamp to calculate `ns_late`, and the schedule function doing the same thing some point later. 

If some event was scheduled at time 100, maybe the callback delay is calculated at time 102, passing a ns_late of 2 into the callback. The callback runs and at the end wants to reschedule itself 10 units later. It calls reschedule with 10 - 2, the scheduling function then gets the current time, which may now be 103 or 104, and adds 8. A inherant drift happens due to continued calling to get a timestamp and some "lost time".

This PR changes the callback such that you can register an event which will loop itself, and will automatically use the event time + reschedule time. So if you event scheduled at time 100, with a reschedule of 10, it will *always* reschedule it for time 100 + 10, and then 110 + 10, and then 120 + 10, etc. The *only* time a timestamp is taken, is on the very first schedule to start the event, so it'll reschedule with the appropriate time.

Here's some timestamp logging from hidbus for example, which aims for 15,000,000:
old https://pastebin.com/iPq2rFK6
new https://pastebin.com/JxTAj9ZB

The current frame timings are larger, and in particular the log timestamp itself is all over the place as the time drifts forward. In the new, it's able to fix quite firmly on 3758 -> 8758 -> 3758 -> 8758.

This is spun out of Project Andio, which needs a much more accurate callback than coretiming was giving, due to the real-time deadline of audio output. PR'ing separately due to merge conflicts.

### DIFF
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -245,8 +245,7 @@ struct KernelCore::Impl {
             });
 
         const auto time_interval = std::chrono::nanoseconds{std::chrono::milliseconds(10)};
-        system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), time_interval,
-                                                 preemption_event);
+        system.CoreTiming().ScheduleLoopingEvent(time_interval, time_interval, preemption_event);
     }
 
     void InitializeShutdownThreads() {
--- a/src/core/hle/kernel/kernel.cpp
+++ b/src/core/hle/kernel/kernel.cpp
@@ -234,17 +234,19 @@ struct KernelCore::Impl {
 
     void InitializePreemption(KernelCore& kernel) {
         preemption_event = Core::Timing::CreateEvent(
-            "PreemptionCallback", [this, &kernel](std::uintptr_t, std::chrono::nanoseconds) {
+            "PreemptionCallback",
+            [this, &kernel](std::uintptr_t, s64 time,
+                            std::chrono::nanoseconds) -> std::optional<std::chrono::nanoseconds> {
                 {
                     KScopedSchedulerLock lock(kernel);
                     global_scheduler_context->PreemptThreads();
                 }
-                const auto time_interval = std::chrono::nanoseconds{std::chrono::milliseconds(10)};
-                system.CoreTiming().ScheduleEvent(time_interval, preemption_event);
+                return std::nullopt;
             });
 
         const auto time_interval = std::chrono::nanoseconds{std::chrono::milliseconds(10)};
-        system.CoreTiming().ScheduleEvent(time_interval, preemption_event);
+        system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), time_interval,
+                                                 preemption_event);
     }
 
     void InitializeShutdownThreads() {
--- a/src/core/hle/service/hid/hid.cpp
+++ b/src/core/hle/service/hid/hid.cpp
@@ -97,11 +97,10 @@ IAppletResource::IAppletResource(Core::System& system_,
             return std::nullopt;
         });
 
-    system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), pad_update_ns,
-                                             pad_update_event);
-    system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), mouse_keyboard_update_ns,
+    system.CoreTiming().ScheduleLoopingEvent(pad_update_ns, pad_update_ns, pad_update_event);
+    system.CoreTiming().ScheduleLoopingEvent(mouse_keyboard_update_ns, mouse_keyboard_update_ns,
                                              mouse_keyboard_update_event);
-    system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), motion_update_ns,
+    system.CoreTiming().ScheduleLoopingEvent(motion_update_ns, motion_update_ns,
                                              motion_update_event);
 
     system.HIDCore().ReloadInputDevices();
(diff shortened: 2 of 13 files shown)


FernandoS27 (approved):

src/core/hle/kernel/kernel.cpp
```diff
@@ -234,17 +234,19 @@ struct KernelCore::Impl {
...
                     global_scheduler_context->PreemptThreads();
                 }
-                const auto time_interval = std::chrono::nanoseconds{std::chrono::milliseconds(10)};
-                system.CoreTiming().ScheduleEvent(time_interval, preemption_event);
+                return std::nullopt;
             });
 
         const auto time_interval = std::chrono::nanoseconds{std::chrono::milliseconds(10)};
-        system.CoreTiming().ScheduleEvent(time_interval, preemption_event);
+        system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), time_interval,
```
FernandoS27, line 248: start time should be the same as resschedule to have the same functionality as before.

src/core/hle/service/hid/hid.cpp
```diff
@@ -74,26 +74,35 @@ IAppletResource::IAppletResource(Core::System& system_,
...
 
-    system.CoreTiming().ScheduleEvent(pad_update_ns, pad_update_event);
-    system.CoreTiming().ScheduleEvent(mouse_keyboard_update_ns, mouse_keyboard_update_event);
-    system.CoreTiming().ScheduleEvent(motion_update_ns, motion_update_event);
+    system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), pad_update_ns,
+                                             pad_update_event);
+    system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), mouse_keyboard_update_ns,
+                                             mouse_keyboard_update_event);
+    system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), motion_update_ns,
+                                             motion_update_event);
```
FernandoS27, line 105: ditto.

src/core/hle/service/hid/hidbus.cpp
```diff
@@ -50,21 +50,22 @@ HidBus::HidBus(Core::System& system_)
...
-        [this](std::uintptr_t user_data, std::chrono::nanoseconds ns_late) {
+        [this](std::uintptr_t user_data, s64 time,
+               std::chrono::nanoseconds ns_late) -> std::optional<std::chrono::nanoseconds> {
             const auto guard = LockService();
             UpdateHidbus(user_data, ns_late);
+            return std::nullopt;
         });
 
-    system_.CoreTiming().ScheduleEvent(hidbus_update_ns, hidbus_update_event);
+    system_.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), hidbus_update_ns,
```
FernandoS27, line 60: ditto.

src/core/hle/service/nvflinger/nvflinger.cpp
```diff
@@ -67,21 +67,20 @@ NVFlinger::NVFlinger(Core::System& system_, HosBinderDriverServer& hos_binder_dr
...
-
-            this->system.CoreTiming().ScheduleEvent(future_ns, composition_event);
+            return std::chrono::nanoseconds(GetNextTicks()) - ns_late;
         });
 
     if (system.IsMulticore()) {
         vsync_thread = std::jthread([this](std::stop_token token) { SplitVSync(token); });
     } else {
-        system.CoreTiming().ScheduleEvent(frame_ns, composition_event);
+        system.CoreTiming().ScheduleLoopingEvent(std::chrono::nanoseconds(0), frame_ns,
```
FernandoS27, line 82: ditto.

src/core/hle/service/nvflinger/nvflinger.cpp
```diff
@@ -67,21 +67,20 @@ NVFlinger::NVFlinger(Core::System& system_, HosBinderDriverServer& hos_binder_dr
...
+               std::chrono::nanoseconds ns_late) -> std::optional<std::chrono::nanoseconds> {
             const auto lock_guard = Lock();
             Compose();
 
-            const auto ticks = std::chrono::nanoseconds{GetNextTicks()};
-            const auto ticks_delta = ticks - ns_late;
-            const auto future_ns = std::max(std::chrono::nanoseconds::zero(), ticks_delta);
-
-            this->system.CoreTiming().ScheduleEvent(future_ns, composition_event);
+            return std::chrono::nanoseconds(GetNextTicks()) - ns_late;
```
FernandoS27, line 76: this is not equivalent.

src/core/memory/cheat_engine.cpp
```diff
@@ -184,10 +184,12 @@ CheatEngine::~CheatEngine() {
...
     event = Core::Timing::CreateEvent(
         "CheatEngine::FrameCallback::" + Common::HexToString(metadata.main_nso_build_id),
-        [this](std::uintptr_t user_data, std::chrono::nanoseconds ns_late) {
+        [this](std::uintptr_t user_data, s64 time,
+               std::chrono::nanoseconds ns_late) -> std::optional<std::chrono::nanoseconds> {
             FrameCallback(user_data, ns_late);
+            return std::nullopt;
         });
-    core_timing.ScheduleEvent(CHEAT_ENGINE_NS, event);
+    core_timing.ScheduleLoopingEvent(std::chrono::nanoseconds(0), CHEAT_ENGINE_NS, event);
```
FernandoS27, line 192: ditto.

PR #8564, opened 2022-07-10
yuzu: Streamline broken Vulkan handling

lat9nq:
I'm looking to do some Windows-specific work in the future that involves just a fork and watching the child. Since this *actually* means using CreateProcess and starting over with a fresh instance of the yuzu frontend, I thought I would practice it by reworking my broken Vulkan handling on yuzu's boot, which was #8393.

Previously, when yuzu boots, it would start the Vulkan loader and, in the event that the host environment is not working, crash. The next time the user starts the application, yuzu is able to detect the crash and adjust accordingly.

The approach in this PR instead uses a child process whose sole purpose is to start the Vulkan loader. If the child crashes, the parent marks the current instance as not having Vulkan available. In effect the user only has to start yuzu once to see the error message. To fix it, the user has to just restart yuzu to check if Vulkan is working since the `has_broken_vulkan` setting is no longer read or saved to a config.

I think one last thing I need to mention is that I noticed that selecting the shader_backend setting in custom configurations was not working after the previous PR. I just disabled setting shader_backend in the event that Vulkan is broken.


src/yuzu/startup_checks.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    // Wait until the processs exits and get exit code from it
+    DWORD exit_code = STILL_ACTIVE;
+    while (exit_code == STILL_ACTIVE) {
+        const int err = GetExitCodeProcess(process_info.hProcess, &exit_code);
+        if (err == 0) {
+            std::fprintf(stderr, "GetExitCodeProcess failed with error %d\n", GetLastError());
+            break;
+        }
+    }
```
lat9nq, line 68: WaitForSingleObject

src/yuzu/startup_checks.cpp
```diff
@@ -0,0 +1,133 @@
...
+        std::fprintf(stderr, "Failed to initialize Vulkan: %s\n", exception.what());
+    }
+}
+
+bool StartupChecks(const char* arg0, bool* has_broken_vulkan) {
+#ifdef _WIN32
+    // Check environment variable to see if we are the child
+    char variable_contents[8];
+    const DWORD startup_check_var =
+        GetEnvironmentVariableA(STARTUP_CHECK_ENV_VAR, variable_contents, 8);
```
lat9nq, line 39: Clean up this variable on the way out

lat9nq (commented):
Reminders to myself for tomorrow

Docteh:
Does this just check if the sub process crashes, or can this also do the CheckSuitability? 

The  function writes out to the log and can also `throw vk::Exception(VK_ERROR_FEATURE_NOT_PRESENT);`

Just something I noticed a few times on the discord, it appears that initial Vulkan checks pass, but then later it fails on an extension, and then the application stops responding.

Somebody elses log
[yuzu_log (2).txt](https://github.com/yuzu-emu/yuzu/files/9146077/yuzu_log.2.txt)

```
[   3.299848] Render.Vulkan <Error> video_core\vulkan_common\vulkan_device.cpp:CheckSuitability:862: Missing required extension: VK_EXT_robustness2
[   3.303612] Render.Vulkan <Error> video_core\renderer_vulkan\renderer_vulkan.cpp:RendererVulkan:114: Vulkan initialization failed with error: VK_ERROR_EXTENSION_NOT_PRESENT
[   3.303653] HW.GPU <Error> video_core\video_core.cpp:CreateGPU:54: Failed to initialize GPU: Vulkan initialization error VK_ERROR_EXTENSION_NOT_PRESENT
[   3.308172] Core <Critical> core\core.cpp:Load:263: Failed to initialize system (Error 5)!
```

lat9nq:
@Docteh the purpose of these checks at boot is to check if starting the Vulkan loader causes the entire application to crash. If the Vulkan driver does not meet our requirements, yuzu is still able to handle it without crashing like you have shown with the log file. For the purpose of both this PR and #8393, we don't care if yuzu is able to handle a unsuitable Vulkan driver. We only care if the system's Vulkan configuration makes the application unusable (e.g. the user should be able to open the configuration dialog without crashing the entire application).

IIRC that the application fails to respond after attempting to boot with an unsuitable (but stable) Vulkan driver is an entirely separate issue. From what I remember this happens because video_core decides to cut the cord early while the guest is trying to boot. We then fail to cleanly shut down the rest of the emulated system.

bunnei (approved):

PR #8566, opened 2022-07-11
core: hid: Add fallback for dualjoycon and pro controllers

german77:
Super Mario Galaxy soft locked when dual joycon was selected. This was due the game only enabling pro controller and handheld support at boot. Then later on enabling dual joycon support. 

Original behavior will disconnect the controller since it's not supported at boot.  But after a few HW test with homebrew I found that the controller never stops working it just fallback itself to pro controller then restore the original type once it's supported.

This PR fixes Mario Galaxy with dual joycons.

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -99,8 +99,10 @@ void EmulatedController::ReloadFromSettings() {
     // Other or debug controller should always be a pro controller
     if (npad_id_type != NpadIdType::Other) {
         SetNpadStyleIndex(MapSettingsTypeToNPad(player.controller_type));
+        original_npad_type = npad_type;
     } else {
         SetNpadStyleIndex(NpadStyleIndex::ProController);
+        original_npad_type = npad_type;
     }
 
     if (player.connected) {
@@ -339,6 +341,7 @@ void EmulatedController::DisableConfiguration() {
             Disconnect();
         }
         SetNpadStyleIndex(tmp_npad_type);
+        original_npad_type = tmp_npad_type;
     }
 
     // Apply temporary connected status to the real controller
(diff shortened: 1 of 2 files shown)


src/core/hid/emulated_controller.cpp
```diff
@@ -950,20 +953,50 @@ void EmulatedController::SetSupportedNpadStyleTag(NpadStyleTag supported_styles)
...
+    if (npad_type == NpadStyleIndex::JoyconDual && supported_style_tag.fullkey) {
+        LOG_WARNING(Service_HID, "Reconnecting controller type {} as Pro controller", npad_type);
+        SetNpadStyleIndex(NpadStyleIndex::ProController);
+        Connect();
+        return;
+    }
+
+    // Fallback Pro controllers to Dual joycon
+    if (npad_type == NpadStyleIndex::ProController && supported_style_tag.joycon_dual) {
+        LOG_WARNING(Service_HID, "Reconnecting controller type {} as Pro controller", npad_type);
```
liamwhite, line 994: `Reconnecting controller type {} as Dual Joycons`

doudou64800D:
how do we modify in yuzu the file in hid?

src/core/hid/emulated_controller.cpp
doudou64800D: how do we modify in yuzu the file in hid?

liamwhite:
Note that SMG still crashes with this change if you do not connect any controller. It does not do this on console, instead it just opens the controller applet:
output

Could you implement this behavior as well?

bunnei (approved):

bunnei:
As discussed on Discord, liamwhite's suggestion will be left to a future PR

PR #8542, opened 2022-07-06
gpu_thread: Use the previous MPSCQueue implementation

Morph1984:
The bounded MPSCQueue implementation causes crashes in Fire Emblem Three Houses, use the previous implementation for now.

PR #8573, opened 2022-07-13
video_core: Fix gpu thread deadlock

vonchenplus:
Supersedes PR #8542

Issue caused by bounded queue introduced by PR #8413.

Because the bounded queue is capped by the capacity. The queue has multiple threads push command at the same time, and only the gpu thread is consuming. The problem is that the gpu thread is also adding data to it. When the queue is full, this will cause the gpu thread to be blocked, and the gpu thread will directly deadlock, that is, we will see that the screen will not be refreshed.

This fixes the DRAGON QUEST BUILDERS startup deadlock. Other more tests are welcome.

### DIFF
--- a/src/video_core/dma_pusher.cpp
+++ b/src/video_core/dma_pusher.cpp
@@ -34,7 +34,6 @@ void DmaPusher::DispatchCalls() {
     }
     gpu.FlushCommands();
     gpu.SyncGuestHost();
-    gpu.OnCommandListEnd();
 }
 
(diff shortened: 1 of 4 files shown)


gidoly:
seems to not adresses fire emblem three houses crash.

vonchenplus:
> 

This PR doesn't fix any crashes

vonchenplus:
@gidoly does not reproduce the crash, he refers to the screen freezing, which is not caused by the bounded queue.

Issue #8606, opened 2022-07-18
[1093-1094] Crash on startup when controller connected (Flatpak)


Beanow:
Running on Steam Deck, when I have an Xbox 360 controller connected, Yuzu crashes on startup.
I've not had this issue with 1085.

Here a stacktrace

```
Reading symbols from /app/bin/yuzu...
Reading symbols from /usr/lib/debug//app/bin/yuzu.debug...
(gdb) r
Starting program: /app/bin/yuzu 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffe5e42640 (LWP 17)]
[New Thread 0x7fffdffbe640 (LWP 18)]
[New Thread 0x7fffddaed640 (LWP 19)]
[New Thread 0x7fffdd1ab640 (LWP 20)]
[New Thread 0x7fffdc9aa640 (LWP 21)]
[New Thread 0x7fffcffff640 (LWP 22)]
[New Thread 0x7fffcf7fe640 (LWP 23)]
[New Thread 0x7fffceffd640 (LWP 24)]
[New Thread 0x7fffce7fc640 (LWP 25)]
[New Thread 0x7fffcdffb640 (LWP 26)]
[New Thread 0x7fffcd7fa640 (LWP 27)]
[New Thread 0x7fffccff9640 (LWP 28)]
[New Thread 0x7fffa3fff640 (LWP 29)]
[New Thread 0x7fffa37fe640 (LWP 30)]
[New Thread 0x7fffa2ffd640 (LWP 31)]
[New Thread 0x7fffa27fc640 (LWP 32)]
[New Thread 0x7fffa15eb640 (LWP 33)]
[New Thread 0x7fffa0dea640 (LWP 34)]
[New Thread 0x7fff93fff640 (LWP 35)]
[New Thread 0x7fff937fe640 (LWP 36)]
[New Thread 0x7fff92ffd640 (LWP 37)]
[New Thread 0x7fff927fc640 (LWP 38)]
[New Thread 0x7fff91ffb640 (LWP 39)]
[New Thread 0x7fff917fa640 (LWP 40)]
[New Thread 0x7fff90ff9640 (LWP 41)]
[New Thread 0x7fff7bfff640 (LWP 42)]

Thread 1 "yuzu" received signal SIGSEGV, Segmentation fault.
[...]

Issue #8524, opened 2022-06-30
MinGW compilation, patch required in SDL


psfree:
The following patch is required to compile yuzu on mingw:

``` 
+++ /yuzu/externals/SDL/src/joystick/windows/SDL_windows_gaming_input.c
@@ -34,6 +34,11 @@
 #include <cfgmgr32.h>
 #include <roapi.h>
 
+#ifdef __MINGW32__
+#define __FIReference_1_int __FIReference_1_INT32
+#define __FIReference_1_int_get_Value __FIReference_1_INT32_get_Value
+#define __FIReference_1_int_Release __FIReference_1_INT32_Release
+#endif

struct joystick_hwdata
 {
```

This can be fixed by updating the SDL repo past https://github.com/libsdl-org/SDL/commit/d2c4d74dd0302369251dce9375b49f394e54f76a

lat9nq:
Can you be more specific with regards to "required to compile ... on mingw"? Like do you have issues using MSYS2, or is this for cross-compiling on some system other than Fedora or Arch Linux? Or is this resolving some runtime issue? Asking because we regularly check compiling yuzu on MinGW with and without the bundled SDL2 sources.

psfree:
Using msys2 on windows starting from fresh and following the Building For Windows steps listed.

`[ 51%] Building CXX object externals/dynarmic/src/dynarmic/CMakeFiles/dynarmic.dir/frontend/A64/translate/impl/floating_point_conditional_compare.cpp.obj
[ 51%] Building C object externals/SDL/CMakeFiles/SDL2-static.dir/src/haptic/windows/SDL_dinputhaptic.c.obj
C:/msys64/home/USERNAME/yuzu/externals/SDL/src/joystick/windows/SDL_windows_gaming_input.c: In function 'WGI_JoystickOpen':
C:/msys64/home/USERNAME/yuzu/externals/SDL/src/joystick/windows/SDL_windows_gaming_input.c:641:13: error: unknown type name '__FIReference_1_int'; did you mean '__FIReference_1_INT32'?
  641 |             __FIReference_1_int *full_capacityP, *curr_capacityP;
      |             ^~~~~~~~~~~~~~~~~~~
      |             __FIReference_1_INT32
In file included from C:/msys64/mingw64/include/windows.gaming.input.h:622,
                 from C:/msys64/home/USERNAME/yuzu/externals/SDL/src/joystick/windows/SDL_windows_gaming_input.c:33:
[...]

PR #8607, opened 2022-07-19
externals: Revert SDL2 to release-2.0.20

lat9nq:
`prerelease-2.23.1` appears to have issues on the SteamDeck with external controllers. Revert to 2.0.20 for now (and as opposed to using `prerelease-2.0.19` like before.)

@Beanow if you're comfortable testing an AppImage, I'd appreciate if you could test the PR artifact when it's available with regards to #8606. #8524 will need to be reopened after merge.

If this is merged: @psfree and other MSYS2 users, you likely want to use MSYS2's own SDL2 package. Install `mingw-w64-x86_64-SDL2` and then set `-DYUZU_USE_EXTERNAL_SDL2=OFF -DYUZU_USE_BUNDLED_SDL2=OFF` in your CMake parameters (I should probably consolidate those options in the future...)


german77 (approved):

Beanow:
Looking at the manifest, neither of the build options you mentioned are present.
https://github.com/flathub/org.yuzu_emu.yuzu/blob/e8fef25ed20d303694151b7d5cd938590ba3cc8d/org.yuzu_emu.yuzu.json#L212

I have no idea if, whatever the default is should work well with Flatpak :shrug:

german77:
Your report confirms that SDL 2.23.1 was faulty. It was added on 1093 and removed on 1096

Docteh:
@Beanow that flatpak appears to be for Linux, while the only build options I'm seeing discussed are: 
-DYUZU_USE_EXTERNAL_SDL2=OFF -DYUZU_USE_BUNDLED_SDL2=OFF  

with a suggestion to use them with [MSYS2](https://www.msys2.org/) a build system for windows.

PR #8596, opened 2022-07-16
package MSVC CI Builds differently, and include yuzu.exe

Docteh:
This is related to #8486

    Ninja places the exe files into .\build\bin while MSBuild may place them
    into .\build\bin\Release

    upload.ps1 was originally written for use with Azure Dev Ops to cough up about
    5 files and the script appears to be used for both CI and mainline builds

    GHA (GitHub Actions) makes available a single zip of the items uploaded by
    each Upload action (artifacts directory), so we want to work with that.

    I'm doing changes to upload.ps1 to accomplish this.

    The changes to the verify.yml are as follows

    -DGIT_BRANCH=pr-verify changes the header in yuzu, instead of saying
    HEAD-<hash>-dirty it'll say pr-verify-<hash>

    -DCLANG_FORMAT_SUFFIX=discordplzdontclang tricks the CMake stuff for
    discord-rpc to NOT run clang-format, as this was marking CI builds as
    dirty

    I'm also making it upload just the exe by itself, as the msvc builds are
    quite chunky. but maybe this is unnecessary.

    Currently the MSVC artifact option is a 274MB zip that contains 3 copies
    of the DLLs, and 4 copies of the source tarball, and zero copies of yuzu.exe

    This PR should have msvc artifacts of about 190MB that downloads as 81 MB zip

---

This will fix the missing yuzu.exe as ninja puts the .exe into build\bin\ instead of build\bin\Release

Fixes #8595

PR #8624, opened 2022-07-22
ci,CMake: Drop Conan support for vcpkg

lat9nq:
Between packages breaking, Conan always being a moving target for minimum required CMake support, and now their moving to Conan 2.0 causing existing packages to break, I suppose this was a long time coming. vcpkg isn't without its drawbacks, but at the moment it seems easier on the project to use for external packages.

Mostly removes the logic for Conan from the root CMakeLists file, leaving basic find_package()'s in its place. Sets only the find_package()'s that require CONFIG mode as necessary. clang and linux CI now use the vcpkg toolchain file configured in the Docker container when possible.

mingw CI turns off YUZU_TESTS because there's no way on the container to run Windows executables on a Linux host anyway, and it's not easy to get Catch2 there.

Requires yuzu-emu/build-environments#57

Please wait until I get the MSVC CI working before merge. I don't have a better way to test my changes on MSVC than to do it on a live PR.


.ci/scripts/clang/docker.sh
```diff
@@ -6,7 +6,18 @@ set -e
 ccache -s
 
 mkdir build || true && cd build
-cmake .. -GNinja -DDISPLAY_VERSION=$1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/lib/ccache/clang -DCMAKE_CXX_COMPILER=/usr/lib/ccache/clang++ -DYUZU_ENABLE_COMPATIBILITY_REPORTING=${ENABLE_COMPATIBILITY_REPORTING:-"OFF"} -DENABLE_COMPATIBILITY_LIST_DOWNLOAD=ON -DUSE_DISCORD_PRESENCE=ON -DENABLE_QT_TRANSLATION=ON -DCMAKE_INSTALL_PREFIX="/usr"
+cmake .. \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_CXX_COMPILER=/usr/lib/ccache/clang++ \
+      -DCMAKE_C_COMPILER=/usr/lib/ccache/clang \
+      -DCMAKE_INSTALL_PREFIX="/usr" \
+      -DCMAKE_TOOLCHAIN_FILE=${VCPKG_TOOLCHAIN_FILE} \
```
Docteh, line 14: Is this an environment variable that needs to be passed into the docker? Just asking because I'm looking at what appears to be other variables not passed into docker.

Docteh: Nevermind I see it set in the dockerfile

Docteh:
On the topic of MSVC CI Builds, for #8596 I actually tested things on my own account with a modified `verify.yml` that only contained the MSVC build. 

https://github.com/Docteh/yuzu/commit/395bf40aac9107dbd5031f23d144806fe22f13b8

More of a suggestion if you later find yourself banging your head against something exclusive to a single build.

lat9nq:
Developers: to use vcpkg, you can enable the one included now with yuzu via `-DYUZU_USE_BUNDLED_VCPKG=ON` in your CMake options. Otherwise, you can install it yourself and provide your own with `-DCMAKE_TOOLCHAIN_FILE=[vcpkg root]/scripts/buildsystems/vcpkg.cmake`.

Alternatively, you could just have the necessary packages installed on your system as that should work better (e.g. `find_package(lz4)` should now work more consistently for MinGW).

bunnei (approved):

Docteh:
when building the source tarball for MSVC we might want to exclude a few directories

externals/vcpkg/buildtree
externals/vcpkg/packages
externals/vcpkg/downloads

    source                173,186
    get rid of buildtree:  94,195 
    get rid of packages:   70,562
    get rid of downloads:  31,671

Tachi107:
Just curious, what's the advantage in using a package manager like conan or vcpkg as opposed to using externals?

Docteh:
With conan there are builds of the items, not sure about Linux, but for windows they had prebuilt files for a bunch of version of visual studio, not 2022 yet. I think with vcpkg even though it doesn't cache binaries, I think the project could do it. 

I have heard that builds used to compile the entirety of boost, right now it looks like only as much as needed is compiled.

Tachi107:
I know this has been merged recently, but CMake 3.24 has gained a lot of new features that, here I say it, may make it less painful to use for dependency discovery.

For example, the FetchContent module can now mimic one aspect of Meson's `dependency()`, as it will be now possible to make it try to first look for the dependency on the system with `find_package()` and fall back to downloading it if not found; this can also be configured on the CMake command line, so it is possible to only use bundled or system dependencies if desired.

Support for third party package managers has been greatly improved too, so this might help with yuzu's vcpkg usage.

I've looked at this only briefly, but from what I've read this looks like a great improvement!

As yuzu makes heavy use of a mix of system dependencies, a package manager, and bundled "externals" I believe that CMake 3.24 (not released yet) could really be interesting for the project.

You can find all the references in CMake 3.24's changelog: https://cmake.org/cmake/help/latest/release/3.24.html

PR #8635, opened 2022-07-24
build: override git description on mainline builds

Docteh:
Note: this won't do anything to Early Access builds

right now it appears that the output of like:
git describe  --long --always --dirty
is something like: mainline-636-9139-g0ec190c16-dirty

and the title for yuzu-cmd: HEAD-mainline-636-9139-g0ec190c16-dirty

the goal here is to change this title to:
mainline-<build id>-<shorthash>

so as an example trying for: mainline-1107-0ec190c16

dirty flag on github actions builds are due to discord-rpc
and its desire to run clang-format on itself as part of
compiling. that library is retired and deprecated, so no updates.

Also remove scm_rev.cpp from getting shipped as an artifact. (oops, but very useful)

--------------

$ ./yuzu-cmd -v

1105: yuzu HEAD mainline-636-9124-geb14b2edf-dirty
1107: yuzu HEAD mainline-636-9139-g0ec190c16-dirty
this pr: yuzu mainline 69420-b3e675102

image

image


Docteh:
I could use some input on what to do for Early Access. Maybe in CMake REPO_NAME can be looked at, or something else?

(some random log in #yuzu-support-windows)
[   0.260469] Frontend <Info> yuzu\main.cpp:GMainWindow:300: yuzu Version: yuzu Early Access 2849

https://github.com/yuzu-emu/yuzu/blob/1e67d2b59f6dfd561768db3fb9a8e0c6a16ec9f2/src/yuzu/main.cpp#L299-L303

hopefully its not checking TITLE_BAR_FORMAT_IDLE if its "yuzu Early Access {}"

Docteh:
I bothered Morph into looking at the azure devops side of things.

liamwhite:
Any updates on the ADO changes?

Docteh:
I haven't heard back on it. Was hoping to either hear or see improvements.

Currently for yuzu-cmd on windows:
1194
image

yuzu HEAD mainline-636-9611-g01a2ae888

I'd like it to be like mainline-1194-01a2ae888 (I just realized I mention this in the first comment)

or maybe mainline-1194-gh-01a2ae888 if noting that the hash is from github is important

Tachi107:
> or maybe mainline-1194-gh-01a2ae888 if noting that the hash is from github is important

I think that using `git` instead of `gh` is a more common practice (also because the commit hash is unrelated to GitHub in particular)

PR #8640, opened 2022-07-25
cmake,ci: vcpkg leftovers

lat9nq:
I figured out how to get a specific version of Catch2 installed with vcpkg using their manifest mode, which was one of last major pains from #8624. This allows `YUZU_TESTS` to be enabled with vcpkg in use, and thus I'm much more comfortable enabling vcpkg usage by default. For now I'm only going to do that for MSVC since on MinGW and Linux it is much easier to get required dependencies installed on the system there.

In addition this fixes lz4 woes when building in Debug mode with MSVC by `find_package`'ing lz4 in CONFIG mode first. A major drawback from this approach is vcpkg's `lz4-config.cmake` lacks a version string for us to require. This may need to be brought up as an issue on vcpkg's repository.


CMakeLists.txt
```diff
@@ -157,7 +161,6 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 # =======================================================================
```
Docteh, line 161: The line directly above this
`# System imported libraries`

possibly another silly suggestion, but I think that line should mention vcpkg. 

Two separate suggestions

`# System Imported libraries (optionally from vcpkg)`

`# System/vcpkg imported libraries`

bunnei (approved):

Morph1984 (approved):

PR #8641, opened 2022-07-25
build: Ship vcpkg dlls with MSVC pr-verify builds

Docteh:
With our recent switchover from conan to vcpkg, we're shipping a few
more dll files, these need to be in the full zip.

    cp .\build\bin\*.dll .\artifacts\

also tacking on the fix where we're shipping scm_rev.cpp accidentally


german77 (approved):
Finally the builds from msvc work

PR #8647, opened 2022-07-26
Linux: handle dark system themes nicely

### LINKED ISSUE #8643: [Linux] Dark text on dark system themes
I am using the yuzu flatpak under KDE(wayland) with the breeze dark theme(although i do think this may also affect other desktops too without having tested them)
These buttons do not change their color to a light variant under a dark theme making them unreadable.

Docteh:
yuzu's default theme doesn't specify everything, which is fine for
windows, but in linux anything unspecified is set to the users theme.

Symptoms of this are that a linux user with a dark theme won't think
to change the theme to a dark theme when first using yuzu

Idea here is to try and support arbitrary themes on linux.

preliminary work on a "default_dark" theme, used only as overlay
for any themes that are measured to be dark mode.

Other work done:

FreeDesktop standard icon names:
plus -> list-add
delete refresh, we use view-refresh

remove duplicated icons for qdarkstyle_midnight_blue
referencing icon aliases in the qrc files is the way to go

Note:
Dynamic style changing doesn't appear to work with AppImage

------

I'm not sure where to put the GMainWindow::check_dark_mode function,  maybe it belongs lower in the file, same with location of the additions to main.h

Long term the themes should be split into two parts, regular Qt theme, and yuzu specific

fixes #8643


Docteh:
image

Looks like on Linux we're grabbing icons from the theme, which is fine, but there is no fallback to explicitly grab icons from the program. #8339 might be of reference

When showing the gamelist I see two errors about `QPixmal::scaled: Pixmap is a null pixmap` Is one or both the plus icon, time will tell.

Docteh:
#7755 is also of reference. Turns out the default is to use the colorful icons, nobody told the UpdateUI function I've been staring at, if given an invalid theme it'll actually fall back to default instead of colorful/colorful_dark


AppImage on `Default`, note the total lack of icons.
image


If I don't actually set the icon theme to "default" when using "default"
image

Path A: just ignore this for now. For QIcon with its themeName and themeSearchPaths it looks like it finds 

Path B: conform to FreeDesktop icon naming, because QIcon::fromTheme actually intends to get icons from a theme

star -> rating, example from oxygen
rating
or maybe user-bookmarks

plus -> either folder-new or list-add
chip -> cpu
sd_card -> media-flash

https://commons.wikimedia.org/wiki/Comparison_of_icon_sets

Path C: Essentially re-implement a form of QIcon::fromTheme that runs after checking QIcon::hasThemeIcon

Path D: Leaning towards this one, just use yuzu's icons, and ignore the desktop icons.


```cpp
// In src/yuzu/main.cpp GMainWindow::UpdateUITheme
    QIcon::setThemeName(current_theme);
    QIcon::setThemeSearchPaths(QStringList(QStringLiteral(":/icons")));
```

TL;DR of this comment, looking at QIcon::fromTheme I don't believe we can use this function to grab icons from the users theming, and grab icons from internal themes at the same time.

Docteh:
@Tachi107 I've been meaning to ask, if I make a copy of say dist/qt_themes/default/style.qss do I mark leave the year 2018 in the copy, or make it 2022?

Tachi107:
@Docteh I believe that if you make a copy the same _copy_-right year applies :)

src/yuzu/main.cpp
```diff
@@ -3969,9 +4003,6 @@ void GMainWindow::UpdateUITheme() {
         new_pal.setColor(QPalette::Link, QColor(0, 140, 200, 255));
     }
     qApp->setPalette(new_pal);
```
Docteh, line 4005: need to not setPalette if reacting to a changeEvent (loops)

Docteh:
image
Appllication Style: Breeze
Icons: Buff-plasma
Color Scheme: DarkRed_0-4

Docteh:
In the last image, DOCKED fell back to the text color, but NEAREST and NO AA are grey. I've found that colors can be specified to be `palette(text)` for consistency. Also grabbed the brighter colors for the VULKAN/OPENGL and GPU HIGH/GPU NORMAL
image

image
I'm going to wait for the AppImage to build, see if the dotted lines in the QTreeView are there. Otherwise this is a sensible image of the PR, showing a mainline build vs what the PR does.

Docteh:
AppImage generated by this PR
image

src/yuzu/main.h
```diff
@@ -392,6 +393,10 @@ private slots:
     QTimer mouse_hide_timer;
     QTimer mouse_center_timer;
 
+    QString startup_icon_theme;
+    bool os_dark_mode = false;
+    bool check_dark_mode();
```
liamwhite, line 398: CheckDarkMode

PR #8649, opened 2022-07-26
common: Use PROJECT_SOURCE_DIR to find CMakeModules

lat9nq:
Fixes CMake configuration when yuzu is a submodule of another project.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -16,7 +16,7 @@ find_package(Git QUIET)
 
 add_custom_command(OUTPUT scm_rev.cpp
     COMMAND ${CMAKE_COMMAND}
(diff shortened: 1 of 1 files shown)


Tachi107:
Wouldn't it be better to use [`PROJECT_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html)?

lat9nq:
> Wouldn't it be better to use [`PROJECT_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html)? It always refers to the most recent `project()` call.

Thanks!

Tachi107:
I think you forgot to commit the actual code changes - you've only updated the commit message :)

Morph1984 (approved):

PR #8650, opened 2022-07-26
[Coretiming/NVNFlinger] Improve multi-core vsync timing, and core timing accuracy

Kelebek1:
This re-works the vsync timing a bit for multi-core. After the core timing changes to make it more accurate, we can now use it for vsync flipping as well, which is an improvement over the current sleep method. 

Games in multicore should now be smoother, some quite noticeably (XC2), and can also boost FPS as well (again noticeably XC2), and also eliminates the time drift in Smash. 

Also fixes games that get locked to 58 FPS rather than 60, such as Three Hopes. Images courtesy of Dunastique (the frametime bar at the bottom):
Before: 
unknown2

After:
 unknown

### DIFF
--- a/src/core/core_timing.cpp
+++ b/src/core/core_timing.cpp
@@ -243,16 +243,16 @@ std::optional<s64> CoreTiming::Advance() {
             basic_lock.lock();
 
             if (evt.reschedule_time != 0) {
-                // If this event was scheduled into a pause, its time now is going to be way behind.
-                // Re-set this event to continue from the end of the pause.
-                auto next_time{evt.time + evt.reschedule_time};
-                if (evt.time < pause_end_time) {
-                    next_time = pause_end_time + evt.reschedule_time;
-                }
-
                 const auto next_schedule_time{new_schedule_time.has_value()
                                                   ? new_schedule_time.value().count()
                                                   : evt.reschedule_time};
+
+                // If this event was scheduled into a pause, its time now is going to be way behind.
+                // Re-set this event to continue from the end of the pause.
+                auto next_time{evt.time + next_schedule_time};
+                if (evt.time < pause_end_time) {
+                    next_time = pause_end_time + next_schedule_time;
+                }
 
                 event_queue.emplace_back(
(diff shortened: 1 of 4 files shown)


src/core/core_timing.cpp
```diff
@@ -278,11 +281,23 @@ void CoreTiming::ThreadLoop() {
...
+                    // Only try to sleep if the remaining time is >= 1ms. Take off 500 microseconds
+                    // from the target time to account for possible over-sleeping, and spin the
+                    // remaining.
+                    const auto sleep_time_ns{diff_ns - 500LL * 1'000LL};
+                    const auto sleep_time_ms{sleep_time_ns / 1'000'000LL};
+                    if (sleep_time_ms >= 1) {
+                        event.WaitFor(std::chrono::nanoseconds(sleep_time_ns));
+                    }
+                    const auto end_time{std::chrono::nanoseconds(*next_time)};
+                    while (!paused && GetGlobalTimeNs() < end_time) {
```
liamwhite, line 296: Previously, WaitFor would always stop waiting when a new event was pushed into the heap. When spinning, you need to be sure that you do not wait for longer than you are supposed to if the head of the heap changes.

You might consider having an atomic with the next wait end_time and updating this when you modify the heap, though I'm not sure how it would interact with wait cancellation.

bunnei:
Can we revert https://github.com/yuzu-emu/yuzu/pull/8428 now with these changes? (or does that still break time drift in SSBU?)

liamwhite:
@bunnei we can, but the reversion conflicts with YFC.

Morph1984:
Causes a major regression in Kirby and the Forgotten Land:

Mainline 1115
image

EA 2874
image

This PR
image

Morph1984:
The latest commit resolves the regression above with a ~1 fps drop with sufficient CPU power, and increases CPU demand from 4C to 6C to maintain peak performance

DiegoMadGlez:
> The latest commit resolves the regression above with a ~1 fps drop with sufficient CPU power, and increases CPU demand from 4C to 6C to maintain peak performance

Cpus with lower cores 4 or less will see a performance decrease?

bunnei (approved):

PR #8667, opened 2022-07-29
[SPIRV] Add missed shader defines

Kelebek1:
FSwizzleAdd uses subgroup_local_invocation_id, and ConvertDepthMode uses output_position, but they weren't being set, resulting in some invalid shader opcodes which crashed Vulkan. 

This allows Xenoblade Chronicles 3 to boot with Vulkan. All credits to @liamwhite for this, I was just a guinea pig.

### DIFF
--- a/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
@@ -1306,7 +1306,7 @@ void EmitContext::DefineInputs(const IR::Program& program) {
         subgroup_mask_gt = DefineInput(*this, U32[4], false, spv::BuiltIn::SubgroupGtMaskKHR);
         subgroup_mask_ge = DefineInput(*this, U32[4], false, spv::BuiltIn::SubgroupGeMaskKHR);
     }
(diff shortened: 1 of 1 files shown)


yuwenhua653:
This change does not appear to be effective for AMD GPUs

corang:
MSVC build works for Windows 10 RTX 3090 latest drivers

corang:
> MSVC build works for Windows 10 RTX 3090 latest drivers

There are still crashes that happen the same way, immediate log end and hard crash, but I'm not sure if it's the same kind of issue. There is a specific cutscene that triggers the crash, I'm unsure how to figure out what the issue is since the log doesn't get saved to disk properly.

<details>
  <summary>XC3 Ch 2 Spoiler</summary>
  
  For me it always crashes during the `Consul K` cutscene in Chapter 2
</details>

Starcrater21:
> This change does not appear to be effective for AMD GPUs

Yea, When trying this out, the issue seems to be a Exception Access Violation, 

"Unhandled exception at 0x00007FFD3648C51D (ntdll.dll) in yuzu.exe: 0xC0000005: Access violation accessing location 0x0000000000000000."

necuil:
There is still crashing, which will occur when playing some CGs

liamwhite (approved):

PR #8676, opened 2022-07-30
cmake: add option to check submodules

abouvier:
Better than deleting `.gitmodules` manually :p


Docteh:
What does this help with?

abouvier:
When you non-recursively clone the repo to build yuzu, because you use system libraries instead of the externals. The build fails if you don't delete `.gitmodules` before compiling.

liamwhite (approved):

PR #8681, opened 2022-07-30
tidy up github actions

Docteh:
ci.yml: REUSE task should only run on yuzu-emu/yuzu repository, not on forks.
verify.yml: #8659 is a better fix to the discord-rpc dirty flag

----------

I only need one email if I push something and it fails REUSE ;)


Tachi107:
Fair enough, REUSE checks are pretty annoying in forks

Morph1984 (approved):

PR #8682, opened 2022-07-30
yuzu qt: Add option to create Windows crash dumps

lat9nq:
This makes it possible for layman on Windows to produce a stack trace given any Windows build of yuzu. MinGW builds can do this to some capacity, but the stack trace isn't as reliable there as it is with an MSVC build.

To do this, yuzu will spawn an additional process at the start. The child process becomes what the user sees and interacts with (the yuzu interface we're used to). The parent now sidetracks into a debugging loop and debugs the child. When the child throws a twice unhandled exception, a crash dump is produced in the same directory as the yuzu executable.

Performance is worse as a result of yuzu being debugged, so this is behind an option in the debug settings. Due to the nature of the setting, yuzu has to restart in order to apply or clear the setting. This is covered in a dialog to notify the user when they activate it.


src/yuzu/mini_dump.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <windows.h>
+
+#include <dbghelp.h>
+
```
lioncash, line 6: These functions should probably have a namespace around them

src/yuzu/mini_dump.cpp
```diff
@@ -0,0 +1,195 @@
...
+            //     for. Then when it reaches here again, yuzu will probably crash.
+            // - DBG_CONTINUE on an exception that the debuggee does not handle can set us up for an
+            //     infinite loop of exceptions.
+            ContinueDebugEvent(deb_ev.dwProcessId, deb_ev.dwThreadId, DBG_EXCEPTION_NOT_HANDLED);
+            break;
+        }
+    }
+}
+
+const char* ExceptionName(DWORD exception) {
```
lioncash, line 146: This is only used by the above function, so this can be moved above it and be made internally linked with static

src/yuzu/mini_dump.cpp
```diff
@@ -0,0 +1,195 @@
...
+    case EXCEPTION_STACK_OVERFLOW:
+        return "EXCEPTION_STACK_OVERFLOW";
+    case EXCEPTION_INVALID_DISPOSITION:
+        return "EXCEPTION_INVALID_DISPOSITION";
+    case EXCEPTION_GUARD_PAGE:
+        return "EXCEPTION_GUARD_PAGE";
+    case EXCEPTION_INVALID_HANDLE:
+        return "EXCEPTION_INVALID_HANDLE";
+    default:
+        return nullptr;
```
lioncash, line 193: It's better here to return `"Unknown exception type"`, which makes every exit path safe for formatting

src/yuzu/mini_dump.cpp
```diff
@@ -0,0 +1,195 @@
...
+    char file_name[255];
+    const std::time_t the_time = std::time(nullptr);
+    std::strftime(file_name, 255, "yuzu-crash-%Y%m%d%H%M%S.dmp", std::localtime(&the_time));
+
+    // Open the file
+    HANDLE file_handle = CreateFileA(file_name, GENERIC_READ | GENERIC_WRITE, 0, nullptr,
+                                     CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
+
+    if (file_handle == nullptr || file_handle == INVALID_HANDLE_VALUE) {
+        std::fprintf(stderr, "CreateFileA failed. Error: %d\n", GetLastError());
```
lioncash, line 23: The fprintf calls can be converted over to `fmt::print`

src/yuzu/mini_dump.cpp
```diff
@@ -0,0 +1,195 @@
...
+void DumpFromDebugEvent(DEBUG_EVENT& deb_ev, PROCESS_INFORMATION& pi) {
+    EXCEPTION_RECORD& record = deb_ev.u.Exception.ExceptionRecord;
+
+    HANDLE thread_handle = OpenThread(THREAD_GET_CONTEXT, false, deb_ev.dwThreadId);
+    if (thread_handle == nullptr) {
+        std::fprintf(stderr, "OpenThread failed (%d)\n", GetLastError());
+    }
+
+    // Get child process context
+    CONTEXT context;
```
lioncash, line 52:
```suggestion
    CONTEXT context = {};
```

Will let you get rid of the memset below

src/yuzu/mini_dump.cpp
```diff
@@ -0,0 +1,195 @@
...
+    // Close the file
+    CloseHandle(file_handle);
+}
+
+void DumpFromDebugEvent(DEBUG_EVENT& deb_ev, PROCESS_INFORMATION& pi) {
+    EXCEPTION_RECORD& record = deb_ev.u.Exception.ExceptionRecord;
+
+    HANDLE thread_handle = OpenThread(THREAD_GET_CONTEXT, false, deb_ev.dwThreadId);
+    if (thread_handle == nullptr) {
+        std::fprintf(stderr, "OpenThread failed (%d)\n", GetLastError());
```
lioncash, line 48: Shouldn't this branch return if the thread handle can't be successfully acquired?

src/yuzu/CMakeLists.txt
```diff
@@ -208,6 +208,16 @@ add_executable(yuzu
     yuzu.rc
 )
 
+if (WIN32 AND YUZU_CRASH_DUMPS)
+    target_sources(yuzu PRIVATE
+        mini_dump.cpp
+        mini_dump.h
+    )
+
+    target_link_libraries(yuzu PUBLIC ${DBGHELP_LIBRARY})
```
lioncash, line 217:
```suggestion
    target_link_libraries(yuzu PRIVATE ${DBGHELP_LIBRARY})
```

lat9nq: Thank you!

lat9nq:
The dbghelp header and the Windows 10 SDK minidumpapiset.h header have mismatched opening/closing braces and cannot be compiled there. Need to wait for the Azure pipelines to update before trying again.

Morph1984 (approved):

PR #8686, opened 2022-07-31
ci: Graft source when archiving

lat9nq:
Instead of including yuzu and all the sources it uses directly, include only what specifically belongs to yuzu. Submodules can be downloaded separately later using git since a shallow clone includes minimally all the repository information needed for it.

Shrinks the source archive down from about 32 MB at time of writing to about 8 MB, both numbers after XZ compression.

### DIFF
--- a/.ci/scripts/common/post-upload.sh
+++ b/.ci/scripts/common/post-upload.sh
@@ -8,7 +8,8 @@ cp LICENSE.txt "$DIR_NAME"
 cp README.md "$DIR_NAME"
 
 if [[ -z "${NO_SOURCE_PACK}" ]]; then
-  tar -cJvf "${REV_NAME}-source.tar.xz" src externals CMakeLists.txt README.md LICENSE.txt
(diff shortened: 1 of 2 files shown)


Docteh:
should source be a generated artifact from the verification builds? 

also for the MSVC build, could the source be put into a 7z format? asking for maximum compression saves a little bit more

```
defaults:
-a----         7/31/2022   1:53 AM        7926258 yuzu-butts.7z
-a----         7/31/2022   1:52 AM       38705152 yuzu-butts.tar
-a----         7/31/2022   1:52 AM        7955576 yuzu-butts.tar.xz
-a----         7/31/2022   1:53 AM       11210157 yuzu-butts.zip

maximum compression
7z a -r -bt -mx9 -t7z yuzu-butts-source.7z yuzu-butts
-a----         7/31/2022   2:00 AM        7771092 yuzu-butts-source.7z
```

Morph1984 (approved):

PR #8724, opened 2022-08-06
hid: core: Properly emulate controller color and battery level

german77:
~~The invested time to figure out this issue can't be correlated to the line changed. Qt provided alpha values for the controller color. The switch doesn't have alpha values for the controllers.~~

Properly reports battery and controller colors depending on the controller type.

Fixes Mario Party Superstars displaying the wrong controller icon.
image

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -84,17 +84,18 @@ void EmulatedController::ReloadFromSettings() {
         motion_params[index] = Common::ParamPackage(player.motions[index]);
     }
 
+    controller.colors_state.fullkey = {
+        .body = GetNpadColor(player.body_color_left),
+        .button = GetNpadColor(player.button_color_left),
+    };
     controller.colors_state.left = {
-        .body = player.body_color_left,
-        .button = player.button_color_left,
+        .body = GetNpadColor(player.body_color_left),
+        .button = GetNpadColor(player.button_color_left),
     };
-
-    controller.colors_state.right = {
-        .body = player.body_color_right,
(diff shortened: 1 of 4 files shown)


src/core/hid/emulated_controller.cpp
```diff
@@ -84,18 +84,57 @@ void EmulatedController::ReloadFromSettings() {
         motion_params[index] = Common::ParamPackage(player.motions[index]);
     }
 
+    controller.colors_state.fullkey = {
+        .body =
+            {
+                .b = static_cast<u8>((player.body_color_left >> 16) & 0xFF),
```
FearlessTobi, line 90: Is there any way we can get rid of the duplicated code and manual bitshifting here?

bunnei (approved):

src/core/hid/emulated_controller.cpp
```diff
@@ -84,18 +84,19 @@ void EmulatedController::ReloadFromSettings() {
...
-        .body = player.body_color_left,
-        .button = player.button_color_left,
+        .body = GetNpadColor(player.body_color_left),
+        .button = GetNpadColor(player.button_color_left),
     };
-
-    controller.colors_state.right = {
-        .body = player.body_color_right,
-        .button = player.button_color_right,
+    controller.colors_state.left = {
```
psxlover, line 95: Is this supposed to modify .left? If yes, then the assignment above seems unnecessary, but I would expect that this is supposed to be right.
In fact I don't see why this whole block was modified (apart from adding `GetNpadColor()`, assigning left to fullkey at the end instead of duplicating the code seemed cleaner.

Issue #7310, opened 2021-11-10
Feature Request: Double buffering, Triple buffering or V-sync equilavent for Vulkan


DelghLay:
Feature Request: Double buffering, Triple buffering or V-sync equilavent for Vulkan.

This one will helps to reduce frame skipping to create a more smooth playing experience and the most important thing, reduce excesive screen tearing.
OpenGL has his own exclusive OpenGL V-sync into Yuzu Emulator, why Vulkan couldn't have something similar.

Forcing v-sync from Nvidia Control Panel doesn't solve those problems.

Thanks you all folks!

BreadFish64:
Vulkan by default uses a "mailbox" swapchain which is equivalent to triple buffering. If that mode isn't available then it uses "immediate" mode which is the same as V-Sync off.

DelghLay:
Sorry i'm ignorant but, there is no way to add an option to always force double or triple buffering on Vulkan? (like Cemu).
Sometimes 60 FPS native games displays as if they were less than 60 fps (losing image smoothness) at the same time that Yuzu shows me 60 fps locked of performance, this doesn't occur on OpenGL (i know, sounds strange).
This problem is not related to shader cache compilation or PC performance problems, the only partial solution i've found is activate Rivatunner's Scaleline Sync at 60 FPS + to run Yuzu's Vulkan, but this mess up OpenGL, it's a bit annoying enable and disable constantly that.

Squall-Leonhart:
Believe this is closeable now?

PR #8735, opened 2022-08-08
Make vsync setting work for Vulkan

djrobx:
The Vulkan renderer does not support setting vsync.   It currently prefers mailbox (Triple buffering) mode if found. 

Mailbox mode avoids tearing, but it has a downside: it does not keep the emulation in sync with the real display.   If the emulation and display are very close to precise sync, there can be fits of juddering when the newly generated frames don't decisively fall as next in the display's sequence.  

Related to issue #7310.   I experienced the same behavior as [DelghLay](https://github.com/DelghLay).   Yuzu shows a rock solid 60fps, but on my setup, the display judders for ~30 seconds or so every few minutes.   After switching the code to use FIFO mode, the issue goes away completely and scrolling games like Metroid Dread and Link's Awakening become butter smooth.   I'm using the existing vsync checkbox so I can still disable it for less performant titles. 

A better fix would be to give access to more vulkan swap mode options.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_swapchain.cpp
+++ b/src/video_core/renderer_vulkan/vk_swapchain.cpp
@@ -33,9 +33,10 @@ VkSurfaceFormatKHR ChooseSwapSurfaceFormat(vk::Span<VkSurfaceFormatKHR> formats)
 }
 
 VkPresentModeKHR ChooseSwapPresentMode(vk::Span<VkPresentModeKHR> modes) {
-    // Mailbox doesn't lock the application like fifo (vsync), prefer it
+    // Mailbox (triple buffering) doesn't lock the application like fifo (vsync),
+    // prefer it if vsync option is not selected
     const auto found_mailbox = std::find(modes.begin(), modes.end(), VK_PRESENT_MODE_MAILBOX_KHR);
-    if (found_mailbox != modes.end()) {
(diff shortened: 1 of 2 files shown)


liamwhite (approved):

Missingphy:
> The Vulkan renderer does not support setting vsync. It currently prefers mailbox (Triple buffering) mode if found.
> 
> Mailbox mode avoids tearing, but it has a downside: it does not keep the emulation in sync with the real display. If the emulation and display are very close to precise sync, there can be fits of juddering when the newly generated frames don't decisively fall as next in the display's sequence.
> 
> Related to issue #7310. I experienced the same behavior as [DelghLay](https://github.com/DelghLay). Yuzu shows a rock solid 60fps, but on my setup, the display judders for ~30 seconds or so every few minutes. After switching the code to use FIFO mode, the issue goes away completely and scrolling games like Metroid Dread and Link's Awakening become butter smooth. I'm using the existing vsync checkbox so I can still disable it for less performant titles.
> 
> A better fix would be to give access to more vulkan swap mode options.

hello, noob question

how do I turn on FIFO mode? thanks

djrobx:
With this commit: turn vsync on to get FIFO, leave vsync off for old behavior (mailbox if available).

goldenx86:
Hi, can I mention you and this PR in the next yuzu progress report?

PR #8745, opened 2022-08-09
arm_dynarmic: Fix nullptr fastmem arenas

merryhime:
Many thanks to @vladkosi for investigating the underlying issue regarding non-functioning on earlier versions of Windows. See: merryhime/dynarmic#701.

On earlier versions of Windows, because they do not support modern memory allocation APIs, the fastmem pointer would be nullptr. Yuzu attempts to enable fastmem of exclusive accesses without a valid fastmem arena, which is not a valid state to be in. Dynarmic thus asserts internally when this happens.

Verification would be helpful as I am unable to test this.


vladkosi:
I can confirm that this has fixed the crashing. Tested on Windows 10 versions 1511 and 1803.

liamwhite (approved):

Issue #4043, opened 2020-06-04
FAST RMX - Missing Graphics In-Game


ActualMandM:
yuzu Mainline 271

GeForce RTX 2080
Ryzen R9 3950X

Dynamic Exposure on:
The exposure make the screen almost entirely white.
<img width="961" alt="2020-06-03_17-38-38" src="https://user-images.githubusercontent.com/15317421/83702365-5c0f5000-a5c1-11ea-9619-787b4f5f4292.png">

Dynamic Exposure off:
While you can see outlines of models in the game, you don't see the textures of them.
<img width="961" alt="2020-06-03_17-38-27" src="https://user-images.githubusercontent.com/15317421/83702372-603b6d80-a5c1-11ea-813f-5e850a5bc764.png">

[yuzu_log.zip](https://github.com/yuzu-emu/yuzu/files/4726842/yuzu_log.zip)

bunnei:
Probably still relevant, no need to close :)

legend800:
Still quite broken - missing shaders, broken lighting and runs at 2fps (on a high-end pc)

![2021-01-13 13_02_41-](https://user-images.githubusercontent.com/20983844/104533527-69fd6880-55c7-11eb-9eae-8485e690d82b.png)
![2021-01-13 13_00_51-yuzu Early Access 1315 _ FAST RMX](https://user-images.githubusercontent.com/20983844/104533559-784b8480-55c7-11eb-946d-ac3ecb46e3d1.png)
![2021-01-13 13_01_53-yuzu Early Access 1315 _ FAST RMX](https://user-images.githubusercontent.com/20983844/104533566-7da8cf00-55c7-11eb-850c-765d2814cd2b.png)

benderscruffy:
still broken on Vulkan and Opengl
![01009510001ca000_2021-12-05_19-00-03-418](https://user-images.githubusercontent.com/2696404/144743872-f47e5f91-6aca-4ff3-bbc0-0d8e1320760c.png)
![01009510001ca000_2021-12-05_19-01-22-045](https://user-images.githubusercontent.com/2696404/144743873-cca7b35a-0d2a-4dc1-9119-a48729a15324.png)
![01009510001ca000_2021-12-05_18-59-13-752](https://user-images.githubusercontent.com/2696404/144743870-c3d5300a-f222-47f4-99ca-3fd1ab7f79a3.png)
[...]

PR #8752, opened 2022-08-11
video_code: support rectangle texture

vonchenplus:
This PR is intended to solve the rectangle texture. Considering that the rectangle texture is a legacy texture format, we use texture2d to emulator, and normalize the rectangle texture coordinates when it create fs shader.

Before:
image

After:
image

### DIFF
--- a/src/shader_recompiler/backend/glasm/emit_glasm_image.cpp
+++ b/src/shader_recompiler/backend/glasm/emit_glasm_image.cpp
@@ -67,6 +67,7 @@ std::string_view TextureType(IR::TextureInstInfo info) {
         case TextureType::ColorArray1D:
             return "SHADOWARRAY1D";
         case TextureType::Color2D:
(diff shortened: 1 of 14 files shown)


gidoly:
breaks xenoblade definitive edition rendering. black screen

BSoDGamingYT:
Screenshot 2022-08-12 151424
Screenshot 2022-08-12 151628

Fast RMX and The Touryst begin to render better with this change (At least on OGL)

bunnei (approved):

german77:
Some users need to clean the pipeline cache manually for games to render again. Are you aware into why? If this can't be avoided can we invalidate them automatically

vonchenplus:
I have fixed the shader cache invalidation issue, please do not delete the cache to rebuild yet.

german77:
closes #4043

FernandoS27 (approved):

PR #8756, opened 2022-08-12
Allow audio volume up to 200%

Kelebek1:
Requested by a couple people, allows the volume slider to go up to 200%. Also changes the default volume up key from + to =. It was weird that for volume down you just press -, but volume up required shift+=. Changing the key from + to = lets you skip the shift, as was originally intended I think.


liamwhite (approved):

german77 (approved):

oxixes:
About the volume up key, on some keyboards the `+` key is just one press without shift (while `=` requires shift), like the spanish keyboard, so maybe that should depend on the keyboard layout.

bunnei (approved):

PR #8795, opened 2022-08-20
video_core: support framebuffer crop rect top not zero

vonchenplus:
Before:
image

After:
image


liamwhite (approved):
Fix clang-format and it should be good to go

PR #8805, opened 2022-08-22
yuzu: Force camera output to be saved on a buffer

german77:
Our current camera implementation have a horrible bug. It saves every single frame into a file. This results in millions of images over a long period of play time often breaking file explorer.

This PR forces the output to be saved in a buffer. This comes at a cost. If camera doesn't support this we no longer support that camera.

Sorry for everyone with this issue. If you are on windows and the file explorer doesn't work please open cmd on that folder and type "del IMG_00*.jpg" after a couple of minutes it will delete all images generated by this feature.


goldenx86 (approved):
LGTM

PR #8104, opened 2022-03-28
Moving Icons away from CC BY-ND 3.0 licensed Icons to help with FOSS Packaging

Docteh:
I've seen some comments stating that sharing pre-compiled binaries
is problematic for linux distributions due to the CC BY-ND 3.0 license

In this commit I am tackling the 16x16 icons

The reported status of the "yuzu Web Service" section of Configuration uses two icons, checked and failed

checked is replaced by material-design-icons check_circle
failed is replaced by material-design-icons error

Colors have been set by taking the white pngs of 18x18, cropping to 16x16, and then adjusting Curves using Paint.NET

Other tools are available.

There is also a locked icon that does not appear to be used in the current code. citra uses it for multiplayer

lock is replaced by material-designs-icons lock

Some of the colorful theme icons are clearly from https://github.com/icons8/flat-color-icons/ which is available MIT licensed

Sources for MIT license:
https://github.com/icons8/flat-color-icons/issues/5
https://github.com/icons8/flat-color-icons/blob/master/package.json

This covers plus_folder, bad_folder, chip, folder

Leaving plus, sd_card and star to need further research/replacement

The first two are user visible, the lock icons aren't used right now. 
image

The lock icons are shoddily replaced to avoid having to test and verify that removal doesn't break anything. 

Fixes #8077

PR #8807, opened 2022-08-23
Qt: work around Qt5's font choice for Chinese (in Windows)

Docteh:
On Windows there are currently two fonts used.

The first, does the Menu, QTreeView and Tooltips
Second is Everything else which is a default font.

From inspecting QApplication::font() at runtime
Windows 10 English: QFont(MS Shell Dlg 2,8.25,-1,5,50,0,0,0,0,0)
Windows 11 Japanese:        MS UI Gothic,9   ,-1,5,50,0,0,0,0,0
Windows 11 Traditional Chinese: PMingLiU,9   ,-1,5,50,0,0,0,0,0
Windows 11 Simplified Chinese:    SimSun,9   ,-1,5,50,0,0,0,0,0
Windows 11 Korean:                 Gulim,9   ,-1,5,50,0,0,0,0,0

I initially investigated dynamically changing the font when
the UI language is English, but this was getting quite messy

Qt6 makes changes to default font in some situations, so this
PR is being narrowed in scope to only effect Chinese font choices.
This change only effects rendering of Latin/Cyrillic characters.

----

Comparison of the font choices is in this comment: https://github.com/yuzu-emu/yuzu/pull/8807#issuecomment-1229195994

I was originally going to submit this PR using Comic Sans MS,
but at size 9 it was starting to not be noticeable.

This fixes an issue where between Windows and Qt, a font is being
chosen that looks worse than the rest of the OS.

--------------

IBM Plex Mono is most likely the wrong font

TODO
Right click menus
Tooltips
The about box

IBM Plex Mono, English
image

`\` Character is Japanese yen symbol in that default font.
image

Most replacements we would provide would fix this.
image

----
Extra notes
RFN has been removed in Fira fonts, https://github.com/mozilla/Fira/pull/219 in/around May 2019

TTF fonts seem to be able to hold variants in the same TTF, but this is different from a TTC according to `file`

https://github.com/bluescan/proggyfonts <--- MIT

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -261,6 +261,18 @@ static QString PrettyProductName() {
     return QSysInfo::prettyProductName();
 }
 
+#ifdef _WIN32
+static void OverrideWindowsFont() {
+    // Qt5 chooses these fonts on Windows and they have fairly ugly alphanumeric/cyrllic characters
+    // Asking to use "MS Shell Dlg 2" gives better other chars while leaving the Chinese Characters.
+    const QString startup_font = QApplication::font().family();
+    const QStringList ugly_fonts = {QStringLiteral("SimSun"), QStringLiteral("PMingLiU")};
+    if (ugly_fonts.contains(startup_font)) {
+        QApplication::setFont(QFont(QStringLiteral("MS Shell Dlg 2"), 9, QFont::Normal));
+    }
+}
+#endif
+
 bool GMainWindow::CheckDarkMode() {
 #ifdef __linux__
     const QPalette test_palette(qApp->palette());
@@ -4133,6 +4145,10 @@ int main(int argc, char* argv[]) {
     // Enables the core to make the qt created contexts current on std::threads
     QCoreApplication::setAttribute(Qt::AA_DontCheckOpenGLContextThreadAffinity);
     QApplication app(argc, argv);
+
+#ifdef _WIN32
+    OverrideWindowsFont();
+#endif
 
     // Workaround for QTBUG-85409, for Suzhou numerals the number 1 is actually \u3021
     // so we can see if we get \u3008 instead


src/yuzu/main.cpp
```diff
@@ -269,6 +270,18 @@ bool GMainWindow::CheckDarkMode() {
 #endif // __linux__
 }
 
+static void font_menu_items(QList<QAction *> the_actions) {
```
liamwhite, line 273: maybe SetFontRecursive

src/yuzu/main.cpp
```diff
@@ -269,6 +270,15 @@ bool GMainWindow::CheckDarkMode() {
 #endif // __linux__
 }
 
+static void font_menu_items(QList<QAction*> the_actions) {
+    foreach (QAction* action, the_actions) {
```
lioncash, line 274:
```suggestion
    for (QAction* action : the_actions) {
```

Since there's no need for the Qt-specific macro since C++11

src/yuzu/main.cpp
```diff
@@ -269,6 +270,15 @@ bool GMainWindow::CheckDarkMode() {
...
+        if (action->menu()) {
```
lioncash, line 275: Maybe assign this to a local variable, so that `action->menu()` doesn't need to be called repeatedly like:


```suggestion
        if (auto* menu = action->menu()) {
```

You may also want to leave a comment indicating that this should be switched over to:

```suggestion
        if (auto* menu = QMenu::menuInAction(action)) {
```

when we move to Qt 6, since `menu()` is deprecated in it.

src/yuzu/yuzu.qrc
```diff
@@ -7,4 +7,8 @@ SPDX-License-Identifier: GPL-2.0-or-later
     <qresource prefix="/img">
         <file alias="yuzu.ico">../../dist/yuzu.ico</file>
     </qresource>
+    <qresource prefix="/fonts">
+        <file alias="IBMPlexMono-Italic.ttf">../../dist/fonts/IBMPlexMono-Italic.ttf</file>
+        <file alias="IBMPlexMono-Regular.ttf">../../dist/fonts/IBMPlexMono-Regular.ttf</file>
```
Tachi107, line 12: Is it possible to avoid hardcoding this? Ideally, yuzu should use the system-installed font if available.
For example, on Debian you can install the `fonts-ibm-plex` package to have all variations of the font available.

Tachi107 (commented):
I'm not sure how I feel about this patch. Does this solve an issue?

I think that yuzu using a completely custom font would make it look a bit odd. On my system I expect all apps to use more or less the same font type.

I also fear that a custom font might not play nicely with every possible locale.

Lastly, note that including fonts in yuzu might make the source partially non-free again, effectively nullifying #8104. The RFN clause is particularly [problematic](https://github.com/rsms/inter/issues/282).

Docteh:
My next step on this is to take a look at how it works in linux, I've seen comments that the use of setFont may break picking up theme changes. I wish to see this in action before hiding anything behind `_WIN32`

> IBM Plex Mono is most likely the wrong font
i gotta hit post on this, mainly want you to see the windows screenshots, 

@Tachi107 this where things look ugly on Windows. 

On the main window with default Chinese font, not so bad, because certain widgets QMenuBar, QTreeView, QListView (probably more) don't actually use the default font given on windows. 
image

yuzu English, windows Chinese
image

<details>
 <summary>Other screenshots for context</summary>


yuzu Chinese, windows Chinese
image

yuzu Chinese, windows English
image


</details>

src/yuzu/main.cpp
```diff
@@ -4122,6 +4123,18 @@ int main(int argc, char* argv[]) {
...
+    // MS Shell Dlg 2 is in the list to actually raise the size a bit
+    const QString startup_font = QApplication::font().family();
+    const QStringList ugly_fonts = {QStringLiteral("SimSun"), QStringLiteral("PMingLiU"),
+                                    QStringLiteral("MS UI Gothic"),
+                                    QStringLiteral("MS Shell Dlg 2")};
+    if (ugly_fonts.contains(startup_font)) {
+        QApplication::setFont(QFont(QStringLiteral("MS Shell Dlg 2"), 9, QFont::Normal));
+    }
+#endif
+
```
Docteh, line 4137: I think this should be a function, but naming it is getting set aside.

Docteh:
Qt's default font is being logged in LogRuntimes, I plan on bothering some people to run the verification build, see if there are any other interesting fonts.

Explicitly setting widgets to the default font is limited to `_WIN32`
image

Top is this PR
image

Docteh:
I took a look at Dolphin which uses Qt6, and then did another Qt6 build to confirm, the lack of anti-aliasing in certain languages is fixed in Qt6. I'm minimalizing this PR


1141
image
image


This PR
image
image

Docteh:
I removed the other commit from this PR, and shoved it into #8829 

So this PR is minor tweak to font change (for Qt5), and the other PR contains all the Qt6 stuff.

Docteh:
Cleaned up the comments, did a fresh rebase for October

liamwhite (approved):

Docteh:
I got a note about issues with character representation in Traditional, I suspect its an issue with screenshots posted by me playing loosey goosey with two virtual machines, not a technical problem.

The string being looked at is "Note: Changing language will apply your configuration"
image

PR #8829, opened 2022-08-28
CMake: rework for Qt6 support

Docteh:
This PR rearranges things in the CMake system to make compiling with Qt6 easier

1. Camera API has changed in Qt6, so the camera feature is disabled
2. A previous fix involving QLocale is now version gated.
3. QRegExp replaced with QRegularExpression, see #5343
4. Qt6_LOCATION option added to specify a location to search for Qt6
  (see examples below)
5. windeployqt is used to copy Qt6 files into the build directory on Windows

Notes for Arch Linux
Arch install happened to have qt6-base qt6-declarative qt6-translations installed

mkdir build && cd build
cmake .. -GNinja -DYUZU_USE_BUNDLED_VCPKG=ON -DYUZU_TESTS=OFF -DENABLE_QT6=YES -DYUZU_USE_BUNDLED_QT=NO

Windows (MSVC)
Qt wants users to download precompiled libraries via an online installer, 
it is worth noting that the GPL/LGPL takes precendence over any ...

In the Qt Maintenance tool, under a version, such as 6.3.1
Select "MSVC 2019 64-bit"
Under Additional Libraries Qt Multimedia may be of use for Camera support

For the Web Applet I had to select the following:
PDF Positioning WebChannel WebEngine

mkdir build && cd build
cmake -G "Visual Studio 16 2019" -DQt6_LOCATION=C:/Qt/6.4.0/msvc2019_64/ \
-DENABLE_COMPATIBILITY_LIST_DOWNLOAD=YES -DYUZU_USE_BUNDLED_QT=NO \
-DENABLE_QT_TRANSLATION=YES -DENABLE_QT6=YES ..

Some numbers for reference (msvc2019_64)
```
Qt5 (slimmed down) 508 MB
Qt5.15.2 all in    929 MB
Qt6.3.1           1.71 GB
Qt6.4.0-beta3     1.83 GB
```

--------

As part of bringing in Qt, MSVC builds would need the dll files brought into build/bin/Release and build/bin/Debug

https://www.qt.io/blog/cmake-deployment-api Might be worth a look in regards to handling DLL files


CMakeLists.txt
```diff
@@ -311,18 +313,38 @@ if(ENABLE_QT)
...
     endif()
-    if ((${CMAKE_SYSTEM_NAME} STREQUAL "Linux") AND YUZU_USE_BUNDLED_QT)
-        find_package(Qt5 ${QT_VERSION} REQUIRED COMPONENTS Widgets Concurrent Multimedia DBus ${QT_PREFIX_HINT} ${YUZU_QT_NO_CMAKE_SYSTEM_PATH})
-    else()
-        find_package(Qt5 ${QT_VERSION} REQUIRED COMPONENTS Widgets Concurrent Multimedia ${QT_PREFIX_HINT} ${YUZU_QT_NO_CMAKE_SYSTEM_PATH})
+
+    # Best practice is to ask for all components at once, so they are from the same version
+    set(YUZU_QT_COMPONENTS Widgets Concurrent Multimedia)
+    if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+        list(APPEND YUZU_QT_COMPONENTS DBus)
```
Tachi107, line 320: From the previous code it seems that `DBus` is only needed when `YUZU_USE_BUNDLED_QT` is true, right?

Docteh: DBus is also asked for above when `YUZU_USE_BUNDLED_QT` is false, Line 211 in the left column

Tachi107: Ops sorry, you're right :)

Docteh:
I tested building linux-mingw myself, works, and we (as in not me) can re-run just the one build after https://github.com/yuzu-emu/build-environments/pull/60 is merged.  So marking this ready for review.

CMakeLists.txt
```diff
@@ -311,18 +313,38 @@ if(ENABLE_QT)
...
+    endif()
+    message(STATUS "YUZU_QT_COMPONENTS: " ${YUZU_QT_COMPONENTS})
+
+    if (ENABLE_QT6)
+        find_package(Qt6 COMPONENTS ${YUZU_QT_COMPONENTS} ${YUZU_QT_NO_CMAKE_SYSTEM_PATH})
     endif()
+    if (Qt6_FOUND)
+        if ("${QT_VERSION}" VERSION_LESS_EQUAL "5.15")
+            set(QT_VERSION 6.0.0) # used later in src/yuzu/CMakeLists.txt
+            message(STATUS "yuzu/CMakeLists.txt: Setting QT_VERSION = ${QT_VERSION}")
```
Tachi107, line 336: Is this message needed? I think it's obvious that Qt6 is being used if `find_package(Qt6)` is successfully found

CMakeLists.txt
```diff
@@ -311,18 +313,38 @@ if(ENABLE_QT)
...
+        endif()
+
+        # This may be poorly placed, MUST be done after find_package for Qt6
+        # Without this Qt6 helpfully causes MSVC to generate as cxx_std_17
+        set_target_properties(Qt6::Platform PROPERTIES INTERFACE_COMPILE_FEATURES "cxx_std_23")
```
Tachi107, line 341: Why `cxx_std_23` and not `cxx_std_20`?

CMakeLists.txt
Docteh: I found the CMake system to be very quiet when finding Qt6 successfully or otherwise.  I could drop this one. 

Do you think the other message needs tweaking?
message(STATUS "yuzu/CMakeLists.txt: Qt6 not found/not selected, trying for Qt5")

Docteh:
Changes

1. only setting `cxx_std_23` for MSVC.
2. setting QT_VERSION with Qt6Widgets_VERSION for Qt6
3. When Qt6_LOCATION is specified that path is added to CMAKE_PREFIX_PATH

CMakeLists.txt
Tachi107: > I chose cxx_std_23 as with a Qt5 build the C++ Language Standard shown in MSVC is /std:c++latest, and cxx_std_23 put it back to latest.

OK, but is `/std:c++latest` actually needed?

CMakeLists.txt
Tachi107: Nah it's fine, the important thing is just to tell the user what's happening

CMakeLists.txt
```diff
@@ -311,18 +318,38 @@ if(ENABLE_QT)
...
+        find_package(Qt6 COMPONENTS ${YUZU_QT_COMPONENTS} ${YUZU_QT_NO_CMAKE_SYSTEM_PATH})
+    endif()
+    if (Qt6_FOUND)
+        message(STATUS "yuzu/CMakeLists.txt: Qt6Widgets_VERSION ${Qt6Widgets_VERSION}, setting QT_VERSION")
+        set(QT_VERSION ${Qt6Widgets_VERSION})
+
+        # This may be poorly placed, MUST be done after find_package for Qt6
+        # Without this Qt6 helpfully causes MSVC to generate as cxx_std_17
+        if(MSVC)
+            set_target_properties(Qt6::Platform PROPERTIES INTERFACE_COMPILE_FEATURES "cxx_std_23")
```
Tachi107, line 345: `target_compile_features(Qt6::Platform INTERFACE cxx_std_23)` is a bit more idiomatic.

Anyway the fact that without this the build is done with C++17 is a bit odd.

Doesn't `cxx_std_20` suffice?

I think that the issue is caused by the something else. This is fine as a temporary workaround, but maybe we'll eventually need to find out why it happens and how to fix it properly.

Docteh: This is where Qt sets specifies cxx_std_17
https://github.com/dolphin-emu/ext-win-qt/blob/bd088e81e72243a86cf41db049b62f28d956d4e2/Qt6.3.0/x64/lib/cmake/Qt6/Qt6Targets.cmake#L56-L62
```cmake
set_target_properties(Qt6::Platform PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "WIN32;_ENABLE_EXTENDED_ALIGNED_STORAGE;WIN64;_WIN64;\$<\$<NOT:\$<BOOL:\$<TARGET_PROPERTY:QT_NO_UNICODE_DEFINES>>>:UNICODE;_UNICODE>"
  INTERFACE_COMPILE_FEATURES "cxx_std_17"
  INTERFACE_COMPILE_OPTIONS "\$<\$<COMPILE_LANGUAGE:CXX>:-Zc:__cplusplus;-permissive->;\$<\$<AND:\$<NOT:\$<BOOL:\$<TARGET_PROPERTY:QT_NO_UTF8_SOURCE>>>,\$<COMPILE_LANGUAGE:C,CXX>>:-utf-8>"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/mkspecs/win32-msvc;${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Threads::Threads"
)
```

At  around line 157 we're currently specifying latest for MSVC. `cxx_std_20`  is forced on the verification builds, something relating to caching. But If we're doing cxx_std_23 for Qt5 builds, we should try for it in Qt6 builds as well.
https://github.com/yuzu-emu/yuzu/blob/a83a5d2e4c8932df864dd4cea2b04d87a12c8760/CMakeLists.txt#L157-L158

Tachi107: Then I think that the most appropriate option is setting the `cxx_std_23` compile feature for all the targets and avoid manually fiddling with flags, that (as we are seeing here) can lead to mixed results

Docteh: can I do that in the base CMakeLists.txt or is that editing a bunch of other files?

Tachi107:
You need to edit all the yuzu targets that want to link with C++23. I'd 
do it in a separate commit.

-- 
OpenPGP key: 66DE F152 8299 0C21 99EF  A801 A8A1 28A8 AB1C EE49

Docteh:
Did some digging on MSVC. Code block followed by the result looks potentially confusing to me over on GitHub, but I typed it all out already 🙏 
```cmake
add_compile_options($<$<COMPILE_LANGUAGE:CXX>:/std:c++latest>)
```
This loosely sets latest for `cubeb` `inih` `yuzu`

```cmake
find_package(Qt6 COMPONENTS components here)
Qt6 sets 2017 for `yuzu`
```cmake
set_target_properties(Qt6::Platform PROPERTIES INTERFACE_COMPILE_FEATURES "cxx_std_23")
```
This sets just `yuzu` to latest, after find_package helped
    
```cmake
if(MSVC)
set(CMAKE_CXX_STANDARD 20)
```
This pushes a few things from latest down to 2020 from latest.  video_core common audio_core inih

But I'll be pushing a commit that says 23 to keep things in line with current builds. My understanding is that MSVC is a very special program and is silly sometimes.


Overall looks like we can make soft or hard suggestions, and we should be in favor of more hard suggestions.

Next move is to look up if `set(CMAKE_CXX_STANDARD_REQUIRED ON)` should be done for MSVC

CMakeLists.txt
```diff
@@ -148,11 +149,13 @@ endif()
 # boost asio's concept usage doesn't play nicely with some compilers yet.
 add_definitions(-DBOOST_ASIO_DISABLE_CONCEPTS)
 if (MSVC)
-    add_compile_options($<$<COMPILE_LANGUAGE:CXX>:/std:c++latest>)
```
lioncash, line 151: We should always be setting MSVC to the latest available progress on the next standard and not bolt it down to a single version. This allows us to have a builder available that can catch potential breaking changes (deprecations/removals, etc) regardless of the standard chosen which is beneficial over sticking to a standard, changing over to a new one and potentially breaking a lot of things all at once. It allows things to be incrementally handled as work on standard support progresses.

Docteh: Well, that line asking for latest isn't doing all that well of a job. when I turfed it, only 3 out of 37 were effected, and when I added `set(CMAKE_CXX_STANDARD 20)` yuzu and inih are put back to latest, only cubeb is changed.

```
ALL_BUILD           2014
audio_core          latest
clang-format        2014
common              latest
core                latest
crypto              2014
crypto_obj          2014
cubeb               2014
dynarmic            2020
enet                2014
getopt              2014
glad                2014
host_shaders        2014
inih                latest
input_common        latest  
INSTALL             2014
mbed_crypto         2014
mbedtls             2014
mbedx509            2014
mcl                 2020
network             latest
opus                2014
RUN_TESTS           2014
shader_recompiler   latest
sirit               2020
speex               2014
ssl                 2014
ssl_obj             2014
tests               latest 
usb                 2014
video_core          latest
web_service         latest
yuzu                latest
yuzu-cmd            latest
yuzu-room           latest
Zycore              2014
Zydis               2014
```

lioncash: We generally only care about our code being compiled as latest, externals don't matter and you should always assume they'll do whatever they want. The point is to catch future issues in our code, not the externals, since they follow their own procedures for compiler support.

Also setting it to 20 to get latest isn't forward compatible, because when C++23 gets finalized that cmake command will no longer have the same behavior on an updated version of cmake.

Docteh: Other than `add_compile_options($<$<COMPILE_LANGUAGE:CXX>:/std:c++latest>)` is there any other way to get latest? Just needs to work on one target.

I tried doing it in the top of src/yuzu/CMakeLists.txt but it didn't work

lioncash: have you tried doing 

```cpp
target_compile_options(TargetName PRIVATE /std:c++latest)
```

with a cleared CMake cache?

CMakeLists.txt
```diff
@@ -201,14 +202,20 @@ if (MINGW)
...
 # Qt5 requires that we find components, so it doesn't fit our pretty little find package function
 if(ENABLE_QT)
     set(QT_VERSION 5.15)
+    set(QT5_VERSION 5.15)
 
     # Check for system Qt on Linux, fallback to bundled Qt
     if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
         if (NOT YUZU_USE_BUNDLED_QT)
-            find_package(Qt5 ${QT_VERSION} COMPONENTS Widgets DBus Multimedia)
+            find_package(Qt5 ${QT5_VERSION} COMPONENTS Widgets DBus Multimedia)
```
Tachi107, line 218: Doesn't this mean that CMake will always try to look for Qt5 on Linux?

CMakeLists.txt
```diff
@@ -311,18 +318,33 @@ if(ENABLE_QT)
...
+    message(STATUS "YUZU_QT_COMPONENTS: " ${YUZU_QT_COMPONENTS})
+
+    if (ENABLE_QT6)
+        find_package(Qt6 COMPONENTS ${YUZU_QT_COMPONENTS} ${YUZU_QT_NO_CMAKE_SYSTEM_PATH})
+    endif()
+    if (Qt6_FOUND)
+        message(STATUS "yuzu/CMakeLists.txt: Qt6Widgets_VERSION ${Qt6Widgets_VERSION}, setting QT_VERSION")
+        set(QT_VERSION ${Qt6Widgets_VERSION})
+        # Qt6 sets cxx_std_17 and we need to undo that
+        set_target_properties(Qt6::Platform PROPERTIES INTERFACE_COMPILE_FEATURES "")
```
Tachi107, line 342: `target_compile_features(Qt6::Platfrom INTERFACE "")` would be prettier, but I don't know if it works for clearing properties

Docteh: I looked at dolphin-emu, they're doing the same thing but on Qt5/6::Core instead of Platform.

CMakeLists.txt
Docteh: I wanted to avoid touching the section, but I've moved things around those that section only runs when doing Qt5

Docteh:
Qt Multimedia is being turned into optional requirement. Emulator can be compiled against the regular MSVC2019 option that Qt's online downloader provides. 

Additional Libraries for consideration, from 6.4.0 but should be same with 6.3.1
If you want to set YUZU_USE_QT_WEB_ENGINE to yes, you'll need those 4 items.
Camera: Multimedia
Web Engine: PDF Positioning WebChannel WebEngine

Docteh:
With the exception of this QPlatformNativeInterface thing I successfully compiled in arch Linux with Qt6

/home/kyle/yuzu/src/yuzu/bootmanager.cpp:25:10: fatal error: qpa/qplatformnativeinterface.h: No such file or directory
   25 | #include <qpa/qplatformnativeinterface.h>


```diff
diff --git a/src/yuzu/bootmanager.cpp b/src/yuzu/bootmanager.cpp
index f9f534521..4c31558aa 100644
--- a/src/yuzu/bootmanager.cpp
+++ b/src/yuzu/bootmanager.cpp
@@ -22,7 +22,7 @@
 #endif

 #if !defined(WIN32)
-#include <qpa/qplatformnativeinterface.h>
+#include <6.3.1/QtGui/qpa/qplatformnativeinterface.h>
 #endif

 #include <fmt/format.h>
```

warning near top of qpa/qplatformnativeinterface.h
```cpp
#ifndef QPLATFORMNATIVEINTERFACE_H
#define QPLATFORMNATIVEINTERFACE_H

//
//  W A R N I N G
//  -------------
//
// This file is part of the QPA API and is not meant to be used
// in applications. Usage of this API may make your code
// source and binary incompatible with future versions of Qt.
//
```

src/yuzu/bootmanager.cpp
```diff
@@ -867,6 +874,7 @@ void GRenderWindow::RequestCameraCapture() {
 
     pending_camera_snapshots++;
     camera_capture->capture();
+#endif
 }
 
 void GRenderWindow::OnCameraCapture(int requestId, const QImage& img) {
```
Docteh, line 880: With the references to GetCamera and SetCameraData, I don't understand why this function (GRenderWindow::OnCameraCapture) compiles without QtMultimedia

CMakeLists.txt
Docteh: Tried with `target_compile_features(Qt6::Platform INTERFACE "")` and no effect on how MSVC wants to compile yuzu target (/sdc:c++17)

Tachi107: Ok, thanks for checking :)

Docteh:
Rebased (over 200 commits behind) and tweaked the gate for the Qt5 IR Camera code.

`#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0)) && YUZU_USE_QT_MULTIMEDIA`

Docteh:
Needed rebase after #8932 was merged

CMakeLists.txt
```diff
@@ -213,128 +218,166 @@ if (MINGW)
...
-                    Perhaps `libglu1-mesa-dev` needs to be installed?")
-            endif()
-            pkg_check_modules(QT_DEP_MESA QUIET dri>=20.0.8)
-            if (NOT QT_DEP_MESA_FOUND)
-                message(FATAL_ERROR "Qt bundled pacakge dependency `dri` not found. \
-                    Perhaps `mesa-common-dev` needs to be installed?")
-            endif()
+                # Check that the system GLIBCXX version is compatible
+                find_program(OBJDUMP objdump)
+                if ("${OBJDUMP}" STREQUAL "OBJDUMP-NOTFOUND")
```
Tachi107, line 278: This can be simplified as:

```suggestion
                if (NOT OBJDUMP)
```

See <https://cmake.org/cmake/help/latest/command/if.html#basic-expressions>

CMakeLists.txt
```diff
@@ -213,128 +218,166 @@ if (MINGW)
...
+                    libxcb-xinerama.so.0
+                    libxcb-xkb.so.1
+                    libxcb.so.1
+                    libxkbcommon-x11.so.0
+                    libxkbcommon.so.0
+                )
+                set(UNRESOLVED_QT_DEPS "")
+                foreach (REQUIREMENT ${BUNDLED_QT_REQUIREMENTS})
+                    find_library(BUNDLED_QT_${REQUIREMENT} ${REQUIREMENT})
+                    if ("${BUNDLED_QT_${REQUIREMENT}}" STREQUAL "BUNDLED_QT_${REQUIREMENT}-NOTFOUND")
```
Tachi107, line 336: Same as above

```suggestion
                    if (NOT BUNDLED_QT_${REQUIREMENT})
```

Tachi107 (commented):
This suggestions do not apply specifically to this patch (as you've only changed indentation), but it'd be nice to simplify the code anyway :)

hurricane-dorian:
It is pretty strange but in "Graphic Settings" Resolution resets to 0.5X, Adapting Filter to Nearest and Aliasing to None every time you restart Yuzu. I even tried resetting Yuzu settings to default but no luck.

liamwhite (approved):

PR #8467, opened 2022-06-15
Project yuzu Fried Chicken (Y.F.C.) Part 1

FernandoS27:
This is the very fundamental base of project YFC a project aimed at improving multiple aspects of GPU emulation that have been left in the oblivion.

Part 1 includes:
- Full rework of GPU Driver based on the Re and code from Skyline Emulator with full permission from @bylaws (please check that all crediting and copyright notices are correct).
- Full rework of GPU memory management.
- Initial implementation of GPU Channels, this fixes World of light in SSBU, Deltarune, among a bunch of other games.
- Fixes and accelerations to the DMA engine.
- A series of fixes for Vulkan, Buffer Cache, Shader Decompiler that affected multiple games.
- Rework host frame presentation to be a considerably smoother (fixes Xenoblade DE).

Part 1 does NOT include:
- Performance improvements. (main theme)
- Buffer Cache Rework. (this fixes FE: Warriors)
- Macro HLE Rework. DrawIndirect, DrawElementsIndirect, MultiDrawIndirect, etc (perf for MHR and Koemo tecmo games).
- Fencing Rework with asynchronous buffer and texture downloads.

These will be addressed in part 2.

This are further features that are partially done internally and will be released later after Part 2:
- Accelerate DMA.
- Unnormalized texture coords & rectangular textures.
- Rendering to / from ASTC textures (biggest pain of UE4 games).
- Host Conditional Rendering ? (still under consideration as it only seemed beneficial in 1 game and detrimental in others).

Credits:
- Bylaws (@bylaws ) & Skyline Emulator [Skyline Emulator](https://github.com/skyline-emu/skyline) for providing information on the RE of NVIDIA's kernel driver as well as being the base at which most of the DRV rework had been done.
- AsLody (@asLody ) For finding out the cause of the issues behind Yokai Watch 4, The Witcher 3, Hades, among others. Along some of the solutions applied here.
- Pharynx (@liamwhite ) and toastUnlimited (@lat9nq ) Linux help and compiling bugs fixes on GCC / Clang.
- yuzu Team & Testers for annoying me to rebase / release this as it has been done since April xD.

PR #8833, opened 2022-08-30
ci: Enable building with Visual Studio 2022 (again)

Morph1984:
Since the following https://developercommunity.visualstudio.com/t/Type-alias-lookup-failure-within-paramet/10039150 compiler bug has been fixed, we can finally build with VS 2022 again.


Morph1984:
#8467 fails to compile due to conversion issues

lioncash:
Is this good to go or are there still issues?

lioncash (approved):

PR #8837, opened 2022-08-31
(shader/pipeline)_cache: Raise shader/pipeline cache version

Morph1984:
Since the following commit: https://github.com/yuzu-emu/yuzu/commit/a83a5d2e4c8932df864dd4cea2b04d87a12c8760 , many games will refuse to boot unless the shader/pipeline cache has been invalidated.
Raises the shader/pipeline cache version to force an invalidation.


Morph1984:
That change is fine, since the same is done in [gl_shader_cache](https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/renderer_opengl/gl_shader_cache.cpp#L283)

lioncash (approved):

jacky400:
Wonder if it is possible we can invalidate only partial shader cache instead of invalidating entire shader code

bunnei (approved):

PR #8842, opened 2022-09-01
[audio_core] Rework audio output

### LINKED ISSUE #8857: [Xenoblade Chronicles 2] Update after EA2930, audio breaks and slows down during cutscene playback.
I updated with EA2930 and the error occurred.
During cutscene playback, the entire game slows down, the sound is interrupted, and occasionally freezes.
30 fps drops to about 6 fps and returns to 30 fps when the pre-rendered video finishes playing, and sound synchronization returns to normal.
In some scenes, synchronization may not return.
Both English and Japanese audio are the same.
I think the "[audio_core] Rework audio output #8842" is affecting it.
There seems to be a problem with audio playback during playback of pre-rendered video.

How to check the scene for slowdowns.
Events Theater->Scenario Event->Chapter Four: Aegis->Tragedy of Tora, Mòrag Reappers

Specs:
Windows 11 Version 21H2
i7-11700 2.50GHz
DDR4 32GB 2400MHz
RTX 3080 Ti 12GB - Driver 512.95
Vulkan API

Yuzu:
Yuzu EA 2930 or later(No problem until 2929.)

Game Ver:
2.0.2, 2.1.0

[EA2929_EA2930_log.zip](https://github.com/yuzu-emu/yuzu/files/9483723/EA2929_EA2930_log.zip)

Kelebek1:
Main purpose is to fix audio desync in games using audio out (notably XC3 cutscenes). Playback is now connected to core timing rather than being based on the realtime backend thread. 

Also removed the audio renderer queue and rely on the same stalling if it gets too large, this may cause an issue with coretiming underrunning though, so I might revert this if it's reported as a problem.

Closes #8857

### DIFF
--- a/src/audio_core/sink/null_sink.h
+++ b/src/audio_core/sink/null_sink.h
@@ -3,10 +3,29 @@
 
 #pragma once
 
+#include <string>
+#include <string_view>
+#include <vector>
+
 #include "audio_core/sink/sink.h"
 #include "audio_core/sink/sink_stream.h"
 
+namespace Core {
+class System;
+} // namespace Core
+
 namespace AudioCore::Sink {
+class NullSinkStreamImpl final : public SinkStream {
+public:
+    explicit NullSinkStreamImpl(Core::System& system_, StreamType type_)
+        : SinkStream{system_, type_} {}
+    ~NullSinkStreamImpl() override {}
+    void AppendBuffer(SinkBuffer&, std::vector<s16>&) override {}
+    std::vector<s16> ReleaseBuffer(u64) override {
+        return {};
+    }
+};
+
 /**
  * A no-op sink for when no audio out is wanted.
  */
@@ -15,14 +34,15 @@ public:
     explicit NullSink(std::string_view) {}
     ~NullSink() override = default;
 
-    SinkStream* AcquireSinkStream([[maybe_unused]] Core::System& system,
-                                  [[maybe_unused]] u32 system_channels,
-                                  [[maybe_unused]] const std::string& name,
-                                  [[maybe_unused]] StreamType type) override {
-        return &null_sink_stream;
+    SinkStream* AcquireSinkStream(Core::System& system, u32, const std::string&,
+                                  StreamType type) override {
+        if (null_sink == nullptr) {
+            null_sink = std::make_unique<NullSinkStreamImpl>(system, type);
+        }
+        return null_sink.get();
     }
 
-    void CloseStream([[maybe_unused]] const SinkStream* stream) override {}
+    void CloseStream(SinkStream*) override {}
     void CloseStreams() override {}
     void PauseStreams() override {}
     void UnpauseStreams() override {}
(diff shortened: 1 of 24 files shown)


src/audio_core/sink/null_sink.h
```diff
@@ -15,14 +31,15 @@ class NullSink final : public Sink {
...
     ~NullSink() override = default;
 
-    SinkStream* AcquireSinkStream([[maybe_unused]] Core::System& system,
-                                  [[maybe_unused]] u32 system_channels,
-                                  [[maybe_unused]] const std::string& name,
-                                  [[maybe_unused]] StreamType type) override {
-        return &null_sink_stream;
+    SinkStream* AcquireSinkStream(Core::System& system, u32, const std::string&,
+                                  StreamType type) override {
+        if (null_sink.get() == nullptr) {
```
lioncash, line 36: Equality with nullptr can be done directly with smart pointers

```suggestion
        if (null_sink == nullptr) {
```

src/audio_core/sink/null_sink.h
```diff
@@ -3,10 +3,26 @@
 
 #pragma once
 
+#include <string>
+#include <string_view>
+#include <vector>
+
 #include "audio_core/sink/sink.h"
 #include "audio_core/sink/sink_stream.h"
+#include "core/core.h"
```
lioncash, line 12: This can be a forward declaration

src/audio_core/sink/sink_stream.cpp
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+
+    samples_buffer.Push(samples);
+    queue.enqueue(buffer);
+    queued_buffers++;
+}
+
+std::vector<s16> SinkStream::ReleaseBuffer(u64 num_samples) {
+    static constexpr s32 min = std::numeric_limits<s16>::min();
```
lioncash, line 112: These are just local numeric constants, they don't need a global lifetime

src/audio_core/sink/sink_stream.cpp
```diff
@@ -0,0 +1,259 @@
...
+    std::memcpy(&last_frame[0], &output_buffer[(frames_written - 1) * frame_size],
+                frame_size_bytes);
+
+    if (stalled && queued_buffers <= max_queue_size) {
+        Unstall();
+    }
+}
+
+void SinkStream::Stall() {
+    if (!stalled) {
```
lioncash, line 246: This can be a guard clause

```cpp
if (stalled) {
    return;
}
```

src/audio_core/sink/sink_stream.cpp
```diff
@@ -0,0 +1,259 @@
...
+        stalled = true;
+        system.StallProcesses();
+    }
+}
+
+void SinkStream::Unstall() {
+    if (stalled) {
```
lioncash, line 253: Ditto as:

```cpp
if (!stalled) {
    return;
}
```

src/audio_core/sink/sink_stream.h
```diff
@@ -3,12 +3,17 @@
...
 #include <atomic>
 #include <memory>
+#include <span>
 #include <vector>
 
 #include "audio_core/common/common.h"
 #include "common/common_types.h"
+#include "common/reader_writer_queue.h"
+#include "common/ring_buffer.h"
+#include "core/core.h"
```
lioncash, line 16: This can be a forward declaration

src/audio_core/sink/sink_stream.cpp
Kelebek1: What's the reason to do that instead of keeping it as-is? Just style?

src/audio_core/sink/sink_stream.h
Kelebek1: How do I do this? I can't figure out what syntax I need to do this. They're both templates, but SinkBuffer isn't declared until further down in the file, trying to forward-declare that doesn't work.

```cpp
struct AudioCore::Sink::SinkBuffer;
namespace Common {
template <>
class ReaderWriterQueue<AudioCore::Sink::SinkBuffer>;
template <>
class RingBuffer<s16, 0x10000>;
} // namespace Common
```

lioncash: I meant just turning the core include into a forward declaration (by forward declaring the system instance), not that all of the includes need to be forward declared

src/audio_core/sink/sink_stream.cpp
lioncash: Any refactoring that potentially adds to the code remains unindented unnecessarily and it makes the early-exit/invalid case explicit instead of implicit.

bunnei (approved):

PR #8849, opened 2022-09-02
astc: Enable parallel CPU astc decoding

Morph1984:
Given the issues with GPU accelerated ASTC decoding with NVIDIA's latest drivers, parallelize astc decoding on the CPU.
Uses half the available threads in the system for astc decoding.


goldenx86 (approved):
LGTM if no GPU dev chips in.

bunnei (approved):

PR #8863, opened 2022-09-05
core: hid: Fix GC triggers overwriting ZL and ZR buttons

german77:
Since we use the same input for both triggers and buttons. Trigger inputs from GC controllers where overwriting the input of the ZL and ZR buttons. Usually not noticiable until you enable a property like toggle button.

Fixes #8862

PR #8864, opened 2022-09-06
input_common: Add support for analog toggle

german77:
Allows to toggle buttons that have analog inputs.

### DIFF
--- a/src/common/input.h
+++ b/src/common/input.h
@@ -102,6 +102,8 @@ struct AnalogProperties {
     float offset{};
     // Invert direction of the sensor data
     bool inverted{};
+    // Press once to activate, press again to release
+    bool toggle{};
 };
 
 // Single analog sensor data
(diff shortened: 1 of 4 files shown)


unuruu5760:
I really, really, really, really, really, really thank you. you are the best!!

unuruu5760:
Oh no!!!!!!!!!!!! , I just tested it. But this doesn't work in games.

unuruu5760:
> 



> It does work. But remember this is another PR. #8863 fixes the error in game while this one add support for analog toggle. Wait until both are on mainline.

YEAH ,I tested using this PR but  it didn't work. Is this wrong?

german77:
I mean you need to compile yourself a custom build with both PR merged or just wait until both are on mainline. They are small changes it will be in less than a week

bunnei (approved):

PR #8865, opened 2022-09-07
sdl_driver: Added Joy-Cons to the HD rumble list

Mutcholoko:
New SDL includes Joy-Cons support, and they weren't mentioned on the HasHDRumble() function, which resulted on a return of false.


german77:
Sorry but we haven't updated to latest SDL we are using 2.0.20. New Joycon implementation has a couple of issues see https://github.com/libsdl-org/SDL/issues/6095 besides from already having full dual joycon support we don't have a reason to update yet.

Mutcholoko:
Oh, damn, I should've double checked that then. My bad.
Closing this request.

german77:
No worries. I appreciate the effort.

PR #8867, opened 2022-09-08
CMake: explicitly link mbedcrypto for yuzu-room

### LINKED ISSUE #8866: OpenSSL vs mbedtls
https://github.com/gentoo-mirror/guru/blob/master/games-emulation/yuzu/yuzu-9999.ebuild

That guru thing is build metadata managed by gentoo community.

26: Early in log:  * Skipping submodule mbedtls
155: -- Found OpenSSL: /usr/lib64/libcrypto.so (found suitable version "1.1.1q", minimum required is "1.1")  
1234: yuzu_room.cpp:(.text.startup+0x123d): undefined reference to `mbedtls_base64_decode'

Maybe OpenSSL also provides  a base64_decode function?

` -DENABLE_WEB_SERVICE=no` might not disable yuzu-room target

Two things to run down. 
1) Can we decode base64 without mbedtls library
  The answer is yes, but like do we OpenSSL: https://www.openssl.org/docs/man1.0.2/man3/EVP_DecodeInit.html or something else
2) Does `-DENABLE_WEB_SERVICE=no` still compile yuzu-room target
  I will test this later, maybe file a PR about it.

Edit: github acting up, but ENABLE_WEB_SERVICE does what it says on the tin

Docteh:
Doesn't appear to effect anything regular, but in both Linux and Windows builds it looks like our project has all the libraries available for linking. If this feature is turned off, there is only one thing that quit working, when linking yuzu-room it couldn't find a function called mbedtls_base64_decode

mbedtls is split into three libraries for some reason: 
mbedtls
mbedx509
mbedcrypto

mbedtls_base64_decode is in mbedcrypto

-----

fixes #8866  

ebuild in question https://github.com/gentoo-mirror/guru/blob/master/games-emulation/yuzu/yuzu-9999.ebuild

### DIFF
--- a/src/dedicated_room/CMakeLists.txt
+++ b/src/dedicated_room/CMakeLists.txt
@@ -16,7 +16,7 @@ if (ENABLE_WEB_SERVICE)
     target_link_libraries(yuzu-room PRIVATE web_service)
 endif()
 
-target_link_libraries(yuzu-room PRIVATE mbedtls)
+target_link_libraries(yuzu-room PRIVATE mbedtls mbedcrypto)
 if (MSVC)
     target_link_libraries(yuzu-room PRIVATE getopt)
 endif()


bunnei (approved):

Tachi107:
You could've removed the `mbedtls` library from the list of needed libs, as the code only uses `mbedcrypto`. I'll do it in a separate PR.

PR #8871, opened 2022-09-09
Draft Workaround for SSL ca-certificate issue

Docteh:
usr/lib/libssl.so.1.1 in various linux distributions have differing opinions on where to locate the bundle of certificates used to properly handle https traffic.

Ubuntu 18.04: /usr/lib/ssl/cert.pem
Manjaro (Arch): /etc/ssl/cert.pem

In the AppImage libssl.so.1.1 is being shipped.

Notable because we build the AppImage on a Ubuntu base, and the steam decks use arch.

This is a hastily thrown together PR, so should not be merged without.

The AppRun shell script is modified to specify SSL_CERT_FILE in environment

-----

a bit yolo on this one. Looks like https://github.com/darealshinji/AppImageKit-checkrt was rewritten


Docteh:
trace.27217:openat(AT_FDCWD, "/usr/lib/ssl/cert.pem", O_RDONLY) = -1 ENOENT (No such file or directory)
trace.27217:stat("/usr/lib/ssl/certs/8e363097.0", 0x7fef78dfc250) = -1 ENOENT (No such file or directory)
trace.27217:stat("/usr/lib/ssl/certs/653b494a.0", 0x7fef78dfc250) = -1 ENOENT (No such file or directory)
trace.27217:stat("/usr/lib/ssl/certs/653b494a.0", 0x7fef78dfc250) = -1 ENOENT (No such file or directory)

cp: cannot stat '/usr/lib/ssl/cert.pem': No such file or directory

ooops

For later reference
818:openat(AT_FDCWD, "/usr/lib/ssl/openssl.cnf", O_RDONLY) = -1 ENOENT (No such file or directory)
```
kyle@hausfrau:~$ ls /usr/lib/ssl -lR
/usr/lib/ssl:
total 4
lrwxrwxrwx 1 root root   14 Mar  5  2020 certs -> /etc/ssl/certs
drwxr-xr-x 2 root root 4096 Aug 12 17:40 misc
lrwxrwxrwx 1 root root   20 Jul  4 04:25 openssl.cnf -> /etc/ssl/openssl.cnf
lrwxrwxrwx 1 root root   16 Mar  5  2020 private -> /etc/ssl/private

/usr/lib/ssl/misc:
total 16
-rwxr-xr-x 1 root root 7597 Jul  4 04:25 CA.pl
lrwxrwxrwx 1 root root    8 Jul  4 04:25 tsget -> tsget.pl
-rwxr-xr-x 1 root root 6574 Jul  4 04:25 tsget.pl
```

/etc/ssl/certs has 259 files

links except for
ca-certificates.crt
ssl-cert-snakeoil.pem

the ca-certificates.crt appears to be right. bunch of certificates in a file.
-rw-r--r-- 1 root root 195453 Jun 10 22:04 /etc/ssl/certs/ca-certificates.crt


Over on arch /etc/ssl/cert.pem is a link to 
-r--r--r-- 1 root root 220514 Jul 29 04:51 /etc/ca-certificates/extracted/tls-ca-bundle.pem
has a comment in front of each one with the name.

The note on ARGV0 is 100% unrelated to this, but I think it may be worth looking into
https://docs.appimage.org/packaging-guide/environment-variables.html

> Scenarios where ARGV0 is really useful involve so-called multi-binary AppImages, where the filename in ARGV0 defines which program is called inside the AppImage. This concept is also known from single-binary tools like [BusyBox](https://en.wikipedia.org/wiki/BusyBox), and can be implemented in a custom AppRun script (see [Architecture](https://docs.appimage.org/reference/architecture.html#ref-architecture) for more information).

Docteh:
@lat9nq https://github.com/darealshinji/AppImageKit-checkrt was rewritten to primarily use a shell script for AppRun, but it now has a checkrt binary. I think we should do this, but executed better. 


I think just shipping a copy of ca-certificates from the docker image is a reasonable solution. If the root certicates are sufficiently out of date, might not want such an old yuzu talking to API...

Docteh:
Maybe explicitly add checked to every theme

Docteh:
This icon seemed fine until I pasted a token in,  I thought I did something wrong..
image


Settling on this blue icon when the text is changed in the field. We should come back to this with some local token validation, but I think this PR is kind of needed for AppImages to do LDN, so... its out of scope.

image

src/yuzu/configuration/configure_web.cpp
```diff
@@ -128,20 +128,25 @@ void ConfigureWeb::RefreshTelemetryID() {
...
-        ui->label_token_verified->setPixmap(pixmap);
+        // Empty = no icon
+        ui->label_token_verified->setPixmap(QPixmap());
+        ui->label_token_verified->setToolTip(QString());
     } else {
         user_verified = false;
 
-        const QPixmap pixmap = QIcon::fromTheme(QStringLiteral("failed")).pixmap(16);
+        // Show a warning if it's been changed, clearer than showing failure
+        const QPixmap pixmap = QPixmap(QStringLiteral(":/icons/default/16x16/info.png"));
```
german77, line 138: Any reason you are using the full path rather than letting qt pick the correct file for the theme?

Docteh: the QIcon::fromTheme stuff works 100% great until you listen to the Linux users on the concept of OS themes. 

After some thought, maybe just need to make the default_colorful theme has its own copy of everything.

using the full path JustWorks instead of worrying about current and future fromTheme related problems. Like maybe the icon theme should actually be called like yuzu_app_default or yuzu_app_colorful to avoid collisions

Tachi107:
Is adding 3k lines of third party code necessary? Can't the AppImage thing be added as a submodule or downloaded when needed by the AppImage job?

Docteh:
@Tachi107 definitely willing to take advice on importing the thing. my concern is that https://github.com/darealshinji/AppImageKit-checkrt is kind of restarted. I see an old branch, so maybe if they redo it again we can grab a specific commit later? Should we save a fork just in case? Not a fan of the idea of just blindly grabbing latest, as it might change again substantially.

I don't understand why the checkrt tool needs to copy the GCC Runtime exemption, the exemption doesn't mention that it needs to be included, but I can only try my best to follow licenses. The res.h file is created using a tool `xxd` that the ubuntu docker image doesn't have. So I added a res.h and removed the line from the .gitignore. 

We do need to have a modified AppRun script. I'd prefer to have that inside the yuzu repo somewhere. 

I included the icon changes into this, because I wasn't getting any visible change in the UI when I hit verify. if the certificate bundle is bad I get a null error, but pass/fail on a token doesn't. So I've been wrestling with both changes. I think I got the icon changes nailed down so I can separate it out.

Docteh:
Filed a PR to get checkrt-x86_64 into proper ext-linux-bin repository, see: https://github.com/yuzu-emu/ext-linux-bin/pull/10

I took a closer look at what exec.so I think its unchanged, so using the one already checked in over in yuzu-emu/ext-linux-bin

lat9nq:
We really should be using https://github.com/yuzu-emu/AppImageKit-checkrt at this point. I stalled on it because I asked a review on its first PR and never got a response.

Docteh:
Looking at that PR, I think that the AppRun shell script itself should be part of the yuzu repository.

Specifically I could see a future Wayland PR wanting to remove your suggested change so that the generated AppImage's will actually use Wayland.
```sh
# Force xcb platform for Qt applications
if [ -z "${QT_QPA_PLATFORM}" ]; then
    export QT_QPA_PLATFORM=xcb
fi
```

Looks like the source for checkrt is unchanged?

Docteh:
UN-commenting these temporarily, if you're using a GUI you won't see the extra infos 
#echo ">>>>> $LD_LIBRARY_PATH"
#echo ">>>>> $LD_PRELOAD"

Docteh:
Superseded by #8933, yes. But I want to have the PR functional.

Docteh:
https://github.com/yuzu-emu/AppImageKit-checkrt/pull/3

PR #8878, opened 2022-09-10
Remove pause callbacks from coretiming

Kelebek1:
As requested by @bunnei. Was able to delete a bunch of functions related to pausing in general along with it. Pausing/unpausing/shutdown still working from all my testing but may be jank somewhere.


bunnei:
Thanks!

bunnei (approved):

PR #8898, opened 2022-09-13
Add option "Accept any Internet access"

noneistaken:
This setting, which you can change at runtime unlike "Network Interface", is a workaround for Splatoon 3.
Along with the Project London pull request and the "Enable Auto-Stub" option enabled, this should allow the user to at least enter The Shoal and create a LAN lobby (not a LDN lobby which errors out for some reason).
I have not tested if this allows users to enter each others' games, since I don't have a computer with enough memory to run 2 Yuzu instances.

To elaborate on using this option/toggle:
- **Uncheck** "Accept any Internet access" and **check** "Enable Auto-Stub"
- Launch Splatoon 3
- Once you enter the game, **check** "Accept any Internet access"
- Enter the Shoal
- Hold ZL+ZR+Left Stick to enter LAN mode
- You should be able to create a room now.

### DIFF
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -529,6 +529,7 @@ struct Values {
     Setting<bool> extended_logging{false, "extended_logging"};
     Setting<bool> use_debug_asserts{false, "use_debug_asserts"};
     Setting<bool> use_auto_stub{false, "use_auto_stub"};
+    Setting<bool> accept_any_net{false, "accept_any_net"};
     Setting<bool> enable_all_controllers{false, "enable_all_controllers"};
 
     // Miscellaneous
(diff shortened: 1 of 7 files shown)


noneistaken:
(The forced push was to get rid of a typo from porting these changes from Early Access to master.)

noneistaken:
Forgot to make this more clear, but one should rebase this to either the London pull request or Early Access for it to actually be useful.

FearlessTobi (changes requested):
I'm not a big fan of this solution to be honest. 
For the most of the network requests, data from the network interfaces is needed for them to be sucessful.
Just reporting that internet is available, when there's no interface to make requests with is not a great solution.

This is also the reason why LDN doesn't work with your changes.

noneistaken:
> I'm not a big fan of this solution to be honest. For the most of the network requests, data from the network interfaces is needed for them to be sucessful. Just reporting that internet is available, when there's no interface to make requests with is not a great solution.
> 
> This is also the reason why LDN doesn't work with your changes.

I'm aware, this is only meant to be a workaround for those who do not wish to wait for a proper solution. If this isn't to your liking, please close the PR.

noneistaken:
Stale, closing

PR #8906, opened 2022-09-14
UI: move icons from default into colorful theme.

Docteh:
colorful theme has been default theme for awhile. having colorful theme try and grab icons from other theme doesn't work on Linux.

Also adding two additional icons, info is to hint to the user that they should hit verify after pasting in a token, sync is to show that the verification is occurring.

------
Pulling this out of #8871


dist/qt_themes/colorful/icons/index.theme
```diff
@@ -1,5 +1,4 @@
 [Icon Theme]
-Name=colorful
```
Docteh, line 2: lol (I wanted to remove `Inherits=default` line)

liamwhite (approved):

bunnei (approved):

PR #8910, opened 2022-09-15
video_core: Modify astc texture decode error fill value

vonchenplus:
The moon of LIVE A LIVE is a 9-layer astc texture, but the data to be decoded is very strange, only the first two layers are normal, from the third layer onwards all are 0. I have not found the essential cause of this problem so far.

The current solution is to modify the fill value when an error occurs to prevent the wrong texture from being mixed in when sampling.

Before:
image

After:
image


FernandoS27:
This happens because UE4 games actually render the mipmaps on the go. Our Texture cache, fails to do the conversion.

bunnei (approved):

theboy181:
Is this an actual solution, or is it just a happy accident?

vonchenplus:
> Is this an actual solution, or is it just a happy accident?

just a happy accident, still need to look into it.

PR #8930, opened 2022-09-19
yuzu qt: Add option to disable startup Vulkan check

lat9nq:
The startup check apparently confuses other programs when yuzu launches 2 processes and then quickly closes one of them. Though this isn't really our issue it's also not a big deal for me to add an option to work around that issue.

Notably Steam on Linux considers yuzu closed when the child process closes.


Morph1984 (approved):

LOuroboros:
There's a line inside Yuzu's General Configuration's Debug tab where this config option is located which says "*This will be reset automatically when yuzu closes.*" right below the Advanced options, but doesn't seem to be accurate *(at least not anymore)* considering that the state of this new "Perform Vulkan Startup Check" config option does get saved.

I suppose it would be a good idea to correct that, either in this PR or in a new one? :eyes:

PR #8933, opened 2022-09-20
ci/linux: Use AppRun.sh to start AppImage

lat9nq:
This script still uses the old version of AppImageKit-checkrt (our fork with the `old` branch: https://github.com/yuzu-emu/AppImageKit-checkrt/tree/old), but now we use the shell script version so we can set our own environment variables as the application starts up.

This specific version searches for and sets the correct root CA file to prevent SSL errors in yuzu.

Supersedes #8871

### DIFF
--- a/.ci/scripts/linux/docker.sh
+++ b/.ci/scripts/linux/docker.sh
@@ -35,11 +35,11 @@ rm -vf AppDir/usr/bin/yuzu-cmd AppDir/usr/bin/yuzu-tester
 # Download tools needed to build an AppImage
 wget -nc https://github.com/yuzu-emu/ext-linux-bin/raw/main/appimage/linuxdeploy-x86_64.AppImage
 wget -nc https://github.com/yuzu-emu/ext-linux-bin/raw/main/appimage/linuxdeploy-plugin-qt-x86_64.AppImage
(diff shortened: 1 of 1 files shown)


Docteh:
`exec="./bin/$(sed -n 's|^Exec=||p' $(ls -1 *.desktop))"`

This line leaves %f, and assumes that the executable is in bin/yuzu instead of usr/bin/yuzu
```
[docteh@haus squashfs-root]$ echo ./bin/$(sed -n 's|^Exec=||p' $(ls -1 *.desktop))
./bin/yuzu %f
```

at the bottom
```
$exec "$*"
exit $?
```

In #8871 I changed the bottom to this so that ./yuzu.AppImage -f -g  "/path/to/game" would function
```sh
exec ./usr/bin/yuzu "$@"
```

Docteh:
Followup, AppImage launched, not tested with arguments yet, but I'm at least half-wrong :)

Docteh:
I filed a PR over at https://github.com/yuzu-emu/AppImageKit-checkrt/pull/3 to cover launching games directly.

Kurosawa79:
This still needs to fixed. Cannot launch games direct from Game mode in Steam Deck. Tested 1175 and still doesn't work.

Docteh:
Just tested 1176, launched a game successfully from the terminal

Docteh (approved):

Morph1984 (approved):

PR #8993, opened 2022-09-26
ci/linux: Drop linuxdeploy usage

lat9nq:
Recent versions of Docker appear to cause the Qt linuxdeploy plugin to throw a boost file copy error in some of our build containers (at the moment, specifically on the Patreon releases, but it's also happened on Mainline at least once? So mostly Azure.)

This switches from linuxdeploy to a [script of mine](https://github.com/lat9nq/deploy/blob/main/linux/deploy-linux.sh) I've been working on for a while.


Docteh:
This started happening on Github Actions ☹️

Morph1984 (approved):

PR #9039, opened 2022-10-08
Auto select the SDL audio backend when Cubeb latency is too high

Kelebek1:
Currently we only select cubeb with the auto option, but some audio devices report really large latencies from cubeb, which causes problems with playback, this PR just tests cubeb before selecting it, switching to SDL if cubeb fails to init or the latency is too large (threshold relatively random).

### DIFF
--- a/src/yuzu/configuration/configure_audio.cpp
+++ b/src/yuzu/configuration/configure_audio.cpp
@@ -161,8 +161,8 @@ void ConfigureAudio::InitializeAudioSinkComboBox() {
     ui->sink_combo_box->clear();
     ui->sink_combo_box->addItem(QString::fromUtf8(AudioCore::Sink::auto_device_name));
 
-    for (const char* id : AudioCore::Sink::GetSinkIDs()) {
-        ui->sink_combo_box->addItem(QString::fromUtf8(id));
+    for (const auto& id : AudioCore::Sink::GetSinkIDs()) {
+        ui->sink_combo_box->addItem(QString::fromUtf8(id.data(), static_cast<s32>(id.length())));
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/yuzu/configuration/configure_audio.cpp
```diff
@@ -161,8 +161,8 @@ void ConfigureAudio::InitializeAudioSinkComboBox() {
     ui->sink_combo_box->clear();
     ui->sink_combo_box->addItem(QString::fromUtf8(AudioCore::Sink::auto_device_name));
 
-    for (const char* id : AudioCore::Sink::GetSinkIDs()) {
-        ui->sink_combo_box->addItem(QString::fromUtf8(id));
+    for (const auto& id : AudioCore::Sink::GetSinkIDs()) {
+        ui->sink_combo_box->addItem(QString::fromUtf8(id.data()));
```
Tachi107, line 165: `std::string_view` is not guaranteed to contain a null-terminated string. Maybe here it is better to call `QString::fromUtf8(id.data(), id.length())`?

src/yuzu/configuration/configure_audio.cpp
```diff
@@ -161,8 +161,8 @@ void ConfigureAudio::InitializeAudioSinkComboBox() {
     ui->sink_combo_box->clear();
     ui->sink_combo_box->addItem(QString::fromUtf8(AudioCore::Sink::auto_device_name));
 
-    for (const char* id : AudioCore::Sink::GetSinkIDs()) {
-        ui->sink_combo_box->addItem(QString::fromUtf8(id));
+    for (const auto& id : AudioCore::Sink::GetSinkIDs()) {
+        ui->sink_combo_box->addItem(QString::fromUtf8(id.data(), static_cast<s32>(id.length())));
```
Tachi107, line 165: Nit: this overload takes a `qsizetype`, which is not necessarily an `s32`. It would be nicer to cast to `qsizetype`, if possible :)

liamwhite (approved):

PR #8955, opened 2022-09-25
core: nfp: Rewrite implementation to remove direct access from the frontend

german77:
Removes all hacks used to load an amiibo, adds an unique nfc interface for each npad. Implements functions needed for the cabinet applet. Reorganizes the service to match RE. Makes amiibo keys mandatory to read and write any data.

This fixes amiibo support on games like Shovel knight or erasing progress on SSBU. Only the cabinet applet is left to implement and a few details that I haven't fully understand.

fixes #8496
fixes #8493

PR #9044, opened 2022-10-09
ci/windows: Revert to using GCC for MinGW builds

lat9nq:
Conceptually a revert of #8455

Using MinGW in the future may not be ideal as it does not work very well with crash dumps (#8682).

Switch back to GCC on MinGW. This also gives CI a way to check GCC 12 (as of writing, or whatever version of mingw-gcc Arch happens to be shipping on a given week).


lat9nq:
MinGW build failure is due to #8955, needs #9043

liamwhite (approved):

PR #9054, opened 2022-10-11
CMake: Try add_library "lz4" if "lz4::lz4" is unavailable

Docteh:
Right now this looks like a distro specific problem, but we'll have to see.

Over on Gentoo: with lz4 1.9.3 there is a lz4::lz4 library target, with 1.9.4 it's no longer mentioned in the cmake files provided by the  package. (/usr/lib64/cmake/lz4)

arch and openSUSE have lz4 1.9.4 available so I checked there, they only have .pc files for pkg-config, so asking for "lz4::lz4" works as usual

MSVC does require "lz4::lz4" to be asked for

--------
Bug reported and Fix tested by @FellFromTheSky


liamwhite (approved):

src/common/CMakeLists.txt
```diff
@@ -169,7 +169,11 @@ endif()
 create_target_directory_groups(common)
 
 target_link_libraries(common PUBLIC ${Boost_LIBRARIES} fmt::fmt microprofile Threads::Threads)
-target_link_libraries(common PRIVATE lz4::lz4)
+if (TARGET lz4::lz4)
+  target_link_libraries(common PRIVATE lz4::lz4)
+else()
+  target_link_libraries(common PRIVATE lz4)
```
Tachi107, line 175: This will cause CMake to continue the configuration phase even if the lz4 target doesn't exist, causing a failure at link phase, unless the `lz4` library is found by the linker. But even then, flags required by the lz4 library won't be passed to the linker, and this could potentially lead to issues.

If this is a bug in a rolling release distro, it should be fixed there.

Tachi107: It seems that the issue has been caused by this patch: https://github.com/lz4/lz4/pull/1030

Upstream now provides a CMake Config file when building with CMake, and its targets are part of the `LZ4::` namespace.

Docteh: should I copy the code lower down where for zstd we're asking for _shared if available with a fallback to _static? or just ask for LZ4::lz4_shared for now?

Tachi107: I'd just ask for the shared variant for now, as the upper case namespace is likely to be defined only when using (recent) distribution packages, and they almost always ship a shared library (often they don't even ship a static library at all).

abouvier: Why not just remove `CONFIG` from `find_package(lz4 CONFIG)` and let [Findlz4.cmake](https://github.com/yuzu-emu/yuzu/blob/master/externals/find-modules/Findlz4.cmake) do its magic?

abouvier: vcpkg also exports .pc files, and the find module will detect them.

Tachi107: Do they also work on Windows? I think that `pkg_search_module` requires a `pkg-config` or `pkgconf` binary to be available

abouvier: It works in my [PR](https://github.com/yuzu-emu/yuzu/pull/6833) for both lz4 and zstd inside the yuzu CI Windows image.

Tachi107: GitHub's CI images are so full of stuff it's pretty much impossible to know if things are going to work on a default install of the tested OS... For instance, they include pkg-config by default.

abouvier: yuzu already has a bunch of `find_package(PkgConfig REQUIRED)` which [require the presence of a `pkg-config` executable](https://github.com/Kitware/CMake/blob/ba8f393f78cda4801f074e27de788fb191b43fab/Modules/FindPkgConfig.cmake#L100).

Docteh:
```
CMakeLists.txt:255:            Include(FindPkgConfig REQUIRED) <-- gated by Linux
CMakeLists.txt:389:    include(FindPkgConfig) <-- maybe
CMakeLists.txt:413:    Include(FindPkgConfig REQUIRED) <-- UNIX AND NOT APPLE
```

For the middle one it might not be critical? 
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
-- test PKG_CONFIG_EXECUTABLE-NOTFOUND
-- libusb-1.0 not found.
CMake Warning at CMakeLists.txt:402 (message):
  libusb not found, falling back to externals

I will do a more clean build of yuzu and report back

abouvier: Your fork is not [up to date](https://github.com/yuzu-emu/yuzu/commit/09a87966e06fd8ba1cf40332a6c4ef81f8a1eac8) :p

Docteh: Yup, to check I used following patch, and PKG_CONFIG_EXECUTABLE was set to PKG_CONFIG_EXECUTABLE-NOTFOUND and it kept trucking along

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 20dd1383f..5c0de52c2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,6 +387,7 @@ endif()
 # Ensure libusb is properly configured (based on dolphin libusb include)
 if(NOT APPLE AND NOT YUZU_USE_BUNDLED_LIBUSB)
     include(FindPkgConfig)
+    message(STATUS "test ${PKG_CONFIG_EXECUTABLE}")
     if (PKG_CONFIG_FOUND AND NOT CMAKE_SYSTEM_NAME MATCHES "DragonFly|FreeBSD")
         pkg_check_modules(LIBUSB QUIET libusb-1.0>=1.0.24)
     else()

```

I had happened to have a pkg-config from devkitpro, so I had to remove it to verify a clean build.

Docteh: saw your message about the fork not being up to date,

-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
-- test PKG_CONFIG_EXECUTABLE-NOTFOUND

stuff in the log looks the same to me. even after switching to find_package

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 957df54f5..b5f1b71d8 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,6 +387,7 @@ endif()
 # Ensure libusb is properly configured (based on dolphin libusb include)
 if(NOT APPLE AND NOT YUZU_USE_BUNDLED_LIBUSB)
     find_package(PkgConfig)
+    message(STATUS "test ${PKG_CONFIG_EXECUTABLE}")
     if (PKG_CONFIG_FOUND AND NOT CMAKE_SYSTEM_NAME MATCHES "DragonFly|FreeBSD")
         pkg_check_modules(LIBUSB QUIET libusb-1.0>=1.0.24)
     else()

```

Tachi107: > yuzu already has a bunch of `find_package(PkgConfig REQUIRED)` which [require the presence of a `pkg-config` executable](https://github.com/Kitware/CMake/blob/ba8f393f78cda4801f074e27de788fb191b43fab/Modules/FindPkgConfig.cmake#L100).

Yes, but those are only used on Linux, or when not bundling certain dependencies.

PR #9091, opened 2022-10-17
UI: Add option to hide the compatibility list

Docteh:
Option is added directly below the option for the addons column

Defaulting to hide compatibility list. Changing default works properly.

-----

New option
image

I made sure to test the default because I don't want to see users needing help turning the compatibility list back on when we're ready to show it again.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -819,6 +819,7 @@ void Config::ReadUIGamelistValues() {
     qt_config->beginGroup(QStringLiteral("UIGameList"));
 
     ReadBasicSetting(UISettings::values.show_add_ons);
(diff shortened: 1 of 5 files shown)


legend800:
What about the other undesired columns like File Type and Size? Ideally a solution allows users to hide all of them as none present much value. 

2022-10-18 00_02_34-yuzu Early Access 3033

piplup55:
> What about the other undesired columns like File Type and Size? Ideally a solution allows users to hide all of them as none present much value.
> 
> 2022-10-18 00_02_34-yuzu Early Access 3033

maybe something i can think about

Kewlan:
Perhaps a solution similar to how Dolphin and RPCS3 does it would be good? Right click the top category bar for the options:
image

bunnei (approved):

PR #9113, opened 2022-10-22
service: nfp: Allow amiibos without keys

german77:
Since many users  started using outdated builds instead of dumping the amiibo keys. I decided to add a little workaround by marking amiibos as read only when keys are missing. This enables basic detection of amiibos.

No fallbacks are given to keep the code accurate. Games that use decrypted data will still need keys.

### DIFF
--- a/src/core/hle/service/nfp/amiibo_crypto.cpp
+++ b/src/core/hle/service/nfp/amiibo_crypto.cpp
@@ -9,6 +9,7 @@
 #include <mbedtls/hmac_drbg.h>
 
 #include "common/fs/file.h"
+#include "common/fs/fs.h"
 #include "common/fs/path_util.h"
 #include "common/logging/log.h"
 #include "core/hle/service/mii/mii_manager.h"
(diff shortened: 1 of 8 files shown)


src/core/hle/service/nfp/amiibo_crypto.cpp
```diff
@@ -295,6 +295,16 @@ bool LoadKeys(InternalKey& locked_secret, InternalKey& unfixed_info) {
     return true;
 }
 
+bool IsKeyAvailable() {
+    const auto yuzu_keys_dir = Common::FS::GetYuzuPath(Common::FS::YuzuPath::KeysDir);
+
+    const Common::FS::IOFile keys_file{yuzu_keys_dir / "key_retail.bin",
```
lioncash, line 301: should be more straightforward to do:

```cpp
return Common::FS::Exists(yuzu_keys_dir / "key_retail.bin");
```

no?

german77: Thanks. I was going to ask morph for a better alternative

bunnei (approved):

PR #9127, opened 2022-10-25
video_core: Catch vulkan clear op not all channel need clear

vonchenplus:
This is just a temporary solution,  Auxiliary shader needs to be added to perform specified channel clear.

Fix p5r's battle scenes black.

image


Fuzzy64 (approved):

liamwhite (approved):

bunnei (approved):

VITORUGOP5R:
como se adiciona o sombreador falado?

PR #9134, opened 2022-10-26
audio_in/out_system: Pass Initialize members by value where applicable

lioncash:
applet_resource_user_id isn't actually modified and is just assigned to a member variable, so this doesn't need to be a mutable reference.

Similarly, the device name itself isn't modified and is only moved. We pass by value here, since we can still perform the move, but eliminate a sneaky set of calls that can unintentionally destroy the original string. Given how nested the calls are, it's good to get rid of this potential vector for a use-after-move bug.


Kelebek1:
I had to mess with this because it wasn't setting the name properly. The other references should be able to go too if possible.

liamwhite (approved):

PR #9157, opened 2022-10-30
core: hle: service: acc: Fix ListOpenContextStoredUsers/StoreOpenContext.

bunnei:
- These APIs are used to capture the opened users and allow that state to be persisted across processes.
- They are not intended to just return the system opened users, that is what ListOpenUsers is for.
- Fixes the launch hang with Bayonetta 3.

### DIFF
--- a/src/core/hle/service/acc/acc.cpp
+++ b/src/core/hle/service/acc/acc.cpp
@@ -512,10 +512,11 @@ protected:
 
 class IManagerForApplication final : public ServiceFramework<IManagerForApplication> {
 public:
-    explicit IManagerForApplication(Core::System& system_, Common::UUID user_id_)
+    explicit IManagerForApplication(Core::System& system_,
+                                    const std::shared_ptr<ProfileManager>& profile_manager_)
         : ServiceFramework{system_, "IManagerForApplication"},
           ensure_token_id{std::make_shared<EnsureTokenIdCacheAsyncInterface>(system)},
-          user_id{user_id_} {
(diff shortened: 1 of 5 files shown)


src/core/hle/service/acc/acc.cpp
```diff
@@ -512,10 +512,11 @@ class EnsureTokenIdCacheAsyncInterface final : public IAsyncContext {
 
 class IManagerForApplication final : public ServiceFramework<IManagerForApplication> {
 public:
-    explicit IManagerForApplication(Core::System& system_, Common::UUID user_id_)
+    explicit IManagerForApplication(Core::System& system_,
+                                    std::shared_ptr<ProfileManager>& profile_manager_)
```
lioncash, line 516:
```suggestion
                                    const std::shared_ptr<ProfileManager>& profile_manager_)
```

src/core/hle/service/acc/acc.cpp
```diff
@@ -855,16 +859,14 @@ void Module::Interface::LoadOpenContext(Kernel::HLERequestContext& ctx) {
     // This is similar to GetBaasAccountManagerForApplication
     // This command is used concurrently with ListOpenContextStoredUsers
     // TODO: Find the differences between this and GetBaasAccountManagerForApplication
```
lioncash, line 861: Seems like this TODO can be removed, since the behavior isn't identical anymore

FernandoS27 (approved):

PR #9180, opened 2022-11-04
UI: split up strings relating to content removal

Docteh:
Requested by Italian translator (Fs00 in Discord)

"Remove Installed Game %1?"
"Error Removing %1"

I didn't press for translated strings, so have a taste direct from deepl

Rimuovere il contenuto del gioco installato?
Rimuovere l'aggiornamento del gioco installato?
Rimuovere il DLC del gioco installato?

-------
UI Element in question
image

Request from our discord 
> Hello everyone! I've found out an issue regarding a particular translatable string in Yuzu: "Remove Installed Game %1?".
Since I wasn't able to understand what the placeholder was for, I dug into the code and discovered that the %1 can get replaced with Contents, Update or DLC (https://github.com/yuzu-emu/yuzu/blob/master/src/yuzu/main.cpp#L1999-L2016).
The problem is that the code is assuming that the part leading to the placeholder ("Remove Installed Game") won't change regardless of the placeholder, which is not the case in other languages such as Italian. In the latter language for example, I need to add an article before Contents, Update or DLC which is not the same in all cases.
So, instead of using a placeholder, I'd recommend having 3 different translatable strings for the 3 scenarios ("Remove Installed Game Contents?", "Remove Installed Game Update?", "Remove Installed Game DLC?") in order to leave more room to translators. Thank you!

Note: [Updated code link](https://github.com/yuzu-emu/yuzu/blob/9fc1bcc7b2da398d12327e8111c21e453a4af27d/src/yuzu/main.cpp#L2021-L2036)

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2018,38 +2018,50 @@ static bool RomFSRawCopy(QProgressDialog& dialog, const FileSys::VirtualDir& src
     return true;
 }
 
+QString GMainWindow::GetGameListErrorRemoving(InstalledEntryType type) const {
+    switch (type) {
+    case InstalledEntryType::Game:
+        return tr("Error Removing Contents");
+    case InstalledEntryType::Update:
+        return tr("Error Removing Update");
+    case InstalledEntryType::AddOnContent:
+        return tr("Error Removing DLC");
+    default:
+        return QStringLiteral("Error Removing <Invalid Type>");
+    }
+}
 void GMainWindow::OnGameListRemoveInstalledEntry(u64 program_id, InstalledEntryType type) {
-    const QString entry_type = [type] {
+    const QString entry_question = [type] {
         switch (type) {
         case InstalledEntryType::Game:
-            return tr("Contents");
+            return tr("Remove Installed Game Contents?");
         case InstalledEntryType::Update:
-            return tr("Update");
+            return tr("Remove Installed Game Update?");
         case InstalledEntryType::AddOnContent:
-            return tr("DLC");
+            return tr("Remove Installed Game DLC?");
         default:
-            return QString{};
+            return QStringLiteral("Remove Installed Game <Invalid Type>?");
         }
     }();
 
-    if (QMessageBox::question(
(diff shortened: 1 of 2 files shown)


Fs00:
Thanks for the PR @Docteh! I'm Fs00 here on GitHub too :)
The DeepL translations are correct, you can see that the article (_il_ / _l'_) differs according to the entry type and that's why the single string with placeholder is not enough.

Regarding the code, for the `default:` cases should we instead have an assertion or raise an exception, since they can never be reached? In either case, I'm not a big fan of having strings on Transifex that are never used anywhere...

Docteh:
I forgot about asserts, I'll implement those later

src/yuzu/main.cpp
```diff
@@ -2018,46 +2018,59 @@ static bool RomFSRawCopy(QProgressDialog& dialog, const FileSys::VirtualDir& src
...
+QString GMainWindow::GetGameListErrorRemoving(InstalledEntryType type) {
+    switch (type) {
+    case InstalledEntryType::Game:
+        return tr("Error Removing Contents");
+    case InstalledEntryType::Update:
+        return tr("Error Removing Update");
+    case InstalledEntryType::AddOnContent:
+        return tr("Error Removing DLC");
+    default:
+        return tr("Error Removing");
```
lioncash, line 2030: This should probably be an untranslated string saying "Invalid entry type!" or something indicating a programming error occurred. Translators shouldn't need to worry about this case

Docteh:
Normal strings eg "Error Removing Contents" "Remove Installed Game Contents?"

```
return QStringLiteral("Error Removing <Invalid Type>");
return QStringLiteral("Remove Installed Game <Invalid Type>?");
```

It's a bit clunky because I want to make sure that if an end user sees the message and asks us about it, we can most easily find the problem.

Docteh:
Not sure. 

In GetDirectionName left,right,up,down come from SDL, so there is a chance that they could add something, and without the error in the log, english readers might not immediately notice. Alternatively if SDL changed from "left" to "Left", I wouldn't notice.

In the case of this PR, we're in control of the whole thing. 

I'm leaning slightly against it.

PR #9213, opened 2022-11-08
network: play together in different room

SmileYik:
create a room named PostOffice, it accepts and forwards packets from a game room to another game room and made different room's players can play together.


src/network/room.h
```diff
@@ -140,6 +140,13 @@ class Room final {
      */
     void Destroy();
 
+    /**
+     *
+     */
+    bool SetupMailBox(const std::string& server_address, u16 server_port);
+
+    bool HasMailBox();
```
lioncash, line 148:
```suggestion
    bool HasMailBox() const;
```

src/network/room.cpp
```diff
@@ -1140,4 +1240,15 @@ void Room::Destroy() {
     room_impl->room_information.name.clear();
 }
 
+bool Room::SetupMailBox(const std::string& server_address, u16 server_port) {
```
lioncash, line 1243:
```suggestion
bool Room::SetupMailBox(std::string server_address, u16 server_port) {
```

src/network/room.cpp
```diff
@@ -1140,4 +1240,15 @@ void Room::Destroy() {
...
+    room_impl->post_office.address = server_address;
```
lioncash, line 1244:
```suggestion
    room_impl->post_office.address = std::move(server_address);
```

src/network/room.cpp
```diff
@@ -48,6 +48,15 @@ class Room::RoomImpl {
     IPBanList ip_ban_list;             ///< List of banned IP addresses
     mutable std::mutex ban_list_mutex; ///< Mutex for the ban lists
 
+    struct PostOffice {
+        const u8 WaitPacket = 10; ///< Wait post office's reply max times.
```
lioncash, line 52: If this is always const, it can be turned into a static constexpr member, making the struct a little smaller.

```suggestion
        static constexpr u8 WaitPacket = 10; ///< Wait post office's reply max times.
```

src/network/room_post_office.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    class Mailbox {
+    public:
+        ENetPeer* peer; ///< the remote room's peer
+    };
+    using MailboxList = std::vector<Mailbox>;
+    /// remote rooms
+    MailboxList mailboxes;
+
+    RoomPostOfficeImpl() {}
```
lioncash, line 36:
```suggestion
    RoomPostOfficeImpl() = default;
```

src/network/room_post_office.cpp
```diff
@@ -0,0 +1,164 @@
...
+                break;
+            case ENET_EVENT_TYPE_CONNECT:
+                HandleJoinMailbox(&event);
+                break;
+            }
+        }
+    }
+
+    for (const Mailbox& mailbox : mailboxes) {
+        if (mailbox.peer)
```
lioncash, line 80: This should be braced

src/network/room_post_office.h
```diff
@@ -0,0 +1,46 @@
+// SPDX-FileCopyrightText: Copyright 2017 Citra Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <array>
```
lioncash, line 6: A lot of the includes here seem like they can be moved into the cpp file, since a lot of these seem to be unused in the header.

src/network/room.cpp
```diff
@@ -1037,6 +1070,73 @@ void Room::RoomImpl::HandleClientDisconnection(ENetPeer* client) {
...
+
+inline void Room::RoomImpl::DisconnectPostOffice() {
+    post_office.address = "";
+    if (post_office.peer) {
+        enet_peer_disconnect(post_office.peer, 0);
+        post_office.peer = nullptr;
+    }
+}
+
+inline void Room::RoomImpl::HandlePostOfficeConnection(const ENetEvent* event) {
```
lioncash, line 1121: Ditto

src/network/room.cpp
```diff
@@ -1037,6 +1070,73 @@ void Room::RoomImpl::HandleClientDisconnection(ENetPeer* client) {
...
+
+    // LOG_INFO(Network,
+    //            "Connect to room post office {}:{} success!",
+    //            post_office.address, post_office.port);
+
+    return true;
+}
+
+inline void Room::RoomImpl::DisconnectPostOffice() {
+    post_office.address = "";
```
lioncash, line 1114:
```suggestion
    post_office.address.clear();
```

src/network/network.cpp
```diff
@@ -43,6 +49,11 @@ void RoomNetwork::Shutdown() {
             m_room->Destroy();
         m_room.reset();
     }
+    if (m_room_post_office) {
+        if (m_room_post_office->GetState() == RoomPostOffice::State::Open)
```
lioncash, line 53: This should be braced

FearlessTobi (changes requested):
Apart from issues with the code that were already pointed out above, I'm really trying to think of a usecase where this feature would really be useful or needed.
As far as I understand, the "post office" is meant to connect different rooms together, but I can't find many situations where you can't just make a bigger room to fit more players into it.

SmileYik:
> Apart from issues with the code that were already pointed out above, I'm really trying to think of a usecase where this feature would really be useful or needed.
> As far as I understand, the "post office" is meant to connect different rooms together, but I can't find many situations where you can't just make a bigger room to fit more players into it.

In China, players who use the campus network or other network providers' networks when playing Animal Crossing connect to a single operator's line room server, and it is easy to loss connection, made crash the island and lose their save progress due to the abnormal network of the player who joined the island.
And I think the online experience would be better when players connect to a room server that is more network-friendly to their own network and join other room's people's games while playing.

FearlessTobi:
Thank you for your contribution, but I don't think this code has the chance of being merged into yuzu anytime soon.
I understand the problem with connection losses, especially in China, but we need a different solution than this.

Thank you for your understanding.

PR #9214, opened 2022-11-08
src: Add save file import and export

DD3Boh:
* This is very useful when playing the same game on
  different devices
* Use zip archives with the help of libzip for this
* The archive structure is program_id/save_file

Signed-off-by: Davide Garberi <dade.garberi@gmail.com>

### DIFF
--- a/src/yuzu/game_list.h
+++ b/src/yuzu/game_list.h
@@ -58,6 +58,11 @@ enum class InstalledEntryType {
     AddOnContent,
 };
 
+enum class SaveFileOperation {
+    Export,
+    Import,
+};
+
 class GameList : public QWidget {
     Q_OBJECT
 
@@ -103,6 +108,7 @@ signals:
     void OpenFolderRequested(u64 program_id, GameListOpenTarget target,
                              const std::string& game_path);
     void OpenTransferableShaderCacheRequested(u64 program_id);
+    void HandleSaveFile(SaveFileOperation operation, u64 program_id, const std::string& game_path);
     void RemoveInstalledEntryRequested(u64 program_id, InstalledEntryType type);
     void RemoveFileRequested(u64 program_id, GameListRemoveTarget target,
                              const std::string& game_path);
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -6,6 +6,7 @@
 #include <cmath>
 #include <memory>
 #include <thread>
+#include <zip.h>
 #ifdef __APPLE__
 #include <unistd.h> // for chdir
 #endif
@@ -1226,6 +1227,7 @@ void GMainWindow::ConnectWidgetEvents() {
     connect(game_list, &GameList::OpenFolderRequested, this, &GMainWindow::OnGameListOpenFolder);
     connect(game_list, &GameList::OpenTransferableShaderCacheRequested, this,
             &GMainWindow::OnTransferableShaderCacheOpenFile);
+    connect(game_list, &GameList::HandleSaveFile, this, &GMainWindow::OnHandleSaveFile);
     connect(game_list, &GameList::RemoveInstalledEntryRequested, this,
             &GMainWindow::OnGameListRemoveInstalledEntry);
     connect(game_list, &GameList::RemoveFileRequested, this, &GMainWindow::OnGameListRemoveFile);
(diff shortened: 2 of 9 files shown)


Docteh:
> since handling zip archives properly takes an unreasonable amount of code

how unreasonable are we talking?

> This filetype is nothing more than the game code, followed by the filename of the save, followed by the raw content of the save

That sounds like a [tar file](https://en.wikipedia.org/wiki/Tar_(computing)).

For backups, I think yuzu should use a standard format, as game saves can be used with other emulators, or on hardware.

DD3Boh:
> > since handling zip archives properly takes an unreasonable amount of code
> 
> how unreasonable are we talking?
> 
> > This filetype is nothing more than the game code, followed by the filename of the save, followed by the raw content of the save
> 
> That sounds like a [tar file](https://en.wikipedia.org/wiki/Tar_(computing)).
> 
> For backups, I think yuzu should use a standard format, as game saves can be used with other emulators, or on hardware.

The main problem is that from what I've seen, handling both zip and tar (not 100% sure about the latter) needs external libraries like quazip or karchive, and I just don't think it's the ideal solution. If uncompressed tar files can be handled without additional external libraries outside of Qt/C++, that would be the best solution yeah.

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+            *system, vfs_nand_dir, FileSys::SaveDataSpaceId::NandUser,
+            FileSys::SaveDataType::SaveData, program_id, {}, 0);
+
+        path = Common::FS::ConcatPathSafe(nand_dir, device_save_data_path);
+    }
+
+    const QString qpath = QString::fromStdString(Common::FS::PathToUTF8String(path));
+    const QDir dir(qpath);
+    QFile file;
+    QDataStream dataStream;
```
lioncash, line 2041: Our convention is `snake_case` not `camelCase`. Ditto for a few other variables past this point.

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+                }
+            }
+
+            if (!outFile.open(QIODevice::WriteOnly)) {
+                file.close();
+                return;
+            }
+
+            outFile.write(dataByteArray);
+            outFile.close();
```
lioncash, line 2114: files automatically close when they go out of scope, so this isn't necessary.

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+        }
+
+        file.close();
```
lioncash, line 2117: Ditto

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+    if (operation == 0) {
+        if (!dir.exists()) {
+            QMessageBox::warning(this, tr("Error Opening %1 Folder").arg(open_target),
+                                 tr("Folder does not exist!"));
+            return;
+        }
+
+        QString destinationFile = QFileDialog::getSaveFileName(
+            this, tr("Save Game Files"),
+            QString(QStandardPaths::DocumentsLocation) + QString::fromStdString("/"),
```
lioncash, line 2052:
```cpp
QStringLiteral("%1/%2).arg(QString(QStandardPaths::DocumentsLocation), tr("Yuzu Data Save (*.yzsv)")))
```

seems a bit more appropriate here. That or use QString::fromUtf8 instead of QString::fromStdString to convert a forward slash over to a QString (the latter is quite heavy handed.

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+
+        file.setFileName(destinationFile);
+        file.open(QIODevice::WriteOnly);
+
+        dataStream.setDevice(&file);
+
+        QFileInfoList filesList = dir.entryInfoList();
+
+        for (int i = 0; i < filesList.length(); i++) {
+            qCritical() << "for loop run " << i;
```
lioncash, line 2063: Leftover logging statement?

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+            QFile filefor(dir.absolutePath() + QString::fromStdString("/") +
+                          filesList.at(i).fileName());
+            qCritical() << "filefor " << filefor;
```
lioncash, line 2066: Ditto

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+        file.setFileName(destinationFile);
+        file.open(QIODevice::WriteOnly);
+
+        dataStream.setDevice(&file);
+
+        QFileInfoList filesList = dir.entryInfoList();
+
+        for (int i = 0; i < filesList.length(); i++) {
+            qCritical() << "for loop run " << i;
+            QFile filefor(dir.absolutePath() + QString::fromStdString("/") +
```
lioncash, line 2064: Ditto about either using a format string or QString::fromUtf8

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+            dataStream << filesList.at(i).fileName();
+            dataStream << filefor.readAll();
+
+            filefor.close();
+        }
+        file.close();
+    } else {
+        QString sourceFile = QFileDialog::getOpenFileName(
+            this, tr("Choose Game File"),
+            QString(QStandardPaths::DocumentsLocation) + QString::fromStdString("/"),
```
lioncash, line 2080: Ditto

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+
+        dataStream.setDevice(&file);
+
+        while (!dataStream.atEnd()) {
+            QString fileName, game_id;
+            QByteArray dataByteArray;
+
+            dataStream >> game_id >> fileName >> dataByteArray;
+
+            QFile outFile(qpath + QString::fromStdString("/") + fileName);
```
lioncash, line 2101: Ditto

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+            if (!dir.exists()) {
```
lioncash, line 2102: These if statements can be combined

```suggestion
            if (!dir.exists() && !dir.mkpath(qpath)) {
```

src/yuzu/main.cpp
```diff
@@ -1965,6 +1966,158 @@ void GMainWindow::OnGameListOpenFolder(u64 program_id, GameListOpenTarget target
...
+        dataStream.setDevice(&file);
+
+        QFileInfoList filesList = dir.entryInfoList();
+
+        for (int i = 0; i < filesList.length(); i++) {
+            qCritical() << "for loop run " << i;
+            QFile filefor(dir.absolutePath() + QString::fromStdString("/") +
+                          filesList.at(i).fileName());
+            qCritical() << "filefor " << filefor;
+            if (!filefor.open(QIODevice::ReadOnly))
```
lioncash, line 2067: This should be braced

src/yuzu/game_list.h
```diff
@@ -103,6 +103,8 @@ class GameList : public QWidget {
     void OpenFolderRequested(u64 program_id, GameListOpenTarget target,
                              const std::string& game_path);
     void OpenTransferableShaderCacheRequested(u64 program_id);
+    // operation: 0 = export, 1 = import
+    void HandleSaveFile(bool operation, u64 program_id, const std::string& game_path);
```
lioncash, line 107: This should be using an enum class rather than a bool to specify the file operation to make code more explicit.

e.g.

```cpp
HandleSaveFile(SaveOperation::Export, program_id, path); 
```

is more self-documenting than:

```cpp
HandleSaveFile(0, program_id, path); 
```

This also allows for easier expansion of operations in the future if necessary.

src/yuzu/main.h
```diff
@@ -288,6 +288,7 @@ private slots:
     void OnGameListAddDirectory();
     void OnGameListShowList(bool show);
     void OnGameListOpenPerGameProperties(const std::string& file);
+    void OnHandleSaveFile(bool operation, u64 program_id, const std::string& game_path);
```
lioncash, line 291: Ditto about `operation` being an enum class.

FearlessTobi:
Just fyi, yuzu already ships with libraries that can handle compression, like lz4 and zstd.
I also agree with docteh, that we should use some kind of standard format for this.

DD3Boh:
> I also agree with docteh, that we should use some kind of standard format for this.

If that's the way to go, ok. Which format should I try implementing? Just to make sure that I don't have to change it once already finished and pushed here

merryhime:
+1 for .zip

We had previously libzip as a dependency before boxcat was removed. See 5f6cd32c9e63ebbda120063cb9a282a7f1374d4d for reference.

Morph1984:
Hi there, great work at the attempt at implementing this feature.
However, I'm currently working on a virtual filesystem rewrite which will make said changes incompatible with the implementation I had come up with when it comes to importing and exporting save data.

As others have said above, using a standard container like zip is much more user friendly than inventing a yuzu-specific container for saves.

As for what my implementation exports, the following is an example generated by it:
[exported_save.zip](https://github.com/yuzu-emu/yuzu/files/9965591/exported_save.zip)

DD3Boh:
Okay, nice to know you're working on that already.
So should I just give up on this change or redo it with zip files? My ideal implementation with zip files would be to create zip files containing paths like <game_id>/<savefile>
Just let me know if I should finish it and push it or not

liberodark:
So that's what you wanted to do ^^
I who very often move my backups I think it's a good idea.

Morph1984:
Given that my rewrite is expected to be releasing soon, having 2 different import formats will be a burden to support.

DD3Boh:
Ok, closing this pull request then and I'll wait for your implementation.

PR #9217, opened 2022-11-09
Add break statement in default cases

HidroSaphire:
According to the contributing page (https://github.com/yuzu-emu/yuzu/wiki/Contributing) the default cases should have a break statement

default:
        // Yes, even break for the last case
        break;


lioncash (approved):
Hey, thanks for the contribution! I appreciate it

PR #9223, opened 2022-11-10
Add CPU thread count to log files

goldenx86:
Doesn't differentiate real cores from threads, but it's a safe global way.


src/yuzu/main.cpp
```diff
@@ -361,6 +362,7 @@ GMainWindow::GMainWindow(std::unique_ptr<Config> config_, bool has_broken_vulkan
     }
     LOG_INFO(Frontend, "Host CPU: {}", cpu_string);
 #endif
+    LOG_INFO(Frontend, "Host CPU thread count: {}", processor_count);
```
Morph1984, line 365:
```suggestion
    LOG_INFO(Frontend, "Host CPU Threads: {}", processor_count);
```

bunnei (approved):

Issue #9077, opened 2022-10-15
minecraft nintendo switch edition graphic issues since yuzu 1195+


piplup55:
everything still works however it does show some odd behaviour mostly in the menu from what i can see
edit: i did get it to happen on 1191 but now it seems happen on 1195+

**before 1195:**
![image](https://user-images.githubusercontent.com/100526773/196009706-e64ac070-84d2-41af-83dc-8231b85b03b6.png)
![image](https://user-images.githubusercontent.com/100526773/196009711-2c4ec4b3-5eea-4753-a0a4-e6e9c2ed8525.png)
![image](https://user-images.githubusercontent.com/100526773/196009779-f3e26660-2ddc-4477-a8db-e90a6f64f6d2.png)

**after 1195:**
![image](https://user-images.githubusercontent.com/100526773/196009748-5837e450-a341-44dd-b147-65e45f76486c.png)
![image](https://user-images.githubusercontent.com/100526773/196009752-e2f841e5-0431-4c34-98b6-772e3ebb62aa.png)
![image](https://user-images.githubusercontent.com/100526773/196009756-e85da356-0245-42ad-bc43-c12daa86c57f.png)

**System Specs:**
Linux:
 - OS: Arch Linux
 - CPU: AMD Ryzen 7 3700X
 - GPU: RX6600 XT
 - Memory: 64GB, Swap: N/A
 - GPU Driver: Mesa 22.2.1, radv 22.2.1

gidoly:
can you test build between 1192,1193, 1194?

gidoly:
broken on 1193 likely broken due to #8766

piplup55:
while #8766 did break text rendering a bit it doesn't seem to be the root cause of the issue i report, since text was fixed by something else (used EA 3004)
![image](https://user-images.githubusercontent.com/100526773/200844470-24aea3f9-f7e6-412b-9645-4cbd88a12fbb.png)

PR #9226, opened 2022-11-11
[video_core] Fix a couple regs regressions

Kelebek1:
In my regs PR I removed regs which weren't listed in nVidia's official docs, but that caused regressions in some games due to losing the `invalidate_texture_data_cache` register. I don't know if this reg is correct or not, but we know there are regs which were redacted from the docs, and it seems we do need something to trigger GPU cache invalidation, so I've added it back.

I also added back conservative raster for Vulkan, not sure if the reg is correct or if conservative raster affects any games. Also fixes a mistake with two-sided stencils on OpenGL.

This PR fixes the weird blocky patterns in XC3, and likely helps other games.


piplup55:
#9077 can be closed when this gets merged

liamwhite (approved):

vonchenplus (approved):

bunnei (approved):

PR #9228, opened 2022-11-11
Add break statement in default case

HidroSaphire:
According to the contributing page (https://github.com/yuzu-emu/yuzu/wiki/Contributing) the default cases should have a break statement

### DIFF
--- a/src/yuzu_cmd/yuzu.cpp
+++ b/src/yuzu_cmd/yuzu.cpp
@@ -351,6 +351,7 @@ int main(int argc, char** argv) {
                          "additional help.\n\nError Code: {:04X}-{:04X}\nError Description: {}",
                          loader_id, error_id, static_cast<Loader::ResultStatus>(error_id));
         }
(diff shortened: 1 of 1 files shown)


Docteh:
Hi, for review purposes it might be better to do these in a batch instead of one at a time. 

Personally I'm wondering if breaks are needed when there is a bunch of return statements. Example:
https://github.com/yuzu-emu/yuzu/blob/5eb30c7827959075cd240f39c6ab57d2c2a8197e/src/core/core.cpp#L72-L83

HidroSaphire:
The break I put in this commit is more than valid if you take a good look at the code. I didn't choose the laziest way, I did the code analysis and saw this to fix.

Docteh:
> Plus I can say that there are not only returns in the switch case

I was trying to say that I was curious `switch` statements in situations where the `default:` uses a `return`.

liamwhite (approved):

PR #9229, opened 2022-11-11
Add break for default cases

Docteh:
Visual Studio has an option to search all files in a solution, so I did a search in there for "default:" looking for any missing break statements.

I've left out default statements that `return` something, and that `throw` something, even if via `ThrowInvalidType`

----
UNREACHABLE() also throws

Follow-up to #9217  #9228

### DIFF
--- a/src/audio_core/renderer/performance/performance_manager.cpp
+++ b/src/audio_core/renderer/performance/performance_manager.cpp
@@ -26,6 +26,7 @@ void PerformanceManager::CreateImpl(const size_t version) {
         impl = std::make_unique<
             PerformanceManagerImpl<PerformanceVersion::Version1, PerformanceFrameHeaderVersion1,
                                    PerformanceEntryVersion1, PerformanceDetailVersion1>>();
+        break;
     }
 }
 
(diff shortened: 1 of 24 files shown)


Docteh:
https://github.com/yuzu-emu/yuzu/blob/5eb30c7827959075cd240f39c6ab57d2c2a8197e/src/core/hle/service/nvdrv/devices/nvhost_ctrl_gpu.cpp#L303-L308


https://github.com/yuzu-emu/yuzu/blob/5eb30c7827959075cd240f39c6ab57d2c2a8197e/src/core/hle/service/nvdrv/devices/nvhost_gpu.cpp#L367-L372

For these two I'm not sure if I should add break, or extra `return nullptr;` statements

Docteh:
in src/core/hle/kernel/k_page_table.cpp there are two cases of R_THROW that I added break statements to. 

"R_THROW -> R_RETURN -> return"

But investigation revealed that the R_THROW macro leads to return, so I pulled those two.

liamwhite (approved):

PR #9231, opened 2022-11-11
Add CPU core count to log files

goldenx86:
[GetLogicalProcessorInformation](http://msdn.microsoft.com/en-us/library/ms683194) is used for Windows, which is a chore and a half.
Linux checks for the existence of SMT, as @liamwhite suggested.

### DIFF
--- a/src/common/x64/cpu_detect.cpp
+++ b/src/common/x64/cpu_detect.cpp
@@ -4,14 +4,27 @@
 
 #include <array>
 #include <cstring>
+#include <fstream>
 #include <iterator>
+#include <optional>
 #include <string_view>
+#include <thread>
+#include <vector>
 #include "common/bit_util.h"
 #include "common/common_types.h"
+#include "common/logging/log.h"
 #include "common/x64/cpu_detect.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#endif
 
 #ifdef _MSC_VER
 #include <intrin.h>
+
+static inline u64 xgetbv(u32 index) {
+    return _xgetbv(index);
+}
 #else
 
 #if defined(__DragonFly__) || defined(__FreeBSD__)
@@ -39,12 +52,11 @@ static inline void __cpuid(int info[4], u32 function_id) {
 }
 
 #define _XCR_XFEATURE_ENABLED_MASK 0
-static inline u64 _xgetbv(u32 index) {
+static inline u64 xgetbv(u32 index) {
     u32 eax, edx;
     __asm__ __volatile__("xgetbv" : "=a"(eax), "=d"(edx) : "c"(index));
     return ((u64)edx << 32) | eax;
 }
-
 #endif // _MSC_VER
 
 namespace Common {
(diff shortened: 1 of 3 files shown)


FearlessTobi:
Should this maybe into common/cpu_detect?

src/yuzu/main.cpp
```diff
@@ -286,6 +286,41 @@ bool GMainWindow::CheckDarkMode() {
...
+    int num_processors = 0;
+    for (const auto& information : list) {
+        if (information.Relationship == RelationProcessorCore) {
+            num_processors++;
+        }
+    }
+    return num_processors;
+#elif defined(__unix__)
+    const int thread_count = std::thread::hardware_concurrency();
+    std::ifstream smt("/sys/devices/system/cpu/smt/active");
```
v1993, line 316: This most definitely needs a change. Shouldn't you read contents contents of file to check if SMT is actually enabled (because you can turn it off by e.g. writing `0` to this very file)? And if you actually *don't* need to do that (seems unlikely to me), there are better ways for checking if files exists than trying to open it (not sure what exact FS API yuzu's code standards suggest to use here, I'd personally go with `std::filesystem::exists`).

v1993 (changes requested):

src/yuzu/main.cpp
goldenx86: What do you think of it now?

v1993: Looks reasonable. Have you verified that results are correct if you turn SMT off?

goldenx86: Yup, toast confirmed it with and without HT.

v1993 (approved):

goldenx86:
Moved to common/x64/cpu_detect.

src/common/x64/cpu_detect.cpp
```diff
@@ -192,4 +201,49 @@ const CPUCaps& GetCPUCaps() {
...
+            LOG_ERROR(Frontend, "Failed to query core count.");
+            return std::nullopt;
+        }
+    }
+    int num_processors = 0;
+    for (const auto& information : list) {
+        if (information.Relationship == RelationProcessorCore) {
+            num_processors++;
+        }
+    }
```
Morph1984, line 226:
```suggestion
    // Get the buffer length.
    std::vector<SYSTEM_LOGICAL_PROCESSOR_INFORMATION> buffer;
    DWORD length;
    GetLogicalProcessorInformation(nullptr, &length);
    if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
        LOG_ERROR(Frontend, "Failed to query core count.");
        return std::nullopt;
    }
    list.resize(length / sizeof(SYSTEM_LOGICAL_PROCESSOR_INFORMATION));
    // Now query the core count.
    if (!GetLogicalProcessorInformation(buffer, &length)) {
        LOG_ERROR(Frontend, "Failed to query core count.");
        return std::nullopt;
    }
    return std::count_if(buffer.cbegin(), buffer.cend(), [](const auto& proc_info) { return proc_info.Relationship == RelationProcessorCore;  });
```

lioncash (approved):

PR #9240, opened 2022-11-13
configure_input: Use unique pointers for player_controllers array

ameerj:
Avoids a potential memory leak since these objects never seem to be deleted


lioncash:
I'm pretty sure the parent UI object takes ownership over them once they're added to the parent layout via `addWidget`, which will delete the objects whenever the parent is destroyed. Using unique_ptr here can lead to double-frees

lioncash:
Actually, even before the `addWidget` call, each `ConfigureInputPlayer` instance would already be a child of the parent, since `this` is passed as a parent in the constructor, so these would already be handled by Qt from the get-go.

ameerj:
Ah, a qt quirk. Fair enough.

PR #9241, opened 2022-11-14
config: Always apply input profile's controller type

ameerj:
Previously, input profiles only applied the controller type if it was a `LeftJoycon` or `RightJoycon`. 

This change applies the profile's type for all controllers.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -195,27 +195,16 @@ void Config::ReadPlayerValue(std::size_t player_index) {
 
     auto& player = Settings::values.players.GetValue()[player_index];
 
-    if (player_prefix.isEmpty()) {
-        const auto controller = static_cast<Settings::ControllerType>(
-            qt_config
-                ->value(QStringLiteral("%1type").arg(player_prefix),
-                        static_cast<u8>(Settings::ControllerType::ProController))
-                .toUInt());
+    player.controller_type = static_cast<Settings::ControllerType>(
+        qt_config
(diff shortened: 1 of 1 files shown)


Morph1984:
iirc, the rationale behind only applying the controller type for single joycons is due to them not being able to share the same input mappings as the other controllers due to the different layout. So this allows different controller types to be used with the same profile

ameerj:
@Morph1984 Not sure I fully understand what you mean. Can you give an example input configuration where this change would be problematic?

Morph1984:
For example: An input profile created when a Pro Controller is selected can be re-used when Dual Joycons, Handheld, or GameCube Controller are selected. However, with this change, if the user has selected any of the aforementioned types it will load in the Pro Controller

ameerj:
It would make sense that the controller type that was chosen when the profile was saved would be selected when loaded along with the bindings. For cases where a user wants to more quickly load their Handheld profile for PLGE or their GameCube profile for SSBU, etc.

Morph1984:
While these are valid scenarios, I have not heard of users complaining of input profiles not saving the selected controller type. When this was implemented I too had to pick between genericity and specificity. I've seen input profiles being used more to store custom controller mappings rather than specifically for one game AND controller type. The specificity problem was supposed to be solved by per-game input profiles, which allowed selecting the controller type along side the profile, however I never really got to implementing it fully as I lost the code for it.

Morph1984:
Given that this behavior has been the default for a couple years, there isn't much of an incentive to change the behavior now, as it would result in unexpected results when users using input profiles would find that after selecting the controller type, selecting the input profile would change the controller type.

PR #9247, opened 2022-11-15
configure_profile_manager: Use a custom dialog when deleting a profile

lat9nq:
I was prompted to add this after a user deleted their user profile and found out that it also deleted all the associated save data for that user.

This adds a hopefully more informative dialog that most importantly notifies the user that their saves will be deleted with the user profile. It shows the profile icon, name, and UUID in the dialog.

Screenshot from 2022-11-15 17-41-23

While I was in this segment of the configuration, I also removed the frame from the profile picture in the manager, since it caused the image to be offset by 1px and cropped by 2px.

### DIFF
--- a/src/yuzu/configuration/configure_profile_manager.cpp
+++ b/src/yuzu/configuration/configure_profile_manager.cpp
@@ -115,7 +115,7 @@ ConfigureProfileManager::ConfigureProfileManager(const Core::System& system_, QW
             &ConfigureProfileManager::ConfirmDeleteUser);
     connect(ui->pm_set_image, &QPushButton::clicked, this, &ConfigureProfileManager::SetUserImage);
 
-    confirm_dialog = std::make_unique<ConfigureProfileManagerDeleteDialog>(this);
+    confirm_dialog = new ConfigureProfileManagerDeleteDialog(this);
 
     scene = new QGraphicsScene;
     ui->current_user_icon->setScene(scene);
@@ -246,7 +246,7 @@ void ConfigureProfileManager::ConfirmDeleteUser() {
     confirm_dialog->show();
 }
 
-void ConfigureProfileManager::DeleteUser(const Common::UUID uuid) {
+void ConfigureProfileManager::DeleteUser(const Common::UUID& uuid) {
     if (Settings::values.current_user.GetValue() == tree_view->currentIndex().row()) {
         Settings::values.current_user = 0;
     }
@@ -324,7 +324,7 @@ void ConfigureProfileManager::SetUserImage() {
 }
 
 ConfigureProfileManagerDeleteDialog::ConfigureProfileManagerDeleteDialog(QWidget* parent)
-    : QDialog(parent) {
+    : QDialog{parent} {
     auto dialog_vbox_layout = new QVBoxLayout(this);
     dialog_button_box =
         new QDialogButtonBox(QDialogButtonBox::Yes | QDialogButtonBox::No, Qt::Horizontal, parent);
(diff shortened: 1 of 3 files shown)


src/yuzu/configuration/configure_profile_manager.h
```diff
@@ -47,14 +67,17 @@ class ConfigureProfileManager : public QWidget {
     void SelectUser(const QModelIndex& index);
     void AddUser();
     void RenameUser();
-    void DeleteUser();
+    void ConfirmDeleteUser();
+    void DeleteUser(const Common::UUID uuid);
```
lioncash, line 71:
```suggestion
    void DeleteUser(const Common::UUID& uuid);
```

src/yuzu/configuration/configure_profile_manager.cpp
```diff
@@ -319,3 +322,47 @@ void ConfigureProfileManager::SetUserImage() {
                         new QStandardItem{GetIcon(*uuid), FormatUserEntryText(username, *uuid)});
     UpdateCurrentUser();
 }
+
+ConfigureProfileManagerDeleteDialog::ConfigureProfileManagerDeleteDialog(QWidget* parent)
+    : QDialog(parent) {
```
Morph1984, line 327:
```suggestion
    : QDialog{parent} {
```

src/yuzu/configuration/configure_profile_manager.h
```diff
@@ -3,16 +3,22 @@
...
 #pragma once
 
+#include <functional>
 #include <memory>
 
+#include <QDialog>
 #include <QList>
 #include <QWidget>
 
+#include "common/uuid.h"
```
Morph1984, line 13: Can be a forward declaration

src/yuzu/configuration/configure_profile_manager.h
```diff
@@ -26,6 +32,20 @@ namespace Ui {
 class ConfigureProfileManager;
 }
 
+class ConfigureProfileManagerDeleteDialog : public QDialog {
+public:
+    explicit ConfigureProfileManagerDeleteDialog(QWidget* parent);
+    ~ConfigureProfileManagerDeleteDialog();
+
+    void SetInfo(const QString username, const Common::UUID uuid,
```
Morph1984, line 40:
```suggestion
    void SetInfo(const QString username, const Common::UUID& uuid,
```

src/yuzu/configuration/configure_profile_manager.h
```diff
@@ -47,14 +67,17 @@ class ConfigureProfileManager : public QWidget {
...
+    void ConfirmDeleteUser();
+    void DeleteUser(const Common::UUID uuid);
     void SetUserImage();
 
     QVBoxLayout* layout;
     QTreeView* tree_view;
     QStandardItemModel* item_model;
     QGraphicsScene* scene;
 
+    std::unique_ptr<ConfigureProfileManagerDeleteDialog> confirm_dialog;
```
lioncash, line 79: This should probably not be a unique_ptr, since Qt will delete the dialog when ConfigureProfileManager is freed, this might result in a double free.

This can just be a regular pointer. It'll get handled by qt, since the dialog is given a parent when constructed

src/yuzu/configuration/configure_profile_manager.h
lioncash, line 40:
```suggestion
    void SetInfo(const QString& username, const Common::UUID& uuid,
```

goldenx86:
The things we do for users.

PR #9254, opened 2022-11-17
Dynarmic: Remove inaccurate NaN from Auto CPU settings.

FernandoS27:
Outsourced testing has shown that disabling this option improves stability in Pokemon Scarlet. I cannot verify it myself yet, so I'll leave it as "experimental".

### DIFF
--- a/src/core/arm/dynarmic/arm_dynarmic_64.cpp
+++ b/src/core/arm/dynarmic/arm_dynarmic_64.cpp
@@ -348,7 +348,6 @@ std::shared_ptr<Dynarmic::A64::Jit> ARM_Dynarmic_64::MakeJit(Common::PageTable* 
         if (Settings::values.cpu_accuracy.GetValue() == Settings::CPUAccuracy::Auto) {
             config.unsafe_optimizations = true;
             config.optimizations |= Dynarmic::OptimizationFlag::Unsafe_UnfuseFMA;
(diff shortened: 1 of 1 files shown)


src/core/arm/dynarmic/arm_dynarmic_64.cpp
```diff
@@ -348,7 +348,6 @@ std::shared_ptr<Dynarmic::A64::Jit> ARM_Dynarmic_64::MakeJit(Common::PageTable*
         if (Settings::values.cpu_accuracy.GetValue() == Settings::CPUAccuracy::Auto) {
             config.unsafe_optimizations = true;
             config.optimizations |= Dynarmic::OptimizationFlag::Unsafe_UnfuseFMA;
-            config.optimizations |= Dynarmic::OptimizationFlag::Unsafe_InaccurateNaN;
```
hugomota59, line 351: see this!

Janmm14:
What about making "auto" just actually smart, so it only disables it for certain games?

v1993:
> What about making "auto" just actually smart, so it only disables it for certain games?

Doing it automatically is not exactly feasible and maintaining a database with game-specific settings is something yuzu maintainers would like to avoid.

FernandoS27:
> What about making "auto" just actually smart, so it only disables it for certain games?

that's what per game settings are for.

bunnei (approved):

PR #9261, opened 2022-11-19
common: fiber: Use Win32 impl. for fibers.

bunnei:
Older implementation of fibers that uses Win32 implementation. Improves stability here -- tagging for broader testing.

### DIFF
--- a/src/common/fiber.cpp
+++ b/src/common/fiber.cpp
@@ -1,40 +1,132 @@
 // SPDX-FileCopyrightText: Copyright 2020 yuzu Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
-#include <mutex>
-
 #include "common/assert.h"
 #include "common/fiber.h"
-#include "common/virtual_buffer.h"
+#include "common/spin_lock.h"
 
+#if defined(_WIN32) || defined(WIN32)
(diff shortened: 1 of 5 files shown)


src/common/fiber.cpp
```diff
@@ -132,4 +218,5 @@ std::shared_ptr<Fiber> Fiber::ThreadToFiber() {
     return fiber;
 }
 
-} // namespace Common
+#endif
+} // namespace Common
```
v1993, line 222: Accidentally removed newline at EOF.

src/common/fiber.h
```diff
@@ -50,16 +52,26 @@ class Fiber {
...
     static void FiberStartFunc(boost::context::detail::transfer_t transfer);
     static void RewindStartFunc(boost::context::detail::transfer_t transfer);
+#endif
 
     struct FiberImpl;
     std::unique_ptr<FiberImpl> impl;
 };
 
-} // namespace Common
+} // namespace Common
```
v1993, line 77: Ditto.

v1993 (changes requested):

FearlessTobi:
Before merging we should check if this has any noticeable performance impact in games.

PR #9273, opened 2022-11-19
 Configuration: Add per-game input profiles 

ameerj:
This enables setting custom input profiles per-game. 

Per-game profiles also apply the controller type (i.e. Pro Controller, Handheld, GameCube) that was selected when the profile was saved.

This acts as a temporary compromise until proper Per-Game Input configurations are implemented.

image

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -124,6 +124,10 @@ void Config::Initialize(const std::string& config_name) {
     }
 }
 
+bool Config::IsCustomConfig() {
+    return type == ConfigType::PerGameConfig;
+}
+
 /* {Read,Write}BasicSetting and WriteGlobalSetting templates must be defined here before their
  * usages later in this file. This allows explicit definition of some types that don't work
  * nicely with the general version.
@@ -186,7 +190,7 @@ void Config::WriteGlobalSetting(const Settings::SwitchableSetting<Type, ranged>&
 
 void Config::ReadPlayerValue(std::size_t player_index) {
     const QString player_prefix = [this, player_index] {
-        if (type == ConfigType::InputProfile && global) {
+        if (type == ConfigType::InputProfile) {
             return QString{};
         } else {
             return QStringLiteral("player_%1_").arg(player_index);
@@ -194,8 +198,20 @@ void Config::ReadPlayerValue(std::size_t player_index) {
     }();
 
     auto& player = Settings::values.players.GetValue()[player_index];
+    if (IsCustomConfig()) {
+        const auto profile_name =
+            qt_config->value(QStringLiteral("%1profile_name").arg(player_prefix), QString{})
+                .toString()
+                .toStdString();
+        if (profile_name.empty()) {
+            // Use the global input config
+            player = Settings::values.players.GetValue(true)[player_index];
+            return;
+        }
+        player.profile_name = profile_name;
+    }
 
-    if (player_prefix.isEmpty()) {
+    if (player_prefix.isEmpty() && Settings::IsConfiguringGlobal()) {
         const auto controller = static_cast<Settings::ControllerType>(
             qt_config
                 ->value(QStringLiteral("%1type").arg(player_prefix),
@@ -244,14 +260,6 @@ void Config::ReadPlayerValue(std::size_t player_index) {
                 ->value(QStringLiteral("%1button_color_right").arg(player_prefix),
                         Settings::JOYCON_BUTTONS_NEON_RED)
                 .toUInt();
-
-        // This only applies to per-game configs
-        if (!global) {
-            player.profile_name =
-                qt_config->value(QStringLiteral("%1profile_name").arg(player_prefix), QString{})
-                    .toString()
-                    .toStdString();
-        }
     }
 
     for (int i = 0; i < Settings::NativeButton::NumButtons; ++i) {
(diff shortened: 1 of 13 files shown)


bunnei:
Nice to see this change! Tagging for EA to get more eyes on this.

src/yuzu/configuration/configure_input_per_game.cpp
```diff
@@ -0,0 +1,115 @@
+// SPDX-FileCopyrightText: 2016 Citra Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
```
Tachi107, line 2: Do these files really come from Citra or this is just a copy-paste thing?

ameerj: Good catch!

BSoDGamingYT:
This change while it seems great, doesn't seem to work as one would expect. 

For example...I am testing Pokemon Lets' Go, which only works with the handheld. So I make a profile for Handheld > select my controller > save the profile. 

Now when I load the game, it loads the profile and sets me to Handheld...but there is no input because there is no stored input device. For it to work, I need to go to configure > controls > manually set the input device (Dualsense controller)

german77:
This PR seems to always select Handheld regardless of the current config or per game profile. From I can tell `IsCustomConfig()` always returns true even when no custom profile is set.

ameerj:
> This PR seems to always select Handheld 

Should be fixed now

> Also make the window a bit wider so it fits all tabs.

I don't have this issue on my monitor. If you have a fix feel free to push to this PR.

src/yuzu/configuration/config.cpp
```diff
@@ -388,9 +404,26 @@ void Config::ReadAudioValues() {
...
         ReadPlayerValue(p);
     }
+    ReadGlobalSetting(Settings::values.use_docked_mode);
+
+    // Disable docked mode if handheld is selected
+    const auto controller_type = Settings::values.players.GetValue()[0].controller_type;
+    if (controller_type == Settings::ControllerType::Handheld) {
+        Settings::values.use_docked_mode.SetGlobal(!IsCustomConfig());
+        Settings::values.use_docked_mode.SetValue(false);
+    }
```
german77, line 418:
```suggestion
    if (controller_type == Settings::ControllerType::Handheld) {
        Settings::values.use_docked_mode.SetValue(false);
    }
```
Altering the global state on read makes difficult to track since config file doesn't match with what really happens.

src/yuzu/configuration/config.cpp
```diff
@@ -388,9 +404,26 @@ void Config::ReadAudioValues() {
...
+    Settings::values.players.SetGlobal(!IsCustomConfig());
     for (std::size_t p = 0; p < Settings::values.players.GetValue().size(); ++p) {
         ReadPlayerValue(p);
     }
+    ReadGlobalSetting(Settings::values.use_docked_mode);
+
+    // Disable docked mode if handheld is selected
+    const auto controller_type = Settings::values.players.GetValue()[0].controller_type;
+    if (controller_type == Settings::ControllerType::Handheld) {
+        Settings::values.use_docked_mode.SetGlobal(!IsCustomConfig());
```
german77, line 416:
```suggestion
```
I did a bit of testing. We don't need to change the global setting of docked mode. Since if a handheld profile is selected it will always be overridden with `docked=false` no matter the config. Otherwise we keep weird behaviors like always having handheld mode enabled as I described earlier.

If you really want to store the docked mode. I suggest to add an actual checkbox to Qt.

src/yuzu/configuration/config.cpp
```diff
@@ -1087,20 +1119,25 @@ void Config::SaveAudioValues() {
 void Config::SaveControlValues() {
     qt_config->beginGroup(QStringLiteral("Controls"));
 
+    Settings::values.players.SetGlobal(!IsCustomConfig());
     for (std::size_t p = 0; p < Settings::values.players.GetValue().size(); ++p) {
         SavePlayerValue(p);
     }
+    WriteGlobalSetting(Settings::values.use_docked_mode);
```
german77, line 1126:
```suggestion
    // Disable docked mode if handheld is selected
    const auto controller_type = Settings::values.players.GetValue()[0].controller_type;
    if (controller_type == Settings::ControllerType::Handheld) {
        Settings::values.use_docked_mode.SetGlobal(!IsCustomConfig());
        Settings::values.use_docked_mode.SetValue(false);
    } else {
        Settings::values.use_docked_mode.SetGlobal(true);
    }
    WriteGlobalSetting(Settings::values.use_docked_mode);
```
We can set the global state here when handheld is selected. To allow setting and restoring the config when the game is running or stopped.  

If any other profile is used other than handheld we should keep the global value. Otherwise we need to display this setting on the Qt interface to notify the user that it will be overridden by the custom config .

german77 (changes requested):
Everything looks fine but docked mode needs to be fixed. See the individual comments.

german77 (approved):

v1993 (changes requested):
Qt complains about duped entries yet again...

```
Automatic MOC and UIC for target yuzu
AutoUic: /home/v/compile/yuzu/src/yuzu/configuration/configure_input_per_game.ui: Warning: The name 'input_profile_layout' (QHBoxLayout) is already in use, defaulting to 'input_profile_layout1'.
/home/v/compile/yuzu/src/yuzu/configuration/configure_input_per_game.ui: Warning: The name 'input_profile_layout' (QHBoxLayout) is already in use, defaulting to 'input_profile_layout2'.
/home/v/compile/yuzu/src/yuzu/configuration/configure_input_per_game.ui: Warning: The name 'input_profile_layout' (QHBoxLayout) is already in use, defaulting to 'input_profile_layout3'.
/home/v/compile/yuzu/src/yuzu/configuration/configure_input_per_game.ui: Warning: The name 'input_profile_layout' (QHBoxLayout) is already in use, defaulting to 'input_profile_layout4'.
/home/v/compile/yuzu/src/yuzu/configuration/configure_input_per_game.ui: Warning: The name 'input_profile_layout' (QHBoxLayout) is already in use, defaulting to 'input_profile_layout5'.
/home/v/compile/yuzu/src/yuzu/configuration/configure_input_per_game.ui: Warning: The name 'input_profile_layout' (QHBoxLayout) is already in use, defaulting to 'input_profile_layout6'.
/home/v/compile/yuzu/src/yuzu/configuration/configure_input_per_game.ui: Warning: The name 'input_profile_layout' (QHBoxLayout) is already in use, defaulting to 'input_profile_layout7'.
```

PR #9297, opened 2022-11-22
[audio_core] Fix an OoB with sample sinking

Kelebek1:
Input channels don't start at index 0. In SSHD there's 6 channels with indexes from 0x8 - 0xD, so when only allocating a 6 channel span size, the range is limited to `sample count * 6`, but when sinking it then tries to access `sample count * 8` for the first channel and beyond. So instead just check for the max input index and add 1.

No functional change here, just fixing an error when building with _FORTIFY_SOURCE since the span access was OoB.


MysticExile:
Fixes #9287

liamwhite (approved):

PR #9300, opened 2022-11-23
CMake: Use precompiled headers to improve compile times

ameerj:
This change introduces Precompiled Headers (PCH) for some of the more commonly included headers which take the longest to compile.

Each project/CMake target defines its own PCH in its `precompiled_headers.h` file. 
This leads to repetitive includes and recompilation of the same headers in multiple projects, but this is needed since many compilers require that PCH are compiled with same compiler flags as the source files including them. 
This also adds the flexibility of adding/removing headers in the PCH as needed for each project.

On MSVC I saw a ~50% build time reduction. Other compilers may fare better or worse.

PCH is disabled on MSVC when using buildcache for now due to compilation errors.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -82,8 +82,9 @@ if (MSVC)
         /wd4324 # 'struct_name': structure was padded due to __declspec(align())
     )
 
-    if (USE_CCACHE)
+    if (USE_CCACHE OR YUZU_USE_PRECOMPILED_HEADERS)
     # when caching, we need to use /Z7 to downgrade debug info to use an older but more cachable format
+    # Precompiled headers are deleted if not using /Z7. See https://github.com/nanoant/CMakePCHCompiler/issues/21
         add_compile_options(/Z7)
     else()
         add_compile_options(/Zi)
@@ -149,6 +150,10 @@ else()
         set(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> qcTP <TARGET> <LINK_FLAGS> <OBJECTS>")
         set(CMAKE_CXX_ARCHIVE_APPEND "<CMAKE_AR> qTP  <TARGET> <LINK_FLAGS> <OBJECTS>")
     endif()
+endif()
+
+if (YUZU_USE_PRECOMPILED_HEADERS)
+    set(CMAKE_PCH_INSTANTIATE_TEMPLATES ON)
 endif()
 
 add_subdirectory(common)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -152,10 +152,6 @@ else()
     endif()
 endif()
 
-if (YUZU_USE_PRECOMPILED_HEADERS)
-    set(CMAKE_PCH_INSTANTIATE_TEMPLATES ON)
-endif()
-
 add_subdirectory(common)
 add_subdirectory(core)
 add_subdirectory(audio_core)
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -37,6 +37,7 @@ add_library(common STATIC
     cache_management.cpp
     cache_management.h
     common_funcs.h
+    common_precompiled_headers.h
     common_types.h
     concepts.h
     div_ceil.h
@@ -187,5 +188,4 @@ endif()
 
 if (YUZU_USE_PRECOMPILED_HEADERS)
     target_precompile_headers(common PRIVATE precompiled_headers.h)
-    set(CMAKE_PCH_INSTANTIATE_TEMPLATES ON)
 endif()
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -95,6 +95,7 @@ add_library(common STATIC
     param_package.h
     parent_of_member.h
     point.h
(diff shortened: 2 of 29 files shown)


src/precompiled_headers/precompiled_headers.h
```diff
@@ -0,0 +1,8 @@
+// SPDX-FileCopyrightText: 2022 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <chrono>
```
lioncash, line 6: You would likely want to add `<algorithm>`  and <memory>`  alongside this one, since both headers are usually quite heavyweight (especially the algorithm header)

src/precompiled_headers/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: 2022 yuzu Emulator Project
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+add_library(precompiled_headers STATIC
+    precompiled_headers.cpp
+    precompiled_headers.h
+)
+set_target_properties(precompiled_headers PROPERTIES LINKER_LANGUAGE CXX)
+
+target_link_libraries(precompiled_headers PRIVATE fmt::fmt)
```
lioncash, line 10: was it intended for `<fmt/format.h>` to be added to the precompiled header but was forgotten when making the header? Otherwise shouldn't this not be linked in?

ameerj: `<fmt/format.h>` is included by assert.h, so it needed to be linked. But I will add `<fmt/format.h>` explicitly for clarity. Thanks for catching this.

src/common/CMakeLists.txt
```diff
@@ -183,3 +183,10 @@ else()
   target_link_libraries(common PRIVATE
     $<IF:$<TARGET_EXISTS:zstd::libzstd_shared>,zstd::libzstd_shared,zstd::libzstd_static>)
 endif()
+
+if (YUZU_USE_PRECOMPILED_HEADERS)
+    # Resuing from the precompiled_headers leads to an error detecting _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING
```
Morph1984, line 188:
```suggestion
    # Reusing from the precompiled_headers leads to an error detecting _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING
```

src/CMakeLists.txt
```diff
@@ -80,10 +80,12 @@ if (MSVC)
 
         /wd4100 # 'identifier': unreferenced formal parameter
         /wd4324 # 'struct_name': structure was padded due to __declspec(align())
+        /wd4652 # compiler option 'Warning level (/W<n>)' inconsistent with precompiled header
```
Morph1984, line 83:
```suggestion
        /wd4652 # compiler option 'option' inconsistent with precompiled header
```

src/core/CMakeLists.txt
```diff
@@ -820,3 +820,14 @@ if (ARCHITECTURE_x86_64 OR ARCHITECTURE_arm64)
     )
     target_link_libraries(core PRIVATE dynarmic)
 endif()
+
+if (YUZU_USE_PRECOMPILED_HEADERS)
+    target_precompile_headers(core PRIVATE
+        <algorithm>
```
lioncash, line 826: Is it possible to have a precompiled_header.h file inside the core library's source tree for this? It'd make expanding the header in the future a little nicer, since the build system script files wouldn't need to be touched and source files wouldn't need to be regenerated.

It'd also keep documentation about source file organization inside the source tree and out of the build system

It'd also keep documenting details about the source outside of the build system too

Tachi107:
> PCH is disabled on MINGW and Clang for now due to compilation errors. Perhaps someone with better access to these compilers can investigate if there is interest.

I haven't used precompiled headers a lot, but I seem to remember that on GCC (and probably also Clang) all the consumers of a given PCH must use mostly the same compiler flags used by the producer. Maybe this is the cause of the errors?

ameerj:
@Tachi107 the clang issues are resolved, but the MINGW error does not seem to be caused by compiler flags.
https://github.com/yuzu-emu/yuzu/actions/runs/3528711813/jobs/5919099882
https://pastebin.com/pbGdUa55

src/common/common_headers.h
```diff
@@ -0,0 +1,12 @@
+// SPDX-FileCopyrightText: 2022 yuzu Emulator Project
```
lioncash, line 1: This file name seems a little to generic and should likely be named `common_precompiled_headers`, since it's main use is for precompiled headers.

src/common/string_util.cpp
```diff
@@ -141,7 +141,7 @@ static std::wstring CPToUTF16(u32 code_page, const std::string& input) {
         MultiByteToWideChar(code_page, 0, input.data(), static_cast<int>(input.size()), nullptr, 0);
 
     if (size == 0) {
-        return L"";
+        return std::wstring(size, L'\0');
```
lioncash, line 144: Wouldn't this and the corresponding variant below be able to be replaced with just:

```cpp
return {};
```

?

liamwhite (approved):

PR #9315, opened 2022-11-26
Make configure dialog non-blocking

jaihysc:
Currently, when the configure dialog is open, it prevents inputs to the main window. This is annoying if you are trying to change your keybinds as you have to constantly reopen the configure dialog to test your keys.

To make the window non-blocking, we use .show() instead of .exec() on the configure dialog, and have a slot handle when the confirmation buttons (ok, cancel, x) are pressed.

However, entering the configuration dialog disables inputs to prevent sending inputs to the running game while changing inputs. We address  this by preventing inputs from being sent to the game only when focus is not on the main menu by calling `EmulatedController::EnableConfiguration()` and `EmulatedController::DisableConfiguration()` on focus in and focus out on he main window.

### DIFF
--- a/src/yuzu/configuration/configure_dialog.cpp
+++ b/src/yuzu/configuration/configure_dialog.cpp
@@ -117,6 +117,14 @@ void ConfigureDialog::ApplyConfiguration() {
     Settings::LogSettings();
 }
 
+void ConfigureDialog::MainWindowFocusIn() {
+    input_tab->DisableConfiguration();
+}
+
+void ConfigureDialog::MainWindowFocusOut() {
+    input_tab->EnableConfiguration();
+}
+
 void ConfigureDialog::changeEvent(QEvent* event) {
     if (event->type() == QEvent::LanguageChange) {
         RetranslateUI();
--- a/src/yuzu/configuration/configure_dialog.h
+++ b/src/yuzu/configuration/configure_dialog.h
@@ -46,6 +46,10 @@ public:
 
     void ApplyConfiguration();
 
+public slots:
+    void MainWindowFocusIn();
+    void MainWindowFocusOut();
+
 private slots:
     void OnLanguageChanged(const QString& locale);
 
--- a/src/yuzu/configuration/configure_input.cpp
+++ b/src/yuzu/configuration/configure_input.cpp
@@ -181,6 +181,18 @@ void ConfigureInput::Initialize(InputCommon::InputSubsystem* input_subsystem,
     LoadConfiguration();
 }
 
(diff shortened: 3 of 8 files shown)


Kelebek1:
Not directly relevant to this PR I guess, but when the keyboard app is open you also can't interact with the title bar at all. Would be great if you were able to fix that one too.

jaihysc:
Sorry, I am not too familiar with Yuzu, what is this keyboard app you are referring to?

Kelebek1:
> Sorry, I am not too familiar with Yuzu, what is this keyboard app you are referring to?

When you're entering a name in any game, the keyboard pops up and you can type.

jaihysc:
> When you're entering a name in any game, the keyboard pops up and you can type.

I see, I will look into it.

There seems to be an issue, the first time the configuration is opened, you cannot make inputs to the game, after opening the configuration a second time inputs can be made again. It occurs only the first time the configuration is opened.

german77:
The configure window It's supposed to disable input intentionally by calling  `EmulatedController::EnableConfiguration()` this mode prevents any controller input from reaching the game while you are configuring your mappings.

The controller image at the center displays any input that is configured. So you use that to test and confirm that your mappings are correct. Keyboard is the only input source that doesn't work in this mode due to being used to navigate through the Qt window. You can certainly hook keyboard input to be always active on that window but it wouldn't be a nice experience to be moving around Qt elements while you try to test.

This PR is effectively killing the initialization process on subsequent launches and might cause unexpected issues on the future. The reason second time always work it's because on save `EmulatedController::DisableConfiguration()` is called to allow inputs to the game and the next time you open the window input is not disabled because the configuration window was already initialized

jaihysc:
> The configure window It's supposed to disable input intentionally by calling `EmulatedController::EnableConfiguration()` this mode prevents any controller input from reaching the game while you are configuring your mappings. ...

Would it be feasible if we made it so that it only blocks the inputs while the configure menu is in focus by calling `EmulatedController::EnableConfiguration()` and `EmulatedController::DisableConfiguration()`? Or would there be additional issues involved?

jaihysc:
I fixed the calls to `ConfigureDialog` so it always destructs first then constructs. Regarding the inputs, I changed it such that entering and leaving focus calls `EmulatedController::EnableConfiguration()` and `EmulatedController::DisableConfiguration()`.

jaihysc:
@german77 Could you please review the new changes again? I addressed the issues you outlined.

WhyNotHugo:
Looks like github auto-closed this due to the "fork" being deleted.

This issue is still relevant tho.

PR #9325, opened 2022-11-27
yuzu-cmd: Fix default config value

german77:
`sdl2_config->Get(...)` will only return the default value if the setting doesn't exist. However if it's empty it will not use the default value. This solves lack of audio at boot.


liamwhite (approved):

src/yuzu_cmd/config.cpp
```diff
@@ -90,7 +90,11 @@ static const std::array<std::array<int, 5>, Settings::NativeAnalog::NumAnalogs>
 
 template <>
 void Config::ReadSetting(const std::string& group, Settings::Setting<std::string>& setting) {
-    setting = sdl2_config->Get(group, setting.GetLabel(), setting.GetDefault());
+    std::string setting_value = sdl2_config->Get(group, setting.GetLabel(), setting.GetDefault());
+    if (setting_value.empty()) {
+        setting_value = setting.GetDefault();
+    }
+    setting = setting_value;
```
lioncash, line 97:
```suggestion
    setting = std::move(setting_value);
```

PR #9330, opened 2022-11-27
exclusive_monitor: Don't lock when disabled

merryhime:
Almost nothing makes use of this interface so I don't see this affecting performance much.


liamwhite (approved):

merryhime:
This doesn't work (requires access to internal state).

PR #9353, opened 2022-11-29
video_core: Fine tuning the index drawing judgment logic

vonchenplus:
Previously, we would set vertex(index)_count to 0 after each draw, because we needed to determine if it was an index draw based on the value. 
But now it turns out that macro will initiate some draw calls that rely on the previous value of vertex(index)_count, so we have modified the current logic.

Fixed particle in xc3, but some particles were still found to be faulty.


goldenx86:
Seems to break MK8.

vonchenplus:
> 

What's the regression?

Morph1984:
Also fix the typo on the commit and title

gidoly:
untag until graphical regression on super mario sunshine and galaxy resolve

liamwhite (approved):

PR #9360, opened 2022-11-30
Respect render mode override

Kelebek1:
Noticed that macros interact with this reg, may as well use it.


gidoly:
ready to tagg ea after other graphic regarding pr merges

liamwhite (approved):

PR #9395, opened 2022-12-06
cmake: correct find modules

abouvier:
The alias creation order was wrong. An alias for the imported target could have been created before the library version was validated by `find_package_handle_standard_args`. Now the alias is created only after the library has been found with a compatible version.

The GLOBAL keyword is also removed because [CMake 3.22 can now create alias of non-global imported target](https://cmake.org/cmake/help/v3.22/command/add_library.html#alias-libraries).


Tachi107:
> The GLOBAL keyword is also removed because CMake 3.22 can now create alias of non-global imported target.

Just a note: if I recall correctly, GLOBAL was needed in some cases because the target was created in the externals directory but accessed in a parent directory.

abouvier:
> Just a note: if I recall correctly, GLOBAL was needed in some cases because the target was created in the externals directory but accessed in a parent directory.

Yes, but all the `find_package` are now in the root `CMakeLists.txt`, where they should be :p

liamwhite (approved):

externals/find-modules/FindOpus.cmake
```diff
@@ -1,19 +1,17 @@
...
-find_package(PkgConfig)
-
+find_package(PkgConfig QUIET)
 if (PKG_CONFIG_FOUND)
-    pkg_search_module(opus IMPORTED_TARGET GLOBAL opus)
-    if (opus_FOUND)
-        add_library(Opus::opus ALIAS PkgConfig::opus)
-    endif()
+    pkg_search_module(OPUS QUIET IMPORTED_TARGET opus)
 endif()
```
Tachi107, line 7: @abouvier  just wondering: is checking for `PKG_CONFIG_FOUND` necessary? Couldn't we simply call `pkg_search_module()` unconditionally? If no pkg-config executable is available, the module should simply result as not found, right?

abouvier: Yeah, but if for whatever reason the `FindPkgConfig.cmake` module is not installed, both the macro and the variable (`find_package` only define `PKGCONFIG_FOUND`) will be undefined. Calling an undefined macro is an error, but testing an undefined variable is not.

Tachi107: It cannot be unavailable, it is part of CMake itself. By the same logic we should be checking that `find_package_handle_standard_args()` is defined before calling it, or I could even say "what if `PKG_CONFIG_FOUND` was set by something other than the FindPkgConfig module, and the module itself is unavailable?", but we shouldn't be checking for things that go against the documented behaviour ;)

abouvier: It can be made unavailable with `cmake (...) -DCMAKE_DISABLE_FIND_PACKAGE_PkgConfig=ON` :p 
And the existence of `find_package_handle_standard_args` does not need to be checked because `include(FindPackageHandleStandardArgs)` without the `OPTIONAL` keyword will raise a fatal error if the file does not exist.

PR #9400, opened 2022-12-07
Update README.md: add information about supported Windows versions

SHOJI122:
Add information about supported Windows versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ It is written in C++ with portability in mind, and we actively maintain builds f
 ## Compatibility
 
 The emulator is capable of running most commercial games at full speed, provided you meet the [necessary hardware requirements](https://yuzu-emu.org/help/quickstart/#hardware-requirements).
-Windows users should note that only Windows 10 or newer is supported.
+Windows users should note that only Windows 10 v1803 or newer is supported because of the use of the [Fastmem technology](https://yuzu-emu.org/entry/yuzu-fastmem/), and a custom build without Fastmem might still work for older systems like Windows 7 but is not supported.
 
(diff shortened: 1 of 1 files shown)


piplup55:
technically only windows 10 1803 is supported for fastmem

Norstyrm:
Also if you don't consider fastmem, Windows 7 is still _supported_ afaik, no?

piplup55:
> Also if you don't consider fastmem, Windows 7 is still _supported_ afaik, no?

technically yes but you would have some issues
image

SHOJI122:
Thank you for telling me about the Fastmem technology. I didn't know that, and this information would be help me much about why >Windows 10 v1803 is needed. I have added it in the PR.

piplup55 (commented):
fastmem might get disabled by default if it can't assign 4gb in windows

SHOJI122:
> fastmem might get disabled by default if it can't assign 4gb in windows, custom builds shouldn't be mentioned/discussed in the readme imo other then that it looks fine

Thanks for the suggestion! I didn't know that there are cases that Fastmem would get disabled. I have further optimised the patch.

FearlessTobi:
Just a suggestion, but maybe we should put all of this information on the hardware requirements wiki page instead in order to keep our Readme clean?

SHOJI122:
> Just a suggestion, but maybe we should put all of this information on the hardware requirements wiki page instead in order to keep our Readme clean?

I agree. I firstly read wiki on the website. However, when I clicked "edit this page" I came here and found that I have no permission to edit the wiki page. :(

PR #9433, opened 2022-12-14
build: tweak the find modules even more

Tachi107:
As described in https://github.com/yuzu-emu/yuzu/pull/9395#discussion_r1047456172 checking for `PKG_CONFIG_FOUND` before calling `pkg_search_module()` is unneeded, and some find modules (like [FindFFmpeg.cmake](https://github.com/yuzu-emu/yuzu/blob/a4696285af946588dc33b19e49a3baa0f8b2b60d/externals/find-modules/FindFFmpeg.cmake#L101-L102)) don't do this already. Consequently, this patch removes these checks.

I've sightly corrected the Findhttplib module, changing the `REQUIRED_VAR` from `HTTPLIB_LINK_LIBRARIES` to `HTTPLIB_FOUND`. Instead of explaining why in the commit message I did so in the module itself, so that the reasoning doesn't get lost and the same mistake isn't committed again.

Cc: @abouvier (does this make sense?)


liamwhite (approved):

abouvier:
I think you mixed up `HTTPLIB_INCLUDE_DIRS` and `HTTPLIB_INCLUDEDIR`.
I don't have a pc file for cpp-httplib, but with other libraries `<XXX>_INCLUDEDIR` seems defined even when the value is `/usr/include`.

Tachi107:
> I think you mixed up `HTTPLIB_INCLUDE_DIRS` and `HTTPLIB_INCLUDEDIR`.

You're right, dumb me. I've reverted that change

PR #9441, opened 2022-12-14
Revert "hle: service: audio: Use default service thread."

bunnei:
Reverts yuzu-emu/yuzu#9232


bunnei:
Reports that this causes audio stutter

PR #9452, opened 2022-12-16
hle_ipc: Refactor ReadBuffer to set buffer size upon initialization

ameerj:
Initializing the vector size during initialization is more efficient than a later call to resize()

Performance comparison:
https://quick-bench.com/q/oVUVQBssJFkZieTrmqsKX81qpO4


german77:
Since it always returns wouldn't be better to just remove the else block. It leaves the code a bit cleaner

liamwhite (approved):

PR #9462, opened 2022-12-17
Make player input configuration scrollable

behunin:
Because I had to add a widget context for the scroll area, I indented the ui file by one space and that made the diff look ugly and confusing so here is exactly what else happened there.

Removed the minimumSize property and used instead minimumContentsLength to size the profile combo box with length of 20.
The  ZL/RL threshold sliders max height was increased to 20 from 18.

Supersedes #6714


german77:
Windows has a selectable border and the default size crops the controller image. No scrollbars available at any resolution.
image

Linux has the scrollbar displayed by default and seems to refuse getting as compact as windows equivalent.
image

behunin:
That windows controller image looks huge haha
Steamdeck users are the ones that will really benefit from this PR.
That and small laptops

german77:
Was something wrong? Or you are going to aim to for a different approach?

PR #9485, opened 2022-12-21
Disable automatically opening the console on windows yuzu-cmd builds

ChrisOboe:
This change builds yuzu-cmd on windows as "windows" application (as opposed to "console" application which it did before). 
This leads to that the console doesn't start up by default anymore. When opening in an existing console session it will output to stdout and stderr just like before. 

Here are my reasons why i did this patch:
- Opening a console by default is wierd windows specific behaviour. No other OS i know opens a terminal emulator by default when you run a software. Now yuzu-cmd on windows behaves the same as on linux or mac. 
- There was no way (i know) of preventing opening the console window on windows with yuzu-cmd, this lead to interference with steam-in-home-streaming (which i use a lot) when running yuzu-cmd not in fullscreen. 

This works fine with the visual-studio build. And it should work with the mingw build too. Sadly i couldn't test the mingw build, since i couldn't get it compiled with mingw because of some unreleated libusb error. 

This is my first pr to yuzu. I hope i did everything right.


lioncash (approved):

goldenx86:
Hi. Can I mention you and this PR in the next progress report?

ChrisOboe:
@goldenx86 Hi, yes of course.

PR #9489, opened 2022-12-22
TAS: Increase accuracy of Stick inputs

MonsterDruide1:
As TAS-scripts provide raw inputs for the system, they should not be sanitized by Yuzu. As the function responsible for doing that (https://github.com/yuzu-emu/yuzu/blob/bbb963a31a141979b27615b9c28bcc72be50468e/src/core/hid/input_converter.cpp#L347) cannot be worked around, the simplest way to fix the current inaccuracies is by just hardcoding the range and deadzone to full and none, meaning the sanitization will not actually perform any changes on the data.

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -210,6 +210,13 @@ void EmulatedController::LoadTASParams() {
     tas_stick_params[Settings::NativeAnalog::LStick].Set("axis_y", 1);
     tas_stick_params[Settings::NativeAnalog::RStick].Set("axis_x", 2);
     tas_stick_params[Settings::NativeAnalog::RStick].Set("axis_y", 3);
+
+    // set to optimal stick to avoid sanitizing the stick and tweaking the coordinates
+    // making sure they play back in the game as originally written down in the script file
+    tas_stick_params[Settings::NativeAnalog::LStick].Set("deadzone", 0.0f);
+    tas_stick_params[Settings::NativeAnalog::LStick].Set("range", 1.0f);
+    tas_stick_params[Settings::NativeAnalog::RStick].Set("deadzone", 0.0f);
+    tas_stick_params[Settings::NativeAnalog::RStick].Set("range", 1.0f);
 }
 
 void EmulatedController::LoadVirtualGamepadParams() {


src/input_common/input_poller.cpp
```diff
@@ -832,9 +832,15 @@ std::unique_ptr<Common::Input::InputDevice> InputFactory::CreateHatButtonDevice(
 
 std::unique_ptr<Common::Input::InputDevice> InputFactory::CreateStickDevice(
     const Common::ParamPackage& params) {
-    const auto deadzone = std::clamp(params.Get("deadzone", 0.15f), 0.0f, 1.0f);
-    const auto range = std::clamp(params.Get("range", 0.95f), 0.25f, 1.50f);
+    auto deadzone = std::clamp(params.Get("deadzone", 0.15f), 0.0f, 1.0f);
+    auto range = std::clamp(params.Get("range", 0.95f), 0.25f, 1.50f);
     const auto threshold = std::clamp(params.Get("threshold", 0.5f), 0.0f, 1.0f);
+
+    if (input_engine.get()->GetEngineName() == "tas") {
```
lioncash, line 839:
```suggestion
    if (input_engine->GetEngineName() == "tas") {
```

src/input_common/input_poller.cpp
```diff
@@ -832,9 +832,15 @@ std::unique_ptr<Common::Input::InputDevice> InputFactory::CreateHatButtonDevice(
...
+        deadzone = 0;
```
lioncash, line 840:
```suggestion
        deadzone = 0.0f;
```

src/input_common/input_poller.cpp
```diff
@@ -832,9 +832,15 @@ std::unique_ptr<Common::Input::InputDevice> InputFactory::CreateHatButtonDevice(
...
+        range = 1;
```
lioncash, line 841:
```suggestion
        range = 1.0f;
```

lioncash:
Should probably also leave a comment in the source for why the TAS input-engine is being special-cased

german77:
Please avoid custom workarounds. TAS is hardcoded.  Just define the correct config here https://github.com/yuzu-emu/yuzu/blob/master/src/core/hid/emulated_controller.cpp#L175

MonsterDruide1:
That's exactly what I was looking for, thanks! Changed accordingly.

german77 (approved):

PR #9490, opened 2022-12-22
texture_cache: Use pre-allocated heap buffer for texture swizzles

ameerj:
Currently, heap memory is allocated and deallocated whenever a texture upload/download occurs. This can have significant overhead that is noticeable as stutters, especially since the texture data can be in the order of a few megabytes.

To alleviate this, a pre-allocated buffer is used to hold the swizzle data. This will reduce dynamic memory allocations, and should lead to less stutter as new textures are loaded, or when the GPU Garbage Collector is evicting images.

This can be further improved by using the ScratchBuffer from #9453


gidoly:
while performance remain almost same reduces stutters on texture heavy games such as xenoblade chronicles games, astral chain, kirby the forgotten land

liamwhite (approved):

bunnei (approved):

PR #9495, opened 2022-12-23
yuzu: Automatically refresh device list

german77:
We don't need a refresh button if we always keep the list updated.

### DIFF
--- a/src/yuzu/configuration/configure_input_player.cpp
+++ b/src/yuzu/configuration/configure_input_player.cpp
@@ -738,12 +738,9 @@ ConfigureInputPlayer::ConfigureInputPlayer(QWidget* parent, std::size_t player_i
 
     connect(ui->comboDevices, qOverload<int>(&QComboBox::activated), this,
             &ConfigureInputPlayer::UpdateMappingWithDefaults);
+    ui->comboDevices->installEventFilter(this);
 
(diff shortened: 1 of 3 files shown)


src/yuzu/configuration/configure_input_player.cpp
```diff
@@ -1479,6 +1476,16 @@ void ConfigureInputPlayer::keyPressEvent(QKeyEvent* event) {
     }
 }
 
+bool ConfigureInputPlayer::eventFilter(QObject* f_object, QEvent* f_event) {
+    if (f_object == ui->comboDevices) {
+        if (f_event->type() == QEvent::MouseButtonPress) {
```
FearlessTobi, line 1481: You can just merge those if statements into one and avoid one of the return statements.

FearlessTobi:
Also another consideration: 
Is querying for input devices blocking? We had an issue with loading audio devices for example, as it could take a long time on some platforms.

german77:
Since kraken most input engines don't block the thread. Only custom joycons and SDL have a bit of delay on basically getting ownership of the mutex. I can't notice any meaningful delay, 3.6ms if you want to have some numbers.

liamwhite (approved):

bunnei (approved):

PR #9514, opened 2022-12-28
settings: warn on invalid regon/language combinations

ColinKinloch:
I was trying to figure out why the temperature indicator in BotW was in Fahrenheit.

Since British English is at index 13 it's easy to overlook.

This branch changes the "English" language option to "American English".

It also adds a warning to the settings window when the selected language isn't valid for the selected region.
This warning doesn't appear when the region is set to China, as that isn't an option on my switch.

### DIFF
--- a/src/yuzu/configuration/configure_system.cpp
+++ b/src/yuzu/configuration/configure_system.cpp
@@ -14,6 +14,20 @@
 #include "yuzu/configuration/configuration_shared.h"
 #include "yuzu/configuration/configure_system.h"
 
+constexpr std::array<u32, 7> LOCALE_BLOCKLIST{
+    0b100011100001100000, // Japan
+    0b000001101001100100, // Americas
+    0b100110100001000010, // Europe
+    0b100110100001000010, // Australia
+    0b000000000000000000, // China
+    0b100111100001000000, // Korea
+    0b100111100001000000, // Taiwan
+};
+
+static bool IsValidLocale(u32 region_index, u32 language_index) {
+    return ((LOCALE_BLOCKLIST[region_index] >> language_index) & 1) == 0;
+}
+
 ConfigureSystem::ConfigureSystem(Core::System& system_, QWidget* parent)
     : QWidget(parent), ui{std::make_unique<Ui::ConfigureSystem>()}, system{system_} {
     ui->setupUi(this);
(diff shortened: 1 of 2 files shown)


src/yuzu/configuration/configure_system.cpp
```diff
@@ -14,6 +14,18 @@
 #include "yuzu/configuration/configuration_shared.h"
 #include "yuzu/configuration/configure_system.h"
 
+#define CHECK_LOCALE(region, language) (((LOCALE_BLOCKLIST[region] >> language) & 1) == 0)
```
lioncash, line 17: This can just be a regular translation-unit-scope function like:

```cpp
static bool IsValidLocale(u32 region_index, u32 language_index) {
  return ((LOCALE_BLOCKLIST[region_index] >> language_index) & 1) == 0;
}
```

src/yuzu/configuration/configure_system.cpp
```diff
@@ -34,6 +46,19 @@ ConfigureSystem::ConfigureSystem(Core::System& system_, QWidget* parent)
         }
     });
 
+    const auto locale_check = [this](int index) {
+        bool validLocale = CHECK_LOCALE(ui->combo_region->currentIndex(), ui->combo_language->currentIndex());
```
lioncash, line 50:
```suggestion
        bool valid_locale = CHECK_LOCALE(ui->combo_region->currentIndex(), ui->combo_language->currentIndex());
```

src/yuzu/configuration/configure_system.cpp
```diff
@@ -34,6 +46,19 @@ ConfigureSystem::ConfigureSystem(Core::System& system_, QWidget* parent)
...
+        ui->label_warn_invalid_locale->setVisible(!validLocale);
+        if (!validLocale) {
+            ui->label_warn_invalid_locale->setText(
+                QStringLiteral("Warning: \"%1\" is not a valid language for region \"%2\"")
```
lioncash, line 54:
```suggestion
                tr("Warning: \"%1\" is not a valid language for region \"%2\"")
```

Since this should be translatable

src/yuzu/configuration/configure_system.cpp
```diff
@@ -34,6 +48,22 @@ ConfigureSystem::ConfigureSystem(Core::System& system_, QWidget* parent)
         }
     });
 
+    const auto locale_check = [this](int index) {
+        bool valid_locale =
```
FearlessTobi, line 52: can be const

src/yuzu/configuration/configure_system.cpp
```diff
@@ -14,6 +14,20 @@
...
+    0b000001101001100100, // Americas
+    0b100110100001000010, // Europe
+    0b100110100001000010, // Australia
+    0b000000000000000000, // China
+    0b100111100001000000, // Korea
+    0b100111100001000000, // Taiwan
+};
+
+static bool IsValidLocale(u32 region_index, u32 language_index) {
+    return !(region_index < LOCALE_BLOCKLIST.size() &&
```
lioncash, line 28: seems like this could by just doing:

```cpp
return ((LOCALE_BLOCKLIST.at(region_index) >> language_index) & 1) == 0;
```

which will throw if an out-of-bound access attempt is made.

ColinKinloch: I had considered

```c++
try {
    return ((LOCALE_BLOCKLIST.at(region_index) >> language_index) & 1) == 0;
} catch (std::out_of_range const& exc) {
    return true;
}
```

I was thinking it would be good to have it return `true` when there is any doubt. If a new region is added it'll take a while to figure out what languages are valid.

Should I just leave the throw unhandled?

lioncash: Yeah, it's fine to just let it terminate

liamwhite (approved):

PR #9537, opened 2023-01-01
cmake: improve find_package failure messages

abouvier:
Make failure messages less scary as discussed in https://github.com/yuzu-emu/yuzu/pull/9515.

Also don't try to find a library via `pkg-config` when a CMake config file with unsuitable version has already been found. The `pkg-config` will fail too and it makes the message better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,42 +207,43 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
 # System imported libraries
 # =======================================================================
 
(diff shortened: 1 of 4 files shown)


liamwhite (approved):

Tachi107:
`<PackageName>_CONSIDERED_CONFIGS` is new to me... Turns out you can never truly know CMake well :)

abouvier:
Fair enough, I added a comment and enforced the search mode for every non-required package. The size of error message for required package does not matter since it's fatal anyway.

bunnei (approved):

PR #9539, opened 2023-01-01
video_core/opengl: Added FSR upscaling filter to the OpenGL renderer

Wollnashorn:
Implemented FSR 1.0 in the OpenGL renderer. I can only test it with an AMD gpu at the moment, but I hope it works fine with NVIDIA/Intel cards too.

https://user-images.githubusercontent.com/17367726/210182429-abd37b9e-ca60-4bec-ae0a-7f0b42fedf95.mp4

### How?
The GLSL sources AMD provided of FidelityFX FSR 1.0 can basically be used in OpenGL too as is, even though GPUOpen only mentions support in Vulkan.
In the original Vulkan example, the rendering was done in a compute shader with a custom rasterizer. To make it work in Yuzu's OpenGL renderer in a easy way, the EASU/RCAS functions are called here in a normal fragment shader drawn to a fullscreen triangle. The performance should roughly be the same.

FSR 1.0 has two passes: An upscale pass to an intermediate framebuffer of the output resolution and a subsequent sharpening pass to the target framebuffer. The intermediate framebuffer is resized at `RendererOpenGL::DrawScreen` everytime the output resolution has changed (e.g. user resized the Yuzu window).

I had to somehow fuse the FSR functions defined in vk_fsr.cpp with the new gl_fsr.cpp, so I moved these functions into video_core/fsr.(h/cpp).
Perhaps there is a better place for these functions?

### Why?
Of course the Vulkan API offers better performance than OpenGL and should always be preferred, but there are some games that do have graphical glitches and OpenGL acts as a good fallback there. Either way should be a nice feature to have FSR in OpenGL too.

<table width="100%">
  <tr>
  <th width="50%">Scaleforce</td>
  <th width="50%">FSR 1.0</td>
  </tr>
  <tr>
  <td></td>
  <td></td>
  </tr>
</table>


goldenx86:
Now OpenGL benefits from the golden FSR + SMAA combo.

src/video_core/fsr.cpp
```diff
@@ -0,0 +1,145 @@
+// SPDX-FileCopyrightText: Copyright 2023 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <cmath>
+#include "video_core/fsr.h"
+
+// Reimplementations of the constant generating functions in ffx_fsr1.h
+// GCC generated a lot of warnings when using the official header.
+u32 FSR::AU1_AH1_AF1(f32 f) {
```
lioncash, line 9: Seems like this function can be given internal linkage with static, since it's only used within this source file.

src/video_core/fsr.cpp
```diff
@@ -0,0 +1,145 @@
...
+        0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18,
+        0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18,
+        0x18, 0x18,
+    };
+    const u32 u = Common::BitCast<u32>(f);
+    const u32 i = u >> 23;
+    return base[i] + ((u & 0x7fffff) >> shift[i]);
+}
+
+u32 FSR::AU1_AH2_AF2(f32 a[2]) {
```
lioncash, line 101: Ditto

Austinzironic:
Where is the Yuzu emulator for MacOS?

src/video_core/renderer_opengl/gl_shader_manager.cpp
```diff
@@ -1,2 +1,123 @@
...
+    UnbindCompute();
+}
+
+void ProgramManager::UnbindCompute() {
+    if (is_compute_bound) {
+        is_compute_bound = false;
+        glUseProgram(0);
+    }
+}
+} // namespace OpenGL
```
liamwhite, line 123: missing newline

liamwhite (approved):

vonchenplus (approved):

vonchenplus (approved):

src/video_core/host_shaders/full_screen_triangle.vert
```diff
@@ -21,8 +26,8 @@ END_PUSH_CONSTANTS
 layout(location = 0) out vec2 texcoord;
 
 void main() {
-    float x = float((gl_VertexIndex & 1) << 2);
-    float y = float((gl_VertexIndex & 2) << 1);
+    float x = float((VERTEX_ID & 1) << 2);
```
vonchenplus, line 29: Is this file change necessary? If not please roll back the changes to this file if they are not required as one of my PRs will also need to change this file.

Wollnashorn: It is. The shader wasn't usable in OpenGL's GLSL before, as `gl_VertexIndex` is named `gl_VertexID` in OpenGL.
The same with adding the explicit definition of vertex shader outputs, as it is required with `ARB_separate_shader_objects`. Basically I copied these changes from the SMAA vertex shaders.

gidoly:
Appears to flip the screen on some games with nvidia gpus such as The legend of zelda breath of the wild and mariokart 8 and more.

yuzu-msvc-verify-pr09539-20230105-9bef04138_srao8vBqxA

vonchenplus:
@Wollnashorn You can make adjustments after #9556 been merged. Our PR has some conflicts.

liamwhite:
#9556 has now been merged

goldenx86:
Bug seems to only affect NVIDIA. An RX 6600 was fine, a 1660 Ti remains flipped with the latest artifact.

Wollnashorn:
@goldenx86 Oh boy, that's strange. Thanks for testing.
I rewrote some passages, so that the screen filling quad is used at the presentation stage, just like with all other filters, instead of a fullscreen triangle.
It is probably the cleaner and more maintainable solution anyway, even it might be 0.0001% slower in comparison 😄

Can you test again? Hopefully should be fixed now

goldenx86:
Will test NVIDIA in a bit, decided to test Intel (Gen 12, Xe in this case) as I only have my laptop on hand at the moment, and it produces an interesting result:
shintel

jacky400:
Just tested with Pokemon Legend Arceus , it is rendered upside down.

Wollnashorn:
It should now be fixed. NVIDIA's OpenGL driver was rightfully complaining about the usage of `layout(set)`, which made no sense in OpenGL.

jacky400:
@Wollnashorn Nivida issue fixed and thanks.

goldenx86:
Yup seems like the handheld issue is also fixed, I'll tag for EA.

goldenx86:
@Wollnashorn can I mention you work in the next progress report?

Wollnashorn:
@goldenx86 Yes of course, would be glad!

PR #9572, opened 2023-01-07
Remove memory allocations from maxwell_dma's LaunchDMA

Kelebek1:
These seemed unnecessary, 2 of them are fixed size and small, and the others can re-use the scratch buffers already in the class and skip allocation pretty much always.

### DIFF
--- a/src/video_core/engines/maxwell_dma.cpp
+++ b/src/video_core/engines/maxwell_dma.cpp
@@ -99,9 +99,10 @@ void MaxwellDMA::Launch() {
         if (regs.launch_dma.remap_enable != 0 && is_const_a_dst) {
             ASSERT(regs.remap_const.component_size_minus_one == 3);
             accelerate.BufferClear(regs.offset_out, regs.line_length_in, regs.remap_consta_value);
-            std::vector<u32> tmp_buffer(regs.line_length_in, regs.remap_consta_value);
-            memory_manager.WriteBlockUnsafe(regs.offset_out,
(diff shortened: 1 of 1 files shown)


src/video_core/engines/maxwell_dma.cpp
```diff
@@ -99,9 +99,10 @@ void MaxwellDMA::Launch() {
         if (regs.launch_dma.remap_enable != 0 && is_const_a_dst) {
             ASSERT(regs.remap_const.component_size_minus_one == 3);
             accelerate.BufferClear(regs.offset_out, regs.line_length_in, regs.remap_consta_value);
-            std::vector<u32> tmp_buffer(regs.line_length_in, regs.remap_consta_value);
-            memory_manager.WriteBlockUnsafe(regs.offset_out,
-                                            reinterpret_cast<u8*>(tmp_buffer.data()),
+            read_buffer.resize_destructive(regs.line_length_in * sizeof(u32));
+            std::fill(reinterpret_cast<u32*>(read_buffer.begin()),
+                      reinterpret_cast<u32*>(read_buffer.end()), regs.remap_consta_value);
```
Morph1984, line 104: This is UB

Morph1984 (changes requested):

src/video_core/engines/maxwell_dma.cpp
FernandoS27, line 104: make the read buffer use u32 and then reinterpret down to u16 or u8 in the others, that'll guarantee alignment and remove UB.

PR #9576, opened 2023-01-07
Partially revert #9535

bunnei:
Partially revert #9535 due to reported regressions. I'm just a bot -- @bylaws @gidoly please add context.


gidoly:
it breaks marvel ultimate allicance rendering

bunnei:
Confirmed not the regression.

PR #9612, opened 2023-01-13
1.5X, 7X, 8X resolution scaler options

goldenx86:
Seems to hit the sweetspot for current most common GPUS. Also looks great on 1440p displays.
Slapping an [EXPERIMENTAL] warning because Scarlet/Violet doesn't like it too much.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -153,6 +153,14 @@ void UpdateRescalingInfo() {
         info.up_scale = 6;
         info.down_shift = 0;
         break;
+    case ResolutionSetup::Res7X:
+        info.up_scale = 7;
+        info.down_shift = 0;
+        break;
+    case ResolutionSetup::Res8X:
+        info.up_scale = 8;
+        info.down_shift = 0;
+        break;
     default:
         ASSERT(false);
         info.up_scale = 1;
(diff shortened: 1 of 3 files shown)


goldenx86:
Added 7x and 8x by user trolling/popular demand.

Nerboruto:
perhaps a generic hack that disables dynamic resolution or prevents going below a certain level would be interesting.
anyway i will try it...

goldenx86:
Fast GPU already tries to do that by telling the game that the GPU is at peak performance. How to use that information is up to each game, so custom mods will still be needed.
Just build the PR or download the artifact. Git gud with git.

Nerboruto:
> La GPU veloce prova già a farlo dicendo al gioco che la GPU è al massimo delle prestazioni. Come utilizzare queste informazioni dipende da ogni gioco, quindi saranno comunque necessarie mod personalizzate. Basta costruire il PR o scaricare l'artefatto. Git gud con git.

i think that dynamic resolution destabilizes emulator 
in some way, however i tested it on some games it performs well, I've had some crashes on others, but I don't know if it was due to this... age of calamity, pokemon violet

goldenx86:
Changing the resolution on the fly forces the emulator to recreate the scene and kill the old one, eating tons of RAM and VRAM in the process.
That's why disably dynamic resolution mods and Fast GPU improve performance and stability, besides also keping the game at the best resolution it allows.

bunnei (approved):

PR #9615, opened 2023-01-14
audio_core: Corrective fixes to upsampler

merryhime:
* There was an off-by-one error in convolution operation, where the convolution would only summate 19 values in total instead of the expected 20.
* Make the coefficient arrays a sensible format, they're halves of offsets of a windowed sinc function divvied up as expected for the polyphase interpolator.

### DIFF
--- a/src/audio_core/renderer/command/resample/upsample.cpp
+++ b/src/audio_core/renderer/command/resample/upsample.cpp
@@ -20,25 +20,25 @@ static void SrcProcessFrame(std::span<s32> output, std::span<const s32> input,
                             const u32 target_sample_count, const u32 source_sample_count,
                             UpsamplerState* state) {
     constexpr u32 WindowSize = 10;
-    constexpr std::array<Common::FixedPoint<24, 8>, WindowSize> SincWindow1{
-        51.93359375f, -18.80078125f, 9.73046875f, -5.33203125f, 2.84375f,
-        -1.41015625f, 0.62109375f,   -0.2265625f, 0.0625f,      -0.00390625f,
+    constexpr std::array<Common::FixedPoint<17, 15>, WindowSize> WindowedSinc1{
+        0.95376587f,   -0.12872314f, 0.060028076f,  -0.032470703f, 0.017669678f,
+        -0.009124756f, 0.004272461f, -0.001739502f, 0.000579834f,  -0.000091552734f,
     };
-    constexpr std::array<Common::FixedPoint<24, 8>, WindowSize> SincWindow2{
-        105.35546875f, -24.52734375f, 11.9609375f,  -6.515625f, 3.52734375f,
-        -1.796875f,    0.828125f,     -0.32421875f, 0.1015625f, -0.015625f,
+    constexpr std::array<Common::FixedPoint<17, 15>, WindowSize> WindowedSinc2{
+        0.8230896f,    -0.19161987f,  0.093444824f,  -0.05090332f,   0.027557373f,
+        -0.014038086f, 0.0064697266f, -0.002532959f, 0.00079345703f, -0.00012207031f,
     };
-    constexpr std::array<Common::FixedPoint<24, 8>, WindowSize> SincWindow3{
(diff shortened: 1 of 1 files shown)


Morph1984 (approved):

Kelebek1:
I tried to implement it as close to the decomp as I could get, here are the functions for accuracy reference:

From DSP: https://pastebin.com/RwweKzk4
From SDK: https://pastebin.com/RNzHSf46

I recommend the SDK one, I think it's easier to follow.

merryhime:
> Why did you change which coeffs are used for example? Those at least are clear in the code. Like 160 -> 240 case 1 uses Window1/Window2 in that order, and case 2 uses Window2/Window1 in that order.

The coefficients have a meaning and an intrinsic order.

```
WindowedSinc1 = window(x + 1/6) * sinc(x + 1/6)
WindowedSinc2 = window(x + 2/6) * sinc(x + 2/6)
WindowedSinc3 = window(x + 3/6) * sinc(x + 3/6)
WindowedSinc4 = window(x + 4/6) * sinc(x + 4/6)
WindowedSinc5 = window(x + 5/6) * sinc(x + 5/6)
```

This is why in the `40 -> 240` case these are used in sequence, in the `80 -> 240` case every other one is used.

----

`160 -> 240` is the odd one out because what's actually going on is 3x expansion followed by a 2x decimation.

So what you see is:
```
160 -> 480:
next_input()
output[write_index] = state->history[state->history_output_index].to_int_floor();
calculate_sample(WindowedSinc2, WindowedSinc4);
calculate_sample(WindowedSinc4, WindowedSinc2);
next_input()
output[write_index] = state->history[state->history_output_index].to_int_floor();
calculate_sample(WindowedSinc2, WindowedSinc4);
calculate_sample(WindowedSinc4, WindowedSinc2);
```

160 -> 480 -> 240: (decimation by 2 skips every other operation above):
```
next_input()
output[write_index] = state->history[state->history_output_index].to_int_floor();
// calculate_sample(WindowedSinc2, WindowedSinc4); // decimated
calculate_sample(WindowedSinc4, WindowedSinc2);
next_input()
// output[write_index] = state->history[state->history_output_index].to_int_floor(); // decimated
calculate_sample(WindowedSinc2, WindowedSinc4);
// calculate_sample(WindowedSinc4, WindowedSinc2); // decimated
```

which is why the 160 -> 240 case is actually the odd duckling

merryhime:
> One thing I'm not sure about, what Q are the 2 halves of the sinc? I've been assuming things are Q8, input/history/output is clearly 8, you turned the coeffs into 15 which may make sense, but since they're only 16 bits, that leaves a sign bit and decimal bits only, I thoguht there would have to be at least 1 significant digit in it. Plus, when you multiply these together, you double the decimal digits right? Like Q8 * Q8 would give you a Q16 result.

The filter coefficients are Q15. None of the values are above 1.0, and you don't need a singificant digit, you just need a sign bit.

History is Q8 yes, it's pretty obvious from the decomp.

During the summation you're doing Q8 * Q15 which would give you a Q23 result. This is why at the end of the summation the value is shifted right by 23.

merryhime:
Looking at yuzu's implementation of `fixed_point`, `operator*` is incorrect, and probably caused your misunderstanding. 

For a reference that the scale factors multiply, see also: https://en.wikipedia.org/wiki/Fixed-point_arithmetic#Multiplication

There is a pretty obvious example:

2.0 * 3.0 = 6.0

But let's say we're multiply two Q4 values:

0x20 * 0x30 = 0x600

Now it's probably very clear to you that the result is Q8.

Similarly a lop-sided version:

0x20 * 0x300 = 0x6000

Which hopefully makes it obvious that Q4 * Q8 -> Q12.

Kelebek1:
Interesting, thanks. I thought multiplication would have to be done the same as addition I guess, multiplying the 0.5 bit by a 0.25 bit just seemed like it'd be wrong. I've almost certainly made that mistake all over the place then...

Definitely fix fixed_point if it's wrong.

Kelebek1 (approved):

bunnei (approved):

PR #9617, opened 2023-01-14
core: hid: Fix stick minimum range

german77:
Stick minimum range is -32,766 verified with HW.  As for why it isn't -32768 is unclear to me but there is a particular game where this matters, `EARTH DEFENSE FORCE: WORLD BROTHERS` is unable to read negative values.
image

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -10,6 +10,7 @@
 
 namespace Core::HID {
 constexpr s32 HID_JOYSTICK_MAX = 0x7fff;
+constexpr s32 HID_JOYSTICK_MIN = 0x7ffe;
 constexpr s32 HID_TRIGGER_MAX = 0x7fff;
 // Use a common UUID for TAS and Virtual Gamepad
 constexpr Common::UUID TAS_UUID =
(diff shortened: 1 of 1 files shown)


MonsterDruide1:
I don't quite get the stuff happening here...
The screenshot shows some `ff` at the end, doesn't that verify the `-32768` instead?
Also, how does a game not work on negative numbers *at all*? Shouldn't that also affect stick movements with less magnitude, like `-10000`?

german77:
I might have explained correctly. The screenshot reads 0xFFFF8002 this is -32766 stored in a s32. -32768 is the minimum possible on a s16.

The game does detect any value from -32766 to 32767 but anything outside that range is ignored. I could RE the game if you want an explanation but since the switch is also clamped to the same range I didn't feel like investing more time in this issue

MonsterDruide1:
Ah, I didn't think about little/big endian encodings, thanks for that explanation.

No REing required, your explanation on the game in the first message just confused me - I get it now, it only reads the values down to `-32766` and ignores everything further. What about a positive limit? Can the highest value (`32767`) still be observed in your HW tests? ... also, just another thought, is the `-32766` maybe just a hardware limit or stick design issue by Nintendo? That wouldn't explain the game ignoring all other values though.

german77:
Joycon outputs an u16 with a max value of 0xFFF. After applying the calibration the final value is the full range of s16. Not all games behave like this. Skyward sword for example has no issues reading values bigger than s16. This actually makes link run 10 times faster with no issues at all.

liamwhite (approved):

Issue #9436, opened 2022-12-14
Allowing changing volume with holding hotkey button


soredake:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What feature are you suggesting?

Right now, when you hold `-` or `=` it will decrease/increase volume only once, it will be nice if holding this hotkeys will continue to decrease/increase volume further.

### Why would this feature be useful?

It will be useful if you want to decrease for example from 100% to 25% without pressing button multiple times.

soredake:
Implemented now.

PR #9637, opened 2023-01-17
allow volume up/down hotkeys to be repeated

SaiKai:
pressing and holding volume up/down hotkeys will now increase/decrease repeatedly.


SaiKai:
this would close #9436

german77:
I kind of hate making the array even bigger and uglier. But I guess that's the easiest way to have something like this working. Seems fine to me I will test it in an hour or two

german77 (approved):

PR #9640, opened 2023-01-18
input_common: reset sdl motion if data is invalid

german77:
Turns out things like the web applet disable motion.  SDL is never aware of this and starts to push null data. This PR fixes the issue by restarting the sensors if they send invalid data after a few seconds.

### DIFF
--- a/src/input_common/drivers/sdl_driver.cpp
+++ b/src/input_common/drivers/sdl_driver.cpp
@@ -40,25 +40,26 @@ public:
     }
 
     void EnableMotion() {
-        if (sdl_controller) {
-            SDL_GameController* controller = sdl_controller.get();
-            has_accel = SDL_GameControllerHasSensor(controller, SDL_SENSOR_ACCEL) == SDL_TRUE;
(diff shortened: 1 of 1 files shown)


src/input_common/drivers/sdl_driver.cpp
```diff
@@ -942,18 +958,18 @@ MotionMapping SDLDriver::GetMotionMappingForDevice(const Common::ParamPackage& p
...
 
-    if (joystick->HasGyro() || joystick->HasAccel()) {
+    if (joystick->HasMotion()) {
         mapping.insert_or_assign(Settings::NativeMotion::MotionRight,
                                  BuildMotionParam(joystick->GetPort(), joystick->GetGUID()));
     }
     if (params.Has("guid2")) {
         joystick2->EnableMotion();
-        if (joystick2->HasGyro() || joystick2->HasAccel()) {
+        if (joystick->HasMotion()) {
```
FearlessTobi, line 967: Should this be joystick2?

german77: yes

liamwhite (approved):

bunnei (approved):

PR #9644, opened 2023-01-19
add volume quicksetting with volume slider

SaiKai:
adding a volume display to the main window which allows for easy display of volume/mute status.
clicking it will display a slider to change the volume

https://user-images.githubusercontent.com/6887590/213465007-1158e9f1-1dc3-41cd-9e3f-da961cd6b33d.mp4

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -3973,7 +3973,7 @@ void GMainWindow::UpdateVolumeUI() {
     volume_slider->setValue(volume_value);
     if (Settings::values.audio_muted) {
         volume_button->setChecked(false);
-        volume_button->setText(tr("VOLUME: MUTE", "Volume percentage (e.g. 50%)"));
+        volume_button->setText(tr("VOLUME: MUTE"));
     } else {
         volume_button->setChecked(true);
         volume_button->setText(tr("VOLUME: %1%", "Volume percentage (e.g. 50%)").arg(volume_value));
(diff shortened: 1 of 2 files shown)


src/yuzu/main.cpp
```diff
@@ -957,6 +957,38 @@ void GMainWindow::InitializeWidgets() {
...
+    volume_popup->setWindowFlags(Qt::FramelessWindowHint | Qt::NoDropShadowWindowHint | Qt::Popup);
+    volume_popup->setLayout(new QVBoxLayout());
+    volume_popup->setMinimumWidth(200);
+
+    volume_slider = new QSlider(Qt::Horizontal);
+    volume_slider->setObjectName(QStringLiteral("volume_slider"));
+    volume_slider->setMaximum(200);
+    volume_slider->setPageStep(5);
+    connect(volume_slider, &QSlider::valueChanged, this, [this](int percentage) {
+        Settings::values.audio_muted = false;
```
Kelebek1, line 970: This should probably be set to true at 0%?

SaiKai: sure..makes sense. should this behaviour also be introduced in the hotkey functions?

Kelebek1: If it's not already there, sure. I'm not actually sure what/where audio_muted is used, just saw it looking at your PR here.

FearlessTobi: It seems to be only used for the "Mute Audio" Hotkey.
I'm not sure if we need to set it to true here.

german77: Mute is for quickly jumping between full volume and zero. Doesn't make much sense to mute at 0% since it doesn't add any benefit.

liamwhite (approved):

bunnei (approved):

Issue #8928, opened 2022-09-18
Feature Request: Multiplayer: Connect to a room using a domain instead of ip address


hitsu4yrkl:
Hey awesome yuzu team, another **very very minor** quality of life suggestion:

Currently using Mainline 1171, the "Multiplayer/Direct Connect" dialogue window only accepts numbers and dots in an ip-like format (3 dots and numbers in between).
Yet, I have a domain pointing to the ip address of the Linux Server running my multiplayer room. In my opinion, it would be much easier to share a short domain like `example.net` with friends than a bulky ip address.

PR #9661, opened 2023-01-23
LDN hostname support in direct connect

SoRadGaming:
This PR is being opened due to an old PR with this feature takings four months to merge. Hopefully, we can get the feature added sooner.

This PR adds support for IPv6 and Hostnames.

Regex Testing:
image

Used in Test at (https://regexr.com/):
`((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|([a-zA-Z0-9]+(-[a-zA-Z0-9]+)*\.)+[a-zA-Z]{2,}`

Used in Yuzu Code:
`"^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$|^([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$|^([a-zA-Z0-9]+(-[a-zA-Z0-9]+)*\\.)+[a-zA-Z]{2,}$"`

Complied and Worked in **Yuzu**, Windows 11.

This PR fixes https://github.com/yuzu-emu/yuzu/issues/8928
This PR fixes https://github.com/yuzu-emu/yuzu/pull/8996


FearlessTobi:
the regex looks good from a first glance

liamwhite:
I'm not opposed to merging this as is. However, it's worth pointing out that this doesn't detect many possible shorthands for IPv6 addresses. Ideally, we should just let the getaddrinfo() call be used to determine whether an address is valid, rather than trying to validate it extensively here.

SoRadGaming:
> I'm not opposed to merging this as is. However, it's worth pointing out that this doesn't detect many possible shorthands for IPv6 addresses. Ideally, we should just let the getaddrinfo() call be used to determine whether an address is valid, rather than trying to validate it extensively here.

That would be a better way of handling it, but it seems to require more than a string input to validate which is already completed when clicking connect. I have updated the IPv6 Regex in accordance with the possible addresses. One issue is that ipv4 and hostnames have a regex conflict, for example (192.168.0.1.domainname.com) is a correct hostname but not ipv4, so the ipv4 regex is useless.

src/yuzu/multiplayer/direct_connect.ui
```diff
@@ -60,10 +60,10 @@
             <item>
              <widget class="QLineEdit" name="ip">
               <property name="toolTip">
-               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;IPv4 address of the host&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;IP address of the host&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
liamwhite, line 63: probably should be `Address of the host`

SoRadGaming: Changed it to `Server Address of the host`.

SoRadGaming:
Updated the UI to the Image Below
image

german77:
Clang format is still failing

src/yuzu/multiplayer/direct_connect.ui
```diff
@@ -85,6 +76,18 @@
...
+                <horstretch>0</horstretch>
+                <verstretch>0</verstretch>
+               </sizepolicy>
+              </property>
+              <property name="maximumSize">
+               <size>
+                  <width>65</width>
+                  <height>50</height>
+               </size>
+              </property>
```
Morph1984, line 90: was this meant to be added

SoRadGaming: `<property name="sizePolicy">
               <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
                <horstretch>0</horstretch>
                <verstretch>0</verstretch>
               </sizepolicy>
              </property>`

The size Policy is useless and was used in testing UI changes, I will remove it. As for **maximumSize** property, I demonstrate below why I added it, as the port can only have a limit of 5 characters there is no need to stretch that box.

With **maximumSize**:
image
With **maximumSize** Stretched:
image

Without **maximumSize**:
image
Without **maximumSize** Stretched:
image

liamwhite:
Could you squash these commits together? I will merge ASAP

SoRadGaming:
> Could you squash these commits together? I will merge ASAP

Sure, I Just Squashed it all into one Commit.

liamwhite (approved):

Issue #8552, opened 2022-07-08
[Feature Request] Controller friendly input confirmation applet


Anabas27:
Description:
Controller confirmation applet is controllable only via mouse/kb which is emulation breaker for many users. Especially if you use your PC as a Gaming Box (Network Streaming or as a Standalone controller controlled PC which boots up straight into a game list).

Request:
There's two solutions - First is an easy but homebrew one and second is the proper way,

1. Make an experimental checkbox somewhere in preferences which will bypass this applet altogether, because in 99.9% cases everything is already setup and you only have to press OK.  If you have this checkbox enabled - auto confirm controllers in the background.
2. Make a controller controlled overlay or window so you can confirm or/and configure everything with a controller (overlay is better because that way you can hide Windows UI altogether)

Requests describing the same problem:

#8132 - Solution presented here is awful and is just a workaround not a solution.
#4211

FernandoS27:
@Its-Rei @Morph1984 I think this a feature we have wanted to add in sometime and the UI was already done. any ideas?

Gamerboi520:
Upvote

Gamerboi520:
If anything id prefer the popup not even appear at all

D1Vad22:
Just started using Yuzu for the first time, adding this feature would be very helpful for controller only setups

mdac3564444:
Please add...

MaLDox77:
Upvote

devcordlesssystems:
Upvote! I use Steam Deck with 2 controllers and TV and I have to stand up, unplug steam deck, press OK on a touch screen, connect it back, go back to sofa each time I start a game :) Never had to change anything there, btw

djeer:
Upvote! I use Steam Deck with 2 controllers and TV and I have to stand up, unplug steam deck, press OK on a touch screen, connect it back, go back to sofa each time I start a game 🙂 Never had to change anything there, btw. I have a great experience in all other aspects and even with this hustle it's still awesome project 👍

Silas888:
Upvote from me too

MrAngularVelocity:
Just chiming in to say I've been using yuzu for years and this small popup breaks my controller only setup. Would love to just hit A and accept whatever is there.

devcordlesssystems:
[...]

PR #9663, opened 2023-01-23
qt: add option to disable controller applet

EBADBEEF:
- add checkbox to disable the controller applet UI
- when controller applet is disabled, use the yuzu-cmd fallback controller applet that applies controller config based on rules
- See https://github.com/yuzu-emu/yuzu/issues/8552 for some discussion

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -836,6 +836,7 @@ void Config::ReadUIValues() {
     ReadBasicSetting(UISettings::values.pause_when_in_background);
     ReadBasicSetting(UISettings::values.mute_when_in_background);
     ReadBasicSetting(UISettings::values.hide_mouse);
+    ReadBasicSetting(UISettings::values.controller_applet_disabled);
     ReadBasicSetting(UISettings::values.disable_web_applet);
 
     qt_config->endGroup();
@@ -1456,6 +1457,7 @@ void Config::SaveUIValues() {
     WriteBasicSetting(UISettings::values.pause_when_in_background);
     WriteBasicSetting(UISettings::values.mute_when_in_background);
     WriteBasicSetting(UISettings::values.hide_mouse);
+    WriteBasicSetting(UISettings::values.controller_applet_disabled);
     WriteBasicSetting(UISettings::values.disable_web_applet);
 
     qt_config->endGroup();
--- a/src/yuzu/configuration/configure_general.cpp
+++ b/src/yuzu/configuration/configure_general.cpp
@@ -44,6 +44,8 @@ void ConfigureGeneral::SetConfiguration() {
     ui->toggle_background_pause->setChecked(UISettings::values.pause_when_in_background.GetValue());
     ui->toggle_background_mute->setChecked(UISettings::values.mute_when_in_background.GetValue());
     ui->toggle_hide_mouse->setChecked(UISettings::values.hide_mouse.GetValue());
+    ui->toggle_controller_applet_disabled->setEnabled(runtime_lock);
+    ui->toggle_controller_applet_disabled->setChecked(UISettings::values.controller_applet_disabled.GetValue());
 
     ui->toggle_speed_limit->setChecked(Settings::values.use_speed_limit.GetValue());
     ui->speed_limit->setValue(Settings::values.speed_limit.GetValue());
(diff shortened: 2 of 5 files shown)


german77:
You need to do a bit more than this. You need to satisfy the rules of the applet as a fallback. Otherwise you will stay in a infinite loop of the game asking for the applet

EBADBEEF:
@german77 Can you give me an example? Would yuzu-cmd get stuck in a loop in the same case too? If my reading of the source is correct, it uses the existing fallback handler used by yuzu-cmd if no applet is supplied in SetAppletFrontendSet - https://github.com/yuzu-emu/yuzu/blob/f99f618d45ad862c4bc23fc28c91d1c48218a3cb/src/core/hle/service/am/applets/applets.cpp#L250

german77:
This is the set of rules the applet requires https://github.com/yuzu-emu/yuzu/blob/08d4e7c7aff6b4b798b9095c5715368918cdba0b/src/core/frontend/applets/controller.h#L20

Actually yes the cmd version has the same issue. You don't need to go that far mk8 has this issue. Just set `ui->buttonBox->setEnabled` to be always true so you can click the button and try to start a race with the incorrect number of players. To solve it you need to automatically connect controllers and set the correct type to satisfy with the rules. This can be done by accessing the EmulatedController directly

Brief example:
```
    npad_device = system.HIDCore().GetEmulatedController(npad_id);
    npad_device->Connect();
    npad_device->Disconnect();
    npad_device->SetNpadStyleIndex(Core::HID::NpadStyleIndex::ProController);
```

applet

EBADBEEF:
I'm trying to understand what needs to be done. Is what you described different from what DefaultControllerApplet::ReconfigureControllers does? It walks through all the controllers available and makes them conform to the requested parameters using SetNpadStyleIndex (https://github.com/yuzu-emu/yuzu/blob/08d4e7c7aff6b4b798b9095c5715368918cdba0b/src/core/frontend/applets/controller.cpp#L19).

What else is missing? If we don't have enough controllers, what happens? Oh, `hid_core.available_controllers` is always 10. So I guess walks through all possible controllers and connects them (even if they have no real controller backing them).

I've only tested Mario Odyssey (2 players) and it doesn't seem to keep calling the controller applet. I see the following my logs:
```
[ 143.874133] Service.HID <Info> core/hle/service/am/applets/applet_controller.cpp:Initialize:59: Initializing Controller Applet.
[ 143.874143] Service.HID <Info> core/frontend/applets/controller.cpp:ReconfigureControllers:21: called, deducing the best configuration based on the given parameters!
```

If I start Mario Odyssey with one controller plugged in, switch to two player mode, connect second controller, I can successfully use 2-player mode. I don't see ReconfigureControllers being called when the new controller is plugged in.

FearlessTobi:
A solution I'd honestly like much more than this would be to actually implement a controller applet that can be navigated without a mouse, Or trying to at least make the QT applet useable with a controller.

Hixbot2:
> A solution I'd honestly like much more than this would be to actually implement a controller applet that can be navigated without a mouse, Or trying to at least make the QT applet useable with a controller.

Interesting but at least in my case the issue with the applet is not just that it must be used with a mouse.   It is also that the applet is a separate window that requires focus when using, and then the main window need to be re-focused after the applet closes.  This doesn't play nice with some configurations, for example when using Moonlight on Android as the client device and Yuzu on Windows as the stream host.

Ebadbeef's idea to disable or auto-fill and close the applet according to pre-configured rules makes alot of sense in my view.

EBADBEEF:
@german77 I would appreciate another look at this. See earlier comment https://github.com/yuzu-emu/yuzu/pull/9663#issuecomment-1401233317 (quick summary: I believe the existing default controller applet already addresses all of your concerns).

I have since tested more games (including MK8) and haven't noticed any problems.

FearlessTobi:
> It is also that the applet is a separate window that requires focus when using, and then the main window need to be re-focused after the applet closes. 

To be honest I think this could just be fixed by returning the focus back to the main yuzu window after the applet has closed...

coolllman:
When will this PR be merged?

FearlessTobi:
It will be merged when the feedback has been addressed by the author.

EBADBEEF:
@FearlessTobi Do you mean your comments about wanting to navigate the applet with controller and/or handle focus better? I agree that would be good, but I would prefer to keep the option to not show a dialog. Can you convince me that the applet is actually useful?

My use case is a system with 1-4 bluetooth controllers (pro controller style) and displaying on a TV. I launch yuzu from a frontend purely with a controller. To quit yuzu I also use the controller (with a separate program listening for a specific button combo). I even have it set up to turn on and off the TV with the controller (haha). The controllers can get connected/disconnected at will. The keyboard and mouse is a crappy little wireless one that I keep just for emergencies. Maybe I should just be using yuzu-cmd? But then when I do need the UI it is annoying to switch back. The separate configs are also hard to manage. I can't remember but I might have had some other issues with yuzu-cmd.

Side note - sometimes a controller gets "disconnected" in yuzu and I have to go reconnect it in the UI. Mainly noticed on SSBU. I might try to look at that issue later (seems to happen with and without fancy controller applet).

PR #9676, opened 2023-01-25
Revert #9617 and fix it on input_common

german77:
Reverts all the changes from #9617. I made another homebrew to specifically check the weird value. Turns out -32767 is totally possible just unlikely. I confirmed as well by playing the game on the switch. If you put the stick just right you have the same issue as yuzu. So this is a game bug. Not a yuzu one.

Instead I modified the digital stick output to have a maximum range of 99.996% to allow for this game to work properly with keyboard and similar.

### DIFF
--- a/src/input_common/helpers/stick_from_buttons.cpp
+++ b/src/input_common/helpers/stick_from_buttons.cpp
@@ -11,6 +11,11 @@ namespace InputCommon {
 
 class Stick final : public Common::Input::InputDevice {
 public:
+    // Some games such as EARTH DEFENSE FORCE: WORLD BROTHERS
+    // do not play nicely with the theoretical maximum range.
+    // Using a value one lower from the maximum emulates real stick behavior.
+    static constexpr float MAX_RANGE = 32766.0f / 32767.0f;
+
     using Button = std::unique_ptr<Common::Input::InputDevice>;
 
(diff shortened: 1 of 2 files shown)


liamwhite (approved):

src/input_common/helpers/stick_from_buttons.cpp
```diff
@@ -11,6 +11,9 @@ namespace InputCommon {
 
 class Stick final : public Common::Input::InputDevice {
 public:
+    // Some games don't like the theorical maximum so aim for one unit off
```
Morph1984, line 14: Documents the game affected and the rationale behind why one unit lower is used
```suggestion
    // Some games such as EARTH DEFENSE FORCE: WORLD BROTHERS
    // do not play nicely with the theoretical maximum range.
    // Using a value one lower from the maximum emulates real stick behavior.
```

Morph1984: Also modifier scale is not changed? Perhaps you should just std::clamp the value

german77: Modifier scale is by default 50%. The idea is not to completely prevent this from happening. Just making it unlikely to happen. As the maximum value is completely valid. Just not for this particular game.

PR #9682, opened 2023-01-26
shader_recompiler: Remove S32 IR type

ameerj:
The shader IR does not distinguish between signed and unsigned integer types.

Fixes broken shaders when IR validation/graphics debugging is enabled for shaders that were using `BitCastS32F32`, where the resulting S32 argument was passed into opcodes expecting U32.


liamwhite (approved):

src/shader_recompiler/backend/glsl/emit_glsl_bitwise_conversion.cpp
```diff
@@ -48,10 +48,6 @@ void EmitBitCastU64F64(EmitContext& ctx, IR::Inst& inst, std::string_view value)
     ctx.AddU64("{}=doubleBitsToUint64({});", inst, value);
 }
 
-void EmitBitCastS32F32(EmitContext& ctx, IR::Inst& inst, std::string_view value) {
-    ctx.AddF32("{}=ftoi({});", inst, value);
```
Kelebek1, line 52: This bug always bothered me, so if this ends up not being reverted for some reason, fix this at least.

german77:
This format was added on #9001 Please check that the game hasn't regressed

vonchenplus (approved):
LGFM

PR #9690, opened 2023-01-28
yuzu: config: Avoid reading deleted object

german77:
Found this one while debugging. You can't do a for loop and delete entries of the object at the same time otherwise the index might go out of bounds.

### DIFF
--- a/src/yuzu/configuration/input_profiles.cpp
+++ b/src/yuzu/configuration/input_profiles.cpp
@@ -58,13 +58,16 @@ std::vector<std::string> InputProfiles::GetInputProfileNames() {
     std::vector<std::string> profile_names;
     profile_names.reserve(map_profiles.size());
 
-    for (const auto& [profile_name, config] : map_profiles) {
+    auto it = map_profiles.cbegin();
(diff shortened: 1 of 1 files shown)


src/yuzu/configuration/input_profiles.cpp
```diff
@@ -55,12 +55,13 @@ InputProfiles::InputProfiles() {
...
 
 std::vector<std::string> InputProfiles::GetInputProfileNames() {
+    std::vector<std::string> profiles_to_delete{};
     std::vector<std::string> profile_names;
     profile_names.reserve(map_profiles.size());
 
     for (const auto& [profile_name, config] : map_profiles) {
         if (!ProfileExistsInFilesystem(profile_name)) {
-            DeleteProfile(profile_name);
+            profiles_to_delete.emplace_back(profile_name);
```
Morph1984, line 64: You can fix this just by changing the for loop to use iterators, and calling erase on the map directly instead of DeleteProfile, as the check for !ProfileExistsInFilesystem is already satisfied

liamwhite (approved):

PR #9703, opened 2023-01-29
shaders: Fix TXQ with MSAA textures

ameerj:
TXQ can operate on MSAA textures, but has no flag within the instruction indicating if that is the case. The shader backends require knowing if the texture is MSAA as they cannot query LOD related information for MSAA textures.

This implementation relies on an assumption that TLD/`texelFetch` is called on the MSAA texture elsewhere in the shader, which can track if the texture binding is MSAA. 
This assumption will fall apart if there are shaders which query the size of the MSAA texture, but never sample from it. 

Ideally the MSAA mode for the texture is tracked based on its `TICEntry` from the engine environment, as is done here: https://github.com/yuzu-emu/yuzu/blob/3aab57452153c3dfa8591809bc09797da03b44dc/src/shader_recompiler/ir_opt/texture_pass.cpp#L367-L373 

But doing so will invalidate shader caches, so for now, we assume MSAA textures are always sampled with TLD.

With these changes in place, certain submenus in Dokapon UP! finally render.

|      Before | After        |
|---------|----------|
|  dokapon_broken | dokapon_fixed |

### DIFF
--- a/src/shader_recompiler/backend/glasm/emit_glasm_image.cpp
+++ b/src/shader_recompiler/backend/glasm/emit_glasm_image.cpp
@@ -59,6 +59,13 @@ std::string Image(EmitContext& ctx, IR::TextureInstInfo info,
     }
 }
 
+bool IsTextureMsaa(EmitContext& ctx, const IR::TextureInstInfo& info) {
+    if (info.type == TextureType::Buffer) {
+        return false;
+    }
+    return ctx.info.texture_descriptors.at(info.descriptor_index).is_multisample;
+}
+
 std::string_view TextureType(IR::TextureInstInfo info, bool is_ms = false) {
     if (info.is_depth) {
         switch (info.type) {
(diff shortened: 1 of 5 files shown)


liamwhite (approved):

src/shader_recompiler/backend/glasm/emit_glasm_image.cpp
```diff
@@ -59,6 +59,13 @@ std::string Image(EmitContext& ctx, IR::TextureInstInfo info,
     }
 }
 
+bool IsTextureMsaa(EmitContext& ctx, const IR::TextureInstInfo& info) {
```
vonchenplus, line 62: duplicate code

ameerj: There's not much that can be done about that for now. `EmitContext` is a unique class per backend, not shared or abstracted across the backends, hence why spir-v doesn't even follow this pattern.

Ideally when the msaa mode for the texture is read from the TICEntry, that can be encoded in the `TextureInstInfo` flags directly. But that can be done next time there's a need to invalidate shaders.

vonchenplus (approved):

FernandoS27 (approved):

PR #9720, opened 2023-02-03
Game Image with Discord RPC

SoRadGaming:
This PR has been opened due to https://github.com/yuzu-emu/yuzu/pull/9051 being stale.
I decided to use the Yuzu combability wiki to grab images.
image
image

### DIFF
--- a/src/yuzu/discord_impl.cpp
+++ b/src/yuzu/discord_impl.cpp
@@ -4,7 +4,10 @@
 #include <chrono>
 #include <string>
 #include <discord_rpc.h>
+#include <fmt/format.h>
+#include <httplib.h>
 #include "common/common_types.h"
+#include "common/string_util.h"
 #include "core/core.h"
 #include "core/loader/loader.h"
 #include "yuzu/discord_impl.h"
@@ -14,7 +17,6 @@ namespace DiscordRPC {
 
 DiscordImpl::DiscordImpl(Core::System& system_) : system{system_} {
     DiscordEventHandlers handlers{};
-
     // The number is the client ID for yuzu, it's used for images and the
     // application name
     Discord_Initialize("712465656758665259", &handlers, 1, nullptr);
@@ -29,23 +31,74 @@ void DiscordImpl::Pause() {
     Discord_ClearPresence();
 }
 
+static std::string GetGameString(const std::string& title) {
+    // Convert to lowercase
+    std::string icon_name = Common::ToLower(title);
+
+    // Replace spaces with dashes
+    std::replace(icon_name.begin(), icon_name.end(), ' ', '-');
+
+    // Remove non-alphanumeric characters but keep dashes
+    std::erase_if(icon_name, [](char c) { return !std::isalnum(c) && c != '-'; });
+
+    // Remove dashes from the start and end of the string
+    icon_name.erase(icon_name.begin(), std::find_if(icon_name.begin(), icon_name.end(),
+                                                    [](int ch) { return ch != '-'; }));
+    icon_name.erase(
+        std::find_if(icon_name.rbegin(), icon_name.rend(), [](int ch) { return ch != '-'; }).base(),
+        icon_name.end());
+
+    // Remove double dashes
+    icon_name.erase(std::unique(icon_name.begin(), icon_name.end(),
+                                [](char a, char b) { return a == '-' && b == '-'; }),
+                    icon_name.end());
+
+    return icon_name;
+}
+
 void DiscordImpl::Update() {
     s64 start_time = std::chrono::duration_cast<std::chrono::seconds>(
                          std::chrono::system_clock::now().time_since_epoch())
                          .count();
+    const std::string default_text = "yuzu is an emulator for the Nintendo Switch";
+    const std::string default_image = "yuzu_logo";
+    std::string game_cover_url = "https://yuzu-emu.org";
     std::string title;
+
+    DiscordRichPresence presence{};
+
     if (system.IsPoweredOn()) {
         system.GetAppLoader().ReadTitle(title);
-    }
-    DiscordRichPresence presence{};
-    presence.largeImageKey = "yuzu_logo";
-    presence.largeImageText = "yuzu is an emulator for the Nintendo Switch";
-    if (system.IsPoweredOn()) {
+
+        // Used to format Icon URL for yuzu website game compatibility page
+        std::string icon_name = GetGameString(title);
+
+        // New Check for game cover
+        httplib::Client cli(game_cover_url);
+
+        if (auto res = cli.Head(fmt::format("/images/game/boxart/{}.png", icon_name).c_str())) {
+            if (res->status == 200) {
+                game_cover_url += fmt::format("/images/game/boxart/{}.png", icon_name);
+            } else {
+                game_cover_url = "yuzu_logo";
+            }
+        } else {
+            game_cover_url = "yuzu_logo";
+        }
+
+        presence.largeImageKey = game_cover_url.c_str();
+        presence.largeImageText = title.c_str();
+
+        presence.smallImageKey = default_image.c_str();
+        presence.smallImageText = default_text.c_str();
         presence.state = title.c_str();
         presence.details = "Currently in game";
     } else {
-        presence.details = "Not in game";
+        presence.largeImageKey = default_image.c_str();
+        presence.largeImageText = default_text.c_str();
+        presence.details = "Currently not in game";
     }
+
     presence.startTimestamp = start_time;
     Discord_UpdatePresence(&presence);
 }
--- a/src/yuzu/CMakeLists.txt
+++ b/src/yuzu/CMakeLists.txt
@@ -353,7 +353,7 @@ if (USE_DISCORD_PRESENCE)
         discord_impl.cpp
         discord_impl.h
     )
-    target_link_libraries(yuzu PRIVATE DiscordRPC::discord-rpc)
+    target_link_libraries(yuzu PRIVATE DiscordRPC::discord-rpc httplib::httplib)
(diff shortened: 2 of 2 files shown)


src/yuzu/discord_impl.cpp
```diff
@@ -33,20 +33,54 @@ void DiscordImpl::Update() {
...
+
+        // Remove dashes from the start and end of the string
+        temp_title.erase(temp_title.begin(),
+                         std::find_if(temp_title.begin(), temp_title.end(),
+                                      [](int ch) { return !std::isspace(ch); }));
+
+        // Remove double dashes
+        temp_title.erase(std::unique(temp_title.begin(), temp_title.end(),
+                                     [](char a, char b) { return a == '-' && b == '-'; }),
+                         temp_title.end());
```
Morph1984, line 64: Hopefully a better API can be provided in the future, but for now this could work

src/yuzu/discord_impl.cpp
```diff
@@ -33,20 +33,54 @@ void DiscordImpl::Update() {
...
-        system.GetAppLoader().ReadTitle(title);
-    }
+
     DiscordRichPresence presence{};
-    presence.largeImageKey = "yuzu_logo";
-    presence.largeImageText = "yuzu is an emulator for the Nintendo Switch";
+
     if (system.IsPoweredOn()) {
+        system.GetAppLoader().ReadTitle(title);
+        std::string temp_title = title;
```
Morph1984, line 45:
```suggestion
        std::string icon_name = title;
```

src/yuzu/discord_impl.cpp
Morph1984: I would also document that this string manipulation is used to extract the icon from the yuzu compat wiki

Morph1984:
Changes look good overall, just rename the commit to describe the changes better

src/yuzu/discord_impl.cpp
SoRadGaming: > I would also document that this string manipulation is used to extract the icon from the yuzu compat wiki

Added "Used to format Icon URL for yuzu website game compatibility page", as for the future, game id seems the best way to manage it.

SoRadGaming:
This is what it looks like when the game is not in the wiki
image

src/yuzu/discord_impl.cpp
```diff
@@ -33,20 +33,57 @@ void DiscordImpl::Update() {
...
+        presence.largeImageKey = default_image.c_str();
+        presence.largeImageText = default_text.c_str();
+        presence.details = "Currently not in game";
     }
+
     presence.startTimestamp = start_time;
     Discord_UpdatePresence(&presence);
 }
-} // namespace DiscordRPC
+} // namespace DiscordRPC
```
liamwhite, line 89: missing newline

liamwhite:
IMO, we should check the image for existence first, so that the broken icon doesn't appear with games not in the compatibility wiki. Otherwise, looks good to me.

src/yuzu/discord_impl.cpp
```diff
@@ -33,20 +33,57 @@ void DiscordImpl::Update() {
     s64 start_time = std::chrono::duration_cast<std::chrono::seconds>(
                          std::chrono::system_clock::now().time_since_epoch())
                          .count();
+    std::string default_text = "yuzu is an emulator for the Nintendo Switch";
```
FearlessTobi, line 36: can be const. dito below

SoRadGaming:
> IMO, we should check the image for existence first, so that the broken icon doesn't appear with games not in the compatibility wiki. Otherwise, looks good to me.

Agreed, but adding a wildcard yuzu logo png would be better for the site. *.png should point to a default image instead. I was planning to open a PR to add this, but I have never used Hugo; I'm more of a ReactJS guy. TLDR: This should be fixed server side, not the client side.

liamwhite:
As far as I know, there's no way to do it with a wildcard rule because it's a static site.

SoRadGaming:
> As far as I know, there's no way to do it with a wildcard rule because it's a static site.

You're right; Hugo doesn't support the feature, but other frameworks do. I will add a quick fetch URL check. What library do you guys use to fetch URLs? I cant find any.

liamwhite:
cpp-httplib is available from externals

src/yuzu/discord_impl.cpp
```diff
@@ -29,23 +30,82 @@ void DiscordImpl::Pause() {
     Discord_ClearPresence();
 }
 
+std::string GetGameString(const std::string& title) {
+    std::string icon_name = title;
+
+    // Convert to lowercase
+    std::transform(icon_name.begin(), icon_name.end(), icon_name.begin(), ::tolower);
```
lioncash, line 37: We have [ToLower()](https://github.com/yuzu-emu/yuzu/blob/master/src/common/string_util.h#L16) in string_util.h , which would let you collapse this as:

```cpp
std::string icon_name = ToLower(title);
```

src/yuzu/discord_impl.cpp
```diff
@@ -29,23 +30,82 @@ void DiscordImpl::Pause() {
     Discord_ClearPresence();
 }
 
+std::string GetGameString(const std::string& title) {
```
lioncash, line 33:
```suggestion
static std::string GetGameString(const std::string& title) {
```

since it's only used in this file.

src/yuzu/discord_impl.cpp
```diff
@@ -29,23 +30,82 @@ void DiscordImpl::Pause() {
...
+        if (auto res = cli.Get(fmt::format("/{}.png", icon_name).c_str())) {
+            if (res->status == 200) {
+                game_cover_url += fmt::format("{}.png", icon_name);
+            } else {
+                LOG_INFO(Frontend, "Game cover not found, using default yuzu logo : {}",
+                         res->status);
+                game_cover_url = "yuzu_logo";
+            }
+        } else {
+            auto err = res.error();
```
lioncash, line 91: maybe join this into the log call, as:

```cpp
LOG_ERROR(Frontend, "Error getting game cover: {}", res.error());
```

src/yuzu/discord_impl.cpp
```diff
@@ -29,23 +30,82 @@ void DiscordImpl::Pause() {
...
+    std::string icon_name = title;
+
+    // Convert to lowercase
+    std::transform(icon_name.begin(), icon_name.end(), icon_name.begin(), ::tolower);
+
+    // Replace spaces with dashes
+    std::replace(icon_name.begin(), icon_name.end(), ' ', '-');
+
+    // Remove non-alphanumeric characters but keep dashes
+    icon_name.erase(std::remove_if(icon_name.begin(), icon_name.end(),
```
lioncash, line 43: seems like this can be shortened into:

```cpp
std::erase_if(icon_name, [](char c) { return !std::isalnum(c) && c != '-'; });
```

SoRadGaming:
Still need to fix the fact that I keep getting error 2 (as the res), with no documentation on what that means. Assuming a connection issue.

src/yuzu/discord_impl.cpp
```diff
@@ -29,23 +31,84 @@ void DiscordImpl::Pause() {
...
+        std::string icon_name = GetGameString(title);
+
+        // New Check
+        // Http GET request to check if the game cover exists TODO remove this
+        LOG_INFO(Frontend, "Checking if game cover exists: {}{}.png", game_cover_url.c_str(),
+                 icon_name.c_str());
+
+        httplib::Client cli(game_cover_url.c_str());
+
+        if (auto res = cli.Get(fmt::format("/{}.png", icon_name).c_str())) {
```
Tachi107, line 83: [Recent versions of cpp-httplib](https://github.com/yhirose/cpp-httplib/releases/tag/v0.11.0) use `std::string` instead of C strings, is this conversion still necessary?

SoRadGaming: Error when not using C string 
image

SoRadGaming: turn out its some issue with.png at the end of the URL and yuzu domain, as google.com works fine, but googles .png links don't, ill have to change `cli.Get()`.

SoRadGaming:
> Still need to fix the fact that I keep getting error 2 (as the res), with no documentation on what that means. Assuming a connection issue. (Status code -1 so connection issue)

turn out its some issue with.png at the end of the URL and the yuzu domain, as google.com works fine, but Google .png links don't, ill have to change `cli.Get()`.

src/yuzu/discord_impl.cpp
Tachi107: Ok, yuzu's bundled cpp-httplib is still at version 0.10.8, so yeah until we update the submodule passing a C string is required

liamwhite, line 83: The string object returned by fmt::format goes out of scope after .c_str() is called, resulting in undefined behavior with the const char* passed to cli.Get(). You'll need to store the returned string somewhere while you call the method.

Also, consider using HEAD (cli.Head()) to check for existence instead of GET.

Tachi107:
On Sat Feb 4, 2023 at 9:12 PM CET, liamwhite wrote:
> The string object returned by fmt::format goes out of scope after .c_str() is called, resulting in undefined behavior with the const char* passed to cli.Get(). You'll need to store the returned string somewhere while you call the method.

Isn't the lifetime of the returned string extended until the end of the
expression?

liamwhite:
Oops, that is correct in this case because the full-expression is the outer call. I edited the comment.

SoRadGaming:
**No Image (404 anything but 200 ok):**
image
**Status (200)**
image
**In Menu**
image

liamwhite (approved):

bunnei (approved):

PR #9744, opened 2023-02-08
gpu_thread: Remove OnCommandListEndCommand

behunin:
Call rasterizer->ReleaseFences() directly

Tegra::Control::Scheduler::Push calls Tegra::DmaPusher::DispatchCalls which calls Tegra::GPU::OnCommandListEnd

This in turn would make the gpu thread produce to its own command queue.

After this is tested and shows no regressions I will bring back the bounded queue. I believe this is the root cause of the problems the bounded queue encountered.

### DIFF
--- a/src/video_core/gpu.cpp
+++ b/src/video_core/gpu.cpp
@@ -99,7 +99,7 @@ struct GPU::Impl {
 
     /// Signal the ending of command list.
     void OnCommandListEnd() {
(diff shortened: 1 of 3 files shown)


vonchenplus:
I have previously PR a similar solution at #8573

liamwhite (approved):

PR #9746, opened 2023-02-08
texture_cache: OpenGL: Implement MSAA uploads and copies

ameerj:
This is an initial implementation of MSAA image uploads and copies in OpenGL. The main challenge for MSAA support has been replicating the memory layout of MSAA textures, since the APIs forbid calling most typical memory operations on MSAA textures (likely because the memory layout could be device/driver dependent).

Implementing this on Vulkan proved to be a lot more tedious, so for now, it remains unsupported.

This was tested on Sonic Forces, which uses MSAA textures for the background of the map screen:

https://user-images.githubusercontent.com/52414509/217416312-7b7917cf-0a2d-4481-a9a4-7563dc5a9e6e.mp4

https://user-images.githubusercontent.com/52414509/217416319-9d1e4deb-83e4-4cdd-9e8b-04b3200d71f7.mp4

Further testing on other titles known to use MSAA would be appreciated.

### DIFF
--- a/src/video_core/texture_cache/util.cpp
+++ b/src/video_core/texture_cache/util.cpp
@@ -573,10 +573,6 @@ u32 CalculateUnswizzledSizeBytes(const ImageInfo& info) noexcept {
     if (info.type == ImageType::Buffer) {
         return info.size.width * BytesPerBlock(info.format);
     }
-    if (info.num_samples > 1) {
-        // Multisample images can't be uploaded or downloaded to the host
-        return 0;
-    }
     if (info.type == ImageType::Linear) {
         return info.pitch * Common::DivCeil(info.size.height, DefaultBlockHeight(info.format));
     }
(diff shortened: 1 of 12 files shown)


src/video_core/texture_cache/util.cpp
```diff
@@ -703,7 +699,7 @@ ImageViewType RenderTargetImageViewType(const ImageInfo& info) noexcept {
 std::vector<ImageCopy> MakeShrinkImageCopies(const ImageInfo& dst, const ImageInfo& src,
                                              SubresourceBase base, u32 up_scale, u32 down_shift) {
     ASSERT(dst.resources.levels >= src.resources.levels);
-    ASSERT(dst.num_samples == src.num_samples);
+    // ASSERT(dst.num_samples == src.num_samples);
```
lioncash, line 702: Was this accidentally left in?

liamwhite (approved):
Approved for content, but please resolve the comment above

bunnei (approved):

PR #9749, opened 2023-02-08
buffer_base: Partially revert changes from #9559

ameerj:
This fixes a regression where Yoshi's Crafted World (and potentially other titles) would crash/enter an infinite loop when GPU Accuracy was set to "Normal"


gidoly:
appear to fixes pokemon scarlet/violet regression which causes land scape will disappear randomly and tree would flicker with gpu high

goldenx86 (approved):
LGTM

PR #9772, opened 2023-02-11
video_core: Create indirect buffer instead of dynamic create

vonchenplus:
IndirectDraw causes device loss in some cases, according to the analysis it is caused by buffer, possibly a bug in the driver. the current solution is to create a separate buffer.

Nintendo Switch Sports can enter the game now.

image

### DIFF
--- a/src/video_core/buffer_cache/buffer_base.h
+++ b/src/video_core/buffer_cache/buffer_base.h
@@ -289,6 +289,11 @@ public:
         return cpu_addr;
     }
 
+    /// Update the base CPU address of the buffer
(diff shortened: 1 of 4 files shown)


liamwhite (dismissed):

liamwhite:
Regresses rendering in Pokemon Scarlet
0100a3d008c5c000_2023-02-11_20-19-14-832

Intended is this
0100a3d008c5c000_2023-02-11_20-31-06-940

vonchenplus:
abnormal data in indirect buffer, further analysis required

PR #9778, opened 2023-02-11
gpu_thread: Use bounded queue

behunin:


Morph1984:
Causes crashes in Fire Emblem Three Houses

bunnei (approved):

PR #9781, opened 2023-02-11
kernel/svc: Fix undefined info_id

ColinKinloch:
The build fails due to `info_id` being used before it's defined.
From #9742


v1993:
Why not just move this trace below info_id definition?

ColinKinloch:
Should I change it?

liamwhite (approved):

PR #9792, opened 2023-02-12
Fix borderless hardware flip presentation; exponential performance increase

marat569:
Due to the frame buffer being sized incorrectly, yuzu is unable to engage hardware flip within a borderless presentation.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -3168,7 +3168,7 @@ void GMainWindow::ShowFullscreen() {
         window->setWindowFlags(window->windowFlags() | Qt::FramelessWindowHint);
         const auto screen_geometry = GuessCurrentScreen(window)->geometry();
         window->setGeometry(screen_geometry.x(), screen_geometry.y(), screen_geometry.width(),
-                            screen_geometry.height() + 1);
+                            screen_geometry.height());
         window->raise();
         window->showNormal();
     };


Morph1984 (changes requested):
The +1 is required so that windows treats it as a borderless window rather than a fullscreen exclusive window.
This setting was introduced to allow windows to appear on top of the current render window, as these are used for applets such as the software keyboard and error overlays.
If you require the window to be exactly the same size as the display, the exclusive fullscreen setting exists.

marat569:
There is no reason to use Fullscreen Exclusive 2023, outside of preference.

Thanks to all the hard work from Micorsoft and hardware vendors; we now get all the benefits FSE used to provide, but in borderless.

Intentionally adding the +1 is just griefing users.

Morph1984:
I think you're misunderstanding the issue here. If you resize the window to exactly the screen dimensions, windows will treat it as a fullscreen exclusive window despite all the attributes being set for a borderless window.
Trust me, I've already tried it when I made the borderless window setting in #6135 
If it had worked, I wouldn't have added the +1 there

Morph1984:
If you do not believe me, you can test the generated PR artifact and see whether alt-tabbing is fast or slow. That is the test I've used to determine whether it is fullscreen exclusive or borderless.

marat569:
> I think you're misunderstanding the issue here. If you resize the window to exactly the screen dimensions, windows will treat it as a fullscreen exclusive window despite all the attributes being set for a borderless window. Trust me, I've already tried it when I made the borderless window setting in #6135 If it had worked, I wouldn't have added the +1 there

A borderless window with windowed optimizations for direct flip is exactly what we need. 

The current behavior with the +1 shows you an image that takes up an entire monitor, but is in Composed: Flip due to the +1 -- essentially a hack forcing the user to waste their time to go into FSE 

Without the +1 we'd be in hardware independent flip (or hardware composed: independent flip if MPOs are present) -- this is how things should be.

Morph1984:
I understand that, but you're missing the issue that I mentioned earlier, is that windows treats the borderless window exactly the same as fullscreen exclusive.

Stimpy7314:
> > I think you're misunderstanding the issue here. If you resize the window to exactly the screen dimensions, windows will treat it as a fullscreen exclusive window despite all the attributes being set for a borderless window. Trust me, I've already tried it when I made the borderless window setting in #6135 If it had worked, I wouldn't have added the +1 there
> 
> A borderless window with windowed optimizations for direct flip is exactly what we need.
> 
> The current behavior with the +1 shows you an image that takes up an entire monitor, but is in Composed: Flip due to the +1 -- essentially a hack forcing the user to waste their time to go into FSE
> 
> Without the +1 we'd be in hardware independent flip (or hardware composed: independent flip if MPOs are present) -- this is how things should be.

There's also a setting within the nvidia control panel to enable such features for opengl-vulkan games. DXGI Swapchain, I believe.

EndlesslyFlowering:
> I think you're misunderstanding the issue here. If you resize the window to exactly the screen dimensions, windows will treat it as a fullscreen exclusive window despite all the attributes being set for a borderless window. Trust me, I've already tried it when I made the borderless window setting in #6135 If it had worked, I wouldn't have added the +1 there

this is not Windows' fault but the fault of AMD and Nvidia's driver
they have optimisations to convert windows that are borderless and match the screen resolution (maybe even stretched ones too) into fullscreen exclusive. This applies for both OpenGL and Vulkan. iirc OpenGL doesn't even have FSE and Vulkan got an extension later but both AMD and Nvidia already had their driver hacks in place.

liamwhite:
In exclusive fullscreen, applets like the web applet and on-screen keyboard do not work; this does not happen in bordered fullscreen, but will happen if you use this PR (because Windows will, for whatever reason, treat it like exclusive fullscreen).

marat569:
I mean I don't know if you guys are just anti-borderless or something; but I've been manually re-sizing the framebuffer to fit for a while now; and well life is good.

Especially with 22h2/nvidia dxgi layered swapchain; everything just works so much better.

Having the +1 is just like how modern d3d12 games that have no GDI presentation add a "fullscreen exclusive" option because uneducated users on steam complain; not knowing FSE is pretty much obsolete for some time now -- essentially creating a placebo setting to placate the people.

Morph1984:
I am all ears for borderless if windows actually treats it as such. I after all was the one who added it to solve an issue. If we can get the window to be treated properly at the correct size, then go for it.

Morph1984:
I apologize for the assumption I made earlier that you understood the issue I had when I tried to implement the borderless windowed mode. But I hope you do now understand why I wasn't able to eliminate the +1 from the code

Morph1984:
I have a 1080p display at 100% scaling, and testing the generated artifact above, both borderless windows and FSE act exactly the same (same alt tab behavior and swkbd applet does not show up)

marat569:
> 

Feel free to download presentmon https://github.com/GameTechDev/PresentMon -- and compare the presentations.

With the wrong resolution, you won't achieve hardware flip with the +1

Morph1984:
> > 
> 
> Feel free to download presentmon https://github.com/GameTechDev/PresentMon -- and compare the presentations.
> 
> With the wrong resolution, you won't achieve hardware flip with the +1

I understand why the +1 is undesirable, I'm explaining why I can't remove it

liamwhite:
The performance is not the issue here -- I'm unsurprised that it's better. The issue is that because the window is treated like exclusive fullscreen when you match the dimensions, it _breaks important functionality of the application_, including the web applet (used for manuals) and the software keyboard applet (used for inputting text). These don't work in exclusive fullscreen, and don't work in borderless fullscreen anymore with this PR applied.

NativeGamer081:
I belive that merging this commit will allow SpecialK HDR Retrofit to work as intend with Yuzu using the Vulkan API.  SpeicalK doesn't work with Yuzu's Vulkan API, the screen is enlarged as shown in the image below, however SpeicalK does work with Ryujinx's Vulkan API. It's only when I use OpenGL with Yuzu that SpeicalK works. 
Screenshot 2023-02-12 193823

marat569:
> I believe that merging this commit will allow Special K HDR Retrofit to work as intended with Yuzu using the Vulkan API. Special K doesn't work with Yuzu's Vulkan API, the screen is enlarged as shown in the image below, however Special K does work with Ryujinx's Vulkan API. It's only when I use OpenGL with Yuzu that Special K works. Screenshot 2023-02-12 193823

Right.

Nvidia's DXGI interlop does not like windows that get resized or are outside of the display's resolution.

So having the +1 griefs a fair bit.

@Morph1984 @liamwhite  why not consider a config option; where advanced users can opt-out of the +1 -- a win-win scenario for all parties

liamwhite:
> where advanced users can opt-out of the +1 -- a win-win scenario for all parties

This should be what the exclusive fullscreen setting does, no?

marat569:
> > where advanced users can opt-out of the +1 -- a win-win scenario for all parties
> 
> This should be what the exclusive fullscreen setting does, no?

FSE and hardware accelerated borderless are two different things; especially with alt tab behavior -- or when trying to do something like the nvidia interlop -> inject reshade -> HDR

giias:
another way to "fix" this situation would be to use a dxgi swapchain/interop layer to basically have the game use d3d11/dxgi for final presentation instead of vulkan.

normally vulkan games get "promoted" to fullscreen exclusive by default (unless you do something like the +1 solution implemented here), which yes (full screen) has an "always-on-top" situation that prevents the web applet and other windows/apps from layering/showing on top of the game. This "promotion" to fullscreen exclusive could be prevented from happening if dxgi was used for windowed/borderless... but... getting to that can be tricky...

nvidia semi-recently added a dxgi interop that can be enabled for opengl and vulkan games in the nvidia control panel. The setting there is "Vulkan/OpenGL present method" and need to set it to "Prefer layered on DXGI Swapchain" instead of the other options there in the nvidia control panel. However, for it to work properly (in actual windowed/borderless mode) one should also set OGL_DX_PRESENT_DEBUG to 0x00080001 -- this can be done in the "Nvidia Profile Inspector" under the hidden section 8 - Extra (also, this setting/section might be hidden until the "Show unknown settings from NVIDIA predefined profiles" magnifying glass icon is selected inside the top bar in the nvidia profile inspector). 

also, the 0x00080001 flag has to be typed (or copy/pasted) in the field for OGL_DX_PRESENT_DEBUG -- since 0x00080001 is not actually an available option to select in the drop down menu there, and basically this was found to disable at the driver level the "promotion" to Full Screen Exclusive (FSE) for games running with vulkan/dxvk, This together with the "Prefer layered on DXGI Swapchain" option from the nvidia control panel can get you a proper windowed/borderless game that is vulkan(or a game using dxvk) but presented with dxgi as if it was a d3d11 game, and the game can then run in windowed/borderless with hardware independent flip presentation (basically the same performance as FSE/legacy flip), and since it's proper windowed/borderless without an FSE or "always-on-top" flag, it can allow other windows and apps to appear on top of the game (though if you do place other windows/apps on top of the game, hopefully you have active multi-plane overlays (MPOs) so that the windows/apps on top of the game don't drop your game from the high performance "hardware independent flip" (or "hardware composed: independent flip") to the lower performance "composed flip" presentation. The "composed flip" presentation is higher latency and is known to make G-sync/VRR inactive). 

kinda the problem with this method also though is that it relies on using basically a hidden nvidia driver flag, and there's not an AMD equivalent currently to my knowledge at least. Additionally, nvidia's dxgi interop apparently has some problems currently where resizing a game window sometimes doesn't work properly and may even freeze a game (or this was the case a while back at least... maybe it's better now after the latest nvdia driver update ? but i doubt it)

that said, since apparently with yuzu there's also an exclusive fullscreen setting that keeps the window exactly the same size as the display, one could use nvidia's dxgi interop to present the vulkan game with dxgi and then even have the _Special K_ software hook into the dxgi layer and ensure that the FSE game is forced into windowed/borderless using "[Display.Output]ForceWindowed=true" in Special K's config. This apparently allows yuzu vulkan/fse games to be presented and work as if they were a d3d11 windowed/borderless game with hardware independent flip presentation (which also would let you have G-Sync/VRR and scRGB HDR on windowed/borderless through Special K). Marat could use this method at least as long as he still has an nvidia gpu.

Special K currently doesn't hook into the vulkan api, so it can't work in vulkan games unless the games are being presented with dxgi. For example, the game Doom Eternal used vulkan but also a dxgi interop layer for their HDR implementation, and SK is known to actually be able to hook into that vulkan game only if/when using HDR -- also I believe a delayed global injection of Special K was needed for that game last time this was done. I've personally also successfully hooked Special K in a d3d9 game running with dxvk/vulkan and then presented through dxgi via nvidia's interop (Special K hooked into the dxgi final presentation interop layer). This allowed for scRGB HDR with this d3d9 game, which used to not be an option (Special K's scRGB HDR only works with games using dxgi).

I think this whole PR/conversation may have started because someone wanted HDR with yuzu vulkan games and they ran into a problem with this... 

anyway, i agree with not outright removing the +1 with the windowed/borderless setting here unless a way is found and applied to present the game with dxgi or some way (that is not exclusive to one gpu vendor) to prevent the vulkan game from being sent to FSE land -- since that is undesirable by default due to FSE's always-on-top situation preventing other apps and windows from layering on top of the game as it has been mentioned a few times yeah.

giias:
@NativeGamer081 did you set OGL_DX_PRESENT_DEBUG to 0x00080001 in the nvidia Profile Inspector and also set [Display.Output]ForceWindowed=true in Special K's config file for the game you're trying to run with yuzu ? also, i believe you gotta do that with the game using the exclusive fullscreen setting in yuzu. I suspect you didn't and that your issue would be fixed if you do. I bet marat hasn't tried this either

hm and the dpi scaling approach to prevent the vulkan game from going into FSE that @Morph1984 found is interest... 
it doesn't solve the situation for people who like to run the display at 100% scaling (I myself prefer 100% scaling on my 4k 43" display), but at least it could be useful to people with 200% scaling (or devicePixelRatioF() is > 1) yeah. I don't know if this has been implemented or if there is enough benefit/demand (but I guess it's fine if it doesn't break anything anyway)

giias:
> > > where advanced users can opt-out of the +1 -- a win-win scenario for all parties
> > 
> > 
> > This should be what the exclusive fullscreen setting does, no?
> 
> FSE and hardware accelerated borderless are two different things; especially with alt tab behavior -- or when trying to do something like the nvidia interlop -> inject reshade (vulkan -> dx11) -> HDR

I think they get that, @marat569, and i think you're forgetting that vulkan on borderless gets "promoted" to FSE

you're asking for a setting that sort of already exists (without realizing it) -- just not with the label you'd expect...

yuzu's exclusive fullscreen setting is basically what you want (because it doesn't have the +1 thing). You'd then want to use nvidia/SK to basically override yuzu's exclusive fullscreen vulkan game into dxgi windowed/borderless. Try it :)

Kelebek1:
How can the applets disappear? Shouldn't they just be composited into the framebuffer by Yuzu in nvflinger before the frame is ever sent out? How can they disappear?

NativeGamer081:
> 

I did everything you said, even down to adding [Display.Output]ForceWindowed=true to the SpecialK.ini for Yuzu, yet I still get the issue shown in the image.  I did everything, from making sure Vulkan/OpenGL present method = Prefer layered on DXGI Swapchain in Nvidia Control Panel, to adding OGL_DX_PRESENT_DEBUG to 0x00080001 in Nvidia Profile Inspector for Yuzu, and and even setting Display.Output]ForceWindowed=true to the SpecialK.ini for Yuzu. 
Again, Ryujinx works perfectly fine using the Vulkan API with Special K.
Screenshot 2023-02-13 062615

giias:
> > 
> 
> I did everything you said, even down to adding [Display.Output]ForceWindowed=true to the SpecialK.ini for Yuzu, yet I still get the issue shown in the image. I did everything, from making sure Vulkan/OpenGL present method = Prefer layered on DXGI Swapchain in Nvidia Control Panel, to adding OGL_DX_PRESENT_DEBUG to 0x00080001 in Nvidia Profile Inspector for Yuzu, and and even setting Display.Output]ForceWindowed=true to the SpecialK.ini for Yuzu. Again, Ryujinx works perfectly fine using the Vulkan API with Special K. Screenshot 2023-02-13 062615

did you actually set it to exclusive fullscreen in yuzu ? 

the exclusive fullscreen setting shouldn't have the +1... 

also, instead of doing it in a yuzu profile, set OGL_DX_PRESENT_DEBUG to 0x00080001 in the global profile in the nvidia profile inspector (so in: _GLOBAL_DRIVER_PROFILE (Base Profile)) 

hopefully that works for you. 

I don't have yuzu set up right now, but forcing a vulkan game out of FSE has worked for me and other people with the method I mentioned. Beyond that, I'm not sure what else to tell you...

FearlessTobi:
@Kelebek1 As far as I understand, the applets are drawn in QT and not as part of the emulated operating system.
The ideal solution would be to support LLE applets so we can get rid of all these stupid workarounds.
But with our current state of the kernel and multiprocess emulation, I don't think that's feasible anytime soon.

NativeGamer081:
> @Kelebek1 As far as I understand, the applets are drawn in QT and not as part of the emulated operating system. The ideal solution would be to support LLE applets so we can get rid of all these stupid workarounds. But with our current state of the kernel and multiprocess emulation, I don't think that's feasible anytime soon.

Yes, I did all that, yet the result is still the same.

NativeGamer081:
> > > 
> > 
> > 
> > I did everything you said, even down to adding [Display.Output]ForceWindowed=true to the SpecialK.ini for Yuzu, yet I still get the issue shown in the image. I did everything, from making sure Vulkan/OpenGL present method = Prefer layered on DXGI Swapchain in Nvidia Control Panel, to adding OGL_DX_PRESENT_DEBUG to 0x00080001 in Nvidia Profile Inspector for Yuzu, and and even setting Display.Output]ForceWindowed=true to the SpecialK.ini for Yuzu. Again, Ryujinx works perfectly fine using the Vulkan API with Special K.
> 
> did you actually set it to exclusive fullscreen in yuzu ?
> 
> the exclusive fullscreen setting shouldn't have the +1...
> 
> also, instead of doing it in a yuzu profile, set OGL_DX_PRESENT_DEBUG to 0x00080001 in the global profile in the nvidia profile inspector (so in: _GLOBAL_DRIVER_PROFILE (Base Profile))
> 
> hopefully that works for you.
> 
> I don't have yuzu set up right now, but forcing a vulkan game out of FSE has worked for me and other people with the method I mentioned. Beyond that, I'm not sure what else to tell you...

Yes, I did all that, yet the result is still the same.

NativeGamer081:
Also, If I run Yuzu windowed without going into full screen, this is the result. 
Screenshot 2023-02-13 165545

NativeGamer081:
Also, I don't know if this helps, but if I run Yuzu in windowed mode, this is the result. 
Again, this issue only occurs with Vulkan API on Yuzu, OpenGL doesn't exhibit this issue. FYI, Ryujinx works fine on both Vulkan and OpenGL.
Desktop Screenshot 2023 02 13 - 17 11 44 06

Morph1984:
Further findings in #9811 show that it has issues with HDR

NativeGamer081:
> 

Same thing occurs if I disable HDR.

NativeGamer081:
> Further findings in #9811 show that it has issues with HDR

Same thing occurs if I disable HDR.

NativeGamer081:
> Further findings in #9811 show that it has issues with HDR

The same thing occurs even if I disable HDR! This issue is specific to Yuzu using Vulkan. As I said, and I'll repeat myself one last time. Vulkan with Ryujinx doesn't exibit this issue. If I want to use Special K with Yuzu, it can only be done using OpenGL. 
Screenshot 2023-02-18 155651

Morph1984:
If Special K doesn't work with Vulkan, then what makes you believe that this PR would solve the issue? It is an indication of a deeper issue

NativeGamer081:
> If Special K doesn't work with Vulkan, then what makes you believe that this PR would solve the issue? It is an indication of a deeper issue

It does work with Vulkan, only with Ryujinx. This indicates it's an issue with Yuzu's vulkan backend, or qt.

liamwhite:
@NativeGamer081 If the code change in this pull request does not solve the issue, I'm not sure I understand why it's relevant here (it should be a separate feature request on the tracker).

NativeGamer081:
> @NativeGamer081 If the code change in this pull request does not solve the issue, I'm not sure I understand why it's relevant here (it should be a separate feature request on the tracker).

I guess I'll have to use Ryujinx henceforth, If I want to use HDR with Vulkan on a Switch emulator, since you people can't acknowledge there's an issue with Yuzu's code. Otherwise, if I wish to play Yuzu in HDR, i'm forced to use OpenGL which has alot of shader compilation stutter.

Morph1984:
> It does work with Vulkan, only with Ryujinx. This indicates it's an issue with Yuzu's vulkan backend, or Qt.
I'm implying yuzu's Vulkan, not Ryujinx.
> since you people can't acknowledge there's an issue with Yuzu's code
There was a claim that this PR solves said issue, but it clearly does not. No one here indicated that there was a deeper issue until you showed screenshots of a yuzu window that was not fullscreen, and since that doesn't work, then this PR won't solve that issue either.

We are not against solving these issues, but when the issue is mischaracterized we take issue with that, because claims are being made that are not true.

marat569:
His issue is less with vulkan/yuzu and how nvidia handles the dxgi interlop layer.

The DXGI layer R E A L L Y dislikes any form of resizing/odd shaped windows -- ideally you'd get a swapchain with a frame buffer the same size as the display.

NativeGamer081:
> There was a claim that this PR solves said issue, but it clearly does not. No one here indicated that there was a deeper issue until you showed screenshots of a yuzu window that was not fullscreen, and since that doesn't work, then this PR won't solve that issue either.
> 
But as you have already seen, even in Fullscreen the issue persists.

NativeGamer081:
> His issue is less with vulkan/yuzu and how nvidia handles the dxgi interlop layer.
> 
> The DXGI layer R E A L L Y dislikes any form of resizing/odd shaped windows -- ideally you'd get a swapchain with a frame buffer the same size as the display.

 But as you have already seen, even in Fullscreen the issue persists, even if I disable HDR. OpenGL with Yuzu doesn't exhibit this issue. Ryujinx works with both Vulkan and OpenGL. How does this not indicate the issue lays with Yuzu, if Ryujinx works fine with Vulkan and OpenGL, yet I can only get Special K to work with Yuzu only with OpenGL.

FearlessTobi:
Thanks for your contribution, however I'm going to close this PR for the following reasons:
 1. Changing yuzu's code to remove the `+1` breaks QT overlays such as our software keyboard, because they can't be shownontop of the render window anymore. As far as I know, there's no fix we can implement for this on our side, except for this ugly workaround.
 2. It causes issues with HDR, as described more closely in https://github.com/yuzu-emu/yuzu/pull/9811
 3. There's already the option to enable Exclusive Fullscreen in the yuzu settings.

PR #9802, opened 2023-02-14
Re-add the invalidate_texture_data_cache register

Kelebek1:
This register seems to be real, and stops the undefined behaviour flickering on XC3 terrain. We used to have this, but it seems to have been removed by accident.

### DIFF
--- a/src/video_core/engines/maxwell_3d.cpp
+++ b/src/video_core/engines/maxwell_3d.cpp
@@ -186,6 +186,7 @@ bool Maxwell3D::IsMethodExecutable(u32 method) {
     case MAXWELL3D_REG_INDEX(launch_dma):
     case MAXWELL3D_REG_INDEX(inline_data):
     case MAXWELL3D_REG_INDEX(fragment_barrier):
+    case MAXWELL3D_REG_INDEX(invalidate_texture_data_cache):
     case MAXWELL3D_REG_INDEX(tiled_cache_barrier):
         return true;
     default:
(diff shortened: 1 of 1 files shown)


liamwhite (approved):

src/video_core/engines/maxwell_3d.cpp
```diff
@@ -375,6 +375,9 @@ void Maxwell3D::ProcessMethodCall(u32 method, u32 argument, u32 nonshadow_argume
         return;
     case MAXWELL3D_REG_INDEX(fragment_barrier):
         return rasterizer->FragmentBarrier();
+    case MAXWELL3D_REG_INDEX(invalidate_texture_data_cache):
```
vonchenplus, line 378: Add invalidate_texture_data_cache method to IsMethodExecutable, Otherwise method will not be invoke

vonchenplus (changes requested):

vonchenplus (approved):

goldenx86:
Testing seems to show this is a partial solution.
Depending of camera angles, the behaviour changed from showing a mix of both textures, to showing only one of them.




Here's the same wall texture in Breath of the Wild, just at different camera angles.

AaronKelley:
I did a build based on this PR for testing myself, and it seems to have made things notably better in XC3, but like @goldenx86 said it's a partial solution.  Some spots where I was able to reproduce terrain flickering before seem to have been fixed, but I'm still able to spot some.  It seems more like geometry flickering than texture flickering most of the time, though, so it might be a different issue...  (I cannot reproduce it in OpenGL, only Vulkan.  2× scaling / high accuracy, GeForce RTX 3080 Ti.)

Animation2
Animation1

bunnei (approved):

PR #4949, opened 2020-11-19
main: Enable High DPI fixes for Qt >= 5.14

Morph1984:
This uses Qt's new high DPI application attributes for scaling the current window.
However, these aren't perfect as scaling with non integer scales will cause artifacts in UI, icons and other elements.
Therefore, we use a heuristic to select an appropriate integer scale value depending on the current screen resolution and applies this to the application.

Fixes #4904
Fixes #7383
Fixes #8051

PR #9815, opened 2023-02-16
Qt: Fix mouse scalling

### LINKED ISSUE #7853: R analog stick is not centered when mouse inputs are binded to it
**Version:** yuzu 908 (2022-02-04)
**Issue:** R analog stick is not centered when mouse inputs are binded to it 

The control settings are all the default ones for keyboard/mouse input as seen below:

Following pictures display the debugging controller P1 when moving the cursor to different corners:
**Top left:**

**Top right:**

**Bottom right:**

**Bottom left:**

From these pictures the centre appears to be at the top left of the screen. 
Currently in BoTW, the camera is stuck looking at the ceiling and can only rotate clockwise.

There are similar results when binding the L analog (centre is not centered).

german77:
Since #4949 any user that had a window scale different from 100% will have wrong mouse coordinates. This PR fixes the issue by removing the workaround for scaling since now it should be 1:1 at any scale. 

This PR also splits mouse input into 3 different devices. A physical mouse, touch and finally one for controllers. This solves the issue Zeikken85 had where having `emulated mouse` enabled had a weird behavior on stick input. As a final protection against setting invalid configs a warning message will show up if emulated mouse is enabled.

Fixes #7853

### DIFF
--- a/src/core/hid/emulated_console.cpp
+++ b/src/core/hid/emulated_console.cpp
@@ -23,7 +23,8 @@ void EmulatedConsole::SetTouchParams() {
 
     // We can't use mouse as touch if native mouse is enabled
     if (!Settings::values.mouse_enabled) {
-        touch_params[index++] = Common::ParamPackage{"engine:mouse,axis_x:10,axis_y:11,button:0"};
+        touch_params[index++] =
+            Common::ParamPackage{"engine:mouse,axis_x:0,axis_y:1,button:0,port:2"};
     }
 
     touch_params[index++] =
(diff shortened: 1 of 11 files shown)


frykillr1985:
issue still present.

https://user-images.githubusercontent.com/28325592/219393951-2ab5132e-b31a-4b95-870a-125437c92306.mp4

with your test build.

Zeikken85:
With @german77 the issue was found.
Need to untick Mouse in advanced settings of the controllers menu.
Also make sure that Mouse Panning is active while in-game.
image

liamwhite (approved):

PR #9817, opened 2023-02-16
yuzu: Write to config file on important config changes

german77:
Qt isn't always writing changes on save. This causes config to be lost on crash. This PR ensures all changes are always saved on the file.

Additionally restarting a game didn't refresh the per game config since the game wasn't stopped before booting the game again.

### DIFF
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -1103,6 +1103,7 @@ void Config::SaveValues() {
     SaveRendererValues();
     SaveAudioValues();
     SaveSystemValues();
+    qt_config->sync();
 }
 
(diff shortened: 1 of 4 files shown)


liamwhite (approved):

theboy181:
:) drooooool!

theboy181:
I was able to make a setting then force quite the app, and settings were committed!  I tested global and per game. I have noticed a different issue that may be unrelated to QT. 

When you set a per game setting for 21:9 then load the game and use there reset function, the game will be in a 21:9 window but display as 16:9

I hope this can be fixed too.

liamwhite (approved):

lioncash (approved):

PR #9822, opened 2023-02-17
 buffer_cache: Add logic for non-NVN storage buffer tracking

ameerj:
TL;DR Fixes some graphical issues and a crash during the initial loading screen in `Doom Eternal` with the latest update, and potentially other titles.

#### Problematic assumptions

In NVN, global memory ops such as `STG/LDG` generally act on SSBOs whose address and size is stored in a driver reserved constant buffer at index 0. The memory layout for this reserved memory is 16 byte aligned in the format:
```
u32 ssbo_0_addr_lo
u32 ssbo_0_addr_hi
u32 ssbo_0_size
u32 padding
```
Currently, the shader Global To Storage pass uses this information to add an optimization to directly bind the ssbo that the memory op is acting on, rather than go through the indirection of the constant buffer.

Doom Eternal uncovered a flaw in this optimization. This title seems to be using global memory ops on raw SSBO addresses (potentially something equivalent to `VK_KHR_buffer_device_address`) stored in custom constant buffers, which do not follow the format of NVN's cbuf_0 convention.

The layout of the cbuf Doom Eternal uses is tightly packed addresses, with no indicator of the expected binding size.
```
u32 ssbo_0_addr_lo
u32 ssbo_0_addr_hi
u32 ssbo_1_addr_lo
u32 ssbo_1_addr_hi
...
```
The optimization pass would sometimes read these global memory operations assuming they were in the NVN convention, which would then use the adjacent buffer's `addr_lo` as the size for the ssbo, leading to unmapped reads and memory overflows.

#### Solution (for now)

For SSBOs that are not in the NVN constant buffer, the size can be estimated based on the arguments to `MapBufferEx` for its address.
This essentially queries the size of the mapped buffer that the SSBO addr resides in, and binds that entire buffer range for the shader to access.

This seems overkill, since the SSBO for the specific draw call/dispatch may be a subregion of the entire buffer.
I'm not sure how else to confirm the SSBO size, so for now this seems like the safest approach.

### DIFF
--- a/src/shader_recompiler/ir_opt/global_memory_to_storage_buffer_pass.cpp
+++ b/src/shader_recompiler/ir_opt/global_memory_to_storage_buffer_pass.cpp
@@ -35,6 +35,7 @@ struct Bias {
     u32 index;
     u32 offset_begin;
     u32 offset_end;
+    u32 alignment;
 };
 
 using boost::container::flat_set;
@@ -349,7 +350,8 @@ std::optional<StorageBufferAddr> Track(const IR::Value& value, const Bias* bias)
             .index = index.U32(),
             .offset = offset.U32(),
         };
-        if (!Common::IsAligned(storage_buffer.offset, 16)) {
+        const u32 alignment{bias ? bias->alignment : 8U};
+        if (!Common::IsAligned(storage_buffer.offset, alignment)) {
             // The SSBO pointer has to be aligned
             return std::nullopt;
         }
@@ -371,6 +373,7 @@ void CollectStorageBuffers(IR::Block& block, IR::Inst& inst, StorageInfo& info) 
         .index = 0,
         .offset_begin = 0x110,
         .offset_end = 0x610,
+        .alignment = 16,
     };
     // Track the low address of the instruction
     const std::optional<LowAddrInfo> low_addr_info{TrackLowAddress(&inst)};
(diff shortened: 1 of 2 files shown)


Kelebek1:
Is it possible that cbuf index 0 could still not be in the expected format? Would it not be worth checking that read cbuf size and falling back to the entire range if the size is a silly value, instead of forcibly returning + 8 for cbuf 0?

ameerj:
> Is it possible that cbuf index 0 could still not be in the expected format? Would it not be worth checking that read cbuf size and falling back to the entire range if the size is a silly value, instead of forcibly returning + 8 for cbuf 0?

It's a nice thought, but I foresee a few issues with this. First is that who decides what a "silly" value is? And the fallback is very likely going to overshoot the size compared to what it should be.

Since cbuf 0 is driver reserved afaik, any corruption there would most likely point to an issue with nvdrv emulation.

FernandoS27:
> > Is it possible that cbuf index 0 could still not be in the expected format? Would it not be worth checking that read cbuf size and falling back to the entire range if the size is a silly value, instead of forcibly returning + 8 for cbuf 0?
> 
> It's a nice thought, but I foresee a few issues with this. First is that who decides what a "silly" value is? And the fallback is very likely going to overshoot the size compared to what it should be.
> 
> Since cbuf 0 is driver reserved afaik, any corruption there would most likely point to an issue with nvdrv emulation.

The environment has a method called "IsPropietaryDriver" which allows you to check that. In general, the const buffer 0 fallback hack should only be used in that case.

ameerj:
> The environment has a method called "IsPropietaryDriver" which allows you to check that. In general, the const buffer 0 fallback hack should only be used in that case.

@FernandoS27 Not entirely sure what you have in mind. But given that it would potentially alter the existing logic, and this PR does not, I would rather the driver specific checks be done and tested in a follow-up PR.

liamwhite (approved):

PR #9826, opened 2023-02-18
Allow AMD proprietary drivers to disable vsync

goldenx86:
The original logic for presentation was done before the vsync toggle in graphics > advanced was added, afaik. 

NVIDIA drivers can run games at framerates higher than the display's refresh rate while in mailbox (triple buffer) mode, but AMD proprietary drivers get locked to the refresh rate of the display. While having tearing is undesired, 

Smash players use high framerate mods to reduce the hardcoded input delay of the game, and not everyone has access to 120+ Hz monitors. 

Letting <120 Hz AMD users have the choice between being able to use the 120 FPS mod with tearing, or stopping tearing and sticking to 60 FPS is a better option IMO. To do this, just use the VSync toggle in advanced graphic settings.

ayymd

### DIFF
--- a/src/video_core/renderer_vulkan/vk_swapchain.cpp
+++ b/src/video_core/renderer_vulkan/vk_swapchain.cpp
@@ -33,20 +33,29 @@ VkSurfaceFormatKHR ChooseSwapSurfaceFormat(vk::Span<VkSurfaceFormatKHR> formats)
     return found != formats.end() ? *found : formats[0];
 }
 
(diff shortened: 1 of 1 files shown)


coolllman:
I have tearing with this pr (rtx3060 and vsync off), before this i haven't tearing

goldenx86:
The idea is to allow tearing with vsync disabled, as that solves issues with non-NVIDIA hardware.
The previous implementation always forced vsync on NVIDIA.

I'm closing the PR, a proper UI list with all 3 options is preferred. I'll start it after the progress report.

Issue #9840, opened 2023-02-21
Link error when system cpp-httplib using


svalx:
### Issue

On Linux build with **cpp-httplib** by package manager installed, I get a linking error - no openssl symbols founded. This patch fixing it. Maybe it should be included in the mainline?

```
Index: yuzu-0.1339/externals/CMakeLists.txt
===================================================================
--- yuzu-0.1339.orig/externals/CMakeLists.txt
+++ yuzu-0.1339/externals/CMakeLists.txt
@@ -103,7 +103,6 @@ add_subdirectory(sirit EXCLUDE_FROM_ALL)
 # httplib
 if (ENABLE_WEB_SERVICE AND NOT TARGET httplib::httplib)
     if (NOT WIN32)
-        find_package(OpenSSL 1.1)
         if (OPENSSL_FOUND)
             set(OPENSSL_LIBRARIES OpenSSL::SSL OpenSSL::Crypto)
         endif()
Index: yuzu-0.1339/CMakeLists.txt
===================================================================
--- yuzu-0.1339.orig/CMakeLists.txt
+++ yuzu-0.1339/CMakeLists.txt
@@ -214,6 +214,7 @@ find_package(nlohmann_json 3.8 REQUIRED)
 find_package(Opus 1.3 MODULE)
 find_package(ZLIB 1.2 REQUIRED)
 find_package(zstd 1.5 REQUIRED)
+find_package(OpenSSL 1.1)

 if (NOT YUZU_USE_EXTERNAL_VULKAN_HEADERS)
     find_package(Vulkan 1.3.238 REQUIRED)
@@ -242,6 +243,9 @@ endif()
 if (ENABLE_WEB_SERVICE)
     find_package(cpp-jwt 1.4 CONFIG)
     find_package(httplib 0.11 MODULE)
+    if (httplib_FOUND AND OPENSSL_FOUND)
+       set(LINK_OPENSSL ON)
+    endif()
 endif()

 if (YUZU_TESTS)
Index: yuzu-0.1339/src/web_service/CMakeLists.txt
===================================================================
--- yuzu-0.1339.orig/src/web_service/CMakeLists.txt
+++ yuzu-0.1339/src/web_service/CMakeLists.txt
@@ -19,6 +19,10 @@ add_library(web_service STATIC
 create_target_directory_groups(web_service)
[...]

PR #9841, opened 2023-02-21
externals: Update cpp-httplib to latest

abouvier:
`find_package(httplib 0.11 MODULE)` will refuse library with version >= 0.12.

Maybe solve https://github.com/yuzu-emu/yuzu/issues/9840.

In fact this won't solve this particular issue because the library involved uses pkgconfig file instead of cmake config file (so any version greater than 0.11 would be accepted), but the update is welcome anyway :p

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -241,7 +241,7 @@ endif()
 
 if (ENABLE_WEB_SERVICE)
     find_package(cpp-jwt 1.4 CONFIG)
-    find_package(httplib 0.11 MODULE)
+    find_package(httplib 0.12 MODULE)
 endif()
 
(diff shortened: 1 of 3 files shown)


liamwhite (approved):

Tachi107:
This should also update various places that needlessly convert `std::string`s to C strings when calling cpp-httplib, since lots of places that accepted a `const char*` now ask for an `std::string`

Tachi107:
On Tue Feb 21, 2023 at 8:58 PM CET, Alexandre Bouvier wrote:
> There are more than these two?

I guess so :)

I haven't checked myself, but grep is always your friend ;)

PR #9848, opened 2023-02-22
input_common: Implement dedicated motion from mouse

german77:
Yuzu already had motion from mouse. However it was pretty bad. Metroid prime on hybrid mode uses only gyro data. Even better just two axis. This is 1:1 mapping and provides excellent camera control with mouse panning on this game.

The dedicated motion doesn't include any clamping of inputs and has a different sensitivity allowing to be used without changing the config. As for other games they will probably still have bad motion as they use full motion instead of just gyro.

### DIFF
--- a/src/core/hid/emulated_controller.cpp
+++ b/src/core/hid/emulated_controller.cpp
@@ -363,7 +363,17 @@ void EmulatedController::ReloadInput() {
                     SetMotion(callback, index);
                 },
         });
-        motion_devices[index]->ForceUpdate();
+
+        // Restore motion state
+        auto& emulated_motion = controller.motion_values[index].emulated;
+        auto& motion = controller.motion_state[index];
+        emulated_motion.ResetRotations();
+        emulated_motion.ResetQuaternion();
+        motion.accel = emulated_motion.GetAcceleration();
+        motion.gyro = emulated_motion.GetGyroscope();
+        motion.rotation = emulated_motion.GetRotations();
+        motion.orientation = emulated_motion.GetOrientation();
+        motion.is_at_rest = !emulated_motion.IsMoving(motion_sensitivity);
     }
 
     for (std::size_t index = 0; index < camera_devices.size(); ++index) {
--- a/src/core/hid/motion_input.cpp
+++ b/src/core/hid/motion_input.cpp
@@ -10,6 +10,8 @@ MotionInput::MotionInput() {
     // Initialize PID constants with default values
     SetPID(0.3f, 0.005f, 0.0f);
     SetGyroThreshold(ThresholdStandard);
+    ResetQuaternion();
+    ResetRotations();
 }
 
 void MotionInput::SetPID(f32 new_kp, f32 new_ki, f32 new_kd) {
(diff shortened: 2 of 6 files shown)


liamwhite (approved):

saeedAlshaikh:
Hello , thanks a ton for all of these efforts.

Whenever I chose hybrid controls on Metroid prime and than toggle mouse panning via a hotkey , I can't move the camera at all.
Is there a special setting for gyro ? because on dual mode I can move with mouse freely with a slight lag.

german77:
Hybrid requires the button R to be pressed to move the camera

SixelAlexiS90:
> 

Hi!
You DON'T have to use the "Hybrid" controls but the "Dual Stick" controls instead.
After that go in Emulation > Configure > Controls, emulate the Pro Controller and select "Keyboard" under "Input Device" (not "Keyboard/Mouse" since it will map the mouse on the right stick and you don't want that or it will go in conflict with the motion controls we are going to setup).

Still in the "Controls" window, click on "Motion 1" and when it says "Shake!" just click with the left click of the mouse and you will see "mouse" instead of "Shake!".
This means that you mapped the mouse movement on the motion controls.

Don't forget that you have to manually bind all buttons on keyboard and mouse to make it actually working like a regular FPS game.
If you don't want to mess up with that for now, here you can find a profile that I've modified to work on Yuzu, I've already set up general controls and the "Motion 1" setting as well:

[Metroid Prime Remastered (M+K).txt](https://github.com/yuzu-emu/yuzu/files/10957621/Metroid.Prime.Remastered.M%2BK.txt)

Rename the ".txt" file in a ".ini" instead and put it in:
C:\Users\*username*\AppData\Roaming\yuzu\config\input\

You can load the profile in the Emulation > Configure > Controls page.

It's not optimal yet (see navigating the map) but it's a good base to make the game act overall like a mouse and keyboard game.


Once in-game, after setting "Dual Stick" controls, go down on "Camera" and select "Gyro + Stick".
After that you can change the camera sens by changing the "Camera Gyro Sensitivity X" and "Camera Gyro Sensitivity Y" settings.

I suggest using a "+10" sens on both axis (you will have to start from "0" and count the steps with the d-pad) since the game adds a weird negative camera acceleration on very low sens (see "0" on both axis) and jittery movements on higher sens (see "+50").

After that you tweak the Yuzu "Mouse Sensitivity" percentage (under Emulation > Configure > Controls > Advanced, near the "Enable mouse panning" setting) and your mouse DPI accordingly.

For a 25cm/360° sensitivity I use "+10" in-game Camera Gyro Sensitivity and "10%" on "Mouse Sensitivity" with a 1800 DPI mouse.

For lower mouse DPI you can use an higher "Mouse Sensitivity" instead, you will have to find a balance by tweaking this setting depending on your favorite DPI setting.

**IMPORTANT**: remember to enable the "Enable mouse panning" setting while in-game with the "Ctrl+F9" shortcut, this will make the mouse cursor stick on the center of the screen and make the mouse actually work with gyro controls.

Press "Ctrl+F9" again if you need to continue use the PC for changing settings and stuff.

**IMPORTANT 2**: be sure to close any other window and programs once you go fullscreen in Yuzu and enable the mouse panning. If there is any window open under Yuzu once you go in fullscreen, the mouse will move erratically. I've already reported the issue and hopefully it will be fixed at some point. Enjoy :)

PR #9849, opened 2023-02-22
texture_cache: Add asynchronous ASTC texture decoding

ameerj:
Currently, the GPU thread blocks waiting for ASTC textures to be decoded.

This PR adds another weapon in the war against ASTC: deferring ASTC texture decoding to a separate thread, unblocking the GPU thread. This significantly reduces the notorious stutter caused by ASTC.
The setting can be found under the Advanced Graphics configuration tab.

This uses the CPU decoding implementation, since using the compute shader for this would have been much more complicated for potentially unnoticeable gains.

This does result in some missing graphics while the texture is being decoded, but these are usually quick to clear up unless the system has a low thread count or the title uses many ASTC textures.

An example of missing graphics from async ASTC loading (1 ASTC decode thread):

https://user-images.githubusercontent.com/52414509/220536773-a2508884-1755-4ae5-9432-92f8e945c5a6.mp4

### DIFF
--- a/src/common/scratch_buffer.h
+++ b/src/common/scratch_buffer.h
@@ -23,6 +23,7 @@ public:
           buffer{Common::make_unique_for_overwrite<T[]>(initial_capacity)} {}
 
     ~ScratchBuffer() = default;
+    ScratchBuffer(ScratchBuffer&&) = default;
 
     /// This will only grow the buffer's capacity if size is greater than the current capacity.
     /// The previously held data will remain intact.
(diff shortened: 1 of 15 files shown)


gidoly:
async cpu astc loading
 
https://user-images.githubusercontent.com/66776795/220783143-d7e42f27-432b-4893-8dde-32dfce4f7033.mp4

vs

compute shader decode

https://user-images.githubusercontent.com/66776795/220783511-f67002e2-e553-4df7-a27f-fd3f901d4887.mp4

compute shader is now slower than cpu decoding with async.
my cpu is 5800x3d 16 which has thread and gpu is 4090

gidoly:
also can you make this hack as none default so user can not see graphic glitch with default setting. it will be nightmare to manage support with this default on.it also only benifit few excepton like astral chain and bayonetta 3 which makes me prefer to be off as default to avoid confusing issue report caused by other games

ameerj:
Defaulting this setting on was accidental, thanks for catching that!

liamwhite (approved):

vasishath:
I might sound complete noob here but is it not possible to cache the decoded ASTC?

PR #9877, opened 2023-02-28
cmake: use correct boost imported targets

abouvier:
The true imported target name since CMake 3.15 is `Boost::headers` instead of `Boost::boost`. Also remove old hacks not needed anymore (hopefully :p).


liamwhite (approved):
Very nice simplification

PR #9889, opened 2023-03-02
core_timing: Reduce CPU usage on Windows

Morph1984:
This PR aims to reduce CPU usage on Windows caused by the CoreTiming event loop in 2 ways:

1. Reducing the cost of querying the current time
This was done by implementing a high resolution `SteadyClock`, which is ~2.5x faster compared to querying RDTSC.
While `QueryPerformanceCounter` has a lower precision than RDTSC, it offers sufficient precision (microsecond precision) for our needs.
2. Increasing the timer resolution
Increasing the timer resolution from 1ms to 0.5ms allows us to have 500 microsecond sleeps, which significantly reduces the amount of time needed during the spin-wait.
This was achieved by utilizing the undocumented `ntdll` functions `NtSetTimerResolution` and `NtDelayExecution`.

This results in significantly lower CPU usage in core timing heavy games such as Kirby and the Forgotten Land, Super Smash Bros. Ultimate, and many others. 
Furthermore, this may also improve performance for CPUs that have low core counts and/or are power limited, such as those used in laptops or gaming handhelds.

Additional changes were made to this PR, including:
1. Targeting the Windows 10 SDK
2. Setting the current process to HighQoS, which is a hint to the Windows scheduler to prioritize performance cores over efficiency cores, which is beneficial for processors such as Intel's Alder and Raptor Lake

### DIFF
--- a/src/common/wall_clock.cpp
+++ b/src/common/wall_clock.cpp
@@ -81,4 +81,9 @@ std::unique_ptr<WallClock> CreateBestMatchingClock(u64 emulated_cpu_frequency,
 
 #endif
 
+std::unique_ptr<WallClock> CreateStandardWallClock(u64 emulated_cpu_frequency,
+                                                   u64 emulated_clock_frequency) {
+    return std::make_unique<StandardWallClock>(emulated_cpu_frequency, emulated_clock_frequency);
+}
+
 } // namespace Common
--- a/src/common/wall_clock.cpp
+++ b/src/common/wall_clock.cpp
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: Copyright 2020 yuzu Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include "common/steady_clock.h"
 #include "common/uint128.h"
 #include "common/wall_clock.h"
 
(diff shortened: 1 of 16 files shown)


Hasster1:
Are there any specific configuration that users need to use? For me (i5-4210u, GeForce 840m) games either gets stuck on launching, or freeze the emulator after it.

src/core/core_timing.cpp
```diff
@@ -186,14 +191,14 @@ void CoreTiming::ResetTicks() {
 
 u64 CoreTiming::GetCPUTicks() const {
     if (is_multicore) {
```
FernandoS27, line 193: since you are here add `[[likely]]` here.

src/core/core_timing.cpp
```diff
@@ -186,14 +191,14 @@ void CoreTiming::ResetTicks() {
...
-        return clock->GetCPUCycles();
+        return cpu_clock->GetCPUCycles();
     }
     return ticks;
 }
 
 u64 CoreTiming::GetClockTicks() const {
     if (is_multicore) {
```
FernandoS27, line 200: ditto.

src/common/wall_clock.cpp
```diff
@@ -11,53 +12,41 @@
...
+        return std::chrono::duration_cast<std::chrono::milliseconds>(GetTimeNS());
     }
 
     u64 GetClockCycles() override {
-        std::chrono::nanoseconds time_now = GetTimeNS();
-        const u128 temporary =
-            Common::Multiply64Into128(time_now.count(), emulated_clock_frequency);
-        return Common::Divide128On32(temporary, 1000000000).first;
+        const u128 temp = Common::Multiply64Into128(GetTimeNS().count(), emulated_clock_frequency);
+        return Common::Divide128On32(temp, NS_RATIO).first;
```
FernandoS27, line 36: since NS_RATIO is always larger than emulated_clock_frequency. You could turn this operations into a less costly 128-bits fixed point multiplication. This is very important because this particular function is called by the game's code often.

goldenx86:
@Hasster1 we don't target support of dual core CPUs.

Morph1984:
@Hasster1 Is this an issue with the PR or yuzu in general?

Hasster1:
> @Hasster1 Is this an issue with the PR or yuzu in general?

PR.

Hasster1:
I think that settings don't really matter, but it really feels like the more i launch the game, the more the chances that it'll work for a little more time. Here are the issues i've found for now: 
1 - Trying to stop the game at any point - either gets stuck on "closing software..." (most of the times, the emulator itself doesn't freeze) or exits the emulator without any messages. Trying to restart the game just freezes the emulator
2 - When the game goes past launching process, it might boot without audio, which (probably) increases the time for the game to eventually freeze.
3 - Lighter games have a better chance to not get stuck at launching.

It looks like logs aren't really useful when the game can't progress past the launching, but i'm still going to drop them here.
logs with "stopgameexit" at the end indicate that the emulator closed itself after i've pressed "Stop game", instead of me ending the emulator's process through the task manager.

[terraria_infinite_loading_stopgameexit.txt](https://github.com/yuzu-emu/yuzu/files/10880057/terraria_infinite_loading_stopgameexit.txt)
[terraria_ingame_freeze_noaudio.txt](https://github.com/yuzu-emu/yuzu/files/10880059/terraria_ingame_freeze_noaudio.txt)
[terraria_ingame_freeze_withaudio_stopgameexit.txt](https://github.com/yuzu-emu/yuzu/files/10880060/terraria_ingame_freeze_withaudio_stopgameexit.txt)
[terraria_infinite_loading.txt](https://github.com/yuzu-emu/yuzu/files/10880062/terraria_infinite_loading.txt)


I'm not sure if it'll help in any way, but i hope that it will.

Morph1984:
@Hasster I notice that you were testing the Windows MinGW build. Can you test the Windows MSVC build?
Also to ensure that it isn't my PR causing the issue, can you test the Windows MinGW build from #9884 and see if it exhibits the same behavior?

Tachi107:
Is `ntdll`'s ABI stable? I seem to remember that Windows only maintains compatibility for the Win32 API, but I'm probably wrong.

Hasster1:
You're right, the MinGW build on other PRs also causes the same hanging issue, and msvc actually works with no problems (except those that might appear because of this PR, but that's not the point here).

Morph1984:
> Is `ntdll`'s ABI stable? I seem to remember that Windows only maintains compatibility for the Win32 API, but I'm probably wrong.

Considering that they still maintain compatibility with DOS by disallowing filenames like COM, etc, it's safe to say that MS will not remove these user mode functions that have existed for a very long time

src/common/wall_clock.cpp
Morph1984: This is not used by the event clock

Morph1984: This can be optimized in a future PR by optimizing this into a multiplication by 102/100 and using libdivide

liamwhite (approved):

someshittyeletronicmusicband:
Just saw the progress report and these two undocumented functions being mentioned, `NtSetTimerResolution` and `NtDelayExecution`, while saying they are "new" and from Windows 10 SDK. The thing is, they have been around for almost 3 decades now. There are entries dating from 1997/2000 around the web documenting them, so I wouldn't it count as a benefit of dropping 7 and 8/8.1 support since they could very well support them.

http://undocumented.ntinternals.net/
https://web.archive.org/web/20050619080659/www.sysinternals.com/Information/HighResolutionTimers.html

PR #9896, opened 2023-03-04
Check all swizzle components for red, not just [0]

### LINKED ISSUE #9805: Story of seasons wonderful life visual glitch

### Affected Build(s)

Yuzu 1154

### Description of Issue

There are there lines that appear everywhere at certain times of day in game. Its a universal glitch everyone who trys to run the game on yuzu experiences.

### Expected Behavior

4b4916c5e7a47384a5afbd7abc285b4c68dce928

### Reproduction Steps

it happens on title screen and when loading any part of the game. its worse indoors

### Log File

dont know how to get log file

### System Configuration

CPU:
GPU/Driver:nvidia 2070
RAM:8gigs
OS:windows

Kelebek1:
Third time's the charm? Thanks to @bylaws for helping out with this one.

Instead of only checking component[0] for red, check all components. Red and green usage here should be exclusive (as per ByLaws), swizzles should only contain either red or green and not both. If no red is found it's assumed green, but if there's no green either and all the components are either zero or one, then that should also be fine, since they'll return constants anyway it doesn't matter if it tries to pull from depth or stencil.

Closes #9805. Hopefully for the last time.

### DIFF
--- a/src/video_core/renderer_opengl/gl_texture_cache.cpp
+++ b/src/video_core/renderer_opengl/gl_texture_cache.cpp
@@ -112,13 +112,17 @@ GLenum ImageTarget(Shader::TextureType type, int num_samples = 1) {
     return GL_NONE;
 }
 
-GLenum TextureMode(PixelFormat format, bool is_first) {
+GLenum TextureMode(PixelFormat format, std::array<SwizzleSource, 4> swizzle) {
+    bool any_r =
+        std::ranges::any_of(swizzle, [](SwizzleSource s) { return s == SwizzleSource::R; });
     switch (format) {
     case PixelFormat::D24_UNORM_S8_UINT:
     case PixelFormat::D32_FLOAT_S8_UINT:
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_opengl/gl_texture_cache.cpp
```diff
@@ -112,13 +112,17 @@ GLenum ImageTarget(Shader::TextureType type, int num_samples = 1) {
     return GL_NONE;
 }
 
-GLenum TextureMode(PixelFormat format, bool is_first) {
+GLenum TextureMode(PixelFormat format, std::span<SwizzleSource> swizzle) {
```
Morph1984, line 115: Passing this small 4 byte array by copy is likely to be faster than passing a span
```suggestion
GLenum TextureMode(PixelFormat format, std::array<SwizzleSource, 4> swizzle) {
```

Kelebek1: Should I just pass it by reference?

liamwhite (approved):

PR #9906, opened 2023-03-05
input_common: Increase mouse sensitivity range

german77:
After talking with some metroid fans we arrived at the conclusion that high dpi mouses move really fast and even 1% still too fast for yuzu. Motion from mouse also felt with a considerable delay and a bit choppy.

I addressed this issue by making the default sensitivity 50% instead of 10%, removing the smoothing filter, lowering the motion threshold, clamping the rotation speed and accumulating all inputs until they are delivered. This improves the feedback and play ability with mouse and keyboard.

### DIFF
--- a/src/input_common/drivers/mouse.cpp
+++ b/src/input_common/drivers/mouse.cpp
@@ -3,6 +3,7 @@
 
 #include <thread>
 #include <fmt/format.h>
+#include <math.h>
 
 #include "common/param_package.h"
 #include "common/settings.h"
@@ -11,8 +12,9 @@
 
 namespace InputCommon {
 constexpr int update_time = 10;
-constexpr float default_stick_sensitivity = 0.022f;
-constexpr float default_motion_sensitivity = 0.008f;
+constexpr float default_stick_sensitivity = 0.0044f;
+constexpr float default_motion_sensitivity = 0.0003f;
+constexpr float maximum_rotation_speed = 2.0f;
 constexpr int mouse_axis_x = 0;
 constexpr int mouse_axis_y = 1;
 constexpr int wheel_axis_x = 2;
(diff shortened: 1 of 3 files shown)


src/input_common/drivers/mouse.cpp
```diff
@@ -99,11 +101,13 @@ void Mouse::UpdateMotionInput() {
...
-        last_motion_change *= 0.93f;
-    } else {
-        last_motion_change.z *= 0.93f;
+    const float rotation_velocity = std::sqrt(last_motion_change.x * last_motion_change.x +
+                                              last_motion_change.y * last_motion_change.y);
+
+    if (rotation_velocity > maximum_rotation_speed / sensitivity) {
+        const float multiplier = maximum_rotation_speed / rotation_velocity / sensitivity;
+        last_motion_change.x = last_motion_change.x * multiplier;
+        last_motion_change.y = last_motion_change.y * multiplier;
```
Kewlan, line 110: This is to limit the max speed, right? Is this needed? If not, what's the purpose of this when using a mouse? I tested without this and think it's better for quick turnarounds.

Similar here: 
https://github.com/yuzu-emu/yuzu/blob/bd2aba183104bb88d66c9e48983b0840cd7339e4/src/input_common/drivers/mouse.cpp#L145-L150 Is this the same but for slow movements? I took this away too and it feels easier to aim more precise.

liamwhite (approved):

Kewlan (commented):
I tested this now and it feels much better! The main issue now is the input delay. When testing with uncapped FPS it feels perfect!

Some questions:

src/input_common/drivers/mouse.cpp
german77: I decided it would be a good idea to limit the turn speed to something decent since you can get crazy spins. If you feel it's limiting you we can bump the value up, right now is around 540° per second.

As for lines 145-150 those aren't related to motion but rather the analog stick so if you feel that changes something is pure placebo.

german77: As for the input delay I can't do much about it. Since we are already at the lowest we can achieve.

Kewlan: Increasing the max speed would be nice. For me `maximum_rotation_speed` doesn't need to be more than `2.0f`. But if an increased fps mod gets released, would this need to be even higher, or would lower work too since the game updates more frequently?

And yes, seems like the slow movement stuff was placebo!

bunnei (approved):

PR #9786, opened 2023-02-12
YFC - Engines: Implement Accelerate DMA Texture.

FernandoS27:
This implements host acceleration of the DMA engine. This is used in many games to download textures. Only implemented in Vulkan.

This improves performance in:
- Metroid Prime Remaster.
- Super Mario Sunshine.
- Super Mario Galaxy.

Among others.

PR #9913, opened 2023-03-07
AccelerateDMA: Refactor Buffer/Image copy code and implement for OGL

ameerj:
Deduplicates most of the AccelerateDMA Buffer <-> Image copy code added in #9786 and generalizes it within texture_cache.h rather than being rasterizer specific.

Also implements the code on the OpenGL rasterizer.


Kelebek1:
Seems like the image->ScaleUp/ScaleDown stuff is removed in this, is that intended?

FernandoS27:
Well ScaleUp/ScaleDown logic should be readded.

Now I had some plans for improving the perf which inclusing copying to 2 buffers (a async download buffer and a buffer cache buffer). Now that you added OGL, we'll likely need async texture downloads on it.

ameerj:
Scale logic is handled within image UploadMemory/DownloadMemory

FernandoS27 (approved):

PR #9933, opened 2023-03-10
video_core: Update texture format

vonchenplus:
Updates the texture format based on nVidia's open docs.


Kelebek1:
If you're going to use underscores, you should probably use them everywhere and make all formats the same. Right now it has all weird inconsistent styling. Put a note that the struct is in MSB -> LSB format, I think that would help clean up confusion as well.

It would be nice if all PixelFormat formats could be reversed into MSB -> LSB as well. This whole flipping formats back and forth is really confusing.

vonchenplus:
MSB LSB is back to a troublesome thing, the follow-up we slowly amended. underline I can unify a little.

liamwhite (approved):

PR #9951, opened 2023-03-13
configure_audio: Fix output mode setting not saving

Morph1984:


liamwhite (approved):

german77:
Now that I search for usages this setting is also unused. So probably is best to delete it to avoid having confusion in users

bunnei (approved):

PR #9953, opened 2023-03-15
service: nfp: Actually write correct crc

german77:
Slowly getting closer. Previously I kind of stubbed the functions now I actually feed the correct data to generate a valid crc.


liamwhite (approved):

src/core/hle/service/nfp/nfp_device.cpp
```diff
@@ -838,6 +846,31 @@ void NfpDevice::UpdateSettingsCrc() {
...
+        .mii_extension = tag_data.mii_extension,
+        .unknown2 = tag_data.unknown2,
+    };
+
+    std::array<u8, sizeof(CrcData)> data{};
+    memcpy(data.data(), &crc_data, sizeof(CrcData));
+    tag_data.register_info_crc = CalculateCrc(data);
+}
+
 u32 NfpDevice::CalculateCrc(std::span<const u8> data) {
```
Wunkolo, line 874: `boost-crc` is available I believe. So `boost::crc_32_type` can be utilized, such as here for example:

https://github.com/yuzu-emu/yuzu/blob/6d76a54d3768f45bb6aa7a1a4b4999a0b9b77e09/src/input_common/helpers/udp_protocol.cpp#L64C31-L72

bunnei (approved):

PR #9973, opened 2023-03-19
Implement asynchronous presentation

GPUCode:
While #9931 fixed a lot of scheduler stalls that occured in DispatchWork, the backend's ability to generate frames in async is still limited due to the fact that WaitWorker must be called at the end of each frame, which may be expensive depending on what the scheduler is [doing](https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp#L496). Thus the main goal of this PR is to offload swapchain operations to a separate thread to avoid such a wait and allow the backend to have multiple frames in flight.

The present manager class handles all swapchain calls. It maintains a separate thread that acquires and presents any frames that are pushed to the present queue. The GPU thread will send render frames and wait for free ones, to prevent synchronization errors. Plain semaphores and fences are used since I wasn't able to make it work with timeline semaphores.

I've tested this on both NVIDIA and AMD, window resizing seems to work reliably in addition to screenshots. However due to the nature of the changes some things might break so more testing is needed.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_present_manager.cpp
+++ b/src/video_core/renderer_vulkan/vk_present_manager.cpp
@@ -1,0 +1,440 @@
+// SPDX-FileCopyrightText: Copyright 2023 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "common/microprofile.h"
+#include "common/thread.h"
+#include "video_core/renderer_vulkan/vk_present_manager.h"
+#include "video_core/renderer_vulkan/vk_scheduler.h"
+#include "video_core/renderer_vulkan/vk_swapchain.h"
+#include "video_core/vulkan_common/vulkan_device.h"
+
+namespace Vulkan {
+
+MICROPROFILE_DEFINE(Vulkan_WaitPresent, "Vulkan", "Wait For Present", MP_RGB(128, 128, 128));
+MICROPROFILE_DEFINE(Vulkan_CopyToSwapchain, "Vulkan", "Copy to swapchain", MP_RGB(192, 255, 192));
+
+namespace {
+
+bool CanBlitToSwapchain(const vk::PhysicalDevice& physical_device, VkFormat format) {
+    const VkFormatProperties props{physical_device.GetFormatProperties(format)};
+    return (props.optimalTilingFeatures & VK_FORMAT_FEATURE_BLIT_DST_BIT);
+}
+
+[[nodiscard]] VkImageSubresourceLayers MakeImageSubresourceLayers() {
+    return VkImageSubresourceLayers{
+        .aspectMask = VK_IMAGE_ASPECT_COLOR_BIT,
+        .mipLevel = 0,
+        .baseArrayLayer = 0,
+        .layerCount = 1,
+    };
+}
+
+[[nodiscard]] VkImageBlit MakeImageBlit(s32 frame_width, s32 frame_height, s32 swapchain_width,
+                                        s32 swapchain_height) {
+    return VkImageBlit{
+        .srcSubresource = MakeImageSubresourceLayers(),
+        .srcOffsets =
+            {
+                {
+                    .x = 0,
+                    .y = 0,
+                    .z = 0,
+                },
+                {
+                    .x = frame_width,
+                    .y = frame_height,
+                    .z = 1,
+                },
+            },
+        .dstSubresource = MakeImageSubresourceLayers(),
+        .dstOffsets =
+            {
+                {
+                    .x = 0,
+                    .y = 0,
+                    .z = 0,
+                },
+                {
+                    .x = swapchain_width,
+                    .y = swapchain_height,
+                    .z = 1,
+                },
+            },
+    };
+}
+
+[[nodiscard]] VkImageCopy MakeImageCopy(u32 frame_width, u32 frame_height, u32 swapchain_width,
+                                        u32 swapchain_height) {
+    return VkImageCopy{
+        .srcSubresource = MakeImageSubresourceLayers(),
+        .srcOffset =
+            {
+                .x = 0,
+                .y = 0,
+                .z = 0,
+            },
+        .dstSubresource = MakeImageSubresourceLayers(),
+        .dstOffset =
+            {
+                .x = 0,
+                .y = 0,
+                .z = 0,
+            },
+        .extent =
+            {
+                .width = std::min(frame_width, swapchain_width),
+                .height = std::min(frame_height, swapchain_height),
+                .depth = 1,
+            },
+    };
+}
+
+} // Anonymous namespace
+
+PresentManager::PresentManager(Core::Frontend::EmuWindow& render_window_, const Device& device_,
+                               MemoryAllocator& memory_allocator_, Scheduler& scheduler_,
+                               Swapchain& swapchain_)
+    : render_window{render_window_}, device{device_},
+      memory_allocator{memory_allocator_}, scheduler{scheduler_}, swapchain{swapchain_},
+      blit_supported{CanBlitToSwapchain(device.GetPhysical(), swapchain.GetImageViewFormat())},
+      image_count{swapchain.GetImageCount()} {
+
+    auto& dld = device.GetLogical();
+    cmdpool = dld.CreateCommandPool({
+        .sType = VK_STRUCTURE_TYPE_COMMAND_POOL_CREATE_INFO,
+        .pNext = nullptr,
+        .flags =
+            VK_COMMAND_POOL_CREATE_TRANSIENT_BIT | VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT,
+        .queueFamilyIndex = device.GetGraphicsFamily(),
+    });
+    auto cmdbuffers = cmdpool.Allocate(image_count);
+
+    frames.resize(image_count);
+    for (u32 i = 0; i < frames.size(); i++) {
+        Frame& frame = frames[i];
+        frame.cmdbuf = vk::CommandBuffer{cmdbuffers[i], device.GetDispatchLoader()};
+        frame.render_ready = dld.CreateSemaphore({
+            .sType = VK_STRUCTURE_TYPE_SEMAPHORE_CREATE_INFO,
+            .pNext = nullptr,
+            .flags = 0,
+        });
+        frame.present_done = dld.CreateFence({
+            .sType = VK_STRUCTURE_TYPE_FENCE_CREATE_INFO,
+            .pNext = nullptr,
+            .flags = VK_FENCE_CREATE_SIGNALED_BIT,
+        });
+        free_queue.push(&frame);
+    }
+
+    present_thread = std::jthread([this](std::stop_token token) { PresentThread(token); });
+}
+
+PresentManager::~PresentManager() = default;
+
+Frame* PresentManager::GetRenderFrame() {
+    MICROPROFILE_SCOPE(Vulkan_WaitPresent);
+
+    // Wait for free presentation frames
+    std::unique_lock lock{free_mutex};
+    free_cv.wait(lock, [this] { return !free_queue.empty(); });
+
+    // Take the frame from the queue
+    Frame* frame = free_queue.front();
+    free_queue.pop();
+
+    // Wait for the presentation to be finished so all frame resources are free
+    frame->present_done.Wait();
+    frame->present_done.Reset();
+
+    return frame;
+}
+
+void PresentManager::PushFrame(Frame* frame) {
+    std::unique_lock lock{queue_mutex};
+    present_queue.push(frame);
+    frame_cv.notify_one();
+}
+
+void PresentManager::RecreateFrame(Frame* frame, u32 width, u32 height, bool is_srgb,
+                                   VkFormat image_view_format, VkRenderPass rd) {
+    auto& dld = device.GetLogical();
+
+    frame->width = width;
+    frame->height = height;
+    frame->is_srgb = is_srgb;
+
+    frame->image = dld.CreateImage({
+        .sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO,
+        .pNext = nullptr,
+        .flags = VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT,
+        .imageType = VK_IMAGE_TYPE_2D,
+        .format = swapchain.GetImageFormat(),
+        .extent =
+            {
+                .width = width,
+                .height = height,
+                .depth = 1,
+            },
+        .mipLevels = 1,
+        .arrayLayers = 1,
+        .samples = VK_SAMPLE_COUNT_1_BIT,
+        .tiling = VK_IMAGE_TILING_OPTIMAL,
+        .usage = VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT,
+        .sharingMode = VK_SHARING_MODE_EXCLUSIVE,
+        .queueFamilyIndexCount = 0,
+        .pQueueFamilyIndices = nullptr,
+        .initialLayout = VK_IMAGE_LAYOUT_UNDEFINED,
+    });
+
+    frame->image_commit = memory_allocator.Commit(frame->image, MemoryUsage::DeviceLocal);
+
+    frame->image_view = dld.CreateImageView({
+        .sType = VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO,
+        .pNext = nullptr,
+        .flags = 0,
+        .image = *frame->image,
+        .viewType = VK_IMAGE_VIEW_TYPE_2D,
+        .format = image_view_format,
+        .components =
+            {
+                .r = VK_COMPONENT_SWIZZLE_IDENTITY,
+                .g = VK_COMPONENT_SWIZZLE_IDENTITY,
+                .b = VK_COMPONENT_SWIZZLE_IDENTITY,
+                .a = VK_COMPONENT_SWIZZLE_IDENTITY,
+            },
+        .subresourceRange =
+            {
+                .aspectMask = VK_IMAGE_ASPECT_COLOR_BIT,
+                .baseMipLevel = 0,
+                .levelCount = 1,
+                .baseArrayLayer = 0,
+                .layerCount = 1,
+            },
+    });
+
+    const VkImageView image_view{*frame->image_view};
+    frame->framebuffer = dld.CreateFramebuffer({
+        .sType = VK_STRUCTURE_TYPE_FRAMEBUFFER_CREATE_INFO,
+        .pNext = nullptr,
+        .flags = 0,
+        .renderPass = rd,
+        .attachmentCount = 1,
+        .pAttachments = &image_view,
+        .width = width,
+        .height = height,
+        .layers = 1,
+    });
+}
+
+void PresentManager::WaitPresent() {
+    // Wait for the present queue to be empty
+    {
+        std::unique_lock queue_lock{queue_mutex};
+        frame_cv.wait(queue_lock, [this] { return present_queue.empty(); });
+    }
+
+    // The above condition will be satisfied when the last frame is taken from the queue.
+    // To ensure that frame has been presented as well take hold of the swapchain
+    // mutex.
+    std::scoped_lock swapchain_lock{swapchain_mutex};
+}
+
+void PresentManager::PresentThread(std::stop_token token) {
+    Common::SetCurrentThreadName("VulkanPresent");
+    while (!token.stop_requested()) {
+        std::unique_lock lock{queue_mutex};
+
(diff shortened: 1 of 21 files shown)


v1993:
Wouldn't this worsen frame pacing in games that support variable FPS and input (well, technically, presentation) latency overall?

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,335 @@
...
+    present_thread = std::jthread([this](std::stop_token token) { PresentThread(token); });
+}
+
+PresentManager::~PresentManager() = default;
+
+void PresentManager::PushFrame(Frame&& frame) {
+    std::unique_lock lock{queue_mutex};
+    if (frame_queue.size() >= image_count) {
+        frame_cv.wait(lock, [this] { return frame_queue.size() < image_count; });
+    }
```
liamwhite, line 108:
```suggestion
    frame_cv.wait(lock, [this] { return frame_queue.size() < image_count; });
```

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,335 @@
...
+    frame_cv.notify_one();
+}
+
+void PresentManager::WaitPresent() {
+    {
+        // Wait for the frame queue to be empty
+        std::unique_lock queue_lock{queue_mutex};
+        if (!frame_queue.empty()) {
+            frame_cv.wait(queue_lock, [this] { return frame_queue.empty(); });
+        }
```
liamwhite, line 120:
```suggestion
        frame_cv.wait(queue_lock, [this] { return frame_queue.empty(); });
```

GPUCode:
Yes that's a possibility but it shouldn't be too much (hopefully). If it becomes noticeable an option to make presentation synchronous again could be added

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,335 @@
...
+}
+
+Frame PresentManager::PopWait(std::stop_token token) {
+    std::unique_lock lock{queue_mutex};
+    if (frame_queue.empty()) {
+        Common::CondvarWait(frame_cv, lock, token, [this] { return !frame_queue.empty(); });
+        if (token.stop_requested()) {
+            return {};
+        }
+    }
```
liamwhite, line 136:
```suggestion
    Common::CondvarWait(frame_cv, lock, token, [this] { return !frame_queue.empty(); });
    if (token.stop_requested()) {
        return {};
    }
```

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+void PresentManager::PresentThread(std::stop_token token) {
+    Common::SetCurrentThreadName("VulkanPresent");
+    do {
+        Frame frame = PopWait(token);
+        if (token.stop_requested()) {
+            continue;
+        }
+        CopyToSwapchain(frame);
+    } while (!token.stop_requested());
```
liamwhite, line 162:
```suggestion
    while (!token.stop_requested()) {
        Frame frame = PopWait(token);
        if (token.stop_requested()) {
            return;
        }
        CopyToSwapchain(frame);
    }
```

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,335 @@
...
+        .queueFamilyIndex = device.GetGraphicsFamily(),
+    });
+    cmdbuffers = cmdpool.Allocate(image_count);
+}
+
+void PresentManager::PresentThread(std::stop_token token) {
+    Common::SetCurrentThreadName("VulkanPresent");
+    do {
+        Frame frame = PopWait(token);
+        if (token.stop_requested()) {
```
liamwhite, line 158: You might want to change the signature of PopWait to avoid this, like `bool PresentManager::PopWait(Frame& out_frame, std::stop_token token)` (and return false if a frame was not popped)

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,335 @@
...
+        std::unique_lock queue_lock{queue_mutex};
+        if (!frame_queue.empty()) {
+            frame_cv.wait(queue_lock, [this] { return frame_queue.empty(); });
+        }
+    }
+
+    // The above condition will be satisfied when the last frame is taken from the queue.
+    // To ensure that frame has been presented as well take hold of the swapchain
+    // mutex.
+    std::scoped_lock swapchain_lock{swapchain_mutex};
```
liamwhite, line 126: Because PopWait is not called with swapchain_mutex locked, this doesn't make the guarantee that you want here:
- the queue may be emptied by PopWait
- this lock call happens-before the lock call in CopyToSwapchain

You very likely need to implement the lock exchange that I added in the scheduler to make this work properly.

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+void PresentManager::WaitPresent() {
+    // Wait for the frame queue to be empty
+    std::unique_lock queue_lock{queue_mutex};
+    frame_cv.wait(queue_lock, [this] { return frame_queue.empty(); });
+
+    // The above condition will be satisfied when the last frame is taken from the queue.
+    // To ensure that frame has been presented as well take hold of the swapchain
+    // mutex.
+    std::exchange(queue_lock, std::unique_lock{swapchain_mutex});
```
liamwhite, line 120: This still has atomicity issues. Suppose this ordering occurs:
- PopWait empties the queue and releases the lock
- WaitPresent acquires the swapchain mutex and releases the queue mutex
- WaitPresent releases the swapchain mutex
- CopyToSwapchain acquires the swapchain mutex

GPUCode: Hm would taking the swapchain mutex right before PopFrame and then manually unlocking the queue lock here work then?

liamwhite: I tried [that solution](https://github.com/yuzu-emu/yuzu/blob/7878aba00c2a8dfaed78d70bf6f377f020f3a2f6/src/video_core/renderer_vulkan/vk_scheduler.cpp#L164) with the scheduler as well and it [doesn't work](https://github.com/yuzu-emu/yuzu/pull/9931#issuecomment-1464854609), because of signaling order requirements (you end up with lost wakeups).

GPUCode: Ok I changed it up a bit to match your scheduler implementation, should be better now

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+    frame_queue.push(frame);
+    frame_cv.notify_one();
+}
+
+void PresentManager::WaitPresent() {
+    // Wait for the frame queue to be empty
+    std::unique_lock queue_lock{queue_mutex};
+    frame_cv.wait(queue_lock, [this] { return frame_queue.empty(); });
+    queue_lock.unlock();
```
liamwhite, line 116:
```suggestion
    {
        std::unique_lock queue_lock{queue_mutex};
        frame_cv.wait(queue_lock, [this] { return frame_queue.empty(); });
    }
```

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+    // The above condition will be satisfied when the last frame is taken from the queue.
+    // To ensure that frame has been presented as well take hold of the swapchain
+    // mutex.
+    std::unique_lock swapchain_lock{swapchain_mutex};
```
liamwhite, line 121:
```suggestion
    std::scoped_lock swapchain_lock{swapchain_mutex};
```

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,329 @@
...
+            VK_COMMAND_POOL_CREATE_TRANSIENT_BIT | VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT,
+        .queueFamilyIndex = device.GetGraphicsFamily(),
+    });
+    cmdbuffers = cmdpool.Allocate(image_count);
+}
+
+void PresentManager::PresentThread(std::stop_token token) {
+    Common::SetCurrentThreadName("VulkanPresent");
+    while (!token.stop_requested()) {
+        std::unique_lock queue_lock{queue_mutex};
```
liamwhite, line 138:
```suggestion
        std::unique_lock lock{queue_mutex};
```

liamwhite (approved):

behunin:
Games render darker with this change and stopping emulation doesn't work.
Screenshot_20230319_113639

gidoly:
performance seems left overall same but i noticed issue what nehunin mentioned and some games would crash after unlock fps.
you can reproduce crash on super mario odyssey with posted save. load save > and unlock fps.
+ hyrule warriors age of calamity seems to crash while loading  map. issue is reproducable on 1st story map

AliciaBurrito:
To help tease out potential performance I decided to primarily test this on my Ayaneo 2 handheld, which has a Ryzen 6800U. 

Unfortunately I've only been met with device loss errors almost instantly after launching every game. Seems like AMD may be more sensitive to this.

GPUCode:
The darker colors should be fixed by the latest commit, might also address device losses

behunin:
Screenshot_20230320_115313

Looks better in this shot but I still can't get emulation to actually stop.
Nsight wont let me capture a frame with this change either. It stays stuck @ 30%

src/video_core/renderer_vulkan/vk_present_manager.h
```diff
@@ -0,0 +1,67 @@
...
+    void CopyToSwapchain(Frame frame);
+
+private:
+    Core::Frontend::EmuWindow& render_window;
+    const Device& device;
+    Scheduler& scheduler;
+    Swapchain& swapchain;
+    vk::CommandPool cmdpool;
+    vk::CommandBuffers cmdbuffers;
+    std::jthread present_thread;
```
liamwhite, line 58: This needs to come after the condition variables and mutexes to ensure correct destruction order, otherwise it will deadlock on shutdown with libstdc++ / glibc

liamwhite:
I can confirm that the screen output is much darker than intended in Breath of the Wild, even on the title screen (using radv 22.3.6, and amdvlk 2022.Q4.2). Validation layers don't report anything interesting in either case.

behunin:
Stopping emulation works now, still no luck with Nsight though.

GPUCode:
Hm could this be caused by post processing being broken? Can you disable FSR and SMAA just to check? Otherwise I suspect it's srgb related though I'm unable to figure out why.

behunin:
> Hm could this be caused by post processing being broken? Can you disable FSR and SMAA just to check? Otherwise I suspect it's srgb related though I'm unable to figure out why.

No difference

src/video_core/renderer_vulkan/vk_scheduler.h
```diff
@@ -106,6 +106,8 @@ class Scheduler {
         return *master_semaphore;
     }
 
+    std::mutex submit_mutex;
```
lioncash, line 109: Maybe have a function part of the interface like:

```cpp
// or some better name
std::scoped_lock LockForSubmit() {
  return std::scoped_lock{submit_mutex};
}
```

instead of directly exposing the mutex itself

lioncash: Ah, fair enough, I suppose

liamwhite:
In Breath of the Wild:
On yuzu master, the final render to the swapchain image is using a sRGB view of a unorm swapchain image
On this PR branch, the final render to the swapchain image is using an sRGB image

GPUCode:
I've updated the implementation to mimic the swapchain images more closely, thanks @liamwhite for the information.

liamwhite:
As of the latest commit, I can confirm that the dark rendering in Breath of the Wild is fixed.

AliciaBurrito:
Doing another passthrough on my Ayaneo 2 with the latest artifact.

Mainline tested here is 1377. A common issue I'm hitting is rapid crashing, unlocking framerate in many titles can cause yuzu to immediately crash, and other instabilities. Good news is that the color / darkness issues are fixed for me as well though!

**Fire Emblem: Engage**
- Main: 91-93 FPS (Map) | 46-47 FPS (Somniel)
- PR: 88-90 FPS (Map) | 47 FPS (Somniel)

I consistently and reproducibly get crashes when in the Somniel toggling fullscreen on the PR. Interestingly doesn't happen on the map screen.

**Astral Chain**
Device loss

**Bayoneta 3**
Crash at or after Title Screen, not reported as device loss. Crashes on shader_recompiler, even when emptying old shaders.

**Captain Toad's Treasure Tracker**
Crashes at start, last report in log is vk_pipeline_cache.

**Catherine: Full Body**
Crashes randomly upon loading up.

GPUCode:
@AliciaBurrito If you enable graphics debugging does it show anything in the log?

AliciaBurrito:
@GPUCode I redownloaded the latest artifact, and wiped my shaders folder just in case and grabbed some fresh logs with graphics debugging enabled.

Each of the following games crashed before I got to the title screen, so the logs are fairly short. Let me know if there's anything else you'd like me to test for this.

[astral-chain.txt](https://github.com/yuzu-emu/yuzu/files/11057787/astral-chain.txt)
[bayonetta-3.txt](https://github.com/yuzu-emu/yuzu/files/11057788/bayonetta-3.txt)
[fire-emblem-engage.txt](https://github.com/yuzu-emu/yuzu/files/11057789/fire-emblem-engage.txt)

AliciaBurrito:
@GPUCode First time using the Vulkan SDK's debugging features, so let me know if this isn't what you're looking for - but here is the log from both yuzu (with additional logging options enabled) and from Vulkan SDK with the synchronization validation enabled for Bayonetta 3.

If this is accurate I can get logs for more games this way.
[bayonetta-3.txt](https://github.com/yuzu-emu/yuzu/files/11065644/bayonetta-3.txt)
[vulkan-sdk-log.txt](https://github.com/yuzu-emu/yuzu/files/11065645/vulkan-sdk-log.txt)

GPUCode:
Thanks! I've identified several synchronization issues with the code, I'll push an update soon

gidoly:
Any update on this pr?

GPUCode:
Appologies for the wait got a little busy on the citra side, I've pushed a new commit that improves synchronization and also should improve performance by removing the host wait. Would be appreciated if you could retest some of your games.

src/video_core/renderer_vulkan/vk_present_manager.cpp
```diff
@@ -0,0 +1,440 @@
...
+    auto cmdbuffers = cmdpool.Allocate(image_count);
+
+    frames.resize(image_count);
+    for (u32 i = 0; i < frames.size(); i++) {
+        Frame& frame = frames[i];
+        frame.cmdbuf = vk::CommandBuffer{cmdbuffers[i], device.GetDispatchLoader()};
+        frame.render_ready = dld.CreateSemaphore({
+            .sType = VK_STRUCTURE_TYPE_SEMAPHORE_TYPE_CREATE_INFO_KHR,
+            .pNext = nullptr,
+        });
```
behunin, line 119: Missing flag

behunin:
Detail is lost now.
Screenshot_20230401_103557

GPUCode:
Send a log file with graphics debugging enabled

behunin:
It is definitely post processing, cycling through them clears things up.
Funny part is after doing that FSR + SMAA looks good again

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/11130106/yuzu_log.txt)

gidoly:
i have measured pretty good boost on games with resolution scaling vs mainline. only issue i noticed is hyrule warriors age of calamity could crash. 

log:

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/11130704/yuzu_log.txt)

GPUCode:
@gidoly that log file contains a lot of errors, none of which seem related to this PR though. I'm surprised the game doesn't crash already with all these, maybe async present makes it worse?

AliciaBurrito:
@GPUCode Unfortunately still running into instant-crashing for all titles that I've tested. I've compiled yuzu and Vulkan SDK logs for 4 of said games, I hope these help and let me know if there's anything else I can test to help narrow down the crashes 💝 
[vulkan-astral-chain.txt](https://github.com/yuzu-emu/yuzu/files/11132950/vulkan-astral-chain.txt)
[vulkan-breath-of-the-wild.txt](https://github.com/yuzu-emu/yuzu/files/11132951/vulkan-breath-of-the-wild.txt)
[vulkan-fire-emblem-engage.txt](https://github.com/yuzu-emu/yuzu/files/11132952/vulkan-fire-emblem-engage.txt)
[vulkan-mario-kart-8-deluxe.txt](https://github.com/yuzu-emu/yuzu/files/11132953/vulkan-mario-kart-8-deluxe.txt)
[yuzu-astral-chain.txt](https://github.com/yuzu-emu/yuzu/files/11132954/yuzu-astral-chain.txt)
[yuzu-breath-of-the-wild.txt](https://github.com/yuzu-emu/yuzu/files/11132955/yuzu-breath-of-the-wild.txt)
[yuzu-fire-emblem-engage.txt](https://github.com/yuzu-emu/yuzu/files/11132956/yuzu-fire-emblem-engage.txt)
[yuzu-mario-kart-8-deluxe.txt](https://github.com/yuzu-emu/yuzu/files/11132957/yuzu-mario-kart-8-deluxe.txt)

GPUCode:
@AliciaBurrito Unforunately all of the logs seem to cut off when the crash occur. Could you try running yuzu with gdb so we can get a stack trace?

AliciaBurrito:
@GPUCode Setting all of this up on a gaming handheld has been hilarious, but here are some juicy stack traces! Some of these look to be the same thing, but I'm uploading anyhow.

That said I was able to get in-game briefly with Bayonetta 3 before crashing. I hadn't gotten that far prior with the artifact builds. I was able to seemingly play Mario Kart 8 Deluxe fine without issue again unlike prior.

**Bayonetta 3**
image
**Astral Chain**
image
**Fire Emblem: Engage**
image
**Breath of the Wild**
image
**Yoshi's Crafted Island**
image

GPUCode:
Thanks for the stack traces! These are definitely useful

gidoly:
post my crash stack. (though seems alicia stack is exactly same)

devenv_kLuC1XoRGG

GPUCode:
Hm I find it strange that some crashes occur in the presentation manager while others appear completely unrelated to presentation. I will look into it shortly though, thanks

gidoly:
i must heads up that this pr provide decent boost on games

early access:

yuzu_tH3lVKs92u

pr:

async_present_EL6VKiUb9p

GPUCode:
I think the problem is that the present thread calls Finish on the scheduler, which doesn't appear to be thread safe. This would explain the random crashes when drawing. Will push a new commit later today

gidoly:
seems most crashes are fixed now, here remaining 2 crash stack trace.

devenv_7nq1xv5k9H

devenv_RO7ebZgY3f

GPUCode:
I've pushed a potential fix to the descriptor crashes @AliciaBurrito @behunin @gidoly can you retest please?

behunin:
The Nsight issue still persists.
The detail loss is the same as well, goes away once the filter has been changed.
Looks fine when I cycle back to fsr.

AliciaBurrito:
@GPUCode Stability has been rock solid with the latest, including performance from what I can tell! I tested every game in my library. Very exciting :)

gidoly:
seems fixes all crash i have reported. nice job!

AliciaBurrito:
@GPUCode We've gotten reports of FSR broken with this implementation, I've created an issue here: https://github.com/yuzu-emu/yuzu/issues/10037

GPUCode:
FSR should work now

behunin:
FSR issue is resolved on my end.

AliciaBurrito:
Also resolved on my end on both of my devices! Waiting to hear back from someone who mentioned it wasn't working for them even after cycling filters to see if it's fixed for them as well.

solarmystic:
Seeing some very noticeable frametime judder in Astral Chain that was not present prior to this PR in the latest mainline (1401) after this PR got mainline merged.

Compare and contrast the frame time consistency on mainline 1400 (before) and 1401 (after)

bunnei:
Untagging to give other PRs test time

gidoly:
Seems some report suggest that it decrease fps on xc3 with steam deck.maybe we should make toggle fpr this

liamwhite (approved):

bunnei (approved):

PR #9988, opened 2023-03-25
Pass GPU page table by reference inside TextureCache::ForEachImageInRegionGPU

rschlaikjer:
Investigating issues with game playback for Tsukihime -A Piece of Blue Glass Moon- (01001DC01486A000), a `perf` trace seemed to indicate that a significant chunk of time during stutters was being eaten up by hash table node instantiation inside the texture cache:

2023-03-23-232530_2610x800_scrot

On further digging, the issue appears to be a simple missing reference when capturing the gpu page table for a mapper lambda fed into ForEachGPUPage, causing the entire table to be duplicated a few hundred/thousand times per second in certain scenes. Converting the capture to a reference basically eliminates stuttering from this game, and may have an impact on others.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -1616,37 +1616,38 @@ void TextureCache<P>::ForEachImageInRegionGPU(size_t as_id, GPUVAddr gpu_addr, s
         return;
     }
     auto& gpu_page_table = gpu_page_table_storage[*storage_id];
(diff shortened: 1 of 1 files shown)


Kelebek1:
Good catch. I think we have a few problems like this around, I've seen we pass some vectors by copy as well for instance.

Kelebek1 (approved):

goldenx86:

rschlaikjer:
Yep, feel free, thanks.

PR #9990, opened 2023-03-25
Revert #9983

Morph1984:
vcpkg is annoying.


Morph1984:
Closing this in favor of #9989. We are pinning to boost 1.79 for the time being until the regression (whether it is our's or boost's fault) can be isolated.

Issue #9068, opened 2022-10-14
Shin Megami Tensei V - Texture corruption on MC/World elements after heading into World of Shadows since mainline 1191


solarmystic:
## Summary:-

The main character's textures get bugged out after heading into the World of Shadows since **yfc part 1 got merged in mainline 1191**. In addition, abnormally large and sparkly elements appear in the world after exiting the World of Shadows.

Mainline 1190 is the last build not to have the issue.

SMTV is well known to have rendering issues on Vulkan when paired with AMD cards in the past, but this is the first time i've seen it on NVIDIA cards too.

## Reproduction steps:-

1) Remove pipeline shader caches before booting game for consistent reproduction accuracy.
2) Head into World of Shadows from a save point.
3) Select Apotheosis, then Learn Miracles.
4) The MC will display the wrong colored/stripy textures on his skin, as you repeatedly swap back and forth between Sophia and him. 
5) In addition, when you head out back to the game world, there will be abnormally big sparkly elements accompanying your character.

## Screenshots of the issue:-
Discolored/Corrupted textures on MC in mainline 1191:-
![1191mc](https://user-images.githubusercontent.com/4345150/195798553-86ef0708-b895-4f6f-9f45-0dcbf8156610.PNG)

Abnormally large and sparkly elements after exiting World of Shadows:-
![1191sparkly](https://user-images.githubusercontent.com/4345150/195798567-858202a2-1765-4e8f-8d6d-d96bfd8bb8d8.PNG)

Correct textures on the MC in mainline 1190:-
![1190n](https://user-images.githubusercontent.com/4345150/195800434-da6f5ee9-206e-4cc0-843d-d1bcce0db431.PNG)

## Videos of the issue:-

Mainline 1191 (bugged)

https://user-images.githubusercontent.com/4345150/195796875-71dd202e-fb6b-427b-b0aa-0ed9a223a5b0.mp4

Mainline 1190 (working)

https://user-images.githubusercontent.com/4345150/195796855-f9037ea0-e39f-4605-bb04-78dca8c9ed08.mp4

## Logfiles:-

Mainline 1191 (bugged textures)
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/9783946/yuzu_log.txt)

Mainline 1190 (working textures)
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/9783876/yuzu_log.txt)

## System Specifications:-

Intel i5-12400F
[...]

PR #10004, opened 2023-03-28
[texture_cache] Only upload GPU-modified overlaps

### LINKED ISSUE #5255: new texture cache xenoblade definetive edition graphic issue
with new texture cache when first loading this area randomly get this kind of weirdness (water lighting is weird)
(if it not happens it should reproduce with restart game)

area name: fallen arm

how it supposed to look like

area name: colony 9

how it supposed to looks like

log with debug filter enabled:
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/5753998/yuzu_log.txt)

yuzu version: early access 1264 

hardware:
i7 8700k
rtx 2080 super

Kelebek1:
Currently when a new image is inserted into the cache, read it from the guest, upload it, and then we overwrite it with every overlapping image found, regardless of its current state. The data we just read and uploaded from the guest is the most up-to-date available, but the overlaps that already exist in the cache can be varying levels of stale-ness, we don't really know when that data was last modified, and will almost certainly be out-dated vs what we just read. For Xenoblade's cubemaps in particular, we were overwriting new guest data, with older non-modified host data. 

Instead, only copy overlaps which have been GPU-modified. This mirrors what the buffer cache already does on resolving overlaps.

Closes #5255.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -1176,13 +1176,13 @@ ImageId TextureCache<P>::JoinImages(const ImageInfo& info, GPUVAddr gpu_addr, VA
     const size_t size_bytes = CalculateGuestSizeInBytes(new_info);
     const bool broken_views = runtime.HasBrokenTextureViewFormats();
     const bool native_bgr = runtime.HasNativeBgr();
-    std::vector<ImageId> overlap_ids;
+    boost::container::small_vector<ImageId, 4> overlap_ids;
     std::unordered_set<ImageId> overlaps_found;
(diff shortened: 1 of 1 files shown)


gidoly:
also fixes https://github.com/yuzu-emu/yuzu/issues/9068

solarmystic:
> also fixes #9068

I am still getting that issue using the pr build though, i don't think it's really fixed. At least on NVIDIA Turing cards anyway. 


https://user-images.githubusercontent.com/4345150/228753101-0718588f-a299-4ee8-97cf-b44c7a1bbe4d.mp4

liamwhite (approved):

PR #10009, opened 2023-03-29
[Vulkan] Update some barriers

Kelebek1:
Just gives another pass over our Vulkan barriers, adding/changing a few things that seemed missing or could help. Not sure if this fixes or helps any games in particular, and it would be nice to get another set of eyes to check my changes here.


liamwhite (dismissed):

gidoly:
it seems to har xenoblade 2 performance by ~8fps

Kelebek1:
> `srcAccessMask` make written memory available (flushing caches). Any READ_BITs do nothing there.

Doesn't it affect ordering of commands? Without a barrier isn't it possible for a write to begin before a previous read completes? I feel like I've hit that bug multiple times.

K0bin:
> Doesn't it affect ordering of commands? 

No, the `stageMasks` affect the ordering. The `accessMasks` just handle memory availability/visibility.

Issue #9957, opened 2023-03-16
The Legend of Zelda Breath of the Wild AMD graphics card shadow BUG


yemaowuxin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

yuzu Early Access 3459

### Description of Issue

![yuzu_0aWJPPIEOS](https://user-images.githubusercontent.com/48284264/225519115-fc2d2c51-e6c7-4a7e-abf7-1c0e727d80dc.jpg)
![yuzu_gTZdJxiH4Y](https://user-images.githubusercontent.com/48284264/225519164-87f1b4f3-1300-43c3-a73e-9e328907bb34.jpg)
![yuzu_jc0izQZGpG](https://user-images.githubusercontent.com/48284264/225519219-8a1b9094-512e-48ce-823f-cbefdf632a33.jpg)

OpenGL and Vulkan API.
If turn on anisotropic filtering, the shadow will have problems.
All AMD graphics cards have this problem.

### Expected Behavior

Hope to fix the shadow problem.

### Reproduction Steps

AMD graphics cards, turn on anisotropic filtering x16.

### Log File

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/10986817/yuzu_log.txt)

### System Configuration

CPU: i9-10900K
GPU/Driver: AMD RX 6900xt
RAM: 32GB
OS: Windows 11 22H2 (Build 22621.1413)

goldenx86:
Yep it's easy to confirm.
Sadly this is the reason we keep anisotropic in the advanced tab, not all games behave correctly with it, especially if there are driver/hardware differences.
Keep in mind all Switch games expect an NVIDIA behaviour, AMD seems to be doing something different here.

yemaowuxin:
> Yep it's easy to confirm. Sadly this is the reason we keep anisotropic in the advanced tab, not all games behave correctly with it, especially if there are driver/hardware differences. Keep in mind all Switch games expect an NVIDIA behaviour, AMD seems to be doing something different here.

Thank you very much for your answer.

yemaowuxin:
> Yep it's easy to confirm. Sadly this is the reason we keep anisotropic in the advanced tab, not all games behave correctly with it, especially if there are driver/hardware differences. Keep in mind all Switch games expect an NVIDIA behaviour, AMD seems to be doing something different here.

[...]

Issue #6956, opened 2021-09-03
[Iris Xe][Mesa][ANV][Vulkan] Breath of the Wild - Shadow issues


PacoAvelar:
Game : Breath of the Wild (base)
Yuzu Version : EA 2031
CPU : i7 1165G7
RAM : 16GB
GPU : Iris Xe
GPU Driver : Mesa 21.2.1
OS: Debian 11
Kernel : 5.13.13-xanmod1

Issue : Shadow issues
![Screenshot from 2021-09-02 22-03-21](https://user-images.githubusercontent.com/29181120/131944483-1ec6a880-5a5f-4a0b-8e28-99198308ecbd.png)

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7103060/yuzu_log.txt)

FernandoS27:
does this happen on the propietary driver as well.

PacoAvelar:
![image](https://user-images.githubusercontent.com/29181120/132042487-4ef19fb0-c210-426c-8c43-aa25ff9fc6a1.png)


not that one, but other issues

pacoa-kdbg:
Same issue still present on

Yuzu Version : 834
CPU : i7 1165G7
RAM : 16GB
GPU : Iris Xe
GPU Driver : Mesa 21.2.6
OS: Debian 11
Kernel : 5.15.6-xanmod1

FernandoS27:
hmn, how's ANV on OGL?

pacoa-kdbg:
crashes at launch  with this error on  console
../src/intel/isl/isl.c:2105: FINISHME: ../src/intel/isl/isl.c:isl_surf_supports_ccs: CCS for 3D textures is disabled, but a workaround is available.
libpng warning: iCCP: known incorrect sRGB profile
Segmentation fault


if I clear shaders I can get to the menu, but still a segmentation fault happens when loading the save 

QOpenGLContext::swapBuffers() called with non-exposed window, behavior is undefined
Segmentation fault

yuzu log doesnt show any error

FernandoS27:
you should report that to the driver developers though.

skullmonkey911:
Hi,

still having issues with Vulkan (Intel driver 30.0.101.1660) on yuzu 977 in Windows. Is this issue driver or yuzu releated?

Regards,

Stephan

FernandoS27:
[...]

PR #10030, opened 2023-04-04
shader_recompiler: Fix ImageGather rounding on AMD/Intel

Wollnashorn:
This should hopefully fix a long-standing bug affecting AMD (and Intel?) owners in Zelda: Breath of the Wild with artifacts (#9957, #6956) at object edges. These glitches were especially noticeable at the grass in the game. 

BotW uses deferred rendering and renders the object shadows at half the resolution of the framebuffer. Per fragment in the fullscreen triangle shadow mapping pass, four texels of the fullres depth texture are sampled simultaneously with a [textureGather](https://registry.khronos.org/OpenGL-Refpages/gl4/html/textureGather.xhtml) call. Unlike `texelFetch`, `textureGather` works with normalized floating-point texture coordinates.
As a result, the texture coordinates at each fragment are always exactly [on the boundary of the four gathered pixels](https://www.reedbeta.com/blog/texture-gathers-and-coordinate-precision/gather-dark.png) and different rounding precision can lead to different pixels being sampled. [According to a blogpost by Nathan Reed](https://www.reedbeta.com/blog/texture-gathers-and-coordinate-precision/), the actual pixel integer coordinates of the texture are calculated by the GPU by converting the floating-point coordinates to fixed-point with a precision of at least 16.8 bits. Nvidia seems to implement this conversion with a different rounding behaviour at such pixel boundaries than AMD and Intel.

## Solution
The problem can be solved by adding a tiny subpixel offset of 1/512 of the texel size to the texture coordinates in textureGather. To do this, I modified the SPIR-V and GLSL backend of Yuzu's shader recompiler. More precisely, in the SPIR-V/GLSL emitter, the function `EmitImageGather` issues additional shader commands to calculate the offset coordinates. Of course, this offset is not applied on Nvidia cards.

For now this is enabled on AMD and Intel by default. I added `need_gather_subpixel_offset` to the shader recompiler profile. It also might be considerable to make this fix optional with a advanced graphics option, but I don't know.

Note that this does not fix the single horizontal line artifacts when 16x anisotropic filtering is enabled, as this seems to be an unrelated (vendor agnostic?) issue with the game. I think these artifacts only occur at the harsh borders of the mipmap transitions (MIP banding). With 8x or less everything is fine though.

botw-gather-rounding-error
botw-fixed

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -169,6 +169,7 @@ Device::Device(Core::Frontend::EmuWindow& emu_window) {
     has_draw_texture = GLAD_GL_NV_draw_texture;
     warp_size_potentially_larger_than_guest = !is_nvidia && !is_intel;
     need_fastmath_off = is_nvidia;
+    need_gather_subpixel_offset = is_amd;
     can_report_memory = GLAD_GL_NVX_gpu_memory_info;
 
     // At the moment of writing this, only Nvidia's driver optimizes BufferSubData on exclusive
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -169,7 +169,6 @@ Device::Device(Core::Frontend::EmuWindow& emu_window) {
     has_draw_texture = GLAD_GL_NV_draw_texture;
     warp_size_potentially_larger_than_guest = !is_nvidia && !is_intel;
     need_fastmath_off = is_nvidia;
-    need_gather_subpixel_offset = is_amd;
     can_report_memory = GLAD_GL_NVX_gpu_memory_info;
 
     // At the moment of writing this, only Nvidia's driver optimizes BufferSubData on exclusive
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -329,7 +329,9 @@ PipelineCache::PipelineCache(RasterizerVulkan& rasterizer_, const Device& device
 
         .lower_left_origin_mode = false,
         .need_declared_frag_colors = false,
-        .need_gather_subpixel_offset = device.NeedsGatherSubpixelOffset(),
+        .need_gather_subpixel_offset = driver_id == VK_DRIVER_ID_AMD_PROPRIETARY ||
+                                       driver_id == VK_DRIVER_ID_INTEL_PROPRIETARY_WINDOWS ||
+                                       driver_id == VK_DRIVER_ID_INTEL_OPEN_SOURCE_MESA,
 
         .has_broken_spirv_clamp = driver_id == VK_DRIVER_ID_INTEL_PROPRIETARY_WINDOWS,
         .has_broken_spirv_position_input = driver_id == VK_DRIVER_ID_QUALCOMM_PROPRIETARY,
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -329,6 +329,7 @@ PipelineCache::PipelineCache(RasterizerVulkan& rasterizer_, const Device& device
 
         .lower_left_origin_mode = false,
         .need_declared_frag_colors = false,
+        .need_gather_subpixel_offset = device.NeedsGatherSubpixelOffset(),
 
         .has_broken_spirv_clamp = driver_id == VK_DRIVER_ID_INTEL_PROPRIETARY_WINDOWS,
         .has_broken_spirv_position_input = driver_id == VK_DRIVER_ID_QUALCOMM_PROPRIETARY,
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -330,6 +330,8 @@ PipelineCache::PipelineCache(RasterizerVulkan& rasterizer_, const Device& device
         .lower_left_origin_mode = false,
         .need_declared_frag_colors = false,
         .need_gather_subpixel_offset = driver_id == VK_DRIVER_ID_AMD_PROPRIETARY ||
+                                       driver_id == VK_DRIVER_ID_AMD_OPEN_SOURCE ||
+                                       driver_id == VK_DRIVER_ID_MESA_RADV ||
                                        driver_id == VK_DRIVER_ID_INTEL_PROPRIETARY_WINDOWS ||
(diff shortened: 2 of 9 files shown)


goldenx86:
Sudden fix for one of the most asked graphical issues, complete and precise description of the problem and the solution, image examples.
You get a free asado.

src/video_core/renderer_opengl/gl_device.cpp
```diff
@@ -169,6 +169,7 @@ Device::Device(Core::Frontend::EmuWindow& emu_window) {
     has_draw_texture = GLAD_GL_NV_draw_texture;
     warp_size_potentially_larger_than_guest = !is_nvidia && !is_intel;
     need_fastmath_off = is_nvidia;
+    need_gather_subpixel_offset = is_amd;
```
Morph1984, line 172: This should probably be applied for intel. @goldenx86 will be confirming it.

goldenx86:
Can't confirm, Intel's drivers are a piece of shit and crash on their SPIR-V compiler.

Morph1984:
Can confirm it is fixed on Intel:
Before
image
After
image

src/video_core/vulkan_common/vulkan_device.h
```diff
@@ -664,6 +668,7 @@ class Device {
     bool must_emulate_bgr565{};             ///< Emulates BGR565 by swizzling RGB565 format.
     bool dynamic_state3_blending{};         ///< Has all blending features of dynamic_state3.
     bool dynamic_state3_enables{};          ///< Has all enables features of dynamic_state3.
+    bool need_gather_subpixel_offset{};     ///< Needs offset at ImageGather for correct rounding.
```
Morph1984, line 671: This should be set in vk_pipeline_cache and gl_shader_cache respectively using driver ids.

Wollnashorn: Do you mean I should remove the `NeedsGatherSubpixelOffset()` helpers in the device classes and set the profile option directly in vk_pipeline_cache/gl_shader_cache?

Like
```
.need_gather_subpixel_offset = VK_DRIVER_ID_AMD_PROPRIETARY ||
                               VK_DRIVER_ID_INTEL_PROPRIETARY_WINDOWS ||
                               VK_DRIVER_ID_INTEL_OPEN_SOURCE_MESA,
```
instead of
```
.need_gather_subpixel_offset = device.NeedsGatherSubpixelOffset(),
```

Morph1984: Yea, like that

Wollnashorn:
## 16x Anisotropic filtering fix

I also found the cause and a fix of the black lines artifact that happens at 16x anisotropic filtering. [Vulkan doesn't have minification filter that do not use the mipmaps.](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSamplerCreateInfo.html#_description) Yuzu correctly implements equivalent filter with setting the max lod to 0.25 like Khronos suggests.
But in video_core/textures/texture.cpp anisotropic filtering is always turned off for textures with a `mipmap_filter` of `TextureMipmapFilter::None`. I don't know exactly why that leads to the artifacts in BotW, but enabling it also for such textures fixes the glitches and shouldn't do any harm in general (more likely the opposite).

aniso_bug
aniso_fix

I don't know if I should do a new PR for that. It's just a single line change 😄

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -329,6 +329,9 @@ PipelineCache::PipelineCache(RasterizerVulkan& rasterizer_, const Device& device
 
         .lower_left_origin_mode = false,
         .need_declared_frag_colors = false,
+        .need_gather_subpixel_offset = driver_id == VK_DRIVER_ID_AMD_PROPRIETARY ||
```
liamwhite, line 332: This should be needed for RADV and AMD_OPEN_SOURCE too, right?

yemaowuxin:
@Wollnashorn Thanks you fix ImageGather rounding on AMD/Intel.
Now the everything is normal when anisotropic filtering is set system default.
When the anisotropic filtering is set to X16 or lower, there are still black lines.
But if set the anisotropic filtering is set  system default at yuzu, and open amd software:Adrenalin edition add yuzu.exe to Gaming, and set the anisotropic filtering is set to X16 at amd software, everything is normal.
It seems that there is a problem with Yuzu's anisotropic filtering.
RadeonSoftware_8drHGXKNff

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
Wollnashorn: I don't know. Do they need this fix too?

@goldenx86 mentioned a couple times that it doesn't occur with the Linux drivers (e.g. here https://community.citra-emu.org/t/any-way-to-make-breath-of-the-wild-look-better/544137/4)

liamwhite: I definitely experience these graphical issues in Breath of the Wild on Linux drivers.

liamwhite (approved):

bunnei (approved):

bunnei:
Thank you for the great fix @Wollnashorn !
If you're interested in working a little closer with the development team, please consider joining our [Discord](https://discord.com/invite/u77vRWY) (unless you are already there, and I just missed it!). Feel free to send me a DM `bunnei#0166` and I'd be happy to grant you speech in our development channel.

yemaowuxin:
> @yemaowuxin The AMD driver setting to enforce anisotropic filtering is only applied to old DX9 applications, so it has no effect on Yuzu.
> 
> As much as I understand, the bug is caused by the shadowmap depth texture being sampled with anisotropic filtering when you turn on AF in Yuzu, which doesn't make much sense for a shadowmap. I guess the BotW devs just used the wrong sampler and didn't bother because it's fine running on the default settings on the Switch. I think the best solution would be a mod that sets `mipmap_filter = TextureMipmapFilter::None` and `max_anisotropy = 0` for that one sampler, but I have no idea about modding.
> 
> That being said, apparently the game does not even use anisotropic filtering anyways on the terrain textures, so there isn't really a visual improvement there by using AF.

I understand. Thank you very much for your answer!

Wollnashorn:
Thanks @bunnei! I will join the Discord server soon, looking forward to do more fixes and improvements in the future

goldenx86:
The interesting part of the AF issue is that RDNA3 (and only RDNA3) suffers from it even at default. Gotta wait for cheaper ones to come out I guess.

Zhanghailin1995:
When using the AMD 7900XT, TOTK also encountered similar issues. Can these improvements fix the issues encountered in TOTK?

Squall-Leonhart:
> @yemaowuxin The AMD driver setting to enforce anisotropic filtering is only applied to old DX9 applications, so it has no effect on Yuzu.

This is actually incorrect, the description is out of date and the AF setting applies to all supported API's.

PR #10056, opened 2023-04-15
core: audio: return result when audio_out initialize failed

vonchenplus:
IGS Classic Arcade Collection [0100BA001821A000] is now accessible to play.

### DIFF
--- a/src/core/hle/service/audio/audout_u.cpp
+++ b/src/core/hle/service/audio/audout_u.cpp
@@ -49,12 +49,6 @@ public:
         };
         // clang-format on
         RegisterHandlers(functions);
-
-        if (impl->GetSystem()
-                .Initialize(device_name, in_params, handle, applet_resource_user_id)
-                .IsError()) {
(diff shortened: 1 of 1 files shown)


src/core/hle/service/audio/audout_u.cpp
```diff
@@ -287,6 +281,13 @@ void AudOutU::OpenAudioOut(HLERequestContext& ctx) {
 
     auto audio_out = std::make_shared<IAudioOut>(system, *impl, new_session_id, device_name,
                                                  in_params, handle, applet_resource_user_id);
+    result = audio_out->GetImpl()->GetSystem().Initialize(device_name, in_params, handle,
+                                                          applet_resource_user_id);
+    if (result.IsError()) {
```
FearlessTobi, line 286: might be a good idea to still log here

FearlessTobi (approved):

FernandoS27 (approved):

PR #10059, opened 2023-04-15
Qt: RoundPreferFloor scaling mode for win HiDPI

stuken:
The HiDPI fixes introduced a while back have made a mess of scaling on windows with 4k displays and 150% scaling. This manifests as a huge increase of the window and font sizes and scaling artifacts on the audio and control screens.

Lets use the RoundPreferFloor enum to not force Qt to round up the scaling at 150% and regain the correct behaviour from before.

image
image

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -4585,7 +4585,7 @@ static void SetHighDPIAttributes() {
 
     if (max_ratio > real_ratio) {
         QApplication::setHighDpiScaleFactorRoundingPolicy(
-            Qt::HighDpiScaleFactorRoundingPolicy::Round);
+            Qt::HighDpiScaleFactorRoundingPolicy::RoundPreferFloor);
     } else {
         QApplication::setHighDpiScaleFactorRoundingPolicy(
             Qt::HighDpiScaleFactorRoundingPolicy::Floor);


Morph1984:
Integer scaling was used due to artifacts when using PassThrough scaling. I'm not sure what the scaling artifacts are that you mention with 150% rounding up to 200%.
However, the rationale behind 150% rounding to 200% is since 100% at 150% causes the UI to be a lot smaller.

AliciaBurrito:
@stuken Can you provide before/after pics from your own personal setup, and include more details in what issues you were seeing prior?

stuken:
PR build:image
Mainline:
image
Side by side: 
image
Both set at 720p windows:
image

Both are clean oob settings yet the mainline ui is scaled way larger in comparison to the pr and has more sliders being obscured. It looks to me that it isnt 200% based off the baseline but 200% of 150%. Prior to the HiDPI fixes there were no noticable issues on my end. The UI was perfectly useable with no obscuring of elements. Now setting a 1080p window goes far beyond the bounds of a 4k screen which is incorrect.

krisrok:
Can confirm this behaviour. Sadly, fiddling with Window's compatibility options does not help either.

Morph1984:
The issue with scaling is that it depends on both the resolution and the scaling factor of the display and user preference, no one solution works for every display. This makes it worse for people who prefer the round up behavior vs round down.
I would consider a solution that makes this user adjustable, but not one that changes the current behavior.

Morph1984:
This issue is an example of the round down behavior being undesirable:
https://github.com/yuzu-emu/yuzu/issues/10880

PR #10068, opened 2023-04-19
Allow passing `--bind-address` to dedicated room.

twitchax:
### Impetus

In some scenarios, it is important to allow for dedicated room servers to specify the bind address.  For example, `fly.io` requires that UDP services bind to a special address aliased by `fly-global-services`, as routing for UDP packets when bound to `0.0.0.0` does not function properly (has to do with how their global routing works).  In scenarios such as that, the dedicated server should allow for the user to set a custom bind address.

*These changes are _backward compatible_, as the execution continues normally without error, and the previous behavior is preserved, if the new argument is not set.*

### Description

* Allow for the user to pass `--bind-address` to the dedicated room executable.  Otherwise, default to the existing default, which appears to be `0.0.0.0`.
* Add an info statement to the log if an option is not passed, informing the log consumer of the default.

### DIFF
--- a/src/dedicated_room/yuzu_room.cpp
+++ b/src/dedicated_room/yuzu_room.cpp
@@ -49,6 +49,7 @@ static void PrintHelp(const char* argv0) {
              " [options] <filename>\n"
              "--room-name         The name of the room\n"
              "--room-description  The room description\n"
+             "--bind-address      The bind address for the room\n"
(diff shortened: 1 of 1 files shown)


FearlessTobi (approved):

bunnei (approved):

goldenx86:
Hi, can I mention this change in the next progress report?

PR #10074, opened 2023-04-21
Account for a pre-added offset when using Corner sample mode for 2D blits

Kelebek1:
Blit is being passed out-of-bounds ranges when using corner sampling, due to a pre-added offset. Pokemon Scarlet throws a lot of validation errors with invalid ranges, like (2, 1922) rather than the expected (0, 1920).


FernandoS27 (approved):

v1993:
Fixes DoF and a few other effects in 2.5D Kirby games.

PR #10075, opened 2023-04-21
Silence network spam

Kelebek1:
I thought we already silenced these before, but the spam from these is horrible in some games. I left the errors to print once if they're necessary to see.

### DIFF
--- a/src/core/internal_network/network.cpp
+++ b/src/core/internal_network/network.cpp
@@ -356,7 +356,7 @@ NetworkInstance::~NetworkInstance() {
 std::optional<IPv4Address> GetHostIPv4Address() {
     const auto network_interface = Network::GetSelectedNetworkInterface();
     if (!network_interface.has_value()) {
-        LOG_ERROR(Network, "GetSelectedNetworkInterface returned no interface");
+        LOG_DEBUG(Network, "GetSelectedNetworkInterface returned no interface");
         return {};
     }
 
--- a/src/core/hle/service/nifm/nifm.cpp
+++ b/src/core/hle/service/nifm/nifm.cpp
@@ -218,7 +218,7 @@ public:
 
 private:
     void Submit(HLERequestContext& ctx) {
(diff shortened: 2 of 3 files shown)


Morph1984:
Just make these LOG_DEBUG and no need for the extra stuff

FearlessTobi:
one of these about the network interface should stay as error, since it is useful for debugging ldn connection issues.

Kelebek1:
> Just make these LOG_DEBUG and no need for the extra stuff

I wanted to avoid the performance hit, since they're each called dozens of times per second.

Kelebek1:
Made them all debug instead as per Morph's comment.

FearlessTobi:
What I'm saying is, one of them should log *once* as error so it can be used to diagnose issues.
The other ones can all be LOG_DEBUG.

src/core/internal_network/network.cpp
```diff
@@ -356,7 +356,7 @@ NetworkInstance::~NetworkInstance() {
 std::optional<IPv4Address> GetHostIPv4Address() {
     const auto network_interface = Network::GetSelectedNetworkInterface();
     if (!network_interface.has_value()) {
-        LOG_ERROR(Network, "GetSelectedNetworkInterface returned no interface");
+        LOG_DEBUG(Network, "GetSelectedNetworkInterface returned no interface");
```
FearlessTobi, line 359: this one should still be logged as error once

MonsterDruide1:
As far as I understand the project, `(stubbed)` being logged means that a function has been called which is not properly implemented/emulated by Yuzu. Shouldn't we rather implement those functions instead of just silencing the error?

bunnei (approved):

src/core/internal_network/network.cpp
FearlessTobi: This comment has not been addressed. @Kelebek1 @bunnei

PR #10082, opened 2023-04-22
Refactor Accelerate DMA and do downloads through TC.

FernandoS27:
This fixes Pokemon Sword and Hyrule warriors AoC regressions.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_texture_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_texture_cache.cpp
@@ -1,10 +1,11 @@
-// SPDX-FileCopyrightText: Copyright 2019 yuzu Emulator Project
-// SPDX-License-Identifier: GPL-2.0-or-later
+// SPDX-FileCopyrightText: Copyright 2023 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-3.0-or-later
 
 #include <algorithm>
 #include <array>
 #include <span>
 #include <vector>
+#include <boost/container/small_vector.hpp>
 
 #include "common/bit_cast.h"
 #include "common/bit_util.h"
@@ -1341,16 +1342,20 @@ void Image::UploadMemory(const StagingBufferRef& map, std::span<const BufferImag
     UploadMemory(map.buffer, map.offset, copies);
 }
 
-void Image::DownloadMemory(VkBuffer buffer, VkDeviceSize offset,
+void Image::DownloadMemory(std::span<VkBuffer> buffers_span, VkDeviceSize offset,
                            std::span<const VideoCommon::BufferImageCopy> copies) {
     const bool is_rescaled = True(flags & ImageFlagBits::Rescaled);
     if (is_rescaled) {
         ScaleDown();
     }
+    boost::container::small_vector<VkBuffer, 1> buffers_vector{};
+    for (auto& buffer : buffers_span) {
+        buffers_vector.push_back(buffer);
+    }
     std::vector vk_copies = TransformBufferImageCopies(copies, offset, aspect_mask);
     scheduler->RequestOutsideRenderPassOperationContext();
-    scheduler->Record([buffer, image = *original_image, aspect_mask = aspect_mask,
-                       vk_copies](vk::CommandBuffer cmdbuf) {
+    scheduler->Record([buffers = std::move(buffers_vector), image = *original_image,
+                       aspect_mask = aspect_mask, vk_copies](vk::CommandBuffer cmdbuf) {
         const VkImageMemoryBarrier read_barrier{
             .sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER,
             .pNext = nullptr,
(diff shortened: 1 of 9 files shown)


src/video_core/texture_cache/texture_cache.h
```diff
@@ -661,27 +661,40 @@ template <class P>
...
-            image.DownloadMemory(download_map, copies);
-            download_map.offset += Common::AlignUp(image.unswizzled_size_bytes, 64);
+        if (any_none_dma) {
+            auto download_map = runtime.DownloadStagingBuffer(total_size_bytes, true);
+            for (const PendingDownload& download_info : download_ids) {
+                if (download_info.is_swizzle) {
+                    Image& image = slot_images[download_info.object_id];
+                    const auto copies = FullDownloadCopies(image.info);
+                    image.DownloadMemory(download_map, copies);
+                    download_map.offset += Common::AlignUp(image.unswizzled_size_bytes, 64);
```
Kelebek1, line 689: You calculate total_size_bytes with only the unswizzled size, but then when using the download_map you align it up to 64. If any alignup happens, won't you go out of bounds, since you'll have to use more bytes than total_size?

src/video_core/renderer_vulkan/vk_texture_cache.cpp
```diff
@@ -1,10 +1,11 @@
-// SPDX-FileCopyrightText: Copyright 2019 yuzu Emulator Project
-// SPDX-License-Identifier: GPL-2.0-or-later
+// SPDX-FileCopyrightText: Copyright 2023 yuzu Emulator Project
```
FearlessTobi, line 1: we don't adjust the copyright date when a file is changed.

FernandoS27: We should though.

FearlessTobi: Can you back that claim up? I don't find anything when googling around.
And we haven't done this in the past either

Moonlacer: I was under the impression that we update the date when a big enough change was pushed

Tachi107: From the point of view of copyright protection, updating copyright years is pretty useless (see [this article](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change) if you're interested). It may be useful to tell at a glance when the file was first created and when significant modifications were made to it though.

Substituting the year, as done here, is questionable; substituting the license is almost always unnecessary (and potentially inappropriate, as you'd technically have to ask all the contributors to that file to agree on the license change).

piplup55:
the ground is not solid on opengl

opengl:
image

vulkan:
image

src/video_core/renderer_vulkan/vk_texture_cache.cpp
FernandoS27: > From the point of view of copyright protection, updating copyright years is pretty useless (see [this article](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change) if you're interested). It may be useful to tell at a glance when the file was first created and when significant modifications were made to it though.
> 
> Substituting the year, as done here, is questionable; substituting the license is almost always unnecessary (and potentially inappropriate, as you'd technically have to ask all the contributors to that file to agree on the license change).

we agreed in upgrading to GPLv3 a long time ago and it was not even necessatry as the license before was GPLv2+ not just GPLv2. ThIt's not really changing the license. It's fixing it to reflect the current license.

Tachi107: > we agreed in upgrading to GPLv3 a long time ago and it was not even necessary as the license before was GPLv2+ not just GPLv2. It's not really changing the license. It's fixing it to reflect the current license.

I'm aware of that, but changing the license of existing files is not necessary, since, as you say, they were already GPLv2+, so you're allowed to distribute yuzu as a whole as GPLv3 to comply with the Apache-licensed portions of the NVflinger implementation, without having to ask contributors for nothing.

On the other hand, you cannot change the license of individual files from GPL-2.0-or-later to GPL-3.0-or-later without asking the contributors to the file if that's ok, since they have agreed to license the file under v2, and you're taking this away.

In case it's not clear, the license of individual source files is different from the license under which yuzu is distributed.

I know it doesn't matter much here, but still it's important to know how these things work just in case it happens to matter in the future.

FernandoS27: Thanks for pointing that out. I'll talk it out with the team as this is not very sound.

Do you have some legal backing to those claims. I fully understand that in the case of external code for which the organization is not a copyright holder.

Tachi107: > Do you have some legal backing to those claims.

Nope, I don't have legal backing (I'm not a lawyer etc etc), but this is what I've collected so far from working as a Debian maintainer and an open source contributor in general, since we take great care of license compliance and things like that.

> I fully understand that in the case of external code for which the organization is not a copyright holder.

Yeah, as I said in this specific case you _can_ do it since you're the sole copyright holder, but, for example, if I submit a patch that modifies significant parts of a bunch of files and those files are licensed under the GPL-2.0-or-later, you cannot change their license to GPL-3.0-or-later without asking me unless you remove my changes, since we are now both copyright holders of those portions of code.

This is my understanding of how things work - as with everything, I may be wrong, though.

bunnei (approved):

PR #10086, opened 2023-04-23
core_timing: Use CNTPCT as the guest CPU tick

Morph1984:
Previously, we were mixing the raw CPU frequency and CNTFRQ.
The raw CPU frequency (1020 MHz) should've never been used as CNTPCT (whose frequency is CNTFRQ) is the only counter available.

Additional changes to this PR include:
- Simplifying NativeClock
  - This impl significantly reduces the time taken to measure RDTSC by 55-60%.
  - Furthermore, this impl removes the changes in #9917, as the change below made it unnecessary. 
- ~~Reverting #8428~~ As per the comments below, somehow this causes issues again.
  - It turns out we were not acquiring a lock prior to signaling the variable.
- Using compile time arithmetic in TimeSpanType


liamwhite (dismissed):

Morph1984:
Update: I've removed the revert commit for the time being, and it will be investigated separately. This PR shouldn't be in EA forever

liamwhite (approved):

legend800:
Potential regression from this logged here:  https://github.com/yuzu-emu/yuzu/issues/10610

Morph1984:
@legend800 Please check if the regression has been fixed

legend800:
Well, it's better. It boots normally again, shows the initial splash screens but then permanent black screen at the cinematic. No way to bypass. 

On Master, cinematic loads (really slow) and you can skip it with A and then play the game perfectly. 

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/11684465/yuzu_log.txt)

bunnei (approved):

bunnei:
Merging based on Morph's comment from 2 weeks ago.

legend800:
Sure, but TMNT Shredder's Revenge will now be 100% unplayable on Master in addition to EA. Before the game was perfect on SC. https://github.com/yuzu-emu/yuzu/issues/10610 

While a more accurate implementation, ideally a fix is provided to whatever the incorrect implementation was masking (or prioritized as a fast follow) to achieve parity with previous game compatibility. It's probably easier to do the RCA/fix now vs. someone picking this up at a later time after more code refactoring.

PR #10089, opened 2023-04-23
Remove fast path stream buffer for Vulkan

### LINKED ISSUE #9688: Xenoblade 2 Cloud Flickering Issues

### Affected Build(s)

Mainline 1321

### Description of Issue

Flickering cloud effects, both on title screen and in game. Happens no matter what graphics settings I try (including both Vulkan and OpenGL). And FWIW also happens on RyujiNX.

Vulkan:
yuzu_5mJonnEBOC
yuzu_BHouqft2hN
Video: https://user-images.githubusercontent.com/70397032/215156123-d45adacc-dd12-4b1f-afd8-e3727656827d.mp4

OpenGL:
yuzu_uWOplyYBSP
yuzu_wI4vfWcOpv
Video: https://user-images.githubusercontent.com/70397032/215156806-c8f84a1c-a6b4-4c31-afcb-21014a071aa3.mp4

### Expected Behavior

The clouds to look like on console.  (not flickering)

### Reproduction Steps

Start a new game, the first fight vs the crab you can see the issue, as well as on the Chapter 1 title screen. 

### Log File

Vulkan [yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/10521541/yuzu_log.txt)

OpenGL [yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/10521559/yuzu_log.txt)

### System Configuration

CPU: AMD Ryzen 5 2600X Six-Core Processor 3.60 GHz
GPU/Driver: AMD Radeon RX 6700 XT (with latest driver)
RAM: 16GB @ 2666MHz
OS: Windows 10

Kelebek1:
I've looked at the code here and can't see any glaring issues with it, but there's some bug with it when the same piece of memory switches from fast to non-fast and visa versa. The performance loss from disabling this should be negligable. 

If anyone wants to investigate more, please do. Repro I'm using is the Crossroads of Destiny cutscene in XC2. Last cutscene for Chapter 1, just as Mets starts firing the cannons at Seiryuu right near the end of the scene. Some printing around this fast path code for the buffer with CpuAddr 0x2097471000 will show the issue.

In the bad frames, this buffer is used, and gets regularly copied to from the stream buffer during the cloud draws: 
image
Whereas in the good frames (when the clouds are drawn), all input uniform buffers use the Stream Buffer, with no copies:
image

Disabling the fast path fixes the cloud flickering in XC2, but worth testing the performance. Closes #9688

EDIT: This is exposing bugs in our buffer cache, so marking as a draft for now. Even though this causes visual issues, they're not caused by this PR, the code I'm removing is just hiding them. The stream buffer forcibly reading from guest memory regardless of cpu/gpu modification state, and skipping all cached writes etc, seems to be what's hiding it.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -778,11 +778,12 @@ void BufferCache<P>::BindHostGraphicsUniformBuffer(size_t stage, u32 index, u32 
     const u32 size = std::min(binding.size, (*uniform_buffer_sizes)[stage][index]);
     Buffer& buffer = slot_buffers[binding.buffer_id];
     TouchBuffer(buffer, binding.buffer_id);
-    const bool use_fast_buffer = binding.buffer_id != NULL_BUFFER_ID &&
-                                 size <= uniform_buffer_skip_cache_size &&
-                                 !memory_tracker.IsRegionGpuModified(cpu_addr, size);
-    if (use_fast_buffer) {
-        if constexpr (IS_OPENGL) {
+
+    if constexpr (IS_OPENGL) {
+        const bool use_fast_buffer = binding.buffer_id != NULL_BUFFER_ID &&
+                                     size <= uniform_buffer_skip_cache_size &&
+                                     !memory_tracker.IsRegionGpuModified(cpu_addr, size);
+        if (use_fast_buffer) {
             if (runtime.HasFastBufferSubData()) {
                 // Fast path for Nvidia
                 const bool should_fast_bind =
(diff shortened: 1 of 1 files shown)


Kelebek1:
One of the YFC PRs has fixed the regressions with this, so re-opening.

v-fox:
This has caused emulation thread hang on startup with latest Legend of Zelda DLC (+ YuzuFix & VisualFixes & 30fps) like this:
```
[  31.781864] Service.HID <Info> core/hid/emulated_controller.cpp:SetPollingMode:1212: Set polling mode 0, device_index=3
[  31.973344] Core.ARM <Critical> core/arm/dynarmic/arm_dynarmic_64.cpp:ExceptionRaised:150: Cannot execute instruction at unmapped address 0x00000000000000
[  31.973353] Core.ARM <Error> core/arm/arm_interface.cpp:LogBacktrace:82: Backtrace, sp=0000000000000000, pc=0000000000000004
[  31.973358] Core.ARM <Error> core/arm/arm_interface.cpp:LogBacktrace:84: Module Name         Address             Original Address    Offset              Symbol
[  31.973359] Core.ARM <Error> core/arm/arm_interface.cpp:LogBacktrace:85: 
[  32.082656] Core.ARM <Error> core/arm/arm_interface.cpp:LogBacktrace:90: unknown             0000007100000004    0000000000000004    0000000000000004    
[  32.082664] Core.ARM <Error> core/arm/arm_interface.cpp:LogBacktrace:90: unknown             0000007100000000    0000000000000000    0000000000000000    
[  43.123256] Frontend <Warning> ./yuzu/bootmanager.h:ForceStop:107: Force stopping EmuThread
```
Surprisingly with both Vulkan and OpenGL. Unless I just really screwed up my test build by something else.

There is also other weird bug which causes complete segfault on emu-thread startup with any game if yuzu is built with clang. Peculiarly, hang happens before segfault, so hanging saves yuzu from crashing completely on build with clang and this patch together. But it's inoperable either way.

PR #10107, opened 2023-04-28
Allow fully customised controller hotkeys

### LINKED ISSUE #10093: Home prepended to all controller bindings in the Hotkey menu

### Affected Build(s)

Mainline 1409, EA-3531

### Description of Issue

When binding a controller shortcut in the hotkey menu all bindings created with a controller will be prepended with `Home+`. It is possible to circumvent this by creating the mapping through editing the .ini file, which will then correctly display the mapping in the Hotkey menu on next launch.

### Expected Behavior

To map key presses and combinations such as `Minus+B` correctly.

### Reproduction Steps

Requires: Plugged in controller.

1. Load up either build 1409/3531
2. Emulation > Configure > Hotkeys
3. Double click a Controller Hotkey option (ie. "Toggle Status Bar")
4. Press the X button on the controller
5. Observe that it attempts to map Home+X instead

### Log File

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/11315471/yuzu_log.txt)

### System Configuration

CPU: Ryzen 3900X
GPU/Driver: Nvidia RTX3090
RAM: 32GB
OS: Fedora 37

OR

Steam Deck w/ SteamOS 3.5

grimkor:
This PR moves away from prefixing `Home+` on all controller hotkey keybinds and instead allows for fully customisable mapping.

Things to note are:
- Work has gone into protecting the TAS functionality from triggering controller hotkeys.
- I moved away from a separate Home/Screenshot button, it looked like the main purpose of this was for protecting against TAS inputs which I have found a method to do without needing to single out these buttons.
- This is my first time working with C++ and I am very open that I have gaps in better ways of doing some of this work.

Unrelated to the work but there is an issue I found where you can record while in playback mode that probably deserves its own issue.

Closes  #10093.

### DIFF
--- a/src/yuzu/configuration/configure_hotkeys.cpp
+++ b/src/yuzu/configuration/configure_hotkeys.cpp
@@ -48,7 +48,9 @@ ConfigureHotkeys::ConfigureHotkeys(Core::HID::HIDCore& hid_core, QWidget* parent
 
     connect(poll_timer.get(), &QTimer::timeout, [this] {
         const auto buttons = controller->GetNpadButtons();
-        if (buttons.raw != Core::HID::NpadButton::None) {
+        const auto home_pressed = controller->GetHomeButtons().home != 0;
+        const auto capture_pressed = controller->GetCaptureButtons().capture != 0;
+        if (home_pressed || capture_pressed) {
             SetPollingResult(buttons.raw, false);
             return;
         }
@@ -154,8 +156,10 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
             model->setData(index, previous_key);
             return;
         }
-
-        const QString button_string = tr("Home+%1").arg(GetButtonName(button));
+        const auto home_pressed = this->controller->GetHomeButtons().home != 0;
+        const auto capture_pressed = this->controller->GetCaptureButtons().capture != 0;
+        const QString button_string =
+            GetButtonCombinationName(button, home_pressed, capture_pressed);
 
         const auto [key_sequence_used, used_action] = IsUsedControllerKey(button_string);
 
@@ -174,72 +178,83 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
     poll_timer->start(200);     // Check for new inputs every 200ms
     // We need to disable configuration to be able to read npad buttons
     controller->DisableConfiguration();
-    controller->DisableSystemButtons();
 }
 
 void ConfigureHotkeys::SetPollingResult(Core::HID::NpadButton button, const bool cancel) {
     timeout_timer->stop();
     poll_timer->stop();
+    (*input_setter)(button, cancel);
     // Re-Enable configuration
     controller->EnableConfiguration();
-    controller->EnableSystemButtons();
-
-    (*input_setter)(button, cancel);
 
     input_setter = std::nullopt;
 }
 
-QString ConfigureHotkeys::GetButtonName(Core::HID::NpadButton button) const {
+QString ConfigureHotkeys::GetButtonCombinationName(Core::HID::NpadButton button,
+                                                   const bool home = false,
+                                                   const bool capture = false) const {
     Core::HID::NpadButtonState state{button};
+    QString button_combination;
+    if (home) {
+        button_combination.append(QStringLiteral("Home+"));
+    }
+    if (capture) {
+        button_combination.append(QStringLiteral("Screenshot+"));
+    }
     if (state.a) {
-        return QStringLiteral("A");
+        button_combination.append(QStringLiteral("A+"));
     }
     if (state.b) {
-        return QStringLiteral("B");
+        button_combination.append(QStringLiteral("B+"));
     }
     if (state.x) {
-        return QStringLiteral("X");
+        button_combination.append(QStringLiteral("X+"));
     }
     if (state.y) {
-        return QStringLiteral("Y");
+        button_combination.append(QStringLiteral("Y+"));
     }
     if (state.l || state.right_sl || state.left_sl) {
-        return QStringLiteral("L");
+        button_combination.append(QStringLiteral("L+"));
     }
     if (state.r || state.right_sr || state.left_sr) {
-        return QStringLiteral("R");
+        button_combination.append(QStringLiteral("R+"));
     }
     if (state.zl) {
-        return QStringLiteral("ZL");
+        button_combination.append(QStringLiteral("ZL+"));
     }
     if (state.zr) {
-        return QStringLiteral("ZR");
+        button_combination.append(QStringLiteral("ZR+"));
     }
     if (state.left) {
-        return QStringLiteral("Dpad_Left");
+        button_combination.append(QStringLiteral("Dpad_Left+"));
     }
     if (state.right) {
-        return QStringLiteral("Dpad_Right");
+        button_combination.append(QStringLiteral("Dpad_Right+"));
     }
     if (state.up) {
-        return QStringLiteral("Dpad_Up");
+        button_combination.append(QStringLiteral("Dpad_Up+"));
     }
     if (state.down) {
-        return QStringLiteral("Dpad_Down");
+        button_combination.append(QStringLiteral("Dpad_Down+"));
     }
     if (state.stick_l) {
-        return QStringLiteral("Left_Stick");
+        button_combination.append(QStringLiteral("Left_Stick+"));
     }
     if (state.stick_r) {
-        return QStringLiteral("Right_Stick");
+        button_combination.append(QStringLiteral("Right_Stick+"));
     }
     if (state.minus) {
-        return QStringLiteral("Minus");
+        button_combination.append(QStringLiteral("Minus+"));
     }
     if (state.plus) {
-        return QStringLiteral("Plus");
+        button_combination.append(QStringLiteral("Plus+"));
     }
-    return tr("Invalid");
+    if (button_combination.isEmpty()) {
+        return tr("Invalid");
+    } else {
+        button_combination.chop(1);
+        return button_combination;
+    }
 }
 
 std::pair<bool, QString> ConfigureHotkeys::IsUsedKey(QKeySequence key_sequence) const {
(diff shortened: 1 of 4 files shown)


src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -174,74 +178,87 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
...
     // We need to disable configuration to be able to read npad buttons
     controller->DisableConfiguration();
-    controller->DisableSystemButtons();
+    // controller->DisableSystemButtons();
 }
 
 void ConfigureHotkeys::SetPollingResult(Core::HID::NpadButton button, const bool cancel) {
     timeout_timer->stop();
     poll_timer->stop();
+    (*input_setter)(button, cancel);
```
grimkor, line 187: Moved above so lines 159-160 work

grimkor:
Updated the PR based on feedback from @german77 regarding npad state order and how I can't just tack on to it.

The only thing I would like help to figure out with this PR is that the line linked below has to be there otherwise Home/Capture buttons get stuck after binding them and I could not figure out where that's being caused.

https://github.com/yuzu-emu/yuzu/pull/10107/files#diff-9dcf85d028051089f92291485b82cf879ecbbe41d5121ab88a5deb2afc195448R191

grimkor:
Upon feedback I've changed the listener for binding a shortcut to require a Home or Capture button rather than any NPad button, this puts the restriction to requiring 2 buttons, one of which MUST be a Capture or Home button.

Interestingly I've noticed that in mainline if you hold Capture and press any button on the controller it'll re-trigger the held button's hotkey (in this case trigger a screenshot). This makes sense given how it currently looks for hotkeys but might be something to look at.

I'm new to this and was wondering if we could take a list of inputs in a given polling period (eg 200ms) and trigger the hotkey with the highest number of buttons required? This won't completely remove the problem but it would stop things like Screenshot+X triggering both Screenshot  and Screenshot+X bindings unless they were pressed pretty far apart.

src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -154,8 +156,10 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
             model->setData(index, previous_key);
             return;
         }
-
-        const QString button_string = tr("Home+%1").arg(GetButtonName(button));
+        bool is_home_pressed = this->controller->GetHomeButtons().home != 0;
```
FearlessTobi, line 159:
```suggestion
        const auto home_pressed = this->controller->GetHomeButtons().home != 0;
```

src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -154,8 +156,10 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
...
+        bool is_capture_pressed = this->controller->GetCaptureButtons().capture != 0;
```
FearlessTobi, line 160:
```suggestion
        const auto capture_pressed = this->controller->GetCaptureButtons().capture != 0;
```

src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -174,74 +178,86 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
...
 
-    (*input_setter)(button, cancel);
+    // Grimkor: Can't see why but Home/Capture button gets stuck unless we do this afterwards.
+    controller->ResetSystemButtons();
 
     input_setter = std::nullopt;
 }
 
-QString ConfigureHotkeys::GetButtonName(Core::HID::NpadButton button) const {
+QString ConfigureHotkeys::GetButtonCombinationName(Core::HID::NpadButton button, bool home = false,
```
FearlessTobi, line 196:
```suggestion
QString ConfigureHotkeys::GetButtonCombinationName(Core::HID::NpadButton button, const bool home_pressed, const bool capture_pressed) const { 
```

src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -174,74 +178,86 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
...
+                                                   bool capture = false) const {
     Core::HID::NpadButtonState state{button};
+    QString some_string = QStringLiteral("");
```
FearlessTobi, line 199:
```suggestion
    QString button_combination;
```

src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -174,74 +178,86 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
...
-        return QStringLiteral("Plus");
+        some_string.append(QStringLiteral("Plus+"));
+    }
+    if (some_string.isEmpty()) {
+        return QStringLiteral("invalid");
+    } else {
+        some_string.chop(1);
+        return some_string;
     }
-    return tr("Invalid");
```
FearlessTobi, line 242: I don't see why this needed to be changed here

src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -174,74 +178,86 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
...
 }
-
```
FearlessTobi, line 244: this newline shouldn't have been removed

src/yuzu/main.cpp
```diff
@@ -3532,6 +3530,20 @@ void GMainWindow::OnTasReset() {
...
 
+bool GMainWindow::CanPerformControllerHotkey(QString action_name) {
+    auto [tas_status, current_tas_frame, total_tas_frames] = input_subsystem->GetTas()->GetStatus();
+    std::regex playback_regex = std::regex("TAS.*?(Start|Reset).*");
+
+    switch (tas_status) {
+    case InputCommon::TasInput::TasState::Running:
+        return std::regex_match(action_name.toStdString(), playback_regex);
+    case InputCommon::TasInput::TasState::Recording:
+        return action_name.toStdString() == "TAS Record";
```
FearlessTobi, line 3541: Does this work when the UI is in a language different from English?

FearlessTobi (changes requested):
I've only reviewed the code itself.
For the input-specific stuff, @german77 please take a look.

MonsterDruide1:
By the way, recording during TAS playback is more or less intended. It's a positive side effect that can be used to "extend" existing TASes, by playing back the existing file, then taking over and doing more inputs manually.

src/yuzu/main.cpp
grimkor: I checked again and while it does work on all languages I think your question is legitimate and this is a weak solution. What do you think if instead I change `LinkActionShortcut()` to the following:

`void LinkActionShortcut(QAction* action, const QString& action_name, bool tas_allowed);`

And it defaults to `false` to save putting in a value every time?

@german77 maybe a question for you?

grimkor:
> By the way, recording during TAS playback is more or less intended. It's a positive side effect that can be used to "extend" existing TASes, by playing back the existing file, then taking over and doing more inputs manually.

Cool I'll make sure that functionality stays intact then, thanks for letting me know.

src/yuzu/configuration/configure_hotkeys.cpp
```diff
@@ -154,8 +156,10 @@ void ConfigureHotkeys::ConfigureController(QModelIndex index) {
             model->setData(index, previous_key);
             return;
         }
-
-        const QString button_string = tr("Home+%1").arg(GetButtonName(button));
+        const auto home_pressed = this->controller->GetHomeButtons().home != 0;
+        const auto capture_pressed = this->controller->GetCaptureButtons().capture != 0;
+        const QString button_string =
```
FearlessTobi, line 161:
```suggestion
        const QString button_string = GetButtonCombinationName(button, home_pressed, capture_pressed);
```

src/yuzu/main.cpp
```diff
@@ -3483,11 +3490,11 @@ void GMainWindow::OnTasStartStop() {
     if (!emulation_running) {
         return;
     }
-
```
FearlessTobi, line 3486: should not have been changed

src/yuzu/main.cpp
```diff
@@ -3483,11 +3490,11 @@ void GMainWindow::OnTasStartStop() {
...
     // Disable system buttons to prevent TAS from executing a hotkey
     auto* controller = system->HIDCore().GetEmulatedController(Core::HID::NpadIdType::Player1);
     controller->ResetSystemButtons();
 
+
```
FearlessTobi, line 3497: should not have been changed

src/yuzu/main.cpp
```diff
@@ -3500,7 +3507,6 @@ void GMainWindow::OnTasRecord() {
         return;
     }
 
-    // Disable system buttons to prevent TAS from recording a hotkey
```
FearlessTobi, line 3503: should not have been changed

FearlessTobi (changes requested):
Some of my older comments also have not been addressed yet.

grimkor:
> Some of my older comments also have not been addressed yet.

Sorry mate I'll get that fixed, thought I had them all.

src/yuzu/configuration/configure_hotkeys.cpp
grimkor: Moved it back to `tr()`, if you are asking about the change to the whole method then it allows for more complex hotkeys to be bound such as `Home+Screenshot+L+R` , which would be nice for some shortcuts like restarting/stopping emulation making it harder to fumble and easier to trigger on purpose.

src/yuzu/main.cpp
```diff
@@ -7,6 +7,7 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
+#include <regex>
```
german77, line 10: This isn't used anymore

german77 (approved):

PR #10110, opened 2023-04-28
vk_pipeline_cache: Skip compute pipelines on Intel proprietary drivers

Morph1984:
Intel's SPIR-V shader compiler is broken. For now, skip compiling any compute pipelines until they fix this issue.
This is not a perfect workaround, as there are a small subset of non-compute pipelines that still cause it to crash, but this should cover the majority of crashes.
It is unfortunate that even with a test case reported 6 months ago the issue has not been fixed in favor of fixing "[the most popular games and apps](https://github.com/IGCIT/Intel-GPU-Community-Issue-Tracker-IGCIT/issues/159#issuecomment-1527618555)".
Intel, you can do better than this.


goldenx86:
For the record, Intel is not interested in fixing the issue at all.
https://github.com/IGCIT/Intel-GPU-Community-Issue-Tracker-IGCIT/issues/159#issuecomment-1527618555

bunnei (approved):

PR #10119, opened 2023-04-30
Attempt at improving HD Rumble emulation

marius851000:
I use a game controller that is technically comatible with HD Rumble (a DualSense), but Yuzu totally ignore the fact, and doesn’t emulate it other than using standard rumble (I may try to add proper support for it, but I’ll need to learn more about SDL and likely CPP, hoping I don’t have to having to generate the wave myself, like I did in a python prototype).

This change the amplitude of the rumble based upon the frequency. This allow to have some sort of feel for the frequency. It first have no change for frequency from 0 to 100 Hz, then lower linearly from 100 Hz to 600 Hz (Mario Odyssey in practice doesn’t go over 420).

Those parameter might be changed, but this still seems to provide better rumble.

I also increased the max duration of Rumble. There is an effect on the SMO scooter that, when not moving, continue to vibrate. I didn’t went up-to disabling it due to being worried it will continue to vibrate even in the event of a crash, thought.

*ps: I’ve had far better performance against EA than mainline, but seemingly nothing related to rumble*

### DIFF
--- a/src/input_common/drivers/sdl_driver.cpp
+++ b/src/input_common/drivers/sdl_driver.cpp
@@ -109,14 +109,37 @@ public:
     }
 
     bool RumblePlay(const Common::Input::VibrationStatus vibration) {
-        constexpr u32 rumble_max_duration_ms = 1000;
+        constexpr u32 rumble_max_duration_ms = 2000;
+        constexpr f32 low_start_sensitivity_limit = 140.0;
+        constexpr f32 low_width_sensitivity_limit = 400.0;
+        constexpr f32 high_start_sensitivity_limit = 200.0;
+        constexpr f32 high_width_sensitivity_limit = 700.0;
+        // Try to provide some feeling of the frequency by reducing the amplitude depending on it.
+        f32 low_frequency_scale = 1.0;
+        if (vibration.low_frequency > low_start_sensitivity_limit) {
+            low_frequency_scale =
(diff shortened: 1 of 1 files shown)


german77:
I like the idea but seems like the effect is too strong. Low frequency range is from 40Hz-628Hz with the default at 160Hz, High is from 80Hz-1256Hz with the default at 320Hz.

So basically any game that relies on high frequency will output zero amplitude most of the time.  I suggest different scales for both high and low. Make the minimum output to 25% of the original amplitude value. So in extreme cases we still have some rumble signal.

marius851000:
Done. And I have to say, that’s indeed a really good idea. (and took time to make some tuning too) I can now feels in a kinda satisfying way the scooter, and I can locate moons in the ground (in Mario Odyssey, which is the only game I have on my PC that use the HD Rumble).

I still plan to looks into making proper use of the DualSense coils, but It seems to be complicated, and might just wait until SDL does something about that.

german77:
From what I read you can send raw commands to ds5 controllers with sdl. This was done for the triggers so it's technically possible to actually use HD rumble. But yeah it's quite complicated to get it working without any documentation.

Code looks good now but I need to test it for a bit to ensure nothing is out of place.

german77:
Going to leave it a day in EA to see if there are any issues. For now code looks good.

PR #10125, opened 2023-04-30
configuration: Expose separate swap present modes

lat9nq:
See https://github.com/yuzu-emu/yuzu/pull/10125#issuecomment-1530846657 for updated info

Previously, yuzu would try and guess which V-Sync mode to use given different scenarios, but apparently we didn't always get it right. This exposes the separate modes in a drop-down the user can select.

If a mode isn't available in Vulkan, it defaults to FIFO. This also promotes V-Sync from adv. graphics to graphics, and removes it from custom configurations.

~~The swap interval values in bootmanager needs to be looked over, as I have no idea if those values are accurate.~~


src/core/telemetry_session.cpp
```diff
@@ -85,6 +85,18 @@ static const char* TranslateNvdecEmulation(Settings::NvdecEmulation backend) {
     return "Unknown";
 }
 
+constexpr const char* TranslateVSyncMode(Settings::VSyncMode mode) {
```
lioncash, line 88:
```suggestion
static constexpr const char* TranslateVSyncMode(Settings::VSyncMode mode) {
```

lat9nq:
There's some changes I want to add to make this less cumbersome to configure.

lat9nq:
I'm choosing to continue using this PR since the main objective has not changed, though the implementation has been overhauled a bit.

For Vulkan, configure_graphics now exposes the different present modes available from the driver. This means that we are only showing modes that the driver supports, and the user cannot select an unsupported mode. Changing the V-Sync mode during runtime is also available now for Vulkan.

Incorrect language around "Double Buffering" or "Triple Buffering" has been removed. FIFO is considered "V-Sync On", and Immediate "V-Sync Off". Though, users likely want to pick Mailbox if it is available.

Unlocking the framerate now picks Mailbox, then Immediate, then defaults to the current mode depending on what is available. When re-locking the framerate, it returns to the user-requested presentation mode. (If the user is already using Mailbox or Immediate, nothing changes here.)

If the user picks OpenGL, the V-Sync options are On or Off.

lat9nq:
Conflicts with #9973 so untagging for now.

src/yuzu/qt_common.h
```diff
@@ -0,0 +1,12 @@
...
+
+#pragma once
+
+#include <QWindow>
+#include "core/frontend/emu_window.h"
+
+namespace YuzuQtCommon {
+Core::Frontend::WindowSystemType GetWindowSystemType();
+Core::Frontend::EmuWindow::WindowSystemInfo GetWindowSystemInfo(QWindow* window);
+} // namespace YuzuQtCommon
```
FearlessTobi, line 12: should have a newline here

src/yuzu/qt_common.h
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: 2023 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <QWindow>
+#include "core/frontend/emu_window.h"
+
+namespace YuzuQtCommon {
```
FearlessTobi, line 9:
```suggestion
namespace QtCommon {
```

src/yuzu/configuration/configure_graphics.cpp
```diff
@@ -4,20 +4,76 @@
 // Include this early to include Vulkan headers how we want to
 #include "video_core/vulkan_common/vulkan_wrapper.h"
 
+#include <algorithm>
```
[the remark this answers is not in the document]
lat9nq: I used [include-what-you-use](https://include-what-you-use.org/) to determine what headers/forward decls needed to be here. I mean implicitly these headers were already included, but for facilitating code navigation and actually being able to see what this file uses, I prefer to have everything visible up front. (The commits that I did this with have `[IWYU]` appended to them.)

I can tone that back if it's preferred?

FernandoS27 (approved):

src/yuzu/configuration/configure_graphics.cpp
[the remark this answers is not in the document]
lat9nq: Yeah I understand. I want to do that, but it's a large undertaking, especially since IWYU wants to include i.e. qimage.h instead of QImage. Makes it more difficult to let it automatically make the changes.

What I did here was understand that I may have added headers to these files that I didn't need anymore, and knew of a tool that cleans this up. And while we're at it, just clean up the rest of the headers/forward decls.

bunnei (approved):

PR #10142, opened 2023-05-02
GPU: implement missing ASTC

FernandoS27:
This implements the missing ASTC formats remaining. So finally we can forget about that format for good. It also fixes some bugs with its classes.


Morph1984:
Missing changes here:
https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/renderer_opengl/gl_texture_cache.cpp#L851-L869

Morph1984 (approved):

bunnei (approved):

PR #10162, opened 2023-05-05
yuzu-sdl,audio_core: Remove antiquated warning ignore

lat9nq:
Issue was fixed a long time ago, both by SDL2 and in yuzu by including SDL2 as a system library.

Partially reverts f973274 (the other files seemed to have been cleaned up by now).

### DIFF
--- a/src/audio_core/sink/sdl2_sink.cpp
+++ b/src/audio_core/sink/sdl2_sink.cpp
@@ -3,22 +3,13 @@
 
 #include <span>
 #include <vector>
+#include <SDL.h>
 
 #include "audio_core/common/common.h"
 #include "audio_core/sink/sdl2_sink.h"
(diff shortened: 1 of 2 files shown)


abouvier:
The true include path is `<SDL2/SDL.h>` by the way.

abouvier:
Right, I come from [the future](https://github.com/libsdl-org/SDL/issues/6575).

v1993:
Interesting. Technically, that's what SDL3 rather than SDL2 recommends (although [SDL3 FAQ needs an update](https://wiki.libsdl.org/SDL3/FAQDevelopment#do_i_include_sdl.h_or_sdlsdl.h)). IMO it would make the most sense to change include to new form when yuzu eventually switches to SDL3.

liamwhite (approved):

PR #10178, opened 2023-05-06
Fix read access violation after 2 hours of gameplay in PMD: RTDX

### LINKED ISSUE #10126: Advance wars crashes after 1-2 hours

### Affected Build(s)

Mainline 1408

### Description of Issue

Yuzu closes to desktop after some time while playing this game. No ram leaks or other kind of weird behavior happens while playing.

### Expected Behavior

The game shouldn't crash over time

### Reproduction Steps

Play the game for about 1-2 hours until finally decides to crash. Seems to be between scenes so just leaving the emulator running will not trigger the crash 

### Log File

The log is always cut short. Here is the actual place where it breaks
image

### System Configuration

CPU: i7 8700
GPU/Driver: NVIDIA GTX 1060 6GB 531.29.0.0
RAM: 16GB
OS: Windows 10 Version 21H2 (Build 19044.2846)

### LINKED ISSUE #10168: Pokémon Mystery Dungeon crashing after a few hours of gameplay
### Issue

Report from https://community.citra-emu.org/t/pokemon-mystery-dungeon-rescue-team-dx-crashes-after-2-hours-of-gameplay/772370/11

image

i7-13700K
Windows 11 Version 22H2 (Build 22621.1555)
32GB
RTX 3070, 531.68

The issue happens in both OpenGL and Vulkan.

ronikirla:
fixes #10168 
potentially also fixes #10126

OP from the support thread linked in the first issue here. Now I may not exactly be the most qualified person to work on this, but yesterday a Twitch chatter who claims to work at Google took a look at my issue and seems to be "almost certain" that this should fix the issue. Apparently, due to company policy, they are not allowed to contribute to emulators.

Tried out the fix today, and for the first time ever I was able to play through the entire game up to the credits in a single sitting without a crash. Can't really explain much of this myself, but apparently devs of this should be able to "immediately understand" the change.

### DIFF
--- a/src/common/address_space.inc
+++ b/src/common/address_space.inc
@@ -72,7 +72,7 @@ MAP_MEMBER(void)::MapLocked(VaType virt, PaType phys, VaType size, ExtraBlockInf
                 }
             }()};
 
-            if (block_end_predecessor->virt >= virt) {
+            if (block_end_predecessor != blocks.begin() && block_end_predecessor->virt >= virt) {
                 // If this block's start would be overlapped by the map then reuse it as a tail
                 // block
                 block_end_predecessor->virt = virt_end;


SwitchMan92:
Basically it checks block_end_predecessor is not at the start before calling prev() on it, hence avoiding unauthorized stack access.

liamwhite (approved):
Nice catch, thanks!

german77:
Advance wars is stable with this change. Pokémon mystery dungeon should also be fixed by this and probably many other games.

utku-kaan:
Good catch! This fix ensures that the first block isn't reused as a tail. It also prevents potential issues caused by calling `prev()` on the first block, which would lead to unauthorized memory access.

PR #10181, opened 2023-05-07
configure_graphics: Add option to enable compute pipelines for Intel proprietary

lat9nq:
Some games' compute pipelines cause yuzu to crash on the Intel proprietary Vulkan driver (e.g. Xenoblade Chronicles 3). Other game not only work fine but also absolutely need compute pipelines to render properly (e.g. Pokemon Brilliant Diamond/Shining Pearl, Super Smash Bros. Ultimate).

Adds an option to Advanced Graphics for users to enable compute pipelines with. This option only affects the Intel proprietary Vulkan driver. The option is visible **iff** the Intel driver was found during Vulkan device enumeration.

Continuation of #10110


src/yuzu/configuration/configure_per_game.cpp
```diff
@@ -48,8 +48,9 @@ ConfigurePerGame::ConfigurePerGame(QWidget* parent, u64 title_id_, const std::st
     audio_tab = std::make_unique<ConfigureAudio>(system_, this);
     cpu_tab = std::make_unique<ConfigureCpu>(system_, this);
     general_tab = std::make_unique<ConfigureGeneral>(system_, this);
-    graphics_tab = std::make_unique<ConfigureGraphics>(system_, this);
     graphics_advanced_tab = std::make_unique<ConfigureGraphicsAdvanced>(system_, this);
+    graphics_tab = std::make_unique<ConfigureGraphics>(
+        system_, [&]() { graphics_advanced_tab->ExposeComputeOption(); }, this);
```
FearlessTobi, line 53: why do we need to call this here, instead of doing it in the graphics tab constructor?
This all seems a bit convoluted.

lat9nq: When I asked internally, @liamwhite and @goldenx86 agreed the setting should go in Graphics Advanced.

lat9nq: Maybe I didn't understand what you were asking? I created a lambda in the constructor that just calls that function. The lambda doesn't get called until it's needed. I can't pass the method directly because I need to pass a function that isn't directly tied to a class.

liamwhite (approved):

PR #10207, opened 2023-05-09
service: nfp: Allow to load with a different amiibo id

german77:
Some games don't allow to load the same amiibo more than once. This generates a random id so the game always thinks it's a new one.

This feature can be enabled in advanced controller settings.

### DIFF
--- a/src/yuzu/configuration/configure_input_advanced.ui
+++ b/src/yuzu/configuration/configure_input_advanced.ui
@@ -2728,6 +2728,22 @@
                      </widget>
                    </item>
                    <item row="7" column="0">
+                     <widget class="QCheckBox" name="random_amiibo_id">
+                       <property name="toolTip">
+                         <string>Allows unlimited uses of the same Amiibo in games that would otherwise limit you to one use.</string>
+                       </property>
+                       <property name="minimumSize">
+                         <size>
+                           <width>0</width>
+                           <height>23</height>
+                         </size>
+                       </property>
+                       <property name="text">
+                         <string>Use random Amiibo ID</string>
+                       </property>
+                     </widget>
+                   </item>
+                   <item row="8" column="0">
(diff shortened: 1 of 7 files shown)


src/yuzu/configuration/configure_input_advanced.ui
```diff
@@ -2728,6 +2728,19 @@
...
                    <item row="7" column="0">
+                     <widget class="QCheckBox" name="random_amiibo_id">
+                       <property name="minimumSize">
+                         <size>
+                           <width>0</width>
+                           <height>23</height>
+                         </size>
+                       </property>
+                       <property name="text">
+                         <string>Use random amiibo ID</string>
```
lioncash, line 2739: May want to add a tooltip explaining why you'd even want to enable this setting. That or phrase it like "Allow unlimited uses of same Amiibo" or something similar that's user-friendly

src/yuzu/configuration/configure_input_advanced.ui
```diff
@@ -2728,6 +2728,22 @@
                      </widget>
                    </item>
                    <item row="7" column="0">
+                     <widget class="QCheckBox" name="random_amiibo_id">
+                       <property name="toolTip">
+                         <string>Allow unlimited uses of same Amiibo</string>
```
lioncash, line 2733: Since it's a tooltip, you might want to expand this as.

```suggestion
                         <string>Allows unlimited uses of the same Amiibo in games that would otherwise limit you to one use.</string>
```

FearlessTobi:
Is there any reason to make this a setting?
Are there any cases where a random id is **not** wanted?

german77:
Yes. SSBU for example will error out if the amiibo ID changes on subsequent loads. The uuid also plays an important role in encryption so we can't change the uuid for the whole session either.

PR #10243, opened 2023-05-12
Correctly track render target index in the framebuffer for image aspects

Kelebek1:
The framebuffer sequentially adds only valid render targets to `image_ranges`. However, for clears, we look up the image aspect by render target index, and since the framebuffer skips invalid RTs, this can lead to a mis-match between the requested clear RT and the aspect of the RT we look up.

In Tears of the Kingdom, it uses a framebuffer with colour render targets 0/1/2/4, where 3 is invalid. Our framebuffer adds these into `image_ranges` as 0/1/2/3, and then the depth gets index 4. At the start of the renderpass it tries to clear RT index 4 with a colour clear, which looks up `image_ranges[4]` in `framebuffer->HasAspectColorBit()` and returns false as that's now the depth target, and the clear is skipped entirely. This PR just adds a quick mapping between real RT index and the new framebuffer index.

This fixes the red dot on the right of the screen in Tears of the Kingdom, and all of the big graphical corruption in handheld.


Squall-Leonhart:
I'm curious why this issue only manifests under 1.0.0 for most people, 1.1.0 has neither the red dot or the neon explosion for most.

FernandoS27 (approved):

PR #10265, opened 2023-05-13
input_common: Make amiibo scanning less demanding

german77:
Scanning for amiibos is a really demanding task. We don't need to scan every frame so I reduced the scan rate to 5hz and reduced the number of attempts to get valid data to 1 instead of 7.

This makes the right joycon usable while using the amiibo rune on totk. Small stutters will still happen but nothing like current master

### DIFF
--- a/src/input_common/helpers/joycon_protocol/nfc.h
+++ b/src/input_common/helpers/joycon_protocol/nfc.h
@@ -32,6 +32,9 @@ public:
     bool IsEnabled() const;
 
 private:
+    // Number of times the function will be delayed until it outputs valid data
+    static constexpr std::size_t AMIIBO_UPDATE_DELAY = 15;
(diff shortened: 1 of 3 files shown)


liamwhite (approved):

src/input_common/helpers/joycon_protocol/nfc.h
```diff
@@ -32,14 +32,16 @@ class NfcProtocol final : private JoyconCommonProtocol {
     bool IsEnabled() const;
 
 private:
+    static constexpr std::size_t AMIIBO_UPDATE_DELAY = 15;
```
lioncash, line 35: maybe have a comment indicate what the update delay is in terms of here

liamwhite (approved):

PR #10294, opened 2023-05-14
service: hid: Use span instead of vector reference

german77:
Prefer span rather than vector references.


liamwhite (approved):

Kelebek1:
You're still using a vector for vibration_values. Can you use the data from ctx directly, get rid of both vibration_values and vibration_device_handles, and skip the memcpy as well?

PR #10302, opened 2023-05-14
Rework GC to be based on available card memory

Kelebek1:
Currently we're using hard-coded values for the GC memory limits, but we have access to the device's heap limits already, and through the device mem info extension we know the currently-available allocateable memory too. This PR changes the GC to use the card's heap limits, hopefully making it more appropriate for low VRAM cards and high VRAM cards.

I also decided to re-work the GC logic to not call Finish() so much. In a lot of places in our code we do download -> wait -> memcpy -> download -> wait -> memcpy -> download -> wait -> memcpy. It's more preferable to just download -> download -> download -> wait -> memcpy -> memcpy -> memcpy. It just means we need some extra loops to split these sections. This is something I want to extend to other places as well after this, as it gives us the possibility to combine copies on the same images/buffers and make fewer API calls. This change is making up most of the diff here.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -23,42 +23,94 @@ BufferCache<P>::BufferCache(VideoCore::RasterizerInterface& rasterizer_,
     common_ranges.clear();
     inline_buffer_id = NULL_BUFFER_ID;
 
-    if (!runtime.CanReportMemoryUsage()) {
-        minimum_memory = DEFAULT_EXPECTED_MEMORY;
-        critical_memory = DEFAULT_CRITICAL_MEMORY;
-        return;
-    }
-
     const s64 device_memory = static_cast<s64>(runtime.GetDeviceLocalMemory());
-    const s64 min_spacing_expected = device_memory - 1_GiB - 512_MiB;
-    const s64 min_spacing_critical = device_memory - 1_GiB;
-    const s64 mem_threshold = std::min(device_memory, TARGET_THRESHOLD);
-    const s64 min_vacancy_expected = (6 * mem_threshold) / 10;
-    const s64 min_vacancy_critical = (3 * mem_threshold) / 10;
-    minimum_memory = static_cast<u64>(
-        std::max(std::min(device_memory - min_vacancy_expected, min_spacing_expected),
-                 DEFAULT_EXPECTED_MEMORY));
-    critical_memory = static_cast<u64>(
-        std::max(std::min(device_memory - min_vacancy_critical, min_spacing_critical),
-                 DEFAULT_CRITICAL_MEMORY));
+    const u64 device_mem_per = device_memory / 100;
+    minimum_memory = device_mem_per * 25;
+    expected_memory = device_mem_per * 50;
+    critical_memory = device_mem_per * 80;
+    LOG_INFO(HW_GPU, "Buffer cache device memory limits: min {} expected {} critical {}",
+             minimum_memory, expected_memory, critical_memory);
 }
 
 template <class P>
 void BufferCache<P>::RunGarbageCollector() {
-    const bool aggressive_gc = total_used_memory >= critical_memory;
-    const u64 ticks_to_destroy = aggressive_gc ? 60 : 120;
-    int num_iterations = aggressive_gc ? 64 : 32;
-    const auto clean_up = [this, &num_iterations](BufferId buffer_id) {
+    if (total_used_memory < minimum_memory) {
+        return;
+    }
+    bool is_expected = total_used_memory >= expected_memory;
+    bool is_critical = total_used_memory >= critical_memory;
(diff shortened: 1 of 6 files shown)


liamwhite:
This should actually fix #10245.

eXxe-elcior:
I tried this one (yuzu-msvc-verify-pr10302-20230514-bd3bf7f30.exe) it works as you stated. I'm not a dev, but I noticed some weird graphical hitches, when I open the menu to select a weapon, shield or a bow and exists by releasing the keys according to the choice, the textures in the world become dark for a sec and gets back to normal. As if it gets dropped when I open any game menu (Weapons, food, outfits, save menu...), and when I exist it gets reloaded again as if it's not in the memory. this causes a split second of stutter. The amazing thing that I noticed is that VRAM doesn't surpass 4GB at all. it reaches 4.5GB and drops again, to 3.9 and stabilizes to 4-4.1GB. 
Test was done on LoZ : Tears of the Kingdom on AMD Ryzen R7 5700X - 16GB 3200MT/s - RX 6700XT.
Settings : every hack is enabled on Vulkan 2X Resolution with CPU Auto.
I hope this helps.

Joyfal:
EA 3591 (with this PR) has some regressions with the weapon switcher & menus in TOTK. Big lag during/after switching weapons. Issue is not nearly as bad in 3589. 

Specs: 13600k, 32GB RAM, 3080Ti

sharkiller:
> I tried this one (yuzu-msvc-verify-pr10302-20230514-bd3bf7f30.exe) it works as you stated. I'm not a dev, but I noticed some weird graphical hitches, when I open the menu to select a weapon, shield or a bow and exists by releasing the keys according to the choice, the textures in the world become dark for a sec and gets back to normal. As if it gets dropped when I open any game menu (Weapons, food, outfits, save menu...), and when I exist it gets reloaded again as if it's not in the memory. this causes a split second of stutter. The amazing thing that I noticed is that VRAM doesn't surpass 4GB at all. it reaches 4.5GB and drops again, to 3.9 and stabilizes to 4-4.1GB. Test was done on LoZ : Tears of the Kingdom on AMD Ryzen R7 5700X - 16GB 3200MT/s - RX 6700XT. Settings : every hack is enabled on Vulkan 2X Resolution with CPU Auto. I hope this helps.

i experience the same. a lot of graphicals bugs. just stand still open the ingame settings, close the ingame settings and all the floor textures became dark for a second. 
also a lot of items textures or UI became corrupted the first time they are shown so you need to change between menus to get fixed. 
items images on the sellers overlay also get corrupted the first time you see the item.

Sincasios:
I'm not sure if is related, but the two comments that are here speaks about graphical issues after some time (maybe for the memory usage of the VRAM?).

Yesterday I saw some graphical issues that are increasing over the time. For example, like this:
yuzu Early Access 3589 _ The Legend of Zelda_ Tears of the Kingdom (64-bit) _ 1 1 0 _ NVIDIA 14_05_2023 16_34_21

Ind when you teleport to other place, the teleport animation (the green lines that apperas) are laggy and the animation is not fluid (but the game was fluid). I restarted yuzu and it fixed the teleport visual artifats .

This with a RTX3060 with 12 GB

Sincasios:
I captured what I'm saying:
https://youtu.be/jEgC8LopiAo

In this case it seems that the problem starts when I enter the depths, and when I exit, the first teleport is OK (on start and end)

SweViver:
The new fix added into 3591 did work perfectly on my 4090, no issues at all and VRAM only hitting 7-9Gb. With 3592 and 3593, the VRAM spikes at 22-23Gb again and major slowdowns happen or framerate just gets stuck at 20-30fps instead of previous solid 60fps.

QA-Colin:
> The new fix added into 3591 did work perfectly on my 4090, no issues at all and VRAM only hitting 7-9Gb. With 3592 and 3593, the VRAM spikes at 22-23Gb again and major slowdowns happen or framerate just gets stuck at 20-30fps instead of previous solid 60fps.

This was removed after 3591 due to reported rendering issues. Needs additional fixes

bryantmh:
Just commenting that I've had a similar experience in TotK as SweViver on my RTX 4080. The game is basically unplayable after about 30 minutes when the VRAM gets filled on the latest EA build. 3591 with this PR integrated runs flawlessly. I'd love to see this PR get some more work so we're not stuck on old versions.

Squall-Leonhart:
This PR could also be used in tandem with enabling VK_EXT_pageable_device_local_memory, which will enable Turing and newer to avoid some cases of OOM on the Device Local memory heap.

NFSS10:
I tested this with my 4GB RX 580 and I could finally reach the botoom of the depths. It still crashs, even with 0.5x resolution. The VRAM usage rises until 4GB and after some time it crashes.

imagem
It freezes at the first arrow and crashes at the second arrow

This doesn't seem to fully fix the crashing when entering the depths problem but it's progress👍

PR #10347, opened 2023-05-16
Allow vertical scrolling in input configuration

redstrate:
On screens that are limited in height (like the Steam Deck) some parts of the input configuration window are cut off. This includes the motion control configuration, which is a little annoying as you can't easily "resize" a portable device:

image

I originally got around this by changing the screen orientation, but I think we could probably improve this through the UI. So now, if the screen cannot fit the input configuration it will start scrolling:

image

Note that the vertical scrollbar should disappear when the window is the minimum size or bigger, so this wouldn't affect users with large desktop screens. I also disabled the frame rendering, so there is a frame-inside-frame nonsense going on.

Also a bonus I set a minimum size for the two ZR/ZL sliders near the top, because on certain themes like KDE Plasma's Breeze the handles are _huge_ and got cut off.

The controller input bottom panel also gets shoved into the bottom. Should I instead pin that at the bottom, as I think that's what it was originally meant to do?

### DIFF
--- a/src/yuzu/configuration/configure_input.cpp
+++ b/src/yuzu/configuration/configure_input.cpp
@@ -3,6 +3,7 @@
 
 #include <memory>
 #include <thread>
+#include <QScrollArea>
 
 #include "core/core.h"
 #include "core/hid/emulated_controller.h"
@@ -112,7 +113,13 @@ void ConfigureInput::Initialize(InputCommon::InputSubsystem* input_subsystem,
 
     for (std::size_t i = 0; i < player_tabs.size(); ++i) {
         player_tabs[i]->setLayout(new QHBoxLayout(player_tabs[i]));
-        player_tabs[i]->layout()->addWidget(player_controllers[i]);
+
+        auto scroll_area = new QScrollArea(player_tabs[i]);
+        player_tabs[i]->layout()->addWidget(scroll_area);
+        scroll_area->setWidget(player_controllers[i]);
+        scroll_area->setWidgetResizable(true);
+        scroll_area->setFrameShape(QFrame::Shape::NoFrame);
+
         connect(player_controllers[i], &ConfigureInputPlayer::Connected, [&, i](bool is_connected) {
             // Ensures that the controllers are always connected in sequential order
             if (is_connected) {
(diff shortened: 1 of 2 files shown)


v1993:
IMO ideally interface should scale down if it does not fit the window, but I'm not sure if this can be cleanly/easily done in Qt.

FearlessTobi:
There's also this PR which essentially does the same thing: https://github.com/yuzu-emu/yuzu/pull/9499
Maybe it has some useful changes to port over to this one?

german77:
I have an issue with this PR. On windows the scroll bars are visible on the default size. Left image is PR9499 right one is PR10347
image
Current master has the same window size but it's able to fit everything without a need of a scroll bars
image

redstrate:
> IMO ideally interface should scale down if it does not fit the window, but I'm not sure if this can be cleanly/easily done in Qt.

Not with QtWidgets at least, there's not a clean way to scale things AFAIK. It might be possible to adjust the lahyout further in the .ui to accommodate smaller screens but I feel like the scrollbar solution is the easiest & least intrusive option. It's a small change, so if anybody comes along and makes the necessary adjustments it should be easy to revert this commit :-)

> There's also this PR which essentially does the same thing: #9499 Maybe it has some useful changes to port over to this one?

I don't know how I didn't catch that before writing my own :laughing: It looks like it does the same thing in practice, although it looks like it has a bunch of unrelated and unexplained .ui changes (although funnily enough, they also increase the minimum height of those sliders).

> I have an issue with this PR. On windows the scroll bars are visible on the default size. Left image is PR9499 right one is PR10347
> Current master has the same window size but it's able to fit everything without a need of a scroll bars

I spent a bit of time digging into this, and this is seems to be expected behavior. We use adjustSize() to adjust the settings dialog, and according [to the documentation](https://doc.qt.io/qt-6/qwidget.html#adjustSize) Qt will shrink the dialog:

> For windows, the screen size is also taken into account. If the [sizeHint](https://doc.qt.io/qt-6/qwidget.html#sizeHint-prop)() is less than (200, 100) and the size policy is [expanding](https://doc.qt.io/qt-6/qsizepolicy.html#Policy-enum), the window will be at least (200, 100). The maximum size of a window is 2/3 of the screen's width and height.

I'm not sure why there is a difference between master and the two PRs honestly, but on my system there is no difference between master & PR10347 (they both take screen height into account, and shrink). On master, part of the input window is cut off but when scaling the display back to 100% - to gain more screen height - the settings dialog scales to accommodate as expected.

Issue #9876, opened 2023-02-28
Custom Game Speed for various reasons


Waikuteru:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What feature are you suggesting?

Hello,

I'm a Zelda Breath of the Wild modder and the Custom Game Speed (CPU timer in Cemu) helped me a lot developing mods.

Speeding of cutscenes or just being able to walk faster is such a timesaver! I would really appreciate it if you add this feature. If you do it, it would be nice to have hotkeys. For example Shift/Alt * F1 for x1 Game Speed, Shift/Alt + F2 for x2 Game Speed, and so on...

I want to bring to your attention that modding will be a huge deal of this emulator as well. Soon, when Zelda Tears of the Kingdom is out, mods are expected to be produced and any help to make this faster will also ultimately result into more users for your emulator.

I also hope that the mod loading time gets improved soon. It takes quite soom time, even loading 100MBs. I'm not sure WHY it is so slow, so if you could explain me, that would be nice! 

### Why would this feature be useful?

For faster mod developing and to get to certain game locations faster, or even videos where it shows 24hours in a timelapse.

kiri11:
It would be also helpful to increase FPS without mods: just unlock FPS and decrease game speed to match it.

Issue #9280, opened 2022-11-20
More Fine-tuned Speed Limiter Hotkeys/Favourites


NemesisTWarlock:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What feature are you suggesting?

Hotkeys to decrease/increase the CPU Speed Limiter (For example, "Ctrl +" To increase the limiter by 10%, or "Ctrl -" to decrease it by 10%).  A further, more detailed suggestion is the possibility to perhaps switch between two different user chosen limiter values (Such as "Ctrl+1" or "Ctrl+2").

### Why would this feature be useful?

This is mostly an Accessiblility/QoL feature. For instance, a player playing Bayonetta would be able to set the game speed to 50% to slow down the combat during difficult sections, and then when completed, resume the game at 100% speed. 

The current implementation requires the user to pause the game, open the game specific menu, and then manually type in the requested speed each time. This would simply be a way to hotkey that interaction.

Many other emulators have similar slowdown/speedup hotkeys.

goldenx86:
Yes please.

FearlessTobi:
This can probably just be ported from the Citra PR.

PR #10363, opened 2023-05-17
add hotkeys to increase/decrease speed percent

NebulaGazer:
Adds the hotkeys CTRL+= and CTRL+- to increase/decrease the game speed percent by a step of 50. 
Satisfies issues #9876 and #9280. Solves the same issue that PR #10324 is going for. 
Can enhance down the line for a configurable step size via config window, but I think +-50 is going to cover most current cases.

### DIFF
--- a/src/yuzu/configuration/configure_general.ui
+++ b/src/yuzu/configuration/configure_general.ui
@@ -42,7 +42,7 @@
                <string>%</string>
               </property>
               <property name="minimum">
-               <number>1</number>
+               <number>0</number>
               </property>
               <property name="maximum">
                <number>9999</number>
--- a/src/yuzu/configuration/config.cpp
+++ b/src/yuzu/configuration/config.cpp
@@ -70,7 +70,7 @@ const std::array<int, 2> Config::default_ringcon_analogs{{
 // This must be in alphabetical order according to action name as it must have the same order as
 // UISetting::values.shortcuts, which is alphabetically ordered.
 // clang-format off
-const std::array<UISettings::Shortcut, 22> Config::default_hotkeys{{
+const std::array<UISettings::Shortcut, 24> Config::default_hotkeys{{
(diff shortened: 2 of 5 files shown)


src/yuzu/configuration/configure_general.ui
```diff
@@ -42,7 +42,7 @@
                <string>%</string>
               </property>
               <property name="minimum">
-               <number>1</number>
+               <number>0</number>
```
MonsterDruide1, line 45: This should probably not be changed, or is there a specific reason you'd want to do this?

NebulaGazer: Makes the GUI limits consistent with the defined field limits:
https://github.com/yuzu-emu/yuzu/blob/master/src/common/settings.h#L464
If the fields are left as is today, you can use the hotkey to bring your speed down to 0%, but then the GUI reports 1%. From a user POV, if you went down two steps (from 100%), you'd expect to be at 0%, not 1%. Then if you went back up two steps, you'd be at 101% when you thought you were at 100%.

german77:
I would prefer if this setting has more resolution at lower speeds than 100%. Take a look at how the volume slider works with hotkeys.

Also what's the goal on being able to reach 0%. I would clamp this value to 5% or 1% if you want to be consistent with the UI

```
void GMainWindow::OnDecreaseVolume() {
    Settings::values.audio_muted = false;
    const auto current_volume = static_cast<s32>(Settings::values.volume.GetValue());
    int step = 5;
    if (current_volume <= 30) {
        step = 2;
    }
    if (current_volume <= 6) {
        step = 1;
    }
    Settings::values.volume.SetValue(std::max(current_volume - step, 0));
    UpdateVolumeUI();
}
```

german77:
Something like this sounds good to me 1, 2, 3, ..., 10, 20, 30, 40, 50, 75, 100, 125, 150, 200, 250, ...

PR #10376, opened 2023-05-18
cmake: apply defaults to all externals

abouvier:
- Implicitly create static library by default for all externals `add_library()`.
- Implicitly skip install rules for all `add_subdirectory()` in the externals directory.
- ~~Remove header files from target sources.~~
- Disable building useless tools/examples in some projects.

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -8,15 +8,21 @@ set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
 # Disable tests in all externals supporting the standard option name
 set(BUILD_TESTING OFF)
 
+# Build only static externals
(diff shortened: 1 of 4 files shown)


liamwhite (approved):

Tachi107:
> Remove header files from target sources.

As far as I know listing headers there is needed for them to be properly showed in Visual Studio, but I don't use that IDE myself so it may be not needed anymore

abouvier:
> As far as I know listing headers there is needed for them to be properly showed in Visual Studio, but I don't use that IDE myself so it may be not needed anymore

Fair enough, I didn't think about that.

Tachi107:
Looks good to me now 👍

liamwhite (approved):

PR #10389, opened 2023-05-20
externals: Update sdl to 2.26.5

german77:
Lately multiple users have issues with controllers. Bump sdl2 to latest available. Depends on https://github.com/yuzu-emu/ext-windows-bin/pull/36


zeltermann:
Should probably try 2.28.0 when it comes out, which is soon as mentioned by this comment: https://github.com/libsdl-org/SDL/issues/7809#issuecomment-1587870847

german77:
We can update to 2.26.8 as soon is available. We have no real need to use an specific version.

FearlessTobi:
Causes https://github.com/yuzu-emu/yuzu/issues/10786 ?

eUnisk:
> Should probably try 2.28.0 when it comes out, which is soon as mentioned by this comment: [libsdl-org/SDL#7809 (comment)](https://github.com/libsdl-org/SDL/issues/7809#issuecomment-1587870847)

Agree to use 2.28.0 as "This is the last scheduled release of SDL 2.0."

Squall-Leonhart:
> Causes #10786 ?

Yes, its fixed in the 2.28RC

eUnisk:
> Causes #10786 ?

I had a similar issue, I lodged a case and it's been fixed in the 2.28 RC.

PR #10397, opened 2023-05-21
time: Auto detect system time zone

lat9nq:
Given the "auto" time zone setting, some parts of yuzu would offset the time to the system time zone when asked for the time, while others would just default to GMT. This uses C++20's tzdb to find the system time zone string and load it from the system archives, trying to make yuzu consistent with itself.

This also removes GetExternalRtcTime. It was giving the TimeManager a time zone offset time, which is inaccurate, and deleting only the offset part of the expression made the entire function redundant.


lat9nq:
tzdb apparently is limited to libstdc++ 13.

KanameMadoka:
I think this change is causing my system to break. Apparently  libstdc++ 13 is not installed by default for a normal windows user (and probably not required by any previous code)

german77:
Do you have the name of the dll. This shouldn't need anything more than stuff that already exist. Only windows requirement is to have https://aka.ms/vs/17/release/vc_redist.x64.exe installed

KanameMadoka:
> Do you have the name of the dll. This shouldn't need anything more than stuff that already exist. Only windows requirement is to have https://aka.ms/vs/17/release/vc_redist.x64.exe installed

Very interesting, installing that certainly resolved the issue. Thanks!

All I have is this error message. 
yuzu error

liamwhite (approved):
Code is fine but I'm slightly concerned about the potential for increased support burden due to users having outdated vcredist.

lat9nq:
This is causing installs that are missing the Switch's system archives to not report time zone correctly. I'm pondering how to handle this situation.

igoticecream:
Since this change, TOTK save date and time are wrong, I've tried default, auto and CET (current) but no luck

src/core/file_sys/system_archive/time_zone_binary.cpp
Tachi107, line 1: How was this file generated?

lat9nq: I dumped the system firmware, found the archive that contained the time zone data, and then converted the basic time zone data to the hex digits that you see with a small C program that I wrote. The data is essentially just the latter half of that generated from building the sources from here: https://www.iana.org/time-zones, but also with a few more bytes at the end stripped out. Their license specifies that it is public domain, if that is the concern.

Tachi107: Makes sense, thanks for the explanation!

PR #10399, opened 2023-05-21
Cookies

behunin:
This is for future use with the new compatibility reporting system.


src/web_service/cookie.h
```diff
@@ -0,0 +1,98 @@
...
+class Cookie {
+public:
+    explicit Cookie(std::string& name, std::string& value);
+    Cookie() = default;
+    ~Cookie() = default;
+
+    enum RawForm { NameAndValueOnly, Full };
+    enum class SameSite { Default, None, Lax, Strict };
+
+    static inline bool isTerminator(char c) {
```
lioncash, line 23:
```suggestion
    static constexpr bool isTerminator(char c) {
```

Ditto for the other ones

src/web_service/cookie.h
```diff
@@ -0,0 +1,98 @@
...
+
+    static inline bool isValueSeparator(char c) {
+        return isTerminator(c) || c == ';';
+    }
+
+    static inline bool isWS(char c) {
+        return c == ' ' || c == '\t' || c == '\r' || c == '\n';
+    }
+
+    static size_t advanceTo(const std::string& text, size_t from) {
```
lioncash, line 35: Our naming conventions use CamelCase, not camelCase for functions (ditto for all other occurrences). variables should also be snake_case

lioncash, line 35:
```suggestion
    static size_t advanceTo(std::string_view text, size_t from) {
```

src/web_service/cookie.h
```diff
@@ -0,0 +1,98 @@
...
+        // RFC 2616 defines linear whitespace as:
+        //  LWS = [CRLF] 1*( SP | HT )
+        // We ignore the fact that CRLF must come as a pair at this point
+        // It's an invalid HTTP header if that happens.
+        while (from < text.size()) {
+            if (isWS(text.at(from)))
```
lioncash, line 41: `.at` doesn't need to be used if the bounds of the string is already known, just use normal indexing with the subscript operator

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1051 @@
...
+#include <sstream>
+
+#include "common/string_util.h"
+#include "cookie.h"
+
+namespace WebService {
+
+#define TM_YEAR_BASE 1900
+
+Cookie::Cookie(std::string& name, std::string& value) : m_name{name}, m_value{value} {}
```
lioncash, line 13: Pass by value and move

```suggestion
Cookie::Cookie(std::string name, std::string value) : m_name{std::move(name)}, m_value{std::move(value)} {}
```

this allows the caller to either copy if it needs to, or move into the cookie, completely eliminating copies altogether, making the interface way more flexible

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1051 @@
...
+
+void Cookie::setExpirationDate(const std::tm& date) {
+    m_expirationDate = date;
+}
+
+std::string Cookie::domain() const {
+    return m_domain;
+}
+
+void Cookie::setDomain(const std::string& domain_) {
```
lioncash, line 51:
```suggestion
void Cookie::setDomain(std::string domain_) {
```

and std::move into the member variable

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1051 @@
...
+    m_domain = domain_;
+}
+
+std::string Cookie::path() const {
+    return m_path;
+}
+
+void Cookie::setPath(const std::string& path_) {
```
lioncash, line 59: ditto

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1051 @@
...
+    m_path = path_;
+}
+
+std::string Cookie::name() const {
+    return m_name;
+}
+
+void Cookie::setName(const std::string& cookieName) {
```
lioncash, line 67: ditto

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1051 @@
...
+    m_name = cookieName;
+}
+
+std::string Cookie::value() const {
+    return m_value;
+}
+
+void Cookie::setValue(const std::string& value_) {
```
lioncash, line 75: ditto

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1051 @@
...
+    m_value = value_;
+}
+
+Cookie::SameSite Cookie::sameSiteFromString(std::string str) {
```
lioncash, line 79:
```suggestion
Cookie::SameSite Cookie::sameSiteFromString(std::string_view str) {
```

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1051 @@
...
+ * EST/EDT: Eastern Standard/Daylight Time
+ * AST: Atlantic Standard Time
+ * NST: Newfoundland Standard Time
+ * GMT: Greenwich Mean Time
+ * BST: British Summer Time
+ * MET: Middle Europe Time
+ * EET: Eastern Europe Time
+ * JST: Japan Standard Time
+ */
+typedef enum {
```
lioncash, line 120:
```suggestion
enum TimeToken {
```

Is sufficient, there's no need for the typedef

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1051 @@
...
+    return true;
+}
+
+std::pair<std::string, std::string> Cookie::nextPair(const std::string& text, size_t& position,
+                                                     bool isNameValue) {
+    // format is one of:
+    //    (1)  name
+    //    (2)  name = value
+    //    (3)  name = quoted-string
+    const int length = text.size();
```
lioncash, line 847:
```suggestion
    const size_t length = text.size();
```

behunin:
Thanks for the review @lioncash
I will make your changes when Azure updates msvc to 17.6, hopefully that takes care of the missing members for `std::tm`

src/web_service/cookie.h
```diff
@@ -0,0 +1,98 @@
...
+
+namespace WebService {
+
+class Cookie {
+public:
+    explicit Cookie(std::string& name, std::string& value);
+    Cookie() = default;
+    ~Cookie() = default;
+
+    enum RawForm { NameAndValueOnly, Full };
```
lioncash, line 20:
```suggestion
    enum class RawForm { NameAndValueOnly, Full };
```

Realized I forgot to make a comment for this one

FearlessTobi:
Why exactly do we need this?
Our web services are all using Json Web Tokens for authentication and it works pretty well.

behunin:
> Why exactly do we need this? Our web services are all using Json Web Tokens for authentication and it works pretty well.

While JWT do work well for authorization, they are kind of costly to create/parse because of the size of data and the hashing/encoding that is done to them.

Cookies involve a lot less data to parse and are also the default authentication method for the new compatibility system that we are working on.

FearlessTobi:
You lost me a bit here...
Currently the compat list submission server is already using JWTs and seems to handle the load of parsing JWTs just fine.
From what I've seen on our internal monitoring, there's not a huge load of testcases coming in anyway.

Therefore I don't understand why we need to add another form of authentication when we already have an auth system that works just fine.

behunin:
The system you are referring to is a different system than the one I am talking about.
New stuff in the works.

PR #10411, opened 2023-05-22
Limit device access memory for integrated devices to 4 GB

### LINKED ISSUE #10407: TOTK / System Crash on Steam Deck after memory filled up

### Affected Build(s)

Early Access 1601

### Description of Issue

Steam Deck user here. While playing around with a build containing #10398, I noticed that the VRAM seems to either improperly being reported back to the garbage collector or the VRAM / texture garbage collector is being run too late.  If the VRAM exceeds 4 or 4.1 GB, it will eventually grow farther above 4.5 GB with BC3 or even 5 GB with uncompressed. The RAM is also filled up above 13 GB, which will lead to system crash of the deck (i need to force reboot it then)
Using Game version 1.1.1

### Expected Behavior

memory should not fill entirely

### Reproduction Steps

I have been testing this by jumping down into the depths from the surface (dying at the bottom) and reloading the save a few times. I also ran on foot between the outpost near at the castle and the hole to the depths south of it.

### Log File

I just tested by observation of the steam deck performance overlay (for now)

### System Configuration

CPU: AMD
GPU/Driver: AMD Radeon VanGogh with Vulkan RADV driver
RAM: 16 GB
OS: Steam OS

scorpion81:
Hardly limiting the device access memory to 4 GB for integrated vulkan devices here. Vulkan seems to detect the Steam Deck's APU as integrated device. This limit works for the Steam Deck in order not to go above 4 GB VRAM usage any more (above this value the likelihood to crash when the RAM exceeds 12 GB as well raises).

But there will be perhaps a detection mechanism necessary for detecting the real memory limit for integrated vulkan devices. Those likely might have small limits anyway, but what about integrated GPUs on machines with > 16 GB RAM, aka larger amounts ?

This might fix #10407 and similar issue reports.

Tested on my steam deck and VRAM usage stayed below 4 GB, even when jumping down the gloomy depths. But it can become  a bit slow when swapping starts, tho.

Note, this is just a simple idea on how to approach this. There may be better solutions around.

### DIFF
--- a/src/video_core/vulkan_common/vulkan_device.cpp
+++ b/src/video_core/vulkan_common/vulkan_device.cpp
@@ -1034,7 +1034,7 @@ void Device::CollectPhysicalMemoryInfo() {
     }
     const s64 available_memory = static_cast<s64>(device_access_memory - device_initial_usage);
     device_access_memory = static_cast<u64>(std::max<s64>(
-        std::min<s64>(available_memory - 8_GiB, 4_GiB), static_cast<s64>(local_memory)));
+        std::min<s64>(available_memory - 8_GiB, 4_GiB), std::min<s64>(local_memory, 4_GiB)));
 }
 
 void Device::CollectToolingInfo() {


liamwhite (approved):

Squall-Leonhart:
Intel has some potential issues with integrated vram detection given they expose both fixed and dvmt controls, and users with such systems might be configuring stuff all wrong, these systems also typically have a max DVMT of 2GB total.

AL1ghtm3ter:
Hiya, a very noob question is, how do I use the files in Commits? Do I need to change any code or anything else?

AL1ghtm3ter:
> > 
> 
> > > 嗨，一个非常菜鸟的问题是，我如何使用 Commits 中的文件？我需要更改任何代码或其他任何内容吗？
> > 
> > 
> > 就这里关闭的 pullrequest 而言，它已经合并到 master 中，并且应该在合并后的每个构建中都可用。代码现在也是 master 分支的一部分。
> 
> But I have updated to the latest version and one problem is that I am still experiencing flashbacks, I have checked the performance monitor and found that it is still due to lack of memory. And I've tried every different block of VRAM, from 256m to 4G, but still no significant improvement, so I'm very confused

CryoUtilities

AL1ghtm3ter:
> > 
> 
> > > 嗨，一个非常菜鸟的问题是，我如何使用 Commits 中的文件？我需要更改任何代码或其他任何内容吗？
> > 
> > 
> > 就这里关闭的 pullrequest 而言，它已经合并到 master 中，并且应该在合并后的每个构建中都可用。代码现在也是 master 分支的一部分。
> 
> But I have updated to the latest version and one problem is that I am still experiencing flashbacks, I have checked the performance monitor and found that it is still due to lack of memory. And I've tried every different block of VRAM, from 256m to 4G, but still no significant improvement, so I'm very confused

Hiya, I think I've solved the problem by using CryoUtilities and adjusting the VRAM to 4Gb and I haven't had any flashbacks, at least as of now.

goldenx86:
Hi @scorpion81 can I mention your work in the progress report?

scorpion81:
Hey @goldenx86, yes you can mention it :+1:

PR #10420, opened 2023-05-23
input_common: Redesign mouse->stick panning for better responsiveness

jtoomim:
Previously, the mouse processing code had some issues that reduced mouse control quality. No matter how you configured it or how you used the mouse, it was impossible to be able to make both fine inputs and maximum inputs to the control stick with the same configuration, as inputs were clamped to the range, and *outputs* were multiplied by the sensitivity setting. It was also impossible to make brief inputs, as two built-in exponential moving average filters in the mouse driver caused any mouse inputs, no matter how brief, to continue to produce stick outputs for several hundred milliseconds before dissipating.

This commit is an attempt to address those issues. With this commit, an entirely different model of processing inputs is used. Instead of applying the sensitivity to the stick output, this commit applies the sensitivity at the beginning of processing to the mouse input. Next, inputs from the mouse are accumulated into a buffer variable. When the stick is updated, stick output is clamped to the range of the stick, and that value is subtracted from the input buffer. Any *excess* motion still remaining in the buffer gets decayed exponentially (similar to one of the exponential moving averages before). This gives a good compromise between responsiveness (brief mouse movement causes a brief stick response) and inertia (large mouse movement which exceeds the stick's instantaneous dynamic range causes stick movement that persists for a short time afterwards, mimicking inertia). Finally, to improve the linearity of the response, the stick motion is square-rooted to amplify small movements.

This commit changes the semantics of the sensitivity option. It's no longer a percent, and "100(%)" no longer means anything special, so the input configuration UI has been changed to allow values up to 10,000 instead of being limited to 100. The changes to the filtration also affect the sensitivity. There are some magic constants in the code to try to keep the sensitivity about the same with a given value, but it's not exact, and users are encouraged to play around with it again on the new code.

These changes have only been tested in TotK, but they improve aiming with bows, the scope, and hand abilities dramatically. It also improves the ability to carefully select items in the quick menus.

### DIFF
--- a/src/input_common/drivers/mouse.cpp
+++ b/src/input_common/drivers/mouse.cpp
@@ -88,13 +88,26 @@ void Mouse::UpdateStickInput() {
         return;
     }
 
-    const float sensitivity =
-        Settings::values.mouse_panning_sensitivity.GetValue() * default_stick_sensitivity;
+    auto mouse_change = last_mouse_change;
+    auto move_distance = mouse_change.Length();
 
-    // Slow movement by 4%
-    last_mouse_change *= 0.96f;
-    SetAxis(identifier, mouse_axis_x, last_mouse_change.x * sensitivity);
-    SetAxis(identifier, mouse_axis_y, -last_mouse_change.y * sensitivity);
+    // Perform at most 1 unit of buffered mouse movement
+    if (move_distance > 1.0f) {
+        mouse_change *= 1.0f / move_distance;
+        move_distance = mouse_change.Length();
+    }
+
+    last_mouse_change -= mouse_change;
+
+    // Decay remaining buffered movement by 5%
+    last_mouse_change *= 0.95f;
+
+    // Stick response is nonlinear, and is not sensitive enough to fine changes
+    const auto sqrt_distance = sqrt(move_distance);
+    mouse_change *= sqrt_distance / move_distance;
+
+    SetAxis(identifier, mouse_axis_x, mouse_change.x);
+    SetAxis(identifier, mouse_axis_y, -mouse_change.y);
 }
 
 void Mouse::UpdateMotionInput() {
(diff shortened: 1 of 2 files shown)


jtoomim:
Note: I'll be traveling starting tomorrow and will be away from my desktop computer for the next couple months, so I won't be able to make any other changes or amend this PR any time soon. If any changes are needed, someone else will have to do them.

german77:
This change feels pretty bad. It might work for TOTK, but we need to take into account all other Switch games. In general, it just needs better tuning and reintroducing the average between inputs. The following points need to be taken care of:

Default sensitivity should be aimed so that 50% feels normal in both non-panning and panning modes. This adjustment is made with "default_stick_sensitivity," which you got rid of. That's why you suddenly needed to increase sensitivity to 10000.

The deadzone is a bit tricky. We want slight movements to be detected, and not all games have the same internal deadzone. Some have a lower deadzone, while others have a slightly higher one. In any case, we need to ensure this works with many games, such as XC3 (high deadzone) and LM3 (low deadzone). Three units seem to be the best spot, but we can lower it to 2.5 and still be fine-ish. One unit is practically no deadzone, as the minimum mouse movement is one pixel.

We must avoid jitter at all costs. Hence why I sacrificed a bit of responsiveness by averaging inputs to reduce jitter. Not all games make this noticeable, but games like Doom make a huge deal out of it. The jitter is so noticeable that it makes the mouse input feel like garbage.

The maximum range can be increased. Initially, I kept it at 8 units since larger values would take a while to jump to the other side with the average value. This pull request doesn't have an average, so it's fine with this removed, but then we encounter the jitter issue again.

I took the liberty of taking two small videos with the same level of sensitivity 100% to compare the differences
Mainline ->
https://github.com/yuzu-emu/yuzu/assets/5944268/e55a74ec-d28e-4efd-8d8d-986d9fb8023c

PR 10420 ->
https://github.com/yuzu-emu/yuzu/assets/5944268/ae8b2da3-58c2-4552-996e-f86b237ed1de

german77:
Superseded by #10495

PR #10435, opened 2023-05-24
Texture cache: revert wrong acceleration assumption

FernandoS27:
This fixes the regressions in UE4 games.


TOTK-LOVER:
The performance of my computer's CPU is too weak,I need  optimization about CPU,plz

TOTK-LOVER:
> The performance of my computer's CPU is too weak,I need optimization about CPU,plz

In the TOTK

liamwhite (approved):

PR #10465, opened 2023-05-27
input_common: set input thread delay of joycon to 15ms

colorfulshark:
According to my test, the time during between two `IsPayloadCorrect == true` would be 15ms in most of time even though we pull the data every 2ms. This causes a lot useless `status == 0` condition, which means no data received and wastes CPU cycles.

It turns out 15ms is already enough for imperceptible operation and somehow reduces lag on my platform (i5-7300hq + gtx1060 + CRS8510).

Consider that Switch console also requests data every 15ms, I believe it would be fine for us to do so.

### DIFF
--- a/src/input_common/helpers/joycon_driver.cpp
+++ b/src/input_common/helpers/joycon_driver.cpp
@@ -128,8 +128,9 @@ void JoyconDriver::InputThread(std::stop_token stop_token) {
     Common::SetCurrentThreadName("JoyconInput");
     input_thread_running = true;
 
-    // Max update rate is 5ms, ensure we are always able to read a bit faster
-    constexpr int ThreadDelay = 2;
+    // Max update rate is 5ms, but in normal the console asks Joy-Con for an update
+    // every 15ms, this can help reducing CPU pressure on some weak platform
+    constexpr int ThreadDelay = 15;
     std::vector<u8> buffer(MaxBufferSize);
 
     while (!stop_token.stop_requested()) {


german77:
While this is true the joycon are prone to huge delays. this makes the packets stack and if you don't read any faster you will have seconds of input delay

german77:
Also 15ms only applies to input reports, command replies can be much faster

colorfulshark:
Hi @german77 , thanks for replying!

Do you mean SDL has a buffer which will queue the packets? But from what I understand, only if we call `SDL_hid_read_timeout` will SDL ask data from the device, So even though the timeout is enlarged, the function call will return early if the data is realy ready, or the input thread will just wait (sleep maybe?).

So the timeout seems to be an insurance to make sure the read call will return if something gets stuck. Please correct me if I'm wrong.

german77:
Not SDL. We are using hidapi. An yeah it has a queue since I only receive full packets.

german77:
The timeout defines how long should we wait for data. If a packet is available it will read x bytes. If the buffer full returns the current data. If it's not waits for the next package.

You can for example read fewer bytes than the input report. Yet you still receive the data in the correct order. In a few words the first byte is 0x21 which is the input report mode

german77:
I want to close this PR. So we can take the middle ground which is 5ms. That should lower the polling rate by 2.5 times. If this does impact you in performance that amount should be noticiable while keeping a polling rate fast enough to keep the package queue clean.

colorfulshark:
Hi @german77 ,

Agree to you, I have also tested 5ms pull rate on my side, and that still reduced loop speed a lot, so it should be reasonable for common platform. Besides, I noticed BetterJoy also uses 5ms delay. Anyway, this PR could be closed.

PR #10471, opened 2023-05-27
Wait indefinitely when audio buffer queue is too big

Kelebek1:
This was a change sugegsted by ByLaws to stop some audio desync when the emu core gets frozen, e.g by IO on load screens. He thinks it can help with audio crackling by limiting the queue size. (copium)


liamwhite (approved):
Completely fixes the audio crackling on my end

PR #10476, opened 2023-05-28
OpenGL: Make use of persistent buffer maps in buffer cache

ameerj:
Persistent buffer maps were already used by the texture cache, this extends their usage for the buffer cache.

This substantially improves performance, nearly doubling the performance i was seeing in TotK and Pokemon Scarlet.

Before |  After
pkmn-before | pkmn-after
totk-before | totk-after

In my testing, using the memory maps for uploads was slower than the existing `ImmediateUpload` path, so the memory map usage is limited to downloads for the time being.
This also allows for implementing async buffer downloads on OpenGL, which can potentially improve performance even further.

Further testing to ensure stability is needed.


gidoly:
Niceeeee, Can we get vulkan optimization too?

gidoly:
after few test lgtm no apparent regression either

Curlypla:
Hmmm, my game is stuck on a black screen after loading, decent CPU usage at 0% while opengl works fine on my EA version.**

ameerj:
@Curlypla can you please provide a log with your system specs and the game you're trying to run?

Curlypla:
> @Curlypla can you please provide a log with your system specs and the game you're trying to run?

My bad, I was using the artifact's windows.zip instead of the .exe, which works perfectly.

scorpion81:
Hello, 
been testing this with an own dev build on steam deck. 
Switched to OpenGL - Spirv - High - Handheld, also tested Docked. After ~5 - 10 minutes roaming around the first lookout tower, a crash happens. 
And things run rather slow, compared to Vulkan. Ran it in desktop mode, barely 20 fps. 
But note, could be that the game mode reduced TDP still was in effect here as well. 
Will test again with unlimited TDP -> with Vulkan it can reach ~30fps then, but at the "cost" of 80+ degrees Celsius on the CPU and a loud fan noise. 
With 20 fps it will be with a reduced TDP about 60-65 C, but that is another story.

Anyways, I have been testing for 3 times. First run, there seem to have been leftovers from my test before with GLSL. 
See 
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/11585461/yuzu_log.txt)
Most interesting is the end with the message: 
```
[  29.930419] Shader <Warning> shader_recompiler/ir_opt/global_memory_to_storage_buffer_pass.cpp:CollectStorageBuffers:392: Storage buffer failed to track, using global memory fallbacks
```
Another run after clearing the OpenGL shader cache yielded:
[yuzu_log_opengl_spirv2.txt](https://github.com/yuzu-emu/yuzu/files/11585471/yuzu_log_opengl_spirv2.txt)
Interesting lines:
```
[ 460.862306] Debug <Critical> video_core/rasterizer_accelerated.cpp:operator():34: Assertion Failed!
Count may underflow!
[ 460.892511] Debug <Critical> video_core/rasterizer_accelerated.cpp:operator():32: Assertion Failed!
Count may overflow! 
``` 
and
```
[ 599.584699] HW.Memory <Error> core/memory.cpp:operator():237: Unmapped ReadBlock @ 0x00000010D7AC6000 (start address = 0x00000010D7AB9000, size = 69632)
[ 599.584707] HW.Memory <Error> core/memory.cpp:operator():237: Unmapped ReadBlock @ 0x00000010D7AC7000 (start address = 0x00000010D7AB9000, size = 69632)
[ 599.584709] HW.Memory <Error> core/memory.cpp:operator():237: Unmapped ReadBlock @ 0x00000010D7AC8000 (start address = 0x00000010D7AB9000, size = 69632)
[ 599.584712] HW.Memory <Error> core/memory.cpp:operator():237: Unmapped ReadBlock @ 0x00000010D7AC9000 (start address = 0x00000010D7AB9000, size = 69632)
```

at the end. Seems the device runs either out of VRAM or RAM. I have set the UMA buffer size to 4 GB btw.
Is in the OpenGL version of the renderer a similar memory / garbage collector limit calculation like with Vulkan ?

See also https://github.com/yuzu-emu/yuzu/pull/10411 .
There I just took Vulkan in consideration, when attempting to limit the VRAM usage to 4 GB on integrated devices.

ameerj:
@scorpion81 It sounds like there may be a shader crash on SPIR-V.  How does this compare with mainline? Do you experience crashes there after some time?

GLSL is slower at building shaders but will speed up after caching the built shaders and shouldn't experience any shader building crashes

scorpion81:
Ok, tested with mainline and spir-v now, also crashes after a couple of minutes. I have been watching the memory usage tho, RAM went over 11.5 GB and VRAM over 5.5+, swap is at 1 GB, UMA buffer at 4 GB. The whole device crashed and rebooted. Most likely out of memory.
Log: 
[yuzu_log_opengl_spirv_mainline.txt](https://github.com/yuzu-emu/yuzu/files/11585583/yuzu_log_opengl_spirv_mainline.txt)

grimkor:
Not sure on the specifics of this PR but I'm experiencing a drop in performance on Vulkan if I add this PR to my build. I'll attach logs and screenshots below:

[yuzu_log-main.txt](https://github.com/yuzu-emu/yuzu/files/11617893/yuzu_log-main.txt)
[yuzu_log-with-pr.txt](https://github.com/yuzu-emu/yuzu/files/11617894/yuzu_log-with-pr.txt)
main
with-pr

liamwhite:
@grimkor This PR didn't change anything for Vulkan as far as I can tell. You're sure it's this one?

liamwhite (approved):

PR #10482, opened 2023-05-28
yuzu: Disable game list while game is running

german77:
Qt solution to Qt problem. Turns out the game list is still active in background so you can boot a game while a game is running. The easiest way to go with this is to simply disable the game list so no actions can be triggered. This prevents softlocks at booting games

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -1798,6 +1798,7 @@ void GMainWindow::BootGame(const QString& filename, u64 program_id, std::size_t 
     }
 
     system->SetShuttingDown(false);
+    game_list->setDisabled(true);
 
     // Create and start the emulation thread
     emu_thread = std::make_unique<EmuThread>(*system);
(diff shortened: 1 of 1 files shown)


src/yuzu/main.cpp
```diff
@@ -1757,6 +1757,7 @@ void GMainWindow::BootGame(const QString& filename, u64 program_id, std::size_t
     // Save configurations
     UpdateUISettings();
     game_list->SaveInterfaceLayout();
+    game_list->setDisabled(true);
```
Morph1984, line 1760: This should only be disabled when the game is successfully loaded, note the early returns below

src/yuzu/main.cpp
```diff
@@ -1993,6 +1994,9 @@ void GMainWindow::OnEmulationStopped() {
     // When closing the game, destroy the GLWindow to clear the context after the game is closed
     render_window->ReleaseRenderTarget();
 
+    // Enable game list
+    game_list->setDisabled(false);
```
Morph1984, line 1998: setEnabled

liamwhite (approved):

PR #10486, opened 2023-05-29
yuzu-qt: Load Vulkan device info at startup

lat9nq:
Loading it when the configuration opens now incurs a noticeable delay (as a result of #10125). We also don't need to rediscover the same data repeatedly each time the configuration opens.

Moves Vulkan device info discovery to yuzu's startup as opposed to the configure_graphics constructor.

### DIFF
--- a/src/yuzu/CMakeLists.txt
+++ b/src/yuzu/CMakeLists.txt
@@ -210,6 +210,8 @@ add_executable(yuzu
     util/url_request_interceptor.h
     util/util.cpp
     util/util.h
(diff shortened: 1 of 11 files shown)


lat9nq:
> Won't this cause a noticeable delay at startup now though?

Internally when I first wrote this, I offered to put this on its own thread. It didn't seem worth it at the time since the delay isn't more than a second anyway. Let me know if you have a different opinion/experience on this.

liamwhite (approved):

PR #10492, opened 2023-05-29
externals: update mbedtls to 3.4.0

abouvier:
Follow-up of https://github.com/yuzu-emu/yuzu/pull/7044.
The only two modifications added in https://github.com/yuzu-emu/mbedtls are enabled by default in Mbed-TLS version 3. So you might as well use the latest version available in repos.


Tachi107:
I'd wait for this until the MbedTLS devs release an LTS from the 3.x branch, as e.g. in Debian we only package LTS releases.

In any case, if I recall correctly the issue with using upstream MbedTLS is that Morph eventually wants to switch yuzu to using a custom MbedTLS fork with some performance improvements (see the [aes-ni](https://github.com/yuzu-emu/mbedtls/commits/mbedtls-2.28-aesni-shani) branch for an example), so I'm sure if this will ever get merged

comex:
I have an upcoming PR that uses mbedtls to implement the SSL service.  While the current submodule version works okay, newer versions of mbedtls have two features which are useful in this context:

- Improved support for exporting session keys - which can be used to implement `SSLKEYLOGFILE`, which can be used to [allow Wireshark to decrypt packet captures](https://wiki.wireshark.org/TLS).
- Support for TLS 1.3.  This is a pretty big deal; it's just not relevant to Yuzu's existing use of mbedtls, since it doesn't actually use the TLS part.
 
That said, those features aren't strictly needed.  And there is also the option of having the SSL backend be something else entirely, keeping mbedtls's use limited to low-level crypto.  I already have plans to use Schannel on Windows.

comex:
…Scratch that.  I decided that with the… [eclectic variety](https://stackoverflow.com/a/60386677) among Linux distros of where exactly they put their trusted certs, combined with the fact that it's supposed to be overridable with two independent environment variables that do slightly different things (`SSL_CERT_DIR ` and `SSL_CERT_FILE`), it would be more reliable to just use the system OpenSSL.  I now have that implemented along with Schannel.

Tachi107:
Would it make sense to use OpenSSL for both TLS and crypto?

comex:
> Would it make sense to use OpenSSL for both TLS and crypto?

Not really, since OpenSSL is only one of multiple backends depending on the platform.  If you're wondering why Yuzu can't just use OpenSSL on every platform, [my PR](https://github.com/yuzu-emu/yuzu/pull/10912) is now up and the description addresses this.

german77:
Does this PR supersedes #7044 ?

abouvier:
Well yes, it's the same goal but with a more recent mbedtls version.

Issue #10481, opened 2023-05-28
Enable mouse panning option not reading from Settings INI


jcp219:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

Mainline 1448

### Description of Issue

The 'Enable mouse panning' option is not persisted on exit/restart of Yuzu. From reviewing the code, it is not reading from or writing to the settings ini file.

[src/yuzu/configuration/config.cpp Line 442](https://github.com/yuzu-emu/yuzu/blob/d6db422098ad12d95396a23f60f005893eb9b720/src/yuzu/configuration/config.cpp#LL442C5-L442C44)

[Should be writting here at src/yuzu/configuration/config.cpp Line 1159](https://github.com/yuzu-emu/yuzu/blob/d6db422098ad12d95396a23f60f005893eb9b720/src/yuzu/configuration/config.cpp#L1159)

### Expected Behavior

Enabling Mouse Panning should be persisted between sessions by writing it's value to the config ini and reading from it as well.

### Reproduction Steps

1. Open Yuzu
2. Navigate to Emuluation > Configure...
3. Navigate to Controls > Advanced
4. Check the checkbox for 'Enable mouse panning'
5. Press OK
6. Exit/Reopen Yuzu
7. 'Enable mouse panning' is now disabled.

### Log File

N/A

### System Configuration

OS: Windows 11 22H2

german77:
This is intentional. Mouse panning locks your mouse to the center of your screen. We had issues where users enabled this setting accidentally and couldn't disable it.

Leading to many support issues, at the end is a non persistent setting.

jcp219:
Thanks for the reply.  Perhaps the setting could be read from the .ini but not written to it? This would allow advanced users to force the setting enabled by setting mouse_panning=1 in the qt-config.ini file and not have to enable it every time they open Yuzu, but still prevent accidental activations by inexperienced users. I can closed this issue put this in as a feature request if that would be better.

ZReC:
there's a hotkey to toggle mouse panning, extremely useful.

Martchus:
>  Perhaps the setting could be read from the .ini but not written to it? This would allow advanced users to force the setting enabled by setting mouse_panning=1 in the qt-config.ini file 

That sounds useful to me.

> there's a hotkey to toggle mouse panning, extremely useful.

That sounds even more useful, indeed. Has that hotkey already been implemented? If yes, what is it?

german77:
[...]

PR #10495, opened 2023-05-29
input_common: Redesign mouse panning

bm01:
I wanted to give a try at fixing the issues raised by #10420 (whose author is away for some months and can't update the PR), and I ended up with a somewhat more conservative solution.

Attempts to fix the following issues:

1. Sensitivity setting affecting the stick's range, making either the max value unreachable or min value too high if not set to 50%.
Fixed by applying the sensitivity on the mouse input alone, before any clamping.

2. Very small movements being impossible to perform.
Fixed by removing the line that sets the mouse event's length to a minimum of 3. Testing shows that it is rarely higher than 1 even with a low mouse polling rate, so setting it to 3 meant some details were lost.

3. Stick "resistance" feeling inconsistent.
This is more subjective. The buffer decay is applied in both `Move` (9% per mouse event) and `UpdateStickInput` (4% per update tick). Comment says "Average mouse movements" but this is really just a double exponential decay that depends not just on the buffer's current value and how much time has passed (for example I think a high polling rate mouse meant a stronger decay).
Fixed by applying it in `UpdateStickInput` only.

4. Slow stick recentering time causing slippery (high end inertia) movements.
`mouse_panning_timeout` was used to recenter a stick after a certain amount of update with no user input, but a high value resulted in slippery movements and a low one tended to make the stick recenter too easily during slow movements (killing any build-up). This can be fixed by changing the decay function and make the recenter time from a high length must faster without affecting the time it takes to recenter from a short length too much. The following graph shows the decay (Y) applied to `last_mouse_change` on each update based on the stick length (X). (Update: was slightly tweaked to be less aggressive)
chart

5. Lack of configurability.
Only the sensitivity could be configured, and globally.
The stick's decay strength and minimum amount can now be configured to fit the user preference. Another setting was added to help counteract against a game's built-in deadzone. This last setting and sensitivity were split for both X and Y axis. Because the user may want to have different settings between games, the configure button was moved from the advanced widget to the first player input widget, and values are saved in the profiles.

6. https://github.com/yuzu-emu/yuzu/issues/10481
Implemented what was suggested. The enable setting is no longer saved but is still read (effectively a read-only setting).

This still isn't perfect by any mean:
- Qt's mouse event appears to lack granularity, but I don't think we have much choice.
- The update interval is somewhat slow (10ms) so buffer decay is never going to feel very smooth.
- Buffer decay is tied to the update interval so it can't be changed without making it either slower or faster.
- Probably as a consequence of above, slippery movement currently cannot be completely prevented without introducing various issues.

### DIFF
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -524,9 +524,16 @@ struct Values {
     Setting<bool> tas_loop{false, "tas_loop"};
 
     Setting<bool> mouse_panning{false, "mouse_panning"};
-    Setting<u8, true> mouse_panning_sensitivity{50, 1, 100, "mouse_panning_sensitivity"};
-    Setting<bool> mouse_enabled{false, "mouse_enabled"};
+    Setting<u8, true> mouse_panning_x_sensitivity{50, 1, 100, "mouse_panning_x_sensitivity"};
+    Setting<u8, true> mouse_panning_y_sensitivity{50, 1, 100, "mouse_panning_y_sensitivity"};
+    Setting<u8, true> mouse_panning_deadzone_x_counterweight{
+        0, 0, 100, "mouse_panning_deadzone_x_counterweight"};
+    Setting<u8, true> mouse_panning_deadzone_y_counterweight{
+        0, 0, 100, "mouse_panning_deadzone_y_counterweight"};
+    Setting<u8, true> mouse_panning_decay_strength{22, 0, 100, "mouse_panning_decay_strength"};
+    Setting<u8, true> mouse_panning_min_decay{5, 0, 100, "mouse_panning_min_decay"};
 
+    Setting<bool> mouse_enabled{false, "mouse_enabled"};
     Setting<bool> emulate_analog_keyboard{false, "emulate_analog_keyboard"};
     Setting<bool> keyboard_enabled{false, "keyboard_enabled"};
 
--- a/src/input_common/drivers/mouse.cpp
+++ b/src/input_common/drivers/mouse.cpp
@@ -76,9 +76,6 @@ void Mouse::UpdateThread(std::stop_token stop_token) {
         UpdateStickInput();
         UpdateMotionInput();
 
-        if (mouse_panning_timeout++ > 20) {
-            StopPanning();
-        }
         std::this_thread::sleep_for(std::chrono::milliseconds(update_time));
     }
 }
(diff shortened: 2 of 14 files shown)


german77:
Oh nice. I will take a look soon. From a quick glance did you removed the initial deadzone? 

This part in specific https://github.com/yuzu-emu/yuzu/blob/53fd08b360a393af2e6bfdddc5c549492fcb40bf/src/input_common/drivers/mouse.cpp#L145-L150

bm01:
> From a quick glance did you removed the initial deadzone?

I did. It caused the second issue I mention in the PR. In my personal experience most of the mouse movements triggered that deadzone because the length is almost always inferior to 1, but this could be because I play with 1800 DPI mouse and 1000 Hz polling rate. I tried lowering them and I didn't feel like removing it caused any undesirable effect. Is it really needed?

german77:
I haven't tested yet. But I will provide a few games where the lack of deadzone is an issue just give me a few hours.

german77:
Had a chance to test. Here you can see SMO doesn't react to the mouse at all even when it's moving quite a bit. The sensitivity seems a bit low. I can't quite reach the borders unless I really move fast the mouse.
https://github.com/yuzu-emu/yuzu/assets/5944268/4be213bb-d876-4429-aff2-1a67dc47d8f8

german77:
Much better this time. I will try to poke the code a bit to make the stick go to the center even with a deadzone but it's overall good. 

UI Config sucks. We probably need to make a separate category for mouse config since those two input boxes have no explanation XD.

bm01:
I tried to figure a proper solution to that but it was either recentering the stick after an arbitrarily high amount of time, or take the risk of doing it too early and make small movements harder to perform, due to clearing any buffer build-up too low to actually result in any movement but actually contributing towards the next (if that makes any sense).

I agree about the UI. I'll give a try at improving it. How about a "Configure" button that opens a new window similar to "Motion / Touch"? I was also thinking about separating horizontal and vertical sensitivity and deadzone (assuming I can figure out the maths), and also exposing the two float here that now control stick decay:
https://github.com/yuzu-emu/yuzu/blob/465d7c830c165a47d03d3f3a8dd196bc694c4ba5/src/input_common/drivers/mouse.cpp#L112-L113

bm01:
Here's how it could look:

Screenshot 2023-06-01 064002

Interestingly, users can choose to set a decay strength of 0% and a minimum decay of 5%, and effectively get something that's closer to what the original formula was.

With split sensitivity settings I'm not sure what should be done about this though:
https://github.com/yuzu-emu/yuzu/blob/9e7656a597d4a1942a4db3ae996e321766a69d76/src/input_common/drivers/mouse.cpp#L122-L126

bm01:
Screenshot 2023-06-01 170319

Does it expect `2.0f` here? I didn't get any warning about it in Visual Studio. Anyway I don't want to trigger too many actions :p

german77:
Yes. 2.0f. Don't worry about the actions. they are there to help you make sure nothing breaks.

bm01:
I just noticed the deadzone setting in Controls > Player is affecting mouse panning too, which isn't necessarily a problem but for some reason I ended up with one and my tests were done thinking there was none.

Setting it to 0% allows for even finer controls in TOTK, which appears to have a lower built-in deadzone than whatever I had.

Just a thought but maybe a warning could be added in that new widget if the deadzone isn't set to of 0 and range to 100, because I don't think there's any reason to have them to any other value when mouse panning is enabled. But I couldn't find a way to get those values from `ConfigureMousePanning`.

bm01:
Moved the configure button from the advanced widget to the first player input widget, and made it so that the settings are saved in the profiles.

Added a warning that appears only the right stick deadzone and range aren't set to the default values.

Screenshot 2023-06-03 203226

Updated the PR's original post to reflect those changes since it didn't mention the new settings at all.

bunnei:
@german77 ping to take a look at the recent changes here!

german77 (approved):
Sorry for the delay. It's working great now.

PR #10504, opened 2023-05-30
input_common: rename PAGE_SIZE to avoid conflict

12101111:
See also: https://github.com/yuzu-emu/yuzu/issues/8779


german77 (approved):

Kelebek1:
Should probably be called YUZU_PAGESIZE to match the previous time this was done. Or just include wherever that's defined and use that instead.

PR #10538, opened 2023-06-01
web_service: Add cookie parser

behunin:
This is for future use with the new compatibility reporting system.


src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1052 @@
...
+            result += "; HttpOnly";
+        if (m_same_site != SameSite::Default) {
+            result += "; SameSite=";
+            result += Cookie::SameSiteToString(m_same_site);
+        }
+    }
+    return result;
+}
+
+} // namespace WebService
```
FearlessTobi, line 1052: This should have a newline here

src/web_service/cookie.h
```diff
@@ -0,0 +1,98 @@
...
+    std::string m_path;
+    std::string m_comment;
+    std::string m_name;
+    std::string m_value;
+    SameSite m_same_site = SameSite::Default;
+    bool m_secure = false;
+    bool m_http_only = false;
+};
+
+} // namespace WebService
```
FearlessTobi, line 98: dito.

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1052 @@
...
+    case Cookie::SameSite::Strict:
+        return "Strict";
+    case Cookie::SameSite::Default:
+        break;
+    }
+    return "";
+}
+
+/*
+ * see: https://searchfox.org/mozilla-central/source/nsprpub/pr/src/misc/prtime.c#931
```
FearlessTobi, line 105: The project your are linking here seems to use the MPL license. 
Have you made absolutely sure that it is compatible with ours?

FearlessTobi (changes requested):
I think there also were some unadressed concerns from the original PR.

src/web_service/cookie.cpp
behunin: Yes
Mozilla is very chill about this
https://www.mozilla.org/en-US/MPL/2.0/combining-mpl-and-gpl/

src/web_service/cookie.cpp
```diff
@@ -0,0 +1,1052 @@
...
+
+    result->tm_isdst = dst_offset * 60;
+    result->tm_gmtoff = zone_offset * 60;
+    result->tm_zone = zone_name;
+
+    return true;
+}
+
+std::pair<std::string, std::string> Cookie::NextPair(const std::string& text, size_t& position,
+                                                     bool isNameValue) {
```
lioncash, line 843:
```suggestion
                                                     bool is_name_value) {
```

Ditto for other similarly camelCased names below
```

File: src/web_service/cookie.h [hunk 5]
```diff
@@ -0,0 +1,98 @@
...
+    static std::string SameSiteToString(SameSite samesite);
+
+    static std::pair<std::string, std::string> NextPair(const std::string& text, size_t& position,
+                                                        bool isNameValue);
+
+    static std::vector<Cookie> ParseAllCookies(const std::string& cookieString);
+    static std::vector<Cookie> ParseCookie(const std::string& cookieString);
+
+private:
+    std::tm m_expiration_date;
```
lioncash, line 87:
```suggestion
    std::tm m_expiration_date{};
```

So this is always in a deterministic state on construction

src/web_service/cookie.h
```diff
@@ -0,0 +1,98 @@
...
+        // reached the end
+        return text.size();
+    }
+
+    bool IsSecure() const;
+    void SetSecure(bool enable);
+    bool IsHttpOnly() const;
+    void SetHttpOnly(bool enable);
+    SameSite SameSitePolicy() const;
+    void SetSameSitePolicy(SameSite sameSite);
```
lioncash, line 56:
```suggestion
    void SetSameSitePolicy(SameSite same_site);
```

Ditto for other camelCased names below

behunin:
Thank you @lioncash & @FearlessTobi 
17.6 did not bring the necessary changes to msvc but 17.7 is going to bring an update to llvm 16 so I will wait for that to arrive. I assume that will take care of the missing code....

src/web_service/cookie.cpp
FearlessTobi: I am not sure if what we are doing is sufficient here.


> Modified MPL-licensed Files
> 
> A more complicated case occurs when a developer receives a file solely under the terms of the MPL, makes copyrightable changes to that file, and then combines the file with other (L)GPL-licensed files to form a Larger Work. In this situation, the developer may distribute the Larger Work, including the file, under the terms of both the MPL and the (L)GPL, as long as > the developer complies with MPL's conditions on distribution, including the file-level copyleft.

> In order to comply with these terms, great care should be taken to ensure that the new additions made to the MPL-licensed file can be licensed under the MPL, not just the (L)GPL. Additions licensed solely under the (L)GPL would conflict with the terms of the MPL and deprive the upstream MPL project of the ability to reincorporate changes made to the file. For example, copying and pasting content from an (L)GPL-only file into the MPL+(L)GPL file would create a problem. A project may wish to consult with a lawyer to ensure that the terms of the licenses are not violated when new patches are made.

> Once the modifications are made, the top of the file should appear as in the previous Section "[Unmodified MPL-licensed Files - Adding an (L)GPL notice](https://www.mozilla.org/en-US/MPL/2.0/combining-mpl-and-gpl/#unmodified-mpl-licensed-files-adding-an-lgpl-notice)", indicating to recipients that the file is distributed under the terms of both licenses. To the extent new, copyrightable material has been added, and the authors wish to add individual copyright notices, those may be done above the (L)GPL header, alongside the MPL notices (reflecting the fact that such modifications must be under both MPL and (L)GPL).

behunin: Here is my understanding.
I didn't take a whole file, I took one function and made changes that are not copyrightable, the changes I made did not enhance the way the function works, they just adapted from Mozilla's implementation of ``tm`` to ``std::tm``. The only other change from the original is removing localization of the domain so that we don't have to link another library.

Since my changes can never be any use to Firefox or Mozilla, there is no harm done.

PR #10551, opened 2023-06-01
Store button's position in the overlay as a percentage 

PabloG02:
Allows the buttons to reposition themselves if the window is resized.


t895 (approved):
LGTM, works great!

bunnei (approved):

PR #10567, opened 2023-06-02
android: Support for Picture in Picture

AbandonedCart:
Porting this from other implementations I've submitted. 

It's Picture in Picture similar to what's seen when using YouTube. Since it's not actual media, the swapping play / pause button isn't handled by media controls. Having both buttons with conditional activation is the next best thing.

### DIFF
--- a/src/android/app/src/main/AndroidManifest.xml
+++ b/src/android/app/src/main/AndroidManifest.xml
@@ -60,6 +60,8 @@ SPDX-License-Identifier: GPL-3.0-or-later
             android:theme="@style/Theme.Yuzu.Main"
             android:launchMode="singleTop"
             android:screenOrientation="userLandscape"
+            android:supportsPictureInPicture="true"
+            android:configChanges="orientation|screenSize|smallestScreenSize|screenLayout|uiMode"
             android:exported="true">
 
             <intent-filter>
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
@@ -31,7 +31,6 @@ import android.view.View
 import android.view.inputmethod.InputMethodManager
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
-import androidx.core.content.getSystemService
 import androidx.core.view.WindowCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsControllerCompat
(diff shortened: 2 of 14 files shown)


t895:
I really like the change! Just thought I should mention a couple things.

For some reason the emulation surface appears squished
image

And would you mind adding an option in settings to toggle this option?

AbandonedCart:
Shouldn't be a problem as soon as I find a way to actually run what I'm writing :/

AbandonedCart:
What would you put it under? It's not really an "emulator" setting. General?

t895:
General should be fine

AbandonedCart:
I'm not entirely familiar with your settings design yet, so hopefully that's right. It works, at least.

AbandonedCart:
I may be stuck adding secondary icons if the tint can't be set.

AbandonedCart:
I have one more thing I was playing around with before this one is wrapped up. I've never actually tried rebuilding the parameters on the fly to see if I can swap buttons while it's running. That would make the play / pause swap more natural.

bunnei:
@AbandonedCart looks like the aspect ratio is still a little bit off, any idea what is going on here?
image

AbandonedCart:
It mirrors your settings in the app. Are you saying you want it to be fullscreen in PiP regardless of user settings? It could get ugly, since it's not a conventional design (meaning the Android views do not match the emulator viewport)

t895 (commented):
I looked over the code and I'm pretty happy with everything. I would just like to see the emulation window not being squished while in PiP. I also noticed a bug that was my fault. It will be fixed shortly with the addition of #10622

AbandonedCart:
This doesn't handle if there is a change to aspect ratio on the fly. I am still looking into how to handle that without it being a huge inefficiency. It does, however, update aspect ratio when starting and returning from pip for newer android and (ironically) whenever it's launched automatically on older builds.

AbandonedCart:
So there's a slight logic problem. The layout you have and the layout you want are not compatible without overriding the user setting and the emulator is not designed to handle that. As I mentioned before, it will get ugly.

AbandonedCart:
You got to the right conclusion from the wrong direction. Actually, the issue is that you are rendering to part of a view, but PiP does not look at that and say "Ok, so just the part that's rendering, right?" It sees the view parameters, which aren't what the emulator is actually using. The black bars are part of the view and changing the aspect of PiP without setting the emulator to fill that view means you just have a scaled version of the previous view. A lot of it has to do with how your implementation does most of the work on the native side, rather than setting the native side to just output a full screen and then handling it on the Android side.

AbandonedCart:
I think the random delay will make a lot more sense when I wrap up the next commit.

t895:
I don't really understand what you're saying since (to my understanding) we use a view that takes up the entire screen and then we select a 16:9, 4:3, etc. chunk of it to render emulation to. Are you saying that the emulation surface follows the relative size of the view/display? For example, if I have a device that is 19:9 in landscape, when I go to PiP, it will appear squished. But if I have a device that is 16:9 in landscape and enter PiP, it will appear normally?

AbandonedCart:
What I'm saying is you have a surface view. That surface view is, for example 1920 x 1080. Let's say you are rendering the emulator at 1800 x 1080 to that surface. PiP doesn't know that. All it knows is that there is a 1920 x 1080 view. The black bars are "part of the view" in the eyes of Android.

t895:
Then in that case, it would be best to create a custom view that changes its aspect ratio based on the user setting right?

AbandonedCart:
I can deal with either honestly. The holdup is my crashing Ubuntu VM. It's down to whether you want to handle all of the aspect stuff on Android or in C.

t895:
Just so we're not flipping settings and reloading the surface, making the custom view would probably be best for simplicity.

AbandonedCart:
You're still going to be messing with settings to some extent, but ignoring them instead of actually changing any.

t895:
Also this is what I meant for making things easier here #10633

AbandonedCart:
Yeah, that's the "resize the view, not the rendering" concept I was describing. Thanks.

bunnei:
@AbandonedCart apologies that this got caught in the crossfire of us merging the lime PR. Github automatically closed this when we merged `lime` to `master`. Can you please rebase and reopen this PR onto the `master` branch? Looking forward to seeing this merged!

PR #10583, opened 2023-06-02
AccelerateDMA: Fix incorrect check in Buffer<->Texture copies

ameerj:
This PR improves performance, which I verified on some homebrew tests I made. 
But, I don't know which titles are likely to be affected by this change.

> `// No need to waste time on an image that's synced with guest`

This comment makes me think the idea for this check was to skip downloading images that haven't been modified on the GPU, but the early return here made it so that the slower unaccelerated path was taken instead, which still downloaded the image.

The logic is also flawed because it makes the **assumption** that the destination for the DMA copy is the same address as the CPU address for the image, which is "synced with guest" and thus needs no updating.
This assumption can't really be made.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -850,15 +850,11 @@ void TextureCache<P>::PopAsyncFlushes() {
 template <class P>
 ImageId TextureCache<P>::DmaImageId(const Tegra::DMA::ImageOperand& operand, bool is_upload) {
     const ImageInfo dst_info(operand);
-    const ImageId dst_id = FindDMAImage(dst_info, operand.address);
-    if (!dst_id) {
+    const ImageId image_id = FindDMAImage(dst_info, operand.address);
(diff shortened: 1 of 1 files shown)


gidoly:
seems there is no decent boost on dma heavy games. but still nice to have for frame times consistancy sake :)))
astralchain
build: 61fps
master:61fps
mario galaxy
build:88fps
master:87fps
mertroid prime
build: 417fps
master:411fps

astralchain master vs dma graph
CapFrameX_QrOzrukxI9
CapFrameX_IxCvGcxrB3
CapFrameX_AwzliEjpOF

ameerj:
Thank you for testing @gidoly, those are nice graphs 😄

liamwhite (approved):

goldenx86:
Could potentially cause instability issues in KOEI games https://github.com/yuzu-emu/yuzu/issues/10695

gidoly:
follow up for golden message. i have verified with user that this pr is cause

PR #10595, opened 2023-06-03
android: Replace deprecated and Java code

AbandonedCart:
A couple random `;` left behind and replacements for code that was deprecated in API 30 (the minimum of the app)

### DIFF
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
@@ -5,7 +5,6 @@ package org.yuzu.yuzu_emu.activities
 
 import android.app.Activity
 import android.content.Context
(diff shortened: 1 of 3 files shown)


AbandonedCart:
Are you trying to support foldables? It looks like some semi-implemented code that was in a commit about it, but there's no actual support for anything beyond landscape.

bunnei (approved):

bunnei:
@AbandonedCart thanks! Yes, we're working towards better support for foldables. At the moment, no one one the team owns one, so we haven't put a ton of effort here yet, but we're hoping to see this improve.

AbandonedCart:
>At the moment, no one one the team owns one, so we haven't put a ton of effort here yet, but we're hoping to see this improve.

I have a Z Flip 3. As is, the controller gets clumped into the top left of the bottom screen, but I have a feeling that's something that requires the native portrait / landscape support being implemented and then it should fix itself.

PR #10598, opened 2023-06-03
Return the amount of dedicated VRAM under OpenGL

scorpion81:
This is an attempt to prevent overflowing the actual VRAM on lower-end integrated devices. The former query returned a significantly higher value, for example it was above 8 GB on a steam deck, which has only a maximum UMA Framebuffer size of 4 GB. This can be set in the BIOS.

### DIFF
--- a/src/video_core/renderer_opengl/gl_device.cpp
+++ b/src/video_core/renderer_opengl/gl_device.cpp
@@ -286,7 +286,8 @@ void main() {
 
 u64 Device::GetCurrentDedicatedVideoMemory() const {
     GLint cur_avail_mem_kb = 0;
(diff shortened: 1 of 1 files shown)


scorpion81:
Hmm, seems I have tested this not correctly beforehand. 
I was sure that this change would limit the usage of VRAM, but there basically is not much difference to mainline yet. Currently I am testing with GL_GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX. The limiting works there partially, but with a rather strict 2-2.5 GB limit. 
I am also looking into using also GL_GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX in another function as part of the video memory usage calculation.

scorpion81:
should be superseded by #10619 now, hence closing.

PR #10603, opened 2023-06-04
core,common: Implement missing time zone data/computations

lat9nq:
Time zone information in yuzu is calculated using the same logic as https://www.iana.org/time-zones, however a number of bugs in the existing implementation has made it inconsistent with some users in the [previous PR](https://github.com/yuzu-emu/yuzu/pull/10397). This caused the previous attempt of using firmware time zone data problematic.

This PR tries to improve on that by fixing a number of bugs present in the time zone computations, as well as implementing missing code that was required. This also corrects an attempt at a bug fix in the previous PR by referencing the actual implementation.

In addition, yuzu presently only implements the GMT time zone data for the synthesized archive, utilized when the user installation lacks system firmware. The previous attempt relies on that data being present and later tried to add the root directory time zone data, but this is not sufficient as there are many time zones with various nuances. Now, [lat9nq/tzdb_to_nx](https://github.com/lat9nq/tzdb_to_nx) provides time zone data that is nearly equivalent to the Switch's (excepting the [Issues](https://github.com/lat9nq/tzdb_to_nx/issues) on that repo), and this PR includes that data through CMake by converting it to headers then linking the generated interface library.

Lastly, the previous PR tries to search for the closest time zone among the root directory time zones, when either the system archive lacks the time zone needed, or when the C++ implementation lacks `tzdb`. In this PR, I felt as though the root directory time zones weren't sufficient for our needs. Instead, if the host is on an hour offset, it just defaults to `Etc/GMT<offset>`. If the host is not on an hour offset, then it tries to map the host GMT offset to one of the time zones that also use that offset, DST considered. For hosts that need this path, this may not be optimal, but it at least covers all of the modern time zones. https://www.timeanddate.com/time/time-zones-interesting.html was used as a reference for the off time zones. Please let me know what I'm missing if any.

Supersedes #10397


externals/nx_tzdb/CMakeLists.txt
FearlessTobi, line 1: Should we submodule this?

lat9nq: Currently my repo isn't Windows-compatible, so not immediately. I could make tzdb_to_nx compatible with Windows to a point, but the sources from iana.org need to be built with a GNU toolchain. I don't have a solution for the MSVC toolchain on that.

bunnei (approved):

PR #10609, opened 2023-06-04
android: Add settings for variable orientation

AbandonedCart:
This is by no means ready to go, but it's the groundwork for trying to unravel the awkward orientation to support foldables and regular portrait layouts.

### DIFF
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
@@ -144,7 +144,7 @@ class EmulationActivity : AppCompatActivity(), SensorEventListener {
         startMotionSensorListener()
 
         NativeLibrary.notifyOrientationChange(
-            EmulationMenuSettings.landscapeScreenLayout,
+            EmulationMenuSettings.screenLayout,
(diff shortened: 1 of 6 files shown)


t895:
Currently we don't support a portrait layout and using your setting creates some weird results. For now could you just have options for auto, landscape, and reverse-landscape?

Additionally, having this option within the overlay options menu isn't really appropriate. Could you instead add a new option to the drawer or put it in the settings activity?

@Schplee thoughts on this?

AbandonedCart:
I can't seem to place this in draft and it appears there was some confusion in regards to stating it wasn't ready.

t895:
Sorry for the misunderstanding there. I'm excited to see this when it's ready.

AbandonedCart:
No worries. The weirdness is a known issue, but portrait is also necessary for portrait foldables. I threw the setting where I could get to it while I test different ways to go about it.

AbandonedCart:
So the story is that building on Mac 10.14 is no longer realistic due to Homebrew deprecations and upgrading isn't really an option because Adobe CS6 cost too much to throw out. Building through Actions worked (slowly) for testing stuff I was confident about, but ~25 minutes a build and needing to push to test wasn't really effective. I finally got the VM set up, though. Anyway...

PR #10618, opened 2023-06-04
android: Add licenses page

t895:
Please let me know if I should add any more licenses now or in the future.

### DIFF
--- a/src/android/app/src/main/res/values/strings.xml
+++ b/src/android/app/src/main/res/values/strings.xml
@@ -104,6 +104,7 @@
     <string name="contributors">Contributors</string>
     <string name="contributors_description">Made with \u2764 from the yuzu team</string>
     <string name="contributors_link">https://github.com/yuzu-emu/yuzu/graphs/contributors</string>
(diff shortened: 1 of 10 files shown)


src/android/app/src/main/res/values/strings.xml
```diff
@@ -346,4 +347,511 @@
     <string name="use_black_backgrounds">Use Black Backgrounds</string>
     <string name="use_black_backgrounds_description">When using the dark theme, apply black backgrounds.</string>
 
+    <!-- Licenses screen strings -->
```
FearlessTobi, line 350: can we make it so these are not translatable?

t895: For all of them or just the license text?

FearlessTobi: I think it's fine if everything in the license section except the Strings `licenses`  and `licenses_description` stays in English.
I have never seen it localized and it just created additional work on Transifex.

AbandonedCart: Licenses aren't translated.

bunnei (approved):

PR #10623, opened 2023-06-04
service: nfc: Add backup support

german77:
Since writing to amiibos with Joy-Con is now supported, all we need is backup support for amiibo data, just like on the Switch. This pull request will create a backup every time a valid amiibo is loaded or saved. The game will be able to detect the backup and restore it accordingly, fixing any corrupted amiibo tags.

As amiibo data is stored, you can use Yuzu to dump your amiibo bin from a Joy-Con or Pro Controller. You can find all the amiibo files in `%appdata%/amiibo/backup`.

### DIFF
--- a/src/common/fs/fs_paths.h
+++ b/src/common/fs/fs_paths.h
@@ -10,6 +10,7 @@
 
 // Sub-directories contained within a yuzu data directory
 
+#define AMIIBO_DIR "amiibo"
 #define CACHE_DIR "cache"
 #define CONFIG_DIR "config"
 #define DUMP_DIR "dump"
--- a/src/common/fs/path_util.cpp
+++ b/src/common/fs/path_util.cpp
@@ -114,6 +114,7 @@ public:
 #endif
 
         GenerateYuzuPath(YuzuPath::YuzuDir, yuzu_path);
+        GenerateYuzuPath(YuzuPath::AmiiboDir, yuzu_path / AMIIBO_DIR);
         GenerateYuzuPath(YuzuPath::CacheDir, yuzu_path_cache);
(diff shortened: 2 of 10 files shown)


liamwhite (approved):

src/common/fs/fs_paths.h
```diff
@@ -21,6 +21,7 @@
 #define SDMC_DIR "sdmc"
 #define SHADER_DIR "shader"
 #define TAS_DIR "tas"
+#define AMIIBO_DIR "amiibo"
```
Morph1984, line 24: Should be in lexicographical order

src/common/fs/path_util.cpp
```diff
@@ -120,6 +120,7 @@ class PathManagerImpl {
         GenerateYuzuPath(YuzuPath::SDMCDir, yuzu_path / SDMC_DIR);
         GenerateYuzuPath(YuzuPath::ShaderDir, yuzu_path / SHADER_DIR);
         GenerateYuzuPath(YuzuPath::TASDir, yuzu_path / TAS_DIR);
+        GenerateYuzuPath(YuzuPath::AmiiboDir, yuzu_path / AMIIBO_DIR);
```
Morph1984, line 123: ditto

src/common/fs/path_util.h
```diff
@@ -23,6 +23,7 @@ enum class YuzuPath {
     SDMCDir,        // Where the emulated SDMC is stored.
     ShaderDir,      // Where shaders are stored.
     TASDir,         // Where TAS scripts are stored.
+    AmiiboDir,      // Where Amiibo backups are stored.
```
Morph1984, line 26: ditto

PR #10641, opened 2023-06-06
android: Fix ci builds with Java 17

AbandonedCart:
The configuration requires 17, but the default in Actions is an older version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,6 +129,11 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: recursive
+      - name: set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'adopt'
       - name: Set up cache
         uses: actions/cache@v3
         with:


mrcmunir:
Seeems need to removed  -XX:MaxPermSize=512m  and replace with -XX:MaxMetaspaceSize=512 will fix that Unrecognized VM option 'MaxPermSize=512m' error.


https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html#removed-java-options

It's deprecated option from JDK 8 and up.

AbandonedCart:
> Seeems need to removed -XX:MaxPermSize=512m and replace with -XX:MaxMetaspaceSize=512 will fix that Unrecognized VM option 'MaxPermSize=512m' error.
> 
> https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html#removed-java-options
> 
> It's deprecated option from JDK 8 and up.

Yes, but that's the gradle.properties

That entire thing needs a review, but this is only fixing the ci part.

german77 (approved):

PR #10643, opened 2023-06-06
android: Improve Gradle build configuration

AbandonedCart:
org.gradle.jvmargs=
-Xms512m  <- Set the IDE to start with 512m allocated. 
-XX:MaxMetaspaceSize=512m <- Convert the MaxPermSize to new option 
-XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8 (previous options)

android.useAndroidX=true

kotlin.code.style=official

kotlin.parallel.tasks.in.project=true <- parallel compilation tasks in the same project

android.defaults.buildfeatures.buildconfig=true

### DIFF
--- a/src/android/gradle.properties
+++ b/src/android/gradle.properties
@@ -9,8 +9,9 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xms512m -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
 android.useAndroidX=true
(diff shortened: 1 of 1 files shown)


AbandonedCart:
Combines with https://github.com/yuzu-emu/yuzu/pull/10641 to allow verification to pass.

t895:
I'd prefer that you don't enable jetifier since we don't have any incompatible libraries and it will only add extra time to builds this way.

t895:
Also is there a difference in build times by changing the max to 3GBs now? It's a pretty small app with most of the compilation time happening for native code. If this doesn't help much, it just takes longer to build on low-memory systems right?

AbandonedCart:
Is it safe to assume you're building locally on a Chromebook equivalent?

AbandonedCart:
You're limiting it to 2. At that point, you're better off not setting a value at all. However, if you need that limitation, then I can redirect the adjustments to the ci. They can also be set there.

AbandonedCart:
For now, I removed the property entirely. It will start with 512 and the IDE / JVM will set the limit. This can be done for an instance of Android Studio by going to Help -> Edit Custom VM Options and setting an -Xmx value (without the declaration)

t895 (approved):

PR #10655, opened 2023-06-07
CMakeLists: Force C++20 on MSVC due to conflicts with C++23 modules

Morph1984:
The latest version of MSVC STL brings C++23 standard library modules, which conflict with precompiled headers. Disabling with /experimental:module- has no effect, so force C++20 in the meantime while we wait for module support in other compilers.


liamwhite (approved):

ACGNnsj:
This problem could be avoided by adding ClCompile option '<BuildStlModules>false</BuildStlModules>' to MSBuild files. However, it seems difficult to achieve this via CMake as far as I know.

PR #10668, opened 2023-06-08
Combine vertex/transform feedback buffer binding into a single call

Kelebek1:
Currently we bind every vertex and transform feedback buffer individually, which can lead to over a thousand extra API calls per frame. This PR combines them into a temp data structure and binds them at once instead. On buffer deletion, index and vertex buffers are also made conditional and specific, rather than invalidating the index and all vertex buffers every time any buffer is deleted.


Squall-Leonhart:
Any benchmarks done?

gidoly:
~8% boost in general

liamwhite (approved):

Kelebek1:
> ~8% boost in general

It's not 8%, it's like <1%

bunnei (approved):

PR #10686, opened 2023-06-09
android: Fix input overlay version check

t895:


liamwhite (approved):

AbandonedCart:
https://github.com/yuzu-emu/yuzu/blob/master/src/android/app/build.gradle.kts#L58

Android R is API 30. You don't need a check at all.

t895:
> https://github.com/yuzu-emu/yuzu/blob/master/src/android/app/build.gradle.kts#L58
> 
> Android R is API 30. You don't need a check at all.

I'm aware. This is just for when someone wants to build it themselves on a lower API level.

AbandonedCart:
> I'm aware. This is just for when someone wants to build it themselves on a lower API level.

That doesn't really make sense, though. It sounds like instead of throwing an error that will cause someone to look into it, you're suppressing the error so they have to go digging line by line for where you set a minimum API without any handling for the lower ones.

t895:
I'd prefer to have the input overlay load with the small drawback of not insetting the view and allow emulation instead of throwing an error and crashing on lower API levels. It's not worth creating a new codepath for Android 10 and below to inset the view since we're not supporting it. This is just to allow the bare minimum.

AbandonedCart:
So you want it to build for a lower API without any changes and simply not work as expected?

t895:
Yes

AbandonedCart:
> Yes

Well that seems malicious, but whatever then.

bunnei (approved):

PR #10692, opened 2023-06-10
Improve path splitting speed

BreadFish64:
Creates a range closure so the components can be iterated without allocating a container.
Use the range closure to speed up `VfsDirectory::GetFileRelative`

Unfortunately the C++ standard has no direct equivalent to range-v3's `split_when_view`. If pulling range-v3 in is not okay, then building a vector of string_views like this is still relatively cheap:
```cpp
std::vector<std::string_view> SplitPathComponentsView(std::string_view filename) {
    std::vector<std::string_view> components;
    const char* component_begin = filename.data();
    const char* const end = component_begin + filename.size();
    for (const char* it = component_begin; it != end; ++it) {
        const char c = *it;
        if (c == '\\' || c == '/') [[unlikely]] {
            if (component_begin != it) {
                components.emplace_back(std::string_view{component_begin, it});
            }
            component_begin = it + 1;
        }
    }
    if (component_begin != end) {
        components.emplace_back(std::string_view{component_begin, end});
    }
    return components;
}
```
I benchmarked this by looping `CreateRomFS` for ToTK with several mods. Both the range adaptor and vector of string_view approaches doubled the speed. The range adapter is about 8% faster than making a vector of string_views.

Original
Original
Vector of string_view
Vector of string_view
Range adaptor closure
Range adaptor closure

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -279,6 +279,7 @@ find_package(LLVM MODULE COMPONENTS Demangle)
 find_package(lz4 REQUIRED)
 find_package(nlohmann_json 3.8 REQUIRED)
 find_package(Opus 1.3 MODULE)
+find_package(range-v3 REQUIRED)
 find_package(ZLIB 1.2 REQUIRED)
 find_package(zstd 1.5 REQUIRED)
 
--- a/src/common/fs/path_util.cpp
+++ b/src/common/fs/path_util.cpp
@@ -350,17 +350,11 @@ std::string_view RemoveTrailingSlash(std::string_view path) {
 }
 
 std::vector<std::string> SplitPathComponents(std::string_view filename) {
-    std::string copy(filename);
-    std::replace(copy.begin(), copy.end(), '\\', '/');
-    std::vector<std::string> out;
-
-    std::stringstream stream(copy);
-    std::string item;
-    while (std::getline(stream, item, '/')) {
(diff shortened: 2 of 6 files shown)


Kelebek1:
Would a boost small_vector of string views and just iterating normally through the string not do the same thing?

BreadFish64:
I hesitated to try a small_vec before because it would be pretty large on the stack, but I tried a 16 element small_vec and it is faster.
image

boludoz:
Why not use std::filesystem?

FearlessTobi:
Please also retest how much this improves speeds on latest master vs this PR.
We made quite a few optimizations to our fs code in the meantime.

boludoz:
Could you use native std::filesystem or std::experimental::filesystem from stdc++17?, just tell me what I should redesign and I'll do it.

liamwhite:
@boludoz I think this might be a misunderstanding of the code - VfsFile/VfsDirectory is for emulated files and directories, which are layered on top of std::filesystem for interaction with host files.

boludoz:
> @boludoz I think this might be a misunderstanding of the code - VfsFile/VfsDirectory is for emulated files and directories, which are layered on top of std::filesystem for interaction with host files.

The code above splits everything into characters and not into folders and files, it also doesn't support non-ANSII chars

PR #10726, opened 2023-06-11
android: Adapt EmulationActivity to navigation component

t895:
Some work I forgot to merge since I got caught up with the launch

### DIFF
--- a/src/android/app/build.gradle.kts
+++ b/src/android/app/build.gradle.kts
@@ -9,6 +9,7 @@ plugins {
     id("org.jetbrains.kotlin.android")
     id("kotlin-parcelize")
     kotlin("plugin.serialization") version "1.8.21"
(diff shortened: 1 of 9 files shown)


AbandonedCart:
```
emulationFragment = navHostFragment.childFragmentManager.fragments.find {
            it is EmulationFragment
        } as? EmulationFragment
```

AbandonedCart:
I rolled a fix into the PiP changes and removed the need for the variable. I called the listener from the fragment and passed the activity context. It's actually a lot cleaner that way, since everything is in one place.

bunnei (approved):

AbandonedCart:
You may need to have the ci run `vcpkg update` to fix the failure if that keeps happening. It may just be the cache this used, though.

PR #10739, opened 2023-06-12
Re-enable SDL's `CPUinfo` subsystem

zeltermann:
TLDR: Disabling `CPUinfo` triggers a very nasty SDL audio subsystem bug, which **utterly** breaks SDL's JACK output on Linux, so no audio when one wanted to do `SDL_AUDIODRIVER=jack bin/yuzu`. This is a workaround for that.

Below is the offending SDL code in question, if anyone is curious. Without `CPUinfo`, it thinks there is no SIMD support (so it doesn't use SIMD granule type conversion), while also thinking there is no lack of SIMD support, because for example on x86 `SDL_SSE2_INTRINSICS` is still defined (so `NEED_SCALAR_CONVERTER_FALLBACKS` isn't set, so it doesn't use the fallback granule type conversion functions either), thus all the `SDL_Convert_x_to_y` functions are null and there is no way for other parts of SDL's audio subsystem to work properly:

https://github.com/libsdl-org/SDL/blob/4e0f94ea7d58d4a1aac624421d73110232df3efd/src/audio/SDL_audiotypecvt.c#L939C1-L978

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -63,8 +63,9 @@ if (YUZU_USE_EXTERNAL_SDL2)
         # Yuzu itself needs: Atomic Audio Events Joystick Haptic Sensor Threads Timers
         # Since 2.0.18 Atomic+Threads required for HIDAPI/libusb (see https://github.com/libsdl-org/SDL/issues/5095)
         # Yuzu-cmd also needs: Video (depends on Loadso/Dlopen)
+        # CPUinfo also required for SDL Audio, at least until 2.28.0 (see https://github.com/libsdl-org/SDL/issues/7809)
         set(SDL_UNUSED_SUBSYSTEMS
-            CPUinfo File Filesystem
+            File Filesystem
             Locale Power Render)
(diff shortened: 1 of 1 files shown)


v1993 (commented):
Nice catch! Probably worth reporting to SDL too since it's a rather unexpected problem for unsuspecting users.

zeltermann:
I think it's best to keep `CPUinfo` there for a while, even after when SDL 2.28.0 is out (https://github.com/libsdl-org/SDL/milestone/8). I have not looked into other parts of SDL's internal code, but
- removing it might break other things
- we might not have SIMD audio granule conversion if the SDL devs screw this up even harder
- keeping `CPUinfo` probably lets SDL use SIMD whenever possible anyways

zeltermann:
The SDL bug is actually already fixed in https://github.com/libsdl-org/SDL/pull/7464, not sure which SDL release it will land (it didn't land in SDL 2.26.5), but nonetheless `CPUinfo` should be enabled so that the audio subsystem gets to use SIMD.

liamwhite (approved):

PR #10744, opened 2023-06-12
video_core: Improved anisotropic filtering heuristics

### LINKED ISSUE #9800: Metroid Prime Remastered Texture Filtering Not Working Correctly

### Affected Build(s)

All recent early access builds.

### Description of Issue

The texture filtering settings that are selected in the options do not apply to this game. There is a hard falloff in texture quality after a certain distance regardless of the texture filtering setting used.

### Expected Behavior

Yuzu:

Image00002

Ryujinx (No texture filtering issue):

Image00001

### Reproduction Steps

No reproduction steps. The error just occurs.

Below is a save file that can be used to quickly see the issue. Simply walk out of the door, stay in the doorway to kill the three flying enemies, and then stand on the following platform area to clearly see the texture filtering bug.
[save.zip](https://github.com/yuzu-emu/yuzu/files/10727205/save.zip)

### Log File

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/10727196/yuzu_log.txt)

### System Configuration

CPU: 5800X3D
GPU/Driver: 6900 XT / 22.11.2
RAM: 32 GB 4400 MHz CL 17 DDR4 G.Skill TridentZ
OS: Windows 10 x64

Wollnashorn:
Currently anisotropic filtering is only added to trilinear texture samplers, but there are some games such as BotW, TotK and Metroid Prime Remastered, where terrain textures are sampled with the nearest neighbor mipmap mode. As result the AF setting has no effect on these textures.

This PR adds an option in the advanced graphics tab to enable anisotropic filtering for all mipmap modes. There might be some rare instances where this could cause graphical glitches, like maybe in a post-processing effect using mipmap trickery or something, but I haven't encountered such problems yet and expect it to be very unlikely.

Fixes #9800

### DIFF
--- a/src/video_core/renderer_vulkan/vk_texture_cache.h
+++ b/src/video_core/renderer_vulkan/vk_texture_cache.h
@@ -279,8 +279,17 @@ public:
         return *sampler;
     }
 
+    [[nodiscard]] VkSampler HandleWithoutAnisotropy() const noexcept {
+        return *sampler_without_anisotropy;
+    }
+
+    [[nodiscard]] bool HasAddedAnisotropy() const noexcept {
+        return static_cast<bool>(sampler_without_anisotropy);
+    }
+
 private:
     vk::Sampler sampler;
+    vk::Sampler sampler_without_anisotropy;
 };
 
 class Framebuffer {
--- a/src/video_core/renderer_vulkan/vk_texture_cache.h
+++ b/src/video_core/renderer_vulkan/vk_texture_cache.h
@@ -279,17 +279,17 @@ public:
         return *sampler;
     }
 
-    [[nodiscard]] VkSampler HandleWithoutAnisotropy() const noexcept {
-        return *sampler_without_anisotropy;
+    [[nodiscard]] VkSampler HandleWithDefaultAnisotropy() const noexcept {
+        return *sampler_default_anisotropy;
     }
 
     [[nodiscard]] bool HasAddedAnisotropy() const noexcept {
-        return static_cast<bool>(sampler_without_anisotropy);
+        return static_cast<bool>(sampler_default_anisotropy);
     }
 
 private:
     vk::Sampler sampler;
-    vk::Sampler sampler_without_anisotropy;
+    vk::Sampler sampler_default_anisotropy;
 };
 
 class Framebuffer {
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -236,7 +236,6 @@ void RestoreGlobalState(bool is_powered_on) {
     values.bg_blue.SetGlobal(true);
     values.enable_compute_pipelines.SetGlobal(true);
     values.use_video_framerate.SetGlobal(true);
-    values.use_aggressive_anisotropic_filtering.SetGlobal(true);
 
     // System
     values.language_index.SetGlobal(true);
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -236,6 +236,7 @@ void RestoreGlobalState(bool is_powered_on) {
     values.bg_blue.SetGlobal(true);
     values.enable_compute_pipelines.SetGlobal(true);
     values.use_video_framerate.SetGlobal(true);
+    values.use_aggressive_anisotropic_filtering.SetGlobal(true);
 
     // System
     values.language_index.SetGlobal(true);
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -483,6 +483,8 @@ struct Values {
         AstcRecompression::Uncompressed, AstcRecompression::Uncompressed, AstcRecompression::Bc3,
         "astc_recompression"};
     SwitchableSetting<bool> use_video_framerate{false, "use_video_framerate"};
+    SwitchableSetting<bool> use_aggressive_anisotropic_filtering{
+        false, "use_aggressive_anisotropic_filtering"};
 
     SwitchableSetting<u8> bg_red{0, "bg_red"};
     SwitchableSetting<u8> bg_green{0, "bg_green"};
--- a/src/common/settings.h
+++ b/src/common/settings.h
@@ -483,8 +483,6 @@ struct Values {
         AstcRecompression::Uncompressed, AstcRecompression::Uncompressed, AstcRecompression::Bc3,
         "astc_recompression"};
     SwitchableSetting<bool> use_video_framerate{false, "use_video_framerate"};
-    SwitchableSetting<bool> use_aggressive_anisotropic_filtering{
-        false, "use_aggressive_anisotropic_filtering"};
 
     SwitchableSetting<u8> bg_red{0, "bg_red"};
     SwitchableSetting<u8> bg_green{0, "bg_green"};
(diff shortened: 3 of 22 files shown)


gidoly:
i think we used to deafult this bahavior but broke some game. reproducable on pr artifact. nice to have as option if it improves visual on several games though.
might worth documenting game that breaks with it here.
example of breaking game
Astral chain:
before
yuzu_dQyeFpwtev

after
yuzu_g587aPwz3D

xenoblade games:
before
yuzu_4nfrzFYhAD

after
yuzu_18BuJpeZRa

fire emblem three houses:
before
yuzu_IDNwn74DfV

after
yuzu_quQVXoiHuZ

Wollnashorn:
Thanks for testing @gidoly and for the confirmation, that it can introduce such artifacts.. So it certainly should be used with care and of course not be enabled by default.

FernandoS27:
I don't think an option is necessary but a better eurostic. We should not apply anisotropic on Depth Compare for instance.

gidoly:
seems result is same i couldn't find improvement with lastest commit with both option on and off

Wollnashorn:
@gidoly Now it should actually be improved (hopefully). AF was added even when there were no mipmaps at all and that's silly I just realized 😅 Thanks again for testing!

gidoly:
seems fixes artifacting on xenoblade games with option on. astral chan and feth remains same

gidoly:
hmm seems build without this changes has no issue on feth and astral chain with af filter. seems some behavior got applied globally? regardless of setting?

Wollnashorn:
@gidoly I had an oversight in it, thanks for catching. It is fixed now. Also I added the requirement that samplers must have a normal LOD range from 0 to at least 1, this should fix the clouds in Fire Emblem

gidoly:
confirmed that issues i pointed got addressed.
maybe it is out of scope of this pr but may worth to investigate these af issue while we are at it.
1.astral chain with option on
off
yuzu-msvc-verify-pr10744-20230614-601868f20_zWsAgZ8vUy

on
yuzu-msvc-verify-pr10744-20230614-601868f20_dkCUpx8w8I

2. mario kart 8 deluxe issue on meow meow meadows with af in general
default
yuzu-msvc-verify-pr10744-20230614-601868f20_LKxG9qC089


x16
yuzu-msvc-verify-pr10744-20230614-601868f20_pOYVVmcUy3

seems option on works better in general after those commits maybe we can default this bahvior and get rid of option once we addressw what i said above.

Wollnashorn:
I've now had a chance to look into the bug in Astral Chain, it's a bit tricky there. The game uses color grading with a LUT, which is a 2D image atlas. It looks like sampling this LUT with anisotropy causes the wrong colors getting indexed.

astral_aniso_renderdoc

The same texture sampler is used all over the place though and there is no way to tell from the sampler configuration alone that it will cause any problems.
A way to address it would be to check the associated image views when the pipeline decriptors are set, and fallback to an alternative without anisotropy. @FernandoS27 I think you had something like this in mind with the image formats, but I had not yet the idea.

In this case, the LUT texture does not even have mips, so anisotropy should be disabled for this particular image view.

src/video_core/renderer_vulkan/vk_texture_cache.h
```diff
@@ -279,8 +279,17 @@ class Sampler {
...
+        return *sampler_without_anisotropy;
+    }
+
+    [[nodiscard]] bool HasAddedAnisotropy() const noexcept {
+        return static_cast<bool>(sampler_without_anisotropy);
+    }
+
 private:
     vk::Sampler sampler;
+    vk::Sampler sampler_without_anisotropy;
```
Wollnashorn, line 292: Maybe this should be better named `sampler_isotropic`? 🤔

xTeixeira:
> @xTeixeira Thanks for reporting, can you test with the latest change again?

xTeixeira:
image

Appears to be fixed.

Although now it crashes with the [AF fix mod](https://github.com/HolographicWings/TOTK-Mods-collection/tree/main/Mods/Graphics/Anisotropic%20Filtering%20Fix%20-%20Set%20AF%20to%2016%20in%20Yuzu%20advanced%20graphics%20settings).
(Not sure if this mod is still needed with this change?)

Wollnashorn:
Nice that it's presumably fixed. My AF fix mod is pretty much obsolete with this new heuristics, but I still recommend it, as trilinear sampling with anisotropic filtering has still less visible seams/mip banding than bilinear + anisotropic.
It is strange that it's crashing for you, I just tried it and it worked fine. Do maybe other mods intervene somehow?

xTeixeira:
Tested again (including latest changes) with all mods disabled except AF fix. Still crashing for me during the first few frames of the main menu screen.

v1993:
I'm getting segfaults from within nvidia driver with this PR shortly after starting Kirby Star Allies/Kirby's Return to Dream Land DX - maybe in more games, those are just two I had around to test. Newly introduced option does not affect segfaults - they happen both with it enabled and disabled. I have anisotropic set to 4x globally. Worth noting is that only Vulkan crashes - OpenGL starts fine.

xTeixeira:
> Newly introduced option does not affect segfaults

Just tested here too. Also crashing with the option disabled.

Wollnashorn:
Hm, that's strange, had no crashes so far on AMD. I've now slightly corrected it, that the fallback sampler uses the unaltered max anisotropy value - I don't think that this fixes the segfaults, but maybe worth a try. Otherwise I need to ask a friend to debug on his NVIDIA card

xTeixeira:
Yeah still segfaults unfortunately

bunnei:
@Wollnashorn yeah can confirm that the last commit causes a segfault with Kirby Star Allies on nvidia:

```
 	nvoglv64.dll!00007ffced9d75ac()	Unknown
 	nvoglv64.dll!00007ffced9dacac()	Unknown
>	[Inline Frame] yuzu.exe!Vulkan::vk::CommandBuffer::PushDescriptorSetWithTemplateKHR(VkDescriptorUpdateTemplate_T *) Line 1024	C++
 	yuzu.exe!Vulkan::GraphicsPipeline::ConfigureDraw::__l2::<lambda_2>::operator()(Vulkan::vk::CommandBuffer cmdbuf) Line 530	C++
 	yuzu.exe!Vulkan::Scheduler::TypedCommand<`Vulkan::GraphicsPipeline::ConfigureDraw'::`2'::<lambda_2>>::Execute(Vulkan::vk::CommandBuffer cmdbuf) Line 141	C++
 	[Inline Frame] yuzu.exe!Vulkan::Scheduler::CommandChunk::ExecuteAll(Vulkan::vk::CommandBuffer) Line 28	C++
 	yuzu.exe!Vulkan::Scheduler::WorkerThread(std::stop_token stop_token) Line 182	C++
 	[External Code]	
```

Untagging from EA for now.

Wollnashorn:
@bunnei Thanks, that was helpful. I think I now know what's causing the UB, will fix it soon

FernandoS27:
I think you could even force trilinear on this conditions but I would leave that to a future PR

Wollnashorn:
Hope it is fixed now. I overlooked that `TextureCache::GetGraphicsSampler`/`GetComputeSampler` can add new samplers and thus causing the slot vector to reallocate, so storing pointers to these slots for later use was plain undefined behaviour.

I like the idea to force trilinear for suitable texture-sampler pairs

gidoly:
confirmed that it fixes crash on kirby games as well

gidoly:
untag due to merge conflict with #10777 for timing being

Wollnashorn:
Maybe it is working already well enough that we can get rid of the toggle to match all mipmap modes? That should solve the merge conflict too

gidoly:
yeah. if you can take a look on mk8 lastly it will be prefect

FernandoS27 (approved):

PR #10760, opened 2023-06-13
android: Declare languages in locales_config.xml

FearlessTobi:
This is required to make per-app language swithcing possible on Android 13.


src/android/app/build.gradle.kts
```diff
@@ -79,6 +79,9 @@ android {
             abiFilters += listOf("arm64-v8a")
         }
 
+        val locales = listOf("en", "de", "es", "fr", "it", "ja", "nb", "pl", "pt-rBR", "pt-rPT", "ru", "uk", "zh-rCN", "zh-rTW")
+        resourceConfigurations.addAll(locales)
+
```
t895, line 84: These lines are unnecessary as of AGP 8.0

t895 (approved):

PR #10767, opened 2023-06-14
android: Linting

t895:
This adds a build check to make sure all Kotlin code is consistent. There's also a nice auto formatting option across the whole project if you run `gradlew ktlintFormat`. The check currently runs every time you build but I could split this out into CI only if that's more appropriate.

Now all Kotlin code has been formatted via this linter.

This also enables the default android linting that checks for potential problems when using certain APIs.

### DIFF
--- a/src/android/app/build.gradle.kts
+++ b/src/android/app/build.gradle.kts
@@ -44,16 +44,6 @@ android {
         jniLibs.useLegacyPackaging = true
     }
 
-    lint {
-        // This is important as it will run lint but not abort on error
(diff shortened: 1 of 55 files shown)


AbandonedCart:
I think having it run automatically is a necessary evil, honestly, but you could always filter it to only run on assemble (or whatever it is when you do run, not just rebuild). 

One of the things I like most about the git setup you have is that if I try to commit something where I deleted a line and the tab got missed, it errors out and says "this repo has whitespace restricts" or some such error. I've caught a lot of overlooked tabs with that.

liamwhite (approved):

PR #10788, opened 2023-06-15
Add an option to set the amiibo directory

AbandonedCart:
I imagine it can't really be this easy, but I have minimal experience with C / Qt / Visual Studio. What I do have is a ROG Ally with a microsd card slot and a very risky hard drive replacement procedure for a device that only released two days ago.

Anyway, it was worth a shot if it can make the folder a little more accessible.

### DIFF
--- a/src/yuzu/configuration/configure_filesystem.cpp
+++ b/src/yuzu/configuration/configure_filesystem.cpp
@@ -19,6 +19,8 @@ ConfigureFilesystem::ConfigureFilesystem(QWidget* parent)
             [this] { SetDirectory(DirectoryTarget::NAND, ui->nand_directory_edit); });
     connect(ui->sdmc_directory_button, &QToolButton::pressed, this,
             [this] { SetDirectory(DirectoryTarget::SD, ui->sdmc_directory_edit); });
+    connect(ui->amiibo_directory_button, &QToolButton::pressed, this,
+            [this] { SetDirectory(DirectoryTarget::Amiibo, ui->amiibo_directory_edit); });
     connect(ui->gamecard_path_button, &QToolButton::pressed, this,
             [this] { SetDirectory(DirectoryTarget::Gamecard, ui->gamecard_path_edit); });
     connect(ui->dump_path_button, &QToolButton::pressed, this,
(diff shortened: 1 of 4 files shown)


german77:
My overall idea is to make an amiibo manager. So you can import, export, restore and edit your files. Easily. So having direct access to the amiibo folder isn't necessary. 

At the moment this directory is useless since it's the same if you use any other directory on your PC to load the amiibos. I just needed a place to store backups which happen to be a great way to get dumps of your amiibos.

german77:
Yuzu can load a virtual amiibo from a valid tag dump, in either encrypted or plain format. It can also dump amiibos with a joycon or pro controller. 

I just don't support emuiibos, foomibos or any other format. Although I been considering supporting json amiibos since they seem a convenient format rather than raw binary.

Tagmo is a good tool but they are a big red flag since it can clone tags.

AbandonedCart:
> Tagmo is a good tool but they are a big red flag since it can clone tags.

What do you mean?

> I just don't support emuiibos, foomibos or any other format.

I'm not really sure about emuiibo, but wumiibo are amiibo generated from an ID for 3ds and foomiibo are just the same concept, but fixed for Switch. You already support them if you support bin files.

german77:
Clone tags as burn fresh tags with an amiibo. This is not legal for comercial use.

If other bin formats follow the structure of a real amiibo dump then we just need to change the validation checks https://github.com/yuzu-emu/yuzu/blob/master/src/core/hle/service/nfc/common/amiibo_crypto.cpp#L19. These checks are in place to validate the integrity of the data. To avoid using partial dumps or corrupted dumps and avoiding damaging data from files that aren't amiibos. 

This might also cause issues if the binary file doesn't have the proper file size. https://github.com/yuzu-emu/yuzu/blob/master/src/input_common/drivers/virtual_amiibo.cpp#L109

AbandonedCart:
> Clone tags as burn fresh tags with an amiibo. This is not legal for comercial use.

I knew what you meant by clone, not by saying it was a red flag. It allows you to make backups of amiibo you own, not unlike making a backup of a game. You still need the keys and the amiibo for that. It's a bit like the pot calling the kettle there. The foomiibo take the ID and empty data and simply create a bin file in the right structure. Nothing needs to be changed for them to work. The caveat is that the serial number would fail any sort of validation (date, production facility, etc) because it's completely random.

AbandonedCart:
> 

I think you're thinking of the UID. I'm talking about the serial number of the statue.

https://www.reddit.com/r/amiibo/comments/3jw88z/comment/cusxmoj/

AbandonedCart:
If you want to stick with reading bin files, I'm not trying to convince you otherwise. I was only mentioning that the possibility of another option exists. The virtual files made from TagMo aren't an option, though. They would work out of the box, but I removed the ability to export them to files to deter distribution.

Either way, it sounds like the actual PR isn't relevant, so i'll close this.

Issue #10757, opened 2023-06-13
[Android] Allow batch install of multiple dlcs


Raxcoms:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What feature are you suggesting?

Allow selecting multiple dlcs to be installed at once

### Why would this feature be useful?

Would facilitate installing dlcs for games that have a bunch of it (i.g. SSBU)

AbandonedCart:
I got you, buddy. ;)

t895:
Addressed by #10794

PR #10794, opened 2023-06-15
android: Add support for multiple installs

AbandonedCart:
Very simple way of replicating the desktop installation process for multiple files.

Outputs a success if all installs work or a fail if at least one fails. Lists the number of files that succeed, overwrite, or fail and any failure causes.

Resolves the request in https://github.com/yuzu-emu/yuzu/issues/10757

### DIFF
--- a/src/android/app/src/main/res/values/strings.xml
+++ b/src/android/app/src/main/res/values/strings.xml
@@ -104,12 +104,14 @@
     <string name="share_log_missing">No log file found</string>
     <string name="install_game_content">Install game content</string>
     <string name="install_game_content_description">Install game updates or DLC</string>
-    <string name="install_game_content_failure">Error installing file to NAND</string>
-    <string name="install_game_content_failure_description">Game content installation failed. Please ensure content is valid and that the prod.keys file is installed.</string>
-    <string name="install_game_content_failure_base">Installation of base games isn\'t permitted in order to avoid possible conflicts. Please select an update or DLC instead.</string>
-    <string name="install_game_content_failure_file_extension">The selected file type is not supported. Only NSP and XCI content is supported for this action. Please verify the game content is valid.</string>
-    <string name="install_game_content_success">Game content installed successfully</string>
-    <string name="install_game_content_success_overwrite">Game content was overwritten successfully</string>
+    <string name="install_game_content_failure">Error installing file(s) to NAND</string>
+    <string name="install_game_content_failure_description">Please ensure content(s) are valid and that the prod.keys file is installed.</string>
+    <string name="install_game_content_failure_base">Installation of base games isn\'t permitted in order to avoid possible conflicts.</string>
+    <string name="install_game_content_failure_file_extension">Only NSP and XCI content is supported. Please verify the game content(s) are valid.</string>
+    <string name="install_game_content_failed_count">%1$d installation error(s)</string>
+    <string name="install_game_content_success">Game content(s) installed successfully</string>
+    <string name="install_game_content_success_install">%1$d installed successfully</string>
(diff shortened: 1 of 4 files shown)


src/android/app/src/main/res/values/strings.xml
```diff
@@ -104,12 +104,14 @@
...
-    <string name="install_game_content_failure">Error installing file to NAND</string>
-    <string name="install_game_content_failure_description">Game content installation failed. Please ensure content is valid and that the prod.keys file is installed.</string>
-    <string name="install_game_content_failure_base">Installation of base games isn\'t permitted in order to avoid possible conflicts. Please select an update or DLC instead.</string>
-    <string name="install_game_content_failure_file_extension">The selected file type is not supported. Only NSP and XCI content is supported for this action. Please verify the game content is valid.</string>
-    <string name="install_game_content_success">Game content installed successfully</string>
-    <string name="install_game_content_success_overwrite">Game content was overwritten successfully</string>
+    <string name="install_game_content_failure">Error installing file(s) to NAND</string>
+    <string name="install_game_content_failure_description">\nGame content(s) installation failed. Please ensure content is valid and that the prod.keys file is installed.</string>
```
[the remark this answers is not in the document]
AbandonedCart: It needs to be moved to the stringbuilder, but it's for the
```
# installed
# overwritten
# failed
Reasons things failed
```
dialog

AbandonedCart:
This is a draft because it's written in a text editor. I'd prefer to do some proper testing, but also don't love the logic and was open to better ways to structure the dialog.

AbandonedCart:
I think the dialog may be one {} too soon. I'm also not entirely sure what to return for the newInstance since it's on longer a reflection of any one file.

AbandonedCart:
Not pretty, but actually working.

AbandonedCart:
It depends on the new dialog fragment. The alternative was to include this entire PR in that one, but then I'd be assuming this gets merged first to rebase that one. Another option was to include the fragment in both and then whichever is second has merge conflicts. In the end, I figured it was easier to drop a commit if that wasn't desired over trying to break it up.

AbandonedCart:
`/home/runner/work/yuzu/yuzu/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt:6:1 Imports must be ordered in lexicographic order without any empty lines in-between`

... except, as far as I can see, there aren't any issues.

AbandonedCart:
As long as 50% of people use optimize imports, it will fix itself fairly fast if they get disorganized.

src/android/app/src/main/java/org/yuzu/yuzu_emu/utils/DeviceUtil.kt
```diff
@@ -0,0 +1,49 @@
...
+package org.yuzu.yuzu_emu.utils
+
+import android.app.ActivityManager
+import android.content.Context
+import java.util.Locale
+
+class DeviceUtil(context: Context) {
+
+    private val Long.floatForm: String
+        get() = String.format(Locale.US, "%.2f", this.toDouble())
```
FearlessTobi, line 13: Shouldn't this be using the locale of the user instead?
German uses the comma as the digit seperator instead of the point for example.

AbandonedCart: That's a regex.

AbandonedCart:
Looks like vcpkg had some sort of a network error and suddenly Android Studio started putting the kotlin imports above the internal imports, which is what klint wanted. I don't know what to think anymore

AbandonedCart:
Rebased, but unchanged

bunnei (approved):

PR #10795, opened 2023-06-15
input_common: Add foomiibo support

german77:
Foomiibo is a plain amiibo that has been autogenerated. The only difference is that they contain an extra header at the end compared to a regular amiibo dump. Add this file size to supported amiibo formats.

### DIFF
--- a/src/input_common/drivers/virtual_amiibo.cpp
+++ b/src/input_common/drivers/virtual_amiibo.cpp
@@ -82,6 +82,7 @@ VirtualAmiibo::Info VirtualAmiibo::LoadAmiibo(const std::string& filename) {
     switch (nfc_file.GetSize()) {
     case AmiiboSize:
     case AmiiboSizeWithoutPassword:
+    case AmiiboSizeWithSignature:
         data.resize(AmiiboSize);
         if (nfc_file.Read(data) < AmiiboSizeWithoutPassword) {
             return Info::NotAnAmiibo;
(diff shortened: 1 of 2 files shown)


german77:
Honestly I never heard about that signature before. I'm only basing on what the NTAG215 protocol says. But now that you mention it is part of the security stuff of a NTAG. Which I wasn't aware of an is exactly as you describe it. I'm unsure why this part is relevant as from my RE is never used at all by the switch. 

```
Each NTAG21x UID is signed with a NXP private key and the resulting 32-byte signature
is stored in a hidden part of the NTAG21x memory during IC production.
```

The rest of the file contents already meet with the minimum validation I do to the amiibos. So this kind of format already has full support on yuzu.

AbandonedCart:
It's not used in conventional amiibo functionality, but it is how you get the hardware ID of an N2 Elite. I used it for the sake of making the Foomiibo "recognizable" from regular amiibo. The reason I don't allow the files to be saved to userspace as files is that each one is freshly generated every time it is accessed. The only instance that exists is the one written to a tag. The second a file is copied, it is no longer unique and the purpose falls apart.

german77:
Got it. Thanks for the clarification.

AbandonedCart:
Thanks for hearing me out. It wouldn't be such a touchy subject if not for how quickly it gets mistaken for piracy.

liamwhite (approved):

PR #10797, opened 2023-06-16
time: Various time zone fixes

lat9nq:
This fixes a few things I wasn't made aware of until after #10603 was merged.

The Flatpak build is not able to download anything during CMake configure as it has no internet access, so it couldn't download the time zone archive. Without time zone data, a C++ assertion would fail when booting a game. The affected area had an entirely separate bug in that the existing code was comparing the wrong booleans. In searching for the issue, my time zone homebrew revealed to me that I had spelled Antarctica wrong in the previous PR, and so none of that data was accessible in yuzu.

This corrects those issues. For the Flatpak, I submoduled [eggert/tz](https://github.com/eggert/tz) (checked out to `2022c`) in [lat9nq/tzdb_to_nx](https://github.com/lat9nq/tzdb_to_nx), and then submoduled lat9nq/tzdb_to_nx here in externals. If the required build tools cannot be found (i.e. MSVC toolchain), or if `tzdb2nx` cannot run on the host (e.g. Windows), it will download the binary archive as before.


externals/nx_tzdb/CMakeLists.txt
```diff
@@ -1,24 +1,60 @@
...
+
+add_library(nx_tzdb INTERFACE)
+
+find_program(GIT git)
+find_program(GNU_MAKE make)
+find_program(GNU_DATE date)
 
+set(CAN_BUILD_NX_TZDB true)
+
+if ("${GIT}" STREQUAL "GIT-NOTFOUND")
```
Tachi107, line 14: This can be simplified to `if (NOT GIT)`, see [the docs](https://cmake.org/cmake/help/latest/command/if.html#basic-expressions)

Also, in general, `if ("${VAR}")` is generally equivalent to the simpler `if (VAR)`

lat9nq: Thanks, I wasn't aware of this.

bunnei (approved):

PR #10801, opened 2023-06-16
android: Fix aspect ratio when rotating screen

AbandonedCart:
The "locked" aspect ratio is based on landscape, so allowing the activity to start in portrait means landscape will use the width and height of portrait to dictate the landscape aspect ratio. The end result is an emulation surface that is short and wide. By launching in landscape, this sets the proper aspect ratio at the cost of a quick transition.

### DIFF
--- a/src/android/app/src/main/AndroidManifest.xml
+++ b/src/android/app/src/main/AndroidManifest.xml
@@ -53,6 +53,7 @@ SPDX-License-Identifier: GPL-3.0-or-later
         <activity
             android:name="org.yuzu.yuzu_emu.activities.EmulationActivity"
             android:theme="@style/Theme.Yuzu.Main"
+            android:screenOrientation="userLandscape"
             android:supportsPictureInPicture="true"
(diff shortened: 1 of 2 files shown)


bunnei (approved):

goldenx86:
Screenshot_20230702_050811_yuzu

Doesn't seem to be working as intended.

pacoa-kdbg:
Same here, stretches

AbandonedCart:
> Same here, stretches

Disable the stretch to fit aspect. It's working as intended for your settings.

goldenx86:
Yep, even with a forced 16:9 aspect ratio, the image stretches.

AbandonedCart:
Any idea when it stopped working?

pacoa-kdbg:
I am on latest EA, but if there is some artifact I can install and test to confirm if the PR was working on my device

AbandonedCart:
The PR was definitely working. It's a matter of tracking down what caused it to stop.

PR #10808, opened 2023-06-16
android: Expose settings

t895:
Exposes settings for CPU debugging, a combined fastmem toggle, and an audio engine switcher


AbandonedCart:
Do you want a mute / unmute for PiP? This was the missing part to allow muting that without changing the entire system audio.

t895:
If you want to add one go ahead. Sounds useful.

AbandonedCart:
> If you want to add one go ahead. Sounds useful.

I like it for long cutscenes. I can switch to something else and let it run, but without it blaring the audio. I'll put it together for when this hits and I can base it.

bunnei (approved):

PR #10811, opened 2023-06-16
android: Add a PiP interface to mute / unmute

AbandonedCart:
This is a bit of a hack using the setting in lieu of a direct native interface, but adds a mute / unmute to PiP.


AbandonedCart:
I think I need a better way to handle this. I don't know if loading the settings will be too much overhead for the RAM impaired.

german77:
We have a setting called `Settings::values.audio_muted`. Please use this to enable or disable audio. If you are worried about loading the setting on kotlin you could make a jni function to disable it from the emulated side.

german77:
Just a bit curious what's the motivation behind this setting. Most phones have dedicated buttons to control the volume.

AbandonedCart:
> Just a bit curious what's the motivation behind this setting. Most phones have dedicated buttons to control the volume.

This isn't changing device volume. It's muting picture in picture. Unlike pause, it allows a game to continue without sound. It's pretty straightforward

AbandonedCart:
The JNI needed the attention.

src/android/app/src/main/java/org/yuzu/yuzu_emu/NativeLibrary.kt
```diff
@@ -313,6 +313,21 @@ object NativeLibrary {
...
 
+    /**
+     * Mutes emulation sound
+     */
+    external fun muteAudio(): Boolean
+
+    /**
+     * Unmutes emulation sound
+     */
+    external fun unMuteAudio(): Boolean
```
FearlessTobi, line 324:
```suggestion
    external fun unmuteAudio(): Boolean
```

src/android/app/src/main/res/values/strings.xml
```diff
@@ -375,6 +375,8 @@
     <string name="picture_in_picture_description">Minimize window when placed in the background</string>
     <string name="pause">Pause</string>
     <string name="play">Play</string>
+    <string name="mute">Mute</string>
+    <string name="sound">Sound</string>
```
FearlessTobi, line 379: I think "Sound" is a bit ambigous. Can we just name it "Unmute" like we do with the function names already?

src/android/app/src/main/java/org/yuzu/yuzu_emu/NativeLibrary.kt
AbandonedCart: You used "unPause" for the emulator, so are they supposed to inconsistent?

src/android/app/src/main/res/values/strings.xml
AbandonedCart: I used sound because it's not shown and "unmute" comes up as a typo, but I guess that would have been more relevant if there weren't a mountain of typo warnings already.

AbandonedCart: This is also going to make them harder to differentiate down the road, hence why "unpause" was implemented as "play" but I also imagine this was viewed out of context.

src/android/app/src/main/java/org/yuzu/yuzu_emu/NativeLibrary.kt
AbandonedCart: I'll just fix both. After the inconsistency between the header and cpp, it's probably for the best.

AbandonedCart:
Just to clarify, responding to the requested changes is only meant to provide the logic in case it matters. At the end of the day, the changes get made and can always be reverted if that reasoning actually matters.

bunnei (approved):

PR #10835, opened 2023-06-18
vulkan_device: Restrict compute disable only to affected Intel drivers

lat9nq:
Followup to #10181

Intel finally fixed the compute issue, so we can disable it here where needed. Currently the only fixed driver is a beta version, so this just checks for affected drivers. As a side effect, this also re-enables compute on older unsupported Intel graphics devices where they never had the compute issue in the first place.

### DIFF
--- a/src/video_core/vulkan_common/vulkan_device.h
+++ b/src/video_core/vulkan_common/vulkan_device.h
@@ -10,6 +10,7 @@
 #include <vector>
 
 #include "common/common_types.h"
+#include "common/logging/log.h"
 #include "common/settings.h"
 #include "video_core/vulkan_common/vulkan_wrapper.h"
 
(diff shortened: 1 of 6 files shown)


src/video_core/vulkan_common/vulkan_device.h
```diff
@@ -579,6 +585,22 @@ class Device {
...
+            const u32 minor = VK_API_VERSION_MINOR(driver_version);
+            const u32 patch = VK_API_VERSION_PATCH(driver_version);
+            if (major == 0 && minor == 405 && patch < 286) {
+                LOG_WARNING(
+                    Render_Vulkan,
+                    "Intel proprietary drivers 0.405.0 until 0.405.286 have broken compute");
+                return true;
+            }
+        }
+        return {};
```
FearlessTobi, line 601:
```suggestion
        return false;
```

liamwhite (approved):

PR #10865, opened 2023-06-21
android: Clean up file extension checks

t895:
Things should be much more consistent now.

I saw one user say that renaming an extension from ".NSP" to ".nsp" allowed their game to appear in their games library. I wasn't able to repro but I tried to make case-checking better here anyways.


src/android/app/src/main/java/org/yuzu/yuzu_emu/utils/FileUtil.kt
```diff
@@ -326,25 +325,9 @@ object FileUtil {
...
-        cursor.close()
-
-        if (fileName == null) {
-            return false
-        }
-        return fileName.substring(fileName.lastIndexOf(".") + 1).contains(extension)
+    fun getExtension(uri: Uri): String {
+        val fileName = getFilename(uri)
+        return fileName.substring(fileName.lastIndexOf(".") + 1)
+            .lowercase(Locale.ENGLISH)
```
AbandonedCart, line 331: Android provides a `.lowercase()` for situations where locale shouldn't be a factor at all.

bunnei (approved):

PR #10888, opened 2023-06-23
android: (native) Parameter types from Android Studio

AbandonedCart:
Android Studio marked these parameters as errors because it is an instance, not a class, that is being passed from Java.

After the pip mute fixes were pushed, Android Studio was able to index the file a little better.

### DIFF
--- a/src/android/app/src/main/jni/native.cpp
+++ b/src/android/app/src/main/jni/native.cpp
@@ -528,23 +528,24 @@ static Core::SystemResultStatus RunEmulation(const std::string& filepath) {
 
 extern "C" {
 
-void Java_org_yuzu_yuzu_1emu_NativeLibrary_surfaceChanged(JNIEnv* env, jclass clazz, jobject surf) {
+void Java_org_yuzu_yuzu_1emu_NativeLibrary_surfaceChanged(JNIEnv* env, jobject instance,
+                                                          jobject surf) {
     EmulationSession::GetInstance().SetNativeWindow(ANativeWindow_fromSurface(env, surf));
     EmulationSession::GetInstance().SurfaceChanged();
 }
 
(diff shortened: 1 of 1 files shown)


FearlessTobi:
Please fix clang format.

AbandonedCart:
> Please fix clang format.

Was it really that urgent?

AbandonedCart:
JNI doesn't like `const auto` but marking every instance with [[maybe-unused]] is a mess. Working out a better way to target the type, not the invocation.

AbandonedCart:
Not sure why the website, atom, and Android Studio all have the same settings and still try to format the files with entirely different spacing.

AbandonedCart:
This supplements auto and const auto with the types jauto and jconst replace assigning [[maybe_unused]] to every instance, as Android Studio is unable to identify usage for auto.

liamwhite (approved):

german77:
Probably out of this topic. But I don't have any good way to communicate with you. I want to add writing support to amiibos on android. But documentation is kind of poor for this and I need some help to know what to do.

From what I have read I first need to send the auth password then apparently send all writes in just 4 bytes blocks. Until I saved everything on the tag. I was expecting to be able to write way more data on a single function call but seems like it's simply not allowed.

AbandonedCart:
I'm not sure I understand the purpose. Didn't you post on TagMo that you could already read and write from a controller and isn't being able to clone an amiibo what you called a red flag?

At this point, even writing something new will more or less be derived from TagMo, so there is also some concern about that. I've submitted a lot of stuff I wrote for TagMo, but I'm not sure how I feel about a yuzu copyright getting slapped over the core code.

AbandonedCart:
That aside, conventional NFC works in pages. The pages are small, but it doesn't require a delay between chunks like Bluetooth. Specialty hardware, such as the N2, allows you to stream a read or write without dividing into pages, but that's also because the underlying target is not a standard tag.

Split the data into pages and loop through them. Write the password. Write the lock. Both are based on the first page.

PR #10901, opened 2023-06-24
input_common: Make use of new SDL features

german77:
With new SDL version joycons stopped working correctly. Address this by updating the flags. Battery level is also a proper event now. So update it like any other input as well.

There's one bug remaining. Player leds are always off on switch controllers. This needs to be fixed upstream so I will address that issue in another PR.

### DIFF
--- a/src/input_common/drivers/sdl_driver.cpp
+++ b/src/input_common/drivers/sdl_driver.cpp
@@ -150,6 +150,8 @@ public:
         if (sdl_controller) {
             const auto type = SDL_GameControllerGetType(sdl_controller.get());
             return (type == SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO) ||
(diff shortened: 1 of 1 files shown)


liamwhite (approved):

src/input_common/drivers/sdl_driver.cpp
```diff
@@ -466,6 +461,13 @@ void SDLDriver::HandleGameControllerEvent(const SDL_Event& event) {
         }
         break;
     }
+    case SDL_JOYBATTERYUPDATED: {
+        if (auto joystick = GetSDLJoystickBySDLID(event.csensor.which)) {
```
slouken, line 465: This should be `event.jbattery.which`

german77: Ty

german77:
This PR is related to the player fix. https://github.com/yuzu-emu/ext-windows-bin/pull/42 as well.

Issue #8489, opened 2022-06-21
real ring-con in fitness ring adventure


taifengdian:
I have been paying attention to the use of real ring con in Yuzu. It seems that german77 has implemented this function, but it has not been integrated into the released version. I want to know what the problem is?

fedeshoker:
Hi! Any updates for real ring con ? Thank you

taifengdian:
it's looks like no progress ！

HisNoodlyServant:
Also going to join in on the request. Be cool as heck to be able to use my Ring Fit with the Deck.

taifengdian:
chech this post every week

Gordoras:
I would really appreciate an update :)
It would really be great to use the real ring con and play the game how it is supposed to be played.

taifengdian:
sad

pktiuk:
> The problem is that it needs a custom joycon driver. I have made a driver but it's purely experimental. The ring does work there but the feature as such needs to be polished so it doesn't crash yuzu or the driver under any circumstance


Maybe it would be good to just put it into some test/nightly release to let people test and break it (and collect logs) to know where it fails.

german77:
The driver is ready and needs to be tested. Ping me on discord(Narr the reg) if you want to join to the testing

dario7218:
Can i join the testing?

pktiuk:
Unluckily currently I am not able to test it, but I see you already have another volunteer. :)

dario7218:
Is the testing going on?

german77:
Yes. about 7 have tested. Original joycons and ring work perfectly. 3rd party not at all. Right now I'm finishing the code to add it to EA

nahlahman:
I'm really excited about this feature. Thank you for working on it!

taifengdian:
what a wonderful news, about one year waiting is valuable. Thanks very much to german77's great work.

netrunner7:
+1 !!!

taifengdian:
@german77 I have tried the early access 3246 version and couldn't get the ring-con work. I don't know if the version doesn't get ready for ring-con or I operated somewhere wrong?  could you show me  some guide. thank you very much.

german77:
Ringcon requires a very specific setup to work properly. Logs will print whatever the ring was detected or not. 

-Custom driver enabled.
-Player 1 as right joycon. With right joycon mapped.
-Player 2 as dual joycon. With left joycon mapped.
[...]

PR #10908, opened 2023-06-25
Clarify Ring-Con configuration message in UI

kiri11:
It is not obvious how the left Joy-Con should be set up.
Clarify message and mention that it should be left physical, but dual emulated according to [this comment](https://github.com/yuzu-emu/yuzu/issues/8489#issuecomment-1546490439).

Without this clarification one might think to set up left controller as dual both physical and emulated, and this won't work. In fact, this confusion has already happened as we can see from comment history on the same issue #8489.

### DIFF
--- a/src/yuzu/configuration/configure_ringcon.ui
+++ b/src/yuzu/configuration/configure_ringcon.ui
@@ -23,7 +23,7 @@
       </size>
      </property>
      <property name="text">
-      <string>If you want to use this controller configure player 1 as right controller and player 2 as dual joycon before starting the game to allow this controller to be detected properly.</string>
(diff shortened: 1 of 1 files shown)


src/yuzu/configuration/configure_ringcon.ui
```diff
@@ -23,7 +23,7 @@
       </size>
      </property>
      <property name="text">
-      <string>If you want to use this controller configure player 1 as right controller and player 2 as dual joycon before starting the game to allow this controller to be detected properly.</string>
+      <string>To use Ring-Con configure player 1 as right JoyCon both physical and emulated, and player 2 as left JoyCon physical and dual (!) JoyCon emulated before starting the game.</string>
```
AbandonedCart, line 26: "right Joy-Con**,** both" (Joy-Con is also hyphenated, by the way)

kukukuan2023:
EA 3708 ring-con works.
3709 reports driver exception: 8.

It's probably not related to this UI opt.

PR #10916, opened 2023-06-25
 OpenGL: Add Local Memory warmup shader for Nvidia

ameerj:
Here's a fun one.

Nvidia's OpenGL drivers see a significant performance uplift in shaders using local memory if the previous shader that was executed uses more local memory (sort of..)

To take advantage of this discovery, this change introduces a dummy shader that allocates the maximum allowed local memory, but returns immediately without executing. This tricks the driver into boosting the performance of the later executing shaders which use local memory.

Here's a local memory torture test homebrew used to test this: 
[textured_cube_compute.zip](https://github.com/yuzu-emu/yuzu/files/11861207/textured_cube_compute.zip)

In this homebrew, my performance went from 7 to 8 FPS, up to 40 to 42 FPS, matching Vulkan's performance.

### DIFF
--- a/src/shader_recompiler/backend/glasm/emit_glasm.cpp
+++ b/src/shader_recompiler/backend/glasm/emit_glasm.cpp
@@ -461,7 +461,7 @@ std::string EmitGLASM(const Profile& profile, const RuntimeInfo& runtime_info, I
         header += fmt::format("R{},", index);
     }
     if (program.local_memory_size > 0) {
-        header += fmt::format("lmem[{}],", program.local_memory_size);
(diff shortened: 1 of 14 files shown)


Kelebek1:
Does this not apply to Vulkan as well? All the local memory usage should be well beyond the graphics API. If this is OpenGL-specific, any idea why?

ameerj:
This perf quirk does not affect Vulkan, only OpenGL, and it affects both GLSL and GLASM.
I can only guess at this point, but my assumption is that there's some aggressive optimization to limit lmem size in the GL driver, which gets reused in subsequent shader executions if they too use lmem.

gidoly:
~2% boost in general in games

liamwhite (approved):

PR #10930, opened 2023-06-27
settings: Catch runtime_error, fallback time zone

lat9nq:
Windows lets you select time zones that fail in their own C++ implementation library. Evidently from the stack trace, we get a runtime error to work with, so catch it and use the fallback.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright 2021 yuzu Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <exception>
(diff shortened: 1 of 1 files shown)


src/common/settings.cpp
```diff
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright 2021 yuzu Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <exception>
+#include <stdexcept>
 #if __cpp_lib_chrono >= 201907L
```
Morph1984, line 6: `#include <future>`

Morph1984 (approved):

PR #10934, opened 2023-06-27
cmake: allow using system VMA library

abouvier:
Follow-up of https://github.com/yuzu-emu/yuzu/pull/10473.
It works on my system, but let's see if there is an error here.


AbandonedCart:
I have a feeling you could fix the Linux builds by throwing`vulkan-sdk` in the packages installed during setup.

abouvier:
No, the error is in upstream `VulkanMemoryAllocator`. They should link the library with `Vulkan::Headers`, but not use `find_package(Vulkan)` when `Vulkan::Headers` is already defined (by yuzu's own `find_package(Vulkan)` or `add_subdirectory(Vulkan-Headers)`).

abouvier:
Ttentative fix in https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/pull/345.

liamwhite (approved):

PR #10942, opened 2023-06-28
Memory Tracking: Add mechanism to register small writes when gpu page is contested by GPU

FernandoS27:
This PR adds a mechanism to store small CPU writes when a GPU page is being written by the GPU. It assumes coherency happens at a cache line level (64 bytes). If the page is not contested it will be fully invalidated.

This fixes:
* Pokemon Violet and scarlet Vertex explosions (keep it at High accuracy for god's sake).
* Rain/Snow in Zelda BotW and TotK.
* Weird particle movement in SMO.
* Particle corruption in Zelda TotK.
* Corrupting Thumbnail in Xenoblade DE.
* Eyebrow animation in Xenoblade 3.

EDIT: After reading the TRM, the coherency depends if the page bypasses the cache (CPU writes/reads directly to that page in RAM) or if it uses the CPU cache. We assume the second so the accuracy remains 64 bytes. In the future we may increase the accuracy for cases where the CPU writes directly to RAM and skips caching. The GPU writing is already designed for coherent writes.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -27,6 +27,7 @@
 #include "core/file_sys/savedata_factory.h"
 #include "core/file_sys/vfs_concat.h"
 #include "core/file_sys/vfs_real.h"
+#include "core/gpu_dirty_memory_manager.h"
 #include "core/hid/hid_core.h"
 #include "core/hle/kernel/k_memory_manager.h"
 #include "core/hle/kernel/k_process.h"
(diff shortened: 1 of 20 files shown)


gidoly:
Ci failed

FernandoS27:
GCC should now compile with 64 bit atomics.

liamwhite (approved):

PR #10943, opened 2023-06-28
android: Input overlay updates

t895:
Now within the Input Overlay file, there is an `OVERLAY_VERSION` code that will determine when the overlay will be reset. This is intended for breaking changes like the ones we had with the additions of percentage based layouts or the addition of foldable/portrait layouts.

This also includes L3/R3 buttons with icons provided by @Schplee 

Screenshot_20230628-145030

### DIFF
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
@@ -212,9 +212,9 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
             }
             if (!isInFoldableLayout) {
                 if (newConfig.orientation == Configuration.ORIENTATION_PORTRAIT) {
-                    binding.surfaceInputOverlay.orientation = InputOverlay.PORTRAIT
+                    binding.surfaceInputOverlay.layout = InputOverlay.PORTRAIT
                 } else {
-                    binding.surfaceInputOverlay.orientation = InputOverlay.LANDSCAPE
+                    binding.surfaceInputOverlay.layout = InputOverlay.LANDSCAPE
                 }
             }
             if (!binding.surfaceInputOverlay.isInEditMode) {
@@ -260,7 +260,9 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
             .remove(Settings.PREF_CONTROL_SCALE)
             .remove(Settings.PREF_CONTROL_OPACITY)
             .apply()
-        binding.surfaceInputOverlay.post { binding.surfaceInputOverlay.resetButtonPlacement() }
+        binding.surfaceInputOverlay.post {
+            binding.surfaceInputOverlay.resetLayoutVisibilityAndPlacement()
+        }
     }
 
     private fun updateShowFpsOverlay() {
@@ -337,7 +339,7 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
                         binding.inGameMenu.layoutParams.height = it.bounds.bottom
 
                         isInFoldableLayout = true
-                        binding.surfaceInputOverlay.orientation = InputOverlay.FOLDABLE
+                        binding.surfaceInputOverlay.layout = InputOverlay.FOLDABLE
                         refreshInputOverlay()
                     }
                 }
@@ -410,9 +412,9 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
                 R.id.menu_toggle_controls -> {
                     val preferences =
                         PreferenceManager.getDefaultSharedPreferences(YuzuApplication.appContext)
-                    val optionsArray = BooleanArray(15)
-                    for (i in 0..14) {
-                        optionsArray[i] = preferences.getBoolean("buttonToggle$i", i < 13)
+                    val optionsArray = BooleanArray(Settings.overlayPreferences.size)
+                    Settings.overlayPreferences.forEachIndexed { i, _ ->
+                        optionsArray[i] = preferences.getBoolean("buttonToggle$i", i < 15)
                     }
 
                     val dialog = MaterialAlertDialogBuilder(requireContext())
@@ -436,7 +438,7 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
                     dialog.getButton(AlertDialog.BUTTON_NEUTRAL)
                         .setOnClickListener {
                             val isChecked = !optionsArray[0]
-                            for (i in 0..14) {
+                            Settings.overlayPreferences.forEachIndexed { i, _ ->
                                 optionsArray[i] = isChecked
                                 dialog.listView.setItemChecked(i, isChecked)
                                 preferences.edit()
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/overlay/InputOverlay.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/overlay/InputOverlay.kt
@@ -51,15 +51,23 @@ class InputOverlay(context: Context, attrs: AttributeSet?) :
 
     private lateinit var windowInsets: WindowInsets
 
-    var orientation = LANDSCAPE
+    var layout = LANDSCAPE
 
     override fun onLayout(changed: Boolean, left: Int, top: Int, right: Int, bottom: Int) {
         super.onLayout(changed, left, top, right, bottom)
 
         windowInsets = rootWindowInsets
 
-        if (!preferences.getBoolean("${Settings.PREF_OVERLAY_INIT}$orientation", false)) {
-            defaultOverlay()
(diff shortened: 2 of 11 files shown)


src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
```diff
@@ -436,7 +436,7 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
                     dialog.getButton(AlertDialog.BUTTON_NEUTRAL)
                         .setOnClickListener {
                             val isChecked = !optionsArray[0]
-                            for (i in 0..14) {
+                            for (i in Settings.overlayPreferences.indices) {
```
AbandonedCart, line 439: Why not `Settings.overlayPreferences.forEachIndexed`?

t895: It does the same thing so it doesn't really matter

AbandonedCart: It is a lot cleaner and easier to understand, but serves the same purpose.

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
```diff
@@ -410,9 +410,9 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
                 R.id.menu_toggle_controls -> {
                     val preferences =
                         PreferenceManager.getDefaultSharedPreferences(YuzuApplication.appContext)
-                    val optionsArray = BooleanArray(15)
-                    for (i in 0..14) {
-                        optionsArray[i] = preferences.getBoolean("buttonToggle$i", i < 13)
+                    val optionsArray = BooleanArray(Settings.overlayPreferences.size)
+                    for (i in Settings.overlayPreferences.indices) {
+                        optionsArray[i] = preferences.getBoolean("buttonToggle$i", i < 13 || i > 14)
```
AbandonedCart, line 415: Wouldn't this just be "not 14" then?

src/android/app/src/main/java/org/yuzu/yuzu_emu/overlay/InputOverlay.kt
```diff
@@ -670,7 +715,11 @@ class InputOverlay(context: Context, attrs: AttributeSet?) :
...
         R.integer.SWITCH_STICK_L_X_FOLDABLE,
-        R.integer.SWITCH_STICK_L_Y_FOLDABLE
+        R.integer.SWITCH_STICK_L_Y_FOLDABLE,
+        R.integer.SWITCH_BUTTON_STICK_L_X_FOLDABLE,
+        R.integer.SWITCH_BUTTON_STICK_L_Y_FOLDABLE,
+        R.integer.SWITCH_BUTTON_STICK_R_X_FOLDABLE,
+        R.integer.SWITCH_BUTTON_STICK_R_Y_FOLDABLE
     )
 
     private fun getResourceValue(orientation: String, position: Int): Float {
```
AbandonedCart, line 725: Any reason not to change this "orientation" to "layout" like the rest?

AbandonedCart:
Is it really necessary to reset the entire layout every time a change is made? 

I understand that when portrait was added, it was a "breaking" change for the foldable layout that was never fully implemented. For portrait, there were no previous settings and landscape was untouched. This would have meant sacrificing someone's landscape settings for a layout with default settings and one their device may not even support. 

That said, shouldn't the version be divided to the layout, not unlike the buttons. This allows you to choose if all layouts need to be reset or, in the case of foldables, only the one impacted.

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
t895: 13 and 14 are the home and screenshot buttons which are disabled by default. I could swap this to be `i != 13 or 14` for a little more simplicity though.

AbandonedCart: I have a feeling the reason they were last was the exclusion. Since that didn't pan out, why not make them first? Instead of 0 until size, you'd do 2 until size and skip iterating them entirely.

t895: Sorry the last comment was silly from me. It was because of the exclusion but when I started writing this change I wanted to avoid creating problems with the existing layout. But since this change will trigger a reset, that should be fine.

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
t895: A little I suppose, sure I'll add it

src/android/app/src/main/java/org/yuzu/yuzu_emu/overlay/InputOverlay.kt
```diff
@@ -685,125 +734,141 @@ class InputOverlay(context: Context, attrs: AttributeSet?) :
         // Each value represents the position of the button in relation to the screen size without insets.
         preferences.edit()
             .putFloat(
-                ButtonType.BUTTON_A.toString() + "-X$orientation",
+                "${Settings.PREF_BUTTON_A}-X$orientation",
```
AbandonedCart, line 737: Good catch on these. They should have been concatenated before. That's my fault.

t895:
> Is it really necessary to reset the entire layout set every time a change is made?
> 
> I understand that when portrait was added, it was a "breaking" change for the foldable layout that was never fully implemented. For portrait, there were no previous settings and landscape was untouched. This would have meant sacrificing someone's landscape settings for a layout with default settings and one their device may not even support.
> 
> That said, shouldn't the version be divided to the layout, not unlike the buttons. This allows you to choose if all layouts need to be reset or, in the case of foldables, only the one impacted.

I suppose I could make this slightly differently where I could have a major version number that could reset all layouts and then a sub-version number that would only reset its corresponding layout if need-be.

t895:
Just pushed an attempt at what I mentioned in the previous comment. I haven't tested it yet though.

t895:
Now the major overlay versions and per-layout sub-versions are tested and ready

AbandonedCart:
> Now the major overlay versions and per-layout sub-versions are tested and ready

Thanks. It's a little more tedious to track, but will save dealing with a meltdown or two.

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
AbandonedCart: I didn't realize it until the second comment, so I completely understand.

FearlessTobi:
Why is the L3 button on the right? That seems a bit counterintuitive to me

FearlessTobi (commented):
Code looks good to me

t895:
@Schplee figures that you'll be moving a lot with the left stick with your right thumb free to press the L3/R3 buttons

Schplee:
> @Schplee figures that you'll be moving a lot with the left stick with your right thumb free to press the L3/R3 buttons

Correct. Crouching, zooming, or other tertiary functions are better handled by the opposite hand so that the player can focus on uninterrupted movement.

AbandonedCart:
> Correct. Crouching, zooming, or other tertiary functions are better handled by the opposite hand so that the player can focus on uninterrupted movement.

Isn't L3 the button triggered by pressing down the left joystick, though? It seems like the purpose of being able to configure placement was for preferences like this one, but that the defaults were supposed to resemble a physical controller.

Schplee:
> > Correct. Crouching, zooming, or other tertiary functions are better handled by the opposite hand so that the player can focus on uninterrupted movement.
> 
> Isn't L3 the button triggered by pressing down the left joystick, though? It seems like the purpose of being able to configure placement was for preferences like this one, but that the defaults were supposed to resemble a physical controller.

The goal of the default layout is to provide reasonable defaults. This does end up looking a bit like a physical controller, but matching the physical layout of a controller is not the goal of the defauly layout.

AbandonedCart:
> The goal of the default layout is to provide reasonable defaults. This does end up looking a bit like a physical controller, but matching the physical layout of a controller is not the goal of the default layout.

I guess I don't really see a scenario where the explanation justifies it being awkward the other 99% of the time. Most games provide a click to toggle option, which is more or less the cost of replacing a physical controller with emulated buttons (assuming you don't want to design them to act like their physical counterparts). No games really consider that you'll already be holding down the L3 with your right hand when expecting you to press a button or trigger.

bunnei (approved):

PR #10949, opened 2023-06-29
android: Rework MemoryUtil

t895:
~~The API for accessing the total amount of system memory has not allowed apps to see memory hidden from the kernel until API 34 (Android 14). As a result the memory warning would appear inaccurate and inconsistent between devices. (Reports 7/8GBs on my Pixel 7 and 6/8GBs on @goldenx86 's S23)~~

~~If there is a way to get the full system memory amount on lower API levels (there has to be since apps like CPU-Z have been around for ages and I think you could do this via JNI), I'll just add it later on.~~

I reworked this PR so we don't have to worry as much about API level (and avoid JNI entirely).

### DIFF
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
@@ -34,11 +34,14 @@ import androidx.core.view.WindowCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsControllerCompat
 import androidx.navigation.fragment.NavHostFragment
+import androidx.preference.PreferenceManager
 import org.yuzu.yuzu_emu.NativeLibrary
 import org.yuzu.yuzu_emu.R
+import org.yuzu.yuzu_emu.YuzuApplication
 import org.yuzu.yuzu_emu.databinding.ActivityEmulationBinding
 import org.yuzu.yuzu_emu.features.settings.model.BooleanSetting
 import org.yuzu.yuzu_emu.features.settings.model.IntSetting
+import org.yuzu.yuzu_emu.features.settings.model.Settings
 import org.yuzu.yuzu_emu.features.settings.model.SettingsViewModel
 import org.yuzu.yuzu_emu.model.Game
 import org.yuzu.yuzu_emu.utils.ControllerMappingHelper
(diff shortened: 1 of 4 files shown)


AbandonedCart:
Wouldn't it be more appropriate to determine what the minimum requirement of visible RAM would be? Seems a bit silly to say you can only know your device isn't going to work if it's running a version of the OS that has barely come out of beta and likely only works with devices that have enough RAM.

src/android/app/src/main/java/org/yuzu/yuzu_emu/activities/EmulationActivity.kt
```diff
@@ -106,17 +107,19 @@ class EmulationActivity : AppCompatActivity(), SensorEventListener {
...
-            ).show()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            if (MemoryUtil.isLessThan(MemoryUtil.REQUIRED_MEMORY, MemoryUtil.Gb)) {
+                Toast.makeText(
+                    this,
+                    getString(
+                        R.string.device_memory_inadequate,
+                        MemoryUtil.getDeviceRAM(),
+                        NumberFormat.getInstance().format(MemoryUtil.REQUIRED_MEMORY) +
+                            " ${getString(R.string.memory_gigabyte)}"
```
AbandonedCart, line 118: This seems excessive. The main string could be changed to `%2$d %3$s` and the two passed as individual parameters. There's no reason for concatenation of two individual values. That was only to include the literal here.

t895:
I do agree it's a bit silly but we'd have a similar problem since the system will automatically kill background processes when we start allocating a bunch of memory. The initial reading will be inconsistent.

AbandonedCart:
Google Play allows you to set a RAM minimum, so you are only really targeting sideloads. Between that and it only being available for Android 14 (which is now enforcing a 4GB minimum for 64-bit), it's probably better off removed or limited to part of a log. Logging was how I use it. Even if you tell someone they have 4GB and you need 6 GB, most are going to try it, but logging that let me know so I wasn't chasing around an imaginary issue.

t895:
It's still a handy feature so I wouldn't remove it (I guess if anything it should really only warn the user once). We can't stop users from sideloading too so it could just be a good thing to show no matter what.

AbandonedCart:
Once or as a note in the settings are viable options. My concern was it becoming a nag notice. Since total RAM is a finite value and the requirements of the emulator or unlikely to change much, it would become a nuisance.

liamwhite (approved):

PR #10952, opened 2023-06-29
video_core: vulkan_device: Remove superfluous is_qualcomm check.

bunnei:
- I think we added this for one bad Qualcomm driver, but seems fine now.


bunnei:
This is not right, need to rethink and test more devices.

PR #10953, opened 2023-06-29
Texture cache: Fix YFC regression due to code testing

FernandoS27:
This fixes a regression with NVN virtual images not being correctly remapped when sparse memory changes to mapped.

It should fix regressions in UE4 games like bravely default, SMT V and Pikmin 4 demo.


Squall-Leonhart:
https://github.com/yuzu-emu/yuzu/issues/10962

PR #10964, opened 2023-06-30
video_core: vulkan_device: Fix S8Gen2 dynamic state checks.

bunnei:
Take 2 -- These are needed on S8Gen2, but not older.


liamwhite (approved):

bylaws:
You shouldn't special case per soc here, it's the drivers that are broken

bunnei:
@bylaws Yep you're right -- marking as do-not-merge for now until I have a chance to figure out which driver broke things

PR #10969, opened 2023-06-30
kernel: Synchronize

Morph1984:
Synchronize implementations of svcReplyAndReceive and svcWaitSynchronization


liamwhite (approved):
Note: there is a TOCTTOU bug with using IsValidVirtualAddressRange (frankly, this method probably shouldn't exist at all, and is a mistake on my behalf), but this code does not change the correctness of what is done here.

PR #10974, opened 2023-07-01
vulkan: Improvements to macOS surface creation

Steveice10:
* Migrates from the legacy VK_MVK_macos_surface to the recommended VK_EXT_metal_surface. We can fetch the required CAMetalLayer from the window using a simple objc_msgSend trick, same as in Citra.
* Fixes a preferences crash on macOS following the addition of present mode detection.
  * The code for this passes in a GMainWindow for surface creation, which is not configured to have a Vulkan surface. On other platforms this seems to work, but on macOS this fails due to the window not having a CAMetalLayer.
  * Instead, create a dummy invisible window with surface type VulkanSurface to perform this test.

### DIFF
--- a/src/video_core/vulkan_common/vulkan_instance.cpp
+++ b/src/video_core/vulkan_common/vulkan_instance.cpp
@@ -19,11 +19,9 @@
 #include <windows.h>
 // ensure include order
 #include <vulkan/vulkan_win32.h>
-#elif defined(__APPLE__)
-#include <vulkan/vulkan_macos.h>
 #elif defined(__ANDROID__)
 #include <vulkan/vulkan_android.h>
-#else
+#elif !defined(__APPLE__)
 #include <X11/Xlib.h>
 #include <vulkan/vulkan_wayland.h>
 #include <vulkan/vulkan_xlib.h>
@@ -68,7 +66,7 @@ namespace {
         break;
 #elif defined(__APPLE__)
     case Core::Frontend::WindowSystemType::Cocoa:
-        extensions.push_back(VK_MVK_MACOS_SURFACE_EXTENSION_NAME);
+        extensions.push_back(VK_EXT_METAL_SURFACE_EXTENSION_NAME);
         break;
 #elif defined(__ANDROID__)
     case Core::Frontend::WindowSystemType::Android:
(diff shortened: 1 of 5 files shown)


liamwhite (approved):

comex:
FWIW, I ran into the same issue with CAMetalLayer, but was able to work around it by just creating a new (invisible) window to do the tests on:
```diff
diff --git a/src/yuzu/vk_device_info.cpp b/src/yuzu/vk_device_info.cpp
index 9bd1ec686..0e613e46b 100644
--- a/src/yuzu/vk_device_info.cpp
+++ b/src/yuzu/vk_device_info.cpp
@@ -24,7 +24,9 @@ Record::~Record() = default;
 void PopulateRecords(std::vector<Record>& records, QWindow* window) try {
     using namespace Vulkan;
 
-    auto wsi = QtCommon::GetWindowSystemInfo(window);
+    QWindow test_window(window);
+    test_window.setSurfaceType(QWindow::VulkanSurface);
+    auto wsi = QtCommon::GetWindowSystemInfo(&test_window);
 
     vk::InstanceDispatch dld;
     const auto library = OpenLibrary();
```

Steveice10:
> FWIW, I ran into the same issue with CAMetalLayer, but was able to work around it by just creating a new (invisible) window to do the tests on:
> 
> ```diff
> diff --git a/src/yuzu/vk_device_info.cpp b/src/yuzu/vk_device_info.cpp
> index 9bd1ec686..0e613e46b 100644
> --- a/src/yuzu/vk_device_info.cpp
> +++ b/src/yuzu/vk_device_info.cpp
> @@ -24,7 +24,9 @@ Record::~Record() = default;
>  void PopulateRecords(std::vector<Record>& records, QWindow* window) try {
>      using namespace Vulkan;
>  
> -    auto wsi = QtCommon::GetWindowSystemInfo(window);
> +    QWindow test_window(window);
> +    test_window.setSurfaceType(QWindow::VulkanSurface);
> +    auto wsi = QtCommon::GetWindowSystemInfo(&test_window);
>  
>      vk::InstanceDispatch dld;
>      const auto library = OpenLibrary();
> ```

Thanks, this works great and is a lot cleaner!

liamwhite (approved):

lat9nq (approved):

PR #10996, opened 2023-07-02
Use spans over guest memory where possible instead of copying data

Kelebek1:
#10457 aimed to reduce memory allocations when copying from guest memory, this PR goes another step and aims to eliminate the data copy entirely.

The minimum guarantee we can have of data contiguity is 1 page (4Kb), but in reality, almost all guest memory is allocated contiguously in host memory. We can exploit that and use spans directly over guest memory, rather than requiring separate reads and writes every time we interact with it. From testing various games, this is able to eliminate a non-trivial amount of copying, and can give a noticable perf boost to a bunch of games, I get ~10% on SMO's metro scene for example. XCDE seems to get similar results.

Still could use some more testing that it doesn't break any games, and if anyone has any ideas for making a cleaner interface/making it more usable then I'm all ears. I wanted it to not be inconvenient or annoying to use over current read/write calls.

### DIFF
--- a/src/core/core_timing.cpp
+++ b/src/core/core_timing.cpp
@@ -70,7 +70,7 @@ void CoreTiming::Initialize(std::function<void()>&& on_thread_init_) {
         -> std::optional<std::chrono::nanoseconds> { return std::nullopt; };
     ev_lost = CreateEvent("_lost_event", empty_timed_callback);
     if (is_multicore) {
-        timer_thread = std::make_unique<std::thread>(ThreadEntry, std::ref(*this));
+        timer_thread = std::make_unique<std::jthread>(ThreadEntry, std::ref(*this));
     }
 }
 
@@ -255,7 +255,6 @@ void CoreTiming::ThreadLoop() {
 #ifdef _WIN32
                     while (!paused && !event.IsSet() && wait_time > 0) {
                         wait_time = *next_time - GetGlobalTimeNs().count();
-
                         if (wait_time >= timer_resolution_ns) {
                             Common::Windows::SleepForOneTick();
                         } else {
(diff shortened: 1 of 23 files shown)


src/core/core_timing.cpp
```diff
@@ -243,6 +243,9 @@ std::optional<s64> CoreTiming::Advance() {
 }
 
 void CoreTiming::ThreadLoop() {
+#ifdef _WIN32
+    const auto timer_resolution_ns = Common::Windows::GetCurrentTimerResolution().count();
+#endif
```
Morph1984, line 248: This is not safe, as other programs may change this value

Kelebek1: > This is not safe, as other programs may change this value

Why would it not be "safe?" In the case where a longer timer decreases to a short timer (1.e 10ms -> 5ms), then we just wake up at the same original rate, 10ms. In the case where a shorter increases to a longer one (5ms -> 10ms), then the sleep is just delayed to the new longer timer (10ms) and gives the same result as if we had read the value and slept for 10ms.

How often would that even happen anyway, that the global timer would be changed? I feel like it would be extremely rare, and likely only done specifically by a user, which is on them. Not worth keeping the OS call in there running at 10000 times per second.

Morph1984: #10998

Morph1984:
Conflicts with #10969

src/core/core_timing.cpp
FernandoS27: I also think core timing changes are out of the scope of this PR

Squall-Leonhart:
Does this have a chance of causing File Descriptor related crashes?

Anuskuss:
Crashes SMO with async ASTC on Linux+Nvidia:
```
[  19.333744] HW.GPU <Info> ./video_core/texture_cache/texture_cache.h:QueueAsyncDecode:1223: Queuing async texture decode
Unhandled SIGSEGV at rip 0x000055aa8b047255
Segmentation fault (core dumped)
```

gidoly:
untag due to cause more frequent vertex explosion on xenoblade 3 and likely others

FernandoS27 (approved):

bunnei (approved):

Anuskuss:
> **Edit**: Also losing 30 FPS.

Performance has not returned.

Issue #10610, opened 2023-06-04
[Regression] TMNT Shredder's Revenge no longer loads on EA builds


legend800:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

yuzu Early Access 3635

### Description of Issue

Something in EA currently broke TMNT Shredder's Revenge. It's been broken for a few weeks now. 

Now with Multi-threaded off (only way to get to the opening cinematic), it's permanent black screen on boot with the following error:
Service <Error> core\hle\service\sockets\bsd.cpp:PollImpl:531: File descriptor handle=0 is not allocated

With multi-threaded on, game will boot and you'll get a few splash screens, but it'll freeze/crash at the opening cinematic as it always has. 

Works fine still on Mainline:
<img width="814" alt="2023-06-03_23_39_10-Switch" src="https://github.com/yuzu-emu/yuzu/assets/20983844/60c3a688-95c2-4300-b6ab-e6314b7d7b15">

### Expected Behavior

Game loads on single-threaded mode like it used to.

### Reproduction Steps

Boot game with multi-threaded off.

### Log File

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/11646166/yuzu_log.txt)

### System Configuration

CPU: AMD5900
GPU/Driver: nvidia 3070
RAM: 32gb
OS: Win11

Squall-Leonhart:
It hasn't frozen, its just taking longer to boot.

legend800:
Mainline is not affected. I know what I'm doing here. And I'm not talking about the cinematic. I know you have to skip that. I'm talking about boot/initial splash screens.

Looks like it does eventually load the splash screen if you have give it literally 2-3mins. BUT even once you skip the opening cinematic, the game is significantly slower to load screens (1+ to boot gameplay) and in-game the speed is 70% slower. 

So something in EA really affected single threaded emulation. Everything is perfect in Mainline - loads are snappy with full speed. 

EA:
<img width="814" alt="EA" src="https://github.com/yuzu-emu/yuzu/assets/20983844/6e2708ad-f9b9-4c17-9414-aef2e8a1efa8">
[EA.txt](https://github.com/yuzu-emu/yuzu/files/11646815/EA.txt)

Mainline:
<img width="815" alt="Mainline" src="https://github.com/yuzu-emu/yuzu/assets/20983844/9af92d74-6073-4a3a-a548-44d9eafc6767">
[...]

PR #10998, opened 2023-07-02
core_timing: Remove GetCurrentTimerResolution in CoreTiming loop

Morph1984:
Other programs may change this value, but if thousands of syscalls in this loop is undesirable, then we can just set this once.


liamwhite (approved):

Squall-Leonhart:
this appears  to have fixed the slow boot behavior in https://github.com/yuzu-emu/yuzu/issues/10610, however, the game now force closes after the intros instead.

PR #11009, opened 2023-07-03
Fix regression by unreal engine fix pr

gidoly:
It seems that pr #10953 regressed Metroid Prime Remastered rendering.

This pr bring back code that removed by #10953 pr. Reason why mentioned pr fixed smt v and other ue4 game rendering seems it enable the code that marked as comment rather than cause by code that this pr trying to bring back.
I have tested ue4 games and metroid prime to see if it regress any game but couldn't found one. since i am not dev this will need to get approve from other devs like @FernandoS27 .

Before
yuzu_jy1I1M2V9x

after
yuzu_ULOjPPZIuQ

### DIFF
--- a/src/video_core/texture_cache/texture_cache.h
+++ b/src/video_core/texture_cache/texture_cache.h
@@ -598,6 +598,13 @@ void TextureCache<P>::UnmapGPUMemory(size_t as_id, GPUVAddr gpu_addr, size_t siz
                             [&](ImageId id, Image&) { deleted_images.push_back(id); });
     for (const ImageId id : deleted_images) {
         Image& image = slot_images[id];
+        if (True(image.flags & ImageFlagBits::CpuModified)) {
+            return;
+        }
+        image.flags |= ImageFlagBits::CpuModified;
+        if (True(image.flags & ImageFlagBits::Tracked)) {
+            UntrackImage(image, id);
+        }
         if (True(image.flags & ImageFlagBits::Remapped)) {
             continue;
         }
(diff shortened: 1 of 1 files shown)


v1993 (changes requested):
This definitely looks wrong even if I don't know logic of what's going on here.

First of all, `return`ing from function there doesn't make sense, it likely should be `continue`. Second, you're now untracking image twice.

Accounting for this gives something like

```C++
        if (True(image.flags & ImageFlagBits::CpuModified)) {
            continue;
        }
        image.flags |= ImageFlagBits::CpuModified;
        if (True(image.flags & ImageFlagBits::Remapped)) {
            continue;
        }
        image.flags |= ImageFlagBits::Remapped;
        if (True(image.flags & ImageFlagBits::Tracked)) {
            UntrackImage(image, id);
        }
    }
```

Although I have no clue in what order those flags should be checked/set.

gidoly:
yeah i just reverted code change so no idea how it works. i opened pr to dev take a look on it

gidoly:
appears to break metroid with your suggested change

src/video_core/texture_cache/texture_cache.h
```diff
@@ -606,7 +610,7 @@ void TextureCache<P>::UnmapGPUMemory(size_t as_id, GPUVAddr gpu_addr, size_t siz
             UntrackImage(image, id);
         }
     }
-}
+}
```
v1993, line 613: Trailing spaces make clang-format unhappy

PR #11012, opened 2023-07-03
Fix regression by unreal engine fix pr #11009

gidoly:
It seems that pr https://github.com/yuzu-emu/yuzu/pull/10953 regressed Metroid Prime Remastered rendering.

This pr bring back code that removed by https://github.com/yuzu-emu/yuzu/pull/10953 pr. Reason why mentioned pr fixed smt v and other ue4 game rendering seems it enable the code that marked as comment rather than cause by code that this pr trying to bring back.
I have tested ue4 games and metroid prime to see if it regress any game but couldn't found one. since i am not dev this will need to get approve from other devs like @FernandoS27 .

before
yuzu_jy1I1M2V9x

after
yuzu_ULOjPPZIuQ

reopen cuz it auto closed original pr while switch branch name and less commit for bonus


FernandoS27:
I don't have time to think if unifying both approaches is great but I'll just mark it as ea test and see.

liamwhite (approved):

PR #11030, opened 2023-07-05
settings: Disable C++20 time zone path on MSVC

lat9nq:
Even though it compiles and runs fine on the latest Windows versions, older LTSC builds will crash due to lacking support somewhere in the OS.

For now just disable it for MSVC until either Microsoft fixes this or we no longer support 1809 LTSC.

Needs testing by someone using an old build of Windows 10.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -26,7 +26,8 @@ std::string GetTimeZoneString() {
 
     std::string location_name;
     if (time_zone_index == 0) { // Auto
-#if __cpp_lib_chrono >= 201907L
+#if __cpp_lib_chrono >= 201907L && !defined(_MSC_VER)
+        // TODO: Remove `!defined(_MSC_VER)` when we no longer support Windows 10 1809 LTSC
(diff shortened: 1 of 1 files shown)


Squall-Leonhart:
Drop 1809, no reason to support anyone Microsoft isn't.

[Windows 10, versions 1803 and 1809 end of servicing](https://learn.microsoft.com/en-us/lifecycle/announcements/windows-10-1803-1809-end-of-servicing)

FearlessTobi:
I don't see much reason to break compatibility if it can be worked around very easily.

lat9nq:
> I don't see much reason to break compatibility if it can be worked around very easily.

Yeah this was kind of my reasoning. Probably some day in the semi-near future we can revert this (maybe when NVIDIA stops supporting these OS versions.) For the time being, this felt like an extremely niche reason to drop compatibility for an OS, especially when the workaround already exists and works but wasn't enabled.

Morph1984 (approved):

lat9nq:
For the record, it was tested and confirmed to be needed on this Windows version.

goldenx86:
LTSC users are saved again.

PR #11041, opened 2023-07-07
ci: Download and install Vulkan SDK directly from LunarG

Morph1984:


Steveice10:
There's a GitHub Action available for this with support for caching and specifying which components you want: https://github.com/citra-emu/citra/blob/master/.github/workflows/ci.yml#L177

Morph1984:
> There's a GitHub Action available for this with support for caching and specifying which components you want: https://github.com/citra-emu/citra/blob/master/.github/workflows/ci.yml#L177

This seems to be slower than just using the installer provided by Khronos, as it has to build the component.

bunnei (approved):

Steveice10:
> > There's a GitHub Action available for this with support for caching and specifying which components you want: https://github.com/citra-emu/citra/blob/master/.github/workflows/ci.yml#L177
> 
> This seems to be slower than just using the installer provided by Khronos/LunarG, as it has to build the component.

The build is quick and the result is cached so it's usually instant.

FireBurn:
Do we actually require 1.3.256 or is that the version being used as an external dependency?

```
if (NOT YUZU_USE_EXTERNAL_VULKAN_HEADERS)
    find_package(Vulkan 1.3.256 REQUIRED)
endif()
```

Gentoo always uses the "blessed" sdk tags so we're still on 1.3.250 and wont get a new version until the next sdk release

leiserfg:
Vulkan modules have a lot of versions, but they are actually for internal development and apps and distros should use the versions of the SDK (those listed in https://vulkan.lunarg.com/sdk/home). Distributing yuzu is a headache because it hardcodes versions of some components that are not part of the SDK releases.

PR #11050, opened 2023-07-07
input_common: set `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` to 0

SuperSamus:
This allows to share the mappings between Nintendo and non-Nintendo controllers.

([Documentation of the hint, which is set to 1 by default](https://github.com/libsdl-org/SDL/blob/92b3c53c92971e685254fd89f89ce6bde8cea60e/include/SDL_hints.h#L512-L532))

Compiled locally to test. (On a Pro Controller and an Xbox controller)

(I have mainly done it because I have this hint set to 0 globally on my system via environment variable, and Yuzu's default mapping was assuming it was set to 1)

⚠ WARNING: Breaks the controller configuration for existing users who are using a Nintendo controller and are not using the direct driver.

### DIFF
--- a/src/input_common/drivers/sdl_driver.cpp
+++ b/src/input_common/drivers/sdl_driver.cpp
@@ -523,6 +523,8 @@ SDLDriver::SDLDriver(std::string input_engine_) : InputEngine(std::move(input_en
     }
 
     SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_SWITCH_PLAYER_LED, "1");
+    // Share the same button mapping with non-Nintendo controllers
+    SDL_SetHint(SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS, "0");
 
     // Disable hidapi driver for xbox. Already default on Windows, this causes conflict with native
     // driver on Linux.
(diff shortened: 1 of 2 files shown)


Morph1984:
Is it possible to retrieve the environment variable if it is set? Maybe someone else has it set to 1 and this PR assumes it is 0

SuperSamus:
To be fair, there's no reason why someone would set it to 1 with an environment variable, since it's the default.
Still, `SDL_SetHintWithPriority` exists.

german77:
Changes look good. Since it breaks profile compatibility I'm going to leave it on mainline merge for a few days. If we have too many complains we might need to reject this PR.

3356533 (approved):
Hey

Squall-Leonhart:
This is a bad idea in general, The SDL button hints system is there so that instead of numerical buttons you get the mapping as per the button on the control itself.

german77:
Were aren't going to add another setting just for this. I took this change as good just because it simplifies mapping code.

We actually use the absolute position of the button. Not the button value.

This PR only affects switch, GC and wii controllers. From them switch and GC have their custom driver and no one uses wii for yuzu. So in practical terms this PR only affects  pro controller users

OldManKain:
> Were aren't going to add another setting just for this. I took this change as good just because it simplifies mapping code.
> 
> We actually use the absolute position of the button. Not the button value.
> 
> This PR only affects switch, GC and wii controllers. From them switch and GC have their custom driver and no one uses wii for yuzu. So in practical terms this PR only affects pro controller users

We had to revert this due to all Official Switch controllers would input the swapped button in input settings.   If you press A on your controller A should show up, not B.   This person just wants everyone controller to be an xbox controller.

Squall-Leonhart:
> We had to revert this due to all Official Switch controllers would input the swapped button in input settings. If you press A on your controller A should show up, not B. This person just wants everyone controller to be an xbox controller.

This is what i meant would happen, I should have worded it better.

german77:
If we press a button it shows as a number. We have no button names for SDL
image

Squall-Leonhart:
> If we press a button it shows as a number. We have no button names for SDL image

i know, a flaw imo.

based on Kain's comment though it seems like https://github.com/libsdl-org/SDL/issues/3967 occured

german77 (approved):

bunnei (approved):

PR #11067, opened 2023-07-09
android: Don't prompt to save user data on uninstall

t895:
While this can be convenient in some scenarios, this will be a big problem for users trying to sideload different APK versions. If they forget the last one they had installed, they could have problems installing a new copy.


liamwhite (approved):

Tachi107:
Can't this be made configurable at runtime?

t895:
> Can't this be made configurable at runtime?

This isn't possible currently but there are plans to support other (more flexible) backup options.

PR #11070, opened 2023-07-11
android: Visualize disabled home options

t895:
Allow for displaying options in the home options that are disabled with messages that explain why they are disabled.

This includes reasoning for the GPU driver installation button.


FearlessTobi (approved):
Code looks fine, I can't comment on the wording

bunnei (approved):

Issue #10987, opened 2023-07-01
Master Detective Archives: Rain Code Crashing at specific moment


callmegabriell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

Early Acess 3731

### Description of Issue

the game, even while pressing ctrl + u (unlock fps) seems to be locked at 30FPS, and, after the first enemy is defeated in the deduction part, the game crashes

https://github.com/yuzu-emu/yuzu/assets/126564414/84229f14-f6bd-4a6b-bf84-614761e4dfdc

### Expected Behavior

to not crash

### Reproduction Steps

play the game normally until the first enemy

### Log File

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/11927224/yuzu_log.txt)

### System Configuration

CPU: I5-4590
GPU/Driver: RX 460 4GB
RAM: 8GB
OS: Windows 11

callmegabriell:
[01004800197F0000.zip](https://github.com/yuzu-emu/yuzu/files/11927235/01004800197F0000.zip)
here's the save file

LOuroboros:
I can confirm that this same crash is happening to me too.

Likely GPU related, given the presence of that pesky "Vulkan Device Loss" error, I pressume? It's just like it happened to me in Pokémon Legends Arceus and Persona 5 Royal *(see #9591 and #10600)*.

Here's my logs, in case they can be useful at all.
[yuzu_log_without_barrierFeedbackLoops.txt](https://github.com/yuzu-emu/yuzu/files/11931240/yuzu_log_without_barrierFeedbackLoops.txt)
[yuzu_log_with_barrierFeedbackLoops.txt](https://github.com/yuzu-emu/yuzu/files/11931241/yuzu_log_with_barrierFeedbackLoops.txt)

![yuzu_20230702_164059672](https://github.com/yuzu-emu/yuzu/assets/4485172/6f4c40bd-78a1-433e-80fd-33c61fb036de)

spidercraft:
i do not know if that can help but for some reason switching to open GL ( nvidia only shaders ), makes the game have a huge lag spike when this happen but manage to go trough the crash

spidercraft:
i do not know if that can help but for some reason switching to open GL ( nvidia only shaders ), makes the game have a huge lag spike when this happen but manage to go trough the crash

[...]

PR #11098, opened 2023-07-15
buffer_cache: Increase number of texture buffers

GPUCode:
Should fix crashes when defeating the first boss in Master Detective Archives: Rain Code (#10987)
The game configures a compute pipeline with 18 texture buffers which overflows compute_texture_buffers and causes garbage data to be written in compute_uniform_buffer_sizes.

According to the vulkan spec the limit for texture buffers is `maxPerStageDescriptorSampledImages`. Looking at the switch vulkan driver in [vulkaninfo](https://vulkan.gpuinfo.org/displayreport.php?id=15757#properties) the limit appears much higher than the current 16, so I am not exactly sure on what the correct value is. Bumped it to 32 to match the number of bits in `enabled_compute_texture_buffers` in the buffer cache.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -442,6 +442,11 @@ void BufferCache<P>::UnbindComputeStorageBuffers() {
 template <class P>
 void BufferCache<P>::BindComputeStorageBuffer(size_t ssbo_index, u32 cbuf_index, u32 cbuf_offset,
                                               bool is_written) {
+    if (ssbo_index >= channel_state->compute_storage_buffers.size()) [[unlikely]] {
+        LOG_ERROR(HW_GPU, "Storage buffer index {} exceeds maximum storage buffer count",
+                  ssbo_index);
+        return;
+    }
     channel_state->enabled_compute_storage_buffers |= 1U << ssbo_index;
     channel_state->written_compute_storage_buffers |= (is_written ? 1U : 0U) << ssbo_index;
 
@@ -464,6 +469,11 @@ void BufferCache<P>::UnbindComputeTextureBuffers() {
 template <class P>
 void BufferCache<P>::BindComputeTextureBuffer(size_t tbo_index, GPUVAddr gpu_addr, u32 size,
                                               PixelFormat format, bool is_written, bool is_image) {
+    if (tbo_index >= channel_state->compute_texture_buffers.size()) [[unlikely]] {
+        LOG_ERROR(HW_GPU, "Texture buffer index {} exceeds maximum texture buffer count",
+                  tbo_index);
+        return;
+    }
     channel_state->enabled_compute_texture_buffers |= 1U << tbo_index;
     channel_state->written_compute_texture_buffers |= (is_written ? 1U : 0U) << tbo_index;
     if constexpr (SEPARATE_IMAGE_BUFFERS_BINDINGS) {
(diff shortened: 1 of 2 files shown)


Kelebek1:
Seems like a lot of mobile devices only support 16. Everything from Huawei, Samsung, Mali and some Intel stuff, dunno if any of those are relevant, but I think 16 was set to be conservative? Aren't there total limits, maximums that can be set for all stages, in addition to per stage?

GPUCode:
NUM_TEXTURE_BUFFERS is per stage here I believe. The vulkan driver curiously reports the same number (1048576) for both per stage and total limit, but that's still far higher than 32. As far as mobile devices go, I think its better to respect those limits in the specific backends. Bumping it to 32 won't cause any more buffers to bound than before.

zeff9801:
Vulkan does not specify a minimum value for maxPerStageDescriptorSampledImages,
To determine the correct value you can use the Vulkan API to query the properties of the device. The VkPhysicalDeviceProperties structure includes a field called limits which contains information about various limits, including maxPerStageDescriptorSampledImages. You can use the vkGetPhysicalDeviceProperties function to retrieve these properties

turbervillejoseph (approved):

andreact00:
so how do resolve the problem? Someone please help me,i want to continue the game but it crashes everytime on that point

Kohega:
Confirming it does fix the crash after the battle
https://github.com/yuzu-emu/yuzu/assets/45211407/82aa8500-fad4-4e5b-b8cd-f9e16c76e8a0

image

mr-sihc:
what about this is an "early acces feature" (early acces was just said to have experimental features)

this is a game crashing bug

qweri0p:
@mr-sihc 

Early Access is a way for PR's to be tested before being accepted into mainline yuzu. You can get the Early Access builds by becoming a patron at yuzu's patreon.

While yes, this is a game breaking bug, just because you cannot play your game doesn't mean that untested fixes should be released in mainline. If you wish to play the game without becoming a patron, you should clone the origin and build yuzu yourself. The steps to that are in the readme.

mr-sihc:
> @mr-sihc
> 
> Early Access is a way for PR's to be tested before being accepted into mainline yuzu. You can get the Early Access builds by becoming a patron at yuzu's patreon.
> 
> While yes, this is a game breaking bug, just because you cannot play your game doesn't mean that untested fixes should be released in mainline. If you wish to play the game without becoming a patron, you should clone the origin and build yuzu yourself. The steps to that are in the readme.

i dont play the game, but a bug fix thats in the GIT version could be tested by a lot more people and achieve the same thing, its just a for people to also play thier games without paying

i appreciete the tutorial tho

v1993:
@mr-sihc

You seem to be confused about git version. It's actually the *least* feature-complete one and is intended to be used for development, not by end users. You can add changes on top of it, either manually or automatically; the latter is what gives us

* Mainline build = git + `mainline-merge` PRs
* Early Access build =  git + `mainline-merge` PRs + `early-access-merge` PRs

Since all PRs are publicly available one is free to make their own (unsupported!) build that includes some of them, which is what @qweri0p suggested. As you can see, suggestion to have something in git but not mainline makes no sense...

As for why use `early-access-merge` instead of `mainline-merge` - this is the kind of a change that might cause serious regressions on some setups, so initial testing is limited to those explicitly willing to do it. It's pretty common to later re-tag fixes with potential regressions as `mainline-merge` before they get accepted into repository to ensure further testing against regressions.

mr-sihc:
thanks for clearing me up on this, this is just how i am used to people managing their projects, they have an experimental git for testing, which people can use for the latest untested fixes, and sometimes tag a stable version or have a seperate stable repo and merge things when it was tested enough, so you "git" repo in this is the early access version if i am correct

liamwhite (approved):

PR #11102, opened 2023-07-16
android: fix links to re-dumping guides

v1993:
Looks like names of those sections changed on yuzu website in the meantime.

Also - I'm not completely sure, but it looks like translated versions have unnecessary backslashes?


v1993:
Removed modifications to per-language files to avoid conflicts with transifex per suggestion on discord.

FearlessTobi (approved):

PR #11144, opened 2023-07-24
Enable controller interaction in Controller Applet

flodavid:
Allows to add controllers and validate current configuration when the Controller Applet is opened.

Add a controller if possible (less than maximum) by using "Right" and remove by using "Left" (more than minimum).
Use "A" to validate and "B" to cancel. If less controllers than minimum are selected and the user tries to validate, an error message is shown. It disappears when the number of controllers is increaed.

### DIFF
--- a/dist/qt_themes/default/style.qss
+++ b/dist/qt_themes/default/style.qss
@@ -115,6 +115,10 @@ QWidget#connectedControllers {
     background: transparent;
 }
 
+QWidget#closeButtons {
+    background: transparent;
+}
+
 QWidget#playersSupported,
 QWidget#controllersSupported,
 QWidget#controllerSupported1,
--- a/dist/qt_themes/qdarkstyle/style.qss
+++ b/dist/qt_themes/qdarkstyle/style.qss
@@ -1380,6 +1380,10 @@ QWidget#connectedControllers {
     background: transparent;
 }
 
+QWidget#closeButtons {
+    background: transparent;
+}
+
 QWidget#playersSupported,
 QWidget#controllersSupported,
 QWidget#controllerSupported1,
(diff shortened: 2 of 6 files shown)


flodavid:
Here is the error message by the way. I only tested it on Windows with Snipperclips, that allows from 1 to 4 players and does not keeps controllers connected when the applet opens
image

flodavid:
```
Downloading https://www.nasm.us/pub/nasm/releasebuilds/2.16.01/win64/nasm-2.16.01-win64.zip
warning: Download failed -- retrying after 1000ms
```
I do not think I can do anything to fix this

german77:
Is not your fault. Nasm has network problems

german77:
Finally got some time to test. I was able to quickly configure the applet with the controller. However on MK8 my inputs weren't recognized anymore even when the controller itself was configured correctly. 

This might happen because the controller stay in configuration mode. Make sure DisableConfiguration or DisableAllControllerConfiguration is called when the applet exits

flodavid:
Thank you for taking the time to test. The call is already made in qt_controller.cpp: 
```cpp
QtControllerSelectorDialog::~QtControllerSelectorDialog() {
    controller_navigation->UnloadController();
    system.HIDCore().DisableAllControllerConfiguration();
}
```
I thought I reproduced the bug, but in fact the thing I did was :
- Open the Player selection screen and leave all cursors on Mario
- Select Mario for player 1 (by accidently pushing "A")
- Open the Controller Applet with "+"

And then the first controller did not seem to be recognized anymore. But by pushing "B" to unselect Mario, I could move the cursor again.
Can you check it is not what you did too ?

flodavid:
> Finally got some time to test. I was able to quickly configure the applet with the controller. However on MK8 my inputs weren't recognized anymore even when the controller itself was configured correctly.

Could you test again @german77 ?

german77 (approved):

PR #11159, opened 2023-07-25
Enable to use controller to close a game

flodavid:
- Add General setting to choose if a confirm dialog is shown when stopping
- Show the right confirm dialog if wanted
  - Reuse dialog window that ask to close the game
  - Create generic method to ask close confirmation
- Add "L + Plus + Minus" default shortcut to Stop emulation
- Add "R + Plus + Minus" default shortcut to Restart emulation
- Create generic question dialog based on TAS dialog
  - It allows controller interaction on most dialogs


flodavid:
Small fix. The return value of "question()" was always true

src/yuzu/configuration/config.cpp
```diff
@@ -112,7 +112,8 @@ const std::map<Settings::ShaderBackend, QString> Config::shader_backend_texts_ma
 // This must be in alphabetical order according to action name as it must have the same order as
 // UISetting::values.shortcuts, which is alphabetically ordered.
 // clang-format off
-const std::array<UISettings::Shortcut, 22> Config::default_hotkeys{{
+const std::array<UISettings::Shortcut, 23> Config::default_hotkeys{{
+    {QStringLiteral(QT_TRANSLATE_NOOP("Hotkeys", "Ask to Stop Emulation")),    QStringLiteral(QT_TRANSLATE_NOOP("Hotkeys", "Main Window")), {QStringLiteral("Ctrl+E"),  QStringLiteral("Plus+Minus"), Qt::WindowShortcut, false}},
```
FearlessTobi, line 116: A "Stop Emulation" Hotkey already exists.

FearlessTobi (dismissed):
I am not sure I understand what this is supposed to achieve.
Users can already bind a controller shortcut to stop their game.

src/yuzu/configuration/config.cpp
flodavid: This one asks before closing the game. It would a bit dangerous to the close the game with a controller shortcut without a warning.

FearlessTobi: I agree with v1993 here.

flodavid: I think it is the changes you wanted. I am not sure how to handle saving the settings though I tried to guess.

Here is the setting in the General tab 
image

v1993: Yep, that's what I had in mind, although I'm not sure if "always" should be the default.

Additionally, you want to show a different string to user if you're querying not because game requested exit, but because you've selected to always confirm it.

flodavid: I decided on "Always" by default because, for example in Breath of the Wild, the game does not seem to prevent closing (maybe in cutscenes or something, I did not check). So if you get a new weapon or anything and inadvertently use the shortcut to close the game, you lose your progession since the last save.

flodavid:
I decided on "Always" by default because, for example in Breath of the Wild, the game does not seem to prevent closing (maybe in cutscenes or something, I did not check). So if you get a new weapon or anything and inadvertently use the shortcut to close the game, you lose your progession since the last save.

v1993:
It *might* make sense to have an option "if triggered from controller or requested by game", although I'd like to hear others' opinions on this. I'd prefer to avoid asking by default in cases where it previously did not, but it's not critical either IMO.

flodavid:
In my opinion, the problem was the same when there was no controller shortcut. Loosing your last few minutes of gameplay by hitting F5 by mistake can be annoying

flodavid:
On the contrary, Kirby and the Forgotten Land seems to always request not to exit

flodavid:
Rebased on master and use new setting method

flodavid:
I noticed that when a simple dialog is opened, the controller stills controls the game. For applets (like Controller Applet), it is not the case. Is this a desired behavior ?

flodavid:
Screens of new elements 
image
image
image

FearlessTobi:
Please fix clang-format. Also, unrelated changes should go into their own PRs.

flodavid:
> Please fix clang-format. Also, unrelated changes should go into their own PRs.

Should be good now, sorry. I removed the ci format commit too.

flodavid:
Rebased on master to resolve conflict

german77:
Tested the PR, works well. However the default hotkey for close game is plus+minus this has a slight chance it can be used on games. Home menu is a safe button because that will never be used in a game.

liamwhite:
how about Home-+, since I occasionally confuse capture and home on my pro controller?

flodavid:
All shortcuts using Home are already used. It would surprise me if a game uses Plus+Minus. And it avoids conflicts with Steam that uses Home for Big Picture and Windows for Game Bar (likely to be on a gaming hardware running Yuzu)

german77:
Can we make it at least a 3 key combination? Like L+Plus+Minus that should be very unlikely to trigger by accident.

flodavid:
I did the change, and took advantage of it to add a shortcut to restart the game. While looking at it, I saw that OnRestartGame calls ShutdownGame(), not OnShutdownBeginDialog(), so the "Closing software" page is not shown and the game is restarted immediatly. Is it the desired behavior ?
Moreover, should we stop the play_time_manager before restarting or is it unnecessary ?

flodavid:
rebased on master and resolve conflict in main.cpp

german77:
Please fix clang format issues.

german77 (approved):

Squall-Leonhart:
This change inadvertently caused the confirmation to be raised twice when stopping via the menu. Once when initially clicking stop and again once the games list starts to populate, it also lacks safety checks to ensure Shutdown doesn't run twice (in the case of spamming the stop hotkey this could occur) and trigger a race condition that corrupts the state.

PR #11163, opened 2023-07-26
Revert "Blacklist EDS3 blending from new AMD drivers"

Moonlacer:
AMD has fixed this issue with public driver 23.7.2, so the blacklist is no longer needed

### DIFF
--- a/src/video_core/vulkan_common/vulkan_device.cpp
+++ b/src/video_core/vulkan_common/vulkan_device.cpp
@@ -554,14 +554,6 @@ Device::Device(VkInstance instance_, vk::PhysicalDevice physical_, VkSurfaceKHR 
     }
 
     sets_per_pool = 64;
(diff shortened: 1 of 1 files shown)


goldenx86:
And for those that don't update their drivers:

image

FearlessTobi:
Shouldn't we just change the check to not apply for newer drivers anymore?

Moonlacer:
> Shouldn't we just change the check to not apply for newer drivers anymore?

I feel like that would be a bad idea, as we don't want to support a potentially broken driver. All we're going to say if update to the latest version if something goes wrong, anways

goldenx86:
@FearlessTobi it would skip the current one. AMD didn't update the version number.

Moonlacer:
That too, Golden's complaining about it to AMD

goldenx86 (approved):
LGTM

PR #11165, opened 2023-07-27
vulkan_device: Return true if either depth/stencil format supports blit

Morph1984:
On devices that don't support D24S8 but supports D32S8, this should still return true if D32S8 supports src and dst blit


FearlessTobi:
What is the status of this?

liamwhite (approved):

PR #11171, opened 2023-07-27
(WIP) Compatibility Report Rewrite

behunin:
This is far from done, just opening as a draft to start discussion.

What should we add?
What should we remove?

Thoughts, and/or opinions appreciated.

Login endpoint for this is not ready yet so even if you did compile and run this, you wont get past the first page.


abouvier:
Make the compatibility report not only per game but per base game + update version.

src/core/telemetry_session.cpp
```diff
@@ -223,21 +223,29 @@ void TelemetrySession::AddInitialInfo(Loader::AppLoader& app_loader,
         AddField(Telemetry::FieldType::Session, "ProgramId", formatted_program_id);
 
         std::string name;
+        std::string version;
         app_loader.ReadTitle(name);
+        FileSys::NACP control;
+        app_loader.ReadControlData(control);
+        version = control.GetVersionString();
```
liamwhite, line 230: Does this actually read the update version?

src/core/telemetry_session.cpp
```diff
@@ -290,4 +298,89 @@ bool TelemetrySession::SubmitTestcase() {
...
+
+bool TelemetrySession::GetSpeedLimitUse() {
+    return field_collection.GetField<bool>("Renderer_UseSpeedLimit");
+}
+
+u16 TelemetrySession::GetSpeedLimit() {
+    return field_collection.GetField<u16>("Renderer_SpeedLimit");
+}
+
+bool TelemetrySession::GetDishCahcheUse() {
```
liamwhite, line 321: GetDiskCacheUse

src/core/telemetry_session.cpp
```diff
@@ -290,4 +298,89 @@ bool TelemetrySession::SubmitTestcase() {
 #endif
 }
 
+std::string TelemetrySession::GetAudioSink() {
```
liamwhite, line 301: It might be better to just collect all settings and then filter out the token and  pathnames. (See src/common/settings.cpp:LogSettings)

src/yuzu/compatdb.h
```diff
@@ -1,43 +1,359 @@
...
 enum class CompatibilityStatus {
-    Perfect = 0,
-    Playable = 1,
+    Perfect = 7,
+    Playable = 8,
     // Unused: Okay = 2,
-    Ingame = 3,
-    IntroMenu = 4,
-    WontBoot = 5,
+    Ingame = 9,
```
liamwhite, line 33: I'd imagine the comment about `Okay` can be removed?

src/core/telemetry_session.cpp
behunin: For some games yes

src/core/telemetry_session.cpp
behunin: I made all these getters so that I could pass each type to the `nlohmann::json` in the report and have it take care of formatting them all. Makes for a much cleaner report review string.

Fafner1000:
Compatibility seems to take forever to update. I have 20 have 13 of them either say not tested or ingame/intro but I have already finished those 13 games without major issues. Funny thing is that they are super popular games like zelda, xenoblade, pokemon, fire emblem. How can they be not tested?

behunin:
> As a sugestion maybe recluting someone to update the list. There is not much point to having it if is just says not tested even afters several years of the game release.

There is a rewrite of the yuzu website in the works, when it is ready there will be new data from these new reports.

src/core/telemetry_session.cpp
```diff
@@ -290,4 +298,89 @@ bool TelemetrySession::SubmitTestcase() {
...
+    return field_collection.GetField<std::string>("Audio_SinkId");
+}
+
+bool TelemetrySession::GetMultiCoreUse() {
```
FearlessTobi, line 305: Almost all of these getters are already exposed via our nornal settings interface.

src/web_service/telemetry_json.cpp
```diff
@@ -126,5 +126,4 @@ bool TelemetryJson::SubmitTestcase() {
 
     return value.result_code == WebResult::Code::Success;
 }
-
```
FearlessTobi, line 129: unnecessary change

src/yuzu/CMakeLists.txt
```diff
@@ -373,9 +372,7 @@ target_compile_definitions(yuzu PRIVATE
     -DQT_NO_URL_CAST_FROM_STRING
 )
 
-if (YUZU_ENABLE_COMPATIBILITY_REPORTING)
```
FearlessTobi, line 376: are you sure this still works with the setting removed?

src/yuzu/compatdb.cpp
```diff
@@ -1,210 +1,1075 @@
...
+           "</body>"
+           "</html>"));
+    header->setTextFormat(Qt::RichText);
+    header->setWordWrap(true);
+
+    login = tr("Login");
+    logout = tr("Logout");
+    ready = tr("Ready");
+
+    cookie_verified = new QLabel();
```
FearlessTobi, line 99: We use .ui files for this purpose. Do not declare these in the code.

src/yuzu/compatdb.cpp
```diff
@@ -1,210 +1,1075 @@
...
+    connect(report_login, &QPushButton::clicked, this, &IntroPage::Login);
+    connect(report_logout, &QPushButton::clicked, this, &IntroPage::Logout);
+    connect(login_yes, &QRadioButton::toggled, this, &QWizardPage::completeChanged);
+}
+
+bool IntroPage::isComplete() const {
+    return login_yes->isChecked();
+}
+
+void IntroPage::OnLogin(QNetworkReply* reply) {
```
FearlessTobi, line 134: We already have an http client in `common` to use for network requests.

src/yuzu/compatdb.cpp
```diff
@@ -1,210 +1,1075 @@
...
+QString graphical_none_ = QString::fromLatin1("graphical.none");
+QString audio_major_ = QString::fromLatin1("audio.major");
+QString audio_minor_ = QString::fromLatin1("audio.minor");
+QString audio_none_ = QString::fromLatin1("audio.none");
+QString comment_text = QString::fromLatin1("comment.text");
+QString comment_use = QString::fromLatin1("comment.use");
+QString screenshot_yes_ = QString::fromLatin1("screenshot.yes");
+QString screenshot_no_ = QString::fromLatin1("screenshot.no");
+
+const char* ADD = "/entity/yuzu_report?_format=json";
```
FearlessTobi, line 39: seems like these can be std::strings

src/yuzu/compatdb.h
```diff
@@ -1,43 +1,358 @@
...
+    std::string* csrf;
+    std::string* report;
+    json json_report;
+
+    QProgressBar* upload_progress;
+
+    QNetworkAccessManager& network_manager;
+
+    int imgs;
+    std::array<json, 3> dcus;
```
FearlessTobi, line 296: please give this variable a more meaningful name

src/yuzu/compatdb.cpp
```diff
@@ -1,210 +1,1075 @@
...
+    auto url = *host + LOGOUT + l_token;
+    connect(&network_manager, &QNetworkAccessManager::finished,
+            [this](QNetworkReply* reply) { OnLogout(reply); });
+
+    QNetworkRequest request;
+    request.setUrl(QUrl(QString::fromStdString(url)));
+    request.setHeader(QNetworkRequest::ContentTypeHeader,
+                      QVariant::fromValue<QString>(content_json));
+    request.setAttribute(QNetworkRequest::CookieLoadControlAttribute, QNetworkRequest::Manual);
+    request.setRawHeader(COOKIE, cookie->c_str());
```
FearlessTobi, line 205: We use JWTs for yuzu web_service authentication.
I see no good reason why cookies are needed here.

src/yuzu/configuration/configure_web.cpp
```diff
@@ -178,3 +180,7 @@ void ConfigureWeb::SetWebServiceConfigEnabled(bool enabled) {
     ui->label_disable_info->setVisible(!enabled);
     ui->groupBoxWebConfig->setEnabled(enabled);
 }
+
+void ConfigureWeb::ChangeURL() {
```
FearlessTobi, line 184: I cannot find where this is being used

src/yuzu/main.cpp
```diff
@@ -3326,7 +3332,15 @@ void GMainWindow::OnMenuReportCompatibility() {
 
     if (!Settings::values.yuzu_token.GetValue().empty() &&
         !Settings::values.yuzu_username.GetValue().empty()) {
-        CompatDB compatdb{system->TelemetrySession(), this};
+        CompatDB compatdb{Settings::values.report_api_url.GetValue(),
```
FearlessTobi, line 3335: I think the CompatDB can already access these fields via the Settings interface.

FearlessTobi (changes requested):
This PR definitely needs some major work before it can be merged.

src/common/settings.h
```diff
@@ -507,6 +507,10 @@ struct Values {
     Setting<std::string> yuzu_username{linkage, std::string(), "yuzu_username",
                                        Category::WebService};
     Setting<std::string> yuzu_token{linkage, std::string(), "yuzu_token", Category::WebService};
+    Setting<std::string> report_api_url{linkage, "https://yuzu-cms.ddev.site", "report_api_url",
```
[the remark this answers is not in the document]
behunin: This is going to replace the old compatibility report once the new one is running.

src/yuzu/compatdb.cpp
behunin: The new system uses cookies instead of JWT.
JWT are expensive to parse and so they will not be used in this new system.
Not only that but having this new system authenticate and authorize against the JWT endpoint would create to much chatter between the two services.

src/yuzu/main.cpp
behunin: CompatDB does not have access to the settings class

src/yuzu/compatdb.cpp
behunin: The problem here with using the .ui approach is that all the functions for setting and getting fields are protected.
https://doc.qt.io/qt-6/qwizardpage.html#protected-functions

This makes it a huge PITA to use the wizard properly when the fields are defined in the .ui file.

src/yuzu/compatdb.h
behunin: I don't want to rename that as the meaningful name would be `direct_creator_upload_responses`. Which is too long IMHO.
Would you settle for a comment?

src/yuzu/configuration/configure_web.cpp
behunin: You are right, my bad.
I will remove that.

behunin: Remembered why I did this, it will be removed before merge.

src/core/telemetry_session.cpp
```diff
@@ -223,21 +223,29 @@ void TelemetrySession::AddInitialInfo(Loader::AppLoader& app_loader,
...
+                name = me->GetApplicationName();
+                version = me->GetVersionString();
             }
         }
 
         if (!name.empty()) {
             AddField(Telemetry::FieldType::Session, "ProgramName", name);
         }
+        if (!version.empty()) {
+            AddField(Telemetry::FieldType::Session, "ProgramVersion", version);
```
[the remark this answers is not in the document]
behunin: Not sure what you mean.... Why would it need that?

src/yuzu/compatdb.h
german77: I see you aren't used to nintendo big names. But yeah I agree with this one dcus has no meaning to me. I would prefer to use a typedef like `using DirectCreatorUploadResponses = std::array<json, 3>;` that way the object type will provide a meaningful name while the variable name can be short

behunin: Took your advise along with a comment linking to the document page.

Raistlin43:
Fix Crash problems in games that start in NCE with 0 fps like Dbz Kakarot.

The problem is shown here with explanatory videos... 
https://github.com/yuzu-emu/yuzu/issues/12246

According to a developer it is a mapping problem and can be solved by rooting the device by adding this sunfish:/ # echo 524288 > /proc/sys/vm/max_map_count

But it is necessary to find a solution without the need to rotate... There are many games that when activating the NCE the fps goes to 0 and crashes.

src/common/settings.h
```diff
@@ -566,6 +566,10 @@ struct Values {
...
```
FearlessTobi, line 569: Before this gets merged, we should make sure that we are not linking to some random third-party website in our code.

src/common/settings.h
```diff
@@ -566,6 +566,10 @@ struct Values {
...
+                                        Category::WebService};
+    Setting<std::string> yuzu_cookie{linkage, std::string(), "yuzu_cookie", Category::WebService};
```
FearlessTobi, line 571: We already have a jwt-based system for all web service authentication.

src/yuzu/compatdb.cpp
FearlessTobi: Can you elaborate more on this? You should be able to access those fields by using the ui pointer of qt.
In fact that's what the current compatdb wizard is doing to access various fields.

src/yuzu/compatdb.cpp
FearlessTobi: I don't really agree with this. 
If our client really has a memory-leak we should just fix it instead, since significant parts of yuzu still use it for communication.

src/yuzu/compatdb.cpp
```diff
@@ -1,210 +1,1111 @@
...
+                           os_, settings_values_, *network_manager));
+    setPage(Page_Screenshots, new ScreenshotsPage(&host, &cookie, &csrf_token, &report,
+                                                  screenshot_path_, *network_manager));
+    setPage(Page_ThankYou, new ThankYouPage());
+
+    network_errors = new QLabel();
+    network_errors->setWordWrap(true);
+    setSideWidget(network_errors);
+
+    connect((IntroPage*)page(Page_Intro), &IntroPage::UserChange, this, &CompatDB::OnUserChange);
```
FearlessTobi, line 1030: we do not use c-style casts

src/yuzu/compatdb.cpp
```diff
@@ -1,210 +1,1111 @@
...
+    done(QDialog::Rejected);
+}
+
+void CompatDB::ReportMade(QByteArray report_) {
+    report = report_.toStdString();
 }
 
-void CompatDB::EnableNext() {
-    button(NextButton)->setEnabled(true);
+void CompatDB::CSRF(QByteArray csrf) {
```
FearlessTobi, line 1109:
```suggestion
void CompatDB::SetCSRFToken(const QByteArray csrf) {
```

src/yuzu/compatdb.h
```diff
@@ -1,43 +1,380 @@
-// SPDX-FileCopyrightText: 2017 Citra Emulator Project
+// SPDX-FileCopyrightText: 2023 Yuzu Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #pragma once
 
-#include <memory>
-#include <QFutureWatcher>
+#include <QButtonGroup>
```
FearlessTobi, line 6: It seems like some of these can be forward declard (or removed entirely after migrating to using .ui files).

src/yuzu/configuration/configure_web.cpp
```diff
@@ -93,6 +94,7 @@ void ConfigureWeb::SetConfiguration() {
         ui->username->setText(QString::fromStdString(Settings::values.yuzu_username.GetValue()));
     }
 
+    ui->edit_url->setText(QString::fromStdString(Settings::values.report_api_url.GetValue()));
```
FearlessTobi, line 97: Why does the user need to specify their own compat report url?

src/yuzu/main.cpp
```diff
@@ -3635,7 +3641,65 @@ void GMainWindow::OnMenuReportCompatibility() {
 
     if (!Settings::values.yuzu_token.GetValue().empty() &&
         !Settings::values.yuzu_username.GetValue().empty()) {
-        CompatDB compatdb{system->TelemetrySession(), this};
+        Loader::AppLoader& app_loader = system->GetAppLoader();
```
FearlessTobi, line 3644: Almost all of this data here is already gathered in `core/telemetry_session.cpp`.
You can just access the data from there instead.

FearlessTobi (changes requested):
Some comments...
I also unresolved some of the older conversations where there are still open questions.

src/yuzu/compatdb.cpp
FearlessTobi: I just asked narr again, and he doesn't know about such a leak in our http client.

behunin: The leak was specifically around images being sent for the discord presence API.
Another problem I have with using the cpp-http client is that it lacks a way to integrate with the progress bar that qt provides. While I am sure I could find a way around that, I would much rather not waste time on that when it is already done with the qt client.

src/common/settings.h
behunin: Correct, this is only a temporary link. It will be changed to the proper url when we are ready to open reporting to the public.

src/common/settings.h
behunin: While that is true, using jwt from that system would create a lot of chatter between the reporting server and the forum server which is not desirable. The jwt is still used to authenticate the user, but after the user is authenticated they will get a cookie for use with all report API calls.

src/yuzu/compatdb.cpp
behunin: I understand why c-style casts are not used in functional code due to how they resolve to a different c++ cast depending on context but in this case it does not matter because the cast is just so that the proper function can be referenced in the second argument to connect.

src/yuzu/configuration/configure_web.cpp
behunin: This is only temporary and will be removed before merge.

src/yuzu/main.cpp
behunin: I wasn't happy with doing that as I had to create getters for each specific data point that I wanted. Both you and byte were not happy with that approach so I just said forget telemetry.

behunin:
https://github.com/yuzu-emu/yuzu/pull/11171#discussion_r1421674398

The problem with using the .ui approach with the QtWizard is that the functions for reading/writing field data are protected. https://doc.qt.io/qt-6/qwizardpage.html#protected-functions
This is why the current wizard does not gather the playability grade correctly. It has no access to the fields it needs to determine playability.

src/yuzu/compatdb.cpp
behunin: Yes that is what is being done currently but that is not the way the QtWizard was designed to be used.
Fields should be registered and set per page and then called with the field(name) function.

src/yuzu/compatdb.cpp
german77: We don't use c-style cast at all. In all yuzu codebase there's not a single one.

behunin: Ah, gotcha

FearlessTobi (changes requested):

src/yuzu/compatdb.cpp
```diff
@@ -1,210 +1,1122 @@
...
+bool CommentPage::isComplete() const {
+    if (comment->checkedId() == 1) {
+        return true;
+    }
+    if (editor->document()->isEmpty()) {
+        return false;
+    }
+    return true;
+}
+// clang-format off
```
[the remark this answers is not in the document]
behunin: Screenshot_20231211_142558

This is what clang format suggests. It looks ridiculous.

Issue #10285, opened 2023-05-14
The Legend of Zelda Skyward Sword HD visual bug/glitch


Trsxh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

Mainline 1431

### Description of Issue

A few seconds after launching the game the whole system will look like this until you close the emulation and restart the explorer process
![Captura de pantalla 2023-05-13 231539](https://github.com/yuzu-emu/yuzu/assets/80502661/c85f2717-a047-49ee-b074-c2dd45175912)
![Captura de pantalla 2023-05-13 231958](https://github.com/yuzu-emu/yuzu/assets/80502661/96018298-4559-48c1-90bd-16f4d0dc26b5)
![Captura de pantalla 2023-05-13 232016](https://github.com/yuzu-emu/yuzu/assets/80502661/89d62fa6-5e84-4123-a2d7-656667cb769b)

### Expected Behavior

It definitely should not look like this

### Reproduction Steps

Start the game and after a few seconds it starts to look like this

### Log File

[yuzu_log.zip](https://github.com/yuzu-emu/yuzu/files/11471018/yuzu_log.zip)

### System Configuration

CPU: AMD Ryzen 5 1600
GPU/Driver: AMD RX 570 / AMD Adrenalin 23.4.3 
RAM: 12GB
OS: Windows 11

Squall-Leonhart:
This is because your explorer process is consuming a  bunch of vram due to a bug, you might find it behaves ok in the next mainline build thanks to https://github.com/yuzu-emu/yuzu/pull/10288

Amer-xyz:
They mention this issue in the april progress report, the suggestion being staying on driver 23.3.1

blakbin:
im using EA3602 this happen on vulkan. Whole screen will look like that until im closing Yuzu then right clicking the desktop 3~5times and Adrenaline 'Driver Timeout' popup came out.
![sk-error](https://github.com/yuzu-emu/yuzu/assets/3345247/97b94c42-4c96-4115-8041-55ac9adf5290)

while using openGL-GLSL the buggy part only inside the emulator, not the whole screen.
![sk-errorg1](https://github.com/yuzu-emu/yuzu/assets/3345247/6cc6d7d3-f2b1-48bb-a738-39d09f07bb0b)
[...]

PR #11182, opened 2023-07-30
Revert the Revert! Reverts the Revert of "Blacklist EDS3 blending from new AMD drivers"

Moonlacer:
Reverts #11163 

*Sighs*, turns out AMD hasn't fixed EDS3 blending for Polaris, they only fixed it for RDNA2 in 23.5.2
Thanks to [Blakbin](https://github.com/yuzu-emu/yuzu/issues/10285#issuecomment-1656729216) for bringing this to my attention so quickly

Further testing will be done before we implement a proper workaround/fix

### DIFF
--- a/src/video_core/vulkan_common/vulkan_device.cpp
+++ b/src/video_core/vulkan_common/vulkan_device.cpp
@@ -558,7 +558,7 @@ Device::Device(VkInstance instance_, vk::PhysicalDevice physical_, VkSurfaceKHR 
         !features.shader_float16_int8.shaderFloat16 &&
         properties.properties.driverVersion >= VK_MAKE_API_VERSION(0, 2, 0, 258)) {
         LOG_WARNING(Render_Vulkan,
-                    "AMD's GCN4 architecture has broken extendedDynamicState3ColorBlendEquation");
+                    "AMD GCN4 has broken extendedDynamicState3ColorBlendEquation");
         features.extended_dynamic_state3.extendedDynamicState3ColorBlendEnable = false;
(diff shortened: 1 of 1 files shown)


liamwhite (approved):

FernandoS27:
Just add a check for Polaris instead, we got one in it.

src/video_core/vulkan_common/vulkan_device.cpp
```diff
@@ -554,6 +554,14 @@ Device::Device(VkInstance instance_, vk::PhysicalDevice physical_, VkSurfaceKHR
     }
 
     sets_per_pool = 64;
+    if (extensions.extended_dynamic_state3 && is_amd_driver &&
+        properties.properties.driverVersion >= VK_MAKE_API_VERSION(0, 2, 0, 270)) {
```
FernandoS27, line 558: replace this for (!features.shader_float16_int8.shaderFloat16)

FernandoS27: that makes it only affect polaris & older.

src/video_core/vulkan_common/vulkan_device.cpp
```diff
@@ -554,6 +554,14 @@ Device::Device(VkInstance instance_, vk::PhysicalDevice physical_, VkSurfaceKHR
...
+        LOG_WARNING(Render_Vulkan,
+                    "AMD drivers after 23.5.2 have broken extendedDynamicState3ColorBlendEquation");
```
goldenx86, line 560: Since newer cards are fixed on the latest drivers, I would replace this for:
"GCN4 has broken extendedDynamicState3ColorBlendEquation"

PR #11219, opened 2023-08-05
Allow searching by a substring of the title ID

zeltermann:
Now I can just type `0001000` and easily find qlaunch.

### DIFF
--- a/src/yuzu/game_list.cpp
+++ b/src/yuzu/game_list.cpp
@@ -214,13 +214,17 @@ void GameList::OnTextChanged(const QString& new_text) {
             const int children_count = folder->rowCount();
             for (int j = 0; j < children_count; ++j) {
                 ++children_total;
+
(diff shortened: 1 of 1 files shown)


src/yuzu/game_list.cpp
```diff
@@ -214,13 +214,17 @@ void GameList::OnTextChanged(const QString& new_text) {
             const int children_count = folder->rowCount();
             for (int j = 0; j < children_count; ++j) {
                 ++children_total;
+
                 const QStandardItem* child = folder->child(j, 0);
+
+                qulonglong program_id = child->data(GameListItemPath::ProgramIdRole).toULongLong();
```
FearlessTobi, line 220:
```suggestion
                const auto program_id = child->data(GameListItemPath::ProgramIdRole).toULongLong();
```

liamwhite (approved):

PR #11221, opened 2023-08-05
service: pctl: Implement functions needed for QLaunch

german77:
Same as with other PR. Verified with homebrew on HW. Get's a bit closer to launching QLaunch.

### DIFF
--- a/src/core/hle/service/pctl/pctl_module.cpp
+++ b/src/core/hle/service/pctl/pctl_module.cpp
@@ -6,6 +6,7 @@
 #include "core/file_sys/control_metadata.h"
 #include "core/file_sys/patch_manager.h"
 #include "core/hle/service/ipc_helpers.h"
+#include "core/hle/service/kernel_helpers.h"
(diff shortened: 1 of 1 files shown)


zeltermann:
@german77 What services are you currently working on but has not pushed yet? I have a few other services I stubbed in while trying to launch the Mii editor, and I can make a PR if you don't mind.

Also, is there a better way to launch applets like QLaunch? I've just been loading the appropriate NCA file containing the QLaunch title on yuzu.

german77:
You can see all the changes here, this includes implementations for mii applet https://github.com/german77/yuzu/tree/home_menu2 

I'm making PR as I review the changes with proper time and care to avoid introducing errors. Using the NCA is the best method I have at the moment. Right now Qlaunch works on the EULA screen but has memory errors booting the actual home menu.

Mii applet doesn't has any error but it just sits idling. I'm not sure why that happens

german77:
If you got something to render I would be interested.

liamwhite (approved):

PR #11278, opened 2023-08-12
Mark accelerated DMA destination buffers and images as GPU-modified

### LINKED ISSUE #10080: Sid Meier's Civilization VI - The game crashes with Vulcan and looks glitchy with OpenGL

### Affected Build(s)

1407

### Description of Issue

Sid Meiers Civilization VI [010044500C182000][v0] is unplayable:

1. If I run it with Vulcan backend, it crashes when I start a new game (during the loading screen with a Civ leader) ([yuzu_log_Vulcan_crash.txt](https://github.com/yuzu-emu/yuzu/files/11301075/yuzu_log_Vulcan_crash.txt))
2. If I  run it with OpenGL, it boots successfully to the map, but some shaders are missing and black holes are on the image (opengl_glitch)

### Expected Behavior

The game just runs and looks like on the Switch

### Reproduction Steps

Steps to reproduce both issues:
1. Start emulation
2. Skip videos
3. Press Continue
4. Single player > Play now
5. Wait for it
6. (for opengl) Press Start game

### Log File

(attached above)

### System Configuration

CPU: Core i7-1195G7
GPU/Driver: integrated
RAM: 16GB
OS: Windows 10

Kelebek1:
The accelerated engine functions in 2d/dma/upload do GPU <-> GPU copies where possible, skipping a write to the CPU-side destination address, but we're currently not marking the destination buffer/image as being modified in DMA, which they should be, otherwise any future reads will not trigger a download and read the wrong data.

The non-accelerated paths in all engines are likely also causing some issues, due to the fact that they don't mark the relevant image/buffer as CPU-modified. I'm not sure how to correctly find them from the various engines, and if this is really even an issue, so I'll leave that to @FernandoS27 to check.

We also now fallback to a software blit when encountering an address which isn't valid for the given image's format. That case needs some more research.

Closes #10080

### DIFF
--- a/src/video_core/renderer_opengl/gl_rasterizer.cpp
+++ b/src/video_core/renderer_opengl/gl_rasterizer.cpp
@@ -1335,7 +1335,8 @@ bool AccelerateDMA::DmaBufferImageCopy(const Tegra::DMA::ImageCopy& copy_info,
     }
     const u32 buffer_size = static_cast<u32>(buffer_operand.pitch * buffer_operand.height);
     static constexpr auto sync_info = VideoCommon::ObtainBufferSynchronize::FullSynchronize;
-    const auto post_op = VideoCommon::ObtainBufferOperation::DoNothing;
+    const auto post_op = IS_IMAGE_UPLOAD ? VideoCommon::ObtainBufferOperation::DoNothing
+                                         : VideoCommon::ObtainBufferOperation::MarkAsWritten;
     const auto [buffer, offset] =
         buffer_cache.ObtainBuffer(buffer_operand.address, buffer_size, sync_info, post_op);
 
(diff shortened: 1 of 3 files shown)


src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -839,8 +840,14 @@ bool AccelerateDMA::DmaBufferImageCopy(const Tegra::DMA::ImageCopy& copy_info,
     const std::span copy_span{&copy, 1};
 
     if constexpr (IS_IMAGE_UPLOAD) {
+        image->flags &= ~VideoCommon::ImageFlagBits::CpuModified;
+        image->flags |= VideoCommon::ImageFlagBits::GpuModified;
```
FernandoS27, line 844: this should be done through the texture cache as the modification tic must be updated.

src/video_core/renderer_opengl/gl_rasterizer.cpp
```diff
@@ -1344,8 +1345,14 @@ bool AccelerateDMA::DmaBufferImageCopy(const Tegra::DMA::ImageCopy& copy_info,
     const std::span copy_span{&copy, 1};
 
     if constexpr (IS_IMAGE_UPLOAD) {
+        image->flags &= ~VideoCommon::ImageFlagBits::CpuModified;
+        image->flags |= VideoCommon::ImageFlagBits::GpuModified;
```
FernandoS27, line 1349: ditto

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -839,8 +840,12 @@ bool AccelerateDMA::DmaBufferImageCopy(const Tegra::DMA::ImageCopy& copy_info,
     const std::span copy_span{&copy, 1};
 
     if constexpr (IS_IMAGE_UPLOAD) {
+        texture_cache.PrepareImage(image_id, true, false);
```
FernandoS27, line 843: PrapareImage is private member method, so ypu may want to expose it.

liamwhite (approved):

Issue #10802, opened 2023-06-16
Rendering Regression on Dust: An Elsyian Tale on EA3649


JWatersMeet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

Early Access 3649

### Description of Issue

The menu screen is corrupted on Dust: An Elsian Tale on EA 3649. This doesn't happen on EA 3648.
Wonder if it's due to  #10668 ?
Dust: AET is an XNA game, Bastion (which is also XNA) seems to suffer from extreme flickering in this build too.

https://github.com/yuzu-emu/yuzu/assets/136824455/5f28e7da-ac5f-463a-b9bb-a2993862b321

### Expected Behavior

Should render as per the below, in EA 3648.

https://github.com/yuzu-emu/yuzu/assets/136824455/0b3e1875-bb0f-4438-abc9-3f4c071667c7

### Reproduction Steps

Run Dust: An Elsyian Tale. Wait for menu to fully load. Notice corruption.

### Log File

To follow.

### System Configuration

CPU: Ryzen 5900HX
GPU/Driver: RTX 3080 - 31.0.15.3118
RAM: 16GB
OS: Windows 11

JWatersMeet:
[yuzu_log_dust.txt](https://github.com/yuzu-emu/yuzu/files/11771382/yuzu_log_dust.txt)

Log now attached.

JWatersMeet:
Can confirm this has resolved the issue with Dust AET, many thanks @Kelebek1 .
Will raise a seperate issue for Bastion, looks unrelated.

PR #11349, opened 2023-08-22
video_core: Fix moltenvk crash on macos 

vonchenplus:
1. Set vertex buffer num to 16, because mvk have when using more than 16.
2. Add missing scissor update when viewport scale offset disable.
3. Fix d24r8/s8d24 convert shader build error in moltenvk

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -989,6 +989,19 @@ void RasterizerVulkan::UpdateScissorsState(Tegra::Engines::Maxwell3D::Regs& regs
     if (!state_tracker.TouchScissors()) {
         return;
     }
+    if (!regs.viewport_scale_offset_enabled) {
+        const auto x = static_cast<float>(regs.surface_clip.x);
+        const auto y = static_cast<float>(regs.surface_clip.y);
+        const auto width = static_cast<float>(regs.surface_clip.width);
+        const auto height = static_cast<float>(regs.surface_clip.height);
+        VkRect2D scissor;
+        scissor.offset.x = static_cast<u32>(x);
+        scissor.offset.y = static_cast<u32>(y);
+        scissor.extent.width = static_cast<u32>(width != 0.0f ? width : 1.0f);
+        scissor.extent.height = static_cast<u32>(height != 0.0f ? height : 1.0f);
+        scheduler.Record([scissor](vk::CommandBuffer cmdbuf) { cmdbuf.SetScissor(0, scissor); });
+        return;
+    }
     u32 up_scale = 1;
     u32 down_shift = 0;
     if (texture_cache.IsRescaling()) {
--- a/src/video_core/buffer_cache/buffer_cache_base.h
+++ b/src/video_core/buffer_cache/buffer_cache_base.h
@@ -62,7 +62,11 @@ using BufferId = SlotId;
 using VideoCore::Surface::PixelFormat;
 using namespace Common::Literals;
 
+#ifdef __APPLE__
+constexpr u32 NUM_VERTEX_BUFFERS = 16;
+#else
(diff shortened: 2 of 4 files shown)


Kelebek1:
I already did this and it breaks games, I had to revert it.

gidoly:
seems cause no issues from my end might worth ea tag and let people test it

vonchenplus:
> I already did this and it breaks games, I had to revert it with [e681f56](https://github.com/yuzu-emu/yuzu/commit/e681f5678c3df6d2e1779cb824615517ddae46d7).
> 
> Did you test this PR with a load of games?

Which game had the problem before?

Kelebek1:
Tested and can confirm this breaks #10802 again

Kelebek1:
Ok so the dirty flags only track changes to the 3d registers affecting binding, they say nothing about the state of the data being pointed to by a bind region. The same vertex buffer binding may be valid between draws, but was written to by the guest and its data needs uploading regardless of rebinding.

I saw a few things though, firstly binding sizes of 0 are already excluded in UpdateVertexBuffers, and a NULL_BINDING is used instead. We can check for that null binding ID instead just to make it more clear. I think this will work more as you want and fixes the bug:

```diff
diff --git a/src/video_core/buffer_cache/buffer_cache.h b/src/video_core/buffer_cache/buffer_cache.h
index f0f450edb..15b919b83 100644
--- a/src/video_core/buffer_cache/buffer_cache.h
+++ b/src/video_core/buffer_cache/buffer_cache.h
@@ -758,9 +758,20 @@ void BufferCache<P>::BindHostVertexBuffers() {
     auto& flags = maxwell3d->dirty.flags;
     for (u32 index = 0; index < NUM_VERTEX_BUFFERS; ++index) {
         const Binding& binding = channel_state->vertex_buffers[index];
+        if (binding.buffer_id == NULL_BINDING.buffer_id) {
+            continue;
+        }
+
         Buffer& buffer = slot_buffers[binding.buffer_id];
         TouchBuffer(buffer, binding.buffer_id);
+
+        auto cpu_modified = IsRegionCpuModified(binding.cpu_addr, binding.size);
+        if (!flags[Dirty::VertexBuffer0 + index] && !cpu_modified) {
+            continue;
+        }
+
         SynchronizeBuffer(buffer, binding.cpu_addr, binding.size);
+
         if (!flags[Dirty::VertexBuffer0 + index]) {
             continue;
         }
@@ -774,8 +785,6 @@ void BufferCache<P>::BindHostVertexBuffers() {
     if (any_valid) {
         host_bindings.max_index++;
         for (u32 index = host_bindings.min_index; index < host_bindings.max_index; index++) {
-            flags[Dirty::VertexBuffer0 + index] = false;
-
             const Binding& binding = channel_state->vertex_buffers[index];
             Buffer& buffer = slot_buffers[binding.buffer_id];
```

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -989,6 +989,19 @@ void RasterizerVulkan::UpdateScissorsState(Tegra::Engines::Maxwell3D::Regs& regs
...
+    if (!regs.viewport_scale_offset_enabled) {
+        const auto x = static_cast<float>(regs.surface_clip.x);
+        const auto y = static_cast<float>(regs.surface_clip.y);
+        const auto width = static_cast<float>(regs.surface_clip.width);
+        const auto height = static_cast<float>(regs.surface_clip.height);
+        VkRect2D scissor;
+        scissor.offset.x = x;
+        scissor.offset.y = y;
+        scissor.extent.width = width != 0.0f ? width : 1.0f;
+        scissor.extent.height = height != 0.0f ? height : 1.0f;
```
Kelebek1, line 1001: You'll need to consider rescaling here, like it is below.

liamwhite (approved):

PR #11350, opened 2023-08-22
ui: Added padding to the reset button

BenjaminHalko:
Some buttons did not have enough padding, now they do!

Before:
image

After:
image


BenjaminHalko:
Apparently adding padding to ALL buttons breaks a few things, so I changed it to only add padding to the reset button, which was the main issue in the first place.

liamwhite (approved):

PR #11380, opened 2023-08-26
android: Settings rework

t895:
Before I started work on getting per-game settings ready, I needed to clean up issues with duplicated code for reading from settings. As I went through I noticed additional functionality that we haven't implemented for android settings (at least not exposed to the user) like temporary or paired settings. I got pretty carried away and ended up with a full rework of the existing system.

After this, per-game settings should be fairly easy to implement.

Here's a list of changes
- Create interface for getting settings from native code
- The settings activity is now using the navigation component which comes with simpler navigation logic and more consistent material transition animations
animations

- Settings that are marked as paired in native code will now be reflected in the android UI with hide/show animations
hideshow

- All emulation settings are now searchable via a button in the first page of Advanced Settings
search

- All settings dialogs now restore state properly (except for date/time) and survive configuration changes
staterestore


skylineyyds:
NCE when？😂

goldenx86:
Source is there, code it in.

t895:
(Forgot to mention) I'd also like @lat9nq to take a look at the code I added to the common settings to make sure I'm not interfering with the desktop app.

src/common/settings.h
```diff
@@ -510,6 +510,10 @@ struct Values {
 
     // Add-Ons
     std::map<u64, std::vector<std::string>> disabled_addons;
+
+    // Android
+    Setting<bool> picture_in_picture{linkage, true, "picture_in_picture", Category::Android};
+    Setting<s32> screen_layout{linkage, 5, "screen_layout", Category::Android, Specialization::Default, true, true};
```
lat9nq, line 516: I would like to see these moved out of common into the respective frontend if they do not pertain to yuzu generally. The Qt frontend's UISettings::Values struct harbors settings specific to that frontend as an example. I'm not sure if the Android frontend already has like infrastructure or not, but it would be ideal to set that up for not only these two settings but also more in the future.

For clarity sake, leaving the Android enumeraton value in Settings::Category is fine by me since it's an enum class.

t895: I just pushed a change so that android settings live in a separate file now. Could you let me know if that follows an appropriate pattern? Also after this I'm very interested in moving all android settings currently stored in shared preferences to the ini file.

goldenx86 (approved):
LGTM

PR #11416, opened 2023-08-30
android: Support dynamic app shortcuts

t895:
After launching any game, an associated app shortcut is added to the launcher. You can long press the app icon and add any of these generated shortcuts to your home screen.

shortcut

### DIFF
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/adapters/GameAdapter.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/adapters/GameAdapter.kt
@@ -3,6 +3,8 @@
 
 package org.yuzu.yuzu_emu.adapters
 
+import android.content.Intent
+import android.graphics.drawable.BitmapDrawable
 import android.net.Uri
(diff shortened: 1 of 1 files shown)


skylineyyds:
NCE

sakuka47:
It works like a charm, whether directly on the home screen or in the application page.

FearlessTobi (approved):

Arkuarius:
Doesn't work on MIUI 💀

RenanSD007:
> Doesn't work on MIUI 💀

I tested it here and it seems that the default Launcher doesn't let you add dynamic shortcuts, I tested it with another launcher (Lawnchair and Nova Launcher) and it worked .

Arkuarius:
> > Doesn't work on MIUI 💀
> 
> I tested it here and it seems that the default Launcher doesn't let you add dynamic shortcuts, I tested it with another launcher (Lawnchair and Nova Launcher) and it worked.

I see...

Still though, I would've liked a separate button inside Yuzu for adding a shortcut to the homescreen, @t895.

PR #11425, opened 2023-08-31
android: Use StateFlow instead of LiveData

t895:
StateFlows are the Kotlin standard way of updating UI state and have extra features that we could take advantage of in the future.


FearlessTobi (commented):
The code looks fine, but I have no experience with this.
So please just make sure everything is tested thoroughly.

t895:
Just a quick note here - I need to do some adjustments to this before a merge. I found a bug yesterday that improperly updates the game list. I also should change the way I'm using SavedStateHandle to only restore certain kinds of UI state, not all.

liamwhite (approved):

PR #11428, opened 2023-09-01
Rework ADSP into a wrapper for apps

Kelebek1:
Currently the ADSP is really just the audio renderer in disguise, because that's the only app implemented. This change just moves the AudioRenderer to its own thing inside the ADSP, and makes it easier to add other apps in future. 

Shouldn't be any functional changes here.


src/audio_core/adsp/adsp.h
```diff
@@ -0,0 +1,53 @@
+// SPDX-FileCopyrightText: Copyright 2022 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <memory>
+#include <mutex>
```
FearlessTobi, line 7: i dont see this being used in this header

src/audio_core/adsp/adsp.h
```diff
@@ -0,0 +1,53 @@
+// SPDX-FileCopyrightText: Copyright 2022 yuzu Emulator Project
```
FearlessTobi, line 1: 2023

src/audio_core/adsp/adsp.cpp
```diff
@@ -0,0 +1,18 @@
+// SPDX-FileCopyrightText: Copyright 2022 yuzu Emulator Project
```
FearlessTobi, line 1: 2023

src/audio_core/adsp/apps/audio_renderer/audio_renderer.h
```diff
@@ -0,0 +1,115 @@
...
+
+namespace AudioCore {
+namespace Sink {
+class Sink;
+}
+
+namespace ADSP::AudioRenderer {
+
+enum Message : u32 {
+    /* 0x00 */ Invalid,
```
FearlessTobi, line 37: you can assign these values directly instead of using comments

src/audio_core/adsp/mailbox.cpp
FearlessTobi, line 1: should probably be deleted?

src/audio_core/adsp/apps/audio_renderer/audio_renderer.h
Kelebek1: I can, but they're all messy and not lined up if you do that. Looks better on the side, but I'll change it.

liamwhite (approved):

PR #11447, opened 2023-09-04
common: add a compile time option to allow disabling portable mode

### LINKED ISSUE #11441: [Bug?]How to stop yuzu installing files into ~/user?

### Affected Build(s)

Tested: 
yuzu(extra), yuzu-mainline-git(AUR)

### Description of Issue

Whenever the ~/user folder is created and yuzu is launched, it uses ~/user folder for data rather than ~/.local/share/yuzu. I have to think this is a bug from the Windows build relating to: https://www.reddit.com/r/yuzu/comments/betu10/comment/el8y2e4/?utm_source=share&utm_medium=web2x&context=3; is there any way I get around this for now? I keep a lot of personal files in ~/user so it is a real pain.

### Expected Behavior

yuzu uses ~/.local/share/yuzu folder for data, not ~/user.

### Reproduction Steps

Create a ~/user folder, launch yuzu, watch ~/user folder get populated as if it were ~/.local/share/yuzu.

### Log File

No logs, here is the output of `ls ~/user` after launching yuzu:
`amiibo  cache  config  dump  keys  load  log  nand  screenshots  sdmc  shader  tas`

### System Configuration

CPU: Ryzen 5600x
GPU/Driver: RTX 3070 535.104.05
RAM: 16GB ddr4
OS: Arch Linux

xcfrg:
yuzu checks for a `user` folder in the current working directory on non-windows and non-android platforms. At least on linux, the CWD will default to `~` if ran through an application launcher, and a folder at `~/user` will enable portable mode.

As installations with a package manager will not make use of portable mode, we should expose a compile time option to allow package maintainers disable this undesirable behavior.

closes #11441

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -151,6 +151,10 @@ add_library(common STATIC
     zstd_compression.h
 )
 
+if (YUZU_ENABLE_PORTABLE)
+    add_compile_definitions(YUZU_ENABLE_PORTABLE)
+endif()
+
 if (WIN32)
   target_sources(common PRIVATE
     windows/timer_resolution.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,8 @@ option(YUZU_ENABLE_LTO "Enable link-time optimization" OFF)
 
 option(YUZU_DOWNLOAD_TIME_ZONE_DATA "Always download time zone binaries" OFF)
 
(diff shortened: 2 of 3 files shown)


src/common/CMakeLists.txt
```diff
@@ -151,6 +151,8 @@ add_library(common STATIC
     zstd_compression.h
 )
 
+add_compile_definitions(YUZU_ENABLE_PORTABLE)
```
Tachi107, line 154:
```suggestion
if (YUZU_ENABLE_PORTABLE)
    add_compile_definitions(YUZU_ENABLE_PORTABLE)
endif()
```

Issue #6253, opened 2021-04-28
SSBU, starting any Mob Smash mode causes guest exception


lat9nq:
In Smash Ultimate, trying to start any of the Mob Smash modes in the "Games & More" category will cause the guest to throw a C++ exception and a guest softlock. Brief reproduction steps would be to select a Mob Smash mode, select stage, character, and start the match. It will crash while loading (should see loading hints on screen).

Tested on b096ec68cdbf6f1064a8b6b855489d38c3e59f6a, though this has never worked from what I remember.

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/6389276/yuzu_log.txt)

rubenguip:
Using version 1226 of yuzu (from today 6 November 2022), Mob smash stays in infinite loading (any of the mob smash modes)

TinyCK:
Yes, same as rubenguip. Mob smash is not work with an infinite loading. Yuzu EA 3708 and 2023 June 25.

PR #11451, opened 2023-09-05
service: mii: Fix default mii database

german77:
For some reason we had 8 default mii. Latest firmware only has 6. This fixes mob smash. Since this game was sensitive to these values I decided to use the correct mii name as well. This was verified with homebrew.

Thanks to byte for debugging. 

Fixes: #6253


liamwhite (approved):

FearlessTobi (approved):
Android CI failure can be disregarded.

PR #11475, opened 2023-09-10
debug: Add renderdoc capture hotkey

GPUCode:
Debugging the texture or buffer cache is often difficult because certain resources are uploaded at specific points and then used in many subsequent frames. Capturing these specific frames in renderdoc is possible but cumbersome. This is my attempt at a solution to this problem.

The PR uses the renderdoc C API and adds a hotkey to start/stop frame captures. This way multiple frames can be recorded and written into the same large capture. Then using the resource inspector the desired frame can be found and analyzed. I've only enabled this if graphics debugging is used as it's a strictly debugging feature.

### DIFF
--- a/src/core/tools/renderdoc.cpp
+++ b/src/core/tools/renderdoc.cpp
@@ -1,0 +1,55 @@
+// SPDX-FileCopyrightText: Copyright 2023 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <renderdoc_app.h>
+
+#include "common/assert.h"
+#include "common/dynamic_library.h"
+#include "core/tools/renderdoc.h"
+
+#ifdef WIN32
+#include <windows.h>
+#else
+#include <dlfcn.h>
+#endif
+
+namespace Tools {
+
+RenderdocAPI::RenderdocAPI() {
+#ifdef WIN32
+    if (HMODULE mod = GetModuleHandleA("renderdoc.dll")) {
+        const auto RENDERDOC_GetAPI =
+            reinterpret_cast<pRENDERDOC_GetAPI>(GetProcAddress(mod, "RENDERDOC_GetAPI"));
+        const s32 ret = RENDERDOC_GetAPI(eRENDERDOC_API_Version_1_6_0, (void**)&rdoc_api);
+        ASSERT(ret == 1);
+    }
+#else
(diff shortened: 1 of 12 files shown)


liamwhite:
It would be nice to not have this connected to the validation layers as they make debugging with renderdoc painfully slow.

GPUCode:
Should I add an addition setting for this, or make it always available?

liamwhite:
Whatever you think is best, I have no strong preference

GPUCode:
Added new setting
εικόνα

src/core/tools/renderdoc.cpp
```diff
@@ -0,0 +1,52 @@
...
+#else
+#include <dlfcn.h>
+#endif
+
+namespace Tools {
+
+RenderdocAPI::RenderdocAPI() {
+#ifdef WIN32
+    if (HMODULE mod = GetModuleHandleA("renderdoc.dll")) {
+        const auto RENDERDOC_GetAPI = (pRENDERDOC_GetAPI)GetProcAddress(mod, "RENDERDOC_GetAPI");
```
FearlessTobi, line 20: should this be a static cast?

vonchenplus:
The performance of continuous capture is very poor and other options may need to be investigated.

vonchenplus:
Often, successive captures are of little use, consider increasing the capture interval.

GPUCode:
Having a capture interval risks not capturing the frame that contains the resource upload, which makes the feature not useful. This is a hard tradeoff as having a small interval, there won't be much performance improvement, while having a larger one will result in many captures and reduce the chances one of them has the wanted frame.

src/core/tools/renderdoc.cpp
liamwhite: reinterpret, probably

src/core/tools/renderdoc.cpp
```diff
@@ -0,0 +1,52 @@
...
+        const s32 ret = RENDERDOC_GetAPI(eRENDERDOC_API_Version_1_6_0, (void**)&rdoc_api);
+        ASSERT(ret == 1);
```
liamwhite, line 22: needs common/assert.h

liamwhite:
@GPUCode needs rebase, also see comment above

liamwhite (approved):

PR #11523, opened 2023-09-17
android: Use 1 worker for shader compilation for all devices

t895:
Users have pointed out that during shader compilation in game (on devices that aren't using Adreno), games can crash unless they go through that game segment at a slower speed. This indicated a problem with using different amounts of workers between each device type.

Fix provided by @liamwhite


bylaws:
Would be better to not disable this for turnip

t895:
I'm actually about to change this to one worker thread for only all android devices (for a moment I thought this was android specific code). Generally trying to avoid using more memory on a constrained platform.

Having more threads will be fine for desktop devices using turnip.

liamwhite (approved):

t895:
Also just for the record. After some testing, it seems like this doesn't eliminate all shader compilation crashes but it does slightly increase performance and reduce the number of crashes.

goldenx86:
The main benefit here besides improving stability may be the saved RAM.

PR #11532, opened 2023-09-18
Tessil/robin-map implemented instead unordered_map and unordered_set

boludoz:
Robin map and robin set implemented, like Skyline, for more performance.


AliciaBurrito:
@boludoz Do you have any examples of the improved performance? eg. games to test, or what numbers you got?

FearlessTobi:
Blindly replacing every unordered_map with robin_map, without providing any performance numbers will not get accepted here.

boludoz:
> Blindly replacing every unordered_map with robin_map, without providing any performance numbers will not get accepted here.

image

boludoz:
```
#include <iostream>
#include <unordered_map>
#include <unordered_set>
#include "tsl/robin_map.h"
#include "tsl/robin_set.h"
#include <chrono>

int main() {
    std::unordered_map<int, int> unorderedMap;
    tsl::robin_map<int, int> robinMap;
    std::unordered_set<int> unorderedSet;
    tsl::robin_set<int> robinSet;

    auto start = std::chrono::high_resolution_clock::now();
    for (int i = 0; i < 1000000; ++i) {
        unorderedMap[i] = i * 2;
    }

    auto end = std::chrono::high_resolution_clock::now();
    std::chrono::duration<double> elapsed_seconds = end - start;

    std::cout << "Time to insert into std::unordered_map: " << elapsed_seconds.count() << " seconds" << std::endl;
    
    start = std::chrono::high_resolution_clock::now();
    for (int i = 0; i < 1000000; ++i) {
        robinMap[i] = i * 2;
    }

    end = std::chrono::high_resolution_clock::now();
    elapsed_seconds = end - start;

    std::cout << "Time to insert into tsl::robin_map: " << elapsed_seconds.count() << " seconds" << std::endl;

    start = std::chrono::high_resolution_clock::now();
    for (int i = 0; i < 1000000; ++i) {
        unorderedSet.insert(i);
    }

    end = std::chrono::high_resolution_clock::now();
    elapsed_seconds = end - start;

    std::cout << "Time to insert into std::unordered_set: " << elapsed_seconds.count() << " seconds" << std::endl;

    start = std::chrono::high_resolution_clock::now();
    for (int i = 0; i < 1000000; ++i) {
        robinSet.insert(i);
    }

    end = std::chrono::high_resolution_clock::now();
    elapsed_seconds = end - start;
    std::cout << "Time to insert into tsl::robin_set: " << elapsed_seconds.count() << " seconds" << std::endl;
    return 0;
}

```

boludoz:
> Blindly replacing every unordered_map with robin_map, without providing any performance numbers will not get accepted here.

I ask you to please reconsider your opinion, perhaps the performance is not too much but it can help. It is a grain of sand that makes the mountain. 30 % more performance in tsl::robin_set for example.

boludoz:
> @boludoz Do you have any examples of the improved performance? eg. games to test, or what numbers you got?

image
I don't give explanations to xenophobic people.

FernandoS27:
I'm sorry our moderators are morons who cannot respect someone's hard work and put you in the hard position of going through benchmarks and stuffs without much necessity. 

Please feel free to reopen.

FernandoS27:
Oh fix clamg format.

gidoly:
i can check if this pr makes difference in games after ci passed. even if doesn't make big difference it is nice to have

boludoz:
> i can check if this pr makes difference in games after ci passed. even if doesn't make big difference it is nice to have

I think the order is fixed there.

gidoly:
getting compiling error on visual studio
devenv_WUDBhnoVfD
i think it is because missing external file as cmake gives me this error
WindowsTerminal_Oe3nlNFJNb

externals/CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ set(BUILD_SHARED_LIBS OFF)
 # Skip install rules for all externals
 set_directory_properties(PROPERTIES EXCLUDE_FROM_ALL ON)
 
+# Robin-map
+add_subdirectory(robin-map)
```
Tachi107, line 18: Could you please add `find_package(tsl-robin-map CONFIG)` to the top-level CMakeLists.txt and change this to the following:

```suggestion
if (NOT TARGET tsl::robin_map)
    add_subdirectory(robin-map)
endif()
```

Also, you don't seem to use the `tsl::robin_map` target anywhere in the CMake files, you should probably add `target_link_libraries(<target-name> PRIVATE tsl::robin_map)` in the relevant place(s).

.gitmodules
```diff
@@ -58,3 +58,6 @@
 [submodule "VulkanMemoryAllocator"]
 	path = externals/VulkanMemoryAllocator
 	url = https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator.git
+[submodule "robin-map"]
+	path = externals/robin-map
+	url = https://github.com/Tessil/robin-map
```
v1993, line 63:
```suggestion
	url = https://github.com/Tessil/robin-map.git
```

boludoz: Thanks sir.

v1993: Looks like this was not committed?

v1993 (changes requested):
`.gitmodules` still not fixed

boludoz:
> `.gitmodules` still not fixed

Sorry, pushed now.

v1993 (commented):
Benchmarking aside - it's worth noting that robin-map has [some differences](https://github.com/Tessil/robin-map#differences-compared-to-stdunordered_map) in its API compared to vanilla `unordered_map`. While most are covered by compile-time checks, a few differences in iterator invalidation should be verified manually to not affect code in question.

v1993 (commented):
Benchmarking aside - it's worth noting that robin-map has [some differences](https://github.com/Tessil/robin-map#differences-compared-to-stdunordered_map) in its API compared to vanilla `unordered_map`. While most are covered by compile-time checks, a few differences in iterator invalidation should be verified manually to not affect code in question.

boludoz:
Fixed submodule

boludoz:
If anyone wants to help, I'm having a hard time adding a simple library.

src/yuzu/multiplayer/chat_room.h
```diff
@@ -64,8 +64,8 @@ public slots:
     bool has_mod_perms = false;
     QStandardItemModel* player_list;
     std::unique_ptr<Ui::ChatRoom> ui;
-    std::unordered_set<std::string> block_list;
-    std::unordered_map<std::string, QPixmap> icon_cache;
+    tsl::robin_set<std::string> block_list;
```
FearlessTobi, line 67: this is not performance critical at all.
dito for many other places

Kelebek1:
I don't know if there's a way to make an external available to all projects by default, but I think you need to add robin map to each project's `target_link_libraries` in their cmake file. 
E.g src/common/cmakelists.txt has
```cmake
target_link_libraries(common PUBLIC Boost::context Boost::headers fmt::fmt microprofile Threads::Threads)
target_link_libraries(common PRIVATE lz4::lz4 zstd::zstd LLVM::Demangle)
```
You'll want robin-map in there, in the public one I guess? And repeat for each project that you're having use it.

FearlessTobi (changes requested):
You need to provide actual performance comparsions from games for this to have a chance to get approved.

boludoz:
For my part, there is nothing more to contribute, if anyone wants to solve the problem, welcome...

PR #11546, opened 2023-09-19
Reduce core timing mutex contention

Kelebek1:
This gives a small reduction in the mutex contention around core timing callbacks.


Azmagisk:
Vamos no se que sea pero te apoyo 🐕‍🦺

PR #11534, opened 2023-09-18
Partial revert of #10433 (Texture Cache Util: Fix block depth adjustment on slices)

Squall-Leonhart:
Alternative to #11528 

The If block in this change was causing some 2D textures to be treated as if their mip 0 was a 3D Slice, this could be ascertained as the same texture viewed from different distances would render fine, but then close up would look like a decoding failure.

It also resulted in some 3D ASTC textures not being scaled appropriate leading to broken graphical effects such as the jagged TOTK recall animation being a circle, as the If block was only accepting the image based on its original info without any adjustments applied.

Fixes #11252  and #10493

Doesn't reintroduce #10255

PR #11554, opened 2023-09-21
Revoke https://github.com/yuzu-emu/yuzu/pull/10433 as a temp solution to fix OLD texture error

Aphcity:
Revoke https://github.com/yuzu-emu/yuzu/pull/10433 as a temp solution to fix OLD texture error in multiple games, especially for Android player who start to play ns games after yuzu for Android released since old versions before mainline 1441 don't have the same issue.

This is ONLY a TEMP solution to fix those issues and I will close this pull request AT ONCE there is a solution found to solve the problem.

Just for those eager players.

Fixed https://github.com/yuzu-emu/yuzu/issues/10456 (tested) and maybe so on. Find artifacts in Actions or my fork.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -4,6 +4,9 @@
 name: 'yuzu verify'
 
 on:
+  push:
+    branches: [master]
+    tags: ["*"]
   pull_request:
     branches: [ master ]
 env:
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -4,9 +4,6 @@
 name: 'yuzu verify'
 
 on:
-  push:
-    branches: [master]
-    tags: ["*"]
   pull_request:
     branches: [ master ]
 env:
--- a/src/video_core/texture_cache/util.cpp
+++ b/src/video_core/texture_cache/util.cpp
@@ -122,9 +122,7 @@ template <u32 GOB_EXTENT>
     return {
         .width = AdjustMipBlockSize<GOB_SIZE_X>(num_tiles.width, block_size.width, level),
         .height = AdjustMipBlockSize<GOB_SIZE_Y>(num_tiles.height, block_size.height, level),
-        .depth = level == 0
-                     ? block_size.depth
-                     : AdjustMipBlockSize<GOB_SIZE_Z>(num_tiles.depth, block_size.depth, level),
+        .depth = AdjustMipBlockSize<GOB_SIZE_Z>(num_tiles.depth, block_size.depth, level),
     };
 }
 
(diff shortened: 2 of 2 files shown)


Kelebek1:
Isn't #11534 fixing this?

Squall-Leonhart:
> Isn't #11534 fixing this?

Was supposed to, i had assumed these textures might be pushed back into the AdjustMipBlockSize code but turns out they are single mip too so they end up going being handled directly any way.

The developer of the translation has already addressed this on their end, but i was hoping to sort it out on Yuzu's too, it seems that the File's blockdepth exceeds what Yuzu expects, so when not running through the adjusted code it works fine but when going through block_size.depth directly the block depth is misinterpretted as 3 or 4 instead of the 0 it should be, I have tinkered with the log2 size to some degree's of better or worse appearance but none of it made it look like it does on hardware.

This particular PR cannot be taken as it returns the broken gloom to TOTK.

Kelebek1:
> I don't really understand the architecture at all, but going by discussions with Liam, i'm assuming that somewhere in the TIC, something involving the block info has less bits available than it technically should so block depth value from the tic is not being masked correctly for these particular textures.
> 
> It's also a possibility that the emulator is right and the hardware has an unexpected quirk that allows this BNTX format to work without issues.

You can compare our struct at https://github.com/yuzu-emu/yuzu/blob/master/src/video_core/textures/texture.h#L190-L192 with the open docs here https://github.com/NVIDIA/open-gpu-doc/blob/master/classes/3d/clb197tex.h#L424-L439. The block width/height/depth should be right.

We only have a single struct though, but really from that file there's 5 different types that are used: Blocklinear, blocklinear color key, 1d, pitch and pitch colour key. Most fields are in the same place, but some switch around, like width, and is why we have both `buffer_high/low_width_minus_one` and `width_minus_one` for instance. Some of the fields may be wrong though, and I do notice looking at it, that there's a gob3d bit right below tile width, which we don't have, and that seems relevant to this problem.

I tried overhauling this struct before, but I couldn't figure out a clean way to split it into 5 structs and have the members be easily accessible without a million getters. If you want to try and improve this struct and make it more accurate to those docs, go ahead. The gob3d bit is likely worth adding at least.

Squall-Leonhart:
> The gob3d bit is likely worth adding at least.

fourth union would look like this in that  case wit the gob3D filling the gap after tile_width_spacing

``` 
           union {
                BitField<0, 3, u32> block_width;
                BitField<3, 3, u32> block_height;
                BitField<6, 3, u32> block_depth;

                BitField<10, 3, u32> tile_width_spacing;
                BitField<13, 1, u32> gob_3d;

                // High 16 bits of the pitch value
                BitField<0, 16, u32> pitch_high;
                BitField<26, 1, u32> use_header_opt_control;
                BitField<27, 1, u32> depth_texture;
                BitField<28, 4, u32> max_mip_level;

                BitField<0, 16, u32> buffer_high_width_minus_one;
            };```

Issue #11550, opened 2023-09-20
Android: The Legend of Nayuta: Boundless Trails (Font/text bug) cant read text ingame


brujo5:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

Android-v77

### Description of Issue

Can't read the text because it's dark

![Screenshot_20230920-111034_yuzu](https://github.com/yuzu-emu/yuzu/assets/4172898/afbf3369-afb3-49c3-bdf3-f812563e3164)
![Screenshot_20230920-111108_yuzu](https://github.com/yuzu-emu/yuzu/assets/4172898/1deb6759-5734-46cd-96fb-82522838c0ff)
![Screenshot_20230920-111143_yuzu](https://github.com/yuzu-emu/yuzu/assets/4172898/3762f199-d23a-4ae4-87ea-ac95e67e2154)

### Expected Behavior

that shouldn't happen

### Reproduction Steps

Justo play normally

### Log File

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/12676553/yuzu_log.txt)

### System Configuration

CPU:sd870
GPU/Driver: Adreno 650, turnip driver 
RAM: 8gb
OS: android 13

PR #11557, opened 2023-09-21
renderer_vulkan: Correct component order for A4B4G4R4_UNORM

GPUCode:
Some games use this format for text rendering but because the component order was not correct it would show up all wrong. So instead we use VK_FORMAT_A4B4G4R4_UNORM_PACK16_EXT which has the order we want. If that format is not available then it is emulated using a swizzle.

yuzu (master) | yuzu (PR)
-|-
master | pr

Fixes https://github.com/yuzu-emu/yuzu/issues/11550


liamwhite (approved):

gidoly:
yuzu-msvc-verify-pr11557-20230921-b4c30bdb7_owLjHXwNZq

fixes text rendering on hajimari no kiseki eastern version. likely will fix other falcom game that has same problem

Issue #3587, opened 2020-03-30
Super Smash Bros. Ultimate Mii Crashes


theincredibleplayer:
Creating a Mii or joining century smash or going into a stage with a mii in it (Find Mii, Tomadachi life) causes the game to softlock.

Creating Mii log:
[yuzu_log.zip](https://github.com/yuzu-emu/yuzu/files/4405668/yuzu_log.zip)
![Screenshot (1105)](https://user-images.githubusercontent.com/54246318/77965723-ed242b00-72e1-11ea-9755-568a103c94ad.png)

Century smash log:
[yuzu_log.zip](https://github.com/yuzu-emu/yuzu/files/4405683/yuzu_log.zip)
crashes after selecting character and starting,doesnt progress to gameplay.
![Screenshot (1106)](https://user-images.githubusercontent.com/54246318/77965714-e7c6e080-72e1-11ea-81db-248ae689250f.png)

tomadachi life stage log:
[yuzu_log.zip](https://github.com/yuzu-emu/yuzu/files/4405732/yuzu_log.zip)

Tested on yuzu Early access 317

hyoretsu:
So can we assume we aren't gonna need a Switch or Ryujinx to create a save on games like Mario Kart 8 Deluxe in the near future?

PR #11561, opened 2023-09-21
am: mii_edit: Implement DB operations

german77:
Allows our HLE implementation of mii edit to create mii. This PR also fixes random mii generation, makes the mii manager shared across sessions as well implements GetIndex which somehow wasn't implemented by previous PR.

With this change creating a mii from any game won't result in any crash. Instead it will automatically generate a random mii.

Fixes: #3587

01006a800016e000_2023-09-21_10-52-35-326

### DIFF
--- a/src/core/hle/service/mii/mii.cpp
+++ b/src/core/hle/service/mii/mii.cpp
@@ -18,8 +18,10 @@ namespace Service::Mii {
 
 class IDatabaseService final : public ServiceFramework<IDatabaseService> {
 public:
-    explicit IDatabaseService(Core::System& system_, bool is_system_)
-        : ServiceFramework{system_, "IDatabaseService"}, is_system{is_system_} {
+    explicit IDatabaseService(Core::System& system_, std::shared_ptr<MiiManager> mii_manager,
+                              bool is_system_)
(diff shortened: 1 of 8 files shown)


liamwhite (approved):

src/core/hle/service/mii/mii.cpp
```diff
@@ -516,41 +518,46 @@ class IDatabaseService final : public ServiceFramework<IDatabaseService> {
...
-        LOG_DEBUG(Service_Mii, "called");
-    }
+void MiiDBModule::GetDatabaseService(HLERequestContext& ctx) {
+    IPC::ResponseBuilder rb{ctx, 2, 0, 1};
+    rb.Push(ResultSuccess);
+    rb.PushIpcInterface<IDatabaseService>(system, manager, is_system);
 
-    bool is_system{};
-};
+    LOG_CRITICAL(Service_Mii, "called");
```
FearlessTobi, line 555: huh

FearlessTobi (approved):

src/core/hle/service/mii/mii.cpp
german77: Why does this keep happening XD

Issue #11010, opened 2023-07-03
Momotaro Dentetsu - Dark Color Rendering on Vulkan


JWatersMeet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

Early Access 3734

### Description of Issue

Momotaro Dentetsu is now working following #10803. However, the color rendering on Vulkan is much darker than an OpenGL. Comparing to it running on my Switch, the OpenGL version is much closer to the actual switch rendering. The issue occurs throughout the game, right from the initial loading screens.

The issue occurs on Vulkan regardless of whether I used my nVidia or AMD GPU.

![Momotaro - Dark Colors (Vulkan)](https://github.com/yuzu-emu/yuzu/assets/136824455/19ccde09-4bf6-42bf-a955-6f708267d215)

### Expected Behavior

Vulkan color rendering should match OpenGL.

![Momotaro - Correct Colors (GLSL)](https://github.com/yuzu-emu/yuzu/assets/136824455/3085f749-ca45-4f23-813b-0cc87655e719)

### Reproduction Steps

Start Momotaro Dentetsu in Vulkan - Notice Colors are Dark
Start Momotaro Dentetsu in OpenGL - Notice Colors Match Switch

### Log File

To follow if needed.

### System Configuration

CPU: Ryzen 5800HX
GPU/Driver: RTX 3080 Mobile, 31.0.15.33318
RAM: 16GB
OS: Windows 11

JWatersMeet:
Can confirm resolved by #11562 
Many thanks!

Issue #7546, opened 2021-12-08
Star Ocean First Departure R rendering issue


NineKain:
![image](https://user-images.githubusercontent.com/72826232/145279727-e5004cbe-481d-417c-87ae-ad723b61363a.png)

Main character is pink

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/7678754/yuzu_log.txt)

NineKain:
![image](https://user-images.githubusercontent.com/72826232/145280024-b4c31a10-cf52-414f-91d2-b8dfa6ff4fe7.png)
opengl solves the issue

benderscruffy:
using the latest EA and mainline
this is still an issue

Hector-Ariceaga:
Still noticing this issue. Not aware of a fix yet with latest drivers and early access build.

gidoly:
Fixed by 11562

PR #11562, opened 2023-09-21
vk_texture_cache: Limit srgb block to transcoding only

GPUCode:
This was originally introduced in https://github.com/yuzu-emu/yuzu/pull/5363. The problem was that UNORM views of SRGB formats could not be attached as storage images in shaders, with the solution being creating the image as UNORM ~~and having the view be SRGB.~~

However this doesn't always work as intended. When the image is used directly without the aid of the view, for example when image is target of vkCmdResolveImage, the colors end up wrong since the colorspace has changed. To fix this, we can do a partial revert of that part and limit the srgb block only when transcoding compressed formats in the GPU.

Fixes https://github.com/yuzu-emu/yuzu/issues/11010

### DIFF
--- a/src/video_core/renderer_vulkan/vk_texture_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_texture_cache.cpp
@@ -120,19 +120,9 @@ constexpr VkBorderColor ConvertBorderColor(const std::array<float, 4>& color) {
     return usage;
 }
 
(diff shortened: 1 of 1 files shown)


liamwhite (approved):

stuken:
Also fixes #7546 and an unreported discolouration issue with Picross S Genesis edition's pixel art backgrounds.

PR #11583, opened 2023-09-25
android: Use measured size of view for input overlay bounds

t895:
Even after updating the androidx window library, this did not fix the issue for all devices. This ensures that the measured size of the overlay will be used instead of a potentially larger one seen by androidx.


liamwhite (approved):

ReyZen45:
Hope next build will fix it

ReyZen45:
Thx for fixing. Allready Test on Build 82, and Overlay Fixed

PR #11604, opened 2023-09-26
Frontend: Remove ability to install xci files

t895:
Just talked with @liamwhite about how you can still install these. Figured we should remove the ability to since XCIs are usually application containers and will fail on installation.


FearlessTobi:
please fix the conflicts

FearlessTobi (approved):

PR #11613, opened 2023-09-27
android: Various play store fixes

t895:
Was looking at play store stats and noticed a few crashes that I could fix quickly.

Fixes an issue where the app would crash if the emulation activity doesn't have a game.
Fixes an issue where we tried to update invalid views in the emulation fragment.
Fixes a couple issues where we tried to update invalid views in the setup fragment.

Additionally don't wait for post callback to update input overlay visibility.

### DIFF
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
@@ -304,6 +304,10 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
 
     override fun onConfigurationChanged(newConfig: Configuration) {
         super.onConfigurationChanged(newConfig)
+        if (_binding == null) {
+            return
+        }
+
         updateScreenLayout()
         if (emulationActivity?.isInPictureInPictureMode == true) {
             if (binding.drawerLayout.isOpen) {
--- a/src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
+++ b/src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
@@ -310,19 +310,13 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
                 binding.drawerLayout.close()
             }
             if (EmulationMenuSettings.showOverlay) {
-                binding.surfaceInputOverlay.post {
-                    binding.surfaceInputOverlay.visibility = View.INVISIBLE
-                }
+                binding.surfaceInputOverlay.visibility = View.INVISIBLE
             }
         } else {
             if (EmulationMenuSettings.showOverlay && emulationViewModel.emulationStarted.value) {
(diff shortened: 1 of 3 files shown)


src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/SetupFragment.kt
```diff
@@ -353,11 +355,15 @@ class SetupFragment : Fragment() {
     }
 
     fun pageForward() {
-        binding.viewPager2.currentItem = binding.viewPager2.currentItem + 1
+        if (_binding != null) {
+            binding.viewPager2.currentItem += 1
```
FearlessTobi, line 359: my kotlin is a bit rusty, but couldn't we do something like
`binding.?viewPager2.currentItem += 1`
here?

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
```diff
@@ -104,10 +105,21 @@ class EmulationFragment : Fragment(), SurfaceHolder.Callback {
...
-        } else {
-            intentGame ?: error("[EmulationFragment] No bootable game present!")
+
+        try {
+            game = if (args.game != null) {
+                args.game!!
+            } else {
+                intentGame!!
+            }
+        } catch (e: Exception) {
```
FearlessTobi, line 115: can we make it more specific what kind of exception we are catching here?
And if is a null pointer exception, just handle it more gracefully?

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/SetupFragment.kt
t895: I guess in this case I could just directly use `_binding`. Probably would make things look cleaner.

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
t895: Yeah I could just catch the NullPointerException but that's about as gracefully as I can handle that situation. This would specifically be catching when someone tries to open a game shortcut where we no longer have access to the game.

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/SetupFragment.kt
t895: Huh, actually I think this is worse. The syntax becomes this and it just looks confusing.
`_binding?.viewPager2?.currentItem?.plus(1)`

src/android/app/src/main/java/org/yuzu/yuzu_emu/fragments/EmulationFragment.kt
t895: *sorry this also catches if you somehow open the emulation activity from within the app with a null reference to the game

FearlessTobi (approved):

Issue #11600, opened 2023-09-26
[Bug] Downscale glitches


boludoz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

yuzu 1568

### Description of Issue

The emulator displays downscaled textures incorrectly in Red Dead Redemption 1.

![Captura de pantalla 2023-09-25 235052](https://github.com/yuzu-emu/yuzu/assets/11653030/5b28739e-7a6f-4459-a67d-9a4f705c640c)
![Captura de pantalla 2023-09-25 235225](https://github.com/yuzu-emu/yuzu/assets/11653030/c84e3e93-0e25-422b-a589-f75536a8d9ed)
![Captura de pantalla 2023-09-01 122250](https://github.com/yuzu-emu/yuzu/assets/11653030/7883a95c-4e8e-4ae6-8f2c-22f18e2696ed)
![Captura de pantalla 2023-09-23 025442](https://github.com/yuzu-emu/yuzu/assets/11653030/4c268411-d082-49dd-87e1-87c59ca0193f)

### Expected Behavior

It should work without those glitches.

### Reproduction Steps

Set the emulator below 1.0x on Intel graphics.

### Log File

[yuzu_log.txt.old.txt](https://github.com/yuzu-emu/yuzu/files/12721585/yuzu_log.txt.old.txt)

### System Configuration

Intel i3 11gen integrated graphics and 8 GB RAM 3200mhz

boludoz:
They seem like minimal glitches but they become annoying and it's a shame because the emulator works well beyond that.

boludoz:
There is a variable called "downscaleable", the problem seems to be there, because when I remove the logic related to it the glitch disappears, but it would break another game.

PR #11625, opened 2023-09-28
Fixed downscale glitches on Intel and Red Dead Redemption at downscale

boludoz:
Fixed downscale glitches on Intel and Red Dead Redemption at downscale #11600 improvements are welcome

### DIFF
--- a/src/video_core/texture_cache/image_info.cpp
+++ b/src/video_core/texture_cache/image_info.cpp
@@ -61,6 +61,7 @@ ImageInfo::ImageInfo(const TICEntry& config) noexcept {
         resources.layers = config.Depth();
         break;
     case TextureType::Texture2D:
+        [[fallthrough]];
     case TextureType::Texture2DNoMipmap:
         ASSERT(config.Depth() == 1);
         type = config.IsPitchLinear() ? ImageType::Linear : ImageType::e2D;
@@ -155,8 +156,7 @@ ImageInfo::ImageInfo(const Maxwell3D::Regs::RenderTargetConfig& ct,
         type = ImageType::e3D;
         size.depth = ct.depth;
     } else {
-        rescaleable = block.depth == 0;
-        rescaleable &= size.height > RescaleHeightThreshold;
+        rescaleable = block.depth == 0 && size.height > RescaleHeightThreshold;
         downscaleable = size.height > DownscaleHeightThreshold;
         type = ImageType::e2D;
         resources.layers = ct.depth;
@@ -193,7 +193,8 @@ ImageInfo::ImageInfo(const Maxwell3D::Regs::Zeta& zt, const Maxwell3D::Regs::Zet
         size.depth = zt_size.depth;
     } else {
         rescaleable = block.depth == 0;
-        downscaleable = size.height > 512;
+        // TODO: Check if this is correct
+        // downscaleable = size.height > DownscaleHeightThreshold;
         type = ImageType::e2D;
         switch (zt_size.dim_control) {
         case Maxwell3D::Regs::ZetaSize::DimensionControl::DefineArraySize:


gidoly:
hmm i will let you know if this affect other games

boludoz:
> hmm i will let you know if this affect other games

That's why it's a draft, I would like it to be a solution only for Intel.

boludoz:
image
Tested in RTX 3080, the biggest problem is that now the gui does not rescale, in any case, the resolution will continue to be as expected.

src/video_core/texture_cache/image_info.cpp
```diff
@@ -155,8 +156,7 @@ ImageInfo::ImageInfo(const Maxwell3D::Regs::RenderTargetConfig& ct,
         type = ImageType::e3D;
         size.depth = ct.depth;
     } else {
-        rescaleable = block.depth == 0;
-        rescaleable &= size.height > RescaleHeightThreshold;
```
v1993, line 159: Any particular reason for this change? Both operands are bool, so there should be no difference and IMO the original is a bit more readable.

src/video_core/texture_cache/image_info.cpp
```diff
@@ -193,7 +193,8 @@ ImageInfo::ImageInfo(const Maxwell3D::Regs::Zeta& zt, const Maxwell3D::Regs::Zet
         size.depth = zt_size.depth;
     } else {
         rescaleable = block.depth == 0;
-        downscaleable = size.height > 512;
+        // TODO: Check if this is correct
+        // downscaleable = size.height > DownscaleHeightThreshold;
```
v1993, line 197: I get that this is a WIP, but I suggest explicitly assigning `false` here for the time being to make it clear what happens in this case.

src/video_core/texture_cache/image_info.cpp
boludoz: No, I just wanted to test if "block.depth == 0" has anything to do with it.

src/video_core/texture_cache/image_info.cpp
boludoz: I would really like to directly avoid those checks and assign true.

FernandoS27: I think that was there for a reason. I'm sure a game gets affected by this.

src/video_core/texture_cache/image_info.cpp
```diff
@@ -61,6 +61,7 @@ ImageInfo::ImageInfo(const TICEntry& config) noexcept {
         resources.layers = config.Depth();
         break;
     case TextureType::Texture2D:
+        [[fallthrough]];
```
FernandoS27, line 64: You don't have yo add fallthrough if there's no block of code between the cases.

boludoz: Sorry, I thought it was good practice for the compiler and reading the code 😁

boludoz:
There is also something that can be irritating and that is changing a 2-condition "switch" for an "if", since in simple or short conditions it is much more efficient to use "if". In any case, this is a draft and they are only suggestions, perhaps the problem with rescaling is a little further away and unfortunately my knowledge is limited.

v1993:
I'd suggest to stop worrying about micro-optimizations for the time being and limit this PR's scope to fixing bugs. If this ever comes up in profiling then it can be addressed separately later on - this code is hardly a bottleneck as of right now.

src/video_core/texture_cache/image_info.cpp
FernandoS27: The compiler is very smart and knows how to optimize code correctly. There's no need to be very verbose unless necessary and when profiling indicates a need.

FernandoS27:
I must agree with v1993. Do not add microoptimizations unless profiling indicates it's necessary. Sacrificing code readability for winning 1 or 2 cpu cycles is always a bad practice.

boludoz:
> I must agree with v1993. Do not add microoptimizations unless profiling indicates it's necessary. Sacrificing code readability for winning 1 or 2 cpu cycles is always a bad practice.



> I'd suggest to stop worrying about micro-optimizations for the time being and limit this PR's scope to fixing bugs. If this ever comes up in profiling then it can be addressed separately later on - this code is hardly a bottleneck as of right now.

Sorry

Kelebek1:
Would be nice to find the real cause of this instead of just avoiding it.

From what you changed it looks like the issue is with arrays? So probably somewhere we're not calculating the downscaled layer size correctly.

boludoz:
> Would be nice to find the real cause of this instead of just avoiding it.
> 
> From what you changed it looks like the issue is with arrays? So probably somewhere we're not calculating the downscaled layer size correctly.

It is precisely what I intend.

src/video_core/texture_cache/image_info.cpp
```diff
@@ -23,7 +23,8 @@ using VideoCore::Surface::PixelFormat;
 using VideoCore::Surface::SurfaceType;
 
 constexpr u32 RescaleHeightThreshold = 288;
-constexpr u32 DownscaleHeightThreshold = 512;
+constexpr u32 DownscaleHeightThreshold = 128;
+constexpr u32 DownscaleHeightThresholdTwo = 1024;
```
v1993, line 27:
```suggestion
constexpr u32 DownscaleHeightThresholdNormal = 128;
constexpr u32 DownscaleHeightThresholdArray = 1024;
```

Or something like this. Current names fail to describe the difference between the two.

FernandoS27:
> > I must agree with v1993. Do not add microoptimizations unless profiling indicates it's necessary. Sacrificing code readability for winning 1 or 2 cpu cycles is always a bad practice.
> 
> > I'd suggest to stop worrying about micro-optimizations for the time being and limit this PR's scope to fixing bugs. If this ever comes up in profiling then it can be addressed separately later on - this code is hardly a bottleneck as of right now.
> 
> Sorry

You don't have to apologize when one starts, one tends to do that. I also had a tendency of doing a lotbof microoptimizations

Anyways, do you know how to use RenderDocs ? If not we can teach you if you pass by our discord and ask for speech in #development

boludoz:
> > > I must agree with v1993. Do not add microoptimizations unless profiling indicates it's necessary. Sacrificing code readability for winning 1 or 2 cpu cycles is always a bad practice.
> > 
> > 
> > > I'd suggest to stop worrying about micro-optimizations for the time being and limit this PR's scope to fixing bugs. If this ever comes up in profiling then it can be addressed separately later on - this code is hardly a bottleneck as of right now.
> > 
> > 
> > Sorry
> 
> You don't have to apologize when one starts, one tends to do that. I also had a tendency of doing a lotbof microoptimizations
> 
> Anyways, do you know how to use RenderDocs ? If not we can teach you if you pass by our discord and ask for speech in #development

Thank you very much sir, I will join tomorrow. Anything to improve and gain new knowledge is welcome. This emulator is very well optimized, especially in the graphics and shaders section, but as you say, one tends to want to optimize everything even more.

src/video_core/texture_cache/image_info.cpp
boludoz: You're right, I took your suggestion, my bad.

src/video_core/texture_cache/image_info.cpp
FernandoS27: can you remove this

FernandoS27:
I'll ask our testers to have a look.

boludoz:
For some reason the game became unplayable.

PR #11684, opened 2023-10-05
Disable push descriptor for Pascal and older nVidia architectures

Kelebek1:
I encountered a bug in Bravely Default 2 around its water rendering, where it flickers in and out every few frames. After some debugging I found it to be *indirectly* caused by push descriptor. If the number of bindings for a draw exceeds the device's push descriptor maximum, we fallback to using regular descriptor binding, otherwise push descriptor is used. There seems to be either a hardware bug or a long-standing driver bug (no driver I've tested seems to fix it, and newer architectures are fine), where switching between descriptor binding and push descriptor will break the descriptors, and causes the flickering.

Confirmed via Bravely Default 2 to affect Pascal and Maxwell, so this PR disables push descriptor for architectures <= Pascal. This PR also re-works the architecture detection a little, adding more architecture options, and reordering the archs to allow < > to be used.

The bug:

https://github.com/yuzu-emu/yuzu/assets/34639600/bff1de28-66b8-4e34-9672-ed88e9773fc2

Here's a save (save slot 1) if anyone has the game with other architectures and wants to test it: https://cdn.discordapp.com/attachments/476077512984756254/1159286167200604202/SaveGameData1

### DIFF
--- a/src/video_core/vulkan_common/vulkan_device.cpp
+++ b/src/video_core/vulkan_common/vulkan_device.cpp
@@ -84,9 +84,12 @@ constexpr std::array VK_FORMAT_A4B4G4R4_UNORM_PACK16{
 } // namespace Alternatives
 
 enum class NvidiaArchitecture {
-    AmpereOrNewer,
+    KeplerOrOlder,
+    Maxwell,
+    Pascal,
+    Volta,
     Turing,
-    VoltaOrOlder,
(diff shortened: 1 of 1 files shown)


benderscruffy:
just tested with your save on my RTX 3060 (Ampere series) and no problem

liamwhite (approved):

Issue #11737, opened 2023-10-10
Desktop shortcut not work


DanielSvoboda:
### Issue

This PR https://github.com/yuzu-emu/yuzu/pull/11705 (Add desktop shortcut support for Windows) presents this error:

![image](https://github.com/yuzu-emu/yuzu/assets/28203966/7b213fc4-4604-47ba-b5d5-40015885215f)
This error happens if I modify my desktop location. Because my desktop is on another unit.
In my version (https://github.com/yuzu-emu/yuzu/compare/master...DanielSvoboda:yuzu:shortcut) I used this code below.

```c++
WCHAR path[MAX_PATH];
SHGetFolderPath(NULL, CSIDL_DESKTOPDIRECTORY, NULL, SHGFP_TYPE_CURRENT, path);
target_directory = path;
```

The code currently has this part:
```c++
target_directory = home_path / "Desktop";
```
  That won't work in Brazil or any other place where the Desktop is not called 'Desktop', here it is called 'Área de Trabalho'.

It is necessary to make some adjustments to the current code remembering the existence of the character 'Á'.

german77:
Seems like this function is windows only. Let me research for one that works on both windows and linux systems

DanielSvoboda:
Yes, this code I shared takes the correct desktop directory without needing ( / "Desktop" ) but it is only for Windows. This code below I believe will work in Linux, I have no way of testing it at the moment...

```c++
#include <cstdlib>
#include <iostream>
#include <array>

std::string getDesktopPath() {
    std::array<char, 128> buffer;
    std::string result;
    FILE* pipe = popen("xdg-user-dir DESKTOP", "r");
    if (!pipe) throw std::runtime_error("popen() failed!");
    while (fgets(buffer.data(), 128, pipe) != nullptr) {
        result += buffer.data();
    }
    auto returnCode = pclose(pipe);
    return result;
}
```

v1993:
@german77 Qt code can use [`QStandardPaths::writableLocation(QStandardPaths::DesktopLocation)`](https://doc.qt.io/qt-6/qstandardpaths.html) for this purpose.

DanielSvoboda:
Thanks for the suggestions, it worked perfectly.
I used QStandardPaths to locate the directories.
And QDir to check, because the previous way didn't work correctly because of the special characters in my language.

PR #11749, opened 2023-10-11
Improvement in Directory Path Detection for Shortcuts

DanielSvoboda:
image

This pull request updates how the directory path for shortcuts is determined. The main changes are:

1. Replaced the use of environment variables to determine the path of the desktop and applications menu with `QStandardPaths::writableLocation`. This change addresses an issue where the desktop path was not correctly identified when its location was customized, as shown in the attached screenshot.

2. Added conversion from `QString` to `std::string` using `toUtf8()`, which correctly handles non-ASCII characters in directory paths. This change ensures that directory paths containing Portuguese words like "área de trabalho" are supported.

3. Replaced directory checking using `Common::FS::IsDir()` with `QDir::exists()`.

These changes should improve cross-platform compatibility and code robustness. Because it couldn't locate my desktop, which wasn't on the C drive, but on the F, and even though localization wouldn't work because it was setting it to find the 'Desktop' folder and in the computer's language it says 'Área de trabalho', that will fix for other languages too.

Thanks for the idea suggested in 'issues' https://github.com/yuzu-emu/yuzu/issues/11737

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2884,7 +2884,7 @@ if (target == GameListShortcutTarget::Desktop) {
                 .arg(QString::fromStdString(target_directory.generic_string())),
             QMessageBox::StandardButton::Ok);
         return;
-     }
+    }
 } else if (target == GameListShortcutTarget::Applications) {
  #if defined(__linux__)
     QString applicationsPath =
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2875,7 +2875,7 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
     // Determine target directory for shortcut
     if (target == GameListShortcutTarget::Desktop) {
         const auto desktop_path = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
-        target_directory = desktopPath.toUtf8().toStdString();
+        target_directory = desktop_path.toUtf8().toStdString();
         const QDir dir(QString::fromStdString(target_directory.generic_string()));
         if (!dir.exists()) {
             QMessageBox::critical(
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2874,7 +2874,7 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
 
     // Determine target directory for shortcut
     if (target == GameListShortcutTarget::Desktop) {
-        QString desktopPath = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
+        const auto desktop_path = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
         target_directory = desktopPath.toUtf8().toStdString();
         QDir dir(QString::fromStdString(target_directory.generic_string()));
         if (!dir.exists()) {
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -2872,36 +2872,36 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
 
     std::filesystem::path target_directory{};
 
-// Determine target directory for shortcut
-if (target == GameListShortcutTarget::Desktop) {
-    QString desktopPath = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
-    target_directory = desktopPath.toUtf8().toStdString();
-    QDir dir(QString::fromStdString(target_directory.generic_string()));
-    if (!dir.exists()) {
-        QMessageBox::critical(
-            this, tr("Create Shortcut"),
-            tr("Cannot create shortcut on desktop. Path \"%1\" does not exist.")
-                .arg(QString::fromStdString(target_directory.generic_string())),
-            QMessageBox::StandardButton::Ok);
-        return;
+    // Determine target directory for shortcut
+    if (target == GameListShortcutTarget::Desktop) {
+        QString desktopPath = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
+        target_directory = desktopPath.toUtf8().toStdString();
+        QDir dir(QString::fromStdString(target_directory.generic_string()));
+        if (!dir.exists()) {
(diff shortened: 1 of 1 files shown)


german77:
Do we still need the macro? Seems like both linux and windows is identical besides the target.

DanielSvoboda:
What do you mean macro?

in windows there is the button
Add to Desktop

and in Linux there are 2
Add to Desktop
Add to Applications Menu

How can you see the link
https://github.com/FearlessTobi/yuzu/blob/9ef9ca0927be1a5ee697f23ccf9f37b74ed12c7c/src/yuzu/game_list.cpp#L563


Add to Desktop (windows and linux) GameListShortcutTarget::Desktop
target_directory = home_path / "Desktop";
/\old / new \/
QString desktopPath = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
target_directory = desktopPath.toUtf8().toStdString();


Add to Applications Menu (linux) GameListShortcutTarget::Applications
target_directory = (xdg_data_home == nullptr ? home_path / ".local/share" : xdg_data_home) / "applications";
/\old / new \/
QString applicationsPath = QStandardPaths::writableLocation(QStandardPaths::ApplicationsLocation);
target_directory = applicationsPath.toUtf8().toStdString();

german77:
I meant to make the code simpler. So it's easier to read instead of having a different path per OS.

 ```switch (target) {
    case GameListShortcutTarget::Desktop: {
        const QString desktopPath =
            QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
        target_directory = desktopPath.toUtf8().toStdString();
        break;
    }
    case GameListShortcutTarget::Applications: {
        const QString applicationsPath =
            QStandardPaths::writableLocation(QStandardPaths::ApplicationsLocation);
        target_directory = applicationsPath.toUtf8().toStdString();
        break;
    }
    default:
        return;
    }

    QDir dir(QString::fromStdString(target_directory.generic_string()));
    if (!dir.exists()) {
       QMessageBox::critical(this, tr("Create Shortcut"),
                             tr("Cannot create shortcut. Path \"%1\" does not exist.")
                                 .arg(QString::fromStdString(target_directory.generic_string())),
                             QMessageBox::StandardButton::Ok);
       return;
    }
```

File: src/yuzu/main.cpp [hunk 1]
```diff
@@ -2869,18 +2871,13 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
...
-    const char* home = std::getenv("HOME");
-#endif
-    const std::filesystem::path home_path = (home == nullptr ? "~" : home);
-    const char* xdg_data_home = std::getenv("XDG_DATA_HOME");
 
+    // Determine target directory for shortcut
     if (target == GameListShortcutTarget::Desktop) {
-        target_directory = home_path / "Desktop";
-        if (!Common::FS::IsDir(target_directory)) {
+        QString desktopPath = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
```
FearlessTobi, line 2877:
```suggestion
        const auto desktopPath = QStandardPaths::writableLocation(QStandardPaths::DesktopLocation);
```

src/yuzu/main.cpp
```diff
@@ -2869,18 +2871,13 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
...
+        target_directory = desktopPath.toUtf8().toStdString();
+        QDir dir(QString::fromStdString(target_directory.generic_string()));
```
FearlessTobi, line 2879:
```suggestion
        const QDir dir(QString::fromStdString(target_directory.generic_string()));
```
also why are we casting from qstring to std::string, then back to qstring?

DanielSvoboda: I was having problems mainly with a 'Á' character on my desktop "F:\Area de Trabalho", as you suggest it should be?

DanielSvoboda:
Thanks german77, this way the code is cleaner.

src/yuzu/main.cpp
```diff
@@ -2869,38 +2871,32 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
...
-                this, tr("Create Shortcut"),
-                tr("Cannot create shortcut in applications menu. Path \"%1\" "
-                   "does not exist and cannot be created.")
-                    .arg(QString::fromStdString(target_directory.generic_string())),
-                QMessageBox::StandardButton::Ok);
-            return;
-        }
+    switch (target) {
+    case GameListShortcutTarget::Desktop: {
+        const QString desktop_Path =
```
FearlessTobi, line 2877:
```suggestion
        const QString desktop_path =
```

german77:
Please fix clang format issues.

DanielSvoboda:
I added an 'if' to make it look like it was before. This code will attempt to use QStandardPaths to find the applications directory. If that fails, it will resort to using the ~/.local/share/applications directory, which is a common location for application shortcuts in Linux.

DanielSvoboda:
some variables that I deleted were used elsewhere, include them where they are used

DanielSvoboda:
Now everything is ok 👍🏼

german77 (approved):

src/yuzu/main.cpp
```diff
@@ -2869,44 +2871,50 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
...
+                              QMessageBox::StandardButton::Ok);
+        return;
     }
 
     const std::string game_file_name = std::filesystem::path(game_path).filename().string();
     // Determine full paths for icon and shortcut
 #if defined(__linux__) || defined(__FreeBSD__)
+    const char* home = std::getenv("HOME");
+    const std::filesystem::path home_path = (home == nullptr ? "~" : home);
+    const char* xdg_data_home = std::getenv("XDG_DATA_HOME");
```
v1993, line 2913: Welp, this was merged already, but anyways: `QStandardPaths::GenericDataLocation` can be used to get this path easily.

german77: I merged this because I'm rewriting this whole function. As it's very OS dependent and just looks like a patched code. Just as an example #11761 is broken due some variables initialized within these macros.

Issue #11761, opened 2023-10-12
Regression: [macOS] Build: Yuzu no longer builds on Mac


shinra-electric:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

N/A

### Description of Issue

Building failed with the following: 
```
[1741/2271] Building CXX object src/co...core.dir/hle/service/caps/caps_a.cpp.o
FAILED: src/core/CMakeFiles/core.dir/hle/service/caps/caps_a.cpp.o 
```

This is likely due to changes introduced in commit 8347e5c which touches `caps_a.cpp`

### Expected Behavior

The last time I built successfully was on the 8th of October, so the change must have happened between then and now.

I should bisect the commits to find the exact one, but I'm a bit short on time and it seems likely that 8347e5c was the culprit.

### Reproduction Steps

Build Yuzu for macOS 

### Log File

[Yuzu terminal.txt](https://github.com/yuzu-emu/yuzu/files/12882140/Yuzu.terminal.txt)

### System Configuration

CPU: M1 Pro
GPU/Driver: M1 Pro
RAM: 16GB
OS: 14.0 Sonoma

german77:
We need a CI for mac. Fixing the issue soon

shinra-electric:
The fix in #11765 did allow the build to get further, but it still fails further along: 

```
[2254/2275] Building CXX object src/yuzu/CMakeFiles/yuzu.dir/main.cpp.o
FAILED: src/yuzu/CMakeFiles/yuzu.dir/main.cpp.o 
```

[Yuzu Terminal.txt](https://github.com/yuzu-emu/yuzu/files/12890556/Yuzu.Terminal.txt)

shinra-electric:
Fixed by #11920

PR #11777, opened 2023-10-13
yuzu: refactor shortcut code

german77:
Makes this feature a bit more OS agnostic. Preventing issues like #11761.

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -77,7 +77,6 @@ static FileSys::VirtualFile VfsDirectoryCreateFileWrapper(const FileSys::Virtual
 #include <QPushButton>
 #include <QScreen>
 #include <QShortcut>
-#include <QStandardPaths>
 #include <QStatusBar>
 #include <QString>
 #include <QSysInfo>
@@ -100,7 +99,6 @@ static FileSys::VirtualFile VfsDirectoryCreateFileWrapper(const FileSys::Virtual
 #include "common/scm_rev.h"
 #include "common/scope_exit.h"
 #ifdef _WIN32
-#include <shlobj.h>
 #include "common/windows/timer_resolution.h"
 #endif
 #ifdef ARCHITECTURE_x86_64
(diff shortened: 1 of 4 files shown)


german77:
Seems to work fine on windows. I still need to confirm on linux.

DanielSvoboda:
I also did the test, as it changes the code I had sent, and it worked perfectly on my Windows. Even though I have my dekstop location on another unit, and the language with special characters.

german77:
I made sure too keep compatibility with your changes.

src/yuzu/main.cpp
```diff
@@ -2962,39 +2902,41 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
...
 
-    QImage icon_data =
+    std::string icon_name{};
+#ifdef _WIN32
+    icon_name = program_id == 0 ? fmt::format("yuzu-{}.ico", game_file_name)
+                                : fmt::format("yuzu-{:016X}.ico", program_id);
+#else
+    icon_name = program_id == 0 ? fmt::format("yuzu-{}.png", game_file_name)
+                                : fmt::format("yuzu-{:016X}.png", program_id);
+#endif
```
v1993, line 2912: I feel like this belongs in `SaveIconToFile`, since that's where platform-specific icon-saving code lives now. It would make sense to handle file extension there as well. This way, `OnGameListCreateShortcut` can be rid of platform-specific code completely.

src/yuzu/main.cpp
```diff
@@ -2962,39 +2902,41 @@ void GMainWindow::OnGameListCreateShortcut(u64 program_id, const std::string& ga
...
 
-    if (!CreateShortcut(shortcut_path.string(), title, comment, icon_path.string(),
-                        yuzu_command.string(), arguments, categories, keywords)) {
+    std::filesystem::path shortcut_path = "";
+#ifdef _WIN32
+    shortcut_path = fmt::format("{}/{}.lnk", target_directory.toUtf8().toStdString(), title);
+#else
+    shortcut_path =
+        fmt::format("{}/yuzu-{}.desktop", target_directory.toUtf8().toStdString(), title);
+#endif
```
v1993, line 2935: Same story - IMO this would be better moved to CreateShortcut.

src/yuzu/util/util.cpp
```diff
@@ -42,7 +50,27 @@ QPixmap CreateCirclePixmapFromColor(const QColor& color) {
...
+
+    const QString applications_path =
+        QStandardPaths::writableLocation(QStandardPaths::ApplicationsLocation);
+
+    if (!applications_path.isEmpty()) {
+        return applications_path;
+    }
+
+    // If Qt fails to find the application path try to use the generic location
+    return QStandardPaths::writableLocation(QStandardPaths::GenericDataLocation);
```
v1993, line 70: I don't think it's a good idea to default to `GenericDataLocation`. It is `~/.local/share` on unix-like and `C:/Users/<USER>/AppData/Local` on Windows.

As an aside, IMO using a `switch` statement would make this function look better.

src/yuzu/util/util.cpp
```diff
@@ -42,7 +50,27 @@ QPixmap CreateCirclePixmapFromColor(const QColor& color) {
     return circle_pixmap;
 }
 
-bool SaveIconToFile(const std::string_view path, const QImage& image) {
+QString GetTargetPath(GameListShortcutTarget target) {
```
v1993, line 53:
```suggestion
QString GetShortcutPath(GameListShortcutTarget target) {
```

Sounds way less generic.

src/yuzu/util/util.cpp
```diff
@@ -136,6 +165,66 @@ bool SaveIconToFile(const std::string_view path, const QImage& image) {
...
+        return false;
+    }
+
+    hres = persist_file->Save(Common::UTF8ToUTF16W(shortcut_path).data(), TRUE);
+    if (FAILED(hres)) {
+        return false;
+    }
+
+    persist_file->Release();
+    shell_link->Release();
```
v1993, line 225: You're leaking pointers in case of failure. Put these into `SCOPE_EXIT` blocks.

v1993 (changes requested):

PR #11786, opened 2023-10-14
host1x/codecs: enable CUDA on Linux

v1993:
The code in question disabled use of CUDA/NVDEC hardware decoder on all Linux machines, including those where it works just fine. The only known setup where this caused problems and led to inclusion of this code is having two NVIDIA+AMD GPUs installed, which is a fairly uncommon case that with this change only requires user to select "CPU Video Decoding" for NVDEC emulation.

I've confirmed that this allows GPU decoding to work on my Linux+Nvidia setup.

### DIFF
--- a/src/video_core/host1x/codecs/codec.cpp
+++ b/src/video_core/host1x/codecs/codec.cpp
@@ -137,16 +137,6 @@ bool Codec::CreateGpuAvDevice() {
                 break;
             }
             if ((config->methods & HW_CONFIG_METHOD) != 0 && config->device_type == type) {
-#if defined(__unix__)
-                // Some linux decoding backends are reported to crash with this config method
-                // TODO(ameerj): Properly support this method
-                if ((config->methods & AV_CODEC_HW_CONFIG_METHOD_HW_FRAMES_CTX) != 0) {
-                    // skip zero-copy decoders, we don't currently support them
-                    LOG_DEBUG(Service_NVDRV, "Skipping decoder {} with unsupported capability {}.",
-                              av_hwdevice_get_type_name(type), config->methods);
-                    continue;
-                }
-#endif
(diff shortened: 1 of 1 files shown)


liamwhite (approved):

vasishath:
This breaks my case. I have a pascal MX150 which doesn't support NVDEC/NVENC but my intel GPU is supported via VAAPI. Yuzu tries to use cuda via ffmpeg, fails and then falls back to CPU video decoding. It should attempt to use other GPU backends as well (if available). 

Also, shouldn't it cache the "last successful" working video backend since everytime a video is played, this whole "available hw codecs" search is performed each time, which results in small stutters that I have noticed when games start a video. Maybe doing this once per app session would be better?

v1993:
Could you attach a log? I suppose some extra checking should be done before selecting CUDA decoder.

vasishath:
> Could you attach a log? I suppose some extra checking should be done before selecting CUDA decoder.

Sure.. I will upload here in a few moments.. These are the logs that would matter here:

```
[  36.660411] Service.NVDRV <Info> video_core/host1x/codecs/codec.cpp:SetTargetCodec:258: NVDEC video codec initialized to VP9
[  36.788061] Service.NVDRV <Info> video_core/host1x/codecs/codec.cpp:CreateGpuAvDevice:140: Using cuda GPU decoder
[vp9 @ 0x7ef2c464d380] Hardware is lacking required capabilities
[vp9 @ 0x7ef2c464d380] Failed setup for format cuda: hwaccel initialisation returned error.
[  36.801693] Service.NVDRV <Info> video_core/host1x/codecs/codec.cpp:GetGpuFormat:56: Could not find compatible GPU AV format, falling back to CPU
```

Issue #6731, opened 2021-07-26
Big meshes when using Vulkan in EA and Mainline Yuzu with Crash Team Racing. 


wesmoen:
The game worked normally before, even for a while when Project Hades got into EA. Since a certain EA version, Meshes of some models will be huge. This issue also came into Yuzu 700, versions prior to 700 don't have this issue.  

I already tried:
- renewing my cache
- Switching to OpenGL (Works as intended)
- Disabling mods
- GPU Accuracy to normal
- Updating GPU drivers. 

Running the game on a i7-2600k, 8GB DDR3-1333, GTX 970.

Can't really wrap my head around it. Thanks and keep up the good work!

![CrashbigobjectRacing](https://user-images.githubusercontent.com/20476068/126966280-0595767d-479f-43b9-8032-a8a323c4cab4.png)
![crashbigracing](https://user-images.githubusercontent.com/20476068/126966281-0a3a0fd3-d9f6-4952-8cae-bad59bd5f311.png)
![crashcurseracing](https://user-images.githubusercontent.com/20476068/126966286-daa9b197-9760-4bb7-bf58-9c7090f1c18a.png)
![CrashLizRacing](https://user-images.githubusercontent.com/20476068/126966289-c793c05b-bbd6-4062-819d-438f290dcbca.png)
![CrashOxideRacing](https://user-images.githubusercontent.com/20476068/126966292-b8951c95-649d-4c67-89bb-60d97167a1b2.png)
![crashtinyracing](https://user-images.githubusercontent.com/20476068/126966295-f1db0a0c-bf78-4702-936f-4c99bbafe517.png)

Log file:
[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/6877221/yuzu_log.txt)

JustBuddy:
+1, having the same issues as OP since the HADES Vulkan update on 700 and 701.
Specs:
i5 10400
GTX 1060 6GB
16GB RAM

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/6883377/yuzu_log.txt)

MauMauMlK:
I have the same problem :(

i5-7400
8GB RAM
GTX 160 6GB

OpenGL works fine, only on vulkan

nyxover:
![image](https://user-images.githubusercontent.com/22452629/128747556-18f86bde-016e-4c05-97c5-8c46d53ad97c.png)
SAME vulkan  👎 Work on OpenGL
i7 1080
[...]

Issue #10026, opened 2023-04-03
Audio Issues - Crash Team Racing Nitro Fueled


Javierdsv:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Build(s)

Early Access 3496

### Description of Issue

On patch 1.0.15 the game has terrible sound cracking and stuttering issues, game itself plays smooth as silk. 
I found that pressing the F4 key two times fixes the issue entirely, but this works only sometimes, and after each race/loading screen the issue is back, and you have to try the F4 two times trick. The times when this won't work, when pressing F4 two times the sound stops entirely and won't come back until you reset the game. 

The issue is exactly as depicted in [this video](https://www.youtube.com/watch?v=hMh3vOA-88M) 

The solution given on the closed [Issue # 9035](https://github.com/yuzu-emu/yuzu/issues/9035) about changing the sound settings to SDL will not work. None of the options there make any difference. 

### Expected Behavior

The expected behavior is for the sound to be correct. And it seems the emulator can already handle the sound for this game as hitting F4 sometimes solves the issue.

### Reproduction Steps

Open the game, and as soon as it reaches the "press a to continue" menu the sound issue begins, if addressed with F4 it also comes back after the loading screen at the start of any race stage.

### Log File

[yuzu_log.txt](https://github.com/yuzu-emu/yuzu/files/11134979/yuzu_log.txt)

### System Configuration

CPU: i7-7700K
GPU/Driver: Nvidia GTX 1080ti / 531.41
RAM: 32 GB
OS: Windows 10

cjunisl:
Exactly the same problem happens here. Tested with v1402 and EA 3507. Previously it was possible to work around this problem by pressing F4 twice to pause the game and resume, but now after pressing F4 the game is completely muted or yuzu crashes.

chon59:
same problem here ! f4 twice is muting the game :(

Tryndameree:
The best solution for the AUDIO problems, try this:

(Windows 10) Go To Task Manager, Go to Details, Find yuzu.exe, Right Click on it, Set Priority > High. Thats it.

Now play the game, or if u already in game just Press 2 times F4, and see if audio has fixed. But the best time to do this is before launching the game, ANY game. One more thingm if u CPU has 8 cores atleast, same on YUZU, right lick Set Affinity, and disable CPU 0, actually you will get more FPS disabling CPU 0, thats actually 1 core disabling, casuing other cores to do more on game... Just google it. 
[...]

PR #11789, opened 2023-10-14
Manually robust on Maxwell and earlier

### LINKED ISSUE #8215: Crash Team Racing - graphical glitches when using Vulkan
When playing CTR nitro fueled under vulkan api, most characters and kart becomes glitched (characters and karts becomes huge and all over the place), doesn't happen in ogl, but ogl seems to have  a memory leak issue.

Tested on latest early acces version.

Kelebek1:
Unfortunately it seems like Maxwell and Pascal cards have broken robust support, or something else strange is happening in Yuzu preventing it from working correctly. Crash Team Racing Nitro Fueled uses wildly invalid buffer indexes (values like 0x3D00001C), and they should just return 0 with robust enabled on the device, but Maxwell seems to instead just % the index with the buffer size, or & it with at least 0x10000 or something, and is returning "valid" results even with large indexes. 0x3D00001C returns the same value as 0x1C. No bits above bit 16 affect the returned result, so 0x20000 and above.

This PR also changes the arithmetic shifts to logical, since we don't want negative indexes, which also affects CTR.

I feel like there's something else strange going on in our shaders or their setup, or in how we enable robust or something which is causing it to fail. Ryujinx doesn't seem to suffer this issue, and the load does return 0, though that may be coincidence or something to do with our 16k buffer size their 4k, not sure. If anyone wants to investigate further, please do.

This should close #8215 and #6731.

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
@@ -111,16 +111,33 @@ Id GetCbuf(EmitContext& ctx, Id result_type, Id UniformDefinitions::*member_ptr,
     } else if (element_size > 1) {
         const u32 log2_element_size{static_cast<u32>(std::countr_zero(element_size))};
         const Id shift{ctx.Const(log2_element_size)};
-        buffer_offset = ctx.OpShiftRightArithmetic(ctx.U32[1], ctx.Def(offset), shift);
+        buffer_offset = ctx.OpShiftRightLogical(ctx.U32[1], ctx.Def(offset), shift);
     } else {
         buffer_offset = ctx.Def(offset);
     }
     if (!binding.IsImmediate()) {
         return ctx.OpFunctionCall(result_type, indirect_func, ctx.Def(binding), buffer_offset);
     }
+
     const Id cbuf{ctx.cbufs[binding.U32()].*member_ptr};
     const Id access_chain{ctx.OpAccessChain(uniform_type, cbuf, ctx.u32_zero_value, buffer_offset)};
-    return ctx.OpLoad(result_type, access_chain);
+    const Id val = ctx.OpLoad(result_type, access_chain);
+
+    if (offset.IsImmediate() || !ctx.profile.has_broken_robust) {
+        return val;
+    }
+
+    const auto is_float = UniformDefinitions::IsFloat(member_ptr);
+    const auto num_elements = UniformDefinitions::NumElements(member_ptr);
+    const std::array zero_vec{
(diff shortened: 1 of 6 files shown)


src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -111,16 +111,30 @@ Id GetCbuf(EmitContext& ctx, Id result_type, Id UniformDefinitions::*member_ptr,
...
+        return val;
+    }
+    std::array<Id, 4> const_vec{
+        ctx.Const(0.0f),
+        ctx.Const(0.0f),
+        ctx.Const(0.0f),
+        ctx.Const(0.0f),
+    };
+    const Id cond = ctx.OpULessThanEqual(ctx.TypeBool(), buffer_offset, ctx.Const(0xFFFFu));
+    const Id zero = ctx.OpCompositeConstruct(result_type, std::span(const_vec));
```
liamwhite, line 136: This will always construct a 4-element composite, but the pointer type may not be a 4-element composite, which will give a shader validation error.

Kelebek1: > This will always construct a 4-element composite, but the pointer type may not be a 4-element composite, which will give a shader validation error.

In the shaders it always made a uvec2, which is the result type. How else do I do it?

liamwhite: It made a uvec2 in the spirv asm? Are you sure?

Kelebek1: > It made a uvec2 in the spirv asm? Are you sure?

Seems to
>         %917 = OpULessThan %u1 %914 %u32_65535
        %918 = OpCompositeConstruct %u32x2 %f32_0 %f32_0 %f32_0 %f32_0
        %919 = OpSelect %u32x2 %917 %916 %918

liamwhite: It used the correct type but provided too many arguments, which won't pass spirv-val

liamwhite (approved):

cjunisl:
Yuzu EA 3940 - GTX 1050

ctr ea 3940

wesmoen:
Had some luck by opening the game without configuration. Looks like the issue is fixed on a GTX970. 

image

image

liamwhite:
@Kelebek1 this check should be done on pascal as well, I believe

cjunisl:
yuzu 1597
GTX 1050

Unfortunately, it doesn't seem to have been fixed on Pascal. 
 
image

There is also an audio bug as reported in #10026

PR #11798, opened 2023-10-15
Update vcpkg.json

YigaUpdate:
I have update "catch2" and "fmt"


abouvier:
You can remove the overrides altogether. This will use the latest version of each library and yuzu doesn't have compatibility issue with them anymore.

PR #11809, opened 2023-10-17
yuzu: Always ask when closing game with controller

german77:
Probably a personal preference but I definitively hate that games always ask on close by default. This changes that back to only when game request. But it will always ask when using the controller hotkey.

### DIFF
--- a/src/yuzu/hotkeys.cpp
+++ b/src/yuzu/hotkeys.cpp
@@ -181,6 +181,10 @@ bool ControllerShortcut::IsEnabled() const {
     return is_enabled;
 }
 
+bool ControllerShortcut::IsActive() const {
+    return active;
+}
+
 void ControllerShortcut::ControllerUpdateEvent(Core::HID::ControllerTriggerType type) {
     if (!is_enabled) {
         return;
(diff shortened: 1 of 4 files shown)


flodavid:
If this is a personal preference, why not just changing it in the settings ? I set it to ask by default with the keyboard shortcut to have the same behaviour between the controller and the keyboard, and because sometimes I want to pause with F4 and may it the F5 key by mistake.
Maybe we could add a setting to allow defining the behaviour depending on the input device ?

german77:
When using applets a pop up will appear when the applet is already closed. This is very annoying.

I use default settings because that's what we recommend to users. 

I know this was changed into 'always ask' because we added the controller shortcut. If you don't like having different behaviour I can revert that and simply keep the setting on 'ask based on game'

flodavid:
As I mentioned in the relative PR, behavior varies depending on games. Some always ask, others never, so I do not know if it would be a good idea not to always ask by default on the controller. At the same time, the shortcut is using 3 buttons now, so you would have very unlucky/clumsy to use it without wanting to.
We can still add an option in the settings to select which behavior the user wants for both shortcut.

reloxx13:
Please dont if its not possible to confirm closing question with controller.

flodavid:
Being able to interact with the dialogs was included in the PR that added the window asking for confirmation

Issue #6828, opened 2021-08-06
[CRASH] [VULKAN] Bravely Default 2 crashes at random


Killian-A:
Bravely Default 2 crashes randomly , can be anytime ( zone load, fight load, menu open ... ) and can happen after 5 or 30 minutes . 
Every crash led to the same backtrace  : 

[yuzu_log.txt.old.zip](https://github.com/yuzu-emu/yuzu/files/6945936/yuzu_log.txt.old.zip)

radmaruto:
this fix crash https://www.youtube.com/watch?v=T41R-eK8ltM&t=320s&ab_channel=daviddegraeve

bunkbail:
Here is a workaround for the crashes caused by Vulkan VRAM memory leaks. In the log file they usually look something like this: 

`HW.Memory <Error> core\memory.cpp:ReadBlockUnsafe:256: Unmapped ReadBlock` 

Download and extract the following file into the game's mod folder and enable the CrashFix mod in yuzu.

[CrashFix.zip](https://github.com/yuzu-emu/yuzu/files/8058685/CrashFix.zip)

What it does is simply changing the in-game UE4 settings under `[/Script/Engine.RendererSettings]` in the `DefaultEngine.ini` file from:

> r.RenderTargetPoolMin=200
r.Streaming.PoolSize=600

to:

> r.RenderTargetPoolMin=400
r.Streaming.PoolSize=1000

Without the mod, I'm getting crashes every few mins. With it, I haven't encountered any crashes for far after 8 hours of gameplay. I'm on Linux and I don't have Windows so I can't test it there but I think it should work the same.

redagito:
Got a random crash during battle, last few lines of the log are
`
[6409.436986] Debug <Critical> video_core/texture_cache/texture_cache.h:operator ():2080: Assertion Failed!
[6409.437156] Debug <Critical> video_core/texture_cache/texture_cache.h:operator ():2080: Assertion Failed!
[6409.437183] HW.Memory <Error> core\memory.cpp:operator ():711: Unmapped GetPointer @ 0x0000001B941B0000`

PR #11847, opened 2023-10-22
emit_glsl_warp: Fix shfl_in_bounds conditional

ameerj:
`readInvocationARB` needs all participating lanes to be active to return valid results if that lane's index is to be read from. The previous `shfl_in_bounds` condition could mask some lanes from participating, causing the data read from their invocation to be undefined.

Instead, the `shfl_in_bounds` check should happen after all lanes participate in `readInvocationARB` to avoid undefined behavior.

Verified to fix previously failing nxgpucatch tests. Likely fixes graphical corruption in some titles.

### DIFF
--- a/src/shader_recompiler/backend/glsl/emit_glsl.cpp
+++ b/src/shader_recompiler/backend/glsl/emit_glsl.cpp
@@ -242,6 +242,7 @@ std::string EmitGLSL(const Profile& profile, const RuntimeInfo& runtime_info, IR
     }
     if (program.info.uses_subgroup_shuffles) {
         ctx.header += "bool shfl_in_bounds;";
+        ctx.header += "uint shfl_result;";
     }
     ctx.code.insert(0, ctx.header);
     ctx.code += '}';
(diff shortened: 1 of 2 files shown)


Kewlan:
I've been playing Bravely Default 2 for around a week. After adding the [Crash Fix](https://github.com/yuzu-emu/yuzu/issues/6828#issuecomment-1038741975) mod I've never had freezes or crashes. (I also use the 60 fps mod.) 
After adding this PR yesterday, I got a freeze at the company logos before the title screen, and one crash around 5 to 10 minutes into gameplay. I removed this PR, played for a while, and didn't get any of those issues again. 

Here's the log with both the freeze and crash:
[yuzu_log_bd2_crash.txt](https://github.com/yuzu-emu/yuzu/files/13066518/yuzu_log_bd2_crash.txt)

vonchenplus:
> I've been playing Bravely Default 2 for around a week. After adding the [Crash Fix](https://github.com/yuzu-emu/yuzu/issues/6828#issuecomment-1038741975) mod I've never had freezes or crashes. (I also use the 60 fps mod.) After adding this PR yesterday, I got a freeze at the company logos before the title screen, and one crash around 5 to 10 minutes into gameplay. I removed this PR, played for a while, and didn't get any of those issues again.
> 
> Here's the log with both the freeze and crash: [yuzu_log_bd2_crash.txt](https://github.com/yuzu-emu/yuzu/files/13066518/yuzu_log_bd2_crash.txt)
> 
> I do see this is marked for opengl, and I was using vulkan, so maybe it was just a huge coincidence?

This PR is only for opengl, it won't have any effect on vulkan.

liamwhite (approved):

PR #11852, opened 2023-10-22
input_common: joycon: Move vibrations to a queue

german77:
Currently switch controllers need to wait for the reply of the controller. This takes time. Games like Mario Wonder love vibration. So you start feeling those little stutters from vibration. We can avoid this by using a queue and just returning a delayed vibration result.

### DIFF
--- a/src/input_common/helpers/joycon_driver.cpp
+++ b/src/input_common/helpers/joycon_driver.cpp
@@ -139,7 +139,7 @@ void JoyconDriver::InputThread(std::stop_token stop_token) {
     input_thread_running = true;
 
     // Max update rate is 5ms, ensure we are always able to read a bit faster
-    constexpr int ThreadDelay = 2;
+    constexpr int ThreadDelay = 3;
     std::vector<u8> buffer(MaxBufferSize);
 
     while (!stop_token.stop_requested()) {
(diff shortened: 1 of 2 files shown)


Kewlan:
Are you able to test with a ProCon connected through USB?
There's an issue where Mario Kart 8 lags a lot that, based on searching through the discord, seems to be caused when a ProCon connected through USB is vibrating.
This PR seems to fix that lag, but now the input freezes after the vibration. In my case, after a race starts I hold Y to accelerate, and the input freezes, there's no lag, but I just go forward.

german77:
Fixed the issue with input. Lag Please test again.

german77:
I did some measurements. Vibration can take as little as 150us and as much as 18ms. With the median being around 4ms. I wonder if it would possible to have consistent times.

Kewlan:
> Fixed the issue with input. Lag Please test again.

Input still freezes, but only _while_ vibrating.

german77:
Are you using SDL or the custom joycon driver?. Seems weird that with last commit.

Kewlan:
I was using direct ProCon driver. I enabled it so long ago I must have forgotten it was marked experimental. 
Disabled it and all input freezing is gone!

german77:
Well this PR only applies to the direct joycon driver. SDL has been working fine for a while

src/input_common/helpers/joycon_driver.cpp
```diff
@@ -163,6 +163,17 @@ void JoyconDriver::InputThread(std::stop_token stop_token) {
...
         }
 
+        if (!vibration_queue.Empty()) {
+            VibrationValue vibration_value;
+            vibration_queue.Pop(vibration_value);
+            last_vibration_result = rumble_protocol->SendVibration(vibration_value);
+        }
+
+        // We can't keep up with vibrations. Start skipping.
+        while (vibration_queue.Size() > 6) {
```
Firepal, line 173: Just curious; what's the heuristic on having a cap of specifically 6 max vibrations?

german77: if we remove the queue limit you have vibrations for a long time after the actual vibration. 6 will have a worst case of 90ms of delay which is too little to notice but big enough to hear it.

liamwhite (approved):

PR #11892, opened 2023-10-26
service: am: Implement ISelfController::SaveCurrentScreenshot

german77:
Fixes the regression caused by #11880 and implements `SaveCurrentScreenshot`. This function is a bit more complicated from the others since the game doesn't provide the image data. This allows to take screenshots on `Pokémon Scarlet` on the latest game update.

0100a3d008c5c000_2023-10-26_17-15-17-000

### DIFF
--- a/src/core/hle/service/am/am.cpp
+++ b/src/core/hle/service/am/am.cpp
@@ -31,6 +31,7 @@
 #include "core/hle/service/apm/apm_controller.h"
 #include "core/hle/service/apm/apm_interface.h"
 #include "core/hle/service/bcat/backend/backend.h"
+#include "core/hle/service/caps/caps_su.h"
 #include "core/hle/service/caps/caps_types.h"
 #include "core/hle/service/filesystem/filesystem.h"
 #include "core/hle/service/ipc_helpers.h"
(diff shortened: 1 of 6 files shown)


abouvier:
What was the regression? :p
You should keep `stbi_write_png_to_func` and fix my `PNGToMemory` function instead. `stbi_write_png_to_mem` is not supposed to be used. At least put the forward declaration in `src/common/stb.h` instead of the upstream header.

german77:
PNGToMemory was writing to nothing. So the saved image was a blank image. 

stbi_write_png_to_func is an async operation which really conflicts with how the service is operating since this function should return if writing to the image failed.

abouvier:
But it's just a wrapper for `stbi_write_png_to_mem`. If `stbi_write_png_to_mem` fails, then `stbi_write_png_to_func` will return. And `PNGToMemory` cannot really fail, it's just a call to `memcpy`. Maybe the vector is wrongly initialized with `reserve`?

german77:
I know it's just a wrapper but the way it returns the data makes it completely unusable. The major issue is that needs to be an static function. This can't modify any members of an AlbumManager unless I make them global. We mustn't use global variables for core.

I could technically move all fs operations inside PNGToMemory so all is saved properly. But then I can't return the proper fs error in case something goes wrong. 

It could be simply lack of experience. But I honestly can't manage to make it work that way.

src/core/hle/service/am/am.cpp
```diff
@@ -702,9 +703,17 @@ void ISelfController::SetAlbumImageTakenNotificationEnabled(HLERequestContext& c
 void ISelfController::SaveCurrentScreenshot(HLERequestContext& ctx) {
     IPC::RequestParser rp{ctx};
 
-    const auto album_report_option = rp.PopEnum<Capture::AlbumReportOption>();
+    const auto report_option = rp.PopEnum<Capture::AlbumReportOption>();
 
-    LOG_WARNING(Service_AM, "(STUBBED) called. album_report_option={}", album_report_option);
+    LOG_INFO(Service_AM, "called. report_option={}", report_option);
```
FearlessTobi, line 708:
```suggestion
    LOG_INFO(Service_AM, "called, report_option={}", report_option);
```

abouvier:
In fact I have no idea how to test this screenshot service. Can you try this function instead? :p

```cpp
static void PNGToMemory(void* context, void* data, int len) {
    std::vector<u8>* png_image = static_cast<std::vector<u8>*>(context);
    unsigned char* png = static_cast<unsigned char*>(data);
    png_image->insert(png_image->end(), png, png + len);
}
```

german77:
Alright that worked. Now it just hit me that context here is actually my array of data. I wonder then why the simple memcpy didn't work

abouvier:
> Alright that worked. Now it just hit me that context here is actually my array of data. I wonder then why the simple memcpy didn't work

Because I don't master vectors :p
`png_image->reserve(len)` allocated enough memory to store the png data, and `memcpy` actually copied the data, but the size of the vector was still 0 and thus considered empty for subsequent operations.

liamwhite (approved):

PR #11933, opened 2023-11-01
ci: android: Use signing key if available

t895:
Credits to the Citra team for the CI code


liamwhite (approved):

FearlessTobi (approved):
Can we merge this before the secrets have been added to the ci?

t895:
It's an optional step so existing builds won't be impacted

PR #11977, opened 2023-11-06
Update CMakeLists.txt

SamayXD:
Updated Comments for better readability.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -21,7 +21,7 @@ if (MSVC)
     # Avoid windows.h from including some usually unused libs like winsocks.h, since this might cause some redefinition errors.
     add_definitions(-DWIN32_LEAN_AND_MEAN)
 
-    # Ensure that projects build with Unicode support.
+    # Ensure that projects are built with Unicode support.
     add_definitions(-DUNICODE -D_UNICODE)
 
     # /W4                 - Level 4 warnings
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ if (MSVC)
         /we4265 # 'class': class has virtual functions, but destructor is not virtual
         /we4388 # 'expression': signed/unsigned mismatch
         /we4389 # 'operator': signed/unsigned mismatch
-        /we4456 # Declaration of 'identifier' hides previous local declaration
+        /we4456 # Declaration of 'identifier' hides a previous local declaration
```
v1993, line 72: These are copied verbatim from warning messages and as such should be kept as is.

src/CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ if (MSVC)
 
         /wd4100 # 'identifier': unreferenced formal parameter
         /wd4324 # 'struct_name': structure was padded due to __declspec(align())
-        /wd4201 # nonstandard extension used : nameless struct/union
+        /wd4201 # nonstandard extension used: nameless struct/union
```
v1993, line 87: Ditto.-

v1993 (changes requested):

src/CMakeLists.txt
SamayXD: Oh Sure.

src/CMakeLists.txt
SamayXD: Is it fine now?

v1993: Yep.

liamwhite (approved):

PR #11984, opened 2023-11-08
shared_translation: Call tr more directly

lat9nq:
Qt can't parse tr called within a macro, so we must call it on each string. Qt still seems unhappy with the way I called tr, but it's getting the strings now nevertheless.

To figure this out, I had to create a workaround so that lupdate would run on my system. The path I created results in a number of warnings across the Qt frontend. I have no ability to tell if these were existing already or not, or if this is an issue with the way I'm calling lupdate.

### DIFF
--- a/src/yuzu/CMakeLists.txt
+++ b/src/yuzu/CMakeLists.txt
@@ -252,6 +252,7 @@ file(GLOB_RECURSE THEMES ${PROJECT_SOURCE_DIR}/dist/qt_themes/*)
 if (ENABLE_QT_TRANSLATION)
     set(YUZU_QT_LANGUAGES "${PROJECT_SOURCE_DIR}/dist/languages" CACHE PATH "Path to the translation bundle for the Qt frontend")
     option(GENERATE_QT_TRANSLATION "Generate en.ts as the translation source file" OFF)
+    option(WORKAROUND_BROKEN_LUPDATE "Run lupdate directly through CMake if Qt's convenience wrappers don't work" OFF)
 
     # Update source TS file if enabled
     if (GENERATE_QT_TRANSLATION)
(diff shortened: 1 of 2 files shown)


FearlessTobi:
The changes in `yuzu/configuration/shared_translation.cpp` look good.

I'm not so sure about the lupdate stuff. I just rebuilt yuzu from scratch with the cmake flags for translations set, and generating the translations worked fine.
What OS/compiler are you on? What cmake command did you use and how did you try to generate the translations?

lat9nq:
> The changes in `yuzu/configuration/shared_translation.cpp` look good.
> 
> I'm not so sure about the lupdate stuff. I just rebuilt yuzu from scratch with the cmake flags for translations set, and generating the translations worked fine. What OS/compiler are you on? What cmake command did you use and how did you try to generate the translations?

```
System:
  Host: Qykopi Kernel: 6.5.8-273-tkg-linux-tkg-pds arch: x86_64 bits: 64
    Desktop: GNOME v: 45.1 Distro: Arch Linux
```
`cmake version 3.27.7`
`cmake --build . --target translations`

CMake configure arguments in vscodium:
```
"cmake.configureArgs": [
  "-DCMAKE_CXX_FLAGS_DEBUG='-O0 -g -pipe -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection'",
  "-DCMAKE_CXX_FLAGS_RELEASE='-O3 -pipe -Xclang -load -Xclang /usr/lib64/LLVMPolly.so -mllvm -polly -mllvm -polly-parallel -mllvm -polly-vectorizer=stripmine -flto=thin -march=native -g'",
  "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON",
  "-DYUZU_TITLE_CONFIG_DIR='/home/lat9nq/repos/yuzu-configs'",
  "-DYUZU_USE_BUNDLED_VCPKG=OFF",
  "-DYUZU_USE_PRECOMPILED_HEADERS=OFF",
  "-DCMAKE_CXX_FLAGS_RELWITHDEBINFO='-O2 -g -pipe -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection'",
  "-DUSE_DISCORD_PRESENCE=OFF",
  "-DENABLE_QT_TRANSLATION=ON",
  "-DGENERATE_QT_TRANSLATION=ON",
  "-DWORKAROUND_BROKEN_LUPDATE=ON"
]
```

Compiler is `clang version 16.0.6`

FearlessTobi:
I tried to reproduce this with @liamwhite on his linux system, and it seems like he had similar problems.
So I guess this is fine to merge?

liamwhite (approved):

purinchu:
I saw this from the Yuzu monthly report.  I just wanted to note that it helps to understand that there's two parts to using translated strings in Qt.

1. Marking the string to be translated, and
2. Loading the translated string based on the source language string.

In Qt, the `tr()` function often performs double-duty for both, but this only works when `tr()` is run on the string to be translated (e.g. `tr("Cancel")` in a dialog). This marks the string `"Cancel"` to be extracted by Qt linguist, and if translations are available at runtime, also allows `tr()` to lookup the translation for "Cancel" at that time.

If you want to use `tr` to translate the contents of a string at runtime indirectly, this *is* possible but you need to have a different way to flag the strings that will be translated so that they are picked up by Qt linguist.

To this end, Qt provides no-op macros that are used to do exactly this, like `QT_TR_NOOP` and related functions. You can e.g. wrap strings in an array of messages in this call to mark them for translation. You will still need to use `tr()` later at the place where the translated string is to be shown, but this time the source text will be in the catalog at runtime.

```cpp
static const char *const dialog_strs[] = {
    QT_TR_NOOP("OK"),
    QT_TR_NOOP("Cancel"),
    QT_TR_NOOP("Apply"),
};

// ...

// will pull translated text if present
const char *btnText = tr(dialog_strs[btnIndex]);
```

There are [some good Qt docs on this](https://doc.qt.io/qt-6/i18n-source-translation.html#c-use-noop-macros) as well.

Apologies if this is all understood by now but it's a common point of friction when trying to localize and internationalize Qt-based apps so I thought it might help.

PR #12007, opened 2023-11-12
core: hid: Split SL and SR buttons

german77:
Plug your leaks! Now you can map SL and SR buttons on dual joycon. Allowing this wario minigame to be playable.

image


FearlessTobi (approved):
Code LGTM.

Will this break any existing controller configs? Does this need additional work for Android?

german77:
It will only break config for single right joycon. The rest would remain the same. Android doesn't have support for SL/SR buttons.

liamwhite (approved):

PR #12008, opened 2023-11-12
Improved The Licance For Viewers

Ishan0981:
•Converted The Licance To .md file


FearlessTobi:
Our License is a legal document and should not be changed randomly.

PR #12014, opened 2023-11-12
android: Shrink branding images

t895:
The yuzu logo could get in the way of the main app content in the settings tab and the about page. This shrinks the logo in both instances and adds a custom landscape layout for the about page.

### Settings tab

Before -
image

After -
image

### About page

Before -
image

After -
image


liamwhite:
This layout makes me wonder if it is even needed to have the branding on the settings pane. I checked a bunch of apps on my phone and yuzu is the only one that puts branding up front when viewing settings.

t895:
There's only a couple apps that I know that have branding this same way but I still think it is a nice extra thing to add. And now since it's much smaller, it isn't in the way and can reasonably be kept.

FearlessTobi (approved):
Code LGTM.

Imo the logo on the about page should be vertically centered though :P

t895:
I considered vertically centering it but that makes it look a little off.

PR #12018, opened 2023-11-13
android: Settings visual cleanup

t895:
- Adds icons and descriptions to the top-level settings page
- Removes dividers between each settings item
- Makes all transitions in settings the same
- Removes General submenu
- Moves Limit Speed to System
- Moves PiP to Graphics
- Moves CPU accuracy to Debug

### Top Level
Before -
image

After - 
image

### Within each submenu
Before -
image

After -
image


FearlessTobi (approved):
Nice, looks much cleaner!
I don't know if we need the "etc" everywhere, but thats probably just personal preference.

t895:
I'm not sure what else I'd put there besides an ellipsis or something. What do you think?

FearlessTobi:
WhatsApp doesn't put anything there for example:

WhatsApp Image 2023-11-13 at 03 21 46(1)

But its probably just a matter of personal preference.

t895:
Well, I think I like that version more. Lemme try it out.

t895:
Yeah I think this is better - pushing now
image

PR #12020, opened 2023-11-13
 Yuzu emulator

Chamatvha:
Aot


german77:
This PR seems like a mistake.

PR #12036, opened 2023-11-15
Query Cache: Disable write syncing on Android

FernandoS27:
This should fix the crashes on Turnip on games that need query summing like Red Dead Redemption and LM3

### DIFF
--- a/src/video_core/renderer_vulkan/vk_query_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_query_cache.cpp
@@ -211,6 +211,13 @@ public:
             return;
         }
         PauseCounter();
+        const auto driver_id = device.GetDriverID();
+        if (driver_id == VK_DRIVER_ID_QUALCOMM_PROPRIETARY ||
+            driver_id == VK_DRIVER_ID_ARM_PROPRIETARY || driver_id == VK_DRIVER_ID_MESA_TURNIP) {
+            pending_sync.clear();
+            sync_values_stash.clear();
+            return;
+        }
         sync_values_stash.clear();
         sync_values_stash.emplace_back();
         std::vector<HostSyncValues>* sync_values = &sync_values_stash.back();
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_vulkan/vk_query_cache.cpp
```diff
@@ -211,6 +211,9 @@ class SamplesStreamer : public BaseStreamer {
             return;
         }
         PauseCounter();
+#ifdef ANDROID
```
liamwhite, line 214: Make this a driver ID check instead of conditional compilation

brujo5:
@FernandoS27 
Sir, I ask this of you since you are currently fixing problems with turnip, could you check a bug that has existed since Yuzu was released for Android with the Dragon Ball Fighterz game?  In the middle of the fights when executing super attacks or final attacks, the game freezes and the use of RAM increases abruptly until it is all consumed and the game crashes.  That bug happens on any Android phone.  I reported it a long time ago but no one paid attention and I'm sure it has to do with what you are fixing.

djdp77:
Just out of curiosity, is this fixing an issue with Yuzu itself or rather you circumventing an issue Turnip has? Might be worth reporting to Mesa if the latter

FernandoS27:
> Just out of curiosity, is this fixing an issue with Yuzu itself or rather you circumventing an issue Turnip has? Might be worth reporting to Mesa if the latter

It's more of a trial and error. Some games that are crashing require some special handling GPU side. So I was looking to see if this was the cause.

djdp77:
Got it, thanks! If it's confirmed that avoids the issue, again, would it be avoiding a problem Yuzu has or Turnip has?

i.e. you are disabling write syncing because does not work with Turnip. For it to work as it does with Qualcomm rather than disabling just it, would change have to come from Yuzu or Turnip?

boludoz:
image

PR #12074, opened 2023-11-18
Implement Native Code Execution (NCE)

GPUCode:
This initially started as a port of Skyline's native code execution (NCE) backend to yuzu, however, due to major architectural differences between the two emulators, it has turned into an (almost) complete reimplementation of the concept. Regardless, Skyline has proved a crucial reference throught the development of this CPU backend, so big credit to them.

**Which games will benefit?**

All games that use a 39-bit address space can utilize the new native code execution backend. This means that 32-bit games, such as Mario Kart 8 Deluxe or Captain Toad: Treasure Tracker, and 36-bit games will not benefit and will fallback to Dynarmic instead. Examples of 36-bit games are the base game of The Legend of Zelda: Breath of the Wild and ARMS. If your game is 36-bit, it is recommended that you install the latest game update, as this turns it into a 39 bit game which will be able to use NCE. NCE improves the overhead of CPU emulation, so GPU bottlenecked games are going to see minimal differences.

**How does Skyline do it?**

Skyline reserves some space before the text section to store a series of patches, small assembly functions that interface between the guest and host. In addition, the `tpidr_el0` register is changed to point to a custom per thread structure that contains useful global state. Each supervisor call is rewritten to jump to a dedicated patch, the same thing occurs for accesses to `tpidr_el0` or `tpidrro_el0` that are redirected to the guest variants. Running SVCs is treated as a callback, the patch will jump to the SvcHandler and return from it when host execution is finished to enter back to the guest. If the SVC causes a reschedule, the current thread is put to sleep in the handler until it's awoken again in the future.

**How does yuzu do it?**

For yuzu, the base principle is the same. For each module we reserve an additional segment called the patch segment, in which patches are written. Patches may be of multiple types, either pre-text (before the text section, like Skyline) or post-data (after the data section). In general, post-data patches are prefered when possible, as they're less intrusive and easier to emit without the need to know their size in advance. Some games, however, contain modules where it's not possible to use post-data patching due to the limits of relative branching, so for that we use the more compatible pre-text patching like Skyline.

To make this separation easier to implement, the generation of the patch segment is split into two parts: the patching step where supervisor calls, accesses to TLS and counter values are detected, with a patch function being emitted for each and the relocation step, where locality sensitive information such as `patch<->module` branch offsets are resolved based on the position of the patch segment relative to the text segment. Patches are emitted using [oaknut](https://github.com/merryhime/oaknut) as a way to avoid hardcoding assembly instructions. This should make maintaing the code much more accessible and less error prone.

yuzu's design necessitates returning to SVCs instead of calling them like a function from guest code. For this reason, two patches are emitted per SVC instruction, a pre-trampoline which is used for `guest->host` transitions and the post-trampoline for `host->guest` transitions. The functions used to call the trampolines from the host are written directly in assembly and not using inline assembly like Skyline. This avoids stack cookie generation in locations where `tpidr_el0` might not be the expected value.

However, there are cases where we need to enter guest code that wasn't paused by an SVC, for example when resuming execution after an interrupt or jumping to the guest code of a newly created thread. For both these cases, UNIX signals are used as a slower, but more generic alternative to the trampolines. The host calls the tkill syscall and copies the guest state in the ucontext provided by the signal handler, thus making the kernel do the actual work of jumping to the guest. This is much slower than the hand-made trampolines, of course, but should be executed less often.

Speaking of interrupts, these are implemented by calling `tkill` on the target running thread and jumping back to the host from the signal handler. If the target thread is not running guest code, it is not interrupted. Interrupting threads can introduce various races, especially when guest state is saved/restored in the trampolines, so spinlocks are also used to ensure proper synchronization.

For UNIX signals on Android, libc symbols are used directly, similar to Skyline. This is required because the Android runtime intercepts all calls to sigaction and doesn't pass the provided functions to the kernel, opting to call them after its own handlers. That presents problems as the ART sigsegv handler accesses `tpidr_el0`, which we have replaced with our own per thread structure.


src/android/app/src/main/jni/default_ini.h
```diff
@@ -153,6 +153,10 @@ use_multi_core =
 use_unsafe_extended_memory_layout =
 
 [Cpu]
+# Configures the preferred cpu backend for executing ARM instructions
```
FearlessTobi, line 156:
```suggestion
# Selects the preferred CPU backend for executing ARM instructions
```

src/android/app/src/main/jni/default_ini.h
```diff
@@ -153,6 +153,10 @@ use_multi_core =
...
+# 0 (default): Dynarmic, 1: Nce
```
FearlessTobi, line 157:
```suggestion
# 0 (default): Dynarmic, 1: NCE
```

src/common/settings.cpp
```diff
@@ -155,6 +156,22 @@ bool IsFastmemEnabled() {
     return true;
 }
 
+static bool is_nce_enabled = false;
+
+void SetNceEnabled(bool is_39bit) {
```
FearlessTobi, line 161: nit: we should probably use a consistent spellking between Nce and NCE

FearlessTobi (commented):
I cannot review the low level stuff, so just some nits

brujo5:
Sorry for the offtopic but its amazing now.

power consumption decrease a lot and the games start instantly and the use of HW resources is much lower

Screenshot_20231118-151436_Galería_1
Screenshot_20231118-144442_yuzu

2.8watts usage just like the switch, thanks to all the devs involved for making it possible 🥰

Carlosskyli (approved):

.codespellrc
```diff
@@ -3,4 +3,4 @@
 
 [codespell]
 skip = ./.git,./build,./dist,./Doxyfile,./externals,./LICENSES,./src/android/app/src/main/res
-ignore-words-list = aci,allright,ba,canonicalizations,deques,froms,hda,inout,lod,masia,nam,nax,nd,optin,pullrequests,pullrequest,te,transfered,unstall,uscaled,vas,zink
+ignore-words-list = aci,allright,ba,canonicalizations,deques,froms,hda,inout,lod,masia,nam,nax,nd,optin,pullrequests,pullrequest,te,transfered,unstall,uscaled,vas,zink,nce
```
liamwhite, line 6:
```suggestion
ignore-words-list = aci,allright,ba,canonicalizations,deques,froms,hda,inout,lod,masia,nam,nax,nce,nd,optin,pullrequests,pullrequest,te,transfered,unstall,uscaled,vas,zink
```

src/common/host_memory.cpp
```diff
@@ -353,6 +361,61 @@ class HostMemory::Impl {
...
+
+static void* ChooseVirtualBase(size_t virtual_size) {
+    constexpr uintptr_t Map39BitSize = (1ULL << 39);
+    constexpr uintptr_t Map36BitSize = (1ULL << 36);
+
+    // Seed the MT with some initial strong randomness.
+    //
+    // This is not a cryptographic application, we just want something more
+    // random than the current time.
+    std::mt19937_64 rng(GetRandomU64());
```
liamwhite, line 380: In retrospect, it's probably okay to seed this with the current time

src/common/host_memory.cpp
```diff
@@ -353,6 +361,61 @@ class HostMemory::Impl {
...
+    }
+
+    return MAP_FAILED;
+}
+#else
+static void* ChooseVirtualBase(size_t virtual_size) {
+    return mmap(nullptr, virtual_size, PROT_READ | PROT_WRITE,
+                MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE, -1, 0);
+}
+#endif
```
liamwhite, line 417: The top section has more space than this
```suggestion
#else

static void* ChooseVirtualBase(size_t virtual_size) {
    return mmap(nullptr, virtual_size, PROT_READ | PROT_WRITE,
                MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE, -1, 0);
}

#endif
```

src/common/signal_chain.h
```diff
@@ -0,0 +1,19 @@
...
+
+#ifndef _WIN32
+
+#include <signal.h>
+
+namespace Common {
+
+// Android's ART overrides sigaction with its own wrapper. This is problematic for SIGSEGV
+// in particular, because ARTs handler access TPIDR_EL0, so this extracts the libc version
+// and calls it directly.
```
liamwhite, line 14:
```suggestion
// Android's ART overrides sigaction with its own wrapper. This is problematic for SIGSEGV
// in particular, because ART's handler accesses tpidr_el0, which conflicts with NCE.
// This extracts the libc symbol and calls it directly.
```

src/core/CMakeLists.txt
```diff
@@ -921,6 +921,22 @@ if (ENABLE_WEB_SERVICE)
     target_link_libraries(core PRIVATE web_service)
 endif()
 
+if (ARCHITECTURE_arm64)
```
liamwhite, line 924: Linux-only for the time being
```suggestion
if (ARCHITECTURE_arm64 AND UNIX AND NOT APPLE)
```
(might need to be (UNIX OR ANDROID), not sure)

src/core/arm/nce/arm_nce.cpp
```diff
@@ -0,0 +1,395 @@
...
+// Verify assembly offsets.
+using NativeExecutionParameters = Kernel::KThread::NativeExecutionParameters;
+static_assert(offsetof(NativeExecutionParameters, native_context) == TpidrEl0NativeContext);
+static_assert(offsetof(NativeExecutionParameters, lock) == TpidrEl0Lock);
+static_assert(offsetof(NativeExecutionParameters, magic) == TpidrEl0TlsMagic);
+
+fpsimd_context* GetFloatingPointState(mcontext_t& host_ctx) {
+    _aarch64_ctx* header = reinterpret_cast<_aarch64_ctx*>(&host_ctx.__reserved);
+    while (header->magic != FPSIMD_MAGIC) {
+        header = reinterpret_cast<_aarch64_ctx*>((char*)header + header->size);
```
liamwhite, line 35:
```suggestion
        header = reinterpret_cast<_aarch64_ctx*>(reinterpret_cast<char*>(header) + header->size);
```

src/core/arm/nce/arm_nce.cpp
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    // We can't handle the access, so trigger an exception.
+    const bool is_prefetch_abort = host_ctx.pc == reinterpret_cast<u64>(info->si_addr);
+    guest_ctx->esr_el1.fetch_or(
+        static_cast<u64>(is_prefetch_abort ? HaltReason::PrefetchAbort : HaltReason::DataAbort));
+
+    // Forcibly mark the context as locked. We are still running.
+    // We may race with SignalInterrupt here:
+    // - If we lose the race, then SignalInterrupt will send us a signal which are masking,
```
liamwhite, line 127:
```suggestion
    // - If we lose the race, then SignalInterrupt will send us a signal we are masking,
```

src/core/arm/nce/arm_nce.cpp
```diff
@@ -0,0 +1,395 @@
+// SPDX-FileCopyrightText: Copyright 2023 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <cinttypes>
+#include <memory>
+
+#include "common/scope_exit.h"
```
liamwhite, line 7:
```suggestion
```

src/core/arm/nce/patch.cpp
```diff
@@ -0,0 +1,472 @@
...
+        oaknut::CodeGenerator rc{m_patch_instructions.data() + rel.patch_offset / sizeof(u32)};
+        rc.dx(RebasePc(rel.module_offset));
+    }
+    for (const Trampoline& rel : m_trampolines) {
+        out_trampolines->insert({RebasePc(rel.module_offset), RebasePatch(rel.patch_offset)});
+    }
+
+    // Cortex-A57 seems to treat all exclusives as ordered, but newer processors do not.
+    // Convert to ordered to preserve this assumption
+    for (u32 i = 0; i < static_cast<u32>(text_words.size()); i++) {
```
liamwhite, line 165: We probably want to build this list in the first step to take better advantage of data locality

src/core/arm/nce/patch.cpp
```diff
@@ -0,0 +1,472 @@
...
+}
+
+size_t Patcher::SectionSize() const noexcept {
+    return Common::AlignUp(m_patch_instructions.size() * sizeof(u32), Core::Memory::YUZU_PAGESIZE);
+}
+
+void Patcher::WriteLoadContext() {
+    // This function was called, which modifies X30, so use that as a scratch register.
+    // SP contains the guest X30, so save our return X30 to SP + 8, since we have allocated 16 bytes
+    // of stack.
```
liamwhite, line 190:
```suggestion
    // SP contains the guest X30, so save return X30 to SP + 8, since we have allocated 16 bytes of
    // stack.
```
makes comment consistent with comment in WriteSaveContext

src/core/arm/nce/patch.cpp
```diff
@@ -0,0 +1,472 @@
...
+    c.STR(W0, X30, offsetof(GuestContext, nzcv));
+    c.LDR(X0, SP, POST_INDEXED, 16);
+
+    // Reload our return X30 from the stack, and return.
+    c.LDR(X30, SP, 8);
+    c.RET();
+}
+
+void Patcher::WriteSvcTrampoline(ModuleDestLabel module_dest, u32 svc_id) {
+    LOG_ERROR(Core_ARM, "Patching SVC {:#x} at {:#x}", svc_id, module_dest - 4);
```
liamwhite, line 259: unnecessary error log

src/core/arm/nce/patch.h
```diff
@@ -0,0 +1,107 @@
+// SPDX-FileCopyrightText: Copyright 2023 yuzu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <span>
+#include <unordered_map>
+#include <vector>
+
+#pragma clang diagnostic push
```
liamwhite, line 10: Convert these to `#pragma GCC diagnostic ...` for GCC (clang has GCC diagnostic compatibility, but not the other way around)

src/core/arm/nce/patch.h
```diff
@@ -0,0 +1,107 @@
...
+#include "common/common_types.h"
+#include "core/hle/kernel/code_set.h"
+#include "core/hle/kernel/k_typed_address.h"
+#include "core/hle/kernel/physical_memory.h"
+
+#include <signal.h>
+
+namespace Core {
+struct GuestContext;
+}
```
liamwhite, line 25:
```suggestion
```

src/core/arm/nce/patch.h
```diff
@@ -0,0 +1,107 @@
...
+
+    void PatchText(const Kernel::PhysicalMemory& program_image,
+                   const Kernel::CodeSet::Segment& code);
+    void RelocateAndCopy(Common::ProcessAddress load_base, const Kernel::CodeSet::Segment& code,
+                         Kernel::PhysicalMemory& program_image, EntryTrampolines* out_trampolines);
+    size_t SectionSize() const noexcept;
+
+    [[nodiscard]] PatchMode Mode() const noexcept {
+        return mode;
+    }
```
liamwhite, line 52:
```suggestion
    size_t GetSectionSize() const noexcept;

    [[nodiscard]] PatchMode GetPatchMode() const noexcept {
        return mode;
    }
```

src/core/core.cpp
```diff
@@ -137,7 +137,6 @@ struct System::Impl {
 
     void Initialize(System& system) {
         device_memory = std::make_unique<Core::DeviceMemory>();
-
```
liamwhite, line 140: Unnecessary change

src/core/hle/kernel/k_page_table_base.cpp
```diff
@@ -88,6 +88,20 @@ Result FlushDataCache(AddressType addr, u64 size) {
...
+    Common::MemoryPermission perms{};
+    if (True(perm & KMemoryPermission::UserRead)) {
+        perms |= Common::MemoryPermission::Read;
+    }
+    if (True(perm & KMemoryPermission::UserWrite)) {
+        perms |= Common::MemoryPermission::Write;
+    }
+    if (True(perm & KMemoryPermission::UserExecute)) {
+        perms |= Common::MemoryPermission::Execute;
+    }
```
liamwhite, line 101: Only if ARCHITECTURE_arm64

src/core/hle/kernel/k_page_table_base.cpp
```diff
@@ -5658,8 +5675,18 @@ Result KPageTableBase::Operate(PageLinkedList* page_list, KProcessAddress virt_a
     }
     case OperationType::ChangePermissions:
     case OperationType::ChangePermissionsAndRefresh:
-    case OperationType::ChangePermissionsAndRefreshAndFlush:
+    case OperationType::ChangePermissionsAndRefreshAndFlush: {
+        const bool read = True(properties.perm & Kernel::KMemoryPermission::UserRead);
+        const bool write = True(properties.perm & Kernel::KMemoryPermission::UserWrite);
+        // todo: this doesn't really belong here and should go into m_memory to handle rasterizer
+        // access todo: ignore exec on non-direct-mapped case
```
liamwhite, line 5682: Fix comment wrapping. Also, worth resolving this in the first place...

src/core/hle/kernel/k_process.h
```diff
@@ -466,6 +468,14 @@ class KProcess final : public KAutoObjectWithSlabHeapAndContainer<KProcess, KWor
 
     static void Switch(KProcess* cur_process, KProcess* next_process);
 
+    std::unordered_map<u64, u64>& GetPostHandlers() noexcept {
+        return m_post_handlers;
+    }
+
+    KernelCore& GetKernel() noexcept {
+        return m_kernel;
+    }
```
liamwhite, line 477: Seems unused?

src/core/loader/nro.cpp
```diff
@@ -194,31 +199,79 @@ static bool LoadNroImpl(Kernel::KProcess& process, const std::vector<u8>& data)
 
     codeset.DataSegment().size += bss_size;
     program_image.resize(static_cast<u32>(program_image.size()) + bss_size);
+    size_t image_size = program_image.size();
+
+#ifdef ARCHITECTURE_arm64
+    const auto& code = codeset.CodeSegment();
+
+    // NROs are always 64-bit programs.
```
liamwhite, line 207:
```suggestion
    // NROs always have a 39-bit address space.
```

src/core/CMakeLists.txt
GPUCode: Maybe AND NOT WIN32 AND NOT APPLE?

src/common/host_memory.cpp
```diff
@@ -415,79 +478,100 @@ class HostMemory::Impl {
...
+        AdjustMap(&virtual_offset, &length);
+
+        int flags = PROT_NONE;
         if (read) {
             flags |= PROT_READ;
         }
         if (write) {
             flags |= PROT_WRITE;
         }
+#ifdef ARCHITECTURE_arm64
```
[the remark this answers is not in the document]
GPUCode: Yeah this is only really makes sense with NCE. At the moment its the responsibility of the caller to check for that, i.e KProcess::LoadModule [checks](https://github.com/GPUCode/yuzu/blob/a880f46b354fd6c1dbeb50017527f84d3b32b26e/src/core/hle/kernel/k_process.cpp#L1219) for this before adding the exec flag. We could definitely limit this further to the NCE setting though

src/core/CMakeLists.txt
liamwhite: how about this: in the root CMakeLists, add
```
if (ARCHITECTURE_arm64 AND (ANDROID OR ${CMAKE_SYSTEM_NAME} STREQUAL "Linux"))
    set(HAS_NCE ON)
    add_compile_definitions(HAS_NCE)
endif()
```
and then just use that for the Linux-specific cases where we are currently checking for ARCHITECTURE_arm64. I will have to do a mac build at some point to make sure.

amazingfate:
Hello, I'm trying to build your branch on ubuntu 22.04 jammy arm64. I got some error: https://paste.next.armbian.com/nuzusehebo.bash
And I have to remove `-Werror=conversion` in `src/core/CMakeLists.txt`.

amazingfate:
Hello, I just added cpu_backend configuration to qt: https://github.com/amazingfate/yuzu/commit/c0abb5b3c46ec4c1564252376b74271ce6f72f60
Feel free to use it.

src/android/app/src/main/jni/default_ini.h
```diff
@@ -153,6 +153,10 @@ use_multi_core =
 use_unsafe_extended_memory_layout =
 
 [Cpu]
+Selects the preferred CPU backend for executing ARM instructions
+# 0 (default): Dynarmic, 1: NCE
+cpu_backend =
+
```
t895, line 159: The config system will automatically add this key into config without adding this to the default_ini and I'm removing this in the config unification anyways

src/android/app/src/main/jni/native.cpp
```diff
@@ -707,6 +707,14 @@ jdoubleArray Java_org_yuzu_yuzu_1emu_NativeLibrary_getPerfStats(JNIEnv* env, jcl
     return j_stats;
 }
 
+jstring Java_org_yuzu_yuzu_1emu_NativeLibrary_getCpuBackend(JNIEnv* env, jclass clazz) {
+    if (Settings::IsNceEnabled()) {
+        return ToJString(env, "NCE");
+    }
+
+    return ToJString(env, "JIT");
+}
```
t895, line 716: You can get the index of the selected backend via the `int` property of the CPU_BACKEND enum that was added in `IntSetting`

src/android/app/src/main/java/org/yuzu/yuzu_emu/features/settings/model/view/SettingsItem.kt
```diff
@@ -77,6 +77,15 @@ abstract class SettingsItem(
...
             )
+            put(
+                SingleChoiceSetting(
+                    IntSetting.CPU_BACKEND,
+                    R.string.cpu_backend,
+                    0,
+                    R.array.cpuBackendNames,
+                    R.array.cpuBackendValues
+                )
+            )
```
t895, line 88: This should be changed based on the architecture. Just make another values array that's used only for x86_64.

src/android/app/src/main/res/values/strings.xml
```diff
@@ -184,6 +184,7 @@
     <string name="frame_limit_enable_description">Limits emulation speed to a specified percentage of normal speed.</string>
     <string name="frame_limit_slider">Limit speed percent</string>
     <string name="frame_limit_slider_description">Specifies the percentage to limit emulation speed. 100% is the normal speed. Values higher or lower will increase or decrease the speed limit.</string>
+    <string name="cpu_backend">CPU Backend</string>
```
t895, line 187: Please use "CPU backend"

src/android/app/src/main/jni/native.cpp
liamwhite: We don't want the selected backend, we want the active backend, which is not a property of the configuration alone

t895: Ah right, nevermind then

src/core/arm/nce/patch.cpp
GPUCode: You mean moving this to the start of RelocateAndCopy or doing it in PatchText? The former makes more sense to me

src/common/settings.cpp
GPUCode: Should I change this to NCE or leave it as Nce?

amazingfate:
Hello, I encountered an issue with mainline linux (ubuntu 22.04, mesa 23.2.1) on my snapdragon 870 tablet. After enabling NCE, game `The Legend of Zelda BOTW` is not rendered well.
With dynarmic as cpu backend:
截图 2023-11-22 23-12-51
With NCE as cpu backend:
截图 2023-11-22 23-16-23

src/core/hle/kernel/k_page_table_base.cpp
GPUCode: I have moved this to memory, please verify this is the intended solution

src/core/arm/nce/patch.cpp
GPUCode: Moved to the top of RelocateAndCopy

src/common/host_memory.cpp
liamwhite: It's a fair bit simpler and doesn't require a call to getrandom.

src/core/arm/nce/patch.cpp
liamwhite: I meant building a list in PatchText, and then applying it in RelocateAndCopy. That would mean that we only have to loop through all of the code instructions once; with it here, we have to loop through all of them twice.

src/common/host_memory.cpp
GPUCode: Changed to current time in nanoseconds

boludoz:
Unfortunately I tried the latest release, activated the option and the game is not launched unlike skyline, I promise logs soon. (Cuphead)

kelve00:
In The Legend of Zelda Link’s Awakening, when activating NCE, the screen flashes constantly, dynarmic rendering works well.

Yuzu build 140
Asus Rog Phone 6  - Android 13
Snapdragon 8+ Gen 1 
Driver turnip 24

src/core/arm/nce/patch.h
```diff
@@ -0,0 +1,98 @@
...
+    None,
+    PreText,  ///< Patch section is inserted before .text
+    PostData, ///< Patch section is inserted after .data
+};
+
+using ModuleTextAddress = u64;
+using PatchTextAddress = u64;
+using EntryTrampolines = std::unordered_map<ModuleTextAddress, PatchTextAddress>;
+
+class Patcher {
```
liamwhite, line 29: The name of this class is Patcher but the file name is patch.h/patch.cpp

src/common/host_memory.cpp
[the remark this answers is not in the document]
GPUCode: Made nce specific

src/core/arm/nce/patch.h
GPUCode: Would you prefer I rename the file or the class?

src/android/app/src/main/java/org/yuzu/yuzu_emu/features/settings/model/view/SettingsItem.kt
GPUCode: Could you list an example of how to use the cpuArchitecture enum in this context? Sorry I have zero knowledge of kotlin 😅

src/core/arm/nce/patch.h
liamwhite: file

PR #12081, opened 2023-11-19
Vulkan: Be more generous with pipeline workers for Android

FernandoS27:
Currently, we have only 1 pipeline worker in Android. Thus we have way too much stutter when making shaders. Since NCE, we can be a bit more generous. This also adds a check for drivers that have broken parallel shader compiling.


liamwhite (approved):

brujo5:
Could a setting be added to enable/disable multithread shaders compiling?

PR #12093, opened 2023-11-19
android: Use file picker intent for save exporter

t895:
The share sheet was causing problems for people who just wanted to back up their saves and didn't have a file manager to copy the zip with. Just saving to a selected location makes things easier for everyone.


liamwhite (approved):

t895:
Named it incorrectly initially, fixed now

PR #12094, opened 2023-11-19
gl_buffer_cache: Batch vertex/tfb buffer binding

ameerj:
Utilize `glBindVertexBuffers` and `glBindBuffersRange` to bind a batch of buffers, rather than binding them one buffer at a time.
This alleviates some API call overhead, and outweighs the benefits of `NV_vertex_buffer_unified_memory`, so the paths using this extension can be removed.

Also includes some other minor optimizations in the area.

`glBindBuffersRange` can be used to batch the UBO/SSBO bindings as well, but that change is a bit intrusive, so it remains as a todo.


gidoly:
Does this require test?

liamwhite (approved):

PR #12100, opened 2023-11-20
translations: Add android translations to transifex config

t895:
This should allow android strings to be automatically transferred between transifex and the app just like Qt


FearlessTobi (approved):
Thanks!

PR #12123, opened 2023-11-21
Explicit this

merryhime:
Implicit `this` is deprecated in C++20.


liamwhite (approved):

abouvier:
Is there a difference between `[this]` and `[=, this]`? Both are used in the code.

v1993:
Wouldn't `[this]` mean "capture the current object" while `[=, this]` be "capture the current object and all referenced local variables by value"?

PR #12130, opened 2023-11-22
Windows version to format script

flodavid:
Enable to check for style locally.
Adding it to a pre-commit hook might be useful


flodavid:
I could not make the target work before, but it is fine now. I guess it is not useful then

PR #12156, opened 2023-11-24
yuzu: Display firmware version

german77:
Since FW version make a difference in system applets. It's actually useful now to display the current firmware version.

image

### DIFF
--- a/src/yuzu/main.cpp
+++ b/src/yuzu/main.cpp
@@ -47,6 +47,7 @@
 #include "core/hle/service/am/applet_ae.h"
 #include "core/hle/service/am/applet_oe.h"
 #include "core/hle/service/am/applets/applets.h"
+#include "core/hle/service/set/set_sys.h"
 #include "yuzu/multiplayer/state.h"
 #include "yuzu/util/controller_navigation.h"
 
@@ -1047,7 +1048,12 @@ void GMainWindow::InitializeWidgets() {
         statusBar()->addPermanentWidget(label);
     }
 
-    // TODO (flTobi): Add the widget when multiplayer is fully implemented
+    firmware_label = new QLabel();
+    firmware_label->setObjectName(QStringLiteral("FirmwareLabel"));
+    firmware_label->setVisible(false);
+    firmware_label->setFocusPolicy(Qt::NoFocus);
+    statusBar()->addPermanentWidget(firmware_label);
+
     statusBar()->addPermanentWidget(multiplayer_state->GetStatusText(), 0);
     statusBar()->addPermanentWidget(multiplayer_state->GetStatusIcon(), 0);
 
@@ -2160,6 +2166,10 @@ void GMainWindow::OnEmulationStopped() {
     game_fps_label->setVisible(false);
     emu_frametime_label->setVisible(false);
     renderer_status_button->setEnabled(!UISettings::values.has_broken_vulkan);
+
+    if (!firmware_label->text().isEmpty()) {
+        firmware_label->setVisible(true);
+    }
 
     current_game_path.clear();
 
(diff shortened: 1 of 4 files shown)


src/yuzu/main.cpp
```diff
@@ -1047,7 +1048,12 @@ void GMainWindow::InitializeWidgets() {
         statusBar()->addPermanentWidget(label);
     }
 
-    // TODO (flTobi): Add the widget when multiplayer is fully implemented
+    firmware_label = new QLabel();
+    firmware_label->setObjectName(QStringLiteral("Firmwarelabel"));
```
liamwhite, line 1052:
```suggestion
    firmware_label->setObjectName(QStringLiteral("FirmwareLabel"));
```

src/yuzu/main.cpp
```diff
@@ -4845,6 +4855,27 @@ bool GMainWindow::CheckFirmwarePresence() {
...
+        LOG_INFO(Frontend, "Installed Firmware: No firmware available");
+        firmware_label->setText(tr("No firmware"));
+    }
+
+    const std::string display_version(firmware_data.display_version.data());
+    const std::string display_title(firmware_data.display_title.data());
+
+    LOG_INFO(Frontend, "Installed Firmware: {}", display_title);
+
+    firmware_label->setText(tr("Firmware %1").arg(QString::fromStdString(display_version)));
```
liamwhite, line 4875: Seems unnecessary to take up more space in the status bar for the word "Firmware"
```suggestion
    firmware_label->setText(QString::fromStdString(display_version));
```

src/yuzu/main.cpp
```diff
@@ -4845,6 +4855,27 @@ bool GMainWindow::CheckFirmwarePresence() {
...
+void GMainWindow::SetFirmwareVersion() {
+    Service::Set::FirmwareVersionFormat firmware_data{};
+    const auto result = Service::Set::GetFirmwareVersionImpl(
+        firmware_data, *system, Service::Set::GetFirmwareVersionType::Version2);
+
+    firmware_label->setVisible(true);
+
+    if (result.IsError()) {
+        LOG_INFO(Frontend, "Installed Firmware: No firmware available");
+        firmware_label->setText(tr("No firmware"));
```
liamwhite, line 4867: Maybe consider hiding the label if there is no firmware?

Also, you are missing a return statement
```suggestion
        LOG_INFO(Frontend, "No firmware available");
        firmware_label->setText(tr("No firmware"));
        return;
```

src/yuzu/main.cpp
```diff
@@ -4845,6 +4855,27 @@ bool GMainWindow::CheckFirmwarePresence() {
...
+    }
+
+    const std::string display_version(firmware_data.display_version.data());
+    const std::string display_title(firmware_data.display_title.data());
+
+    LOG_INFO(Frontend, "Installed Firmware: {}", display_title);
```
liamwhite, line 4873:
```suggestion
    LOG_INFO(Frontend, "Installed firmware: {}", display_title);
```

liamwhite (approved):

daisymlleung:
Just tested it .I'm using FW 17.0 but it shows 14.0

image

german77:
I'm loading data directly from firmware files. So are you sure about that? I find quite difficult to have the system reporting another firmware version. Unless you have more than one firmware on the registered folder.

daisymlleung:
> I'm loading data directly from firmware files. So are you sure about that? I find quite difficult to have the system reporting another firmware version. Unless you have more than one firmware on the registered folder.

I see. Wonder what is the difference towards the prod.keys and title.keys

german77:
Keys shouldn't break this either. Wrong keys should mean that you can't decrypt firmware. Which translates to no firmware installed. 

My suggestion is to delete the registered folder then dump keys and firmware using latest version of the tools

german77:
I intentionally hid it while playing. Our UI is too poluted and doesn't bring any benefit while playing.

StevensND:
> I intentionally hid it while playing. Our UI is too poluted and doesn't bring any benefit while playing.

I still think it's better to show it. Users who are just starting to use the emulator are not going to notice it before and they will have to stop emulation/reopen Yuzu to know it.

german77:
If you lack firmware on games where firmware actually matters it will crash. So you will need to restart yuzu anyways.

If on the future we find this label to be helpful while playing I can change this behavior.

PR #12195, opened 2023-11-27
Shaders: Stub Cbuf accesses for UBO bindings exceeding host device limits

ameerj:
This should resolve shader compilation failures on devices with low UBO limits


Squall-Leonhart:
what should I test this change against game wise?

KreuzerNL:
image
Fixes the regressions in SMW on Pascal 👍

PR #12201, opened 2023-11-27
Fixed controller applet crashing when on FW17+

Macj0rdan:
Fixed this by implementing `SetTouchscreenDimensions`


Kelebek1:
How often is it called? I'd prefer a Debug log rather than Info if possible.

Macj0rdan:
> How often is it called? I'd prefer a Debug log rather than Info if possible.

@german77 Said I should make it infos.
Narr what do you say?

german77:
Just once and only by the controller applet. This function destroys the current state on touch screen on HW. So it will never be spammed since that will break touch.

Kelebek1:
Ok that's fine then.

liamwhite (approved):

PR #12212, opened 2023-11-28
Update Qt version

YigaUpdate:


v1993:
Why? Are there significant issues with those older Qt versions? All this does is prevents building for those with less up-to-date Qt.

